Issue #255, opened 2014-08-16
Max and min coerce value


megawac:
[By spec](https://github.com/CrossEye/ramda/blob/master/test/test.math.js#L224) you guys coerce value types in max/min. That seems plain wrong. Types should be left the way they are and just produce the greatest value compared by `a > b`

``` js
max([{valueOf: constant(5)}, 4, 3, 2, 1]) // => 5
min(["1", 2, 3]) // => 1
```

PR #456, opened 2014-10-19
Abstract max and min, maxWith and minWith to helpers

megawac:
Obviously slight (probably inlined function call) perf hit but it's going to be much better than it was last revision and kills some redundancy. `min` and `max` may also perform better for some edge cases with a slow `valueOf`

### DIFF
--- a/ramda.js
+++ b/ramda.js
@@ -4724,16 +4724,16 @@
     }
 
     function createMaxMinWith(comparator) {
-        return function _maxWith(keyFn, list) {
+        return function(valueComputer, list) {
             if (!(list && list.length > 0)) {
                 return;
             }
             var idx = 0,
                 winner = list[idx],
-                computedWinner = keyFn(winner),
+                computedWinner = valueComputer(winner),
                 computedCurrent;
             while (++idx < list.length) {
-                computedCurrent = keyFn(list[idx]);
+                computedCurrent = valueComputer(list[idx]);
                 if (comparator(computedCurrent, computedWinner)) {
                     computedWinner = computedCurrent;
                     winner = list[idx];
@@ -4814,7 +4814,6 @@
      *      var a = {x: 1}, b = {x: 2}, c = {x: 3};
      *      R.minWith(cmp, [a, b, c]); //=> {x: 1}
      */
-    // TODO: combine this with maxWith?
     R.minWith = curry2(createMaxMinWith(lt));
 
 
--- a/bench/maxWith.bench.js
+++ b/bench/maxWith.bench.js
@@ -1,0 +1,30 @@
+var _ = require('lodash');
+var R = require('..');
+var maxWith = R.maxWith;
+
+var vals = _.chain(_.range(50, 500, 5))
+            .shuffle()
+            .map(function(val) {
+                return {key: val, val: String.fromCharCode(val)};
+            })
+            .value();
+var computer = R.prop('val');
+var maxVal = maxWith(computer);
+
+module.exports = {
+    name: 'maxWith',
+    tests: {
+        '_.max': function() {
+            _.max(vals, computer);
+        },
+        'maxWith(computer, nums)': function() {
+            maxWith(computer, vals);
+        },
+        'maxWith(computer)(vals)': function() {
+            maxWith(computer)(vals);
+        },
+        'maxVal(vals)': function() {
+            maxVal(vals);
+        }
+    }
+};


ramda.js
```diff
@@ -4705,6 +4707,42 @@
...
+                }
+            }
+            return winner;
+        };
+    }
+
+    function createMaxMinWith(comparator) {
+        return function(valueComputer, list) {
+            if (!(list && list.length > 0)) {
+                return;
```
jdalton: current version still needs an init value.

CrossEye: Wow, wonder how we dropped that.  Thanks for the heads-up!

We did discuss the coercing in #255, and decided we were happy with it.

jdalton: On the coerce note I was pointing out that maxBy/minBy don't coerce but min/max do.

CrossEye: Right,  and it's still a reasonable question.  Ramda's focus on values rather than references means that this coercion is reasonable.  But it doesn't particularly imply that it's required.

The main reason that I really do prefer coercion in this case is that if don't do it, then we lose the guarantee that

    max([a, b, ...]) == max([b, a, ...])

and that seems too much to give up.

`maxBy`/`minBy` are very different sorts of beasts.  They really do feel closer to `find`, and the use of references isn't so disconcerting.  Even here, I'm tempted to return a clone of the value, but that would simply be to prove a point; I've seen no practical reason for it.

And here you never could offer such an invariant as the above.

jdalton: Hmm I've not thought of `maxBy/minBy` as find. I thought the callback just allowed customizing the criteria by which something is sorted so a modifier on `max/min`.

CrossEye: Well, I guess that's the distinction I was making.  `max`/`min` in our world are ideally 

    :: [Number] -> Number

But we relax it a bit to allow

    :: [CoerceableToNumber] -> Number

But `maxBy`/`minBy` are

    :: (a -> Number) -> [a] -> a

Your description for `max`/`min` sound more like

    :: [CoerceableToNumber] -> CoerceableToNumber

While we could certainly support that, I prefer a `[Number] -> Number` signature that happens to be slightly relaxed.

jdalton: > Your description for max/min sound more like

I not describing `max/min` here I'm talking about `maxBy`/`minBy`.
I just think it odd that `maxBy/minBy` have entirely different behavior than `min/max`.

I would expect passing `maxBy/minBy` and identity function to have the same output as `max/min`.
Maybe that's complicated here because `max/min` return coerced values, I donno.

CrossEye: It does bother me a bit, too.  I simply don't see any good alternatives.  I think the signatures of `max` and of `maxBy` are correct.  But this means that `max` cannot be implemented as `maxBy(identity)` (which would be `[a] -> a` instead of `[Number] -> Number` (or our slightly relaxed version of that.)

I agree it does feel odd.  But anything that looks to fix it would likely cause something that seems to me much worse:

    max([a, b, ...]) !== max([b, a, ...); for some a, b

At that point this is no longer a number function and simply a partial ordering one.

jdalton: Hmm what if both `max` & `maxBy` had initial values of `-Infinity` and compared values were coerced to numbers that might solve most of the consistency gaps. Since the signature is arrays of numbers anything else could be outside the realm of documented support.

CrossEye: Note that `maxBy` does not share that property.  It's quite possible that 

    maxBy(myFunc, [a, b, ...]) != maxBy(myFunc, [b, a, ...])

And this doesn't bother me at all.

Our algorithm is functionally equivalent to your suggestion.  The problem is that `max` and `maxBy` have two fundamentally different concerns.  One has to do with values, the other with structures which can -- just for comparison purposes -- be represented by a value.  If we wanted to return that value for `maxBy`, we'd have no issue, but we want to return the original structure.

    var grades = [{ltr: 'B+', nbr: 87}, {ltr: 'B+', nbr: 89}, {ltr: 'C', nbr: 74}];
    maxBy(prop('nbr'), grades); //=> {ltr: 'B+', nbr: 89}

I think any of Ramda's `xxxBy` functions can be thought generally of projecting their data before processing it, but then, of course, restoring it before returning it to the user.  Since most projections lose information, there is no issue when things that otherwise wouldn't be equal end up equal.  That's just part of the process.

That there's some projection in `max` (because of the coercion) would be more troubling, but it comes out in the wash, since we end up returning the same numeric value regardless of which original value was coerced into it.

jdalton: Ok cool. Different strokes for different folks.

I started looking into this because of the Ramda benchmark comparing its `max` to `_.max`. However on further digging it appears to be comparing apples to oranges as `_.max`+`identity` is equiv to `_.max` and Ramda's is not.

Issue #512, opened 2014-11-11
a promise-aware version of compose


yuri:
Would Ramda benefit from a promise-aware version of R.compose()? Basically, just like R.compose(), but with automatic unwrapping of promises, so that you could do things like this:

``` javascript
    var getExifDate = promiseAwareCompose(
      R.prop('DateTimeOriginal'),
      R.prop('exif'),
      getExifData
    );
```

even where getExifData() is asynchronous and returns a promise for an array rather than the actual array.

I have an implementation now is actually backwards compatible with R.compose for non-promises, so it could in theory be used as a replacement for R.compose. 

Let me know if you might be interested in something like this and I can prepare a pull request.

PR #524, opened 2014-11-15
Adding promise-aware function composition.

yuri:
Compose would now handle promises returned by any of the composed functions. The resulting function
will evaluate to a promise if any one of the composed functions returns a promise. If all of the composed functions return something other than a promise, then the behaviour of compose() is the same as it was before. In this sense the change is fully backwards compatible.

I added one new test and modified two tests to test promises. I added Q to dev dependencies to test promise support. However, Q is not necessary at run time, and this implementation should work with any promise implementation.

See https://github.com/ramda/ramda/issues/512 for discussion.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "grunt-saucelabs": "^8.3.1",
     "lodash": "latest",
     "orchestrate": "~0.3.4",
+    "q": "^1.1.1",
     "testem": "^0.6.18",
     "uglify-js": "2.4.x",
     "xyz": "0.4.x"
--- a/ramda.js
+++ b/ramda.js
@@ -1316,6 +1316,73 @@
 
 
     /**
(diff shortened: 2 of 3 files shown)


dannyko:
can this be compared to using generators in ES6? e.g. http://taskjs.org/

yuri:
@dannyko: I am not quite sure in what way taskjs is similar. It seems to try to fix a rather different problem. But in terms of using generators in an "await"-like manner, R.pipeP should work fine for that. I mean, you should be able to just "yield" the return value of a function that was composed using R.pipeP.

dannyko:
thanks! to me, they are both similar in that they both try to make it easier to work with promises, in particular to create a "chain" of evaluations that blocks on promises.Perhaps I will better appreciate the differences once I try using them at some point :)

PR #574, opened 2014-11-23
Some handy aliases for array methods

anton-rudeshko:


danny-andrews:
This thread made me smile as I am definitely in the anti-aliasing camp. ;) I think that one intuitively-named function is better than two functions (one intuitively-named and one esoterically-named). I think @anton-rudeshko's `head`,`tail`, `prepend` vs `car`, `cdr`, `cons` example is a great case-study. People familiar with with `car`, `cdr`, `cons` variants are likely to understand what `head`,`tail`, `prepend` means without much effort, but the reverse is not true. As API designers, we should always try to make things the most understandable for everyone, rather than trying to cater to folks from in a particular domain.

I also found it funny that there were two comments that followed the, "I agree with removing aliases, **except for the one(s) that I really like!**" pattern:

> foldl and foldr are fine imo. each for forEach is also fine. The rest I'm anti-alias for

> The only alias I really think we should keep is prop/get

Good show!

PR #619, opened 2014-11-30
remove R.reduce and R.reduceRight

davidchambers:
Many found this [parable](https://github.com/ramda/ramda/pull/574#issuecomment-64124850) persuasive. There's support for removing the remaining :alien:s.

I set to work on a patch to _remove_ `R.foldl` and `R.foldr`, but as I worked my way through the various files I became convinced that it's `R.reduce` and `R.reduceRight` that must go. Here are my reasons:
- `foldl` and `foldr` are well liked. Several people have these in their shortlists of aliases to keep.
- `foldl` and `foldr` are more consistent with our ad hoc naming scheme. We have `lPartial` and `rPartial`, for example, rather than `partial` and `partialRight`.
- We have a function named `unfoldr`. It's inconsistent to have `reduceRight` and `unfoldr`, and `unreduceRight` doesn't exactly roll off the tongue.

I'd love to merge this change and rebase #618 on top of it. The intersection of two ugly things (:alien:s and `.idx` functions) is particularly confusing when every function lives in its own file.

### DIFF
--- a/ramda.js
+++ b/ramda.js
@@ -652,7 +652,7 @@
      *      var add = function(a, b) { return a + b; };
      *      // Adds any number of arguments together
      *      var addAll = function() {
-     *        return R.reduce(add, 0, arguments);
+     *        return R.foldl(add, 0, arguments);
      *      };
      *
      *      // Basic example
@@ -1612,7 +1612,7 @@
         var cache = {};
         return function() {
             if (!arguments.length) {return;}
-            var position = reduce(function(cache, arg) {
+            var position = foldl(function(cache, arg) {
                 return cache[arg] || (cache[arg] = {});
             }, cache, _slice(arguments, 0, arguments.length - 1));
             var arg = arguments[arguments.length - 1];
@@ -1796,7 +1796,7 @@
     // list parameter off.  For instance:
     //
     //     // skip third parameter
-    //     var checkAllPredicates = reduce(andFn, alwaysTrue);
+    //     var checkAllPredicates = foldl(andFn, alwaysTrue);
     //     // ... given suitable definitions of odd, lt20, gt5
     //     var test = checkAllPredicates([odd, lt20, gt5]);
     //     // test(7) => true, test(9) => true, test(10) => false,
@@ -1811,7 +1811,7 @@
      *
      * The iterator function receives two values: *(acc, value)*
      *
-     * Note: `R.reduce` does not skip deleted or unassigned indices (sparse arrays), unlike
+     * Note: `R.foldl` does not skip deleted or unassigned indices (sparse arrays), unlike
      * the native `Array.prototype.reduce` method. For more details on this behavior, see:
      * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce#Description
      *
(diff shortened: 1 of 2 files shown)


Dakuan:
Missing out on `reduce` is a real shame, its mentioned in just about every introduction to FP ever :cry:

buzzdecafe:
i agree, i think these aliases are worth keeping, esp. if I ever get back to transducer integration. 

:+1: for `reduce/reduceRight`

jschomay:
I'm for internal consistency, but I dont see why you can't leave the aliases in there for compatibility.  Our company has completely bought in to ramda, and after updating, our code and tests are throwing errors left and right everywhere we used `reduce`.  Why should a philosophical decision that depreciates common aliases break our software (even if the offending principle is sound)?

semmel:
:question: Where does this come from? I don't mind cleaning up the Ramda api (e.g. all this `.idx`stuff) but this name change of a *core function* seems so *arbitrary* to me. I fear in a future version of Ramda there comes someone else who deprecates `fold` and re-introduces `reduce` again. What a nightmare! :fearful:

I came from a strong JavaScript and C++ background to enjoy doing FP with `Ramda v. 0.8` in our latest web project. Using `reduce` feels quite naturally for me.

TheLudd:
Agreeing with @semmel and @jschomay but I also think this change comes a bit fast. Even though ramda is < v 1.0.0 and no contract is broken by this change, it would have been nice with deprecation warnings so that people could adjust instead of running in to a wall of errors when upgrading. Just a thought for the future.

/Captain Hindshight

CrossEye:
This was definitely too hasty.  The parable was quite convincing.  But I'm afraid that we will need to find a way to support aliases.  I had said that this was the only alias pair I wanted to keep (https://github.com/ramda/ramda/pull/574#issuecomment-64142350) but that for consistency we should probably remove even this one.  I didn't know which one I'd prefer to keep.  I think we chose wrong!

But as I think about it again, I really would like for our dynamic dispatching to grow to support more compatible libraries.  It would be great to be able to work more closely with tools like Highland, RxJs, Bacon, and possibly promise libraries (less likely as their terminology is quite different), as well as the various immutable collections.  In order for that to work, we will probably want to have more, not fewer aliases.  (Of course our current dynamic dispatch doesn't work with aliases in any case, so that would have to be straightened out first.)

But I'm all for putting this one back immediately and pushing out 0.9.1 with this and any doc changes done so far.

buzzdecafe:
0.9.1 is already out, but i also support aliasing for `reduce*` and publishing 0.9.2 ASAP

buzzdecafe:
#769

jschomay:
Thank you! :)

buzzdecafe:
i think we will have to put this in 0.10 because some other changes have snuck in there before it. But we'll be doing that soon.

Aourin:
in favor of alias.  Plus, reading the homepage 
"While Ramda includes all of the favorite list-manipulation functions you expect, e.g. map, filter, reduce, find, and so forth, Ramda is also significantly different from libraries like Underscore and Lo-Dash."

I assumed it was still in there, but I guess with such quick iterations, it isn't easy to keep it docs up to date.

PR #646, opened 2014-12-05
refactor example test runner to separate out doc parsing

kedashoe:
Resubmitting after #618

Only real thing of note I think is that `dox` can no longer determine function names from the comments/source code, as they are no longer there. Easy to get around by using file names. And actually as I'm writing this comment it occurred to me `scripts/build` must do something similar and I see

    var filenameToIdentifier =
        R.pipe(R.split('/'),
               R.last,
               R.split('.'),
               R.head);

. My version was

`path.match(/(\w+)\.js$/)[1]`

ha! I still have a long way to go to get used to this functional stuff :) I was thinking it might make sense to add the function name to all the `@func` tags we have, but if the build script uses the filenames maybe this is not necessary. Also should we add `scripts/common` or what have you for shared functionality like this?

### DIFF
--- a/scripts/doc-parser
+++ b/scripts/doc-parser
@@ -1,0 +1,127 @@
+#!/usr/bin/env node
+
+'use strict';
+
+var R = require('..');
+var fs = require('fs');
+var dox = require('dox');
+var normalize = require('path').normalize;
+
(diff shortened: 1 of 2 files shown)


davidchambers:
I've opened a competing pull request: #735.

kedashoe:
Sounds good, closed via #735

PR #656, opened 2014-12-06
foldObj proposal redux

buzzdecafe:
see #625


buzzdecafe:
clearly no momentum for this, so mothballing it

PR #698, opened 2014-12-19
use JSHint directive to avoid puzzling use of void operator

davidchambers:
A couple of people have asked why we use `void` here. Let's avoid `void` to avoid confusion.


CrossEye:
See https://github.com/ramda/ramda/pull/749#issuecomment-71146470

PR #709, opened 2015-01-01
New Ramda site

saikobee:
### Questions
- How should categories work?

### Todo
- Add favicon
- Squash garbage commits
- Better mobile view
- Stop stripping leading whitespace from code examples
  - This means we need to clean up the `@examples` in JSDoc comments


buzzdecafe:
this is very impressive

CrossEye:
I haven't had time to look at all the code, but the output is phenomenal!

CrossEye:
In a separate discussion, @davidchambers suggested that there was perhaps too much purple and that the code blocks might be better grey.  After seeing this on a desktop, I agree.  The purple is a bit overwhelming.  My color sense is never to be relied on.  But perhaps something like #555 would do, or #404040 if we want greater contrast.

saikobee:
I tried out some of the suggestions made and optimized the page further. I added a few open questions to the issue description.

Check out here: http://saikobee.github.io/ramda/docs/#

CrossEye:
This is looking amazing!

Two additional thoughts:

  - It would be good to include a link to Ramda itself, so that users can try out functions in the console from the doc page.

  - My thought about the Categories is that what's there now is great, but a nice addition would be some sort of selector (multi-select dropdown, checkbox set, on-off buttons) that allowed you to filter the function list (and maybe the content area as well) to a particular category or group of categories.  It would default to "All", but could be user-chosen.  A drop-down would fit well under the Search box, but another control might have to go elsewhere.

  Possibly this would work by pre-filling the search box with, e.g., "[cat=List,Object]", which the user might further refine with a search term such as "map".

saikobee:
@CrossEye 

- I added embedding Ramda in the page to the TODO list. I wouldn't be opposed to a brief link or two at the top, but I think we should keep the docs page lean and put other info on the home page.
- I think perhaps you overestimate how involved people are going to get with searching. With only around a hundred items, I can't imagine anyone using a complicated query. I experimented with filtering a little bit and didn't find it that useful. I think a decent compromise could be simply putting headings in the table of contents and having functions fall in sections. But even then, the distinction can feel arbitrary at times. I'd wager I've been helped as many times as confused by looking for something in the wrong section of the Lo-Dash documentation.

  I guess this is all to say that I think a dropdown seems like the most obvious and useful filter, if we're to even have one.

On another note, I think it could make sense for the home page to not simply insert the README.md. I think README.md should have more useful information for contributors, but that sort of thing should be omitted from the library home page. That being said, I think the README could be trimmed down quite a bit, too.

donnut:
@saikobee, this looks very good. Nice job. 
Maybe a bit of nitpicking from my side:
- the 'dropdown' below the function signature has a triangle pointing down. When I click on it, it doesn't open. I have to click on the signature itself. 
- I would love to have a link to the source code of a specific function, but that maybe a very personal wish
- It would be nice to have the categories back in some form. Ramda is extending fast and I haven't been able to keep up with all new functions. When I can narrow down the list to say object related functions, I have a better overview.
- The color of the argument names in the Parameters is different form the ones in Return (eg. fns and vs)

naamloos
Thanks!

saikobee:
@donnut I've added the first and last points to the TODO. Thanks! I'm currently leaning towards a dropdown for filtering to only a specific category.

As for the source code link, maybe not immediately. I currently don't have it generating a page with the source in it yet.

raine:
One thing I find essential is to be able to see all functions in a specific category, similarly to how the current docs behave. What if the category labels were links that would filter search by functions of that category?

Also, I dislike Menlo. It doesn't look very good in large font sizes.

Good stuff, Brian.

saikobee:
@raine I tried that before but didn't like complicating the search field with extra syntax. I think combining that with a dropdown is worth trying.

I felt like it was good to keep the card headings in a monospace font to be consistent with other occurrences of the word. As for the merits of Menlo... well at least it's not set to Courier in the CSS, right?

saikobee:
I pushed up a new commit and published the gh-pages for a build without Knockout now. Should be noticeably snappier to load the sidebar, especially on a slower connection.

buzzdecafe:
> It would be good to include a link to Ramda itself, so that users can try out functions in the console from the doc page.

or add the repl to the docs somehow: http://ramdajs.com/repl/

CrossEye:
I think I misspoke.  When I mentioned linking to Ramda, I meant only what I see you've already done, embedding Ramda on the page.

Including our ReadMe in the docs was mostly happenstance from the JSDoc template we were using.  I don't see any great reason to include it at all, although I agree that it could use some love regardless.  Our main page on the documentation should probably be handcrafted for that purpose.

I was thinking the same thing that @buzzdecafe suggested, that linking to the newly-built REPL would be useful.  But I was thinking that such could be a later stage of the process, because I was imagining something perhaps interesting: adding links to the REPL for each example in the docs.  I would want that automated, of course, and it could easily be added client-side after load.  A little Run icon would load the source from the example in the REPL.  But that's for later, I believe.  It shouldn't distract from the excellent progress you're making.

fyyyyy:
Nice, i would like to click on the category, e.g. "List", and this would add the text "List" to the search field. The search filed would support Categories as text input. This way, theres no extra interface required.
ramda list

saikobee:
@fyyyyy What if you wanted to filter by name and category? The filter field would need to support something like `list foo` to look for all list functions that have `foo` in them. The nice thing about having the dropdown is it keeps the filter field simple and makes it easier to notice you can filter by categories too.

buzzdecafe:
i'd be happy to merge this as-is, and implement category filtering in a subsequent PR.

saikobee:
@buzzdecafe Good point. Probably best to not let this one PR grow too big. I'd be happy to squash these commits down into something more reasonable and then move current concerns to new issues.

buzzdecafe:
@saikobee yes, pls squash; if @CrossEye and @davidchambers are good with this, I say let's proceed. we can gild the lily later

davidchambers:
:deciduous_tree:

CrossEye:
:herb:

saikobee:
Is this looking sufficiently squashed to y'all?

Also I suppose it's worth mentioning that once this gets merged you'll want to run `grunt publish-gh-pages` to build and deploy the docs to http://ramda.github.io/ramda. And then obviously move over the CNAME file from *ramdocs* and deprecate that repository.

CrossEye:
I think this is sufficiently squashed.

I'm merging now.  But getting this onto the site will have to wait until tomorrow.

saikobee:
Fantastic!

davidchambers:
[ramdajs.com][1] shouldn't point to the new docs until we release 0.9.0, unless we decide the documentation should reference the code in master rather than the latest published version.


[1]: http://ramdajs.com/

buzzdecafe:
i agree with @davidchambers we should wait for 0.9.0 to move the pointer. we should also make an effort to clean up remaining 0.9.0 issues.

Most importantly, :clap: for @saikobee 's great work on this.

CrossEye:
Indeed.  :clap: :fireworks:

saikobee:
Woot! I created new issues for the outstanding issues from this PR.

PR #716, opened 2015-01-08
Major readme update, intro, philosophy, *squashed*, closes #711

### LINKED ISSUE #711: Explaining ramda
Actually i almost skipped this beautiful ramda before even trying, after reading these lines and still not understanding what this library is about

`Using this library should feel as much like using Javascript as possible. Of course it's functional Javascript, but we're not introducing lambda expressions in strings, we're not borrowing consed lists, we're not porting over all of the Clojure functions.`

`Our basic data structures will be normal Javascript objects, and our usual collections will be Javascript arrays. We will not try to reach the point where all the functions have only zero, one, or two arguments. We will certainly try to keep some of the normal features of Javascript that seem to be unusual in functional languages, including variable length function signatures and functions as objects with properties.`

To be fair, my attention span was short, but this would be true for most folks browsing hundreds of github libraries. I was looking at other FP libraries that explained themselves somewhat better to the common JS guy.

Thankfully i had seen http://fr.umio.us/why-ramda/ which did a better job of explaining to me what ramda actually is and convinced me to give ramda a shot.

`Ramda does offer a different style of coding, a style that's taken for granted in purely functional programming languages: Ramda makes it simple for you to build complex logic through functional composition. Note that any library with a compose function will allow you do functional composition; the real point here is: "makes it simple".`

I believe, the first few sentences of your readme should provide a stronger case for why to use ramda. Kind of the elevator pitch.

fyyyyy:
Squashed into single commit

Rewritten ramda introduction
Added What's Different section
Renamed project ramda to ramda
Threw out eweda section
Threw out structure section
Monospace for function names
fixed Lo-Dash spelling

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-Project Ramda
+Ramda
 =============
 
 A practical functional library for Javascript programmers.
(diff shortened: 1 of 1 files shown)


CrossEye:
LGTM.  Thanks for all your work on this.

fyyyyy:
:beer:

Issue #699, opened 2014-12-22
R.eqDeep


davidrusu:
I think it would be useful to have a eqDeep function. When comparing objects it would recursively check inner objects
A possible implementation is at [codepen](http://codepen.io/anon/pen/YPWbaY?editors=101) and in case the codepen goes down the important bits are:
``` javascript
function isSubset(a, b) {
  for (var key in a) {
    if (!(key in b && eqDeep(a[key], b[key]))) {
      return false;
    }
  }
  return true;
}

function eqDeep(a, b) {
  switch (type(a)) {
    case 'Array':
    case 'Object':
      return type(a) === type(b)
          && isSubset(a,b)
          && isSubset(b,a);
    default: 
      return a === b;
  }
};
```
This implementation duplicates effort since ```eqDeep(a,b) = eqDeep(b,a)```. This could be solved with some sort of memoization or as @davidchambers suggested [CANON](https://github.com/davidchambers/CANON).stringify (although we need to think about how to deal with function equality)
thoughts?

PR #718, opened 2015-01-11
added eqDeep

davidrusu:
Discussion at (#699)

### DIFF
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -4903,6 +4903,51 @@
         }, [], fns);
     };
 
+    // The algorithm used to handle cyclic structures is
+    // inspired by underscore's isEqual
+    // RegExp equality algorithm: http://stackoverflow.com/a/10776635
+    var _eqDeep = function _eqDeep(a, b, stackA, stackB) {
+        var typeA = type(a);
+        if (typeA !== type(b)) {
+            return false;
+        }
+        if (eq(a, b)) {
+            return true;
+        }
+        if (typeA == 'RegExp') {
+            // RegExp equality algorithm: http://stackoverflow.com/a/10776635
+            return a.source === b.source && a.global === b.global && a.ignoreCase === b.ignoreCase && a.multiline === b.multiline && a.sticky === b.sticky && a.unicode === b.unicode;
+        }
+        if (Object(a) === a) {
+            if (typeA === 'Date' && a.getTime() != b.getTime()) {
+                return false;
+            }
+            var keysA = keys(a);
+            if (keysA.length !== keys(b).length) {
+                return false;
+            }
+            var idx = stackA.length;
+            while (idx--) {
+                if (stackA[idx] === a) {
+                    return stackB[idx] === b;
+                }
+            }
+            stackA.push(a);
+            stackB.push(b);
+            idx = keysA.length;
+            while (idx--) {
+                var key = keysA[idx];
+                if (!has(key, b) || !_eqDeep(b[key], a[key], stackA, stackB)) {
+                    return false;
+                }
+            }
+            stackA.pop();
+            stackB.pop();
+            return true;
+        }
+        return false;
+    };
+
     /**
      * Assigns own enumerable properties of the other object to the destination
      * object preferring items in other.
@@ -5304,6 +5349,34 @@
      */
     var divide = op(function divide(a, b) {
         return a / b;
+    });
+
+    /**
+     * Performs a deep test on whether two items are equal.
+     * Equality implies the two items are semmatically equivalent.
+     * Cyclic structures are handled as expected
+     *
+     * @func
+     * @memberOf R
+     * @category Relation
+     * @sig a -> b -> Boolean
+     * @param {*} a
+     * @param {*} b
+     * @return {Boolean}
+     * @example
+     *
+     *      var o = {};
+     *      R.eqDeep(o, o); //=> true
+     *      R.eqDeep(o, {}); //=> true
+     *      R.eqDeep(1, 1); //=> true
+     *      R.eqDeep(1, '1'); //=> false
+     *
+     *      var a = {}; a.v = a;
+     *      var b = {}; b.v = b;
+     *      R.eqDeep(a, b); //=> true
+     */
+    var eqDeep = _curry2(function eqDeep(a, b) {
+        return _eqDeep(a, b, [], []);
     });
 
     /**
@@ -6121,6 +6194,7 @@
         dropWhile: dropWhile,
         empty: empty,
         eq: eq,
+        eqDeep: eqDeep,
         eqProps: eqProps,
         evolve: evolve,
         filter: filter,
--- a/src/internal/_eqDeep.js
+++ b/src/internal/_eqDeep.js
@@ -1,0 +1,59 @@
+var eq = require('../eq');
+var has = require('../has');
+var keys = require('../keys');
+var type = require('../type');
+
+// The algorithm used to handle cyclic structures is
+// inspired by underscore's isEqual
+module.exports = function _eqDeep(a, b, stackA, stackB) {
+    var typeA = type(a);
+    if (typeA !== type(b)) {
+        return false;
+    }
+
+    if (eq(a, b)) {
+        return true;
+    }
+
+    if (typeA == 'RegExp') {
+        // RegExp equality algorithm: http://stackoverflow.com/a/10776635
+        return (a.source === b.source) &&
+               (a.global === b.global) &&
+               (a.ignoreCase === b.ignoreCase) &&
+               (a.multiline === b.multiline) &&
+               (a.sticky === b.sticky) &&
+               (a.unicode === b.unicode);
+    }
+
+    if (Object(a) === a) {
+        if (typeA === 'Date' && a.getTime() != b.getTime()) {
+            return false;
+        }
+
+        var keysA = keys(a);
+        if (keysA.length !== keys(b).length) {
+            return false;
+        }
+
+        var idx = stackA.length;
+        while (idx--) {
+            if (stackA[idx] === a) {
+                return stackB[idx] === b;
+            }
+        }
+
+        stackA.push(a);
+        stackB.push(b);
+        idx = keysA.length;
+        while (idx--) {
+            var key = keysA[idx];
+            if (!has(key, b) || !_eqDeep(b[key], a[key], stackA, stackB)) {
+                return false;
+            }
+        }
+        stackA.pop();
+        stackB.pop();
+        return true;
+    }
+    return false;
+};
--- a/test/eqDeep.js
+++ b/test/eqDeep.js
@@ -1,0 +1,104 @@
+/* jshint typed: true */
+
+var assert = require('assert');
+
+var R = require('..');
+
+describe('eqDeep', function() {
+    var a = [];
+    var b = a;
+    it('tests for deep equality of its operands', function() {
+        assert.strictEqual(R.eqDeep(100, 100), true);
+        assert.strictEqual(R.eqDeep(100, '100'), false);
+        assert.strictEqual(R.eqDeep([], []), true);
+        assert.strictEqual(R.eqDeep(a, b), true);
+    });
+
+    it('handles objects', function() {
+        assert.strictEqual(R.eqDeep({}, {}), true);
+        assert.strictEqual(R.eqDeep({a:1, b:2}, {a:1, b:2}), true);
+        assert.strictEqual(R.eqDeep({a:2, b:3}, {b:3, a:2}), true);
+        assert.strictEqual(R.eqDeep({a:2, b:3}, {a:3, b:3}), false);
+        assert.strictEqual(R.eqDeep({a:2, b:3, c:1}, {a:2, b:3}), false);
+    });
+
+    var supportsSticky = false;
+    try { RegExp('', 'y'); supportsSticky = true; } catch (e) {}
+
+    var supportsUnicode = false;
+    try { RegExp('', 'u'); supportsUnicode = true; } catch (e) {}
+
+    it('handles regex', function() {
+        assert.strictEqual(R.eqDeep(/\s/, /\s/), true);
+        assert.strictEqual(R.eqDeep(/\s/, /\d/), false);
+        assert.strictEqual(R.eqDeep(/a/gi, /a/ig), true);
+        assert.strictEqual(R.eqDeep(/a/mgi, /a/img), true);
+        assert.strictEqual(R.eqDeep(/a/gi, /a/i), false);
+
+        if (supportsSticky) {
+            // assert.strictEqual(R.eqDeep(/\s/y, /\s/y), true);
+            // assert.strictEqual(R.eqDeep(/a/mygi, /a/imgy), true);
+        }
+
+        if (supportsUnicode) {
+            // assert.strictEqual(R.eqDeep(/\s/u, /\s/u), true);
+            // assert.strictEqual(R.eqDeep(/a/mugi, /a/imgu), true);
+        }
+    });
+
+    var listA = [1, 2, 3];
+    var listB = [1, 3, 2];
+    it('handles lists', function() {
+        assert.strictEqual(R.eqDeep([], {}), false);
+        assert.strictEqual(R.eqDeep(listA, listB), false);
+    });
+
+    var c = {}; c.v = c;
+    var d = {}; d.v = d;
+    var e = []; e.push(e);
+    var f = []; f.push(f);
+    var nestA = {a:[1, 2, {c:1}], b:1};
+    var nestB = {a:[1, 2, {c:1}], b:1};
+    var nestC = {a:[1, 2, {c:2}], b:1};
+    it('handles recursive data structures', function() {
+        assert.strictEqual(R.eqDeep(c, d), true);
+        assert.strictEqual(R.eqDeep(e, f), true);
+        assert.strictEqual(R.eqDeep(nestA, nestB), true);
+        assert.strictEqual(R.eqDeep(nestA, nestC), false);
+    });
+
+    var date1 = new Date(0);
+    var date2 = new Date(0);
+    var date3 = new Date(1);
+    var date4 = new Date(0);
+    date4.v = 'value';
+    it('handles dates', function() {
+        assert.strictEqual(R.eqDeep(date1, date2), true);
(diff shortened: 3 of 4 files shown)


src/internal/_eqDeep.js
```diff
@@ -0,0 +1,45 @@
+var eq = require('../eq');
+var keys = require('../keys');
+var type = require('../type');
+
+// The algorithm used to handle cyclic structures
+// is heavily inspired by underscore.js'
+// implementation
+module.exports = function _eqDeep(a, b, stackA, stackB) {
+    for (var idx = stackA.length - 1; idx >= 0; idx--) {
```
CrossEye: House style only uses `while` loops.  While this has never been a formal requirement of Ramda code, I think we're pretty loathe to change it.

Usually:

    var idx = -1;
    while (++idx < bound) { /* ... */ }

src/internal/_eqDeep.js
```diff
@@ -0,0 +1,45 @@
...
+    case 'Object':
+        if (keys(a).length !== keys(b).length ||
+            type(a) !== type(b)) {
+            result = false;
+        } else {
+            var keysA = keys(a);
+            var length = keysA.length - 1;
+            result = true;
+
+            while (result && length >= 0) {
```
CrossEye: Again here, we usually loop from 0 up (although this is much less of an issue I think than the no-for-loop one.)

davidrusu:
I've made the changes to the loops and added some tests that I forgot to include the first time.

CrossEye:
Still not quite our usual style, but LGTM.

Any other thoughts?  @buzzdecafe? @davidchambers?

dist/ramda.js
```diff
@@ -4440,6 +4440,43 @@
...
+        case 'Object':
+            if (keys(a).length !== keys(b).length || type(a) !== type(b)) {
+                result = false;
+            } else {
+                var keysA = keys(a);
+                idx = -1;
+                result = true;
+                while (result && ++idx < keysA.length) {
+                    var key = keysA[idx];
+                    if (!(key in b && _eqDeep(b[key], a[key], stackA, stackB))) {
```
megawac: should be hasOwnProperty not simple `in`

dist/ramda.js
```diff
@@ -4440,6 +4440,43 @@
...
+        stackA.push(a);
+        stackB.push(b);
+        var result;
+        switch (type(a)) {
+        case 'Array':
+        case 'Object':
+            if (keys(a).length !== keys(b).length || type(a) !== type(b)) {
+                result = false;
+            } else {
+                var keysA = keys(a);
```
megawac: Can shortcut by checking if `keysA.length === keys(b).length` which would also remove the need for the `in` check I mentioned

dist/ramda.js
```diff
@@ -4440,6 +4440,43 @@
...
+                idx = -1;
+                result = true;
+                while (result && ++idx < keysA.length) {
+                    var key = keysA[idx];
+                    if (!(key in b && _eqDeep(b[key], a[key], stackA, stackB))) {
+                        result = false;
```
megawac: just `return false`

dist/ramda.js
```diff
@@ -4440,6 +4440,43 @@
         }, [], fns);
     };
 
+    // The algorithm used to handle cyclic structures is heavily
+    // inspired by underscore.js' implementation of isEqual
```
davidchambers:
```diff
-    // inspired by underscore.js' implementation of isEqual
+    // inspired by Underscore's isEqual.
```

dist/ramda.js
```diff
@@ -4440,6 +4440,43 @@
...
+        while (--idx >= 0) {
+            if (stackA[idx] === a) {
+                return stackB[idx] === b;
+            }
+        }
+        stackA.push(a);
+        stackB.push(b);
+        var result;
+        switch (type(a)) {
+        case 'Array':
```
megawac: probably should also handle `Dates` and typed arrays

dist/ramda.js
```diff
@@ -4440,6 +4440,43 @@
         }, [], fns);
     };
 
+    // The algorithm used to handle cyclic structures is heavily
+    // inspired by underscore.js' implementation of isEqual
+    var _eqDeep = function _eqDeep(a, b, stackA, stackB) {
+        var idx = stackA.length;
```
megawac: shortcut exit `a === b`?

dist/ramda.js
```diff
@@ -4440,6 +4440,43 @@
...
+        while (--idx >= 0) {
```
davidchambers: Though it's overly tricky in my view, Ramda style is to use `while (idx--)` here, I believe. The postfix `--` makes it possible to rely on type coercion, avoiding the comparison operator.

test/eqDeep.js
```diff
@@ -0,0 +1,39 @@
...
+
+    it('is curried', function() {
+        var isA = R.eqDeep(a);
+        assert.strictEqual(isA([]), true);
+    });
+
+    it('throws if given no arguments', function() {
+        assert.throws(R.eqDeep, TypeError);
+    });
+});
```
buzzdecafe: i'd like to see tests that prove that logically equivalent objects are equal without regard to key order, e.g.:

```javascript
var abc = {a: 1, b: 2, c: 3};
var bac = {b: 2, a: 1, c: 3};
var acb = {a: 1, c: 3, b: 2};
// etc.
assert.strictEqual(R.eqDeep(abc, acb));
assert.strictEqual(R.eqDeep(abc, bac));
assert.strictEqual(R.eqDeep(bac, acb));
// etc.
```

dist/ramda.js
```diff
@@ -4440,6 +4440,43 @@
...
+            if (stackA[idx] === a) {
+                return stackB[idx] === b;
+            }
+        }
+        stackA.push(a);
+        stackB.push(b);
```
megawac: Also, this can be avoided for primitives (Dates, numbers, strings, etc)

test/eqDeep.js
CrossEye: > i'd like to see tests that prove that logically equivalent objects are equal without regard to key order.

I think this implementation does that.

dist/ramda.js
CrossEye: This might be a good time to talk about that style.  I agree that there's no good reason for `while (idx--)`.

Of course, I used to love for loops like this so, maybe my aesthetic taste is not to be trusted:

    for (var idx = arr.length; idx --> 0;) { /* ... */ }

"But it *looks* like an arrow, isn't that cool?"  :blush:

dist/ramda.js
CrossEye: > probably should also handle `Dates` and typed arrays

Are you thinking we should do so with the same implementation this uses for Objects and Arrays?  Obviously that wouldn't work for `Date`s, which have no enumerable properties.  But something like

    date1 - date2 === 0

or

    date1.getTime() === date2.getTime()

would also not capture these differences, if they're important:

    var date1 = new Date(), date2 = new Date();
    date1.color = 'red'; date2.color = 'blue';

Do you have an implementation in mind?

test/eqDeep.js
buzzdecafe: yes, but there are no tests to prove it.

CrossEye: D'oh!

dist/ramda.js
davidrusu: on line 4458 we do check ```keys(a).length !== keys(b).length``` but I'm not sure I understand why this would remove the need for the ```in``` check, can you expand on that

megawac: @davidrusu you shouldn't be calling `keys(a)` twice

That statement I made is slightly inaccurate as it won't handle a value on `a` being undefined while some other value on `b` is set (for instance `eq({a: undefined}, {b: 1})` :/

dist/ramda.js
buzzdecafe: :+1: @megawac

dist/ramda.js
davidrusu: How about
``` javascript
if (date1.getTime() !== date2.getTime()) {
    return false;
} // else continue testing them as objects

Comment (CrossEye) [inline dist/ramda.js]: Well, as I said, that would not capture any additional properties you assigned to those `Date` objects.  I just can't decide if that's an important enough scenario to worry about.

Comment (megawac) [inline dist/ramda.js]: I don't think thats an important concern Scott but you can support both in the object case by doing something like

```js
var value = a.valueOf();
if (isObject(value)) {
   return value === b.valueOf;
}
```

davidrusu: @CrossEye Sorry I wasn't very clear.

``` javascript
if (type(a) === 'Date' || type(b) === 'Object') {
  if (type(a) === 'Date' && a.getTime() !== b.getTime()) {
    return false;
  }
  // now we check it as if it were an 'Object' ie check all attributes
  for (var key in a) {
    if (!(b[key] && eqDeep(a[key], b[key])) {
      return false;
    }
  }
}
```
This would test any added attributes

dist/ramda.js
davidrusu: @davidchambers @CrossEye I'll switch it to ```while (idx--)``` for now

dist/ramda.js
CrossEye: @megawac:  :+1: 

I'm not sure it's particularly important either, but I like the fact that this might also catch some cases for Host or user types that also wouldn't be caught by the remainder of the algorithm.

davidrusu: @megawac I tried adding some tests for typed arrays but jshint was complaining. I opened pull request #723 with the fix

dist/ramda.js
davidrusu: Good idea, I'll use eq(a, b) incase #724 goes through

davidrusu:
Ok, I made the changes, It works with regexs, dates and typed arrays now.

davidchambers:
:deciduous_tree:

Let's merge this and publish 0.9.0. :)

test/eqDeep.js
```diff
@@ -0,0 +1,86 @@
...
+    var date3 = new Date(1);
+    var date4 = new Date(0);
+    date4.v = 'value';
+    it('handles dates', function() {
+        assert.strictEqual(R.eqDeep(date1, date2), true);
+        assert.strictEqual(R.eqDeep(date2, date3), false);
+        assert.strictEqual(R.eqDeep(date1, date4), false);
+    });
+
+    var typArr1 = new ArrayBuffer(10);
```
davidchambers: This will cause the tests to fail in old browsers, I believe.

/cc @buzzdecafe

test/eqDeep.js
```diff
@@ -0,0 +1,86 @@
+/*jshint typed: true */
```
davidchambers: Nit: Please add a space between `*` and `j`.

Also, now's a good time to remove the line you added to __.jshintrc__.

src/internal/_eqDeep.js
```diff
@@ -0,0 +1,64 @@
...
+               (a.ignoreCase === b.ignoreCase) &&
+               (a.multiline === b.multiline);
+    }
+
+    var eqDate = typeA == 'Date';
+    if (eqDate || typeA == 'Object' || typeA == 'Array' || typeA == 'ArrayBuffer' ||
+        typeA == 'Int16Buffer' || typeA == 'Int8Array' || typeA == 'Uint8Array' ||
+        typeA == 'Uint8ClampedArray' || typeA == 'Int16Array' || typeA == 'Uint16Array' ||
+        typeA == 'Int32Array' || typeA == 'Uint32Array' || typeA == 'Float32Array' ||
+        typeA == 'Float64Array') {
```
megawac: I'd probably use a switch here

```js
switch (typeA) {
case 'Object':
case 'Array':
case 'Date':
case 'ArrayBuffer':
case 'Int16Buffer':
case 'Int8Array':
case 'Uint8Array':
case 'Uint8ClampedArray':
case 'Int16Array':
case 'Uint16Array':
case 'Int32Array':
case 'Uint32Array':
case 'Float32Array':
case 'Float64Array':
    // fall through sick
}
```

megawac: Alternatively, `isObject(a) && isObject(b)`

src/internal/_eqDeep.js
```diff
@@ -0,0 +1,64 @@
...
+                return stackB[idx] === b;
+            }
+        }
+
+        stackA.push(a);
+        stackB.push(b);
+        idx = -1;
+        while (++idx < keysA.length) {
+            var key = keysA[idx];
+            if (!b.hasOwnProperty(key) || !_eqDeep(b[key], a[key], stackA, stackB)) {
```
megawac: `{hasOwnProperty: 'haha sucker'}`

src/internal/_eqDeep.js
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    var eqDate = typeA == 'Date';
+    if (eqDate || typeA == 'Object' || typeA == 'Array' || typeA == 'ArrayBuffer' ||
+        typeA == 'Int16Buffer' || typeA == 'Int8Array' || typeA == 'Uint8Array' ||
+        typeA == 'Uint8ClampedArray' || typeA == 'Int16Array' || typeA == 'Uint16Array' ||
+        typeA == 'Int32Array' || typeA == 'Uint32Array' || typeA == 'Float32Array' ||
+        typeA == 'Float64Array') {
+
+        if (eqDate && a.getTime() != b.getTime()) {
```
megawac: I would prefer the `valueOf` check myself but this is good enough for sure

src/internal/_eqDeep.js
```diff
@@ -0,0 +1,64 @@
...
+        while (idx--) {
+            if (stackA[idx] === a) {
+                return stackB[idx] === b;
+            }
+        }
+
+        stackA.push(a);
+        stackB.push(b);
+        idx = -1;
+        while (++idx < keysA.length) {
```
megawac: Nit: I would probably do a `while (idx--) {}` loop here

test/eqDeep.js
davidrusu: @davidchambers Would there be problems with wrapping that test case in an if?
``` javascript
if (ArrayBuffer) {
    // typed array tests
}
````

File: src/internal/_eqDeep.js (hunk 14 above)
Comment (davidrusu) [inline src/internal/_eqDeep.js]: That's much cleaner, I think just ```isObject(a)``` should suffice as we check that the types are equal at line 9.

File: src/internal/_eqDeep.js (hunk 15 above)
Comment (davidrusu) [inline src/internal/_eqDeep.js]: ahh, tricky

File: test/eqDeep.js (hunk 12 above)
Comment (davidchambers) [inline test/eqDeep.js]: You'd need something like this:

```javascript
if (typeof ArrayBuffer !== 'undefined' && typeof Int8Array !== 'undefined') {
    // ...
}
```

This seems like a reasonable approach.

src/internal/_eqDeep.js
davidchambers: Use `Object.prototype.hasOwnProperty.call(b, key)` or Ramda's `has` function.

davidrusu: @davidchambers hmm, something is weird with the ```has``` function
``` javascript
Object.prototype.hasOwnProperty([1], '0'); // => true
R.has([1], '0'); // => false
```
Is this expected behavior?

davidchambers: `R.has` takes the property name as its first argument.

src/internal/_eqDeep.js
```diff
@@ -0,0 +1,61 @@
...
+    if (eq(a, b)) {
+        return true;
+    }
+
+    if (typeA == 'RegExp') {
+        // RegExp equality algorithm: http://stackoverflow.com/a/10776635
+        return (a.source === b.source) &&
+               (a.global === b.global) &&
+               (a.ignoreCase === b.ignoreCase) &&
+               (a.multiline === b.multiline);
```
megawac: That algorithm is insufficient for the `unicode` flag

davidrusu: @megawac Good catch, it doesn't handle the ```sticky``` flag either.

test/eqDeep.js
```diff
@@ -0,0 +1,110 @@
...
+//            assert.strictEqual(R.eqDeep(/\s/y, /\s/y), true);
+//            assert.strictEqual(R.eqDeep(/a/mygi, /a/imgy), true);
+//        }
+//
+//        if (supportsUnicode) {
+//            assert.strictEqual(R.eqDeep(/\s/u, /\s/u), true);
+//            assert.strictEqual(R.eqDeep(/a/mugi, /a/imgu), true);
+//        }
+        // jscs:enable
+        /* jshint ignore: end */
```
davidrusu: jshint and jscs don't recognize the 'sticky' and 'unicode' flags yet, and the version of jscs that we are using doesn't support the disable flag

davidchambers: #739

CrossEye:
This sounds good to me.  Anyone have objections to merging as is?

buzzdecafe:
:cake:

test/eqDeep.js
```diff
@@ -0,0 +1,104 @@
...
+
+        if (supportsSticky) {
+            // assert.strictEqual(R.eqDeep(/\s/y, /\s/y), true);
+            // assert.strictEqual(R.eqDeep(/a/mygi, /a/imgy), true);
+        }
+
+        if (supportsUnicode) {
+            // assert.strictEqual(R.eqDeep(/\s/u, /\s/u), true);
+            // assert.strictEqual(R.eqDeep(/a/mugi, /a/imgu), true);
+        }
```
davidchambers: These assertions should either be uncommented or removed.

CrossEye: The idea was to leave them as reminders to attach them when we are properly able to run them: https://github.com/ramda/ramda/pull/718#commitcomment-9422717

Do you have serious reservations?  I would love it if we could get 0.9.0 out the door this weekend.  And I think this is the only major item left.

davidchambers: > The idea was to leave them as reminders to attach them when we are properly able to run them

I see. I had missed that comment.

> Do you have serious reservations?

No.

There are merge conflicts (likely just in __dist/ramda.js__). I'm happy to resolve the merge conflicts and push the updated commit to master, if you like.

davidchambers:
Could you resolve the merge conflicts, @davidrusu? I'll merge this once you have done so. :)

davidrusu:
@davidchambers sure

PR #721, opened 2015-01-12
build: include header in ramda.min.js

### LINKED ISSUE #676: ramda[.min].js should contain the version number in the header
When upgrading Ramda I've just realized that I couldn't see the version number right off the bat in the source file. Looks like I can query R.version in recent versions but still, it'd cleaner for the version number to be featured in a comment right at the beginning of the non-minified and minified files.

davidchambers:
Closes #676

See also #627/#633

### DIFF
--- a/scripts/header
+++ b/scripts/header
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+set -e
+
+cat <<END
+//  Ramda v$(node -p 'require("envvar").string("VERSION", require("./package.json").version)')
+//  https://github.com/ramda/ramda
+//  (c) 2013-$(git show -s --format=%ai | cut -d - -f 1) Scott Sauyet and Michael Hurley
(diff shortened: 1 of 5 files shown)


scripts/header
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+set -e
+
+cat <<END
+//  Ramda v$(node -p 'require("envvar").string("VERSION", require("./package.json").version)')
+//  https://github.com/ramda/ramda
+//  (c) 2013-$(git show -s --format=%ai | cut -d - -f 1) Scott Sauyet and Michael Hurley
+//  Ramda may be freely distributed under the MIT license.
+
+END
```
davidchambers: Sometimes a shell script is *exactly* the right tool for the job. :)

buzzdecafe:
:cake:

PR #723, opened 2015-01-12
Added 'typed' option to jshint

davidrusu:
I was trying to write some tests using typed arrays but jshint was complaining that it didn't recognize the typed array constructors.

See http://jshint.com/docs/options/#typed for the 'typed' option documentation

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,4 +1,5 @@
 {
+  "typed": true,
   "evil": true,
   "eqnull": true,
   "predef": ["beforeEach", "console", "define", "describe", "it", "module", "process", "require"],


davidchambers:
Ramda targets browsers which don't support typed arrays, so we *should* warn if typed arrays are used, no?

megawac:
Yeah, probably should be handled by a jshint hint `/*jshint typed: true */` in the corresponding tests

davidrusu:
Good catch, I've added the hint to my test. What's standard procedure for removing the changes from the  pull request?

CrossEye:
> What's standard procedure for removing the changes from the pull request?

I'm sorry.  Did you want me to simply revert this change?  Or was there more to it?

davidchambers:
I suggest removing this line from __.jshintrc__ in #718. That pull request can instead use a special comment such as the one suggested by @megawac.

PR #726, opened 2015-01-14
Revert "Added 'typed' option to jshint"

davidrusu:
reason: We want to warn about typed array usage since not all browsers we support have typed arrays implemented.


davidchambers:
Thanks!

Issue #724, opened 2015-01-13
Change R.eq to act like Object.is from ES6


davidrusu:
The implementation is here
http://wiki.ecmascript.org/doku.php?id=harmony:egal

What do you think?

CrossEye:
This makes sense to me.  Always hated ±0  and NaN exceptions.

davidchambers:
:thumbsup:

PR #730, opened 2015-01-16
made R.eq behave as Object.is from ES6

davidrusu:
Related to #724, almost a direct copy of https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is#Polyfill

### DIFF
--- a/src/eq.js
+++ b/src/eq.js
@@ -19,5 +19,13 @@ var _curry2 = require('./internal/_curry2');
  *      R.eq(o, {}); //=> false
  *      R.eq(1, 1); //=> true
  *      R.eq(1, '1'); //=> false
+ *      R.eq(0, -0); //=> false
+ *      R.eq(NaN, NaN); //=> true
  */
-module.exports = _curry2(function eq(a, b) { return a === b; });
+module.exports = _curry2(function eq(a, b) {
+    if (a === 0) {
+        return 1 / a === 1 / b;
+    } else {
+        return a === b || (a !== a && b !== b);
+    }
+});
--- a/test/eq.js
+++ b/test/eq.js
@@ -9,8 +9,13 @@ describe('eq', function() {
     it('tests for strict equality of its operands', function() {
         assert.strictEqual(R.eq(100, 100), true);
         assert.strictEqual(R.eq(100, '100'), false);
+        assert.strictEqual(R.eq('string', 'string'), true);
         assert.strictEqual(R.eq([], []), false);
         assert.strictEqual(R.eq(a, b), true);
+        assert.strictEqual(R.eq(0, -0), false);
+        assert.strictEqual(R.eq(NaN, NaN), true);
+        assert.strictEqual(R.eq(undefined, undefined), true);
+        assert.strictEqual(R.eq(null, undefined), false);
     });
 
     it('is curried', function() {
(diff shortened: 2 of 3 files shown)


src/eq.js
```diff
@@ -19,5 +19,20 @@ var _curry2 = require('./internal/_curry2');
...
+            return 1 / a === 1 / b;
+        }
+        if (a !== a) {
+            return b !== b;
+        }
+        return a === b;
+    } else {
+        return Object.is(a, b);
+    }
+});
```
davidchambers: I don't see much value in using `Object.is` when available. Could the body of the function be an expression?

```javascript
a === 0 && b === 0 ? 1 / a === 1 / b : a === b || a !== a && b !== b
```

davidrusu: Yup, it seems that they are equivalent, although I'm really not a big fan of the ternary operator, how about:
``` javascript
if (a === 0 && b === 0) {
    return 1 / a === 1 / b;
} else {
    return a === b || (a !== a && b != b);
}
```

davidchambers: Works for me.

CrossEye: When we first mentioned using `Object.is`, I passed around a somewhat different version of that ternary as a joke quiz: What does this function do?  Determines if two values are truly equal of course!  :smile: Gotta love JS.

src/eq.js
```diff
@@ -19,5 +19,13 @@ var _curry2 = require('./internal/_curry2');
  *      R.eq(o, {}); //=> false
  *      R.eq(1, 1); //=> true
  *      R.eq(1, '1'); //=> false
+ *      R.eq(0, -0); //=> false
+ *      R.eq(NaN, NaN); //=> true
  */
-module.exports = _curry2(function eq(a, b) { return a === b; });
+module.exports = _curry2(function eq(a, b) {
+    if (a === 0 && b === 0) {
```
megawac: nit: you only really need to check one or the other

src/eq.js
```diff
@@ -19,5 +19,13 @@ var _curry2 = require('./internal/_curry2');
...
+        return 1 / a === 1 / b;
+    } else {
+        return a === b || (a !== a && b != b);
```
megawac: strict for a and not strict for b? I'd go non strict for both here

davidrusu: Oops!

CrossEye:
@davidrusu:

I know that you've had more than your share of little nits to deal with in your recent check-ins, but do you want to fix these last ones so that we can merge this?

davidrusu:
Sorry, It's been a busy past couple of days in school. Thanks for putting up with my beginner mistakes, I'm quite new to js and contributing in general.

CrossEye:
No apologies necessary.  Thanks for your help.

This LGTM.  Anyone else have thoughts?

test/eq.js
```diff
@@ -9,8 +9,12 @@ describe('eq', function() {
     it('tests for strict equality of its operands', function() {
         assert.strictEqual(R.eq(100, 100), true);
         assert.strictEqual(R.eq(100, '100'), false);
+        assert.strictEqual(R.eq('string', 'string'), true);
         assert.strictEqual(R.eq([], []), false);
         assert.strictEqual(R.eq(a, b), true);
+        assert.strictEqual(R.eq(0, -0), false);
+        assert.strictEqual(R.eq(NaN, NaN), true);
+        assert.strictEqual(R.eq(undefined, undefined), true);
```
davidchambers: Let's add an assertion for `R.eq(null, undefined)`.

davidchambers:
:deciduous_tree:

test/eq.js
davidrusu: added in new version

Issue #732, opened 2015-01-17
Switch to the correct implementation of the `curry` function.


aaditmshah:
I recently wrote a question and an answer on StackOverflow illustrating how the current implementations of the `curry` function are wrong:

[How to correctly curry a function in JavaScript?](http://stackoverflow.com/q/27996544/783743)

Please switch to the correct implementation of the `curry` function.

davidchambers:
This seems great. Would you like to submit a pull request?

@buzzdecafe and @CrossEye likely chose the current behaviour for a reason. I wasn't on the scene back then, so I don't know the rationale.

buzzdecafe:
if we can get `project = compose(map, pick)` out of this, then I am all for it. One point where I disagree is what happens when you call a curried function with no arguments. IMO that's an error and should be handled as such.

aaditmshah:
@buzzdecafe There are three things you can do when a curried function is called with no arguments:

1. Return the curried function itself.
2. Throw a &ldquo;no arguments&rdquo; error.
3. Do nothing at all.

The first two cases are pessimistic. You are assuming that your end users might be stupid enough to call a curried function with no arguments:

1. In the first case you allow the end user to call the curried function with no arguments, but you return the curried function itself thereby circumventing the creation of a new curried function.
2. In the second you don't allow the end user to call the curried function with no arguments.

In both cases you have an initial check for no arguments. This check is not part of the core logic of the `curry` function itself, and every time an end user calls the curried function this check will occur. For a correct program (one which never calls a curried function with no arguments) it means that this check will always be `false`. Hence, it's just a waste of CPU cycles.

On the other hand, the third case is optimistic. It assumes that your end users aren't stupid enough to call a curried function with no arguments. Hence it doesn't penalize them with unnecessary &ldquo;no arguments&rdquo; checks. For a correct program (one which never calls a curried function with no arguments) this means that we can skip the useless check, resulting in faster code.

For an incorrect program (one which does call a curried function with no arguments) this means that the `curry` function will unnecessarily create a new curried function every time a curried function is called with no arguments. However, the program will still execute correctly, only slower. In my opinion, this is a reasonable tradeoff.

My mantra is &ldquo;if somebody is stupid enough to call a curried function with no arguments, then they deserve the performance hit. Others shouldn't needlessly be punished just because of a few stupid people.&rdquo;

[...]

PR #733, opened 2015-01-17
make R.always a curried binary function

davidchambers:
As suggested by @aaditmshah in https://github.com/ramda/ramda/issues/732#issuecomment-70383334

### DIFF
--- a/test/always.js
+++ b/test/always.js
@@ -6,19 +6,26 @@ var R = require('..');
 describe('always', function() {
     it('returns a function that returns the object initially supplied', function() {
         var theMeaning = R.always(42);
-        assert.strictEqual(theMeaning(), 42);
         assert.strictEqual(theMeaning(10), 42);
         assert.strictEqual(theMeaning(false), 42);
     });
 
     it('works with various types', function() {
-        assert.strictEqual(R.always(false)(), false);
-        assert.strictEqual(R.always('abc')(), 'abc');
-        assert.deepEqual(R.always({a: 1, b: 2})(), {a: 1, b: 2});
+        assert.strictEqual(R.always(false)('xxx'), false);
+        assert.strictEqual(R.always('abc')('xxx'), 'abc');
+        assert.deepEqual(R.always({a: 1, b: 2})('xxx'), {a: 1, b: 2});
         var obj = {a: 1, b: 2};
-        assert.strictEqual(R.always(obj)(), obj);
+        assert.strictEqual(R.always(obj)('xxx'), obj);
         var now = new Date(1776, 6, 4);
-        assert.deepEqual(R.always(now)(), new Date(1776, 6, 4));
-        assert.strictEqual(R.always()(), undefined);
+        assert.deepEqual(R.always(now)('xxx'), new Date(1776, 6, 4));
+    });
+
+    it('returns a unary function', function() {
+        assert.strictEqual(R.always(42).length, 1);
+    });
+
+    it('is binary', function() {
+        assert.strictEqual(R.always.length, 2);
+        assert.strictEqual(R.always(42, 'xxx'), 42);
     });
 });
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -470,26 +470,6 @@
     };
 
     /**
-     * Returns a function that always returns the given value.
-     *
-     * @func
-     * @memberOf R
-     * @category Function
-     * @sig a -> (* -> a)
-     * @param {*} val The value to wrap in a function
-     * @return {Function} A Function :: * -> val.
-     * @example
-     *
-     *      var t = R.always('Tee');
-     *      t(); //=> 'Tee'
-     */
-    var always = function always(val) {
-        return function () {
-            return val;
-        };
-    };
-
-    /**
      * Wraps a function of any arity (including nullary) in a function that accepts exactly `n`
      * parameters. Unlike `nAry`, which passes only `n` arguments to the wrapped function,
      * functions produced by `arity` will pass all provided arguments to the wrapped function.
@@ -1404,42 +1384,12 @@
     };
 
     /**
-     * A function that always returns `false`. Any passed in parameters are ignored.
-     *
-     * @func
-     * @memberOf R
-     * @category Function
-     * @sig * -> false
-     * @see R.always
-     * @return {Boolean} false
-     * @example
-     *
-     *      R.F(); //=> false
-     */
-    var F = always(false);
-
-    /**
      * @func
      * @memberOf R
      * @category Function
      * @see R.identity
      */
     var I = identity;
-
-    /**
-     * A function that always returns `true`. Any passed in parameters are ignored.
-     *
-     * @func
-     * @memberOf R
-     * @category Function
-     * @sig * -> true
-     * @see R.always
-     * @return {Boolean} `true`.
-     * @example
-     *
-     *      R.T(); //=> true
-     */
-    var T = always(true);
 
     var _append = function _append(el, list) {
         return _concat(list, [el]);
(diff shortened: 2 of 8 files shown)


buzzdecafe:
this means `always` cannot be called with no args.

davidchambers:
> this means `always` cannot be called with no args.

`R.always()`, do you mean? `R.always(undefined)` and `R.always(void 0)` are more explicit; I don't mind losing the shortcut.

If you're saying that the result of partially applying `R.always` cannot be called with no arguments you are correct. I don't see a problem with this, though.

buzzdecafe:
i mean `T = always(true)` is more convenient to call `T()` rather than `T(something)`. If I have to pass a value, why wouldn't I just use `I(true)`?

davidchambers:
If one wants the value `true` one can write `true` rather than `R.T()` or `R.I(true)`. `R.always(x)` is (only?) useful as an argument to a higher-order function such as `R.times`:

```javascript
> R.times(R.always('hi'), 5)
['hi', 'hi', 'hi', 'hi', 'hi']
```

Note that `R.times` does pass an argument to `R.always('hi')`; the argument is simply ignored.

aaditmshah:
I agree with @davidchambers. It makes no sense to call `T` with no arguments.

CrossEye:
I can think of other uses for `R.always(x)()`.  True, they're not particularly FPish, but they seem fairly legitimate to me.  Imagine for instance wanting to stub out -- during early development, as a debugging measure, or for unit testing -- some (obviously not referentially transparent) nullary function such as `getNextId`:

    // temporary work-around : DON'T CHECK IN!
    // var getNextId = function() { /* ... */ }
    var getNextId = R.always(42);

Since the original `getNextId` was called as a nullary function, the plan would be for this one to be called that way too.  This then worries me.

Since there is definitely a loss involved, I think we have to know what we would be gaining by this change.  Does this simply make for a more elegant implementation?  Does it align better with some ideas of functional purity?  Am I missing some other advantage?

davidchambers:
The original question was what should `f.length` be if `f = R.always(42)`? Currently it's 0. In most contexts `f` will be applied to one argument, so the only way for us to reconcile this mismatch is to think of `f` as a variadic function.

If instead we give `f` a length of 1, we can think of `f` as a fixed-arity function. In general when working with Ramda we prefer fixed-arity functions as we like to curry our functions. :)

We are losing functionality, though `getNextId` can easily be defined as `function() { return 42; }`. We are gaining clarity (I believe), and *possibly* practical benefits due to `f` having length 1 rather than 0.

buzzdecafe:
> In most contexts f will be applied to one argument,

how do you know that? 

> We are losing functionality ... [and gaining] possibly practical benefits

a bird in the hand ...

CrossEye:
@davidchambers:

> We are gaining clarity (I believe), and _possibly_ practical benefits due to `f` having length 1 rather than 0.

I'm not certain of the clarity.  Can you explain more fully why you think it's more clear?  Is it only because in some instances we will be calling it as a unary function that we should therefore define it as one?

But, more importantly, can you describe the practical benefits you see from this?  I simply don't see any.

test/always.js
```diff
@@ -6,19 +6,26 @@ var R = require('..');
...
+        assert.deepEqual(R.always(now)('xxx'), new Date(1776, 6, 4));
+    });
+
+    it('returns a unary function', function() {
+        assert.strictEqual(R.always(42).length, 1);
+    });
+
+    it('is binary', function() {
+        assert.strictEqual(R.always.length, 2);
+        assert.strictEqual(R.always(42, 'xxx'), 42);
```
megawac: :-1: `R.always` should only ever create a function in my mind

aaditmshah: I never understood the point of writing `function (any) { any; return val; }`. It's not like `any` is going to have any observable effect. Why not just write `function (any) { return val; }`? I see this same pattern in the `arity` function:

```javascript
module.exports = function(n, fn) {
    switch (n) {
        case 0: return function() {return fn.apply(this, arguments);};
        case 1: return function(a0) {void a0; return fn.apply(this, arguments);};
        case 2: return function(a0, a1) {void a1; return fn.apply(this, arguments);};
        case 3: return function(a0, a1, a2) {void a2; return fn.apply(this, arguments);};
        case 4: return function(a0, a1, a2, a3) {void a3; return fn.apply(this, arguments);};
        case 5: return function(a0, a1, a2, a3, a4) {void a4; return fn.apply(this, arguments);};
        case 6: return function(a0, a1, a2, a3, a4, a5) {void a5; return fn.apply(this, arguments);};
        case 7: return function(a0, a1, a2, a3, a4, a5, a6) {void a6; return fn.apply(this, arguments);};
        case 8: return function(a0, a1, a2, a3, a4, a5, a6, a7) {void a7; return fn.apply(this, arguments);};
        case 9: return function(a0, a1, a2, a3, a4, a5, a6, a7, a8) {void a8; return fn.apply(this, arguments);};
        case 10: return function(a0, a1, a2, a3, a4, a5, a6, a7, a8, a9) {void a9; return fn.apply(this, arguments);};
        default: throw new Error('First argument to arity must be a non-negative integer no greater than ten');
    }
};
```

What's the point of writing `void a0`, etc. at all? BTW, if you are going to write it then I think this is more readable, `function (any) { return any, val; }`.

davidchambers: This is just to appease JSHint, which would otherwise complain about the unreferenced identifier.

CrossEye: Which makes me think, we've turned off either JSHint or JSCS warnings temporarily for something recently.

We should probably do the same here if possible.  When these tools are dictating what we think of as suboptimal practices, we should be skirting them as best we can, no?

davidchambers: Ideally JSHint would allow us to ignore unreferenced parameters, but it does not as far as I know. I don't want to disable this rule entirely as it avoids accidental inclusion of unused code. `void` and `// jshint ignore:line` are some of the tools we have. They're not ideal, but I'm willing to make small sacrifices in exchange for the benefits JSHint offers.

CrossEye: This seems to be all it takes:

```javascript
/*jshint unused:false */
```

I'll try to check in tonight.  This is working locally for `arity`:

```javascript
module.exports = function(n, fn) { /*jshint unused:false */
    switch (n) {
        case  0: return function() {return fn.apply(this, arguments);};
        case  1: return function(a0) {return fn.apply(this, arguments);};
        case  2: return function(a0, a1) {return fn.apply(this, arguments);};
        case  3: return function(a0, a1, a2) {return fn.apply(this, arguments);};
        case  4: return function(a0, a1, a2, a3) {return fn.apply(this, arguments);};
        case  5: return function(a0, a1, a2, a3, a4) {return fn.apply(this, arguments);};
        case  6: return function(a0, a1, a2, a3, a4, a5) {return fn.apply(this, arguments);};
        case  7: return function(a0, a1, a2, a3, a4, a5, a6) {return fn.apply(this, arguments);};
        case  8: return function(a0, a1, a2, a3, a4, a5, a6, a7) {return fn.apply(this, arguments);};
        case  9: return function(a0, a1, a2, a3, a4, a5, a6, a7, a8) {return fn.apply(this, arguments);};
        case 10: return function(a0, a1, a2, a3, a4, a5, a6, a7, a8, a9) {return fn.apply(this, arguments);};
        default: throw new Error('First argument to arity must be a non-negative integer no greater than ten');
    }
};
```

davidchambers:
The ball's in your court, @CrossEye. I'm still in favour of this change, though not strongly. Let's resolve this pull request one way or the other.

CrossEye:
I know that @buzzdecafe is also opposed.  So let's close it for now.  And if you want to try to convince us after we release 0.9, I promise to listen. :smile:

PR #735, opened 2015-01-18
integrate doctest

davidchambers:
This pull request replaces our custom doctest runner with a dependency on [doctest][1], and tweaks the formatting of the doctests accordingly.

Before:

    2 * 3 * 7  //=> 42

After:

    > 2 * 3 * 7
    42

[1]: https://github.com/davidchambers/doctest


src/wrap.js
```diff
@@ -15,18 +15,13 @@ var arity = require('./arity');
...
- *      });
- *      shortenedGreet("Robert"); //=> "Hello Rob"
- *
+ *      > var slashify = R.wrap(R.flip(R.add)('/'), function(f, x) {
+ *      .   return R.match(/\/$/, x) ? x : f(x);
+ *      . })
+ *      > slashify('a')
+ *      'a/'
+ *      > slashify('a/')
+ *      'a/'
```
CrossEye: Was this reversion intentional?

CrossEye:
The plus side of this is large: not having a custom component like this to maintain, depending instead on a presumably more battle-tested solution, is clearly a big advantage.

But there is a real downside as well to my mind.  One significant advantage of our current test code is that it's easy to cut and paste into the console.  The idea is that it's runnable as is.  The doctest version makes that much harder, with no compensating advantages I can see for the user.

This definitely seems easier for someone to paste into a REPL/console/editor:

    R.prop('x', {x: 100}); //=> 100
    R.prop('x', {}); //=> undefined

than this:

    > R.prop('x', {x: 100})
    100
    > R.prop('x', {})
    undefined

Of course to me, the whole notion of running the examples took a while to accept.  It wasn't until one day I had a sudden realization that these were not in any way unit tests but were instead some form of documentation verification that it finally sunk in for me.  So maybe my understanding isn't really complete.

I'm hesitant to object, because clearly @davidchambers put a lot of work into this and I really would rather use a ready-made tool than a home-grown one we have to maintain.  But my objection is real.  So David, any chance doctest is looking to support another format?  :smile:

src/wrap.js
davidchambers: No. Good catch! I'll update it now.

davidchambers:
> One significant advantage of our current test code is that it's easy to cut and paste into the console.

Doctests depict a REPL session. Well-written doctests depict the REPL session of a user familiarizing herself with the function in question, checking the output for the common case then various edge cases. Doctests can provide the benefit of an exploratory REPL session without the context switching.

Elsewhere in your comment you mention pasting doctests into a text editor. This is something one may wish to do, in order to tweak the code then observe the change in output. This is undoubtedly less convenient in the new format.

On the other hand one can copy doctests from the REPL, so the new format is more convenient for authors.

The fact that I've put a lot of work into this pull request shouldn't affect our decision as to whether to accept it. It should be merged or closed on its merits. :)

kedashoe:
> Elsewhere in your comment you mention pasting doctests into a text editor. This is something one may wish to do, in order to tweak the code then observe the change in output. This is undoubtedly less convenient in the new format.

Just glanced briefly, but it looks like this might be as easy as looking for every line that starts with `> `, removing that opening, and throwing a `;` on the end. You could put a "Copy" button on the examples on http://ramda.github.io/ramdocs which does the simple reformatting and shows a textarea with the new editor friendly code selected for easy copy/paste.

davidchambers:
Even better, @kedashoe, would be to apply this transformation at build time. That way documentation generated from the doctests could be in a copy-friendly format.

CrossEye:
> Even better, @kedashoe, would be to apply this transformation at build time.

That works both ways, though.  If the output format is not important to you @davidchambers, but only the format to use for testing the examples, then I would suggest that it might be best to do the conversion only on the way in to doctest.

Issue #736, opened 2015-01-19
`R.compose` should make curried function with correct arity


anton-rudeshko:
Almost same as in https://github.com/ramda/ramda/issues/725

Kinda stange behavior:

```js
R.concat([1], [2]); // [1, 2]
R.concat([1])([2]); // [1, 2]
R.join(',', R.concat([1], [2])); // "1,2"

var concatAndJoin = R.compose(R.join(','), R.concat);
concatAndJoin([1], [2]); // "1,2"
concatAndJoin([1])([2]); // Uncaught TypeError: Cannot read property 'apply' of undefined 
```

Sad =(

------

So I think `R.compose` should return curried function with same arity as it last argument.

```js
R.compose(R.join(','), R.concat); // curried with 2 args, same as R.concat.
```

CrossEye:
Obviously you could call `curry` on this.  And that's been my answer so far.

I've been fighting what I think I'm starting to see as a rearguard action against this.  You can see for instance, https://github.com/ramda/ramda/issues/732#issuecomment-70415923 or #485.  There is a serious issue involved in what to do with functions that don't properly report their arities, either variadic ones or other ones, but I guess I'm starting to be won over to the notion that we're going to have to do this.  I still don't know exactly how to deal with those issues.  But it would be really nice to have the resulting function curried.

Any suggestions for how to handle the variadic or poor-arity-reporting functions?

anton-rudeshko:
> Obviously you could call curry on this. And that's been my answer so far.

I want to insist on Ramda principle listed everywhere:

> Ramda functions are automatically curried. 

And fundamental to FP `compose` isn't curried? Sounds like a joke to me.

Yes, obviously I can `curry` by hand. But what's the point?

I think that without common approach for currying variadic functions in JS (or in Ramda in particular) we need to leave this decision to users because solution may depend strongly on use case.

CrossEye:
> I want to insist on Ramda principle listed everywhere:

>> Ramda functions are automatically curried.

That's usually preceded with "almost all". :smile:  There is a reason for this.  We have a few variadic functions, which cannot comfortably be curried.  But that's different from saying that functions you create using Ramda will be curried as well.  I'm starting to come to think that we should go that route as much as possible, but there are some complexities to work through.


> And fundamental to FP `compose` isn't curried? Sounds like a joke to me.

[...]

PR #737, opened 2015-01-19
return curried function in compose

stoeffel:
I'm new to Ramda but this should solve #736.

### DIFF
--- a/src/internal/_createComposer.js
+++ b/src/internal/_createComposer.js
@@ -1,4 +1,4 @@
-var arity = require('../arity');
+var curryN = require('../curryN');
 
 
 /*
@@ -13,6 +13,6 @@ module.exports = function _createComposer(composeFunction) {
     while (--idx >= 0) {
       fn = composeFunction(arguments[idx], fn);
     }
-    return arity(length, fn);
+    return curryN(length, fn);
   };
 };
--- a/test/compose.js
+++ b/test/compose.js
@@ -65,4 +65,10 @@ describe('compose', function() {
     assert.deepEqual(g(1, 2, 3), [1, 2, 3]);
   });
 
+  it('should return curried function with same arity as it last argument', function() {
+    var concatAndJoin = R.compose(R.join(','), R.concat);
+    assert.strictEqual(concatAndJoin([1], [2]), '1,2');
+    assert.strictEqual(concatAndJoin.length, 2);
+    assert.strictEqual(concatAndJoin([1])([2]), '1,2');
+  });
 });
--- a/test/pipe.js
+++ b/test/pipe.js
@@ -49,4 +49,11 @@ describe('pipe', function() {
     assert.strictEqual(g.length, 3);
     assert.deepEqual(g(1, 2, 3), [1, 2, 3]);
   });
+
+  it('should return curried function with same arity as its first argument', function() {
+    var concatAndJoin = R.pipe(R.concat, R.join(','));
+    assert.strictEqual(concatAndJoin([1], [2]), '1,2');
+    assert.strictEqual(concatAndJoin.length, 2);
+    assert.strictEqual(concatAndJoin([1])([2]), '1,2');
+  });
 });


test/compose.js
```diff
@@ -63,4 +63,10 @@ describe('compose', function() {
         assert.strictEqual(R.compose(f), f);
     });
 
+    it('should return curried function with same arity as it last argument', function() {
+        var concatAndJoin = R.compose(R.join(','), R.concat);
+        assert.strictEqual(concatAndJoin([1], [2]), '1,2');
+        assert.strictEqual(concatAndJoin.length, 2);
+        assert.strictEqual(concatAndJoin([1])([2]), '1,2');
+    });
```
davidchambers: You should add a similar test case for `R.pipe`.

stoeffel: I will do that.

src/internal/_createComposer.js
```diff
@@ -16,7 +17,7 @@ module.exports = function _createComposer(composeFunction) {
                 while (idx--) {
                     fn = composeFunction(arguments[idx], fn);
                 }
-                return arity(length, fn);
+                return curryN(length, arity(length, fn));
```
CrossEye: I don't think there's any reason for both `curryN` and `arity`.  You should be able to simply replace `arity` with `curryN` (untested).

CrossEye:
We still need to decide [whether][co] before we take on [how][fi], I'm afraid.


  [co]: https://github.com/ramda/ramda/issues/736#issuecomment-70555440
  [fi]: https://github.com/ramda/ramda/pull/737/files

test/compose.js
stoeffel: Done.

buzzdecafe:
*ping*

buzzdecafe:
once #771 is in, i think it makes sense to do this as well

stoeffel:
I will rebase as soon as #771 is in.

buzzdecafe:
although #771 got whacked, i think we should go ahead with this, essentially returning `curryN(max(fns), f)` as we are doing elsewhere now,

CrossEye:
I'm still nervous about this.  I agree that this should be the default behavior, but I really would like to find a way to offer a version of `compose` / `pipe` that don't do this in the case that the returned function has named optional parameters.

buzzdecafe:
#985 is how `converge` works today. `compose` should have the same semantics IMO.

stoeffel:
> #985 is how converge works today. compose should have the same semantics IMO.

I think so too.

I rebased this.

src/internal/_createComposer.js
```diff
@@ -1,18 +1,22 @@
-var arity = require('../arity');
+var _noArgsException = require('./_noArgsException');
+var curryN = require('../curryN');
 
 
 /*
  * Returns a function that makes a multi-argument version of compose from
- * either _compose or _composeP.
+ * either _compose or _pCompose.
```
davidchambers: This should be `_composeP` rather than `_pCompose`.

davidchambers:
We've switched from using four spaces for indentation to two spaces. Please adjust the indentation accordingly. :)

davidchambers:
Also, we no longer update __dist/ramda.js__ with each commit, so you should revert the changes to that file.

stoeffel:
Yop, just saw that it will need more than just rebasing :smile: .

stoeffel:
Now, it should be fin.

davidchambers:
This looks good! It might be clearer to use an uncurried function in the test cases. As it stands one wonders whether `R.compose(R.join(','), R.concat)` evaluates to a curried function because `R.concat` is a curried function.

buzzdecafe:
now the question is, do we need to have `composeV` (for variadic) with the old semantics to cover @CrossEye 's case?

otherwise, herd of :cow2:

CrossEye:
I really hate having so many variants of this basic function.

But I think not offering that alternative does a big disservice to JS users who are used to working with functions with optional parameters, especially a final `options` argument.

davidchambers:
Superseded by #1260

PR #741, opened 2015-01-21
currying the output of wrap

### LINKED ISSUE #725: `R.wrap` should make curried function
Kind of strange, isn't it stated that every function in Ramda is autmatically curried?

```js
var wrappedAdd = R.wrap(R.add, function(add, x, y) {
  console.log('adding %s and %s', x, y); 
  return add(x, y); 
});

wrappedAdd(1) // "adding 1 and undefined" => NaN
wrappedAdd(1, 2) // "adding 1 and 2" => 3
```

I particularly interested in wrapping functions for debugging as shown in this example.

CrossEye:
Fixes #725.

(Let's see if it works this time!)

### DIFF
--- a/src/wrap.js
+++ b/src/wrap.js
@@ -1,5 +1,5 @@
 var _concat = require('./internal/_concat');
-var arity = require('./arity');
+var curryN = require('./curryN');
 
 
 /**
@@ -29,7 +29,7 @@ var arity = require('./arity');
  *
  */
 module.exports = function wrap(fn, wrapper) {
-    return arity(fn.length, function() {
+    return curryN(fn.length, function() {
         return wrapper.apply(this, _concat([fn], arguments));
     });
 };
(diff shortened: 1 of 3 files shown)


davidchambers:
:deciduous_tree:

src/wrap.js
```diff
@@ -29,7 +29,7 @@ var arity = require('./arity');
  *
```
anton-rudeshko: Could you please add an example of currying?

CrossEye: @anton-rudeshko: I guess I'm willing to, if you feel strongly about it.  I don't see much of a need.  Part of it is that this function is not one I use often, so I'm not particularly invested in it.  Another part is that I'm (still!) generally less interested in inline documentation.  :-)

Do you have a suggested short example?

anton-rudeshko: Hmm, no example right now. Will do a PR if any come to mind. Thanks for the fix.

PR #744, opened 2015-01-21
Updated ifElse docs.

jethrolarson:
Docs were misusing the word "predicate". I've updated them for clarity.


davidchambers:
Could you run `make` to update __ramda.js__?

jethrolarson:
Maybe later, I just updated via website, so I didn't have that handy.

CrossEye:
I've dropped the ball on this, but I will try this weekend to see if we can get Travis to check back in the generated `ramda.js` on a successful build so that we can add this to `.gitignore` and not have to include artifacts of the build process in our commits.

CrossEye:
Good catch on those changes.  That was definitely misuse of "predicate"!

PR #745, opened 2015-01-22
currying the mixin with op

### LINKED ISSUE #743: mixin function curring
Hi,

I'd like to use `mixin` function in a `pipe`, but it's not possible. I would expect that it behaves the same way as `divide` function does. Does it have any reason why is used `curry` instead of `op`? I think that `curry` should be used only for commutative functions.

I know that I can use converge, but it's not ideal from my point of view.

honzabrecka:
Fixes #743.

### DIFF
--- a/src/mixin.js
+++ b/src/mixin.js
@@ -1,5 +1,5 @@
-var _curry2 = require('./internal/_curry2');
 var _extend = require('./internal/_extend');
+var op = require('./op');
 
 
 /**
@@ -18,7 +18,10 @@ var _extend = require('./internal/_extend');
  *
  *      R.mixin({ 'name': 'fred', 'age': 10 }, { 'age': 40 });
  *      //=> { 'name': 'fred', 'age': 40 }
+ *
+ *      var resetToDefault = R.mixin(R.__, {x: 0});
+ *      resetToDefault({x: 5, y: 2}); //=> {x: 0, y: 2}
  */
-module.exports = _curry2(function mixin(a, b) {
+module.exports = op(function mixin(a, b) {
     return _extend(_extend({}, a), b);
 });
--- a/test/mixin.js
+++ b/test/mixin.js
@@ -36,4 +36,9 @@ describe('mixin', function() {
         var b = {y: 3, z: 4};
         assert.deepEqual(curried(b), {w: 1, x: 2, y: 3, z: 4});
     });
+
+    it('is curried', function() {
+        var curried = R.mixin(R.__, {w: 1, x: 2});
+        assert.deepEqual(curried({x: 3, y: 4}), {w: 1, x: 2, y: 4});
+    });
 });
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -5748,8 +5748,11 @@
      *
      *      R.mixin({ 'name': 'fred', 'age': 10 }, { 'age': 40 });
      *      //=> { 'name': 'fred', 'age': 40 }
+     *
+     *      var resetToDefault = R.mixin(R.__, {x: 0});
+     *      resetToDefault({x: 5, y: 2}); //=> {x: 0, y: 2}
      */
-    var mixin = _curry2(function mixin(a, b) {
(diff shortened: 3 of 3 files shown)


src/mixin.js
```diff
@@ -1,4 +1,4 @@
-var _curry2 = require('./internal/_curry2');
+var _curry2 = require('./op');
```
stoeffel: This should be `var op`

stoeffel:
You should run `make test lint` and add `dist/ramda.js` to your commit.

> Update dist/ramda.js with each commit

Checkout https://github.com/ramda/ramda/blob/master/CONTRIBUTING.md

honzabrecka:
@stoeffel Thanks for your help!

stoeffel:
No problem

src/mixin.js
```diff
@@ -1,6 +1,5 @@
-var _curry2 = require('./internal/_curry2');
 var _extend = require('./internal/_extend');
-
```
davidchambers: Nit: Please leave two empty lines between the imports and the code below.

test/mixin.js
```diff
@@ -36,4 +36,10 @@ describe('mixin', function() {
         var b = {y: 3, z: 4};
         assert.deepEqual(curried(b), {w: 1, x: 2, y: 3, z: 4});
     });
+
+    it('is curried', function() {
+        var curried = R.mixin(undefined, {w: 1, x: 2});
```
davidchambers: This should be `R.__` rather than `undefined`.

src/mixin.js
```diff
@@ -18,7 +17,11 @@ var _extend = require('./internal/_extend');
  *
  *      R.mixin({ 'name': 'fred', 'age': 10 }, { 'age': 40 });
  *      //=> { 'name': 'fred', 'age': 40 }
+ *
+ *      // note: In this example, `__`  is just an `undefined` value.  You could use `void 0` instead
+ *      var resetToDefault = R.mixin(__, {x: 0, y: 0});
```
davidchambers: Let's use `R.__` in the example and delete the note.

CrossEye:
LGTM

src/mixin.js
```diff
@@ -18,7 +18,10 @@ var _extend = require('./internal/_extend');
  *
  *      R.mixin({ 'name': 'fred', 'age': 10 }, { 'age': 40 });
  *      //=> { 'name': 'fred', 'age': 40 }
+ *
+ *      var resetToDefault = R.mixin(R.__, {x: 0, y: 0});
+ *      resetToDefault({x: 5, y: 2}); //=> {x: 0, y: 0}
```
davidchambers: It would be nice for at least one of the map entries in the lower-precedence map to show through:

```javascript
var resetX = R.mixin(R.__, {x: 0});
resetX({x: 5, y: 2}); //=> {x: 0, y: 2}
```

test/mixin.js
```diff
@@ -36,4 +36,10 @@ describe('mixin', function() {
         var b = {y: 3, z: 4};
         assert.deepEqual(curried(b), {w: 1, x: 2, y: 3, z: 4});
     });
+
+    it('is curried', function() {
+        var curried = R.mixin(R.__, {w: 1, x: 2});
+        var b = {x: 3, y: 4};
+        assert.deepEqual(curried(b), {w: 1, x: 2, y: 4});
```
davidchambers: The name `b` implies that `b` is the second argument to `R.mixin` (the first argument being `a`), yet `b` is actually the first argument. I suggest not naming the argument:

```javascript
assert.deepEqual(curried({x: 3, y: 4}), {w: 1, x: 2, y: 4});
```

buzzdecafe:
is this good to go?

CrossEye:
Yes, I think it is.

davidchambers:
@honzabrecka, could you tweak the doctest and resolve the merge conflicts? I'd love to merge this. :)

honzabrecka:
Done. :)

PR #746, opened 2015-01-22
assign sentinel object to R.__

davidchambers:
I've never been comfortable with the value of `R.__` being `undefined`. Why "define" it at all, in this case? I like defining a true sentinel and requiring users to reference it directly or define a convenient :alien::

```javascript
var _ = R.__;
var lessThanTen = R.lt(_, 10);
```

### DIFF
--- a/src/op.js
+++ b/src/op.js
@@ -1,15 +1,12 @@
-var __ = require('./internal/__');
+var __ = require('./__');
 var _noArgsException = require('./internal/_noArgsException');
-var binary = require('./binary');
 var flip = require('./flip');
 var lPartial = require('./lPartial');
-var rPartial = require('./rPartial');
-var unary = require('./unary');
 
 
 /**
  * Uses a placeholder to convert a binary function into something like an infix operation.
- * When called with an `undefined` placeholder (e.g. `R.__`) the second argument is applied to the
+ * When called with the `R.__` placeholder the second argument is applied to the
  * second position, and it returns a function waiting for its first argument.
  * This can allow for more natural processing of functions which are really binary operators.
  *
@@ -26,19 +23,19 @@ var unary = require('./unary');
  *
  *      div(6, 3); //=> 2
  *      div(6)(3); //=> 2
- *      div(__, 3)(6); //=> 2 // note: `__` here is just an `undefined` value.  You could use `void 0` instead
- *      div(__)(3, 6); //=> 2
- *      div(__)(3)(6); //=> 2
+ *      div(R.__, 3)(6); //=> 2
+ *      div(R.__)(3, 6); //=> 2
+ *      div(R.__)(3)(6); //=> 2
  */
 module.exports = function op(fn) {
-    var length = fn.length;
-    if (length !== 2) {throw new Error('Expected binary function.');}
-
+    if (fn.length !== 2) {
+        throw new Error('Expected binary function.');
+    }
     return function _op(a, b) {
         switch (arguments.length) {
             case 0: throw _noArgsException();
-            case 1: return a === __ ? binary(flip(_op)) : unary(lPartial(fn, a));
-            default: return a === __ ? unary(rPartial(fn, b)) : fn(a, b);
+            case 1: return a === __ ? flip(_op) : lPartial(fn, a);
+            default: return a === __ ? flip(fn)(b) : fn(a, b);
         }
     };
 };
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -7,7 +7,7 @@
 
     'use strict';
 
-    var __ = void 0;
+    var __ = { ramda: 'placeholder' };
 
     var _add = function _add(a, b) {
         return a + b;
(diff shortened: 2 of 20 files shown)


buzzdecafe:
:cake:

src/op.js
```diff
@@ -26,19 +23,19 @@ var unary = require('./unary');
...
     return function _op(a, b) {
         switch (arguments.length) {
             case 0: throw _noArgsException();
-            case 1: return a === __ ? binary(flip(_op)) : unary(lPartial(fn, a));
-            default: return a === __ ? unary(rPartial(fn, b)) : fn(a, b);
+            case 1: return a === __ ? flip(_op) : lPartial(fn, a);
+            default: return a === __ ? flip(fn)(b) : fn(a, b);
         }
     };
 };
```
CrossEye: These fixes look worth doing regardless.  What were we smoking?

buzzdecafe:
i think a unique placeholder object is clearer than using `undefined`. always felt a little queasy about using `undefined` as a "magic" value like that. I don't think we need any properties on the object, though.

davidchambers:
> What are the advantages? Right now a user can use any undefined value she has lying around. What would a true sentinel gain the user? What would it gain our codebase?

Clarity. Right now our message to users is along these lines:

> You may use `R.__` as a placeholder. Since the value of `R.__` is `undefined`, you could just as well use `R._` or `R.___` or `R.placeholder` or any other nonexistent property of `R` (or any other object). You may prefer to use `undefined` directly, though in some environments it's safer to use `void 0`. Possibly the best option is to include `var _;` at the top of the file then to use `_` wherever a placeholder is required.

This is what I would like to say instead:

> You may use `R.__` as a placeholder.

I find the alternative acceptable too, though it's not as succinct:

> You may use `undefined` as a placeholder. It's a good idea to include `var _;` at the top of the file to create a conveniently named variable whose value is `undefined`.

Let's pick `R.__` or `undefined`. If we pick the former, let's make it a true sentinel. If we pick the latter, let's remove all the references to `R.__`.

CrossEye:
@davidchambers: 

As ususal, you make a good argument.  I still can't say I like it.  I would rather go with `undefined`, but I believe that had unpleasant side-effects for our example code, each block of which would really need to then declare its own undefined value.

My objection to the sentinel is that `R.__` is ugly and over-verbose for what it does, and requiring users who want the nicer `_` to then assign a Ramda-specific alias in order to use it seems a little demanding.  That they could use whatever alias to `undefined` they chose, even if they also wanted to use it for non-Ramda purposes, is a nicer experience.

But I don't think I can have my :cake: and eat it too.

So here's a very reluctant :+1:

davidchambers:
> My objection to the sentinel is that `R.__` is ugly and over-verbose for what it does, and requiring users who want the nicer `_` to then assign a Ramda-specific alias in order to use it seems a little demanding. That they could use whatever alias to `undefined` they chose, even if they also wanted to use it for non-Ramda purposes, is a nicer experience.

I'm fine with closing this pull request and opening a new one to *remove* `R.__`. My main concern is that we have two ways to do something trivial, which complicates our explanations of placeholders.

CrossEye:
> My main concern is that we have two ways to do something trivial, which complicates our explanations of placeholders.

I think the explanation could be significantly simpler:

> You need to supply an `undefined` value for the placeholder.  We offer the undefined `R.__` as one such value.

davidchambers:
> I don't think we need any properties on the object, though.

True, but I don't see any harm in "labelling" the object to aid debugging.

CrossEye:
> I don't see any harm in "labelling" the object to aid debugging.

I suppose I agree.

But I think `placeholder` is more explicit than `sentinel`:

    {ramda: 'placeholder'}

or

    {ramdaPlaceholder: 'sentinel'}

One other thought.  I'm assuming that we'll stick with our _caveat emptor_, or with Peter Parker's "with great power comes great cliches" (sorry, some translation glitch), for the possibility of someone reassigning the value of `R.__`.  I think that to make it immutable would involve adding exception logic for that file in the build script, something I'd rather avoid.  If we can think of another way, it might be worth doing, though.

buzzdecafe:
> If we can think of another way, it might be worth doing, though.

make `R.__` a reference to `R`? then if it gets reassigned you can always put it back

CrossEye:
>> If we can think of another way, it might be worth doing, though.

> make `R.__` a reference to `R`? then if it gets reassigned you can always put it back

I'm just not sure how we would do that.  `R` is not defined until the distribution is created and the return from the revealing module pattern is constructed.  So it's not available inside the `op` function for such a reference.

davidchambers:
> I think `placeholder` is more explicit than `sentinel`

I agree. I've updated the label to `{ramda: 'placeholder'}`.

As for the possibility of a user assigning another value to `R.__`:

  - we *could* make `R` immutable in environments which provide `Object.freeze`, but this would prevent users from adding their own functions to the object;

  - the current approach is equally "vulnerable", since one could assign `42` to `R.__` while elsewhere assuming `R.__` to be `undefined`; and

  - I'm not in the slightest bit worried as users *should not* mess with `R.__`.

CrossEye:
>  *  we could make R immutable in environments which provide Object.freeze, but this would prevent users from adding their own functions to the object;

Oh no, I would not consider that.  I was merely thinking along the lines of 

    Object.defineProperty(R, '__', {
        enumerable: true,
        configurable: false,
        writable: false,
        value: {ramda: 'placeholder'}
    });

which would then have to be wrapped in some not-ES3 check... not worth it.

>  * the current approach is equally "vulnerable", since one could assign 42 to R.__ while elsewhere assuming R.__ to be undefined

Yes, but as that is likely to be the only reference to the sentinel, it will now be fatal, whereas when we used `undefined`, it was easy to reset, or to work around if you simply used another undefined reference.  Again, I don't think this is a big deal.


>  * I'm not in the slightest bit worried as users should not mess with R.__.

Right, and unless we grow to have jQuery-sized concerns, I don't think we ever really need to worry about how one user deals with another's abuse of our library inside the same scope.  If we ever get to the point of having such worries, this will probably be among the more minor ones.

PR #749, opened 2015-01-23
removing voids from arity

CrossEye:
Like #698.  Works in `grunt jshint`.  Does this fail in `make`?

### DIFF
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -515,7 +515,9 @@
      *      // All arguments are passed through to the wrapped function
      *      takesOneArg(1, 2); //=> [1, 2]
      */
+    /*jshint unused:false */
     var arity = function (n, fn) {
+        /*jshint unused:false */
         switch (n) {
         case 0:
             return function () {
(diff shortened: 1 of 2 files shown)


megawac:
:+1: but what happens when you minify it?

CrossEye:
@megawac:  Not sure what the question is.  So the comment is stripped out and it wont lint properly.  But a minified file shouldn't be passed to lint anyway, right?  Is there some other concern?

megawac:
Will the parameter be removed in the minifed build (i.e. the arity be set incorrectly)

dist/ramda.js
```diff
@@ -515,7 +515,9 @@
      *      // All arguments are passed through to the wrapped function
      *      takesOneArg(1, 2); //=> [1, 2]
      */
+    /*jshint unused:false */
     var arity = function (n, fn) {
+        /*jshint unused:false */
```
megawac: why is it duplicated

CrossEye:
Ahh, I didn't know minifiers might do that.  I'll check.

CrossEye:
Ahh, looks like the minification is a problem:

> WARN: Dropping unused function argument a0 [-:527,29]

I'm going to close this one for now.  Maybe at some point we can also configure the minifier, but that's getting a fair bit trickier.

PR #750, opened 2015-01-23
remove R.__

davidchambers:
This is the alternative to #746. I see no reason to provide `R.__` if its value is and will always be `undefined`. `R.__` will be `undefined` even *after* this change. ;)

### DIFF
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -7,8 +7,6 @@
 
     'use strict';
 
-    var __ = void 0;
-
     var _add = function _add(a, b) {
         return a + b;
     };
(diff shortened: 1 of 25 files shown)


CrossEye:
... and just as I'm (reluctantly, I admit!) coming around to share the point of view you and @buzzdecafe were expressing in #746!  :question: :smiling_imp:

buzzdecafe:
won't fight for it, but i prefer a sentinel object to `undefined`. Using `undefined` feels too magicky. I prefer the specificity of an object that is associated with particular behavior.

CrossEye:
What's sort of winning me over is the notion that someone might want to use `op` on a function

    :: a -> b -> c

where `a === undefined` is not actually an error.  I can't really come up with such a function right now that makes sense with `op`, but I'd rather this API not be tied to my failure of imagination.  So I'm voting for #746 in spite of my previous comments and even my current reluctance.

davidchambers:
Closing in favour of #746

PR #752, opened 2015-01-25
prepublish: ensure generated files are freshly generated

davidchambers:
I noticed after publishing 0.9.0 that the header of __ramda.js__ still mentions v0.8.0. The reason for this is that none of the file's dependencies had changed since the file was last built, so Make did not rebuild the file.

### DIFF
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -1,4 +1,4 @@
-//  Ramda v0.8.0
+//  Ramda v0.9.0
 //  https://github.com/ramda/ramda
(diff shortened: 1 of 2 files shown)


dist/ramda.js
```diff
@@ -1,4 +1,4 @@
-//  Ramda v0.8.0
+//  Ramda v0.9.0
```
davidchambers: I included this change to fix the test failure.

buzzdecafe:
:cake:

PR #755, opened 2015-01-25
build: add Ramda as a dependency

davidchambers:
This should prevent @buzzdecafe from pulling his hair out. :)


buzzdecafe:
> This should prevent @buzzdecafe from pulling his hair out. :)

not much to spare, so this is just in time :smiley:

CrossEye:
:yellow_heart:

PR #758, opened 2015-01-26
make countBy handles reserved words in prototype

404pnf:
Current implementation doesn't handle reserved words in prototype
> c(function (e) { return e; }, ['abc', 'toString'])
{ abc: 1, toString: 'function toString() { [native code] }1' }

Proposed change can handle that
c(function (e) { return e; }, ['abc', 'toString'])
{ abc: 1, toString: 1 }

### DIFF
--- a/src/countBy.js
+++ b/src/countBy.js
@@ -28,5 +28,5 @@ module.exports = _curry2(function countBy(fn, list) {
     return _foldl(function(counts, obj) {
         counts[obj.key] = (counts[obj.key] || 0) + 1;
         return counts;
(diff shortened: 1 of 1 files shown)


davidchambers:
Thank you for identifying and reporting this bug. The pull request as it stands cannot be merged for three reasons:

  - it does not update __ramda.js__;
  - it does not include a regression test; and
  - it uses `Object.create` which was added in ES5, while Ramda aims to be ES3 compatible.

davidchambers:
Closing in favour of #759

PR #759, opened 2015-01-26
ignore inherited properties in R.countBy

davidchambers:
Supersedes #758

I simplified the implementation while I was at it. :)


davidchambers:
This pull request fixes a bug, so we should merge it and publish 0.9.1 before merging any other pull requests.

src/countBy.js
```diff
@@ -25,8 +24,12 @@ var _keyValue = require('./internal/_keyValue');
...
-    }, {}, _keyValue(fn, list));
+    var counts = {};
+    var len = list.length;
+    var idx = -1;
+    while (++idx < len) {
+        var key = fn(list[idx]);
+        counts[key] = (has(key, counts) ? counts[key] : 0) + 1;
+    }
+    return counts;
 });
```
CrossEye: I wonder if we should do something similar to `sortBy` and eliminate the need for `_keyValue`.

davidchambers: Good idea. I've added a second commit to this pull request, which removes `_compareKeys` and `_keyValue`. :)

src/sortBy.js
```diff
@@ -36,5 +34,9 @@ var _pluck = require('./internal/_pluck');
  *      sortByNameCaseInsensitive(people); //=> [alice, bob, clara]
  */
 module.exports = _curry2(function sortBy(fn, list) {
-    return _pluck('val', _keyValue(fn, list).sort(_compareKeys));
+    return clone(list).sort(function(a, b) {
+        var aa = fn(a);
+        var bb = fn(b);
+        return aa < bb ? -1 : aa > bb ? 1 : 0;
```
buzzdecafe: can't you just `return aa - bb;`?

dist/ramda.js
```diff
@@ -4250,7 +4162,11 @@
      *      sortByNameCaseInsensitive(people); //=> [alice, bob, clara]
      */
     var sortBy = _curry2(function sortBy(fn, list) {
-        return _pluck('val', _keyValue(fn, list).sort(_compareKeys));
+        return clone(list).sort(function (a, b) {
+            var aa = fn(a);
+            var bb = fn(b);
+            return aa < bb ? -1 : aa > bb ? 1 : 0;
```
buzzdecafe: likewise

src/sortBy.js
davidchambers: Nope. We're comparing strings:

```javascript
> 'foo' - 'bar'
NaN
```

buzzdecafe: i expected you would have a ready reply. LGTM

CrossEye:
LVGTM

PR #761, opened 2015-01-26
support Date constructor in R.construct and R.constructN

davidchambers:


src/constructN.js
```diff
@@ -33,12 +31,22 @@ var nAry = require('./nAry');
...
  */
 module.exports = _curry2(function constructN(n, Fn) {
-    var f = function() {
-        var Temp = function() {}, inst, ret;
-        Temp.prototype = Fn.prototype;
-        inst = new Temp();
-        ret = Fn.apply(inst, arguments);
-        return Object(ret) === ret ? ret : inst;
-    };
-    return n > 1 ? curry(nAry(n, f)) : f;
```
davidchambers: I don't know why `0` and `1` are treated specially, so I removed the special case. Speak up if you know what makes these two values special.

src/constructN.js
```diff
@@ -33,12 +31,22 @@ var nAry = require('./nAry');
...
+    if (n > 10) {
+        throw new Error('Constructor with greater than ten arguments');
```
CrossEye: Is there no more specific useful error type?  Perhaps not, but `Error` is awful generic.

src/constructN.js
CrossEye: Why curry a zero- or one- argument function?  I think that was the only rationale.

CrossEye:
Good idea.   We should have thought about this back when we changed `nAry` and `arity` to `switch` statements.

src/constructN.js
CrossEye: More to the point, it would cause an error, something I think your version will have to deal with as well:

```javascript
var Foo = function() {
    this.bar = 'baz';
};
var foo = R.construct(Foo);
var f = foo(); // TypeError: Function called with no arguments
```

The 1-case won't throw an error, but also doesn't do anything useful by currying, so why bother adding such a wrapper?

buzzdecafe: with the current "throw on no args" `curry`:

```javascript
function f() { return 'f'; }
c = curry(f);
c(); //=> TypeError: Function called with no arguments
```

so that's why nullary functions at least have to get special treatment.

davidchambers: Ah, yes. I'll add a test case for a zero-arg constructor and reinstate the special case.

src/constructN.js
davidchambers: JavaScript really needs a ValueError. At a stretch one could consider this a RangeError. Would you prefer that?

src/constructN.js
```diff
@@ -33,12 +31,24 @@ var nAry = require('./nAry');
...
+            case  2: return new Fn($0, $1);
+            case  3: return new Fn($0, $1, $2);
+            case  4: return new Fn($0, $1, $2, $3);
+            case  5: return new Fn($0, $1, $2, $3, $4);
+            case  6: return new Fn($0, $1, $2, $3, $4, $5);
+            case  7: return new Fn($0, $1, $2, $3, $4, $5, $6);
+            case  8: return new Fn($0, $1, $2, $3, $4, $5, $6, $7);
+            case  9: return new Fn($0, $1, $2, $3, $4, $5, $6, $7, $8);
+            case 10: return new Fn($0, $1, $2, $3, $4, $5, $6, $7, $8, $9);
+        }
```
davidchambers: `nAry` guarantees `1 <= arguments.length <= 10`, correct?

src/constructN.js
CrossEye: Honestly, that was almost stream of consciousness as I read through it on my phone on the bus.  I don't care all that much.  There are bigger fish to fry.

src/constructN.js
CrossEye: Yes, but an earlier implementation, based on the Function constructor, did not have any particular length limitation.  It was good to be rid of it, even if it meant sacrificing a tiny bit of theoretic flexibility.  When we did that, we should have realized that the same sort of approach would also have fixed the issues with `construct`.  (But I might have the order of events muddled in my mind.)  I'm just glad you caught it now.

buzzdecafe:
this LGTM -- anything holding this up? this could probably got in a patch version as well

PR #764, opened 2015-01-27
add fork button to docs header [ci skip]

buzzdecafe:


jsdoc-template/docs.html.handlebars
```diff
@@ -20,6 +20,9 @@
                 <li class="active"><a href="#">Documentation</a></li>
             </ul>
         </div>
+        <div class="navbar-right">
+          <iframe class="forkme" src="http://ghbtns.com/github-btn.html?user=ramda&repo=ramda&type=fork&size=large" height="30" width="100" frameborder="0" scrolling="0" style="width:100px; height: 30px;" allowTransparency="true"></iframe>
```
davidchambers: `s/&/&amp;/g` would make this "valid" HTML, I believe.

buzzdecafe: i don't think it's necessary to encode the ampersand inside a link, except maybe in xhtml

davidchambers: It's not *necessary* (see this Stack Overflow [question][1]). There's no ambiguity in this case, but the "always encode `&` rule" is much simpler than the alternative (as highlighted by this [answer][2]). I'm not fussed, though; let's keep the ampersands unencoded.


[1]: http://stackoverflow.com/q/3493405/312785
[2]: http://stackoverflow.com/a/3494165/312785

Issue #728, opened 2015-01-15
curryN(3, ...) vs _curry2 / _curry3


neftaly:
Hey everyone,

I'm playing around with an API function to accept an unlimited number of arguments, and am having a little trouble getting it to curry any more than 2/3 arguments. Apologies in advance if I've missed something simple; have only just picked this up.

Is there an existing way to implement `./internal/_curry` (or `./internal/_curryN`)?

---
---

The following works fine for 2 arguments:
```javascript
var _curry2 = require('./internal/_curry2');

module.exports = _curry2(function doSomething() {
    return stuff;
});
```

---

The following works fine for 3 arguments:
```javascript
var _curry3 = require('./internal/_curry3');

module.exports = _curry3(function doSomething() {
    return stuff;
});
```

---

The following fails testing:
```javascript
var curry = require('./curry');

module.exports = curry(function doSomething() {
    return stuff;
});
```
The test result is:
```
1) doSomething is curried:
     TypeError: object is not a function
      at ...
```

---

I would like to be able to do:
```javascript
var _curry = require('./internal/_curry');

module.exports = _curry(function doSomething() {
    return stuff;
});
```

buzzdecafe:
> I'm modifying an API function to accept an unlimited number of arguments,  and am having a little trouble getting it to curry ...

Yeah, that won't work. You can only curry when the arity is known, either on the function `length` or supplied as a parameter to `curryN`

CrossEye:
>> I'm modifying an API function to accept an unlimited number of arguments, and am having a little trouble getting it to curry ...

> Yeah, that won't work. You can only curry when the arity is known, either on the function `length` or supplied as a parameter to `curryN`

And the problem is not an implementation accident.  This is fundamental.  Curried functions, pretty well by definition, accept some arguments (just one in strict definitions, any number in a version like ours) and then check whether all the arguments to the original function have been supplied.  If so, it's called.  If not, it returns a new curried function that awaits the remaining ones.  But this all falls down if if can't tell if all the arguments have been supplied.  Hence we need to know up front how many arguments to expect.

neftaly:
Thankyou for your thoughtful replies. Definitely a case of "the simple thing I missed"; I'll have to think about  how to implement it properly.

neftaly:
[...]

PR #766, opened 2015-01-27
curry functions returned from curry

buzzdecafe:


davidchambers:
Was there a discussion which prompted this change? I believe so. I'd like to reread that thread.

buzzdecafe:
see #732

CrossEye:
The discussion has wandered around a bit.  #736 and #728 also had interesting things to contribute to these ideas.

buzzdecafe:
*ping*: call for an answer

CrossEye:
I'd like to not include this one in 0.9.2.  I think further discussion is warranted.  I'm afraid there are still possible issues here, but I haven't had time to think them through.

buzzdecafe:
this shouldn't go in a patch release, it's a pretty big change

kedashoe:
I am curious to see some killer examples of how this can be used. http://stackoverflow.com/questions/27996544/how-to-correctly-curry-a-function-in-javascript is weird to me as a proof of concept since it seems much cleaner to simply write the function as

var countWhere = R.compose(length, R.filter);

Some other thoughts:
- I think you would not get this behaviour with any of Ramda's functions as they all use curry2 and curry3? But maybe all curried ramda functions that return functions don't need to be curried?
- When users `curry` their own functions, if their function returns a function, they may be surprised that that function is now curried (probably not bad? but how often is it needed/how do I opt out?), or if that function was already curried this will now be double currying it
- Unnecessary overhead? Just allow users to curry their own returned functions if they want to?

It would be nice though to weigh any concerns against some strong examples where it would be difficult/awkward/impossible to achieve the same result without this behaviour.

CrossEye:
To some extent I share @kedashoe's concerns, although on the balance, I think it would be better to be able to say that every Ramda function, _and every function returned from Ramda_ is curried.

But I have another concern that's related to something I've mentioned before for currying the result of `compose`.  (Too lazy/hurried to look for the reference.)  That is, what happens if the resulting function has named optional parameters.  If we curry to length 3, when only the first two are normally supplied, we've broken that user's function.  It's trivially easy to get this as the result of a `compose` call, but I'm having a harder time imagining it from a `curry`, but I can't see why it couldn't happen.  Then things fall down quickly.  And we're to blame.

buzzdecafe:
*ping*

src/curryN.js
```diff
@@ -39,8 +39,13 @@ module.exports = _curry2(function curryN(length, fn) {
         return arity(Math.max(length - (args && args.length || 0), 0), function() {
             if (arguments.length === 0) { throw _noArgsException(); }
             var newArgs = _concat(args, arguments);
+            var value;
             if (newArgs.length >= length) {
-                return fn.apply(this, newArgs);
+                value = fn.apply(this, newArgs);
+                if (typeof value === 'function') {
+                    return curryN(value.length, value);
```
davidchambers: I believe this will return a broken function if `value.length` is 0. Perhaps `curryN` should return the provided function if `length` is 0 or 1. Alternatively it could throw if `length` is 0 rather than returning a function that will throw if called.

```javascript
R.curryN(0, function() { return 42; })()
// => TypeError: Function called with no arguments
```

buzzdecafe: > Perhaps curryN should return the provided function if length is 0 or 1.

good idea, will revise and amend

buzzdecafe:
superseded by #771

Issue #713, opened 2015-01-03
Add category filter to docs


saikobee:
I think we should go with a dropdown and see what people think of that.

CrossEye:
Agreed.

nomosyn:
:+1:

buzzdecafe:
see https://github.com/ramda/ramda.github.io/issues/11

PR #768, opened 2015-01-28
Documentation: filter by category

wildermuthn:
This allows to filter documentation by category, either by clicking on the category label or typing in the name of a category.

My JS beautifier sets tabs to 2 spaces for JS, which is generally the convention I see now.

### DIFF
--- a/docs/main.js
+++ b/docs/main.js
@@ -4,15 +4,20 @@ function toArray(xs) {
     return Array.prototype.slice.call(xs);
 }
 
+function filterTocType(category) {
+    nameFilter.value = category;
+    filterToc();
+}
+
 function filterToc() {
-    var f = filterElement.bind(null, nameFilter.value, '');
+    var f = filterElement.bind(null, nameFilter.value);
     funcs.forEach(f);
 }
 
-function filterElement(nameFilter, categoryFilter, elem) {
+function filterElement(nameFilter, elem) {
     var name = elem.getAttribute('data-name');
     var category = elem.getAttribute('data-category');
-    var matches = strIn(nameFilter, name) || category === categoryFilter;
+    var matches = strIn(nameFilter, name) || strIn(nameFilter, category);
     elem.style.display = matches ? '' : 'none';
 }
 
@@ -46,6 +51,10 @@ function isTopLink(elem) {
 function dispatchEvent(event) {
     var target = event.target;
     var parent = target.parentNode;
+    var category = target.getAttribute('data-category');
+    if (category) {
+        filterTocType(category);
+    }
     if (isTopLink(target)) {
         scrollToTop(target);
     } else {
--- a/jsdoc-template/docs.html.handlebars
+++ b/jsdoc-template/docs.html.handlebars
@@ -42,7 +42,7 @@
             <li class="func" data-name="{{name}}" data-category="{{category}}">
                 <a href="#{{name}}">
                     {{name}}
-                    <span class="label label-category pull-right"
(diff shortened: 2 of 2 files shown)


CrossEye:
@wildermuthn: I'd love to get this included.  Could you fix the [linting error][tr]?

  [tr]: https://travis-ci.org/ramda/ramda/builds/48649900

davidchambers:
Could you revert the indentation changes? Ramda style is four spaces per "tab".

saikobee:
I was under the impression based on our discussion in the initial PR for the new docs that we wanted a dropdown list for filtering. Frankly I think that would be a lot more intuitive than clicking the labels, especially since only the table of contents labels are clickable, not the content area ones.

CrossEye:
I haven't even looked at the implementation.  I don't know what it does.  I agree that this is what we had discussed.  But I would not object to have several different ideas on the table for comparison.

wildermuthn:
@saikobee, I didn't notice #713 (it wasn't tagged as documentation). A drop-down is a good solution.

I made the search field case-insensitive for categories. It's a small but nice improvement. I'd recommend merging it in until #713 can be done.

wildermuthn:
Nevermind, Travis is giving me headaches.

davidchambers:
@wildermuthn, I fixed the formatting error (a missing space between a `,` and the following parameter name). It would be nice to revive this pull request by running the following commands:

```console
$ git checkout docs_filter_categories
$ git reset --hard 94e75ed9da353a1ec25aced0af358fa2b3806255
$ git pull git@github.com:ramda/ramda.git doc-categories
$ git push origin docs_filter_categories --force
```

wildermuthn:
Thanks, David.

mickeyvip:
Hello.

I think it is more intuitive to filter also the documentation pane by chosen category.

Thank you

donnut:
The current filter method of docs doesn't work as expected. Try searching for "function" and you get all results in category Function.

CrossEye:
You're right, this is still not quite right.  When I saw it, I found it an elegant solution, but the possibility of overlap between search term and category name is a serious issue.  My thought would be that the category filters should have a prefix in the search box, perhaps `cat:Function` or `category:Function`.:

    cat:Logic
    ----------------------
    allPass      Logic
    and          Logic
    anyPass      Logic
    cond         Logic
    defaultTo    Logic
    ifElse       Logic
    isEmpty      Logic
    not          Logic
    or           Logic

    co
    ----------------------
    charCodeAt   String
    commute      List
    commuteMap   List
    comparator   Function
    compose      Function
    composeP     Function
    concat       List
    cond         Logic
    construct    Function
    constructN   Function
    contains     List
    containsWith List
    converge     Function
    countBy      Relation


    cat:Function co
    ----------------------
    comparator   Function
    compose      Function
    composeP     Function
    construct    Function
    constructN   Function
    converge     Function

It would also be nice if the actual functions descriptions in the main panel were reduced to those matching the filter, not merely the navigation links.

How does that sound?

donnut:
This sounds good. Maybe clicking on a category label could copy this filter condition into the search box. Deleting it from the search box would then remove the filter condition again.

CrossEye:
Yes. that's what I was thinking.

wildermuthn:
The original request called for a drop-down, which I think is the best solution.

saikobee:
:+1: @wildermuthn

CrossEye:
I honestly have no issue with either the extension to the current technique or to a drop-down.  Anyone want to take a stab at it?

PR #769, opened 2015-01-28
restore aliases for reduce*

buzzdecafe:


CrossEye:
Note that this function is not currently dynamically dispatched, so it's not an issue, but when we want to connect dynamic dispatch with aliases, we're going to have to use separate calls to `checkForMethod` for each alias, or find a different way to do the dispatch.

buzzdecafe:
agreed. for now, i'd like to restore this and get 0.9.2 out

buzzdecafe:
... although that should be a simple matter of changing the implementation to:

```javascript
// in foldl.js:
module.exports = _curry3(_checkForMethod('foldl', _foldl));

// in reduce.js:
module.exports = _curry3(_checkForMethod('reduce', _foldl));
```

CrossEye:
Right, although that means we no longer have simple aliases.

    R.foldl === R.reduce; //=> false

buzzdecafe:
fair enough--i'm not sure if that is bad or not. would you expect `foldl` to dispatch to an object with a `reduce` method? seems a little weird

buzzdecafe:
*ping*

davidchambers:
I have no objections.

jschomay:
Thank you!

PR #771, opened 2015-02-03
curry functions from curry

buzzdecafe:
supersedes #766

### DIFF
--- a/test/curryN.js
+++ b/test/curryN.js
@@ -16,6 +16,48 @@ describe('curryN', function() {
         assert.strictEqual(curried(1, 2, 3), 6);
     });
 
+    it('will curry any returned function if its length > 1', function() {
+        function test(x, y, z) {
+            void z;
+            return source;
+        }
+        var curried = R.curryN(3, test);
+        var retval = curried(1, 2, 3);
+        assert.strictEqual(typeof retval, 'function');
+        assert.strictEqual(retval.length, 4);
+        assert.strictEqual(retval(1, 2, 3, 4), 6);
+        assert.strictEqual(retval(1, 2, 3)(4), 6);
+        assert.strictEqual(retval(1, 2)(3, 4), 6);
+        assert.strictEqual(retval(1)(2)(3)(4), 6);
+    });
+
+    it('will not curry any returned function if its length < 2', function() {
+        function test0(x, y) {
+            return function() { // length == 0
+                return y;
+            };
+        }
+        function test1(x, y) {
+            return function(z) { // length == 1
+                return z + y;
+            };
+        }
+        var curried0 = R.curryN(2, test0);
+        var curried1 = R.curryN(2, test1);
+        var retval0 = curried0(1, 2);
+        var retval1 = curried1(1, 2);
+
+        assert.strictEqual(typeof retval0, 'function');
+        assert.strictEqual(retval0.length, 0);
+        assert.strictEqual(retval0(), 2);
+        assert.strictEqual(retval0(33, 33, 33), 2);
+
+        assert.strictEqual(typeof retval1, 'function');
+        assert.strictEqual(retval1.length, 1);
+        assert.strictEqual(retval1(1), 3);
+        assert.strictEqual(retval1(33), 35);
+    });
+
     it('can be partially applied', function() {
         var curry3 = R.curryN(3);
         var curried = curry3(source);
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -2035,8 +2035,13 @@
                     throw _noArgsException();
                 }
                 var newArgs = _concat(args, arguments);
+                var value;
                 if (newArgs.length >= length) {
-                    return fn.apply(this, newArgs);
+                    value = fn.apply(this, newArgs);
+                    if (typeof value === 'function' && value.length > 1) {
+                        return curryN(value.length, value);
+                    }
+                    return value;
                 } else {
                     return recurry(newArgs);
                 }
--- a/src/curryN.js
+++ b/src/curryN.js
@@ -39,8 +39,13 @@ module.exports = _curry2(function curryN(length, fn) {
         return arity(Math.max(length - (args && args.length || 0), 0), function() {
             if (arguments.length === 0) { throw _noArgsException(); }
             var newArgs = _concat(args, arguments);
+            var value;
             if (newArgs.length >= length) {
-                return fn.apply(this, newArgs);
+                value = fn.apply(this, newArgs);
+                if (typeof value === 'function' && value.length > 1) {
+                    return curryN(value.length, value);
+                }
+                return value;
             } else {
                 return recurry(newArgs);
             }


test/curryN.js
```diff
@@ -16,6 +16,48 @@ describe('curryN', function() {
...
+        var retval = curried(1, 2, 3);
+        assert.strictEqual(typeof retval, 'function');
+        assert.strictEqual(retval.length, 4);
+        assert.strictEqual(retval(1, 2, 3, 4), 6);
+        assert.strictEqual(retval(1, 2, 3)(4), 6);
+        assert.strictEqual(retval(1, 2)(3, 4), 6);
+        assert.strictEqual(retval(1)(2)(3)(4), 6);
+    });
+
+    it('will not curry any returned function if its length < 2', function() {
```
davidchambers: I would write `<= 1` here (but perhaps that's just me).

davidchambers:
:deciduous_tree:

buzzdecafe:
@CrossEye any comment?

CrossEye:
Sorry, flat on my back, with only 4G access, and horrible keyboard.  Dr.'s appt this afternoon.  I'm still worried that some variant of what I suggested for the `compose` version could end up haunting us (https://github.com/ramda/ramda/issues/736#issuecomment-70555440)

The simple question is what happens if the returned now-curried function has named optional parameters?  I really want to do this, but I'm haunted by the this possibility, even if it seems obscure.  Actually maybe even more because it seems obscure.  If we choose to do this, we need to recognize the fact that if a user returns a built-in function from such a call to `curry`, and one engine later decides to add an additional parameter to it -- this does happen! -- then Ramda has broken her application for her, even though she was careful to heed our warning about not returning a function with optional parameters from curry.  How was she to know what that dastardly Mozilla was going to do six months later?

If on the other hand, she had taken the return function and used it directly or even curried it to her expected length, things would continue to work.

The trouble is, for the life of me, I'm unable to come up with a reasonable example of returning a function from a call to curry that itself has named optional parameters.  So I'm wondering if I'm smoking something funny.

buzzdecafe:
that seems like a pretty obscure edge case to me. we should document that possibility and proceed. 

So i guess I'll update the `curry/curryN` docs to reflect that.

CrossEye:
The obscure possibility, I think, is the engine or someone else adding a named parameter to a function one is already using.

But in Javascript, named optional parameters are certainly not obscure.  That's what worries me.  Do we have good reason to assume that functions returned as the result of a curry call -- already an edge case, I realize -- would not have any named optional parameters?  If we don't, then we will be breaking things for users.  _Caveat emptor_ only goes so far.  I won't get a kick out of the doctor responding to my "It hurts when I do _this_" with "Then don't do _this_", since _this_ is breathing.  If we choose to auto-curry these, there is no opt-out.  

As I said, I do want to do it.  But I do have have a strong concern.  I just cant' figure out if it's a legitimate worry about a potential show-stopping, don't-ever-use-that-piece-of-shit-library bug, or a if-you-squint-just-right-while-the-stars-are-aligned-on-a-full-moon-in-a-leap-year potential problem. :smile:

buzzdecafe:
isn't this already a problem if you `curry` a function with named optional params? 

```javascript
var cf = R.curry(function(a, b, c) {
    return a + b + (c || 0); 
});
cf.length; //=> 3
```

Our advice there is "use `curryN`".

CrossEye:
This is a problem if **you** curry a function with a named optional param.  But you have the option to choose `curryN` instead.  Here we're talking automated magic.  The thing that comes out of the `curry` meatgrinder is, honestly, rarely a function.  But when it is, at the moment, it's not curried.  The user would have to decide to curry it herself, with `curry` or `curryN`, if she so chose.  This change would mean that if it is a function, it will automatically be curried to the function's arity, and she no longer has that choice.  That's my concern.

buzzdecafe:
> she no longer has that choice. 

On the contrary, the same advice holds, you can still `curryN` the whole shmear. Since this is such a edge-of-an-edge case, that seems an acceptable workaround to me. It certainly isn't worth holding up useful functionality for.

CrossEye:
Currying the whole shmear [doesn't work][re]:

    var fullName = function(first, last, suffix) {
        return first + ' ' + last + (suffix ? ', ' + suffix : '');
    };
    var f = R.curry(fullName);
    var g = R.curryN(2, f);
    g('Lon', 'Cheney', 'Jr'); //=> 'Lon Cheney, Jr.'
    g('Boris', 'Karloff'); //=> :: String -> String

I may be missing something.  Is this something that people are clamoring for?  I thought this was written mostly on spec.  I was thinking this was a nice-to-have.  I know there was a request to curry the result of `wrap` and another for `compose`.  The only request I knew of for this was the very interesting post from Aadit Shah.  Was there more?

  [re]: http://bit.ly/1CvIokx

buzzdecafe:
It *is* written on spec/nice-to-have. I guess I am more interested in exploring the possibilities at the risk of breaking bad behavior (i.e. named optional params) than foreclosing possibilities in order to protect bad behavior.

CrossEye:
I think we need to start deciding exactly how opinionated we want Ramda to be.  Right now, it works reasonably well with most built-in data structures, but would fail with sparse arrays; we have no problem saying simply "don't do that" for those.  It works well with most fixed-arity functions and gives tools to help work with variadic ones.  It works to some extent with functions that have named optional parameters, but you'd better use `curryN` with the correct `N`.

Ramda itself expresses plenty of strong opinions in its source code: no variadic functions, auto-curry everything, no mutation of user data, very little type-checking; everything built around developing simple functional pipelines.

In other words, so far, Ramda is fairly opinionated mostly in its own API, not in the kind of code a user would write in order to work with Ramda.

We need to decide if we want to move to become more opinionated in the latter sense.  I don't necessarily have an objection.  But it needs to be a conscious decision.  Such things would need to be documented.  And such documentation might be tricky.  And we need to think through the implications very clearly.  What sorts of functions would not work here, not just user functions, but native ones, DOM and other host ones, popular third-party libraries that people might want to have interoperate?

In essense, if we're currying when a user hasn't specifically asked us to do so, we're responsible for changing her function.  So we need to make sure that she knows it will happen.  And we have to make sure it's not particularly likely to break things.  In the `compose` and `wrap` cases, I think we can document it.  But because `curry` is used everywhere, it might be harder to do so.  (Even `compose` scares me a bit.)  That's my concern:  as a user, I might use some function without necessarily even knowing that it has certain additional parameters that I don't care about (`parseInt` has arity 2, for instance) and when it's returned from my Ramda-curried function it no longer works as expected.

davidchambers:
I'd like Ramda to be more opinionated than it currently is. There will be times when a user must wrap an optional-arg function in order to conform to Ramda's expectations, but I'm okay with that. Think of it as a foreign function interface of sorts. :)

CrossEye:
All right, I pulled together an old block of code to demonstrate my worries.

I simplified and Ramda-ized something I used some years back skipping various unrelated parts.  It's an automated table-sorter.  (The original also allowed you to define your own types, override the automated registration of types for a column, and define how the types fit into a type-tree; it also had a custom stable-sort to allow sorting on multiple columns.  But the rest of the API is fairly intact.)  The idea is that this attempts to parse the table data to choose the right type for the column so that you can simply sort on the column number.

You can see it in action at http://tinyurl.com/ot5465d.  (The output is in the console.)

There's a fair bit of code there, but the most relevant bit is the definition of the types:

```javascript
var types = {
    string: {
        test: R.T,
        sortKey: R.identity
    },
    decimal: {
        test: R.match(/^[+-]?((\d+(\.\d*)?)|(\.\d+))$/),
        sortKey: parseFloat
    },
    integer: {
        test: R.match(/^[+-]?[1-9]\d*$/),
        sortKey: parseInt
    },
    isoDate: { // not really accurate
        test: R.match(/^\d{4}-\d{2}-\d{2}$/),
        sortKey: R.identity
    },
    semver: { // again only an approximation
        test: R.match(/^\d\d?(\.\d\d?(\.\d\d?)?)?$/),
        sortKey: R.pipe(R.split('.'), R.foldl(R.useWith(R.add, R.multiply(100), R.I), 0))
    }
};
```

and the code to sort the rows according to a list mapping each column to one of these types (`columnTypes`), and a column number:

```javascript
var sortByColumn = R.curry(function(columnTypes, colNbr, data) {
    return R.sortBy(R.pipe(R.nth(colNbr), columnTypes[colNbr].sortKey), data);
});
```

This function works fine with Ramda the way it is.  If I were to mimic the curry-the-function-returned-from-curry notion in this, via:  

```javascript
var sortByColumn = R.curry(function(columnTypes, colNbr, data) {
    return R.sortBy(R.pipe(R.nth(colNbr), R.curry(columnTypes[colNbr].sortKey)), data);
});
```

(see http://tinyurl.com/n446zeh, also output to the console)

then when sorting by the last column, it does not sort properly.  Because we're in the context of this discussion, it's easy enough to see that the problem is simply that `parseInt` (arity 2) is being curried and then called with only a single value inside `sortBy`.  Then `sort` is called with two identical keys (the partially applied function) and so no actually sorting is done.  But in the wild, finding that, understanding it, and avoiding the temptation to bash the heads of the library authors could be rather difficult.  ("You're telling me that `R.pipe(R.nth(colNbr), sortKey)` can't be passed to `R.sortBy` because that particular `sortKey` has an optional parameter I've never used and don't care about?"  What are you smoking?") 

My concern is that this is not some obscure corner case, but a common way people use Javascript, and, even for those doing more FP, it's something that could sneak into an application, perhaps without them noticing.  So documentation that this is not the best usage of Ramda might not help.  Debugging would be difficult.

Please note that I actually still want to do this.  But I'm trying to think of ways to make this less of a potential issue for users.

davidchambers:
Have you two reached consensus on this? I don't have a strong opinion.

CrossEye:
Don't think so.   Usually we are able to hash some of this stuff out face-to-face, but illness has prevented that.   I should be back in the office Monday.  Maybe we'll be able to clear it up then.

buzzdecafe:
Yes and no. @CrossEye has presented a persuasive argument. I would still love to find a way to square this circle. Pretty frustrating.

davidchambers:
((:bell:)) @buzzdecafe

buzzdecafe:
i think we can close for now, but i expect to circle back to this in future

CrossEye:
I still want to address this, and

PR #775, opened 2015-02-05
use consistent names

davidchambers:
Currently the project mixes three styles:

  - lower-case prefix (e.g. `R.pPipe`)
  - lower-case suffix (e.g. `R.foldl`)
  - upper-case suffix (e.g. `R.liftN`)


buzzdecafe:
`foldL/foldR` just feels *wrong*

TheLudd:
If this change were to be made, please make sure that the API is not broken between two minor versions without any deprecation notice.

honzabrecka:
@TheLudd Until the 1.0.0 is released, the API can be changed without any deprecation notice.

TheLudd:
@honzabrecka That is very correct. But I would say it is still a very bad thing to do which is why we have #769

buzzdecafe:
:+1: @TheLudd it would be nice to help those early adopters, even though it is not _required_

CrossEye:
This is starting to make me think again about a mailing list or forum tool where we could announce such changes before they go live.   Anyone have recent experience setting something up?  Would we want the extras hassle?

buzzdecafe:
expanding on my uneasiness about `foldl`: internal consistency is fine, but we should beware of breaking with expectations from the outside world as well (see `reduce`). I have never seen `foldl` with a capital `L` and would find that very surprising indeed. The other changes don't strike me as problematic.

davidchambers:
> expanding on my uneasiness about `foldl`: internal consistency is fine, but we should beware of breaking with expectations from the outside world as well (see `reduce`). I have never seen `foldl` with a capital `L` and would find that very surprising indeed.

I find it hard to justify `R.foldl` alongside `R.scanL`, `R.partialL`, and `R.mapAccumL`. Although conforming to initial expectations is a good idea, if every API were to do so religiously each would, at best, be as consistent as its predecessor. Entropy would increase over time.

We are not wedded to the names of JavaScript built-ins or Haskell functions or anything else. Ramda's names should be right in the context of Ramda. I don't mind learning an API which doesn't conform to my initial expectations if it's internally consistent. After learning a portion of the system I'll be able to make assumptions—correct assumptions—about how the rest of the system works. Ramda's argument ordering is a good example of this phenomenon.

buzzdecafe:
@davidchambers I agree with you -- 99% :smile: `foldL` is giving me heartburn. i think it makes more sense to *drop* `foldl` for `reduce` than go with `foldL`.

davidchambers:
> it makes more sense to *drop* `foldl` for `reduce` than go with `foldL`.

Works for me! :grinning:

davidchambers:
Would we also change `R.unfoldr` to `R.unreduceRight`? :stuck_out_tongue_winking_eye:

buzzdecafe:
> Would we also change R.unfoldr to R.unreduceRight?

good yet whimsical question! 

i'd be happy with `unfoldr` becoming simply  `unfold`

davidchambers:
Where do you stand, @CrossEye?

CrossEye:
`unfold` sounds good to me.  And I think `reduce` is the winner here.  That would save us the alias heartburn for another day.  Besides that, LGTM.

davidchambers:
Before I make any changes, let's clarify what you're suggesting. Here's my understanding:

  - `R.foldl` → `R.reduce`
  - `R.foldr` → `R.reduceRight`
  - `R.foldlIndexed` → `R.reduceIndexed`
  - `R.foldrIndexed` → `R.reduceRightIndexed`
  - `R.unfoldr` → `R.unfold`
  - `R.lPartial` → `R.partial`
  - `R.rPartial` → `R.partialRight`
  - `R.mapAccumL` → `R.mapAccum`
  - `R.mapAccumR` → `R.mapAccumRight`
  - `R.scanl` → `R.scan`
  - `R.pCompose` → `R.composeP`
  - `R.pPipe` → `R.pipeP`

For consistency `R.unfold` should be `R.unfoldRight`. From the [Haskell docs][1]:

> In some cases, `unfoldr` can undo a `foldr` operation

We're already breaking the fold–unfold symmetry, though, so perhaps this isn't necessary.

I prefer the naming scheme employed by this pull request as it currently stands. If others feel strongly that the `verb`/`verbRight` scheme is preferable I will update this pull request. Consistency is valuable even if the names are not quite to my liking. :)


[1]: https://hackage.haskell.org/package/base-4.7.0.2/docs/Data-List.html#v:unfoldr

buzzdecafe:
_http://stackoverflow.com/questions/23779494/why-the-r-in-unfoldr

I think `unfold` is fine. otherwise, :cow2:

CrossEye:
:herb:

davidchambers:
@CrossEye, does the `:herb:` apply to my comment above or to the pull request in its current form?

CrossEye:
I was responding to your comment; and somehow I missed the final paragraph of it.  I like the names in the comment better than those in the PR.  But I agree that consistency is better than any particular set of names.  I think we learned recently that users are much more invested in `reduce` than `fold`, and if we're not going to have the alias, then the `reduce` versions in this list are probably for the best.

davidchambers:
Okay. I'll update the pull request now.

yuri:
As an alternative to `.pipeP` and `.composeP`, perhaps `.pipeAsync` and `.composeAsync`?

CrossEye:
@yuri: I think `.pipeP`/`.composeP` and/or `pipePromise`/`composePromise` is more clear.  These functions handle only Promises, not the full range of ways in which one can do asynchronous processing in Javascript, including WebWorkers, Server Push, and `setTimeout`.  (Yes, I know that Promises can be wrapped around any of these, but then we're back to dealing with Promises.)

yuri:
@CrossEye The downside of of "composeP" is that it's not immediately clear what "P" stands for. "composePromise" is more explicit, but seems inaccurate, since this function does not "compose a promise". Based on how "Async" suffix is normally used in JavaScript today, I think "composeAsync" will lead people to expect a function that either works with callbacks or with promises. I don't think anyone would expect any of the other behaviours from a functions named this way. Since "compose" and "pipe" are variadic, there is no straightforward way of using callbacks with them, so I think the fact that composeAsync works only with promises shouldn't surprise anyone.

CrossEye:
> "composePromise" is more explicit, but seems inaccurate, since this function does not "compose a promise"

Neither does `reduceRight` reduce a right, but it might still be the best fit for our naming conventions.

However, I don't feel strongly about this, and I don't really object to `Async`.  I really want to get our naming settled down soon, though!

Do others have opinions?

davidchambers:
> Do others have opinions?

Not on this one. All three options seem reasonable.

buzzdecafe:
:no_mouth:

CrossEye:
Let's leave them for now and see if we get questions about them.  If they seem confusing, we can address it again.

PR #776, opened 2015-02-05
Fix issue where memoize does not work for falsy values

peterjoel:
Falsy values (e.g. null, 0, empty string) were not being memoized by memoize().
Additionally, there was an unnecessary conversion of arguments to an array.

### DIFF
--- a/test/memoize.js
+++ b/test/memoize.js
@@ -30,4 +30,21 @@ describe('memoize', function() {
         assert.strictEqual(identity('x'), 'x');
         assert.strictEqual(identity('y'), 'y');
     });
+
+    it('can memoize falsy values', function() {
+        var ctr = 0;
+        var f = R.memoize(function(x) { ctr++; return '' + x; });
+        var values = [0, false, null, undefined, '', NaN];
+        checkFunctionIsIdentityOn(values);
+        // check function has been called once per value
+        assert.equal(values.length, ctr);
+        checkFunctionIsIdentityOn(values);
+        // check function has not been called again for any of the values
+        assert.equal(values.length, ctr);
+        function checkFunctionIsIdentityOn(values) {
+            for (var i = 0; i < values.length; i++) {
+                assert.equal('' + values[i], f(values[i]));
+            }
+        }
+    });
 });
--- a/src/memoize.js
+++ b/src/memoize.js
@@ -42,9 +42,9 @@ module.exports = function memoize(fn) {
     return function() {
         if (!arguments.length) {return;}
         var position = _foldl(function(cache, arg) {
-            return cache[arg] || (cache[arg] = {});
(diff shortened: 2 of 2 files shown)


buzzdecafe:
makes sense, thanks.

davidchambers:
You branched from a very old commit. The files you modified no longer exist. Please pull from master and rebase your changes.

peterjoel:
That's very odd. I forked it from master today for the first time. Anyway... I'll re-submit the pull request now.

test/memoize.js
```diff
@@ -30,4 +30,25 @@ describe('memoize', function() {
...
+        var f = R.memoize(function(x) { ctr++; return x; });
+        var values = [0, false, null, undefined, '', NaN];
+        checkFunctionIsIdentityOn(values);
+        // check function has been called once per value
+        assert.equal(values.length, ctr);
+        checkFunctionIsIdentityOn(values);
+        // check function has not been called again for any of the values
+        assert.equal(values.length, ctr);
+        function checkFunctionIsIdentityOn(values) {
+            values.forEach(function(v) {
```
buzzdecafe: this will fail in ES3

peterjoel: I'm new to JavaScript. Is Array.forEach() not supported in ES3?

Peter

davidchambers: That's right. You could use a `for` loop instead.

CrossEye: There has never been a really strong reason to support older engines running ES3, especially as ES6 is just around the corner, _except_ it has generally been fairly easy to do.  Since there are still some remote corners of the web running on ES3, being able to support them is a nice little bit of bragging rights.  I would definitely fix this, as it's still pretty simple, but at some point, we're going to decide how long we want to keep the ES3-compliance.

peterjoel: Ok, no worries. I've forced pushed to my fork, so that should have updated
in the pull request now.

Peter

CrossEye:
I don't see anything wrong in the Travis output, just the final failure message.

Can someone who understands the output from the `make` command better than me explain [what's wrong][tb] with the build?  Anyone?  Anyone?  Chambers?  :smile:

  [tb]: https://travis-ci.org/ramda/ramda/builds/49688498

buzzdecafe:
i think the built artifact has to be checked in.

davidchambers:
> i think the built artifact has to be checked in.

That's exactly right.

CrossEye:
Oh, of course.  I really hate that.  Someday I'll get around to figuring out how to get Travis to do that part for us.

davidchambers:
Thank you for bringing this to our attention. I was eager to get a fix in, so I opened (and have since merged) #785 which addresses the issues that prevented us from merging this pull request.

Issue #777, opened 2015-02-05
Add a "merge all" function


benperez:
I commonly find myself wishing that there was a function which acted like `R.mixin` but with an entire list of objects. Here's the type signature, implementation.
```javascript
// R.blend :: [{}] -> {}
R.blend = R.foldl(R.mixin, {});
```
Often I am constructing objects for things like database queries from a list of smaller objects.
```javascript
var shouldUpdateFoo = true;
var shouldUpdateBar = false;
var shouldUpdateBaz = true;
var updates = R.blend([
  shouldUpdateFoo ? {foo: 1} : {},
  shouldUpdateBar ? {bar: 2} : {},
  shouldUpdateBaz ? {baz: 3} : {},
]);
// {foo: 1, baz: 3}
```
For me, `R.blend` would conceptually be a cousin of `R.sum`, `R.product`, and `R.flatten`. These functions reduce monoids under the following operators
`R.sum` —> addition with identity element 0
`R.product` —> multiplication with identity element 1
`R.flatten`  —> does more than just this but can be thought of as concatenation with identity element []
`R.blend` —> `R.mixin` with identity element `{}`

davidchambers:
I would use such a function quite often. I'm not sure about the name, though. :P

benperez:
I like the name because it fits well with `R.mixin`, which already sounds like a baking instruction. You can either `R.mixin` one ingredient at a time, or you can throw them all in together and hit `R.blend`.

davidchambers:
That's it, I'm sold!

CrossEye:
:+1:

buzzdecafe:
now we'll have to add functions for grate, liquefy, puree, and makeJulienneFries

PR #778, opened 2015-02-06
Add mergeAll function and tests

benperez:
Fixes https://github.com/ramda/ramda/issues/777

### DIFF
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -6053,6 +6053,23 @@
     });
 
     /**
+     * Merges a list of objects together into one object.
+     *
+     * @func
+     * @memberOf R
+     * @category List
+     * @sig [{k: v}] -> {k: v}
+     * @param {Array} list An array of objects
+     * @return {Object} A merged object.
+     * @see reduce
+     * @example
+     *
+     *      R.mergeAll([{foo:1},{bar:2},{baz:3}]); //=> {foo:1,bar:2,baz:3}
+     *      R.mergeAll([{foo:1},{foo:2},{bar:2}]); //=> {foo:2,bar:2}
+     */
+    var mergeAll = reduce(merge, {});
+
+    /**
      * Divides the second parameter by the first and returns the remainder.
      * Note that this functions preserves the JavaScript-style behavior for
      * modulo. For mathematical modulo see `mathMod`
@@ -6247,6 +6264,7 @@
         maxBy: maxBy,
         memoize: memoize,
         merge: merge,
+        mergeAll: mergeAll,
         min: min,
         minBy: minBy,
         modulo: modulo,
--- a/src/mergeAll.js
+++ b/src/mergeAll.js
@@ -1,0 +1,20 @@
+var merge = require('./merge');
+var reduce = require('./reduce');
+
+
+/**
+ * Merges a list of objects together into one object.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig [{k: v}] -> {k: v}
+ * @param {Array} list An array of objects
+ * @return {Object} A merged object.
+ * @see reduce
+ * @example
+ *
+ *      R.mergeAll([{foo:1},{bar:2},{baz:3}]); //=> {foo:1,bar:2,baz:3}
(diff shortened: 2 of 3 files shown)


src/blend.js
```diff
@@ -0,0 +1,19 @@
...
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig [{k: v}] -> {k: v}
+ * @param {Array} list An array of objects
+ * @return {Object} A merged object.
+ * @see foldl
+ * @example
+ *
+ *      R.blend([{foo:1},{bar:2},{baz:3}]); //=> {foo:1,bar:2,baz:3}
```
davidchambers: It would be nice to demonstrate whether left trumps right or vice versa by having two objects share a key.

test/blend.js
```diff
@@ -0,0 +1,10 @@
+var assert = require('assert');
+
+var R = require('..');
+
+
+describe('blend', function() {
+    it('merges a list of objects together into one object', function() {
+        assert.deepEqual(R.blend([{foo:1}, {bar:2}, {baz:3}]), {foo:1, bar:2, baz:3});
+    });
+});
```
davidchambers: We should probably include a test for inherited properties:

```javascript
function Foo() {}
Foo.prototype.bar = 42;
var foo = new Foo();
var res = R.blend([foo]);
res === foo;  // ???
R.keys(res);  // ???
```

src/blend.js
benperez: fixed :zap:

buzzdecafe:
:+1:

davidchambers:
I like this function. I take issue with its name (though I was greatly amused by the baking analogy). I've never liked the name `R.mixin`, either. The word "mixin" is a noun; the verb is "mix in". `R.mixIn` looks ugly, though. Also, when I hear "mixin" I think OOP.

I suggest the following changes:

  - `R.mixin` → `R.merge`; and
  - `R.blend` → `R.mergeAll`.

buzzdecafe:
do we have a standard suffix for lifting a function to list context?

davidchambers:
> do we have a standard suffix for lifting a function to list context?

We do have `R.insert` and `R.insertAll`.

buzzdecafe:
:cow2:

CrossEye:
@davidchambers You've mentioned before not liking the name "mixin".  I don't care much.  I've seen a lot of names for this, and none of them seems to be particularly dominant.  If you've a strong feeling about it, let's do it now, before we hit 1.0.

davidchambers:
@benperez, please rename the function `R.mergeAll`.

benperez:
Updated

davidchambers:
:deciduous_tree:

jdalton:
Just checking back in on this method. Have folks been happy with the signature `R.mergeAll(array)`?
I'm wondering if there's something to `R.mergeAll(destObj, srcArr)`.

davidchambers:
> I'm wondering if there's something to a signature like `R.mergeAll(destObj, srcArr)`.

`R.reduce(R.merge, destObj, srcArr)` should achieve this, shouldn't it? If you're asking whether there's a function which mutates one of its arguments, the answer is no.

jdalton:
Not about mutating. Just similar to other `All` methods which are like their non-All forms except applying an array of values instead of a single value. The `mergeAll` is a bit different than the other All-forms.

davidchambers:
It doesn't seem too bad to me to write `R.mergeAll(R.prepend(destObj, srcArr))` if one happens to have a single object and an array of objects. I don't think this warrants a dedicated function. Others may disagree, of course.

jdalton:
Yep, I know there are other ways to accomplish things. I was checking in to see if you all had gotten feedback on the signature and thoughts about it related to the other All-forms.

CrossEye:
@jdalton:

Just returning from two-weeks away.  I don't recall any such feedback.

I also don't really think we have any standard behavior for the `All` suffix.  The three functions I can think of with that suffix, `insertAll`, `mergeAll`, and `pickAll` have three different sorts of semantics.

PR #780, opened 2015-02-06
add support for io.js

honzabrecka:
Do we want to support io.js? A few tests are failing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: node_js
 node_js:
 - '0.10'
 - '0.12'
+- 'io.js'
 env:
   global:
   - secure: dgshxCQaY6olKOZnD+BT8R1Vp8uJhwCIUlSnoFji1m+7KIy3zwbbgi/2jkf/SWBCxzlpzDBWUuWbaMJOrJHWfjUaX3JfUY+2gbAPunrofD+42+brfo506rIxr20yaoomsnLxXFLgleLW5FuaezP6KP4lW5IOlWZXOEv0vTvoxQk=


buzzdecafe:
i think we should support IO. which tests are failing?

honzabrecka:
`invert` and `invertObj`

I'll check it why it happens.

buzzdecafe:
looks like it is failing on the "throws on zero args" test

honzabrecka:
`assert.throws` should be used as follows:

```
assert.throws(function() { R.invert(1); });
```

but the problem is, that neither `R.invert(1)` and `R.invert(null)` throw an error, it always return `{}` for those special cases.

So each test should check if result is `typeof result == 'object'`.

buzzdecafe:
IMO `R.invert(1)` and `R.invert(null)` *should* throw an error. So the test is correct. I wonder why it fails on one platform but not the other

honzabrecka:
For me it do not throw an error on any platform, it returns an `{}`.

buzzdecafe:
hmmm. then how did those tests *ever* pass?

```
        assert.throws(R.invertObj(1));
        assert.throws(R.invertObj('1'));
        assert.throws(R.invertObj(null));
        assert.throws(R.invertObj(undefined));
 
```

honzabrecka:
Look at http://nodejs.org/api/assert.html#assert_assert_throws_block_error_message. io.js has better error messages and is more strict, so it expects function as assert.throws first argument.

The `assert.throws(R.invertObj(1))` haven't ever worked, because `R.invertObj(1)` is simply evaluated before the assertion. This is how `assert.throws` works inside:

```
assert.throws = function(f) {
  try {
    f();
    assert.fail('expected error');
  } catch (error) {}
}
```

buzzdecafe:
yes, gotta wrap those tests in lambdas. thanks for catching this

honzabrecka:
I'm curious if that tests will pass. :)

buzzdecafe:
i doubt it, not without rewriting the source

honzabrecka:
Updated to actual master.

davidchambers:
:deciduous_tree: Thanks!

Issue #773, opened 2015-02-04
`assoc` should work on lists as well as plain objects


CrossEye:
Right now:

    var arr = ['a', 'b', 'c', 'd', 'e'];
    R.assoc('2', 'xyz', arr);
    //=>{0: "a", 1: "b", 2: "xyz", 3: "d", 4: "e"}
 
To my mind, this should really return

    //=> ['a', 'b', 'xyz', 'd', 'e']

I'm not sure how straightforward that might be, as the current implementation is concerned with flattening the (enumerable part of the) prototype chain into the newly created object.  But that change looks to me to be the correct API.

davidchambers:
Isn't this what [`R.insert`][1] does?


[1]: http://ramdajs.com/docs/#insert

CrossEye:
Not quite.

    R.insert(2, 'xyz', x); //=> ['a', 'b', 'xyz', 'c', 'd', 'e']

I'm suggesting

    R.assoc(2, 'xyz', x); //=> ['a', 'b', 'xyz', 'd', 'e']

Which is similar to 

    var x = {a: 1, b: 2, c: 3, d: 4, c: 5}
    R.assoc('c', 42, x); //=> {a: 1, b: 2, c: 42, d: 4, c: 5}

CrossEye:
I guess what I'm suggesting for the array case is the equivalent of 

    R.assoc(idx, val, arr) => clone(arr).splice(idx, 1, val);

davidchambers:
Ah, yes. Seems reasonable to me. :)

davidchambers:
Would `R.assoc(99, 'xyz', ['a', 'b', 'c', 'd', 'e'])` produce a sparse array?

How would `R.assoc(-1, 'xyz', ['a', 'b', 'c', 'd', 'e'])` evaluate?

How about `R.assoc('foo', 'xyz', ['a', 'b', 'c', 'd', 'e'])`?

buzzdecafe:
so the signature of `assoc` is contingent upon its arguments? seems bad to me

CrossEye:
All good questions @davidchambers.

I'm mostly concerned about the sparse array issue.  The negative index I would expect should be handled like we handle negative array indices elsewhere, as counters from the end.  And adding an arbitrary property to an Array instance is already quite possible:

    var x = [1, 2, 3];
    x.foo = 'bar';

[...]

Issue #637, opened 2014-12-03
getIn (prop keypaths)


tolmasky:
Facebook's immutable has a nifty getIn method for grabbing a value at a key path:

```javascript
myMap.getIn(["a", "b", "c"], defaultValue)
```

I find myself piping together lots of prop calls to accomplish something similar:

```javascript
R.map(R.pipe(R.prop("id"), R.prop("name")), vars)
```

I made for myself something like getIn:

```javascript
R.map(R.getIn(["id", "name"]), vars)
```

If this is something others could want I could make a pull request.

CrossEye:
@tolmasky Do you want to create a PR for this?

tolmasky:
Hi yeah sorry, I need to split this out from my stuff and submit it. I'll
attempt do it this weekend

On Thursday, January 8, 2015, Scott Sauyet <notifications@github.com> wrote:

> @tolmasky <https://github.com/tolmasky> Do you want to create a PR for
> this?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ramda/ramda/issues/637#issuecomment-69286299>.
>


-- 
Francisco Tolmasky
www.tolmasky.com
tolmasky@gmail.com

tolmasky:
Just providing an update, this touched a couple other functions so I'm just updating all the tests.

CrossEye:
Thanks for the update.

I look forward to seeing it (even if a little sad about losing the old API.)

CrossEye:
Moving this to 0.10.0.  I don't think anyone is desperately waiting on it, and I think we really want to publish 0.9.0.

jethrolarson:
I like the string sugar. I can see both approaches having their use though.
I suppose if you implement the array then the string version can just be `dotPath = compose(path, split('.'))`

tolmasky:
Sorry, know this is taking me forever. But another thing that comes to mind
is that there seems to be no reason to not just support both:

getin(path)
{
if (typeof path === string) return getin(split(“.”))
else …
}

Seems then it works the way everyone expects.


On Fri, Feb 6, 2015 at 3:46 PM, Jethro Larson <notifications@github.com>
wrote:

> I like the string sugar. I can see both approaches having their use though.
> I suppose if you implement the array then the string version can just be dotPath
> = compose(path, split('.'))
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ramda/ramda/issues/637#issuecomment-73334325>.
>



-- 
[...]

PR #781, opened 2015-02-06
add R.dissoc

davidchambers:
The inverse of `R.assoc`; very similar to `R.omit`. See [`clojure.core/dissoc`][1].

[1]: http://clojuredocs.org/clojure.core/dissoc

### DIFF
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -2123,6 +2123,26 @@
     });
 
     /**
+     * Returns a new object that does not contain a `prop` property.
+     *
+     * @func
+     * @memberOf R
+     * @category Object
+     * @sig String -> {k: v} -> {k: v}
+     * @param {String} prop the name of the property to dissociate
+     * @param {Object} obj the object to clone
+     * @return {Object} a new object similar to the original but without the specified property
+     * @example
+     *
+     *      R.dissoc('b', {a: 1, b: 2, c: 3}); //=> {a: 1, c: 3}
+     */
+    var dissoc = _curry2(function dissoc(prop, obj) {
+        return _pickBy(function (val, key) {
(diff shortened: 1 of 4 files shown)


CrossEye:
Makes sense to me.   Of course we've got an issue open (#773) on slightly changing the behavior of `assoc` to work a bit more like `clone`; I'm not sure if it'll get approved.  @buzzdecafe is offering some stern (and reasonable) resistance.  But that behavior wouldn't make so much sense here, which might be the death-knell of that issue, I suppose.

But do we also want a `dissocPath` in parallel with `assocPath`?

    dissocPath('d.e', {a: 1: b: {c: 2}, d: {e: {f: 3, g: 4}, h: 5},i: 6});
    //=> {a: 1, b: {c: 2}, d: {h: 5}, i: 6}

? Or what I imagine will eventually be the API (#637), `dissocPath(['d', 'e'], obj)`?

buzzdecafe:
`*socPath` have a nested-type problem, e.g.: `assocPath('a.b.1.c', 'moo', {a: b: [ {}, {c: 'cow'} ] }) //=> ???`

buzzdecafe:
however, `dissoc` LGTM

:cake:

CrossEye:
@buzzdecafe:

In the current version, it returns 

    {a: {b: {0: {}, 1: {c: 'moo'} } }})

If we follow through on #773, it should have:

    {a: {b: [ {}, {c: 'moo'} ] }})

At least that's what seems reasonable to me.

davidchambers:
If there's consensus for this change, let's merge it and discuss `R.dissocPath` in a separate pull request.

buzzdecafe:
agreed -- LGTM

PR #782, opened 2015-02-07
Fixing lift to work with other Applicatives

CrossEye:
Somewhere along the way, `lift` got broken.  This is just a quick fix and the addition of some regression tests.

To do this, though, I added the `Maybe` type from ramda-fantasy to `test/shared`.

Obviously my git-fu is still a bit lacking as the commit messages are a bit screwed up.

### DIFF
--- a/src/liftN.js
+++ b/src/liftN.js
@@ -1,10 +1,10 @@
 var _ap = require('./internal/_ap');
 var _curry2 = require('./internal/_curry2');
 var _foldl = require('./internal/_foldl');
-var _map = require('./internal/_map');
 var _noArgsException = require('./internal/_noArgsException');
 var _slice = require('./internal/_slice');
 var curryN = require('./curryN');
+var map = require('./map');
 
 
 /**
@@ -31,6 +31,6 @@ module.exports = _curry2(function liftN(arity, fn) {
         throw _noArgsException();
     }
     return curryN(arity, function() {
-        return _foldl(_ap, _map(lifted, arguments[0]), _slice(arguments, 1));
+        return _foldl(_ap, map(lifted, arguments[0]), _slice(arguments, 1));
     });
 });
--- a/test/lift.js
+++ b/test/lift.js
@@ -3,6 +3,7 @@
 var assert = require('assert');
 
 var R = require('..');
+var Maybe = require('./shared/Maybe');
 
 
 var add3 = R.curry(function add3(a, b, c) {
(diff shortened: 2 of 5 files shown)


davidchambers:
> Obviously my git-fu is still a bit lacking as the commit messages are a bit screwed up.

You can edit the commit message by running `git commit --amend`.

test/lift.js
```diff
@@ -40,4 +41,10 @@ describe('lift', function() {
     it('throws on zero arguments', function() {
         assert.throws(R.lift);
     });
+
+    it('works with other functors such as "Maybe"', function() {
+        var addM = R.lift(R.add);
+        assert.deepEqual(addM(Maybe(3), Maybe(5)), Maybe(8));
+    });
```
davidchambers: I'd rather use a simpler type here. Something like [Id][1] but with only the required methods implemented.


[1]: https://github.com/fantasyland/fantasy-land/blob/master/id.js

CrossEye: While I don't mind switching to Id, my thought was to be able to use this for multiple tests across many different functions, so it isn't clear to me what functions are really necessary.  Even here, for the current implementation of `lift` we need `ap` and `map`.  But it might well be reasonable to consider reimplementing it in terms of `of` and `ap`.  I wouldn't want to tie our tests too closely to our current implementation.

davidchambers: Good point. I had not considered the possibility of multiple implementations.

davidchambers:
You may not be fussed, but the commit date is 17 days ago. If you'd like to correct it, you could run `git commit --amend --date="$(date -R)"` to set the commit date to the current date.

Aside from that minor point, LGTM.

CrossEye:
@buzzdecafe: Any input?

src/liftN.js
```diff
@@ -31,6 +31,6 @@ module.exports = _curry2(function liftN(arity, fn) {
         throw _noArgsException();
     }
     return curryN(arity, function() {
-        return _foldl(_ap, _map(lifted, arguments[0]), _slice(arguments, 1));
+        return _foldl(_ap, map(lifted, arguments[0]), _slice(arguments, 1));
```
buzzdecafe: don't you have to use dispatching `ap` here as well?

CrossEye: Internal `_ap` is dynamically dispatched.  The only difference from the public one is that it's not curried.  At some point we should have a discussion about whether this style is worth it.  I don't see this proliferation of functions as being overly helpful.

buzzdecafe:
:cow2:

PR #784, opened 2015-02-09
remove dispatch from R.of

davidchambers:
@buzzdecafe and I discussed this change and decided it's a good idea for two reasons:

  - it's one of the few (or perhaps the only) Ramda function with an optional argument; and
  - it's unnecessary, since `R.of(42, Id)` can be written as `Id.of(42)`.

### DIFF
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -1125,6 +1125,27 @@
     };
 
     /**
+     * Returns a singleton array containing the value provided.
+     *
+     * Note this `of` is different from the ES6 `of`; See
+     * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/of
+     *
+     * @func
+     * @memberOf R
+     * @category Function
+     * @sig a -> [a]
+     * @param {*} x any value
+     * @return {Array} An array wrapping `x`.
+     * @example
+     *
+     *      R.of(null); //=> [null]
+     *      R.of([42]); //=> [[42]]
+     */
+    var of = function of(x) {
+        return [x];
+    };
+
+    /**
      * Accepts a function `fn` and returns a function that guards invocation of `fn` such that
      * `fn` can only ever be called once, no matter how many times the returned function is
      * invoked.
@@ -3459,29 +3480,6 @@
      *      R.nth(-99, list); //=> undefined
      */
     var nth = _curry2(_nth);
-
-    /**
-     *
-     * `of` wraps any object in an Array. This implementation is compatible with the
-     * Fantasy-land Applicative spec, and will work with types that implement that spec.
-     * Note this `of` is different from the ES6 `of`; See
-     * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/of
-     *
-     * @func
-     * @memberOf R
-     * @category Function
-     * @sig a -> [a]
-     * @param {*} x any value
-     * @return {Array} An array wrapping `x`.
-     * @example
-     *
-     *      R.of(1); //=> [1]
-     *      R.of([2]); //=> [[2]]
-     *      R.of({}); //=> [{}]
-     */
-    var of = function of(x, container) {
-        return _hasMethod('of', container) ? container.of(x) : [x];
-    };
 
     /**
      * Returns a partial copy of an object omitting the keys specified.
--- a/src/of.js
+++ b/src/of.js
@@ -1,10 +1,6 @@
-var _hasMethod = require('./internal/_hasMethod');
-
-
 /**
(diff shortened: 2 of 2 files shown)


davidchambers:
Any thoughts, @CrossEye?

CrossEye:
Thanks for the ping.  Too much activity lately.  Can't keep up.

Something feels wrong with this.  If I remember correctly, the main reason for creating this function was the dynamic dispatch.  

The second point seems silly as it's true for all the dynamically dispatched methods:

    map(double, Maybe(42)) ~= Maybe(42).map(double)

but that's not always how we want to write code.  

The first point is more serious.  It tells me there is something wrong with the existing `.of`.  But I'm wondering if there's something wrong with the single array implementation more than with the dynamic dispatch.

Also, we should note that the other way the current code could be used, assuming we have followed the FantasyLand laws correctly, is:

    var i = Id.of(42), m = Maybe.of(42), f = Future.of(42);
    var i2 = R.of(99, i), m2 = R.of(99, m), f2 = R.of(99. f);  // line 2

and we have new Applicatives of the correct types, without necessarily knowing (in line 2) the types of the inputs.  Losing this would bother me, I think.

But for one factor, my preference would actually be to make the second parameter mandatory for the wrap-in-array case too, and force you to pass `[]` to get it to work.  Then we could comfortably use `_checkForMethod` just as it is, and everything would be grand.  Except, then the parameters feel like they're in the wrong order for currying.  And we never want that in Ramda.  So I'm still looking for another alternative.

Anyone got one?

davidchambers:
Note that Ramda's current method of dispatching only works in two out of four cases:

```javascript
function A() {}
A.of = function(x) { ... };

function B() {}
B.prototype.of = function(x) { ... };

// Assume a in an A and b is a B...

var a2 = R.of(99, A);  // Works, since A.of is defined.
var a3 = R.of(99, a);  // Fails, since a.of is not defined.

var b2 = R.of(99, B);  // Fails, since B.of is not defined.
var b3 = R.of(99, b);  // Works, since b.of is defined.
```

It's thus not clear whether the optional argument to `R.of` should be a type or a value. In fact, it depends on the implementation of the type. If we require users to know whether a type provides `of` on the constructor or on the prototype, the user has all the information she needs to eschew `R.of`, as demonstrated below.

> ```
> var i2 = R.of(99, i), m2 = R.of(99, m), f2 = R.of(99. f);
> ```

Depending on the implementation of each of these types, this could be written as:

```javascript
var i2 = i.constructor.of(99);
var m2 = m.constructor.of(99);
var f2 = f.constructor.of(99);
```

or:

```javascript
var i2 = i.of(99);
var m2 = m.of(99);
var f2 = f.of(99);
```

CrossEye:
Ah yes, I thought an `or` in the spec was an `and`.  This is very ugly:

> A value which has an Applicative must provide an of method on itself **or** its constructor object.

Maybe you two are right.  But it would be a real bummer, because I was hoping that our dynamic dispatch of the FL methods would be a way that Ramda would be able to work easily with any FL implementation.

The only other possibility I see would be to use a separate (very ugly and untested) custom dispatching mechanism:

    var of = _curry2(function of(type, val) {
        if (!_isArray(type) || type !== Array && 
                (typeof type.of == "function" || 
                  type.constructor && typeof type.constructor.of == "function")) {
            return (typeof type.of == "function") ? type.of(val) : type.constructor.of(val);
        }
        return [val];
    });

with some sort of named gloss for `of(Array)` if desired.  But this parameter order also feel squirrelly.  

(Scott wanders off mumbling something about `op` and `partialRight`...)

davidchambers:
> I was hoping that our dynamic dispatch of the FL methods would be a way that Ramda would be able to work easily with any FL implementation.

That'll still be the case, just not for `of`. `of` is different from `map` and friends in that it doesn't operate on `this`, so it makes more sense as a "static method" (in Java parlance) than as an "instance method". Therefore I consider it reasonable to require users to write `A.of(x)` or `a.constructor.of(x)` rather than use `R.of`.

CrossEye:
A reluctant :+1: from me.

I will probably challenge this over on FL after doing a bit more thinking.  This "or" really bothers me.

It really means there is no way to do any useful abstraction with Applicatives.  I mean what if I have an `of` on my Applicative prototype that does something different from the `of` on its constructor function?  Which one should I assume does the proper Applicative `of`?  Even worse, in these days of `Object.create`, what is FL even doing assuming that I have a constructor function to define my type?

davidchambers:
Okay. Let's merge this and see what comes of the fantasy-land thread.

buzzdecafe:
see https://github.com/fantasyland/fantasy-land/issues/88

PR #785, opened 2015-02-09
prevent cache misses in R.memoize caused by use of || operator

davidchambers:
Supersedes #776

### DIFF
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -920,55 +920,6 @@
     };
 
     /**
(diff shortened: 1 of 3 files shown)


CrossEye:
LGTM

CrossEye:
> It's the result of the function application that's significant, not the value of any given argument.

Correct.  Perhaps the test case could have somewhat more descriptive text?

davidchambers:
> Perhaps the test case could have somewhat more descriptive text?

Good idea.

PR #786, opened 2015-02-09
Add internal function _depricationWarning.

TheLudd:
The _depricationWarning function provides an easy way to show a
deprication message around a function that has been renamed. By default
it prints the name of the old function and what it has been renamed to.
It also provides the possibility to add an additional message.

_depricationWarning makes sure that it only prints the error message
once per depricated function.

### DIFF
--- a/src/mixin.js
+++ b/src/mixin.js
@@ -1,0 +1,9 @@
+var _deprecationWarning = require('./internal/_deprecationWarning');
+var merge = require('./merge');
+
+module.exports = _deprecationWarning({
+    oldName: 'mixin',
+    newName: 'merge',
+    fn: merge,
+    optionalMessage: 'R.mixin will be removed in v1.0.0'
+});
--- a/deprecation-demo.js
+++ b/deprecation-demo.js
@@ -1,0 +1,7 @@
+var mixin = require('./src/mixin');
+
+var a = {foo: 'bar', baz: 'qux'};
+var b = {foo: 'baz', qux: 'out'};
+
+console.log(mixin(a, b));
+console.log(mixin(a, b));
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -6103,6 +6103,42 @@
     // passing `identity` gives correct arity
     var project = useWith(_map, pickAll, identity);
 
+    var _deprecationWarning = function () {
+        var alreadyLogged = [];
+        function _addLogged(name) {
+            alreadyLogged = append(name, alreadyLogged);
+        }
+        function _isLogged(name) {
+            return contains(name, alreadyLogged);
+        }
+        function logMessage(opts) {
+            var message;
+            if (console) {
+                message = join('', [
+                    'R.',
+                    opts.oldName,
+                    ' has been deprecated. Please use R.',
+                    opts.newName,
+                    ' instead.'
+                ]);
+                if (opts.optionalMessage) {
+                    message = join(' ', [
+                        message,
+                        opts.optionalMessage
+                    ]);
+                }
+                console.warn(message);
+                _addLogged(opts.oldName);
+            }
+        }
+        return function _deprecationWarning(opts) {
+            if (!_isLogged(opts.oldName)) {
+                logMessage(opts);
+            }
+            return opts.fn;
+        };
+    }();
+
     /**
      * Wraps a constructor function inside a curried function that can be called with the same
      * arguments and returns the same type.
@@ -6130,6 +6166,13 @@
     var construct = function construct(Fn) {
         return constructN(Fn.length, Fn);
     };
+
+    var mixin = _deprecationWarning({
+        oldName: 'mixin',
+        newName: 'merge',
+        fn: merge,
+        optionalMessage: 'R.mixin will be removed in v1.0.0'
+    });
 
     var R = {
         F: F,
@@ -6249,6 +6292,7 @@
         merge: merge,
         min: min,
         minBy: minBy,
+        mixin: mixin,
         modulo: modulo,
         multiply: multiply,
         nAry: nAry,
--- a/src/internal/_deprecationWarning.js
+++ b/src/internal/_deprecationWarning.js
@@ -1,0 +1,35 @@
+var append = require('../append');
+var contains = require('../contains');
+var join = require('../join');
+
+module.exports = (function() {
+
+    var alreadyLogged = [];
+
+    function _addLogged(name) {
+        alreadyLogged = append(name, alreadyLogged);
+    }
+
+    function _isLogged(name) {
+        return contains(name, alreadyLogged);
+    }
+
+    function logMessage(opts) {
+        var message;
+        if (console) {
+            message = join('', ['R.', opts.oldName, ' has been deprecated. Please use R.', opts.newName, ' instead.']);
+            if (opts.optionalMessage) {
+                message = join(' ', [message, opts.optionalMessage]);
+            }
+            console.warn(message);
+            _addLogged(opts.oldName);
+        }
+    }
+
+    return function _deprecationWarning(opts) {
+        if (!_isLogged(opts.oldName)) {
+            logMessage(opts);
+        }
+        return opts.fn;
+    };
+}());


TheLudd:
I noticed that #783 was merged and ```mixin``` is used a lot in my projects. I would really like if it was not removed so suddenly but rather allowed for some adjustment time, hence this PR.

This PR has the _depricationWarning applied to ```mixin```. There is also a demo file in the root which you can run to see the feature in action. If this PR is approved, I will remove the demo file before merge.

CrossEye:
Ahh, the joys of running from HEAD, especially in a pre-1.0 library!

Sorry about that.  We are trying to rationalize all names before we hit 1.0... Obviously we caused some real pain with `reduce`.  This one I thought would be less of an issue.

I'll have to give some thought as to the general idea here.  You're right that we do need a better way to make any such transitions easier.  This might well be it, but I need to think it through a bit.  I'd love to hear discussion about other possible approaches.

Several points:

  * You've spelled "deprecation" wrong.  It's an "e" in the middle.
  * You might want to read [CONTRIBUTING.md][co] which asks that you run `make` or the `grunt` equivalent before checking in in order to regenerate `dist/ramda.js`.  This is why the Travis build is failing.
  * I'm not thrilled about all the local variables being used here.  I know that it's common practice in Node modules, but unless I'm mistaken about our build, they'll all end up mixed into our Ramda closure.  It feels dirty.  I'd think I'd rather they were in a local closure for the function.


  [co]: https://github.com/ramda/ramda/blob/master/CONTRIBUTING.md

TheLudd:
I fixed the spelling and created a closure for the local variables. But I still get this error when runnning make:
```
assert.js:87
  throw new assert.AssertionError({
        ^
AssertionError: '_deprecationWarning' === '_deprecationWarningJs'
```
I don't really understand what it is about..

src/mixin.js
```diff
@@ -0,0 +1,9 @@
+var _deprecationWarning = require('./internal/_deprecationWarning.js');
```
CrossEye: You probably don't need the ".js" at the end of the require.

davidchambers:
I've lots to say about this. :)

This breaks [referential transparency][1]. This may be okay in practice, but if we're to break one of the tenets of functional programming we should at least do so knowingly.

In some contexts printing a message and having the program exit successfully is *worse* than having the program crash. Consider, for example, a command-line utility whose only output is via stdout and/or stderr. The Ramda error messages will be intermingled with the program's stderr output, which could have any number of undesirable effects.

> I would really like if [`mixin`] was not removed so suddenly but rather allowed for some adjustment time

There's a one-time cost associated with renaming functions. Even in a dynamically typed language, it isn't difficult in most cases. When I update a project to 0.10.0 I will run:

```console
$ find lib -name '*.js' -print0 | xargs -0 sed -i '' 's/R[.]mixin/R.merge/g'
```

> I still get this error when running make:
>
> ```
> assert.js:87
>   throw new assert.AssertionError({
>         ^
> AssertionError: '_deprecationWarning' === '_deprecationWarningJs'
> ```
>
> I don't really understand what it is about.

The build script asserts that each module required is assigned to a variable whose name exactly matches the last path component of the require path. This is necessary to prevent things from breaking when the various source files are transformed into a single file. The imports are removed in this transformation, so we must name modules consistently.

This change gets a :thumbsdown: from me. In a project with full test coverage updating Ramda is almost painless. Projects with little or no test coverage probably shouldn't use Ramda until its API stabilizes (as will be signified by the release of 1.0.0).


[1]: http://en.wikipedia.org/wiki/Referential_transparency_%28computer_science%29

CrossEye:
@davidchambers 

> I've lots to say about this. :)

(That would probably constitute a short-medium response from me.  :smile:)

Something was bothering me about this, and your point that this would break referential transparency was exactly it.

But @TheLudd is right that we aren't doing enough.  I do wish we had a mailing list at least.

TheLudd:
Ok, here is my $0.02.
While I appreciate the prureishness of ramda with immutable data and single purpose functions I don't come from any "pure" background and have little knowledge in comparison about these laws and so on. However I cannot see the need to fulfill all the laws with functions that are deprecated. 

My experience comes from using a library like Facebooks React which logs messages like these very well. When I upgrade to a new version and see what changes have been made printed nicely I *feel* joy, trust, happiness towards creators and the care by which they create their product. They don't do it because their version is above 1.0.0 and they have to, but rather because they know that their users will appreciate it and that it will make the use of the library easier. And yes, with ramda it is probably easier to do the upgrade, and yes with a good test suite you should have an easy time fixing the errors. But things like this just makes everything smoother. 

Ramda is all over the product we are building. We are still at ```0.8.0``` because of changes like removing ```reduce```. Moreover (tell me if I am wrong), but you have removed ```foldl``` and put back ```reduce``` again, so if I had gone through with that change, I would have to change back again? It is good that you do all this before 1.0.0 but another way of viewing it is that a lot of these changes are being made now, and what the final result will be is hard to say, therefore I urge you to give the users of ramda an option to adapt gradually.

src/mixin.js
TheLudd: ah, pure mistake..

CrossEye:
I'm still floundering on this one.  I want some better mechanism to note API changes than just a changelog-based system.  It's becoming more of an issue as [NPM downloads][dn] are growing quickly (6K+ in December [if you ignore one suspect day], 10K in January, and on track for 15K in February.)

I can't quite decide if this is it.  @TheLudd's makes a good point that referential integrity matters less for functions we've said are deprecated.  But if they still exist and are still Ramda functions, then it would be at least somewhat problematic.  And of course the `_deprecationWarning` function itself is a morass of referentially opaque code.  Even if they logged every time they ran and avoided the local state (users would probably fix them quickly then!) these would still have side effects.

On the other hand, it would be great if it were not some external document that gave the user the information about the updates, but simply using Ramda that told them.  That's what I really like about this.

If I was a fan of heavyweight processes, a nice solution presents itself in using something like this in beta versions, and then stripping it out in final versions.  But there are many more fundamental things I want to do to our build system and our process that I haven't gotten to, so I can't really contemplate this.

The point from @davidchambers about how easy it is to fix is not so relevant to me.  It's more a matter of the surprise of needing to.  And maybe it's because in my job I never work with full test coverage -- I feel very happy with something approaching 75% -- but I don't think that full test coverage on my project is a prerequisite for using a pre-1.0 release.

As I said, I'm very mixed up about this.

@buzzdecafe, I know you're home sick.  If you're online, have you had any thoughts on this?


  [dn]: http://npm-stat.com/charts.html?package=ramda#months

buzzdecafe:
sorry gents, been in a bit of a fog. slowly emerging now ....

anyways, i see the value in this, but I don't like it in the source code. I'd rather see it added as a message in `npm` possibly `postInstall`.

TheLudd:
@buzzdecafe How would that even be possible? Maybe I'm not sure what you mean..

buzzdecafe:
see https://docs.npmjs.com/misc/scripts

I'm suggesting we could add a generic script that we point at from package.json `scripts.postInstall` property. That script could look in a directory (possibly under dist? or lib? or etc?) for a json file describing changes in this version. Then it could log those changes to console.

TheLudd:
Sounds too much of a hassle to me. We could just have a change log online then IMO.

davidchambers:
None of the core committers agrees with the approach taken in this pull request. Let's continue to discuss other approaches.

Aourin:
https://github.com/Aourin/ramda-alias

It isn't elegant or super complete yet. Just getting started, but some collab would be welcome until the full version is released. If anyone wants to help with a temp alias repo to decorate Ramda, feel free to jump in.

buzzdecafe:
@Aourin I like the idea of user-defined aliases. i'd be interested in seeing the build extended to optionally accept a list of aliases they want to include.

Aourin:
I can see that being useful for the Ramda build itself, but I am concerned with users individually needing to flavor Ramda everytime it changes. The ramda-alias would be a temp solution to the base version changes as a stop gap while the naming conventions are getting resolved. 

If you mean a build solution for the alias repo, then that is a good idea. I wasn't sure based off of the placement of these comments now. If so, feel free to jump over to that repo and comment there so we don't clog up this one.

CrossEye:
@Aourin: This looks like a great stopgap.  Hopefully we won't need it for long!

Thanks!

CrossEye:
https://twitter.com/slicknet/status/609100099872882688

TheLudd:
@CrossEye I still don't agree but I have left this discussion behind me when it comes to ramda =)

CrossEye:
Yeah, just saw the tweet and thought of this issue.  Didn't want to reopen debate so much as record the tweet in this context.

PR #787, opened 2015-02-09
fix behaviour of memoized functions when given no arguments

### LINKED ISSUE #791: memoize broken for no args
Moving from 0.8.0 to 0.9.1, my memoized functions broke (not sure if new expected behavior, not specified in release notes I think):

expensiveThingThatGetsCreatedOnce = R.memoize(function()
{
  //expensive initialization process;
  return it;
});

expensiveThingThatGetsCreatedOnce() // undefined

davidchambers:
Fixes #791

We *could* add extra logic to handle caching of nullary functions, but this doesn't seem useful enough to justify the increase in code complexity.

### DIFF
--- a/src/memoize.js
+++ b/src/memoize.js
@@ -1,6 +1,5 @@
-var _reduce = require('./internal/_reduce');
+var _map = require('./internal/_map');
 var has = require('./has');
-var last = require('./last');
 
 
 /**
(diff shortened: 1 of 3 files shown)


src/memoize.js
```diff
@@ -41,7 +41,9 @@ var last = require('./last');
 module.exports = function memoize(fn) {
     var cache = {};
     return function() {
-        if (!arguments.length) {return;}
+        if (arguments.length === 0) {
+            return fn.apply(this, arguments);
```
davidchambers: I realize this could be written as `fn.call(this)` but there's value in using the same approach as in the other branch. This makes it clear that we're calculating the return value the same way in both cases; the only difference is that we don't involve the cache if given no arguments.

CrossEye:
There seems to be a tension between the need expressed in #613 to avoid depending on the function arity for memoization, and the need @tolmasky is expressing in #791 to memoize the expensive call to a nullary function.  An older version of this handled that by having memoize simply return `once(fn)` for nullary functions.  But that won't do if we want to be able to memoize on the actual arguments, regardless of arity.  I think we might need that extra logic...

davidchambers:
> I think we might need that extra logic...

I have something in mind. Let's see what I can do.

davidchambers:
I've updated the pull request. The new implementation is correct more often in addition to being significantly simpler.

CrossEye:
Almost LGTM.  This is very misleading, and took a little time to figure it out:

    assert.strictEqual(f(), 'foobar');
    assert.strictEqual(f(), 'foobar');
    assert.strictEqual(f(), 'foobar');
    assert.strictEqual(count, 3);

Should probably only call `f()` once in there.

I'd seen CANON before.  This is a nice use, even of such a restricted subset of it.

davidchambers:
Good catch, @CrossEye. I added `count` to that example after the fact, forgetting that the default argument values were filled in via recursive calls. I've removed the recursive calls so the final value of `count` is now 1.

CrossEye:
:herb:

davidchambers:
Oops. I forgot to push the update. I'll open a new pull request.

PR #788, opened 2015-02-09
Add defaultValue

TheLudd:
A function that takes two arguments and returns the first argument if
the second one is null/undefined. Otherwize the second one is returned.

Basically a functional version of

    var list = someVariable || [];

### DIFF
--- a/test/defaultTo.js
+++ b/test/defaultTo.js
@@ -1,0 +1,30 @@
+var assert = require('assert');
+
+var R = require('..');
+
+describe('defaultTo', function() {
+
+    var defaultTo42 = R.defaultTo(42);
+
+    it('returns the default value if input is null/undefined', function() {
+      assert.strictEqual(42, defaultTo42(null));
+      assert.strictEqual(42, defaultTo42(undefined));
+    });
+
+    it('returns the input value if it is not null/undefined', function() {
+      assert.strictEqual('a real value', defaultTo42('a real value'));
+    });
+
+    it('returns the input value even if it is considered falsy', function() {
+      assert.strictEqual('', defaultTo42(''));
+      assert.strictEqual(0, defaultTo42(0));
+      assert.strictEqual(false, defaultTo42(false));
+      assert.deepEqual([], defaultTo42([]));
+    });
+
+    it('can be called with both arguments directly', function() {
+      assert.strictEqual(42, R.defaultTo(42, null));
+      assert.strictEqual('a real value', R.defaultTo(42, 'a real value'));
+    });
+
+});
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -1,3 +1,8 @@
+//  Ramda v0.9.1
+//  https://github.com/ramda/ramda
+//  (c) 2013-2015 Scott Sauyet and Michael Hurley
+//  Ramda may be freely distributed under the MIT license.
+
 ;(function() {
 
     'use strict';
@@ -2059,6 +2064,29 @@
      *      R.dec(42); //=> 41
      */
     var dec = add(-1);
+
+    /**
+     * Returns the second argument if it is not null or undefined. If it is null
+     * or undefined, the first (default) argument is returned.
+     *
+     * @func
+     * @memberOf R
+     * @category Logic
+     * @sig a -> b -> a | b
+     * @param {a} val The default value.
+     * @param {b} val The value to return if it is not null or undefined
+     * @return {*} The the second value or the default value
+     * @example
+     *
+     *      var defaultTo42 = defaultTo(42);
+     *
+     *      defaultTo42(null);  //=> 42
+     *      defaultTo42(undefined);  //=> 42
+     *      defaultTo42('Ramda');  //=> 'Ramda'
+     */
+    var defaultTo = _curry2(function _defaultTo(d, v) {
+        return v == null ? d : v;
+    });
 
     /**
      * Finds the set (i.e. no duplicates) of all elements in the first list not contained in the second list.
@@ -6163,6 +6191,10 @@
     };
 
     var R = {
+        F: F,
+        I: I,
+        T: T,
+        __: __,
         add: add,
         all: all,
         allPass: allPass,
@@ -6201,6 +6233,7 @@
         curry: curry,
         curryN: curryN,
         dec: dec,
+        defaultTo: defaultTo,
         difference: difference,
         differenceWith: differenceWith,
         dissoc: dissoc,
@@ -6212,7 +6245,6 @@
         eqDeep: eqDeep,
         eqProps: eqProps,
         evolve: evolve,
-        F: F,
         filter: filter,
         filterIndexed: filterIndexed,
         find: find,
@@ -6234,7 +6266,6 @@
         has: has,
         hasIn: hasIn,
         head: head,
-        I: I,
         identity: identity,
         ifElse: ifElse,
         inc: inc,
@@ -6335,7 +6366,6 @@
         substringTo: substringTo,
         subtract: subtract,
         sum: sum,
-        T: T,
         tail: tail,
         take: take,
         takeWhile: takeWhile,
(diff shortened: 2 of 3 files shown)


TheLudd:
Every now and then I find myself wanting a function like this. If it is wanted, I will add documentation.

megawac:
I'm not sure conventional `curry`ing of this is a good idea

For instance, should this really through an error, I'd expect it to be more or less the same as constant

```js
var def = R.defaultVal('foo');
def();
```

davidchambers:
This problem is well solved by maybes in [Sanctuary][1]:

```javascript
S.or(S.Nothing(), S.Nothing()); // => S.Nothing()
S.or(S.Nothing(), S.Just(42));  // => S.Just(42)
S.or(S.Just(42), S.Nothing());  // => S.Just(42)
S.or(S.Just(42), S.Just(99));   // => S.Just(42)
```

Maybe is a better abstraction here than using null.


[1]: https://github.com/plaid/sanctuary

CrossEye:
@megawac: Why assume that it should act like a constant?  I've never actually been happy with our throws-on-no-arguments rule, but I don't see it any worse here than elsewhere.

@davidchambers: While I agree that `Maybe` is a great abstraction, as is `Either`, `Future`, `IO`, and many others, users of JS are quite used to getting and working with `null`/`undefined` values.  This just offers a functional way to deal with something that otherwise would be an imperative snippet.

TheLudd:
Is the example from @megawac supposed to throw? Because it doesn't when try it...

CrossEye:
@TheLudd: when we curry a function and then you call the resulting function with no arguments, it should throw a `TypeError`, from the internal `_noArgException`.

It happens in http://tinyurl.com/oon8awt.  Perhaps there's something odd happening in `curry2`.

CrossEye:
@TheLudd: when we curry a function and then you call the resulting function with no arguments, it should throw a `TypeError`, from the internal `_noArgException`.

It happens in http://tinyurl.com/oon8awt.  Perhaps there's something odd happening in `curry2`.

TheLudd:
Just noticed that there is an existing special case version of this already in ramda.
http://ramdajs.com/docs/#propOr

CrossEye:
Yes. when I first saw yours, I thought it was a duplicate, until I read a little further.  But I think both are warranted.

buzzdecafe:
looks good -- why not just `default`?

TheLudd:
@buzzdecafe I choose ```defaultValue``` since ```default``` is a keyword in Javascript. But perhaps that doesn't matter.

buzzdecafe:
@TheLudd of course, my mistake

davidchambers:
I'm not a fan of this function but if we are to include it, I think it should be named `R.defaultTo` or `R.withDefault`.

buzzdecafe:
consistent with `propOr` maybe? e.g. `valueOr`?

davidchambers:
The Ramda of my dreams would not include `R.propOr` either. ;)

```javascript
propOr('default', 'name', user)
// can be expressed as two simpler operations...
fromMaybe('default', get('name', user))

propOr('default', '0', xs)
// can be expressed as two simpler operations...
fromMaybe('default', head(xs))
```

Using Maybes would reduce the surface area of the API. Without Maybe, one must provide a specialized version of many functions to handle the failure case:

```
propOr :: a -> String -> Object -> a
headOr :: a -> [a] -> a
findOr :: a -> (a -> Boolean) -> [a] -> a
...
```

If `prop`, `head`, `find`, etc. were to return Maybes, `fromMaybe` would obviate the need for all the above functions.

TheLudd:
@davidchambers Why not just compose? ```R.compose(R.defaultTo(42), R.find(...))```
So I'd say this function really makes ```R.propOr = R.useWith(R.defaultTo, R.prop)```

I agree that ```Maybe``` is a better abstraction but it should really be all or nothing IMHO and as I understand it, ramda has made a choice not to include ```Maybe``` et al.

Anyway. What will it be?
 * defaultTo? <--- my vote
 * withDefault?

davidchambers:
> Why not just compose? `R.compose(R.defaultTo(42), R.find(...))`

This approach does not handle `null` correctly. For example:

```javascript
var safeHead = R.compose(R.defaultTo(42), R.head);

safeHead([]);  // => 42 (correct)
safeHead([99]);  // => 99 (correct)
safeHead([null]);  // => 42 (incorrect: should be null)
```

buzzdecafe:
i agree that should we include `valueOr` or whatever we name it, that would make `propOr` somewhat redundant: `propOr = compose(valueOr('default'), get('prop'))`. I'd rather remove that one in that case.

and i agree that maybes are swell, but i also agree with @CrossEye in keeping the haskell-y stuff separate. Once you start down that road, it is all or nothing, I think, and would represent a huge shift. I don't think such a shift is warranted when we can bolt that stuff on after the fact (sanctuary, ramda-fantasy). I'm content providing a gateway drug to PureScript.

CrossEye:
@davidchambers 

How do  you think we should also be solving this problem (throughout the library)?:

```javascript
var arr  = [];
arr[4] = "x";
safeHead(arr); //=> 42 (should this be `undefined`?)
```

So far we've insisted that the arrays users pass to Ramda are lists of values.  If we want to be able to process arbitrary arrays, including sparse ones, and `null`s and `undefined`s as well as values, we're going to have to think a fair bit about what that means.

My vote, BTW is for `defaultTo`.

TheLudd:
@davidchambers I don't agree that the third case is incorrect. The definition of the function was:

> A function that takes two arguments and returns the first argument if
the second one is null/undefined. Otherwize the second one is returned.

So it doesn't matter where the second argument comes from. If it is an undefined reference or an actual null/undefined value. If it is null/undefined the default value should be returned.

davidchambers:
@CrossEye, the head of `[,,,,'x']` should be `Just(undefined)`. Only the head of `[]` is `Nothing`.

@TheLudd, it depends on whether you consider the head of `[null]` to be equivalent to the head of `[]`. I don't.

CrossEye:
@davidchambers: Only if we're going to use the library to reshape much more of the way that people use JS than we currently do.  That's the problem.  Such things are viral.  And they pull users much further away from idiomatic Javascript.

TheLudd:
@davidchambers But the point of this function is that it has no knowledge about that.  And so `safeHead` could be defined as:
```
safeHead = R.ifElse(R.isEmpty, R.always(42), R.head)
```

TheLudd:
I have changed the name to `defaultTo` and added documentation.

test/defaultTo.js
```diff
@@ -0,0 +1,30 @@
...
+      assert.strictEqual(0, defaultTo42(0));
+      assert.strictEqual(false, defaultTo42(false));
+      assert.deepEqual([], defaultTo42([]));
+    });
+
+    it('can be called with both arguments directly', function() {
+      assert.strictEqual(42, R.defaultTo(42, null));
+      assert.strictEqual('a real value', R.defaultTo(42, 'a real value'));
+    });
+
```
buzzdecafe: 'it is curried' test?

TheLudd: Yes. I used curried by default and added a non curried test. Mostly because I preferred to use the curried version in all assertions instead of passing in the default value everywhere. 

Should I change it? Or just the description somewhere?

buzzdecafe:
:cow2:

TheLudd:
I am not fluent in animal emoticons :smile:

buzzdecafe:
translated: LGTM

CrossEye:
:herb:

(Translates from flora emoticonese as LGTM)

buzzdecafe:
now should we remove/redefine `propOr` as a composition of `defaultTo, get`?

davidchambers:
> now should we remove/redefine `propOr` as a composition of `defaultTo, get`?

That would be a breaking change: `propOr` returns the default value only if the specified property does not exist. Let's leave `propOr` as is. Users can then choose `R.propOr('default', 'name', user)` or `R.defaultTo('default', user.name)` based on their requirements.

buzzdecafe:
fair enough.

PR #789, opened 2015-02-10
preserve object identity in R.sortBy

davidchambers:
/cc @charleswli


CrossEye:
:+1:

Good catch.  I believe this was written when `clone` was in fact a shallow copy.  :smile:

davidchambers:
> I believe this was written when `clone` was in fact a shallow copy.

I believe you're right.

PR #790, opened 2015-02-10
update misleading tests for R.invert and R.invertObj

davidchambers:
As discovered in #780, `R.invert(1)` and friends all evaluate to `{}`. Inexplicably `assert.throws({})` is not a type error, so these tests test nothing. Were these tests correctly written they would fail as these expressions do *not* throw when evaluated. A comedy of errors. :)

### DIFF
--- a/test/invert.js
+++ b/test/invert.js
@@ -6,14 +6,20 @@ describe('invert', function() {
     it('takes a list or object and returns an object of lists', function() {
         assert.equal(typeof R.invert([]), 'object');
         assert.equal(typeof R.invert({}), 'object');
-        assert.throws(R.invert(1));
-        assert.throws(R.invert('1'));
-        assert.throws(R.invert(null));
-        assert.throws(R.invert(undefined));
 
         var inverted = R.invert([0]);
         var keys = R.keys(inverted);
         assert(R.is(Array, inverted[keys.pop()]));
+    });
+
+    it('returns an empty object when applied to a primitive', function() {
+        assert.deepEqual(R.invert(42), {});
+        assert.deepEqual(R.invert('abc'), {});
+    });
+
+    it('returns an empty object when applied to null/undefined', function() {
+        assert.deepEqual(R.invert(null), {});
+        assert.deepEqual(R.invert(undefined), {});
     });
 
     it('returns the input\'s values as keys, and keys as values of an array', function() {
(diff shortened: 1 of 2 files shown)


honzabrecka:
I'm against this PR, because I think that test isn't just for checking if code works as expected, but it's also a live documentation. Test should be removed only if the functionality that it tests is out. Once the edge case is discovered, then it should be properly tested - someone can rely on its current behaviour.

Therefore I would [change the assertions](https://github.com/honzabrecka/ramda/commit/694d58eca980adec1b0f66eb6cbb4cbf60c9738f) or change the current implementation - for those values it would throw an error as was intended.

CrossEye:
I agree with @honzabrecka about the test suite being live documentation.  I'm not sure I would make the argument a general principle, but I believe it applies here.

So yes, @davidchambers, there is a comedy of errors here.  But assuming we wrapped lambdas around these expressions and started failing the tests, we have learned something.  Don't those tests properly establish the API we would want from `invert[Obj]`?  Shouldn't we throw an error there?  Or do you think returning an empty object is the right behavior?

davidchambers:
> Shouldn't we throw an error there? Or do you think returning an empty object is the right behavior?

The result of inverting a value not comprised of key–value pairs is unspecified. We could throw a TypeError for null/undefined, and include tests to assert this, but I don't want to decide whether `new Number(42)` is an invertible object. I'd like to specify the behaviour for objects, arrays, and arguments objects, include tests for these, and let the implementation determine the behaviour of the unspecified cases.

CrossEye:
I can certainly live with leaving it the way it is.  But I think then that we should actually capture that with test.  Perhaps 

    it('returns an empty object when passed a primitive', function() {
         assert.equal(R.invert(42), {});
         assert.equal(R.invert('abc'), {});
    });;

    it('returns an empty object when passed nothing', function() {
         assert.equal(R.invert(null), {});
         assert.equal(R.invert(undefined), {});
    });;

I don't feel we have to exhaustively cover all types of primitives, but a smattering for demonstration purposes would be useful.

davidchambers:
I've updated this pull request to include the suggested tests.

CrossEye:
Sorry, I didn't say it, but we should probably do the equivalent thing on `invertObj`.

davidchambers:
> [W]e should probably do the equivalent thing on `invertObj`.

Yes, of course. There seem to be so many things to fix today; I'm not quite my usual methodical self. Updated.

CrossEye:
LGTM

PR #795, opened 2015-02-12
ignore inherited properties in R.invert

davidchambers:
I also made some minor formatting changes in __invert.js__ and __invertObj.js__.

### DIFF
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -1,3 +1,8 @@
+//  Ramda v0.9.1
+//  https://github.com/ramda/ramda
+//  (c) 2013-2015 Scott Sauyet and Michael Hurley
+//  Ramda may be freely distributed under the MIT license.
+
 ;(function() {
 
     'use strict';
@@ -5371,10 +5376,16 @@
      *
      */
     var invert = function invert(obj) {
-        var props = keys(obj), len = props.length, idx = -1, out = {};
+        var props = keys(obj);
+        var len = props.length;
+        var idx = -1;
+        var out = {};
         while (++idx < len) {
-            var key = props[idx], val = obj[key];
-            out[val] = out[val] || [];
+            var key = props[idx];
+            var val = obj[key];
+            if (!has(val, out)) {
+                out[val] = [];
+            }
             out[val].push(key);
         }
         return out;
@@ -5405,7 +5416,10 @@
      *      //=> { 'alice': '0', 'jake':'1' }
      */
     var invertObj = function invertObj(obj) {
-        var props = keys(obj), len = props.length, idx = -1, out = {};
+        var props = keys(obj);
+        var len = props.length;
+        var idx = -1;
+        var out = {};
         while (++idx < len) {
             var key = props[idx];
             out[obj[key]] = key;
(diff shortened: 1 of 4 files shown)


dist/ramda.js
```diff
@@ -1,3 +1,8 @@
+//  Ramda v0.9.1
+//  https://github.com/ramda/ramda
+//  (c) 2013-2015 Scott Sauyet and Michael Hurley
+//  Ramda may be freely distributed under the MIT license.
+
```
davidchambers: When did we lose our header?

davidchambers: d0033d73458ddf7ff7b742dc7289b9b555993e16

dist/ramda.js
```diff
@@ -5371,10 +5376,16 @@
...
+        var idx = -1;
+        var out = {};
         while (++idx < len) {
-            var key = props[idx], val = obj[key];
-            out[val] = out[val] || [];
+            var key = props[idx];
+            var val = obj[key];
+            if (!has(val, out)) {
+                out[val] = [];
+            }
```
buzzdecafe: is this better than `||`?

dist/ramda.js
CrossEye: Sorry, have been bed-bound for days without good internet.  (Going back to doctor in an hour.)  I finally realized that I could type with a real keyboard by using my phone as a hotspot.  (should have thought of that ages ago.)  But could only connect for some reason from old Windows laptop.  We don't seem to have a full build cycle working on Windows.  Either that or I messed it up.

dist/ramda.js
davidchambers: What are you suggesting, exactly?

dist/ramda.js
davidchambers: No worries. Order will be restored if and when this pull request is merged. :)

I hope you're well again soon.

dist/ramda.js
buzzdecafe: why rewrite 1-liner `out[val] = out[val] || []` to 3-liner:

```
if (!has(val, out)) {
    out[val] = [];
}
```

davidchambers: Because the one-liner is broken. ;)

The current implementation fails the test case.

Aside: I like to use `&&` and `||` with Boolean operands only. Similarly I ensure each of my predicates returns a Boolean and each of my conditions evaluates to a Boolean. I thus avoid the world of JavaScript truthiness.

davidchambers:
Can I get an LGTM or emoji equivalent? :)

CrossEye:
:herb: 

(But just for the record, you're getting this from a steroid-addled brain.  Doc has me on medicines which are fixing my complaint but really messing with my sleep!)

buzzdecafe:
@CrossEye good luck in this year's Tour de France

:cow2:

TheLudd:
@davidchambers 
> Can I get an LGTM or emoji equivalent? :)
haha love this

CrossEye:
@buzzdecafe: I'm looking forward to a Snore de France, I hope.

PR #796, opened 2015-02-12
update misleading description of R.replace

### LINKED ISSUE #727: replace(str, ..., ...) only replacing first occurence
docs: "Returns: string with all the matches replaced. "

```
replace( 'car', '', 'car boat car boat car' )
# expected: " boat  boat "
# actual: " boat car boat car"
```

it seems as if only the first match is replaced

davidchambers:
Closes #727


buzzdecafe:
much beter

PR #799, opened 2015-02-13
dispatch to "filter" in R.reject

davidchambers:
This is nice sugar. Assume I have a value `m` of type `Maybe String`, and I wish to perform the following transformation:

```
Nothing -> Nothing
Just '' -> Nothing
Just a  -> Just a
```

With Ramda I could write:

```javascript
R.filter(R.not(R.isEmpty), m)
```

What I'd really like to write, though, is:

```javascript
R.reject(R.isEmpty, m)
```

This change will enable me (and everyone else) to do so. :)

### DIFF
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -1,3 +1,8 @@
+//  Ramda v0.9.1
+//  https://github.com/ramda/ramda
+//  (c) 2013-2015 Scott Sauyet and Michael Hurley
+//  Ramda may be freely distributed under the MIT license.
+
 ;(function() {
 
     'use strict';
@@ -3968,7 +3973,7 @@
      *      R.reject(isOdd, [1, 2, 3, 4]); //=> [2, 4]
      */
     var reject = _curry2(function reject(fn, list) {
-        return _filter(not(fn), list);
+        return filter(not(fn), list);
     });
 
     /**
@@ -6163,6 +6168,10 @@
     };
 
     var R = {
+        F: F,
+        I: I,
+        T: T,
+        __: __,
         add: add,
         all: all,
         allPass: allPass,
@@ -6212,7 +6221,6 @@
         eqDeep: eqDeep,
         eqProps: eqProps,
         evolve: evolve,
-        F: F,
         filter: filter,
         filterIndexed: filterIndexed,
         find: find,
@@ -6234,7 +6242,6 @@
         has: has,
         hasIn: hasIn,
         head: head,
-        I: I,
         identity: identity,
         ifElse: ifElse,
         inc: inc,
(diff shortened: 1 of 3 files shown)


CrossEye:
My somewhat-addled brain isn't seeing how that helps your case.  But it's a direction I've wanted to go for a long time.  I'd rather be dog-fooding more than we are, and possibly therefore reducing the number of internal functions.  I have no issue with using the curried version internally.

... oh wait, it's probably also dynamically dispatched,  Is that why it helps?  I think I need to go back to bed!

In any case, LGTM.

buzzdecafe:
was a little squeamish about dispatching to a different name, but this is just sugar, so ....

:cow2:

CrossEye:
@buzzdecafe is right to be squeamish here.  Although this might work for @davidchambers' case, I don't like enshrining that behavior in a test case.  We might later want to decide to dispatch to a users `reject` method instead and not treat this as sugar.  So, while I prefer to use `./filter` than to use `./internal/_filter`, I'm afraid of committing to an internal implementation which should really not be part of the public API.

My brain must be doing better.  I'm back to my usual crotchety old self.  :frowning:

buzzdecafe:
@CrossEye do you want to revert this commit?

davidchambers:
> Although this might work for @davidchambers' case, I don't like enshrining that behavior in a test case. We might later want to decide to dispatch to a users `reject` method instead and not treat this as sugar.

The only sensible definition of `reject` is as the inverse of `filter`. This equivalence should always hold:

```javascript
a.reject(p)
// is equivalent to...
a.filter(function(x) { return !p(x); })
```

Since there is only one definition, Ramda should provide it. Furthermore, let's not put ourselves in the position of dispatching to `reject` if provided, `filter` otherwise.

Please correct me if I'm wrong in thinking that `reject` is always the inverse of `filter`.

buzzdecafe:
> let's not put ourselves in the position of dispatching to `reject` if provided, `filter` otherwise.

haha! I think we can all agree on that. I'd be happy to ditch `reject`, it's just sugar, and not worth the extra effort.

TheLudd:
@buzzdecafe 

>haha! I think we can all agree on that. I'd be happy to ditch reject, it's just sugar, and not worth the extra effort.

WDYM? While `reject(f)` definitely should equal `filter(not(f))` I really think that it should be a part of the ramda API. Ramda contains several functions that could be derived from others but still deserve a right to exits. Of those functions `reject` is one of the first ones I would defend.

CrossEye:
Hey, blame the drugs!

I think the docs should explicitly note that this calls `filter`.  But I agree with @TheLudd that this is worth keeping.   Retraction retracted.

PR #804, opened 2015-02-14
Fixes #803: "make clean" no longer removes ramda.min.js

### LINKED ISSUE #803: "make clean; make" leaves you without dist/ramda.min.js
After `make clean; make`, `git status` reports that `dist/ramda.min.js` has been deleted.

I would expect `make` to build both `dist/ramda.js` and `dist/ramda.min.js`.

nicolasv:
`make clean; make` doesn't rebuild ramda.min.js so `git status` marks
it as deleted.

@davidchambers says ramda.min.js is only updated when a new version is
published, so let's not remove it on clean.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ dist/ramda.min.js: dist/ramda.js scripts/header
 
 .PHONY: clean
 clean:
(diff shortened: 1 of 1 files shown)


davidchambers:
:deciduous_tree: Very nice. Thanks, Nic!

Issue #806, opened 2015-02-14
Partial builds?


benjamingr:
Hey, Ramda is kind of big but I'm really enjoying it. Is there any supported/official way to create a partial build containing only the functions I'm using?

buzzdecafe:
Thanks! Yes, it is a little big, and yes there is a way to roll your own. 

`./scripts/build -- src/curry.js src/compose.js src/reduce.js src/map.js src/filter.js > dist/ramda.custom.js`

We should add some documentation for that.

PR #807, opened 2015-02-14
Document partial build functionality

benjamingr:
Adds a description of how to make a partial build of Ramda, closes https://github.com/ramda/ramda/issues/806

I will create a separate PR for the docs repo as it's in a different repo. Note I've also included the /docs fix here.


benjamingr:
Hey, in first look it looks like the index file in the docs repo is generated from the README.md file here - is this the case or should I submit a separate PR for the docs repo?

README.md
```diff
@@ -120,10 +120,18 @@ Or you can inject ramda into virtually any unsuspecting web site using this book
...
 ```
 
+### Partial Builds
+
+It is possible to build Ramda with a subset of the functionality to reduce its file size. Ramda's build system supports this with command line flags. For example if you're using `R.compose`, `R.reduce` and `R.filter` you can create a partial build with:
+
+    ./scripts/build -- src/curry.js src/compose.js src/reduce.js src/map.js src/filter.js > dist/ramda.custom.js
```
buzzdecafe: i'd say that the functions in the description should match the functions in the example. So the preferable way to go is probably to remove `src/curry.js` and `src/map.js` from the example

benjamingr: Roger. Editing.

buzzdecafe:
> Hey, in first look it looks like the index file in the docs repo is generated from the README.md file here - is this the case or should I submit a separate PR for the docs repo?

no need to pr vs. ramda.github.io

benjamingr:
@buzzdecafe thanks. In that case the PR is done - you can merge it if you'd like.

Great work on the library, having something powerful like this in JS is really refreshing, as someone who's never seen the point of underscore given the native methods this has shown me a new light :) Asked that question and started this PR since we're actually deploying an internal app using Ramda. Thanks again for saving me time with the library.

buzzdecafe:
thanks for the kind words. can you please squash to one commit? then looks good to merge

benjamingr:
Done

buzzdecafe:
thanks!

PR #811, opened 2015-02-15
update .npmignore

davidchambers:
I think __/src/__ should be excluded as well. We needn't provide the source files in addition to the combined file.


davidchambers:
I'll merge this then publish v0.10.0. We can discuss whether __/src/__ should be included in future releases.

buzzdecafe:
:+1: on this pr, :-1: on including src/ in the distributed bundle

davidchambers:
> :-1: on including src/ in the distributed bundle

It's currently included, so you're actually :+1: (as I am) on *excluding* __/src/__ from future tarballs.

CrossEye:
:+1: on the :+1: disguised as a :-1:

buzzdecafe:
:-1: on including == :+1: on excluding

no contradiction here

PR #813, opened 2015-02-15
use op-style currying for R.has

davidchambers:


davidchambers:
Any objections? :)

buzzdecafe:
have you benched it? I'm curious to see what it costs

davidchambers:
> have you benched it? I'm curious to see what it costs

There's no cost as we're using uncurried `_has` internally.

buzzdecafe:
i think i responded to the wrong pr

davidchambers:
Does anyone object to merging this pull request? I'd like to pull in these changes regardless of what happens with #817 as there's value in having an uncurried version of such an important function.

buzzdecafe:
do it

:cow2:

PR #816, opened 2015-02-17
clean up root dir, closes #802

buzzdecafe:

PR #820, opened 2015-02-17
Add docs badge to README

rrrene:
Hi there,

I want to propose to add this badge to the README to show off inline-documentation: [Inline docs](http://inch-ci.org/github/ramda/ramda)

The badge links to [Inch CI](http://inch-ci.org) and shows an evaluation by [InchJS](http://trivelop.de/inchjs), a project that tries to raise the visibility of inline-docs. Besides testing and other coverage, documenting your code is often neglected although it is a very engaging part of Open Source.

So far over 500 **Ruby** projects are sporting these badges to raise awareness for the importance of inline-docs and to show potential contributors that they can expect a certain level of code documentation when they dive into your project's code and motivate them to eventually document their own. I would really like to do the same for the **JavaScript** community and roll out support for JS over the coming weeks (early adopters are [forever](https://github.com/foreverjs/forever), [node-sass](https://github.com/sass/node-sass) and [when](https://github.com/cujojs/when)).

Although this is "only" a passion project, I really would like to hear your thoughts, critique and suggestions. Your status page is http://inch-ci.org/github/ramda/ramda

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@ A practical functional library for Javascript programmers.
 [![Build Status](https://travis-ci.org/ramda/ramda.svg?branch=master)](https://travis-ci.org/ramda/ramda)
 [![npm module](https://badge.fury.io/js/ramda.svg)](https://www.npmjs.org/package/ramda)
 [![dependencies](https://david-dm.org/ramda/ramda.svg)](https://david-dm.org/ramda/ramda)
+[![Inline docs](http://inch-ci.org/github/ramda/ramda.svg?branch=master)](http://inch-ci.org/github/ramda/ramda)
 [![Gitter](https://badges.gitter.im/Join Chat.svg)](https://gitter.im/ramda/ramda?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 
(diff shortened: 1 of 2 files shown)


buzzdecafe:
Hi @rrrene 

Looks interesting! At present it is parsing things like `mocha.js` and `ie8-shim.js` which should really be excluded from the evaluation. How would we configure that if we were to add this badge?

rrrene:
You could configure it via a config file. I added it to the PR and the analysis changed accordingly: [Inline docs](http://inch-ci.org/github/ramda/ramda)

Please note that this project is not about getting the fully green bar (albeit ramda doing just that), but it is about highlighting the importance of inline docs to aspiring node developers. I want to achieve that people know that they can dive into open source code to see how it all works. The badge can signal to them that they won't be alone, that there are some inline-docs to guide them.

davidchambers:
Another config file. :\

Inch has an admirable goal, but I'd prefer to wait for demand from users before adding another badge and config file to a project already bursting with badges and config files (see #816).

davidchambers:
Closing due to insufficient excitement.

PR #824, opened 2015-02-18
build: remove dependence on file system ordering

davidchambers:
Different platforms sort `A.js B.js a.js b.js` differently.

This change has another benefit, which is ensuring that…

    $ scripts/build -- src/{foo,bar}.js

and…

    $ scripts/build -- src/{bar,foo}.js

produce the same output.


CrossEye:
:herb:

PR #827, opened 2015-02-19
preserve arity in R.bind

davidchambers:
I discovered this bug while using `R.bind` to assert that `R.curry` and `R.curryN` preserve context.


CrossEye:
Makes sense.  :+1:

PR #819, opened 2015-02-17
support R.__ placeholder in curry functions

davidchambers:
See #817

Changes:

  - add internal `_curryN` function, similar to `op` but capable of currying functions of arbitrary arity;
  - define `curryN` as `_curryN(2, _curryN)`;
  - update `_checkForMethod` to return a function with the appropriate arity;
  - replace occurrences of `_curry2`, `_curry3`, and `op` with `curry`; and
  - remove `_curry2`, `_curry3`, and `op`.

Reasons to like these changes:

  - it will put an end to ad hoc pull requests adding "op-style" currying to one or two functions at a time;
  - it improves consistency✝;
  - partial application is not *only* useful for binary functions;
  - it removes the distinction between "operator" functions and other functions, simplifying Ramda's mental model and making `R.__` easier to explain; and
  - it further promotes partial application, Ramda's key feature.

Reasons to dislike these changes:

  - it's a breaking change✝✝; and
  - it increases the amount of work required to apply a curried function.

✝ Currently commutative functions do not support `R.__`, as it's not useful. For consistency, though, they should. One could then say that `R.curry(f)(R.__, b)(a)` is equivalent to `f(a, b)` for any binary function `f`, in the same way one can currently say that `R.curry(f)(a)(b)` is equivalent to `f(a, b)` for any binary function `f`. (All the discussion in Fantasy Land seems to have got me thinking in terms of laws and whatnot.)

✝✝ Currently `R.divide(R.__)` is equivalent to `R.flip(R.divide)`, which has always seemed to me like too much magic. I doubt anyone is using this form, as the `R.flip` version is much clearer.

Issue #800, opened 2015-02-13
Continious release


TheLudd:
Ramda is a really active library when it comes to contributions and that's just great. However, I don't see that reflected in releases. IMHO, they could come more often. I think it would be great if the library could be released on a regular basis, say every Friday or something like that. The release would just contain whatever had been merged during the week. No need to say "We must get feature X into release Y". 

For this to work I suspect all of this needs to be true (some things I suspect/know are already working) 
 * Release should be a one-command process which would:
    * Run all the tests
    * Deploy to npm and other package managers of choice 
    * Create a tag in git
    * Generate change log
    * Publish the site with updated documentation

There might be a change in work flow regarding breaking changes. If a breaking change is merged that requires a major version bump (or minor in pre 1.0.0 land). And perhaps we want to bundle several breaking changes together in one release to not let the version numbers run amok. This could be addressed by simply not merging breaking changes until it is decided that they should be released. I.e. a breaking change may get a :cow2: stamp and be considered done, but we hold of the merge for a while. That solution might cause a problem for those of you running on the "latest" version. But I assume that solutions can be found for this.

Anyway. Hoping for a nice discussion here.

buzzdecafe:
every friday might be a bit much, but i'd be ok with every other week.

CrossEye:
:+1: for every two weeks.  My only concern is that it's not always as busy as it seems to be now, and pushing just to push is a little crazy, but if we can at least build the routine of pushing every other week if there are any significant changes.

Sorry, I'm being slow here.  Been sick and distracted by conversations over on FantasyLand.  Too much happening.  

I think most of this would be easy to do, except possibly the automated change log.  I know that people have built change-log generators from Git commits, but I prefer the curated approach @davidchambers has been demonstrating.

davidchambers:
Recently we've been publishing a new minor version every six weeks or so. This feels about right to me. I don't think we need a formal release schedule.

TheLudd:
To me, six weeks is way to long, especially in this phase. What is the good argument for letting time pass so long between releases? (assuming that all that was needed for a release was the execution of one command)

CrossEye:
I agree with @davidchambers that we don't have to commit to a formal release schedule.  But I do think we could be better about putting out more of what would (after 1.0) be termed patch-level releases, but here could be used like point releases.  When we have a few new features or a few fixes, even if they're only documentation fixes, we could move from, e.g.,  0.10.3 to 0.10.4 easily enough.

[...]

PR #830, opened 2015-02-21
remove R.appendTo and R.prependTo

davidchambers:
These functions were added in #500 but are no longer necessary thanks to #819.

```javascript
// before           // after
R.appendTo(xs)      R.append(R.__, xs)
R.prependTo(xs)     R.prepend(R.__, xs)
```


CrossEye:
:herb:

buzzdecafe:
:cow2:

CrossEye:
Before we merge this, please see the discussion in #831.  Should we do further function removals/renames in two steps, with a version or two that simply add a deprecated tag to the documentation and changelog before we commit them?

davidchambers:
I'd like to publish v0.12.0 this weekend (unless anyone objects) so people can start using transducers. I'd like to merge this pull request and the others which remove functions deprecated in v0.11.0. Does this sound reasonable?

megawac:
I'd say do a pre-release

CrossEye:
@davidchambers:does x.y.z give us a straightforward way to create _alpha_, _beta_, and _rc_ versions?   If so, then I like @megawac's suggestion.  If not, we really need to figure that out for our post-1.0 world, but I would be fine with releasing 0.12.0 sooner.

davidchambers:
> does x.y.z give us a straightforward way to create _alpha_, _beta_, and _rc_ versions?

Not quite, at this stage. From the [source][1]:

    -i --increment <level>
            Specify the level of the current version number to increment.
            Valid levels: 'major', 'minor', 'patch', 'premajor', 'preminor',
            'prepatch', and 'prerelease'. 'patch' is assumed if this option
            is omitted.

We can do this:

```console
$ node_modules/xyz/node_modules/.bin/semver --increment preminor 0.11.0
0.12.0-0
```

But if we wish to use some prefix we'll need to specify `--preid`, which xyz does not currently support:

```console
$ node_modules/xyz/node_modules/.bin/semver --increment preminor --preid rc 0.11.0
0.12.0-rc.0
```

I'll update xyz and publish a version which supports `--preid`.

If we wish to publish prerelease versions we'll need to adopt the branching strategy outlined in https://github.com/ramda/ramda/issues/800#issuecomment-74550317 (or a similar strategy). This adds significant complexity.

@megawac and @CrossEye, what do you see as the benefits of prerelease versions?


[1]: https://github.com/davidchambers/xyz/blob/v0.5.0/xyz#L22-L26

CrossEye:
> If we wish to publish prerelease versions we'll need to adopt the branching strategy outlined in https://github.com/ramda/ramda/issues/800#issuecomment-74550317 (or a similar strategy). This adds significant complexity.

I don't think we do.  My thought at least was that we would simply move at times from `x.y.z` to `x.(y+1).0-rc.0` and maybe to `-rc.1`, etc. before moving to `x.(y+1).0`.  If we need `alpha` or `beta` we could introduce those too, although I can't see that.

> @megawac and @CrossEye, what do you see as the benefits of prerelease versions?

My reason would be the way I and many others use such tools.  If I'm in a hurry for new features, I might take a beta or rc version and test it, but be ready to roll back because I know that there might be issues.  I don't expect those to be production-ready.  When I take a full `x.y.z` version, I expect the kinks have been worked out.  Publishing something with as dramatic changes as the transducers fix would be a good candidate for such a release.  Honestly if I'd thought about it, the `_curry` changes were probably also in that category.

PR #831, opened 2015-02-21
remove R.propOf

davidchambers:
This function is no longer necessary thanks to #819.

```javascript
// before           // after
R.propOf(obj)       R.prop(R.__, obj)
```


davidchambers:
This shouldn't be merged yet. While working on a patch to curry all the remaining functions I discovered that `R.prop` is manually curried (so does not yet support `R.__`).

CrossEye:
:herb:

buzzdecafe:
:cow2:

TheLudd:
Please no. I use this. Why is one better than the other? I would much prefer `propOf`. What is the cost of keeping it and its siblings?

TheLudd:
At least, leave these functions in a few versions. Have releases where both alternatives exist so that people can upgrade, still have a working application, migrate to the new way of doing things, and then upgrade to a new version where the old stuff is removed. 

And if you don't want the deprecate warning function, at least you can put a deprecated tag on the old functions documentation.

CrossEye:
The point is no so much that `prop(_, obj)` is better than `propOf(obj)`.  Actually, I find it very slightly worse.  But not having to maintain duplicate functionality is quite important.  Having one clear way of how to do things with Ramda is a clear win.  And, once this function is curried the way others are (it isn't now for some nasty historical reasons to do with performance dialogues), we will get the placeholder version for free.  All Ramda functions of more than a single parameter will be curried, as they currently are, but will also accept placeholders for partial application.  That simply makes `propOf` redundant.

I would certainly consider postponing it for a brief time, putting a `deprecated` tag in the documentation and in the changelog for 0.10.1 or some such.  What say you, @davidchambers, @buzzdecafe?

PR #832, opened 2015-02-21
update @param for list functions

davidchambers:
Surprisingly the `list` parameter to `R.head` and friends was once optional. This pull request updates each function's `@param` to match current behaviour.


CrossEye:
:herb:

PR #833, opened 2015-02-23
fix require paths

davidchambers:
The build script goes to [great lengths][1] to ensure that the various modules can be safely concatenated to produce __dist/ramda.js__. It ensures for each `require('...')` that the variable name matches the last path component. It does not, however, ensure that the path component is valid. As a result, one could write…

```javascript
var _has = require('./does/not/exist/_has');
```

and the build script would not complain. __dist/ramda.js__ would be valid, but requiring the module above would lead to an import error.

The solution to this problem is to run the test suite twice: once against __dist/ramda.js__ and once against __index.js__, a temporary file which simply requires and exports each of the individual modules.

[1]: https://github.com/ramda/ramda/blob/v0.10.0/scripts/build#L76-L95

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ DEEDPOLL = node_modules/.bin/deedpoll \
 	--rename fnArity:length \
 	--rename i:idx \
 	--rename index:idx
+MOCHA = node_modules/.bin/mocha
 UGLIFY = node_modules/.bin/uglifyjs
 XYZ = node_modules/.bin/xyz --repo git@github.com:ramda/ramda.git --script scripts/prepublish
 
@@ -48,3 +49,10 @@ setup:
 .PHONY: test
 test: dist/ramda.js
 	npm test
+	find src -name '*.js' -not -path 'src/internal/*' \
+	| sed 's:src/\(.*\)[.]js:exports.\1 = require("./src/\1");:' >index.js
+	sed '/"main":/d' package.json >tmp
+	mv tmp package.json
+	$(MOCHA) -- $(shell find test -name '*.js' -not -name 'installTo.js' -not -name 'test.examplesRunner.js')
+	git checkout -- package.json
+	rm index.js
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -48,3 +49,9 @@ setup:
 .PHONY: test
 test: dist/ramda.js
 	npm test
+	find src -name '*.js' -not -path 'src/internal/*' \
+	| sed 's:src/\(.*\)[.]js:exports.\1 = require("./src/\1");:' >index.js
```
davidchambers: First, create __index.js__ which will contain:

```javascript
exports.__ = require("./src/__");
exports.add = require("./src/add");
exports.all = require("./src/all");
...
exports.zip = require("./src/zip");
exports.zipObj = require("./src/zipObj");
exports.zipWith = require("./src/zipWith");
```

Makefile
```diff
@@ -48,3 +49,9 @@ setup:
...
+	sed -i '' '/"main":/d' package.json
```
davidchambers: Then, remove the `"main"` entry from __package.json__ so the entry point for the package is __index.js__ rather than __dist/ramda.js__.

Makefile
```diff
@@ -48,3 +49,9 @@ setup:
...
+	$(MOCHA) -- $(shell find test -name '*.js' -not -name 'installTo.js' -not -name 'test.examplesRunner.js')
```
davidchambers: Run the test suite (a second time), skipping tests for `R.installTo` since this function requires access to `R` which is not available.

Makefile
```diff
@@ -48,3 +49,9 @@ setup:
...
+	git checkout -- package.json
```
davidchambers: Revert our change to __package.json__. This will also revert any previous unstaged changes.

Makefile
```diff
@@ -48,3 +49,9 @@ setup:
...
+	rm index.js
```
davidchambers: Finally, delete the temporary file.

buzzdecafe:
and `index.js` is dynamically generated from src dir? LGTM

davidchambers:
> `index.js` is dynamically generated from src dir?

That's right. It's created and deleted during the running of `make test`.

Issue #829, opened 2015-02-20
Include src in npm package


matthewp:
It's common for developers using a commonjs loader to require only the package they are using. You would want to do something like

```js
var isEmpty = require('lamda/src/isEmpty');
```

This is especially important on the client where you don't want to bring in dependencies you're not using.

Additionally `src` is usually by convention a folder for files that then get built into the main package (if you were using an es6 transpiler for example).  I know that you are building to `dist` but if modules could be required directly `lib` might be a better place.

CrossEye:
Funny, we just pulled it out.  But you're right, there is at least some good case for restoring it.

@buzzdecafe, @davidchambers: Any compelling reason to exclude it?  Anything more than tidiness?

davidchambers:
> Funny, we just pulled it out.

We discussed the possibility in #811, but haven't actually done so.

CrossEye:
Oh, I just assumed... :-)

Well maybe we should leave `src` in.  I do like this use-case.

matthewp:
@CrossEye What about renaming it to `lib`?

buzzdecafe:
i also like this use case, let's keep it

CrossEye:
@matthewp: Can you make a compelling case for changing folder names?  It's not that I have a major objection, but I also don't want to do this without good reason.  Is there a standard, or an overwhelming consensus around `lib`?

matthewp:
@CrossEye Sure, just that `src` usually denotes code that needs to be built before it can be used. `var isEmpty = require('ramda/src/isEmpty');` **feels** like an unsupported API even if it really is supported. `var isEmpty = require('ramda/lib/isEmpty');` still feels like you're doing something unintended but it should still work.  So my case is basically conventions, `src` is where es6/coffeescript kind of things go and `lib` is where you're library's javascript modules go.

What would be even better would be if you could do `var isEmpty = require('ramda/isEmpty');`.   Unfortunately Node doesn't provide a way to make this work while keeping your code somewhere other than the package's root folder...  You could do what Lodash does and have the npm-published code in a separate branch (he uses the `npm` branch) and not worry about code organization there.  The downside is that this requires an additional build step in your publishing process.

tldr; I'd be happy so long as we can directly require modules whether in src or lib or wherever.  If you did decide to go with option 3 (publishing from a different branch) i'd be happy to contribute there.

davidchambers:
[...]

PR #834, opened 2015-02-23
remove R.installTo

davidchambers:
I've never liked this function, as it promotes what I believe to be a bad practice, but I haven't felt compelled to remove it until now. #829 makes a good case for allowing modules to be required individually; #833 revealed that `R.installTo` does not (and perhaps cannot) work when used this way.

Let's remove the one function in the library which breaks referential transparency.


TheLudd:
Like in the discussion for removing `propOf`: if this *is* to be removed, please consider deprecating it first. It just seems like good practice.

https://github.com/ramda/ramda/pull/831#issuecomment-75371284

davidchambers:
> please consider deprecating it first

That's fine with me. It remains to be seen whether we'll remove it at all. ;)

CrossEye:
I'm going to suspect that I'm the only one who would really miss this.

It's here because this is how I like to work.  I know that it doesn't work for everyone, and that using it the way I prefer could lead to conflicts with other functions, but I simply prefer this:

    pipe(identifierToFilename,
           parse,
           prop('ast'),
           prop('body'),
           takeWhile(where({type: 'VariableDeclaration', kind: 'var'})),
           pluck('declarations'),
           map(tap($pathEq('length', 1))),
           map(head),
           map(tap($pathEq('init.type', 'CallExpression'))),
           map(tap($pathEq('init.callee.type', 'Identifier'))),
           map(tap($pathEq('init.callee.name', 'require'))),
           map(tap($pathEq('init.arguments.length', 1))),
           map(tap($pathEq('init.arguments.0.type', 'Literal'))),
           tap(pipe(pluck('id'),
                        pluck('name'),
                        converge(assert.deepEqual,
                                   I,
                                   sortBy(I)))),
           map(tap(converge(assert.strictEqual,
                                  path('id.name'),
                                  pipe(path('init.arguments.0.value'),
                                         replace(/^[.][/]internal[/]/, './'),
                                         replace(/^[.]{1,2}[/]/, ''),
                                         replace(/[.]([a-z])/g,
                                                   pipe(nthArg(1),
                                                          toUpper)))))),
           map(path('id.name')));

to this:

    R.pipe(identifierToFilename,
           parse,
           R.prop('ast'),
           R.prop('body'),
           R.takeWhile(R.where({type: 'VariableDeclaration', kind: 'var'})),
           R.pluck('declarations'),
           R.map(R.tap($pathEq('length', 1))),
           R.map(R.head),
           R.map(R.tap($pathEq('init.type', 'CallExpression'))),
           R.map(R.tap($pathEq('init.callee.type', 'Identifier'))),
           R.map(R.tap($pathEq('init.callee.name', 'require'))),
           R.map(R.tap($pathEq('init.arguments.length', 1))),
           R.map(R.tap($pathEq('init.arguments.0.type', 'Literal'))),
           R.tap(R.pipe(R.pluck('id'),
                        R.pluck('name'),
                        R.converge(assert.deepEqual,
                                   R.I,
                                   R.sortBy(R.I)))),
           R.map(R.tap(R.converge(assert.strictEqual,
                                  R.path('id.name'),
                                  R.pipe(R.path('init.arguments.0.value'),
                                         R.replace(/^[.][/]internal[/]/, './'),
                                         R.replace(/^[.]{1,2}[/]/, ''),
                                         R.replace(/[.]([a-z])/g,
                                                   R.pipe(R.nthArg(1),
                                                          R.toUpper)))))),
           R.map(R.path('id.name')));


There is one **huge** benefit, though, of removing it: it is the only function that modifies user input data.  Granted it's a very special case, meant for very different use than the rest of the functions in the library.  But it really would be nice to give a categorical statement that Ramda never modifies user data.

I guess I'd probably have to give it a reluctant :+1: and keep it in my back pocket for my own use, but I would like to hear if there are others using it.

buzzdecafe:
it is nice to import the names into global space, but i don't think it's worth it either. i also agree with @TheLudd that we should deprecate and warn for a minor version or two and then remove.

:cow2:

TheLudd:
It could also be kept in another project, let's say we call it `ramda-dirty-stuff`. Where these kind of non pure functions could live.

buzzdecafe:
we have an incipient "impure" extension. "dirty-stuff" might get more hits though :smiley_cat: 

The "impure" extension has languished, probably because none of us is particularly interested in enabling mutating, imperative code.

lucasmreis:
First of all, thanks a lot for the amazing library! We use it on production at work, and it's amazing.

I also prefer writing code like @CrossEye , without all those `R.`, so we just write in the beginning of each module:

    var get = R.get;
    var compose = R.compose;
    var pCompose = R.pCompose;
    var assoc = R.assoc;

For each function we use. It's not pretty to have all that in the beginning of the file, but we feel our actual code gains A LOT by not being written with a lot of `R.`, and we still do not have everything global.

My coworker @jsanchesleao is eager to use ES6 destructuring for that:

   {get, compose, pCompose, assoc} = R;

One more time, thank you very much! :)

davidchambers:
> [T]hanks a lot for the amazing library!

I :heart: enthusiastic, appreciative users. Ramda would not exist in its current form were it not for numerous suggestions and improvements from the community. Thank you for being part of it.

Using destructuring assignment on a module-by-module basis is an elegant way to enjoy `R.`-free programming. :)

lucasmreis:
:+1:

semmel:
@lucasmreis @CrossEye 
Regarding `R.`-free programming in a browser I'd like to throw this snippet in:
```javascript
R.map(function(funcName){ 
		window[funcName] = R[funcName]; 
	},
	R.difference(R.functions(R), R.keysIn(window))
);

window.R = R; // for the rest of things (right now it's find and length)
```

CrossEye:
@semmel: Nice!  I use a slightly different technique, but similar enough.

I've never bothered with the `R.difference` bit, but it makes sense in case we step on the toes of something important.  (I personally never use `window.length` or `window.find`.)  For the foreseeable future, the only difference between `R.keysIn(R)` and `R.functions(F)` is the existence of the placeholder `__` in the former.  That's one I definitely want available; others who do as well should simply switch functions here.

lucasmreis:
@semmel: I always try not to put functions on `window`, but maybe it's just an unnecessary fear! ;)

semmel:
@CrossEye I too hadn't bothered with `R.difference(R.keysIn(R), R.keysIn(window))` until I've recently discovered that the *Stock Android 4.2 Browser* doesn't let you reassign the `window.length` property. 
I.e.
```javascript
window["length"] = function dummy(){ return undefined; }
typeof length // "number" in Stock Android 4.22 Browser, "function" everywhere else
``` 
Thus if you target that browser you must always use `R.length` because the global `length` is not writable.

@lucasmreis 
>I always try not to put functions on window, but maybe it's just an unnecessary fear! ;)

Why is that? Since I doubt that there is a `this` in any Ramda function implementation pointing to `R`, I guess it's the same as simply writing
```html
<script>
function foo(){ bar(); }
function isEmpty(list) {
    return Object(list).length === 0;
};
</script>
```
or is it not?

lucasmreis:
You're right. That example is a simple explanation of how unnecessary the fear is. :+1:

PR #840, opened 2015-02-25
support array-like objects in R.slice

### LINKED ISSUE #839: R.init(arguments) => TypeError: Cannot read property 'apply' of undefined
Is this a known issue?

[Ramda REPL](http://goo.gl/hyAdQ7)

davidchambers:
Fixes #839

### DIFF
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -4420,22 +4420,25 @@
     });
 
     /**
-     * Returns the elements from `xs` starting at `a` and ending at `b - 1`.
+     * Returns a list containing the elements of `xs` from `fromIndex` (inclusive)
+     * to `toIndex` (exclusive).
      *
      * @func
      * @memberOf R
      * @category List
      * @sig Number -> Number -> [a] -> [a]
-     * @param {Number} a The starting index.
-     * @param {Number} b One more than the ending index.
+     * @param {Number} fromIndex The start index (inclusive).
+     * @param {Number} toIndex The end index (exclusive).
      * @param {Array} xs The list to take elements from.
-     * @return {Array} The items from `a` to `b - 1` from `xs`.
+     * @return {Array} The slice of `xs` from `fromIndex` to `toIndex`.
      * @example
      *
      *      var xs = R.range(0, 10);
      *      R.slice(2, 5)(xs); //=> [2, 3, 4]
      */
-    var slice = invoker(2, 'slice');
+    var slice = _curry3(_checkForMethod('slice', function slice(fromIndex, toIndex, xs) {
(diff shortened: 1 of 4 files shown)


megawac:
The flip side is this will break on objects with custom slices

davidchambers:
> The flip side is this will break on objects with custom slices

True. Do you think we should special-case Arguments objects?

davidchambers:
I updated the pull request to use `_checkForMethod`, and added a test case for an object with a custom `slice` method.

dist/ramda.js
```diff
@@ -4420,22 +4420,25 @@
...
-     * @return {Array} The items from `a` to `b - 1` from `xs`.
+     * @return {Array} The slice of `xs` from `fromIndex` to `toIndex`.
      * @example
      *
      *      var xs = R.range(0, 10);
      *      R.slice(2, 5)(xs); //=> [2, 3, 4]
      */
-    var slice = invoker(2, 'slice');
+    var slice = _curry3(_checkForMethod('slice', function slice(fromIndex, toIndex, xs) {
+        return Array.prototype.slice.call(xs, fromIndex, toIndex);
```
buzzdecafe: why not call internal `_slice`?

test/init.js
```diff
@@ -15,4 +15,8 @@ describe('init', function() {
         assert.throws(function() { R.init(undefined); }, TypeError);
         assert.throws(function() { R.init(); }, TypeError);
     });
+    it('handles array-like object', function() {
+        var args = (function() { return arguments; }(1, 2, 3, 4, 5));
+        assert.deepEqual(R.init(args), [1, 2, 3, 4]);
+    });
```
buzzdecafe: important to note that unless the array-like object has its own `slice` method, `R.slice` *always* returns an Array, not a "sliced" version of the input array-like object.

dist/ramda.js
CrossEye: Internal slice only handles positive values.

src/slice.js
```diff
@@ -1,20 +1,24 @@
-var invoker = require('./invoker');
+var _checkForMethod = require('./internal/_checkForMethod');
+var _curry3 = require('./internal/_curry3');
 
 
 /**
- * Returns the elements from `xs` starting at `a` and ending at `b - 1`.
+ * Returns the elements of `xs` from `fromIndex` (inclusive) to
```
CrossEye: Perhaps

* Returns a list containing the elements from...

CrossEye:
LGTM

src/slice.js
davidchambers: Yes, that's better. Updated.

PR #842, opened 2015-02-25
fix signature of R.nth

### LINKED ISSUE #841: Documentation signature of nth is missing the first arg
Although that is kind of trivial, it is missing. It should be Number -> [a] -> [a] instead of [a] -> a

davidchambers:
Fixes #841


buzzdecafe:
:cow2:

PR #850, opened 2015-03-01
fix signature of R.propOr

### LINKED ISSUE #847: propOr documentation re. parameter order vs reality
The documentation states that the parameter order for R.propOr is property name, then defaultValue, then object.

But [in fact](https://github.com/ramda/ramda/blob/master/dist/ramda.js#L4103) it's defaultValue, then propertyName, then object.

davidchambers:
Closes #847

### DIFF
--- a/src/propOr.js
+++ b/src/propOr.js
@@ -10,11 +10,11 @@ var _has = require('./internal/_has');
  * @func
  * @memberOf R
  * @category Object
- * @sig s -> v -> {s: x} -> x | v
- * @param {String} p The name of the property to return.
+ * @sig a -> String -> Object -> a
  * @param {*} val The default value.
+ * @param {String} p The name of the property to return.
  * @param {Object} obj The object to query.
- * @return {*} The value of given property or default value.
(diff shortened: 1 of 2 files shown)


davidchambers:
Could one of you review this change? :)

src/propOr.js
```diff
@@ -10,9 +10,9 @@ var _has = require('./internal/_has');
  * @func
  * @memberOf R
  * @category Object
- * @sig s -> v -> {s: x} -> x | v
- * @param {String} p The name of the property to return.
+ * @sig a -> String -> Object -> a
  * @param {*} val The default value.
+ * @param {String} p The name of the property to return.
  * @param {Object} obj The object to query.
  * @return {*} The value of given property or default value.
```
buzzdecafe: should it be:

>  The value of given property *of the supplied object* or *the* default value.

?

CrossEye: I like the formulation from @buzzdecafe

davidchambers: Nice tweak. Updated.

PR #855, opened 2015-03-01
lensOn for a particular object

buzzdecafe:
This allows you to bind a lens to a particular object. Should the object come first? or last?

### DIFF
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -1853,6 +1853,41 @@
     });
 
     /**
+     * Returns a lens associated with the provided object.
+     *
+     *
+     * @func
+     * @memberOf R
+     * @category Object
+     * @sig {} -> ({} -> v) -> (v -> a -> *) -> (a -> b)
+     * @see R.lens
+     * @param {Function} get A function that gets a value by property name
+     * @param {Function} set A function that sets a value by property name
+     * @return {Function} the returned function has `set` and `map` properties that are
+     *         also curried functions.
+     * @example
+     *
+     *     var xo = {x: 1};
+     *     var xoLens = R.lensOn(xo,
+     *                           function get(o) { return o.x; },
+     *                           function set(v) { return {x: v}; });
+     *     xoLens(); //=> 1
+     *     xoLens.set(1000); //=> {x: 1000}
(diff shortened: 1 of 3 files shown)


CrossEye:
What's the use for this?  Especially with `set`, which presumably would usually return a new object?

But if we're going to do it, I would suggest that the object belongs last, and even if we can't use our normal `curry3` function to do it, then the existing `lens` could be considered just curried version of it.

buzzdecafe:
the motivation is mostly to enable composing with lenses using `map`. at present `lens.map` requires you to pass it the object to map over as well as the mapper function. 

i think these are different enough not to just treat them as curried versions of the same thing.

CrossEye:
makes sense.

:herb:

dist/ramda.js
```diff
@@ -3265,6 +3265,41 @@
...
+     *         also curried functions.
+     * @example
+     *
+     *     var xo = {x: 1};
+     *     var xoLens = R.lensOn(xo,
+     *                           function get(o) { return o.x; },
+     *                           function set(v) { return {x: v}; });
+     *     xoLens(); //=> 1
+     *     xoLens.set(1000) //=> {x: 1000}
+     *     xoLens.map(R.add(1)) //=> {x: 2}
```
davidchambers: Nit: Missing semicolons on the lines above.

buzzdecafe: ok. any thoughts on param order? i'm leaning toward object-last myself now

davidchambers: Object last feels right to me.

CrossEye: yes, definitely object-last.

buzzdecafe: yay we all agree for a change. i will amend and resubmit

buzzdecafe:
params corrected

davidchambers:
Could you rebase and resolve the merge conflicts?

buzzdecafe:
rebased -- take that, @CrossEye! :stuck_out_tongue_winking_eye:

davidchambers:
:deciduous_tree:

CrossEye:
I'm trying to take it!  :wink:  But I seem to have a stubborn system that won't let go.

PR #856, opened 2015-03-02
remove special case for curried function applied to zero arguments

davidchambers:
As discussed in #836


buzzdecafe:
bugs me because calling a curried function with no arguments is an error. *caveat emptor*. 

go ahead though

CrossEye:
:herb:

PR #857, opened 2015-03-02
remove extensions

davidchambers:
These extensions are neglected and have never been fleshed out. If any of these are to live on in some form I suggest separate repositories. This seems to be working well for [ramda-fantasy][1].

[1]: https://github.com/ramda/ramda-fantasy


davidchambers:
Thoughts? Should these be deprecated for one minor version before being removed?

buzzdecafe:
> Should these be deprecated for one minor version before being removed?

are you trolling now? :smile: 

I guess lazylist can move to a ramda repo by itself. And `debug` and `impure` are barely there. And these files were touched 3 months ago and not since. So ..... I guess it is ok to remove.

PR #862, opened 2015-03-02
test: remove redundant test cases for R.__

davidchambers:
Thanks to #836 we needn't test this for each function individually.

### DIFF
--- a/test/concat.js
+++ b/test/concat.js
@@ -35,11 +35,6 @@ describe('concat', function() {
         assert.deepEqual(conc123([4, 5, 6]), [1, 2, 3, 4, 5, 6]);
         assert.deepEqual(conc123(['a', 'b', 'c']), [1, 2, 3, 'a', 'b', 'c']);
     });
-    it('is curried like a binary operator, that accepts an inital placeholder', function() {
-        var appendBar = R.concat(R.__, 'bar');
-        assert.strictEqual(typeof appendBar, 'function');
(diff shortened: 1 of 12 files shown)


buzzdecafe:
:cow2:

CrossEye:
I have mixed feelings about this.

To my mind, test cases both test and describe the public API.  Since this behavior is now universal to Ramda, there's some reason to just assume it for individual functions.  But I'm loathe to drop these descriptive tests simply because our implementation has changed; part of the point of unit tests is to demonstrate constants in an API even as the underlying implementation changes.

So at the moment, I guess I'm :-1: .

davidchambers:
I see your point @CrossEye. It's misleading, though, to perform these assertions for just a handful of functions. It suggests that other functions do not support `R.__`. We could go the other way and add `R.__` tests to all the other functions. What say you?

CrossEye:
That's why my feelings were mixed.  I don't think we should be adding such tests all over the place.  

But perhaps we should as part of the recent changes add a curated few to note that this now works for more than operators.  Maybe the best way would be to write a test or two for each of the various `curry` functions that curry some simple functions and show that the placeholder works there.  (You may have already done so.) 

That would help demonstrate the additional functionality.  And by not removing the existing tests, we show that current functionality is not impaired.

davidchambers:
Okay. It seems we should take a different direction.

Issue #860, opened 2015-03-02
[Question] replaceable implementations for map and list?


sleepyfox:
I'd like to be able to use Ramda with the immutable data structures from Mori, I've seen that Eweda was built around the concept of pluggable List implementations, is this possible with Ramda?

CrossEye:
Unfortunately it's not.  Ramda does do dynamic dispatching of many core functions, and I would not object to extending it to additional ones, but that will probably only work with libraries that present some sort of OO methods.  It works quite well, for instance, with the [FantasyLand][fl] specification, which dictates its functions as object methods.

Eweda was a simpler project and built everything on top of a very few base functions.  That's how we could work with replaceable cores; all we had to do was implement those functions and we had a new implementation.  Unfortunately, it was also deadly slow.

I would love to be able to work with mori, with immutable-js, and the other Immutable collections libraries.  But the will all be a challenge, and mori will probably be more so.

If you have some suggestions for how to go about doing this, I would love to hear them.

  [fl]: https://github.com/fantasyland/fantasy-land

kevinbeaty:
See #878 for initial transducers implementation that  could potentially lead to support of Immutable.js / Mori in the future.

Issue #695, opened 2014-12-14
Supporting more than just arrays


tolmasky:
So, the issue I'm facing is that I want to use ramda everywhere in my project, but I am frequently dealing with other kinds of data structures (for example, facebook's immutable data structures). Currently I'm just wrapping up immutable's methods in curryN's to get similar behavior, but it really would be nice to translate the increasing amount of work in this library in other places (for example, immutable *does* have map, but it doesn't have some of the other funcs, and these just need to be re-written (often times copy-pasting with minimal changes)). I was wondering whether consideration had been given to generalizing these functions, perhaps against JS's iterator spec ( https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/The_Iterator_protocol )?

However, this only covers half the equation because it doesn't really define what the *result* type should be unfortunately. Perhaps (at least internally), iteration methods could be defined in terms of two base functions: next and push, which would form an iterable-instantiable tuple: [next, push]. For example, say we defined map as so:

```javascript
function map(f, iterable)
{
    next = iterable[0],
    push = iterable[1];
    result = push();
    while ([value, next, done] = next())
    {
        if (done)
            return result;

        result = push(result, f(value));
   }
}
```

So, for arrays it would be something like this:

```javascript
function arrayToIterable(anArray)
{
    return [ arrayNext(anArray, 0) concat]
}

function arrayNext(anArray, anIndex)
{
    function ()
    {
        return [anArray[anIndex], anIndex < anArray.length - 1, next(anArray, anIndex + 1)];
    }
}

function arrayPush(lhs, value)
{
    if ( arguments.length === 0)
       return [];

   lhs.push(rhs);

   return lhs;
}
```

So now map(f, arrayToIterable([1,2,3])) works fine. To make it even more painless for the array common case:

```javascript
function toIterable(anObject)
{
    if (anObject.next && anObject.concat)
        return anObject;

    if (Array.isArray(anObject))
        return arrayToIterable(anObject);

    if (anObject.toIterable)
        return anObject.toIterable();

    return objectToIterable(anObject);
}

function map(f, iterable)
{
    iterable = toIterable(iterable); // no op for existing iterables...
}
```

Now map(f, [1,2,3]) works again. And if we fill in the missing objectToIterable:

```javascript
function objectToIterable(anObject)
{
    return [ objectNext, objectPush];

}

function objectNext(anObject, keys, index)
{
[...]

Issue #546, opened 2014-11-19
Iterating objects


tomhicks-bsf:
I have a URL route (e.g. /things/:uid) into which I would like to pass parameters (e.g. { uid: 'abc123' }) to get a real URL (e.g. /things/abc123)

To me this is a simple reduce of the params object beginning with the URL route:

```
reduce(replace, route, params)
```

Because params is an object I have to `toPairs` it, and replace ends up being:

```
function (memo, value) { return memo.replace(":" + value[0], value[1]) }
```

Doing `toPairs`, and accessing `value[0]` and `value[1]` is there purely because Ramda does not iterate over objects, only arrays. Underscore/lodash manage this just fine.

Is there the possibility of having `reduceObj` and `forEachObj`, (like we have `mapObj`) or am I missing the point somehow?

PR #865, opened 2015-03-03
transducers initial commit

kedashoe:
Well, I'll start very briefly with some good bits:

- Everyone gets to write the same awesome Ramda code they've always been writing, only difference is now behind the scenes you'll find transducers for some operations.
- We can get rid of `mapObj`, `mapObjIndexed` (all transducer callbacks will get `value`, `key` and `collection`), `filterIndexed`, etc.
- It is very easy to add support for anything that follows the iterator protocol (Immutable.js already supported, need to have a discussion about how to generalize), streams, etc. (#695, #860, etc)
- Transducer speed scales very well to large data structures.

The bad (one of the? the worst?) part: the code I think is quite difficult to follow. I *really* wanted to find a way to properly visualize the code paths before submitting this PR, but I haven't come up with a great solution. However, I believe things are more or less working so that the core contributors can get an idea of what is going on, decide if you'd like to move in this direction, another direction with transducers, or somewhere else.

You can see some basic usage in `test/transduce.js`. I have simply been running

`make && ./node_modules/.bin/mocha test/transduce.js`

during development. I've also included in `dist/` some examples that follow a couple different code paths (obviouly these would be deleted if any of this is ever merged). I've added a `debugger;` line so if you open up the html examples in a browser and have your console open it should break before it starts doing anything important and you can step through to get a better idea of what's happening. I've started building some UML-ish diagrams as well which I think will be pretty nice when they are done, but not sure when that will be.

### Some notes in no particular order

As http://clojure.org/transducers tells us,

> Composition of the transformer runs right-to-left but builds a transformation stack that runs left-to-right

Now, if you've got pipes and compositions and some of their constituent functions are transducers and some are not, things could get really weird really fast. So all transducers in a **composition** are **piped** left-to-right so that they run right-to-left as you would expect in a composition (actually I've only implemented pipe as that is what I used in my work, we can just deprecate `compose`, right? Jokes! :stuck_out_tongue_winking_eye:  I just did not take the time to implement `compose` as it is simply the mirror of pipe). Let's look at a quick example here. Say you build a pipe

`pipe(a, b, c, d, e, f)`

where a-f are functions. If b, c and d are transducers, they will be composed when the pipe is built, so that when the pipe is actually run you still get a -> (b -> c -> d transducer stack) -> e -> f.

I haven't used this PR to re-implement everything that might possibly be a transducer yet, just 

- `map` and `filter` cause everyone loves mapping and filtering
- `take` to test terminating transduction
- `partitionBy` and `mapcat` because we use those in our implementation

Need to implement curry anywhere since we are "manually" currying.

This also causes problems with `checkForMethod` as currently used (`filter` tests for this and fails), I've just removed that call while the PR is under review.

I took out stream stuff as the default output collection (a write stream?) deserves some discussion.

There are some nice helper functions that are easy to implement like `into` (convert anything to anything else), `toList` (convert anything to a list), `toMap` (you get the idea) which I've not included.

Alright, it's getting late, I'm sure I'm forgetting plenty and regardless there are more todos and questions, but I figured it's best to just get this up and let you guys take the lead from here.

### DIFF
--- a/src/internal/_getTypedFunction.js
+++ b/src/internal/_getTypedFunction.js
@@ -1,0 +1,6 @@
+var _FN_TYPE_KEY = require('./_FN_TYPE_KEY');
+
+module.exports = function _getTypedFunction(type, f) {
+    f[_FN_TYPE_KEY] = type;
+    return f;
+};
--- a/src/internal/_makeXf.js
+++ b/src/internal/_makeXf.js
@@ -1,0 +1,50 @@
+var _XF_TRANSIENT_FLAG = require('./_XF_TRANSIENT_FLAG');
+var __ = require('../__');
+var _getTypedFunction = require('./_getTypedFunction');
+var _isXf = require('./_isXf');
+var _transduce = require('./_transduce');
+var _xfMakeType = require('./_xfMakeType');
+var is = require('../is');
+
+module.exports = function(){
+    
+    var transducable = function transducable(xfType, f) {
+        return function innerTransducable(xf) {
+            return new xfType(f, xf);
+        };
+    };
+
+    var _dispatchXf = function _dispatchXf(xfType, f, xfOrCollection) {
+        // got our reducing function, return our transformer
+        if (_isXf(xfOrCollection)) {
+            return new xfType(f, xfOrCollection);
+        }
+        // got a collection, transduce
+        else {
+            // todo? use apply here so user can pass in however many arguments they want
+            // and they will get passed along to transduce
+            return _transduce(transducable(xfType, f), xfOrCollection);
+        }
+    };
+    
+    var _xfWrapper = function _xfWrapper(xfType) {
+        // all our transducers are wrapped by this function
+        // ie when Ramda is loaded, we are at this point with map, filter, take, etc
+        return _getTypedFunction(_XF_TRANSIENT_FLAG, function _innerXfWrapper(f, xfOrCollectionOrUndefined) {
+            // got our collection
+            if (arguments.length > 1) {
+                return _dispatchXf(xfType, f, xfOrCollectionOrUndefined);
+            }
+            else {
+                return _getTypedFunction(_XF_TRANSIENT_FLAG, function _curryInnerXfWrapper(xfOrCollection){
+                    return _dispatchXf(xfType, f, xfOrCollection);
+                });
+            }
+        });
+    };
+    
+    return function _makeXf(xfDefinition) {
+        var type = _xfMakeType(xfDefinition);
+        return _xfWrapper(type);
+    };
+}();
--- a/src/internal/_reduce.js
+++ b/src/internal/_reduce.js
@@ -1,7 +1,126 @@
-module.exports = function _reduce(fn, acc, list) {
-    var idx = -1, len = list.length;
-    while (++idx < len) {
-        acc = fn(acc, list[idx]);
-    }
-    return acc;
-};
+var _isArray = require('./_isArray');
+var _isIterable = require('./_isIterable');
+var _isXf = require('./_isXf');
+var _xfDefaultInit = require('./_xfDefaultInit');
+var _xfReducer = require('./_xfReducer');
+
+module.exports = function (){
+
+    var isReduced = function(x) {
+        return x.__transducers_reduced__ === true;
+    };
+    
+    var isObject = function isObject(x) {
+        return x instanceof Object;
+    };
+    
+    var arrayReduce = function arrayReduce(xf, init, list) {
+        var acc = init;
+        var i;
+        for (i = 0; i < list.length; ++i) {
+            acc = xf.step(acc, list[i], i, list);
+            if (isReduced(acc)) {
+                acc = acc.value;
+                break;
+            }
+        }
+        return xf.result(acc);
+    };
+    
+    var objectReduce = function objectReduce(xf, init, obj) {
+        var acc = init;
+        var p;
+        for (p in obj) {
+            if (obj.hasOwnProperty(p)) {
+                acc = xf.step(acc, obj[p], p, obj);
+                if (isReduced(acc)) {
+                    acc = acc.value;
+                    break;
+                }
+            }
+        }
+        return xf.result(acc);
+    };
+    
+    var iterableReduce = function(xf, init, iterable) {
+        if (iterable.entries instanceof Function) {
+            return iteratorKeyValReduce(xf, init, iterable, iterable.entries());
+        }
+        else if (typeof Symbol !== 'undefined' && iterable[Symbol.iterator]) {
+            return iteratorValOnlyReduce(xf, init, iterable, iterable[Symbol.iterator]());
+        }
+        else if (iterable['@@iterator']) {
+            return iteratorValOnlyReduce(xf, init, iterable, iterable['@@iterator']());
+        }
+        else if (iterable.next instanceof Function) {
+            return iteratorValOnlyReduce(xf, init, iterable, iterable);
+        }
+        else {
+            throw 'Could not get iterator from ' + iterable.toString();
+        }
+    };
+
+    var iteratorValOnlyReduce = function iteratorValOnlyReduce(xf, init, iterable, iterator) {
+        var acc = init;
+        var step;
+        for (step = iterator.next(); !step.done; step = iterator.next()) {
+            acc = xf.step(acc, step.value, void 0, iterable);
+            if (isReduced(acc)) {
+                acc = acc.value;
+                break;
+            }
+        }
+        return xf.result(acc);
+    };
+
+    var iteratorKeyValReduce = function iteratorKeyValReduce(xf, init, iterable, iterator) {
+        var acc = init;
+        var step;
+        for (step = iterator.next(); !step.done; step = iterator.next()) {
+            acc = xf.step(acc, step.value[1], step.value[0], iterable);
+            if (isReduced(acc)) {
+                acc = acc.value;
+                break;
+            }
+        }
+        return xf.result(acc);
+    };
+
+    var observableReduce = function observableReduce(xf, init, observable, cb) {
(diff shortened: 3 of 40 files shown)


src/internal/_makeXf.js
```diff
@@ -0,0 +1,54 @@
...
+            return _transduce(transducable(xfType, f), xfOrCollection);
+        }
+    };
+    
+    var _xfWrapper = function _xfWrapper(xfType) {
+        // all our transducers are wrapped by this function
+        // ie when Ramda is loaded, we are at this point with map, filter, take, etc
+        return _getTypedFunction(_XF_TRANSIENT_FLAG, function _innerXfWrapper(f, xfOrCollectionOrUndefined) {
+            if (arguments.length === 0) {
+                throw _noArgsException();
```
davidchambers: `_noArgsException` was removed in #856.

src/partitionBy.js
```diff
@@ -0,0 +1,2 @@
+var _partitionBy = require('./internal/_partitionBy');
+module.exports = _partitionBy;
```
davidchambers: There's no need for `_partitionBy` if `partitionBy` is simply an :alien:.

davidchambers:
This is very exciting!

src/partitionBy.js
kedashoe: It is not an :alien: Sorry my comments are all over the place, the only note I made was in `src/internal/_partitionBy.js` that I was following this contract: https://clojuredocs.org/clojure.core/partition-by. It certainly seems to me that how `partition` is defined in Ramda can be useful, just so happened I needed the array to be partitioned exactly as it is defined by clojure's partition-by. I think if Ramda is going to include both, one of the names will have to change as the names `partition` and `partitionBy` do not give much insight into the actual difference.

src/internal/_makeXf.js
kedashoe: I noticed that happened while I was working, I like that change! I'll update the PR.

kedashoe:
I should have noted I followed

https://github.com/cognitect-labs/transducers-js#transformer-protocol

Also I see that this

https://github.com/cognitect-labs/transducers-js#reduced

is different from when I first looked at it, I'll update the PR at some point today

CrossEye:
I'm intrigued by this, but it's going to take time to digest.

I'm also concerned about performance.  To me, one of the big reasons for considering transducers was the promise of increased performance.  I don't know if will just take tweaks, or if there is something fundamental preventing it from operating faster.  Any ideas?

test/transduce.js
```diff
@@ -0,0 +1,90 @@
...
+        var taken = R.pipe(R.filter(even), R.take(2), R.map(square));
+        assert.deepEqual(taken(list), [ 0 , 4 ]);
+    });
+    
+    it('works with objects that support the iterator protocol', function(){
+        var ilist = Immutable.List(R.range(0, 5));
+        var imap = Immutable.Map(obj);
+        var pipe = R.pipe(R.filter(even), R.map(inc));
+        assert.deepEqual(pipe(ilist).toJS(), [ 1, 3, 5 ]);
+        assert.deepEqual(pipe(imap).toJS(), { A: 1, C: 3, E: 5 });
```
buzzdecafe: this is a *huge* win

src/internal/_reduce.js
```diff
@@ -1,7 +1,134 @@
...
+module.exports = function (){
+
+    var isReduced = function(x) {
+        return x.__transducers_reduced__ === true;
+    };
+    
+    var isObservable = function isObservable(x) {
+        return false;
+        // return x instanceof EventEmitter;
+    };
```
buzzdecafe: probably shouldn't include this stub?

buzzdecafe:
it's not clear to me from just reading the code if this will work with external transducer libraries?

src/internal/_getTypedFunction.js
```diff
@@ -0,0 +1,6 @@
+var _FN_TYPE_KEY = require('./_FN_TYPE_KEY');
+
+module.exports = function _getTypedFunction(type, f) {
+    f[_FN_TYPE_KEY] = type;
+    return f;
+};
```
buzzdecafe: this is more a `set`ter than a `get`ter

src/partitionBy.js
CrossEye: I think the point is that there are now two equal references to the same function: '`_partitionBy`' and '`partitionBy`'.  We do this sort of wrapping of internal functions to add currying or `checkForMethod`.  But here there seems to be no point.  It looks as though the internal version could simply be moved into the public implementation.

kedashoe:
> I'm also concerned about performance. To me, one of the big reasons for considering transducers was the promise of increased performance. I don't know if will just take tweaks, or if there is something fundamental preventing it from operating faster. Any ideas?

I did not run any benchmarks, but I think we can safely say any single transducer (ie just calling `R.map`) will be slower than its corresponding one-off implementation.

In order to preserve the current API, there is definitely some overhead when creating transducers and pipes/compositions. But once things get going you get normal transducer behavior, no intermediate collections and all that. I think this should scale well, but we'd need to test it. And I'm sure that we can make some tweaks to help out here and there.

Also somebody really smart once said

> the API is king

:stuck_out_tongue:

src/internal/_reduce.js
kedashoe: oops left in some stream stuff

src/internal/_getTypedFunction.js
kedashoe: Makes sense, what should I call it? _createTypedFunction or _makeTypedFunction or..?

kedashoe:
> it's not clear to me from just reading the code if this will work with external transducer libraries?

I think this is probably not necessary (looks like clojure defines less than 20 transducers), but it should be pretty easy. I should probably tweak how the types are created anyway, I just copied that straight over and it is sort of un-ramda-y

kedashoe:
This also solves #546 et al

CrossEye:
That's one of the things that scared me in your notes.  Does this actually add `index` or `key` parameters to things like `map`, `filter`, and `reduce`?  I really don't like that idea.

src/internal/_getTypedFunction.js
buzzdecafe: dunno! 

also, i think if we are going to start adding metadata to functions we should have a plan for it, and not just proceed in an _ad hoc_ fashion. And I do think we are going to start exploring the use of metadata on funcitons, probably in a `.ramda` namespace. I have an idea I wanna try out ....

kedashoe:
Works with 3rd party transducers now, for what that is worth.

> That's one of the things that scared me in your notes. Does this actually add index or key parameters to things like map, filter, and reduce? I really don't like that idea.

I guess I am with @megawac on this, but we do not have to include the key and collection

src/partitionBy.js
kedashoe: > I think the point is that there are now two equal references to the same function: '_partitionBy' and 'partitionBy'.

ha, that does appear to be what @davidchambers was saying, my fault. Updated.

kevinbeaty:
This is indeed exciting, but I have to ask. is there any reason you didn't base this PR on the `buzzdecafe:xduce` branch?  Seems like we vetted that pretty well... including removing the need for an `_XF_FLAG_`, supporting single value returns (e.g `find`) support for "functor" style method dispatch and probably others...

kedashoe:
> This is indeed exciting, but I have to ask. is there any reason you didn't base this PR on the buzzdecafe:xduce branch?

Just as a way to get things going since I was a lot more comfortable working from some of the code I had previously written and tested. I have absolutely zero problem if you want to submit the buzzdecafe:xduce branch as a separate PR so the maintainers can look at another approach.

kevinbeaty:
> I have absolutely zero problem if you want to submit the buzzdecafe:xduce branch as a separate PR so the maintainers can look at another approach.

Well, that is one of the maintainer's branches, and I was expecting that any day now ;)

@buzzdecafe  Thoughts?

buzzdecafe:
> I'd be happy to clean up that branch and submit it if you would like.

That sounds like an offer to good to refuse :smile:.
i keep *meaning* to do so, but then i wind up chasing some new shiny object.

kevinbeaty:
Shiny objects will get you every time.  That and bunnies.

OK.  I'll try to get something submitted this weekend.

buzzdecafe:
oooo shiny :rabbit:

kevinbeaty:
"Ramduce" alternative implementation in #878

kedashoe:
closed for #878

davidchambers:
I'm very grateful to you for opening this pull request, @kedashoe. It appears to have prompted #878, and some of us were starting to wonder whether anything tangible would come of the "ramduce" project. ;)

test/transduce.js
davidchambers: Does #878 also provide interop with such libraries, @kevinbeaty?

kedashoe: Yep! eg


    assert.deepEqual(
        R.transduce(
            R.map(add(1)),
            R.flip(R.invoker(1, 'push')),
            Immutable.List(),
            Immutable.List([1, 2, 3, 4])
        ).toJS(),
        [2, 3, 4, 5]
    );

davidchambers: Marvellous!

kedashoe:
Sure thing :)

kevinbeaty:
> some of us were starting to wonder whether anything tangible would come of the "ramduce" project. ;)
Hehe.  I was on of those people ;)

Thanks @kedashoe for all your input (dating back to November?).   I think #878 is a subset of some of the things you were trying to accomplish here.  Looking forward to what you come up with.

CrossEye:
> I'm very grateful to you for opening this pull request, @kedashoe.

:+100:

CrossEye:
> I'm very grateful to you for opening this pull request, @kedashoe.

+100

Issue #859, opened 2015-03-02
remove remaining aliases


davidchambers:
Ramda now has just two :alien:s:

  - `R.I` is an :alien: for `R.identity`; and
  - `R.get` is an :alien: for `R.prop`.

I would love to be able to say _Ramda has no aliases_. I don't mind which of the four combinations we keep.

TheLudd:
I certainly prefer `identity`  to `I`
I also think I prefer `prop` to `get` but that is less important for me.

CrossEye:
I'm with @TheLudd on both, although there's a twinge on giving up `I`.

buzzdecafe:
<3 `I`

jethrolarson:
`R.T` and `R.F` are the only single char names now. Are they sticking around?

davidchambers:
> Are they sticking around?

Yes. `R.T` is useful in conjunction with `R.cond`. :)

jethrolarson:
But what about R.truth and R.phase? :P

TheLudd:
@jethrolarson or `R.absolutelyRight` and `R.notInAnyWayCorrect`? ;)

buzzdecafe:
i think we'll go with `R.indubitably` and `R.certainlyNot`

jethrolarson:
+1 for R.indubitably

buzzdecafe:
in that case let's got with `R.indubitably` and `R.dubitably`

jethrolarson:
But literally dubitably means doubtful, so R.dubitably == Maybe(false)

PR #867, opened 2015-03-03
remove R.get

davidchambers:
See #859


buzzdecafe:
this one may be back, depending on what sort of dispatching we want to support

PR #871, opened 2015-03-03
deprecate R.get

davidchambers:
Related to #867

### DIFF
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -4201,6 +4201,7 @@
     var functionsIn = _curry1(_functionsWith(keysIn));
 
     /**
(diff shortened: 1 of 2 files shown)


buzzdecafe:
should we list what version we expect it to be removed in?

davidchambers:
> should we list what version we expect it to be removed in?

I'd prefer not to be tied to a particular version.

I think deprecated functions should only stick around for one minor version. It's important that we continue to progress towards 1.0.0.

buzzdecafe:
:cow2:

CrossEye:
:herb:

Issue #729, opened 2015-01-16
Add anyIndexed function


honzabrecka:
Hi, I'm trying to implement something like this:

```
R.some.idx(function(value, i, list) { return value != R.nth((i + 1) % R.length(list), list); });
```

But Ramda doesn't provide indexed `some` or `any`. Wouldn't it be worth to add `anyIndexed` function?

buzzdecafe:
I don't understand what this predicate proves: 

`function(value, i, list) { return value != R.nth(i, list); }`

as i read this, it says, "`true`, when the value at index `i` is not equal to the value at index `i`". Am I missing something?

we've been over the indexing thing before. I think what we would all like is a way to decorate iterators to take extra args, but default to the single arg case. What do you propose for `anyIndexed`? What is the API/signature?

CrossEye:
We had a long, fairly rancorous discussion about this issue in #452.

There is a proposal, which I'm hoping to revive for version 0.10.0 in #484 to remove the `.idx` functions but allow you to wrap `reduce`, `map`, `filter`, `some`, etc like this:

    addIdx(map)(function(val, idx, all) { /* ... */ }, list);

But I don't understand what you're trying to do in that code block.  Can explain in more detail?

honzabrecka:
Sorry guys, I made a mistake in my example, now it's ok. I would expect that `index` and `list` are extra params.

CrossEye:
This will probably be available when we get the `addIdx` feature in place.  In the meantime, if I understand the code block above, all you're trying to do really is return `true` if there are two or more distinct values in the input list, `false` if the list contains all the same value.  If that's all you want, this would probably do the same:

    pipe(uniq, length, gt(R.__, 1))

There are many other possibilities.

When I have a few minutes, I'll try to explain my philosophical objection to these `idx` functions in general.

CrossEye:
I wanted to discuss my reasons for objecting to the indexed versions of functions like `map`, `filter`, `reduce`, etc.  We had a long discussion about adding them to `some` and `any` back in #452, and a pass at how to replace them with a common wrapper function in #484.  But I never really articulated one of my major reasons for not liking them.

[...]

Issue #484, opened 2014-10-27
Replace all the .idx functions with a common addIdx


CrossEye:
I'm far from certain about this, but based on some discussions recently in #452.  (It's a long thread, you can start with comments from [2014-10-27][ad]) I was thinking about replacing all the controversial `xxx.idx` functions with a single wrapper for callback functions:

```javascript
    /**
     * Creates a wrapper for an iteration callback, such as one that might be
     * supplied to `map`, or `reduce`.  This wrapper supplies the index as well
     * as the standard parameters usually passed.  If the optional `list`
     * parameter is supplied to this function, it also passes that into each
     * iteration.
     *
     * @func
     * @memberOf R
     * @category Function
     * @sig (a -> b -> ... -> n) -> (a -> b -> ... Num -> n)
     * @sig (a -> b -> ... -> n) -> (a -> b -> ... Num -> {*} -> n)
     * @param {Function} fn A list iteration callback function.
     * @param {List} [list] The list that will be iterated
     * @return {Function} A new function that supplies the index (and optionally
     *         the entire list) as well as the existing parameters.
     * @example
     *
     *      var primes = [2, 3, 5, 7, 11];
     *
     *      var convert1 = function(prime, idx) {return prime + "-" + idx;};
     *      map(addIdx(convert1), primes);
     *      //=> ["2-0", "3-1", "5-2", "7-3", "11-4"]
     *
     *      var convert2 = function(prime, idx, list) {
     *          return prime + "-" + idx + "/" + list.length;
     *      };
     *      map(addIdx(convert2, primes), primes)
     *      //=> ["2-0/5", "3-1/5", "5-2/5", "7-3/5", "11-4/5"]
     *
     *      var convert3 = function(soFar, prime, idx) {
     *          return soFar + prime / (idx + 1);
     *      };
     *      reduce(addIdx(convert3), 0, primes);
     *      //=> 2/1 + 3/2 + 5/3 + 7/4 + 11/5 = 2735/60 ~= 9.116666666666667
     */
    R.addIdx = function(fn, list) {
        var idx = 0;
        return function() {
            return fn.apply(this, concat(
                _slice(arguments),
                list ? [idx++, list] : idx++
            ));
        }
    };
```

I haven't tested deeply, but it works for the examples in the comments.  The nicest bit is that most of the time people want these, they just want the index and not the list, and then the wrapper is fairly simple, just a wrap around their function reference or their lambda.

[...]

Issue #452, opened 2014-10-17
callback args


jdalton:
Not sure if this has been brought up but many of your methods like `map` and `some` only provide the `value` to their callbacks instead of the traditional `value, index, object`. Other functional currying-by-default libs, like [FKit](https://github.com/nullobject/fkit), do pass the traditionally expected arguments.

CrossEye:
Above we allow @jdalton to have an open-ended say about his suggestion that Ramda change its API in a way that would make it more like Lo-Dash's.  There are about twenty comments from him above, and the conversation was allowed to continue as long as participants cared about it, even though it grew rather acrimonious; a vigorous exchange of ideas is important to us.

But when I [make the inverse suggestion][ld844] on the Lo-Dash list, that its API get expanded in a way more consistent with Ramda's, he locked the issue, preventing further comment after only my fourth comment on the thread.

What gives?  Why would someone work so hard to promote changing someone else's library, but care so little about criticism of his own that he cuts off debate so soon?

For the record, below was what I was planning on as my final comment on that thread.

  [ld844]: https://github.com/lodash/lodash/issues/844 

----------

>> With the radical changes that `_.rearg` will allow, you still don't offer this simple functionality that other libraries and languages do.

> Sure it does. Devs are free to use `_.ary` and `_.rearg` on any function, Lo-Dash or otherwise. We provide `_.runInContext` and `_.mixin` to allow creating and extending lodash instances. The functionality is simple enough.

It's the difference between

    var res1 = _.mapNoIdx(coll_1, thirdPartyFn);
    // ...
    var res2 = _.mapNoIdx(coll_2, thirdPartyFn);

and 

    var wrappedThirdPartyFn = _.ary(thirdPartyFn, 1);
    // ...
    var res1 = _.map(coll_1, wrappedThirdPartyFn );
    // ...
    var res2 = _.map(coll_2, wrappedThirdPartyFn);

To me, the former is substantially simpler.  And even if you forgo the named function and only wrap the function anonymously, this makes you write code that tweaks the library rather than code that expresses your intent.  To my mind, higher-level libraries should not be doing this.

> Added complexity is minimal with no impact on performance of the common case and low risk due to its narrow check. The result is happy, not annoyed devs.

I'll take your word on performance.  As I've said before, that's not one of my strong suits.  But I am strongly concerned about API design.  And the above smacks of a lazy design.  

>> Moreover, how do you respond to the reasonable request that map extend to other types than array, string, and object?

[...]

PR #872, opened 2015-03-03
Add allIndexed

jeduden:
Currently this commit increases code duplication:
- Unittests have been copied and modified from "all" since the unittests from "all" should also apply to allIndexed.

### DIFF
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -588,6 +588,44 @@
     var all = _curry2(_all);
 
     /**
+     * Like `all`, but but passes additional parameters to the predicate function.
+     *
+     * `fn` receives three arguments: *(value, index, list)*.
+     * Returns `true` if all elements of the list match the predicate, `false` if there are any
+     * that don't.
+     *
+     * @func
+     * @memberOf R
+     * @category List
+     * @sig (a, i, [a] -> Boolean) -> [a] -> Boolean
+     * @param {Function} fn The predicate function.
+     * @param {Array} list The array to consider.
+     * @return {Boolean} `true` if the predicate is satisfied by every element, `false`
+     *         otherwise.
+     * @example
+     *
+     *      var firstIndexOf5ThatIsNotFollowingA4;
+     *
+     *      R.allIndexed(function(num, idx, list) {
+     *        if( num === 5 && list[idx-1] != 4 ) {
+     *          firstIndexOf5ThatIsNotFollowingA4 = idx;
+     *          return false; // use short circuit to stop here.
+     *        } else {
+     *          return true;
+     *        }
+     *      },[4, 5, 1, 2, 5, 6, 5]);
+     */
+    var allIndexed = _curry2(function _all(fn, list) {
+        var idx = -1;
+        while (++idx < list.length) {
+            if (!fn(list[idx], idx, list)) {
+                return false;
+            }
+        }
+        return true;
+    });
+
+    /**
      * Returns a function that always returns the given value.
      *
      * @func
@@ -6163,6 +6201,7 @@
         __: __,
         add: add,
         all: all,
+        allIndexed: allIndexed,
         allPass: allPass,
         always: always,
         and: and,
--- a/src/allIndexed.js
+++ b/src/allIndexed.js
@@ -1,0 +1,40 @@
+var _curry2 = require('./internal/_curry2');
+
+/**
+ * Like `all`, but but passes additional parameters to the predicate function.
+ *
+ * `fn` receives three arguments: *(value, index, list)*.
+ * Returns `true` if all elements of the list match the predicate, `false` if there are any
+ * that don't.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig (a, i, [a] -> Boolean) -> [a] -> Boolean
+ * @param {Function} fn The predicate function.
+ * @param {Array} list The array to consider.
+ * @return {Boolean} `true` if the predicate is satisfied by every element, `false`
+ *         otherwise.
+ * @example
+ *
+ *      var firstIndexOf5ThatIsNotFollowingA4;
+ *
+ *      R.allIndexed(function(num, idx, list) {
+ *        if( num === 5 && list[idx-1] != 4 ) {
+ *          firstIndexOf5ThatIsNotFollowingA4 = idx;
+ *          return false; // use short circuit to stop here.
+ *        } else {
+ *          return true;
+ *        }
+ *      },[4, 5, 1, 2, 5, 6, 5]);
+ */
+module.exports = _curry2(function _all(fn, list) {
+    var idx = -1;
+    while (++idx < list.length) {
+        if (!fn(list[idx], idx, list)) {
+            return false;
+        }
+    }
+    return true;
+});
+
--- a/test/allIndexed.js
+++ b/test/allIndexed.js
@@ -1,0 +1,59 @@
+var assert = require('assert');
+
+var R = require('..');
+
+
+describe('allIndexed', function() {
+    var even = function(n) {return n % 2 === 0;};
+    var T = function() {return true;};
+
+    it('predicate has access to index', function() {
+        var indices = [];
+        var recordIndex = function(n, index) { indices.push(index);return true; };
+        R.allIndexed(recordIndex, [2, 4, 6, 8, 10, 12]);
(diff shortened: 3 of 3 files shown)


CrossEye:
I'm not overly worried about duplication in the test suite; although obviously, even there, DRYer is better.

But I'd like to know more about use-cases for this.  The example in the inline docs doesn't convince me, and there's no real examples in the test cases.

What real problems does it solve?

We've resisted including similar functions before because we could see little need.  See a very long discussion in #452 for example.

I really need to get back to considering #484, which might make this easier to do.

Can you give a convincing argument for this function?

jeduden:
Sure.
General arguments
1. This is a step towards providing indexed versions for all functions that iterate over an array like object. 
2. This makes the usage of the library easier in the sense that you dont need to rethink your algorithms if the one that you got uses the index.
3. if you look at the code of foreach and its siblings , index and list are available.
4. ecmascript versions provide these.
5. If you need the index and ramda does not provide it you might have to add code to recreate it.  Worst case developers add an index variable outside of the callback function and increase it each time. 

Use case based arguments
6. Any algorithm that can be implemented with forEach but that would benefit from the short circuit behavior of all.
7. Find index of unescaped quote. See example in testcase.
8. find index of first pair that where less is true. Used in sorting algos.

buzzdecafe:
I don't see why point 1 should be a goal; point 2 is reasonable, although I have yet to see a use case for index with `all` apart from pairwise processessing. As to 3, I would love to *remove* `forEach` from the library! But it is almost inescapable if you want to work with the DOM. 4 is not a concern for us, I don't think we are trying to recreate native interfaces.

point 5 is a little persuasive, but I am not particularly moved by 6 and 7 which are better implemented with `findIndex`. As noted above, the only use case I've seen that may be simplified with an index to `all` is when processing elements of an array in a pairwise fashion, So far, that has not been an issue.

CrossEye:
I'm not going to have much time before the weekend, but I will take another look then at #484.  I think that would solve this in a generic fashion and still reduce the API footprint.

CrossEye:
Part of my problem is that this code feels very much outside the spirit of Ramda:

    var firstIndexOf5ThatIsNotFollowingA4;
    R.allIndexed(function(num, idx, list) {
      if( num === 5 && list[idx-1] != 4 ) {
        firstIndexOf5ThatIsNotFollowingA4 = idx;
        return false; // use short circuit to stop here.
      } else {
        return true;
      }
    },[4, 5, 1, 2, 5, 6, 5]);

Creating a side-effect inside a call to `allIndexed`, and making that an example of how one would want to use the library is definitely on the wrong track to me.

On the other hand, dealing with pairs of elements (or triples, etc.) from a list seems a very useful possibility, and I could easily see something like:

    R.pairwise(['a', 'b', 'c', 'd']); //=> [['a', 'b'], ['b', 'c'], ['c', 'd']]

which would probably be a gloss on `nWise` or some such:

    R.pairwise = R.nWise(2);

so that it could create tuples of length `n` as necessary.

This still wouldn't solve your whole issue as you need back the index, but together with `addIndex` from #484, I think you could write something simpler for a case like this.

jeduden:
In general i like the idea of pairwise/nwise. And i agree that together
with addIdx it solves my current use cases.

Thinking about pairwise and also nwise, i realized that they are (likely
common) special cases of a  moving window function , that maps an index to
an array of indices, combined with a mapping function, that performs the
element lookup for each of the indices.

One aspect that i am looking a good solution for is the aspect of short
circuiting.
A function like findIndex would support these special cases.

More generally, reduce would also work but currently reduce doesnt offer
short circuiting. (Also es6 dont offer it ).

An extended version of reduce that behaves like the following might be
worth consideration:

Reduce( callback, initial )
Where callback is a function(agg, element, (index, list ), stop )
stop is a function that signals to reduce to stop iterating after the
callback.completes.
Alternative to the stop callback would be to return not simply the return
value but a pair. The result of reduce and a boolean that signals when to
stop.

CrossEye:
`findIndex` is already a concession that doesn't thrill me.  In the abstract, the data type that Ramda works with is _lists_, and lists don't really have indices.  Features like maps, filters, and folds operate on more general data structures as well.  In fact, Javascript doesn't have a `List` type, so we use the closest equivalent, the (non-sparse) array.  But we do try to keep in mind that underneath we want to think about these structures as lists, and we would like one day to ensure our functions work on related data structures that may not have a notion of an index, something like streams, for instance, or trees.  I discuss this at greater length in https://github.com/ramda/ramda/issues/729#issuecomment-70351882.

`findIndex` doesn't well expand in this direction, and so I'm less than thrilled to have it.  I recognize that there are plenty of times its useful, and I'm not looking to get rid of it, but it strikes me as further from the core of Ramda than, say, `find`.

-----

I don't see a reason for `reduce` to have a stop-iteration.  The idea of `reduce` (`fold` in many functional languages) is to fold a data structure down to a value.  That you could stop early is not such a big deal in that circumstance.  Obviously there are other functions that do have a reason to stop iteration, ones like `find`, or short-circuiting ones like `any*` or `all*`.  But I would do those where needed, because in general, operating on a data structure should be thought of as in some way operating on a whole.

jeduden:
> But I would do those where needed, because in general, operating on a
data structure should be thought of as in some way operating on a whole.
It is fine to declare ramda as a library that supports functional
programming for algorithms and data structures that focus on lists/sets
only and that short circuiting and random access data structures and
algorithms are not a focus for Ramda.

While I personally believe that random access and short circuiting have a
place in a pure functional library, it can be very well that Ramda is not
that specific library
If that is the case, it would be helpful for managing expectations about
the library's focus if you could make this more explicit in the mission
statement.

buzzdecafe:
sure, clojure shourt-circuits reduce when it sees

jeduden:
Clojures way to mark values as reduced is clearly an additional alternative provide short circuit functionality to the application.
One thing I am not clear on, if this kind of extension to Ramda's reduce is aligned with the goals of the library and that if I send a pull request that implements one of these approaches has a chance of being accepted.

CrossEye:
I recently published an article on my ideas about the [philosophy of Ramda][pr]; one of the points I stress is that we want to keep Ramda's API shallow and simple.  Functions in Ramda don't do a lot of "well if the parameter is a String, do this, an object do that, or a function do the other."  We don't offer optional parameters anywhere.

I can't imagine a way to update `reduce` with a stop-iteration that didn't run afoul those notions of simplicity.  That doesn't mean there couldn't be another distinct folding function that did have it, if you're interested in presenting one.  Since I haven't personally found much need for one, I might take some convincing to see the benefits.  It would have to demonstrate some usefulness that can't be simply achieved through the current short-circuiting functions.

But @buzzdecafe or @davidchambers might have different feedback.


  [pr]: http://fr.umio.us/the-philosophy-of-ramda/

kedashoe:
Short circuiting `reduce` is necessary for transducers, so if you guys go that way users will ipso facto gain this flexibility.

jeduden:
Yeah. I looked into transducers more indepth yesterday. They seem to
address a lot of my issues.

Transducer-js uses closures approach to shortcicuiting (i.e. wrapping the
return value ). Is that an approach that resonates with the people here ?

CrossEye:
@kedashoe: Does this imply that users of `reduce` will have to change their code?

jeduden:
@CrossEye : No. If wrapping is used, the change is transparent to any implementation that care about short circuiting. 
Both transducers-js and transducer.js use actually the same technique.
see : https://github.com/cognitect-labs/transducers-js/blob/master/src/com/cognitect/transducers.js#L130

kedashoe:
@CrossEye users should be fine, the PR passes the current reduce test suite, I also have a test or two for `reduce` in test/transduce.js. The reducing loop simply checks if the user's function returns `reduced`, eg

https://github.com/kedashoe/ramda/blob/f28db55ed88b541ae3c466f14167c1a8816daf6d/src/internal/_reduce.js#L22

This reminds me, I didn't mention in the PR thread but reduce itself will also be able to handle all data types (objects, Immutable classes, streams, etc)

jeduden:
Looking at https://github.com/ramda/ramda/pull/865 it looks like that it would help my use cases quiet a bit: since provides short circuiting and transducers which themselves somewhat help in my use cases.

buzzdecafe:
see also #878

davidchambers:
If someone can demonstrate how #878 addresses this need we can close this pull request.

PR #873, opened 2015-03-04
adds description of expected function result

donnut:
I propose a more explicit description of the expected return value of the iterator function.

### DIFF
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -1934,7 +1934,8 @@
      * function to each element of a list, passing an accumulating parameter from left to
      * right, and returning a final value of this accumulator together with the new list.
      *
-     * The iterator function receives two values: *(acc, value)*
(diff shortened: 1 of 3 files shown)


buzzdecafe:
:+1: 

i believe the build is failing 'cuz we still require the built artifact to be checked in

donnut:
Ah, that explains the error. Should I do something with this?

buzzdecafe:
i think you can `git add dist/ramda.js`, then `git commit --amend` and then push.

donnut:
Now I've ended up with two commits... Still looking for a 12 weeks MOOC on git basics, or just take time to RTFM.

buzzdecafe:
i feel your pain. can you help out @davidchambers ? alternately you can just create a new branch and push a new PR from there. that may be simplest

donnut:
I'll try to create a new branch.

PR #874, opened 2015-03-04
description iterator function mapAccum*

donnut:
makes PR branch `doc_mapaccum` obsolete

### DIFF
--- a/src/mapAccum.js
+++ b/src/mapAccum.js
@@ -6,7 +6,8 @@ var _curry3 = require('./internal/_curry3');
  * function to each element of a list, passing an accumulating parameter from left to
  * right, and returning a final value of this accumulator together with the new list.
  *
- * The iterator function receives two values: *(acc, value)*
+ * The iterator function receives two arguments: *acc* and *value*, and should return
+ * a tuple *[acc, value]*.
  *
  * @func
  * @memberOf R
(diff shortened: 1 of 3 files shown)


buzzdecafe:
LGTM :cow2:

src/mapAccum.js
```diff
@@ -6,7 +6,8 @@ var _curry3 = require('./internal/_curry3');
  * function to each element of a list, passing an accumulating parameter from left to
  * right, and returning a final value of this accumulator together with the new list.
  *
- * The iterator function receives two values: *(acc, value)*
+ * The iterator function receives two arguments: *acc* and *value*, and should return
+ * a tuple *[acc, value]*.
```
davidchambers: Since `,` binds more tightly than `:`, I parse this as:

    (The iterator function receives two arguments): ((*acc* and *value*), and should return a tuple *[acc, value]*).

We should replace the `:` with a `,` to ensure correct parsing:

    The iterator function receives two arguments, (*acc* and *value*), and should return a tuple *[acc, value]*.

benperez: I think an [em-dash](http://www.thepunctuationguide.com/em-dash.html) actually reads better here.

The iterator function receives two arguments—(*acc* and *value*)—and should return a tuple *[acc, value]*.

davidchambers: :+1: for a pair of em dashes. +:100: for your attention to detail.

CrossEye: I'm fine with any of them.

davidchambers:
Resolved via cf34e3582df4662150afd1ecbf197400bcea63cb

PR #876, opened 2015-03-05
change path functions to operate on lists of strings

### LINKED ISSUE #637: getIn (prop keypaths)

davidchambers:
Closes #637

This pull request introduces four new internal functions: `_assoc`, `_assocPath`, `_dissoc`, and `_dissocPath`. These are defined with as few dependencies as possible. `_assoc`, for example, has no dependencies whereas the existing `assoc` implementation depends on `_extend`, `_map`, `createMapEntry`, `fromPairs`, and `keysIn`!

Most significantly, this pull request changes the type of the `path` parameter from (dot-separated) `String` to `[String]`, and updates the test suite accordingly.


src/internal/_assocPath.js
```diff
@@ -0,0 +1,11 @@
+var _assoc = require('./_assoc');
+var _slice = require('./_slice');
+
+
+module.exports = function _assocPath(path, val, obj) {
```
davidchambers: What should happen if `path` is `[]`?

src/internal/_dissocPath.js
```diff
@@ -0,0 +1,16 @@
+var _assoc = require('./_assoc');
+var _dissoc = require('./_dissoc');
+var _slice = require('./_slice');
+
+
+module.exports = function _dissocPath(path, obj) {
```
davidchambers: What should happen if `path` is `[]`?

buzzdecafe:
we should also update `path*` to have the same api

src/internal/_assocPath.js
buzzdecafe: return a clone of `obj`

src/internal/_dissocPath.js
buzzdecafe: return a clone of `obj`

CrossEye:
Is this the first internal recursion in Ramda?  We've been pretty gun-shy about it after Eweda's stack problems.  I don't mind here.  The paths should never be long enough to worry us, but we need to note it and be wary.

This should close another issue that's been open for a while, but I'm too tired to try to find it.

src/internal/_assocPath.js
CrossEye: I'm thinking a no-op, just returning the obj supplied.

src/internal/_dissocPath.js
CrossEye: Again, I think a no-op is appropriate.

src/internal/_assocPath.js
davidchambers: It's a question of philosophy. One can make an argument for either approach. One could argue that `_assocPath` (and by extension `assocPath`) should *always* return an object to which no one holds a reference. On the other hand one could argue that since Ramda promotes programming without mutation, cloning is unnecessary.

davidchambers: We should also consider the current [behaviour][1] of `R.path`:

> ```javascript
> assert.strictEqual(R.path('', obj), undefined);
> ```


[1]: https://github.com/ramda/ramda/blob/v0.10.0/test/path.js#L25

davidchambers:
> we should also update `path*` to have the same api

Good catch.

`R.pathOn` doesn't make sense now that we're specifying path as a list. Shall we remove it?

test/path.js
```diff
@@ -21,34 +21,34 @@ describe('path', function() {
...
             j: ['J']
         };
-        assert.strictEqual(R.path('a.b.c', obj), 100);
-        assert.strictEqual(R.path('', obj), undefined);
-        assert.strictEqual(R.path('a.e.f.1', obj), 101);
-        assert.strictEqual(R.path('j.0', obj), 'J');
-        assert.strictEqual(R.path('j.1', obj), undefined);
-        assert.strictEqual(R.path('a.b.c', null), undefined);
+        assert.strictEqual(R.path(['a', 'b', 'c'], obj), 100);
+//      assert.strictEqual(R.path([], obj), undefined);
```
davidchambers: We should agree on how this should evaluate.

buzzdecafe:
> R.pathOn doesn't make sense now that we're specifying path as a list. Shall we remove it?

yes. do we need to deprecate first?

davidchambers:
> do we need to deprecate first?

I don't think so. We're making breaking changes to `R.path`, `R.pathEq`, `R.assocPath`, and `R.dissocPath`. I don't see much value in preserving the old behaviour of `R.pathOn` for one release. Furthermore I'm against having the old `R.pathOn` function in a release alongside the new path functions.

buzzdecafe:
agreed. this change has been in the air for a while anyways

src/internal/_assocPath.js
buzzdecafe: `undefined` is what you get for `path('x', {y: 1})`, yes? Seems like the same situation, analogous to `find`. Another _unsafe_ function, to be sure

test/path.js
buzzdecafe: seems to me to be the same case as this: `assert.strictEqual(R.path(['j', '1'], obj), undefined);`

TheLudd:
Just curious, why is this preferred?

buzzdecafe:
> Just curious, why is this preferred?

e.g. imagine you are modeling a filesystem as an object, you could have keys with '.' in them.

src/internal/_assocPath.js
CrossEye: > On the other hand one could argue that since Ramda promotes programming without mutation, cloning is unnecessary.

That's the argument I would make.  I don't see a reason to treat the top-level object any different from the sub-object properties, which are references.

buzzdecafe: noop is fine

buzzdecafe:
:cow2:

src/internal/_assocPath.js
davidchambers: Updated.

src/internal/_dissocPath.js
davidchambers: Updated.

src/internal/_path.js
```diff
@@ -14,11 +14,13 @@
  *      _path(['a', 'b'], {a: {b: 2}}); //=> 2
  */
 module.exports = function _path(paths, obj) {
-    var idx = -1, length = paths.length, val;
-    if (obj == null) { return; }
-    val = obj;
-    while (val != null && ++idx < length) {
-        val = val[paths[idx]];
+    if (obj == null || paths.length === 0) {
```
davidchambers: I don't love the `obj == null` guard, but removing it is a matter for a separate pull request.

test/pathEq.js
```diff
@@ -13,18 +13,23 @@ describe('pathEq', function() {
...
     it('returns false for non existing values', function() {
-        assert.ok(!R.pathEq('c', 'foo', obj));
-        assert.ok(!R.pathEq('c.d', 'foo', obj));
+        assert.strictEqual(R.pathEq(['c'], 'foo', obj), false);
+        assert.strictEqual(R.pathEq(['c', 'd'], 'foo', obj), false);
+    });
+
+    it('accepts empty path', function() {
+        assert.strictEqual(R.pathEq([], 42, {a: 1, b: 2}), false);
+        assert.strictEqual(R.pathEq([], undefined, {a: 1, b: 2}), true);
```
davidchambers: *Should* this evaluate true? Should `R.pathEq(['x'], undefined, {})`?

test/path.js
davidchambers: Updated.

TheLudd:
@buzzdecafe 

> e.g. imagine you are modeling a filesystem as an object, you could have keys with '.' in them.

good point

src/internal/_pickBy.js
```diff
@@ -1,23 +0,0 @@
...
-    var len = props.length;
-    var idx = -1;
-    while (++idx < len) {
-        prop = props[idx];
-        if (test(obj[prop], prop, obj)) {
-            copy[prop] = obj[prop];
-        }
-    }
-    return copy;
-};
```
davidchambers: `_pickBy` is no longer used (except in `pickBy`).

davidchambers:
This pull request is ready for a final review.

src/internal/_assocPath.js
```diff
@@ -0,0 +1,14 @@
...
+    switch (path.length) {
+        case 0:
+            return obj;
+        case 1:
+            return _assoc(path[0], val, obj);
+        default:
+            return _assoc(path[0], _assocPath(_slice(path, 1), val, Object(obj[path[0]])), obj);
```
buzzdecafe: :smiley_cat: 

elegant

test/pathEq.js
buzzdecafe: i think so. Same as `pathEq(['x', 'y'], undefined, {x: {z: 100}}) //=> true`, yes?

buzzdecafe:
:cow2:

test/pathEq.js
davidchambers: It depends on whether one thinks the function should *always* return false if the path does not exist.

```javascript
R.pathEq(['x', 'y'], undefined, {x: {y: undefined}});  // => true, indisputably
R.pathEq(['x', 'y'], undefined, {});  // => true? false?
```

buzzdecafe: As i read it, the question you are asking `pathEq(['x', 'y'], undefined, {})` is: "Is it true that `{}.x.y` is undefined?" The answer is "Yes, it is true."

witoldsz:
It's too bad you did not let users using a path as a string to maintain backward compatibility and for simplicity (Ramda could internally split it on "dot" if an argumet was a string). 

Sometimes it's so much easier and shorter to be able to specify "a.b.c.d" instead of awkward array notation (which is in itself useful some other times).

CrossEye:
It's actually for reasons of simplicity (as opposed to those of [ease][rh]) that we choose for our API not to do such different things on different input types. 

But I'm with you on how nice the string-based API was compared to the array-based one.  But I couldn't argue against the additional flexibility of the current approach.  Luckily, it should be easy to write your own version on top of the existing one. 


witoldsz:
I wish I had convinced you to make an exception, for the sake of "*a practical*" adjective describing the library at the front page :)
Thanks for the *Simple Made Easy* link.

CrossEye:
> I wish I had convinced you to make an exception, for the sake of "a practical" adjective describing the library at the front page

I discuss the notion of what we mean by "practical" in [The Philosophy of Ramda][pr] and further in https://github.com/ramda/ramda/issues/1002#issuecomment-89614280 .  I think there's a good reason not to do what you're suggesting, even though I initially argued for it as well, and even though we did it that way at first.

A number of our functions are dynamically dispatched, and we internally supply the version that works for lists or for plain objects.  And we offer tranduced versions of many of our functions.  But other than that,  we don't vary our behavior based on argument type.  I think that's a pretty clean line to draw.  But creating a version that works the way you want should be as simple as:

    var path = R.useWith(R.path, R.split('.'), R.identity);
    path('a.b.c', {a: {b: {c: 'foo', d: 'bar'}}}); //=> 'foo'


  [pr]: http://fr.umio.us/the-philosophy-of-ramda/

PR #878, opened 2015-03-06
Initial transducers implementation

kevinbeaty:
- Squashed from buzzedecafe:transduce
- Rebased to master
- Discussed in ramduce
- Credits to @buzzdecafe and @kedashoe

Squashed Commits:
- we have a map transducer
- Apply all arguments to transducer as function.
- Use switch case dispatch for transducer function
- port @kedashoe and @kevinbeaty changes
- Fix build and tests.
- _appendXf as single expression and fix require order for bulid
- foldl uses _xwrap for use with functions
- _dispatchable accepts function for use in _groupBy
- _isIterable uses _symIterator
- Fix filter/map to use _dispatchable appropriately
- Curry _xany, _xfilter, _xmap for use with _dispatchable
- Fix idx lint warning in _arrayReduce
- Remove no longer supported transformer dispatch test
- Transducers for take, takeWhile, drop, dropWhile
- get tests working again
- transduce tests
- improve map test coverage
- Add into
- all and any working
- Add find, findLast, findIndex and findLastIndex transducers
- Step default value in any/all
- stepLast no longer requires initial value since transducers always step single value
- Allow null/undefined accumulators when checking reduced
- groupBy transducer

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,6 +50,6 @@ test: dist/ramda.js
 	| sed 's:src/\(.*\)[.]js:exports.\1 = require("./src/\1");:' >index.js
 	sed '/"main":/d' package.json >tmp
 	mv tmp package.json
-	$(MOCHA) -- $(shell find test -name '*.js' -not -name 'installTo.js' -not -name 'test.examplesRunner.js')
+	$(MOCHA) -- $(shell find test -name '*.js' -not -name 'installTo.js' -not -name 'test.examplesRunner.js' -not -path 'test/helpers/*')
 	git checkout -- package.json
 	rm index.js
--- a/src/dropWhile.js
+++ b/src/dropWhile.js
@@ -1,11 +1,16 @@
 var _curry2 = require('./internal/_curry2');
+var _dispatchable = require('./internal/_dispatchable');
 var _slice = require('./internal/_slice');
+var _xdropWhile = require('./internal/_xdropWhile');
 
 
 /**
  * Returns a new list containing the last `n` elements of a given list, passing each value
  * to the supplied predicate function, skipping elements while the predicate function returns
  * `true`. The predicate function is passed one argument: *(value)*.
+ *
+ * Acts as a transducer if a transformer is given in list position.
+ * @see R.transduce
  *
  * @func
  * @memberOf R
@@ -22,8 +27,8 @@ var _slice = require('./internal/_slice');
  *
  *      R.dropWhile(lteTwo, [1, 2, 3, 4]); //=> [3, 4]
  */
-module.exports = _curry2(function dropWhile(pred, list) {
+module.exports = _curry2(_dispatchable('dropWhile', _xdropWhile, function dropWhile(pred, list) {
     var idx = -1, len = list.length;
     while (++idx < len && pred(list[idx])) {}
     return _slice(list, idx);
-});
+}));
--- a/src/internal/_isIterable.js
+++ b/src/internal/_isIterable.js
@@ -1,0 +1,6 @@
+var _symIterator = require('./_symIterator');
+
+
+module.exports = function _isIterable(x) {
+    return (x[_symIterator] != null) || (typeof x.next === 'function');
+};
--- a/src/internal/_isTransformer.js
+++ b/src/internal/_isTransformer.js
@@ -1,0 +1,7 @@
+var _symTransformer = require('./_symTransformer');
+
+
+module.exports = function _isTransformer(obj) {
+    return (obj[_symTransformer] != null) ||
+        (typeof obj.step === 'function' && typeof obj.result === 'function');
+};
--- a/src/internal/_reduce.js
+++ b/src/internal/_reduce.js
@@ -1,7 +1,19 @@
+var _arrayReduce = require('./_arrayReduce');
+var _isIterable = require('./_isIterable');
+var _iterableReduce = require('./_iterableReduce');
+var _xwrap = require('./_xwrap');
+var isArrayLike = require('../isArrayLike');
+
+
 module.exports = function _reduce(fn, acc, list) {
-    var idx = -1, len = list.length;
-    while (++idx < len) {
-        acc = fn(acc, list[idx]);
+    if (typeof fn === 'function') {
+        fn = _xwrap(fn);
     }
-    return acc;
+    if (isArrayLike(list)) {
+        return _arrayReduce(fn, acc, list);
+    }
+    if (_isIterable(list)) {
+        return _iterableReduce(fn, acc, list);
+    }
+    throw new TypeError('reduce: list must be array or iterable');
 };
--- a/src/internal/_stepCat.js
+++ b/src/internal/_stepCat.js
@@ -1,0 +1,54 @@
+var _add = require('./_add');
+var _appendTo = require('./_appendTo');
+var _isTransformer = require('./_isTransformer');
+var _symTransformer = require('./_symTransformer');
+var isArrayLike = require('../isArrayLike');
+var merge = require('../merge');
+
+
+module.exports = (function() {
+  var _result = function(result) { return result; };
+  var _stepCatArray = {
+       init: Array,
+       step: _appendTo,
+       result: _result
+  };
+  var _stepCatString = {
+      init: String,
+      step: _add,
+      result: _result
+  };
+  var _stepCatObject = {
+      init: Object,
+      step: function(result, input) {
+          var key;
+          var vals;
+          var stepObj = {};
+          if (isArrayLike(input)) {
+              key = input[0];
+              vals = input[1];
+              stepObj[key] = vals;
+          } else {
+              stepObj = input;
+          }
+          return merge(result, stepObj);
+      },
+      result: _result
+  };
+
+  return function _stepCat(obj) {
+      if (_isTransformer(obj)) {
+          return obj[_symTransformer] || obj;
+      }
+      if (isArrayLike(obj)) {
+          return _stepCatArray;
+      }
+      if (typeof obj === 'string') {
+          return _stepCatString;
+      }
+      if (typeof obj === 'object') {
+          return _stepCatObject;
+      }
+      throw new Error('Cannot create transformer for ' + obj);
+  };
+})();
--- a/src/internal/_symIterator.js
+++ b/src/internal/_symIterator.js
@@ -1,0 +1,1 @@
+module.exports = (typeof Symbol !== 'undefined') ? Symbol.iterator : '@@iterator';
--- a/src/internal/_xall.js
+++ b/src/internal/_xall.js
@@ -1,0 +1,29 @@
+var _curry2 = require('./_curry2');
+var _reduced = require('./_reduced');
+
+
+module.exports = (function() {
+    function XAll(f, xf) {
+        this.xf = xf;
+        this.f = f;
+        this.all = true;
+    }
+    XAll.prototype.init = function() {
+        return this.xf.init();
+    };
+    XAll.prototype.result = function(result) {
+        if (this.all) {
+          result = this.xf.step(result, true);
+        }
+        return this.xf.result(result);
+    };
+    XAll.prototype.step = function(result, input) {
+        if (!this.f(input)) {
+            this.all = false;
+            result = _reduced(this.xf.step(result, false));
+        }
+        return result;
+    };
+
+    return _curry2(function _xall(f, xf) { return new XAll(f, xf); });
+})();
--- a/src/internal/_xdrop.js
+++ b/src/internal/_xdrop.js
@@ -1,0 +1,24 @@
+var _curry2 = require('./_curry2');
+
+
+module.exports = (function() {
+    function XDrop(n, xf) {
+        this.xf = xf;
+        this.n = n;
+    }
+    XDrop.prototype.init = function() {
+        return this.xf.init();
+    };
+    XDrop.prototype.result = function(result) {
+        return this.xf.result(result);
+    };
+    XDrop.prototype.step = function(result, input) {
+      if (this.n > 0) {
+          this.n -= 1;
+          return result;
+      }
+      return this.xf.step(result, input);
+    };
+
+    return _curry2(function _xdrop(n, xf) { return new XDrop(n, xf); });
+})();
--- a/src/internal/_xfindIndex.js
+++ b/src/internal/_xfindIndex.js
@@ -1,0 +1,31 @@
+var _curry2 = require('./_curry2');
+var _reduced = require('./_reduced');
+
+
+module.exports = (function() {
+    function XFindIndex(f, xf) {
+        this.xf = xf;
+        this.f = f;
+        this.idx = -1;
+        this.found = false;
+    }
+    XFindIndex.prototype.init = function() {
+        return this.xf.init();
+    };
+    XFindIndex.prototype.result = function(result) {
+        if (!this.found) {
+            result = this.xf.step(result, -1);
+        }
+        return this.xf.result(result);
+    };
+    XFindIndex.prototype.step = function(result, input) {
+        this.idx += 1;
+        if (this.f(input)) {
+            this.found = true;
+            result = _reduced(this.xf.step(result, this.idx));
+        }
+        return result;
+    };
+
+    return _curry2(function _xfindIndex(f, xf) { return new XFindIndex(f, xf); });
+})();
--- a/src/internal/_xfindLastIndex.js
+++ b/src/internal/_xfindLastIndex.js
@@ -1,0 +1,26 @@
+var _curry2 = require('./_curry2');
+
+
+module.exports = (function() {
+    function XFindLastIndex(f, xf) {
+        this.xf = xf;
+        this.f = f;
+        this.idx = -1;
+        this.lastIdx = -1;
+    }
+    XFindLastIndex.prototype.init = function() {
+        return this.xf.init();
+    };
+    XFindLastIndex.prototype.result = function(result) {
+        return this.xf.result(this.xf.step(result, this.lastIdx));
+    };
+    XFindLastIndex.prototype.step = function(result, input) {
+        this.idx += 1;
+        if (this.f(input)) {
+            this.lastIdx = this.idx;
+        }
+        return result;
+    };
+
+    return _curry2(function _xfindLastIndex(f, xf) { return new XFindLastIndex(f, xf); });
+})();
--- a/src/internal/_xgroupBy.js
+++ b/src/internal/_xgroupBy.js
@@ -1,0 +1,36 @@
+var _append = require('./_append');
+var _curry2 = require('./_curry2');
+var _has = require('./_has');
+
+
+module.exports = (function() {
+    function XGroupBy(f, xf) {
+        this.xf = xf;
+        this.f = f;
+        this.inputs = {};
+    }
+    XGroupBy.prototype.init = function() {
+        return this.xf.init();
+    };
+    XGroupBy.prototype.result = function(result) {
+        var key;
+        for (key in this.inputs) {
+            if (_has(key, this.inputs)) {
+                result = this.xf.step(result, this.inputs[key]);
+                if (result.__transducers_reduced__) {
+                     result = result.value;
+                     break;
+                }
+            }
+        }
+        return this.xf.result(result);
+    };
+    XGroupBy.prototype.step = function(result, input) {
+        var key = this.f(input);
+        this.inputs[key] = this.inputs[key] || [key, []];
+        this.inputs[key][1] = _append(input, this.inputs[key][1]);
+        return result;
+    };
+
+    return _curry2(function _xgroupBy(f, xf) { return new XGroupBy(f, xf); });
+})();
--- a/src/internal/_xtake.js
+++ b/src/internal/_xtake.js
@@ -1,0 +1,22 @@
+var _curry2 = require('./_curry2');
+var _reduced = require('./_reduced');
+
+
+module.exports = (function() {
+    function XTake(n, xf) {
+        this.xf = xf;
+        this.n = n;
+    }
+    XTake.prototype.init = function() {
+        return this.xf.init();
+    };
+    XTake.prototype.result = function(result) {
+        return this.xf.result(result);
+    };
+    XTake.prototype.step = function(result, input) {
+      this.n -= 1;
+      return this.n >= 0 ? this.xf.step(result, input) : _reduced(result);
+    };
+
+    return _curry2(function _xtake(n, xf) { return new XTake(n, xf); });
+})();
--- a/src/internal/_xtakeWhile.js
+++ b/src/internal/_xtakeWhile.js
@@ -1,0 +1,21 @@
+var _curry2 = require('./_curry2');
+var _reduced = require('./_reduced');
+
+
+module.exports = (function() {
+    function XTakeWhile(f, xf) {
+        this.xf = xf;
+        this.f = f;
+    }
+    XTakeWhile.prototype.init = function() {
+        return this.xf.init();
+    };
+    XTakeWhile.prototype.result = function(result) {
+        return this.xf.result(result);
+    };
+    XTakeWhile.prototype.step = function(result, input) {
+      return this.f(input) ? this.xf.step(result, input) : _reduced(result);
+    };
+
+    return _curry2(function _xtakeWhile(f, xf) { return new XTakeWhile(f, xf); });
+})();
--- a/src/internal/_xwrap.js
+++ b/src/internal/_xwrap.js
@@ -1,0 +1,14 @@
+module.exports = (function() {
+    function XWrap(fn) {
+        this.f = fn;
+    }
+    XWrap.prototype.init = function() {
+        throw new Error('init not implemented on XWrap');
+    };
+    XWrap.prototype.result = function(acc) { return acc; };
+    XWrap.prototype.step = function(acc, x) {
+        return this.f(acc, x);
+    };
+
+    return function _xwrap(fn) { return new XWrap(fn); };
+}());
--- a/src/into.js
+++ b/src/into.js
@@ -1,0 +1,44 @@
+var _curry3 = require('./internal/_curry3');
+var _isTransformer = require('./internal/_isTransformer');
+var _reduce = require('./internal/_reduce');
+var _stepCat = require('./internal/_stepCat');
+
+
+/**
+ * Transforms the items of the list with the transducer and appends the transformed items to
+ * the accumulator using an appropriate iterator function based on the accumulator type.
+ *
+ * The accumulator can be an array, string, object or a transformer. Iterated items will
+ * be appended to arrays and concatenated to strings. Objects will be merged directly or 2-item
+ * arrays will be merged as key, value pairs.
+ *
+ * The accumulator can also be a transformer object that provides a 2-arity reducing iterator
+ * function, step, 0-arity initial value function, init, and 1-arity result extraction function
+ * result. The step function is used as the iterator function in reduce. The result function is
+ * used to convert the final accumulator into the return type and in most cases is R.identity.
+ * The init function is used to provide the initial accumulator.
+ *
+ * The iteration is performed with R.reduce after initializing the transducer.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig a -> (b -> b) -> [c] -> a
+ * @param {*} acc The initial accumulator value.
+ * @param {Function} xf The transducer function. Receives a transformer and returns a transformer.
+ * @param {Array} list The list to iterate over.
+ * @return {*} The final, accumulated value.
+ * @example
+ *
+ *      var numbers = [1, 2, 3, 4];
+ *      var transducer = R.compose(R.map(R.add(1)), R.take(2));
+ *
+ *      R.into([], transducer, numbers); //=> [2, 3]
+ *
+ *      var intoArray = R.into([]);
+ *      intoArray(transducer, numbers); //=> [2, 3]
+ */
+module.exports = _curry3(function into(acc, xf, list) {
+    return _isTransformer(acc) ? _reduce(xf(acc), acc.init(), list)
+                               : _reduce(xf(_stepCat(acc)), acc, list);
+});
--- a/src/transduce.js
+++ b/src/transduce.js
@@ -1,0 +1,49 @@
+var _reduce = require('./internal/_reduce');
+var _xwrap = require('./internal/_xwrap');
+var curryN = require('./curryN');
+
+
+/**
+ * Initializes a transducer using supplied iterator function. Returns a single item by
+ * iterating through the list, successively calling the transformed iterator function and
+ * passing it an accumulator value and the current value from the array, and then passing
+ * the result to the next call.
+ *
+ * The iterator function receives two values: *(acc, value)*. It will be wrapped as a
+ * transformer to initialize the transducer. A transformer can be passed directly in place
+ * of an iterator function.
+ *
+ * A transducer is a function that accepts a transformer and returns a transformer and can
+ * be composed directly.
+ *
+ * A transformer is an an object that provides a 2-arity reducing iterator function, step,
+ * 0-arity initial value function, init, and 1-arity result extraction function, result.
+ * The step function is used as the iterator function in reduce. The result function is used
+ * to convert the final accumulator into the return type and in most cases is R.identity.
+ * The init function can be used to provide an initial accumulator, but is ignored by transduce.
+ *
+ * The iteration is performed with R.reduce after initializing the transducer.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig (c -> c) -> (a,b -> a) -> a -> [b] -> a
+ * @param {Function} xf The transducer function. Receives a transformer and returns a transformer.
+ * @param {Function} fn The iterator function. Receives two values, the accumulator and the
+ *        current element from the array. Wrapped as transformer, if necessary, and used to
+ *        initialize the transducer
+ * @param {*} acc The initial accumulator value.
+ * @param {Array} list The list to iterate over.
+ * @see R.into
+ * @return {*} The final, accumulated value.
+ * @example
+ *
+ *      var numbers = [1, 2, 3, 4];
+ *      var transducer = R.compose(R.map(R.add(1)), R.take(2));
+ *
+ *      R.transduce(transducer, R.appendTo, [], numbers); //=> [2, 3]
+ */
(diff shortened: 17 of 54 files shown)


kevinbeaty:
Alternative to #865

kevinbeaty:
Note, the mocha tests pass, but it is failing on creation of `index.js` ... I still need to look into this.  If anybody could offer any pointers, it would be appreciated.

buzzdecafe:
thanks for doing this so fast @kevinbeaty . 

> it is failing on creation of index.js

have you tried `make clean` and then try again?

kevinbeaty:
> it is failing on creation of index.js
All is well now.

buzzdecafe:
Man, this is a *lot* of work! I dropped the ball on this because i got very discouraged by the perf numbers. However, there is much that is good about this impl (if I say so myself) and there may be opportunities to optimize later. 

I am thinking more and more about using properties on functions to store metadata about them, as recommended long ago by @megawac . I wonder if we could include an *iteration strategy* on a function, s.t. when called e.g. `lazily(map(f))` it would iterate in a lazy fashion based on metadata -- a lazy iteration strategy -- on `map`; or more germane to this case, `transduce(map(f), filter(g))` would use the transducer strategy over the composition. I don't know that this is possible,or how much work it would be, or how invasive. But if there is a way to thread the needle, it could be a way to get the best of several worlds. 

That would address one of my primary concerns about this approach as it stands today: It is *very* invasive, essentially routing all iteration through `reduce`. It costs a lot in terms of typical-use performance, and you will not see the benefits until you get up to very large list sizes.

svozza:
Wow. This is amazing.

kevinbeaty:
>  I wonder if we could include an iteration strategy on a function

I think we could change the iteration strategy internally to the `_transduceDispatch` so I don't think it would be too invasive.   My initial thought would be to include a Ramda meta data property using a Symbol on the function with fallbacks similar to how we are handling iterator and transformer symbols.

Something like:
```javascript
var symRamdaMeta = typeof Symbol !== 'undefined' ? Symbol('ramdaMeta') : '@@ramdaMeta'
fn[symRamdaMeta] = {}

// in _transduceDispatch or transduce... whatever makes sense
fn[symRamdaMeta].iterationStrategy = 'transduce'
```

buzzdecafe:
> My initial thought would be to include a Ramda meta data property using a Symbol on the function with fallbacks similar to how we are handling iterator and transformer symbols.

:+1: 

`fn[symRamdaMeta].iterationStrategy = 'transduce'`

I was imagining a reference to a function, maybe something like:

```javascript
fn[symRamdaMeta].strategy = {
    lazy: lzIter,
    xf: xfIter,
    //... etc.
};
```

then the implementation of `lazily` could query the function and grab the correct iteration strategy. We can fight over design details later.

kevinbeaty:
> I was imagining a reference to a function

Ah.  Gotcha.  That could work well. 

> We can fight over design details later. 

Do you want me to propose something and update the PR, or are you saying this is a separate issue?

buzzdecafe:
> Do you want me to propose something and update the PR, or are you saying this is a separate issue?

I would like to build/see a smaller scale proof of concept first. So I think a separate issue/PR makes sense

kevinbeaty:
I updated the PR with a few minor changes to `_dispatchable` to help reduce the intrusion/performance concerns.  `_dispatchable` now resembles `_checkForMethod` with the exception of dispatching on a transformer given in list position.  Like `_checkForMethod`, the function to call as a fallback is passed in to allow an optimized version in replace of using transducers and `reduce` for the common case.  I changed `map` to use the internal version.  Note I did not need to do anything with function meta data. We might be able to make of use of that later, but it was not necessary with this change.

Using old version (map through reduce in common case on my laptop):

```
Running suite map [lib/bench/map.bench.js]...
_.map x 2,915,839 ops/sec ±4.59% (85 runs sampled)
map(sq, nums) x 339,775 ops/sec ±4.09% (84 runs sampled)
map(sq)(nums) x 307,052 ops/sec ±4.41% (80 runs sampled)
mapSq(nums) x 321,318 ops/sec ±4.33% (81 runs sampled)
Fastest test is _.map at faster than map(sq, nums)
```

Using new version (map using internal _map in common case):
```
Running suite map [lib/bench/map.bench.js]...
_.map x 3,464,050 ops/sec ±2.51% (91 runs sampled)
map(sq, nums) x 2,057,462 ops/sec ±0.89% (100 runs sampled)
map(sq)(nums) x 1,715,320 ops/sec ±0.95% (100 runs sampled)
mapSq(nums) x 1,977,764 ops/sec ±0.96% (98 runs sampled)
Fastest test is _.map at 1.68x faster than map(sq, nums)
```
Does this make you feel better about things?  I can update the other functions to fallback to the internal versions if you like this approach.

buzzdecafe:
wow @kevinbeaty that is quite an improvement

kevinbeaty:
Latest update, after a few more tweaks to `_dispatchable`

```
Running suite map [lib/bench/map.bench.js]...
_.map x 3,526,052 ops/sec ±0.92% (100 runs sampled)
map(sq, nums) x 3,165,482 ops/sec ±0.93% (100 runs sampled)
map(sq)(nums) x 2,337,636 ops/sec ±3.02% (93 runs sampled)
mapSq(nums) x 3,297,122 ops/sec ±1.16% (94 runs sampled)
_.map at 1.07x faster than mapSq(nums)
```

buzzdecafe:
:fireworks:

kevinbeaty:
Yeah. So TIL that `typeof` checks can make a big difference in micro benchmarks. 

So is that a :+1: to convert the rest of the `_dispatchable` functions in this PR to this new approach?  Does it alleviate your concerns?

buzzdecafe:
i'm pretty gobsmacked actually. well done! and yes, let's proceed. Anyone else have $0.02 to contribute?

kevinbeaty:
Ok. Converted the other functions and removed no longer needed helper functions (including `_transduceDispatch`). The diff is cleaner now.  Probably also a good time for interested parties to review.

A few notes:

- Some functions now have "functor method dispatch" that didn't before `all any drop dropWhile find findIndex findLast findLastIndex groupBy` ... If this is controversial, I can remove this by passing `null` as `methodname` for these functions and handle appropriately in `_dispatchable`.  Say the word and I'll make this change.
- `reduce` (and by extension `groupBy`) still has a performance hit due to addition of iterable dispatching and `__reduced__` checks.  Maybe this is OK.  I haven't  looked at it much so there may be opportunity for improvement.

buzzdecafe:
> `all any drop dropWhile find findIndex findLast findLastIndex groupBy`

do any other libs have these object methods that we want to enable point-free composition for? most of these seem to be pretty array specific.

buzzdecafe:
took a quick look:

* highland has `find`; that would be nice support; 
* lazy.js: `every` (`all`); `some` (`any`); this would mean :alien: which is ok by me, but may be a fight; also `drop` and `dropWhile`; also `find`
* etc.

so for the most part i guess we *do* want to dispatch on these; in some cases we may have to alias :scream: if we want interoperability

kevinbeaty:
Would it be better to leave the method dispatch out for now to avoid conflating that discussion with the discussion of transducers?

kevinbeaty:
Also, we could leave some transducers out of this initial PR and add and discuss additions in separate PRs if this chunk is too big, or merits of certain transducers are questionable...

Transducers added in this PR:

- all
- any
- drop
- dropWhile
- filter
- find
- findIndex
- findLast
- findLastIndex
- groupBy
- map
- take
- takeWhile

Of these transducers-js, transducers.js support only:

- drop
- dropWhile
- filter
- map
- take
- takeWhile

Which might be a good set to start with...

megawac:
This is looking sweet (haven't read it all and have only skimmed comments). 2 comments:

Why `step` instead of `next` which is more common for iterators?

It looks like a lot of the `X*` stuff is boiler plate - does it make sense to have a `XBase` and have all of them inherit from it - would save quite a couple of LOC.

buzzdecafe:
thanks @megawac, i'd forgotten how much we did on this. i'm very please @kevinbeatty revived it and got it into this condition. there's probably a lot of room to improve it still, any input gratefully accepted. 

@kevinbeaty 

> Also, we could leave some transducers out of this initial PR and add and discuss additions in separate PRs if this chunk is too big, or merits of certain transducers are questionable...

I'm torn. smaller bites are easier to digest. on the other hand, if we're in for a penny, we're in for a pound. i don't think we want some functions that can compose/iterate transducer-style and some that don't.

buzzdecafe:
and as to `step` vs `next`, i think `step` is the transducer/clojure term. so we may have some work to do to resolve the two names.

kedashoe:
First of all, great work @kevinbeaty!

> I'm torn. smaller bites are easier to digest. on the other hand, if we're in for a penny, we're in for a pound.

Creating new transducer types is pretty easy (and simple, &copy; R Hickey), I don't think it really matters how many are included in the initial commit. Number of transducers aside, I agree with the sentiment and believe you do not go far enough. Ramda is still pre 1.0.0. This is the time to go all in on *something* that addresses all the issues that have been raised over Ramda's life vis a vis iterating over different data types.

Currently we have in `src/map.js`

`module.exports = _curry2(_dispatchable('map', _xmap, _map));`

First of all, now we are maintaining two versions of map? But more to the point, I say get rid of dispatchable all together, replace all the current `_fn` with their `_xfn` counterparts and :dancer: This just seems so much more important to me than keeping old `map` around because.. it is 1.4 times faster than xmap? (Why is it actually still here? A better reason I'm missing?) Whatever else is around to enable this dual world, kill it! (I have not gone through all the code, I may have gotten some details wrong, but I think we get the point).

It's scarier, but Ramda has a great community, if there is anything wrong, someone will find it and fix it. If there are speed gains to be had, someone will realize them.

I think one way we can measure that this is accomplished is that Ramda uses transducers but no `transduce` function needs to be exposed to the user (I would not expose it at all, personally) and data does not need any decorated to be transduced. Nobody wants to `transduce`. Nobody wants to wrap their collections in something every time they want to iterate. They just want to `map` and `filter`.

kedashoe:
@megawac 

> Why step instead of next which is more common for iterators?

see https://github.com/cognitect-labs/transducers-js#transformer-protocol

> It looks like a lot of the X* stuff is boiler plate - does it make sense to have a XBase and have all of them inherit from it - would save quite a couple of LOC.

I actually had just that (even called xfBase) in my initial commit, but it seemed a little un-Ramda (partly because I was also supporting plain objects which complicated things). In any event, there aren't many transducers and they are quite simple having just the 3 methods, I don't think it is worth the perf hit since you don't really gain anything except a little less code

kedashoe:
btw :arrow_up: not an argument for my PR, just general thoughts on the issue. This seems to have more support, I can close #865 if you guys would rather focus here

kevinbeaty:
Oh boy, here we go :)  First of several comments ....

@megawac 
> Why step instead of next which is more common for iterators?

Couple reasons I can think of (stated first, then expanded below):

1. The signature of `next` (both in iterator in generator case) do not match the signature of step.
2. To be compatible with the "Transformer Protocol" defined/used by other libraries.

Regarding 1. `step` is a reducing function that accepts an accumulator and iteration item.  `next` either accepts no arguments (when used with iterators) or a single iteration item (when used with generators).  Although the concept is similar to iterators/generators, the idea is that the accumulator is maintained external to the transducer instead of internal to the iterator.  This allows decoupling the transformation from the input and output sources.

Regarding 2. As @kedashoe  mentioned, this conforms to the [Transformer Protocol](https://github.com/cognitect-labs/transducers-js#transformer-protocol) defined jointly by [transducers-js](https://github.com/cognitect-labs/transducers-js) (from @swannodette )  and [transducers.js](https://github.com/jlongster/transducers.js/blob/master/transducers.js) ( from @jlongster ).  Transducers written for those libraries could be composed with transducers within Ramda.

Maintaining compatibly with this protocol brings immediate support for several other libraries, already including:

- [RxJS](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/gettingstarted/transducers.md)
- [Highland](http://highlandjs.org/#transduce)
- [Kefir](http://pozadi.github.io/kefir/#transduce)
- [js-csp](https://github.com/ubolonton/js-csp)

Less importantly (warning self-promotion)
- [transduce](https://github.com/transduce/transduce) and [transduce-stream](https://github.com/transduce/transduce-stream)

.... I'm sure I missed some and probably many others to come

Although it may not be necessary to support this protocol directly, I think there should be a good reason to diverge from it...

kevinbeaty:
@megawac 

> It looks like a lot of the X* stuff is boiler plate - does it make sense to have a XBase and have all of them inherit from it - would save quite a couple of LOC.

I agree, there is a lot of duplication.  A base class could be used.  Another approach is using a function like [`transformStep`](https://github.com/transduce/transduce#transformstepxfstep).   This could certainly be evaluated...  I'm impartial as long as we maintain some sort of compatibility with the transformer protocol (see preceding comment).  I'm not sure what effect this may have on any benchmarks...

kevinbeaty:
@buzzdecafe 

> I'm torn. smaller bites are easier to digest. on the other hand, if we're in for a penny, we're in for a pound. i don't think we want some functions that can compose/iterate transducer-style and some that don't.

I agree we could be in for a pound.  But I'd really like this PR to focus on the core functionality to make sure we settle on something that we are in at all...  I agree with @kedashoe that once the core is in place, additional transducers are not difficult to add, and that effort could be "crowd sourced" if we have a core to work from.

kevinbeaty:
@kedashoe 

Thanks for the great feedback

> First of all, now we are maintaining two versions of map?

Yes.  This PR initially started with only the transducer version, but we switched back to this due to concerns re. performance and invasiveness as discussed with @buzzdecafe  above.

The first version is a transducer implementation, the second version is the fallback implementation to maintain the current behavior (with existing performance enhancements). It was a tradeoff.  Again, I'd like to get some core functionality in place that isn't too invasive from the initial implementation. And, I'd also like to keep @buzzdecafe happy :)  We can always iterate on the implementation details in several future PRs if we can agree on the core API.

> I say get rid of dispatchable

`_dispatchable` is an evolution of the existing `_checkForMethod`.  It not only supports dispatching to transducers and fallback implementations, it supports the "functor method dispatch" that is already supported within Ramda.  And from @buzzdecafe comment above, it may eventually evolve to support some kind of iteration strategy and potential aliases (but I think both should be discussed external to this PR).

Regardless, I don't know if the debate of removing existing functionality should really be conflated with the addition of transducer functionality.

> I think one way we can measure that this is accomplished is that Ramda uses transducers but no transduce function needs to be exposed to the user 

Turns out this is  difficult to do without getting very hairy, and also trying to work out the revese composition property of transducers with the current array implementations. We spent a good month debating how to do this in `ramduce` gitter room.  I'm not saying it can't be done... but not sure if it needs to be addressed immediately.

>  Nobody wants to transduce. Nobody wants to wrap their collections in something every time they want to iterate. They just want to map and filter.

With curried `into` it's actually not that bad. Want existing Ramda functionality you know and love? It still works!  Want to use transducers?  Use them directly with any library that supports the protocol!  Want to use the magic incremental aggregation that transducers enable, just use (potentially curried) `into`:

```javascript
var ramdaStyle = R.compose(R.map(fn), R.filter(p)(array)
var transducerStyle = R.into([], R.compose(R.filter(p), R.map(fn), array)
// or
var intoArray = R.into([])
var transform = R.compose(R.filter(p), R.map(fn)) // note reversal
var intoArrayStyle = intoArray(transform)

var intoObject = R.into({})

var immutableJsTransformer = ...
var intoCustom = R.into(immutableJsTransformer)
```

@buzzdecafe and I did have a hearty discussion about this a while back, and I think I'm representing the (tentative) conclusion from those discussions fairly, but please correct me if I missed something.

rockymadden:
Absolutely stellar work @kevinbeaty! I'm still working my way through all the details of this thread/PR, but from what I have seen thus far, and of your work with transducers in general, I am enthusiastic.

kevinbeaty:
> I'm sure I missed some and probably many others to come

Looks like baconjs/bacon.js#444 and cujojs/most#90 are coming soon...

kevinbeaty:
@rockymadden Thanks. To be clear, this PR was the result of collaboration with @buzzdecafe @kedashoe and others

CrossEye:
I've been offline for days, and won't have a lot of time for the next several to give this the full review it deserves.  But I'm extremely impressed with what I'm seeing, both here and in @kedashoe's separate PR.  This stuff is fantastic!

kevinbeaty:
Latest push had no new updates. Rebased to master and removed commit to `dist/ramda.js` per changes to `CONTRIBUTING.md`.

buzzdecafe:
ping @davidchambers

Makefile
```diff
@@ -50,6 +50,6 @@ test: dist/ramda.js
 	| sed 's:src/\(.*\)[.]js:exports.\1 = require("./src/\1");:' >index.js
 	sed '/"main":/d' package.json >tmp
 	mv tmp package.json
-	$(MOCHA) -- $(shell find test -name '*.js' -not -name 'installTo.js' -not -name 'test.examplesRunner.js')
+	$(MOCHA) -- $(shell find test -name '*.js' -not -name 'installTo.js' -not -name 'test.examplesRunner.js' -not -name 'listXf.js')
```
davidchambers: It might be better to use `-not -path 'test/helpers/*'` here.

src/dropWhile.js
```diff
@@ -1,6 +1,7 @@
 var _curry2 = require('./internal/_curry2');
+var _dispatchable = require('./internal/_dispatchable');
 var _slice = require('./internal/_slice');
-
+var _xdropWhile = require('./internal/_xdropWhile');
```
davidchambers: Nit: Convention is to have two blank lines between imports and export.

src/internal/_isIterable.js
```diff
@@ -0,0 +1,5 @@
+var _symIterator = require('./_symIterator');
+
+module.exports = function _isIterable(x) {
+    return x[_symIterator] || (x.next && typeof x.next === 'function');
+};
```
davidchambers: Do we want to return a Boolean? This may not be necessary for an internal function.

src/internal/_isTransformer.js
```diff
@@ -0,0 +1,6 @@
+var _symTransformer = require('./_symTransformer');
+
+module.exports = function _isTransformer(obj) {
+    return (obj[_symTransformer] !== void 0) ||
```
davidchambers: Why are we strict here (`!== void 0`) but not in `_isIterable`?

Makefile
kevinbeaty: Agreed.

src/internal/_isTransformer.js
kevinbeaty: No reason.   Should I make the change here or `_isIterable`?

davidchambers: I'd prefer `!= null` in both places.

src/internal/_symIterator.js
```diff
@@ -0,0 +1 @@
+module.exports = (typeof Symbol !== 'undefined') ? Symbol('iterator') : '@@iterator';
```
davidchambers: We could define `_Symbol` like so:

```javascript
module.exports = typeof Symbol !== 'undefined' ? Symbol : function(s) { return '@@' + s; };
```

Then this file could become:

```javascript
var _Symbol = require('./_Symbol');

module.exports = _Symbol('iterator');
```

I haven't checked how many places we reference `Symbol`; this may not be worthwhile.

src/internal/_stepCat.js
```diff
@@ -0,0 +1,62 @@
...
+      if (typeof obj === 'string') {
+          return _stepCatString;
+      }
+      if (typeof obj === 'object') {
+          return _stepCatObject;
+      }
+      throw new Error('Cannot create transformer for ' + obj);
+  }
+
+  return _stepCat;
```
davidchambers:
```javascript
return function _stepCat(obj) {
    ...
};
```

src/internal/_xall.js
```diff
@@ -0,0 +1,35 @@
...
+
+    XAll.prototype.step = function(result, input) {
+        if (!this.f(input)) {
+            this.all = false;
+            result = _reduced(this.xf.step(result, false));
+        }
+        return result;
+    };
+
+    return _curry2(_xall);
```
davidchambers:
```javascript
return _curry2(function _xall(f, xf) { return new XAll(f, xf); });
```

src/internal/_xtake.js
```diff
@@ -0,0 +1,33 @@
...
+    XTake.prototype.init = function() {
+        return this.xf.init();
+    };
+
+    XTake.prototype.result = function(result) {
+        return this.xf.result(result);
+    };
+
+    XTake.prototype.step = function(result, input) {
+      if (this.n-- > 0) {
```
davidchambers: Let's use `--this.n >= 0`.

src/internal/_symIterator.js
kevinbeaty: So far it's only `_symIterator` and `_symTransformer`.   We might actually want to use `Symbol.iterator` instead of `Symbol("iterator")` as it is [well known](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#Well-known_symbols).

In fact the use of `Symbol("transformer")` could be debatable.  It is used in `transducers.js` but not in `transducers-js`...

src/internal/_xtake.js
davidchambers: Actually, this function could be written much more clearly:

```javascript
this.n -= 1;
return this.n >= 0 ? this.xf.step(result, input) : _reduced(result);
```

src/internal/_xtakeWhile.js
```diff
@@ -0,0 +1,32 @@
...
+        return this.xf.result(result);
+    };
+
+    XTakeWhile.prototype.step = function(result, input) {
+        if (this.f(input)) {
+            result = this.xf.step(result, input);
+        } else {
+            result = _reduced(result);
+        }
+        return result;
```
davidchambers: I'd prefer:

```javascript
return this.f(input) ? this.xf.step(result, input) : _reduced(result);
```

src/internal/_xwrap.js
```diff
@@ -0,0 +1,20 @@
...
+        return new XWrap(fn);
+    }
+
+    function XWrap(fn) {
+        this.f = fn;
+    }
+    XWrap.prototype.init = function() {
+        throw new Error('init not implemented on XWrap');
+    };
+    XWrap.prototype.result = identity;
```
davidchambers: We probably don't want `identity` with its `R.__` handling. We could define `_identity` or define it inline.

src/into.js
```diff
@@ -0,0 +1,16 @@
...
+module.exports = _curry3(function into(to, xf, ls) {
+    var acc, fn;
+    if (_isTransformer(to)) {
+        fn = to;
+        acc = to.init();
+    } else {
+        fn = _stepCat(to);
+        acc = to;
+    }
+    return _reduce(xf(fn), acc, ls);
```
davidchambers: I'd prefer:

```javascript
return _isTransformer(to) ? _reduce(xf(to), to.init(), ls)
                          : _reduce(xf(_stepCat(to)), to, ls);
```

This is what I derive from the current code, but I'd rather not have to do the variable substitutions in my head. :)

src/transduce.js
```diff
@@ -0,0 +1,6 @@
+var _reduce = require('./internal/_reduce');
+var curryN = require('./curryN');
+
+module.exports = curryN(4, function(xf, fn, acc, ls) {
```
davidchambers: Is `ls` short for "list"? We don't use this name elsewhere. Could we use `list` instead?

test/transduce.js
```diff
@@ -0,0 +1,66 @@
+var assert = require('assert');
+
+var R = require('..');
+
+
+describe('transduce', function() {
+    var add = function(a, b) {return a + b;};
+    var mult = function(a, b) {return a * b;};
+//    var sq = function(x) { return x * x; };
+//    var gt10 = function(x) { return x > 10; };
```
davidchambers: Remove these?

kevinbeaty:
@davidchambers Great feedback.  Pushed an update with your suggested changes.

src/into.js
```diff
@@ -0,0 +1,9 @@
+var _curry3 = require('./internal/_curry3');
+var _isTransformer = require('./internal/_isTransformer');
+var _reduce = require('./internal/_reduce');
+var _stepCat = require('./internal/_stepCat');
+
+module.exports = _curry3(function into(to, xf, list) {
```
davidchambers: Could you add the JSDoc boilerplate?

src/transduce.js
```diff
@@ -0,0 +1,6 @@
+var _reduce = require('./internal/_reduce');
+var curryN = require('./curryN');
+
+module.exports = curryN(4, function(xf, fn, acc, list) {
```
davidchambers: Here, too.

davidchambers:
It's looking great! I commented on two functions which need documentation, but otherwise this looks ready to merge. :)

buzzdecafe:
:tada: 

fan-f***ing-tastic

kevinbeaty:
Awesome.  Will update the docs tonight.

kevinbeaty:
I'll plan on adding a note in the public functions that support transducers.

Something like

```javascript
/**
  * ...
  * Acts as a transducer if a transformer is given in list position.
  * @see R.transduce
  */
```

And then offer a short explanation to transducer and transformer terminology in the JSDoc for `transduce`

src/internal/_xgroupBy.js
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    XGroupBy.prototype.init = function() {
+        return this.xf.init();
+    };
+
+    XGroupBy.prototype.result = function(result) {
+        var key;
+        for (key in this.inputs) {
+            result = this.xf.step(result, this.inputs[key]);
```
AutoSponge: Should there be a hasOwnProperty check of key?

buzzdecafe: should be *pretty* safe since `this.inputs` should really only be manipulated by the `XGroupBy` instance.

buzzdecafe: ... but it wouldn't hurt to be safer

kevinbeaty: Ok. Will add _has check in the next push

src/internal/_stepCat.js
```diff
@@ -0,0 +1,54 @@
...
+       step: _appendTo,
+       result: _result
+  };
+  var _stepCatString = {
+      init: function() { return ''; },
+      step: _add,
+      result: _result
+  };
+  var _stepCatObject = {
+      init: function() { return {}; },
```
megawac: This can just be `Object` as `function() { return []; }` can be `Array`

kevinbeaty: Interesting.  `String` fine as well, I presume?

megawac: Yepz as long as the user isn't going to be calling them directly

kevinbeaty:
* Updated JSDoc per @davidchambers 
* Added `_has` check to `XGroupBy` per @AutoSponge 
* Updated `_stepCat` to use native functions on `init` per @megawac 
* Added a few tests

buzzdecafe:
any input @CrossEye ?

src/internal/_xfindIndex.js
```diff
@@ -0,0 +1,31 @@
...
+        return this.xf.init();
+    };
+    XFindIndex.prototype.result = function(result) {
+        if (!this.found) {
+            result = this.xf.step(result, -1);
+        }
+        return this.xf.result(result);
+    };
+    XFindIndex.prototype.step = function(result, input) {
+        this.idx++;
```
davidchambers: Nit: I'd prefer `+= 1` here (or prefix `++`).

src/internal/_xfindLastIndex.js
```diff
@@ -0,0 +1,26 @@
...
+        this.lastIdx = -1;
+    }
+    XFindLastIndex.prototype.init = function() {
+        return this.xf.init();
+    };
+    XFindLastIndex.prototype.result = function(result) {
+        return this.xf.result(this.xf.step(result, this.lastIdx));
+    };
+    XFindLastIndex.prototype.step = function(result, input) {
+        this.idx++;
```
davidchambers: Nit: I'd prefer `+= 1` here (or prefix `++`).

src/internal/_xdrop.js
```diff
@@ -0,0 +1,24 @@
...
+    }
+    XDrop.prototype.init = function() {
+        return this.xf.init();
+    };
+    XDrop.prototype.result = function(result) {
+        return this.xf.result(result);
+    };
+    XDrop.prototype.step = function(result, input) {
+      if (this.n > 0) {
+          this.n--;
```
davidchambers: Nit: I'd prefer `-= 1` here (or prefix `--`).

src/internal/_reduce.js
```diff
@@ -1,7 +1,20 @@
...
+    if (typeof fn === 'function') {
+        fn = _xwrap(fn);
+    }
+    if (isArrayLike(list)) {
+        return _arrayReduce(fn, acc, list);
+    } else if (_isIterable(list)) {
+        return _iterableReduce(fn, acc, list);
+    } else {
+        // TODO: better message here please
+        throw new TypeError(typeof list + ' is not reducible');
```
AutoSponge: `else` seems redundant when each condition is terminal (returns or throws).

kevinbeaty:
Incorporated latest tweaks from @AutoSponge and @davidchambers

src/into.js
```diff
@@ -0,0 +1,44 @@
...
+ * be appended to arrays and concatenated to strings. Objects will be merged directly or 2-item
+ * arrays will be merged as key, value pairs.
+ *
+ * The accumulator can also be a transformer object that provides a 2-arity reducing iterator
+ * function, step, 0-arity initial value function, init, and 1-arity result extraction function
+ * result. The step function is used as the iterator function in reduce. The result function is
+ * used to convert the final accumulator into the return type and in most cases is R.identity.
+ * The init function is used to provide the initial accumulator.
+ *
+ * The iteration is performed with R.reduce after initializing the transducer.
```
davidchambers: Terrific description!

megawac:
:+1: looks pretty much good to me. I would say making a `X` base class would be something worth doing in the future to remove some of the redundancy tho

src/into.js
CrossEye: :+1:

src/into.js
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig a -> b -> c -> a
```
davidchambers: This isn't quite right. Perhaps something like this:

    a -> (b -> b) -> [c] -> a

I sense there's some relationship between `a`, `b`, and `c`, but this might only be apparent in `b`'s methods.

test/into.js
```diff
@@ -0,0 +1,50 @@
+var assert = require('assert');
+
+var R = require('..');
+
+
+describe('into', function() {
+    var add = R.curry(function(a, b) {return a + b;});
```
CrossEye: Not particularly important to worry about here, but is there any reason not to use `R.add` here?

CrossEye:
This is fantastic.  I've finally gotten through a once-over of everything and managed to run some performance tests (no significant changes for current benchmarks; presumably there would be speed-ups in combination ones we never wrote.)

I'm very, very impressed!

I was barely listening as the initial passes at this stuff was going on.  I knew _what_ you were talking about but not much about _how_ you were going about it.  This looks really well-done.

I wouldn't delay merging the code for this, as we can address it later.  But I'd like to either have a discussion or to learn about reasons already discussed for choosing function constructors over `Object.create` (even if we shimmed it) or a possibility like this, for `_xmap`:

```javascript
var _curry2 = require('./_curry2');


module.exports = _curry2(function _xmap(f, xf) {
    return {
        xf: xf,
        f: f,
        init: function() {
            return this.xf.init();
        },
        result: function(result) {
            return this.xf.result(result);
        },
        step: function(result, input) {
            return this.xf.step(result, this.f(input));
        }
    }; 
});
```

It's not at all that I think one of these other options is necessarily better, but I think we should consider them... unless this was already hashed out when I was not paying attention.

Really nice work!

test/into.js
kevinbeaty: Nope. Will change

src/into.js
kevinbeaty: Will change here and similarly in transduce.

kevinbeaty:
* `@sig` tweaks per @davidchambers 
* @ `R.add` in test per @CrossEye 
* Rebased to master

kevinbeaty:
@megawac 
> I would say making a X base class would be something worth doing in the future to remove some of the redundancy tho

@CrossEye 

>  But I'd like to either have a discussion or to learn about reasons already discussed for choosing function constructors over Object.create

All options worth exploring.

> I wouldn't delay merging the code for this, as we can address it later.
:+1:  Please :)

CrossEye:
Everyone ok with merging this one?  This is one of our most sweeping changes to date, so I'd really like to hear from @buzzdecafe, @davidchambers, @kedashoe, and (I suppose :wink:) @kevinbeaty if there any objections.  Of course anyone is free to chime in.  (Especially you @megawac. :wink:)

buzzdecafe:
let's do it -- big thanks to @kevinbeaty and @kedashoe for their great work -- particularly reviving this when i had dropped the ball.

One caveat -- this now means we have clojure-style iterators when transducing. SO they support array-like, string, iterator-protocol, and object; this means weird unspecified behavior when iterating objects because of indeterminate iteration order.

AutoSponge:
:+1:

buzzdecafe:
> would this constitute a major or minor version?

since we're pre-1.0 I think we can get away with a minor bump

kevinbeaty:
> One caveat -- this now means we have clojure-style iterators when transducing. SO they support array-like, string, iterator-protocol, and object; this means weird unspecified behavior when iterating objects because of indeterminate iteration order.

Actually, there is no implicit object iteration in this PR for this very reason (we could certainly add it in the future).  It currently supports merging objects on the accumulator side but iteration of objects must be explicit (using ES6 style `Symbol.iterator` or `@@iterator`)

buzzdecafe:
i stand corrected! thanks @kevinbeaty

davidchambers:
:deciduous_tree:

buzzdecafe:
:tada:

megawac:
:confetti_ball:

svozza:
Fantastic work!

kevinbeaty:
Awesome.

I have to admit, hitting that "Delete Branch" button felt pretty good ;)

Thanks all.

PR #879, opened 2015-03-06
simplify implementation of R.omit

davidchambers:
This implementation is simpler in the terms of the number of direct and indirect internal dependencies. If we're to promote the use case outlined in #829 we should avoid defining functions in terms of other functions where practical. I realize this is the antithesis of functional programming!

### DIFF
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -2414,6 +2414,30 @@
     });
 
     /**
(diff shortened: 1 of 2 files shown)


buzzdecafe:
LGTM :cow2:

buzzdecafe:
> I realize this is the antithesis of functional programming!

implementation is orthogonal to interface

PR #881, opened 2015-03-06
add R.none

josepot:
I keep missing a function named `R.none`, which would be the counterpart of [`R.all`](http://ramdajs.com/docs/#all). It's easy to accomplish the same thing by doing `R.all(R.not(fn))` but that's not very readable.

In some occasions I've found myself doing things like:

    var foo = R.map(parseInt, ['1', '2', 'sfa']);
    var noneAreNaN = R.all(R.not(isNaN));
    if (noneAreNaN(foo)) {
      // Do something
    }

And I would like to be able to do this:

    var foo = R.map(parseInt, ['1', '2', 'sfa']);
    if (R.none(isNaN)(foo)) {
      // Do something
    }

### DIFF
--- a/src/none.js
+++ b/src/none.js
@@ -1,0 +1,24 @@
+var _any = require('./internal/_any');
+var _curry2 = require('./internal/_curry2');
+var _dispatchable = require('./internal/_dispatchable');
+var _not = require('./internal/_not');
+var _xany = require('./internal/_xany');
+
+
+/**
+ * Returns `true` if no elements of the list match the predicate,
+ * `false` otherwise.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig (a -> Boolean) -> [a] -> Boolean
+ * @param {Function} fn The predicate function.
+ * @param {Array} list The array to consider.
+ * @return {Boolean} `true` if the predicate is not satisfied by every element, `false` otherwise.
+ * @example
+ *
+ *      R.none(isNaN, [1, 2, 3]); //=> true
+ *      R.none(isNaN, [1, 2, 3, NaN]); //=> false
+ */
+module.exports = _curry2(_not(_dispatchable('any', _xany, _any)));
--- a/test/none.js
+++ b/test/none.js
@@ -1,0 +1,33 @@
+var assert = require('assert');
+
+var R = require('..');
+
+
+describe('none', function() {
+    var even = function(n) {return n % 2 === 0;};
+    var T = function() {return true;};
+
+    it('returns true if no elements satisfy the predicate', function() {
+        assert.strictEqual(R.none(even, [1, 3, 5, 7, 9, 11]), true);
+    });
+
+    it('returns false if any element satisfies the predicate', function() {
+        assert.strictEqual(R.none(even, [1, 3, 5, 7, 8, 11]), false);
+    });
+
+    it('returns true for an empty list', function() {
+        assert.strictEqual(R.none(T, []), true);
+    });
+
+    it('works with more complex objects', function() {
+        var xs = [{x: 'abcd'}, {x: 'adef'}, {x: 'fghiajk'}];
+        function len3(o) { return o.x.length === 3; }
+        function hasA(o) { return o.x.indexOf('a') >= 0; }
+        assert.strictEqual(R.none(len3, xs), true);
+        assert.strictEqual(R.none(hasA, xs), false);
+    });
+
+    it('is automatically curried', function() {
+        assert.strictEqual(R.none(even)([1, 3, 5, 6, 7, 9]), false);
+    });
+});
--- a/src/internal/_not.js
+++ b/src/internal/_not.js
@@ -1,0 +1,3 @@
+module.exports = function _not(f) {
+    return function() {return !f.apply(this, arguments);};
+};
--- a/src/not.js
+++ b/src/not.js
@@ -1,4 +1,5 @@
 var _curry1 = require('./internal/_curry1');
+var _not = require('./internal/_not');
 
 
 /**
@@ -18,6 +19,4 @@ var _curry1 = require('./internal/_curry1');
  *      f(11); //=> false
  *      f(9); //=> true
  */
-module.exports = _curry1(function not(f) {
-    return function() {return !f.apply(this, arguments);};
-});
+module.exports = _curry1(_not);


dist/ramda.js
```diff
@@ -2330,6 +2335,28 @@
...
+     * @sig (a -> Boolean) -> [a] -> Boolean
+     * @param {Function} fn The predicate function.
+     * @param {Array} list The array to consider.
+     * @return {Boolean} `true` if the predicate is not satisfied by every element, `false` otherwise.
+     * @example
+     *
+     *      var xs = R.map(parseInt, ['1', '2', '3', 'foo']);
+     *      xs; //=> [1, 2, 3, NaN]
+     *      R.none(isNaN)(xs); //=> false
+     *      xs.pop();
```
buzzdecafe: should we merge this, i don't think we should have the example *mutate* the list. Better to create e.g. `ys = [1, 2, 3]`

test/none.js
```diff
@@ -0,0 +1,43 @@
...
+
+    it('returns false if any element satisfies the predicate', function() {
+        assert.strictEqual(R.none(even, [1, 3, 5, 7, 8, 11]), false);
+    });
+
+    it('returns true for an empty list', function() {
+        assert.strictEqual(R.none(T, []), true);
+    });
+
+    it('short-circuits on first false value', function() {
```
buzzdecafe: this seems like an implementation detail to me. It's nice that it has this behavior, but do we really need to test for it?

josepot: @buzzdecafe The only thing that I did was adapt [the tests of `R.all`](https://github.com/ramda/ramda/blob/master/test/all.js) for `R.none`, in order to make it consistent with the current code base. But I can remove these tests if you want.

dist/ramda.js
josepot: :+1:

test/none.js
buzzdecafe: was wondering about that. I probably should have looked for myself :smile: 

it's probably me who added that test, too

josepot: np! :)

davidchambers:
This function seems like a reasonable addition.

dist/ramda.js
```diff
@@ -6354,6 +6385,7 @@
         multiply: multiply,
         nAry: nAry,
         negate: negate,
+        none: none,
```
davidchambers: We're no longer updating __dist/ramda.js__ with each commit. Could you rebase your branch and undo the changes to this file?

src/internal/_none.js
```diff
@@ -0,0 +1,9 @@
+module.exports = function _none(fn, list) {
+    var idx = -1;
+    while (++idx < list.length) {
+        if (fn(list[idx])) {
+            return false;
+        }
+    }
+    return true;
+};
```
davidchambers: There's no need for this internal function, since it's only used in the definition of `R.none`.

src/none.js
```diff
@@ -0,0 +1,22 @@
+var _curry2 = require('./internal/_curry2');
+var _none = require('./internal/_none');
+/**
```
davidchambers: Nit: Two blank lines after imports.

src/none.js
```diff
@@ -0,0 +1,22 @@
...
+ * Returns `true` if no elements of the list match the predicate,
+ * `false` if there are any that do.
```
davidchambers: This line could be simplified to:

     * `false` otherwise.

src/none.js
```diff
@@ -0,0 +1,22 @@
...
+ * @param {Function} fn The predicate function.
+ * @param {Array} list The array to consider.
+ * @return {Boolean} `true` if the predicate is not satisfied by every element, `false` otherwise.
+ * @example
+ *
+ *      var xs = R.map(parseInt, ['1', '2', '3', 'foo']);
+ *      xs; //=> [1, 2, 3, NaN]
+ *      var ys = [1, 2, 3];
+ *      R.none(isNaN)(xs); //=> false
+ *      R.none(isNaN)(ys); //=> true
```
davidchambers: You could provide both arguments at once in these examples.

test/none.js
```diff
@@ -0,0 +1,43 @@
...
+        var test = function(n) {count++; return even(n);};
+        var result = R.none(test, [1, 3, 5, 6, 7, 9]);
+        assert(!result);
+        assert.strictEqual(count, 4);
+    });
+
+    it('works with more complex objects', function() {
+        var xs = [{x: 'abcd'}, {x: 'adef'}, {x: 'fghiajk'}];
+        function len3(o) { return o.x.length === 3; }
+        function hasA(o) { return o.x.indexOf('a') > -1; }
```
davidchambers: I'd prefer to see `>= 0` in place of `> -1`.

test/none.js
```diff
@@ -0,0 +1,43 @@
...
+        assert.strictEqual(R.none(len3, xs), true);
+        assert.strictEqual(R.none(hasA, xs), false);
+    });
+
+    it('is automatically curried', function() {
+        var count = 0;
+        var test = function(n) {count++; return even(n);};
+        assert(R.none(test)([1, 3, 5, 6, 7, 9]) === false);
```
davidchambers: Nit: Use `assert.strictEqual` rather than `assert`.

test/none.js
```diff
@@ -0,0 +1,43 @@
...
+
+    it('returns true for an empty list', function() {
+        assert.strictEqual(R.none(T, []), true);
+    });
+
+    it('short-circuits on first false value', function() {
+        var count = 0;
+        var test = function(n) {count++; return even(n);};
+        var result = R.none(test, [1, 3, 5, 6, 7, 9]);
+        assert(!result);
```
davidchambers: Nit: Use `assert.strictEqual` rather than `assert`.

josepot:
@davidchambers :+1: Please let me know if there is anything else that you would like me to change. Thanks!

test/none.js
```diff
@@ -0,0 +1,43 @@
...
+        function len3(o) { return o.x.length === 3; }
+        function hasA(o) { return o.x.indexOf('a') >= 0; }
+        assert.strictEqual(R.none(len3, xs), true);
+        assert.strictEqual(R.none(hasA, xs), false);
+    });
+
+    it('is automatically curried', function() {
+        var count = 0;
+        var test = function(n) {count++; return even(n);};
+        assert.strictEqual(R.none(test)([1, 3, 5, 6, 7, 9]), false);
```
davidchambers: Since we're not asserting `count`, this could be `R.none(even)` rather than `R.none(test)`.

test/none.js
davidchambers: I agree with @buzzdecafe that we should remove this test. We should leave the order in which we test the elements unspecified to avoid tying ourselves to a particular implementation.

CrossEye: Yes, we should remove it from `R.all` as well.

It seemed to make sense at the time, but I wouldn't want to imagine how we'd try to explain it if we were ever to re-implement in a concurrent alternative on some imaginary multi-threaded version of the language.  So it's definitely in fact an implementation detail.

CrossEye:
I agree that this function makes sense.  Some sugar is worth having.

test/none.js
davidchambers: #914

src/none.js
```diff
@@ -0,0 +1,28 @@
...
+ * @memberOf R
+ * @category List
+ * @sig (a -> Boolean) -> [a] -> Boolean
+ * @param {Function} fn The predicate function.
+ * @param {Array} list The array to consider.
+ * @return {Boolean} `true` if the predicate is not satisfied by every element, `false` otherwise.
+ * @example
+ *
+ *      R.none(isNaN)([1, 2, 3, NaN]); //=> false
+ *      R.none(isNaN)([1, 2, 3]); //=> true
```
davidchambers: I like your simplification of the examples. :)

I think it would read better with the `true` example before the `false` one, but I don't know *why* I think this. Also, it would be clearer to provide both arguments at once.

     *      R.none(isNaN, [1, 2, 3]); //=> true
     *      R.none(isNaN, [1, 2, 3, NaN]); //=> false

Aside from this niggle, the pull request looks great!

src/none.js
```diff
@@ -0,0 +1,28 @@
...
+ */
+module.exports = _curry2(function(fn, list) {
+    var idx = -1;
+    while (++idx < list.length) {
```
davidchambers: We should cache `list.length`:

```javascript
var idx = -1;
var len = list.length;
while (++idx < len) {
```

josepot: @davidchambers :+1:  np! I've changed it already. However, I don't quite understand why: I was under the impression that accessing the length of an Array had a complexity of O(1), so I don't see how this will make it different. Would you mind explaining what I'm missing here? Thanks!

davidchambers: Accessing the `length` property of an array is O(1), but we're doing it N times. If `list` contains 100 items, caching the value of the `length` property will save 99 property accesses. It's a minor optimization really, but we do it elsewhere and it's nice to be consistent.

davidchambers:
:deciduous_tree: Very nice patch, @josepot! I'll merge as soon as Travis has finished running the tests.

megawac:
I feel like this one should be implemented via composition

buzzdecafe:
:+1:  @megawac 

`not(any(f))` 

doesn't seem too bad to me

CrossEye:
I feel about this one the same as I do about `reject(fn)` as sugar for `filter(not(fn))`.  It's certainly not essential, but it makes code more readable and comes at little cost.

megawac:
I'm fine with `R.none` being implemented but I'd rather see it implemented as `R.none = R.not(R.any)`

davidchambers:
> I'm fine with `R.none` being implemented but I'd rather see it implemented as `R.none = R.not(R.any)`

Excellent idea! Could you switch to this implementation, @josepot?

josepot:
@davidchambers I can do that, np! However, performance-wise I think that the current implementation would be slightly better.

In order to do that I will have to move the [`not`](https://github.com/ramda/ramda/blob/master/src/not.js#L21) function into the 'internal' folder, so that I can reuse it there. Is that ok?

megawac:
@josepot it should perform comparably (just one extra call to not) if not better (possible short circuiting from the transducers)

davidchambers:
> In order to do that I will have to move the [`not`](https://github.com/ramda/ramda/blob/master/src/not.js#L21) function into the 'internal' folder, so that I can reuse it there. Is that ok?

Sounds good!

josepot:
@davidchambers @megawac @buzzdecafe done! Please, let me know if that's what you were looking for. Thanks!

CrossEye:
> In order to do that I will have to move the [`not`](https://github.com/ramda/ramda/blob/master/src/not.js#L21) function into the 'internal' folder, so that I can reuse it there. Is that ok?

I think this is a mistaken impression.  While there's nothing wrong with doing this, it's also not essential.  I would actually like to reduce the number of `_internal` functions used.

davidchambers:
LGTM. I'll let @CrossEye merge this, in case he has reservations.

CrossEye:
No reservations, just that observation.

PR #883, opened 2015-03-07
remove _pairWith

davidchambers:


dist/ramda.js
```diff
@@ -3263,6 +3246,35 @@
...
+     * @param {Object} obj The object to extract from
+     * @return {Array} An array of key, value arrays from the object's own properties.
+     * @example
+     *
+     *      R.toPairs({a: 1, b: 2, c: 3}); //=> [['a', 1], ['b', 2], ['c', 3]]
+     */
+    var toPairs = _curry1(function toPairs(obj) {
+        var pairs = [];
+        for (var prop in obj) {
+            if (Object.prototype.hasOwnProperty.call(obj, prop)) {
```
buzzdecafe: don't we have a private reference to this?

davidchambers: We do, but with #829 in mind we should avoid internal dependencies where practical.

buzzdecafe: in this case though, wasn't the point of the internal function to grab a reference in case someone messed with the Object prototype? (FWIW, I never found that particularly compelling, but i believe that's why it's there)

davidchambers: If that was the original reason for including `_has`, we're no longer fulfilling the objective. Here's the [current implementation][1]:

```javascript
module.exports = function _has(prop, obj) {
    return Object.prototype.hasOwnProperty.call(obj, prop);
};
```

I don't agree with the original reasoning. If the global environment is damaged, all bets are off.


[1]: https://github.com/ramda/ramda/blob/e29cf515e0714189856927617ec22c7e12667633/src/internal/_has.js

davidchambers: #889

CrossEye: All the activity and I've been offline for days!

And I don't have much time now, unfortunately.  I just want to make a few points:

* I don't think #829 is a good reason for changing our implementations.  I know that @davidchambers has made a number of posts with this rationale, and I simply don't buy it.  I think there are good reasons to ensure that our dependency trees are not horribly deep the way that lodash used to be before a recent refactoring.  But I don't see a reason for every function to try to stand alone either.  If the changes help for other reasons, I have no objections, but let's not make a fetish out of this.  Any `require` mechanism will straighten that out in a reasonable way.  (On a somewhat related note, I _would_ like to remove the separation between `internal/_xyz` and `xyz` where we can.  I think that unless there are really strong reasons for not doing so, we should simply use the curried, dispatched functions internally.)

* The reason people capture `Object.prototype.hasOwnProperty` is generally not because someone might override it on `Object.prototype` but that they might do so on an arbitrary object.  Obviously if they've done it on the global before you take your copy, you're already out of luck, so there's not much use in worrying about that.  But for other objects, if you capture the global version and use that one instead, you won't be caught out by someone's screwy object.  This is not a major concern for me, but it's a small enough effort to ensure that we don't get caught in these circumstances.

* Is there any major reason for this PR besides reducing internal dependencies?  Do you see this as cleaner code?  More performant?  I'm hesitant to do anything base solely upon the dependency tree.

CrossEye: I meant to add that if people want stand-alone implementations, a library like [cherries][ch] might be a good fit.  I haven't really played with it much.  But it's goals were to provide simple individual functional constructs with no dependencies at all.  I guess we'd have to ask @casperin if there are any plans to break it into more easily digestible modules for NPM, but I never thought about such separability being an important goal for Ramda.


  [ch]: https://github.com/casperin/cherries/blob/master/cherries.js

casperin:
@CrossEye (about Cherries): I was not planning to, but if there's any interest, then I'd love to do it. ... in fact I think I might do it anyway. I'm missing a personal project to work on currently.

dist/ramda.js
casperin: @CrossEye (about Cherries): I was not planning to, but if there's any interest, then I'd love to do it. ... in fact I think I might do it anyway. I'm missing a personal project to work on currently.

edit: Sorry, I didn't realise I was writing in the wrong box :(

buzzdecafe:
:cow2:. n.b. `functions/functionsIn` are implemented in a similar fashion.

src/toPairs.js
```diff
@@ -12,11 +10,19 @@ var keys = require('./keys');
...
  * @return {Array} An array of key, value arrays from the object's own properties.
  * @example
  *
  *      R.toPairs({a: 1, b: 2, c: 3}); //=> [['a', 1], ['b', 2], ['c', 3]]
  */
-module.exports = _curry1(_pairWith(keys));
+module.exports = _curry1(function toPairs(obj) {
+    var pairs = [];
+    for (var prop in obj) {
+        if (Object.prototype.hasOwnProperty.call(obj, prop)) {
```
davidchambers: I'll wait to see the outcome of #889 before updating this line.

davidchambers: Changed to `_has`.

PR #887, opened 2015-03-08
use `xs.push(x)` rather than `xs[xs.length] = x`

davidchambers:
We're currently mixing the two approaches. Let's stick with the more commonly seen `xs.push(x)`.

### DIFF
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -87,14 +87,14 @@
         var idx;
         var len1 = set1.length;
         var len2 = set2.length;
-        var result = new Array(len1 + len2);
+        var result = [];
         idx = -1;
         while (++idx < len1) {
-            result[idx] = set1[idx];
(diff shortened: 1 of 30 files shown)


davidchambers:
I'll merge this. If it turns out the other style is preferred, I'll open a pull request to use *that* consistently.

buzzdecafe:
prefer the other way. i think this one should be reverted

davidchambers:
> i think this one should be reverted

It wouldn't be a straight reversion, as we should commit to one approach or the other. If we prefer the other approach we should use it everywhere.

Why do you prefer `xs[xs.length] = x`?

buzzdecafe:
this was a micro-optimization added way back. unless there is a compelling reason to undo those changes, i don't think we should

davidchambers:
#890

PR #888, opened 2015-03-08
replace ++ and -- operators

### LINKED ISSUE #886: avoid ++ and -- operators
Postfix `++` and `--` are excessively tricky operators (to paraphrase [Crockford][1]). `x++` evaluates to the current value of `x` *then* increments `x`. I avoid these operators in my own code and our JSHint config forbids them in projects at Plaid.

Prefix `++` and `--` are less tricky but entirely unnecessary. `++x` can be written as `x += 1`. Again, I have JSHint configured to forbid the use of these operators.

The existence of both prefix and postfix forms is itself a source of confusion.

We have lots of code like this:

```javascript
var idx = -1;
var len = xs.length;
while (++idx < len) { ... }
```

I don't share the devotion to `while` over `for`. I don't like `++`, nor do I like initializing `idx` to a value one less than the first index. The code above can be expressed more clearly as a `for` loop:

```javascript
for (var idx = 0, len = xs.length; idx < len; idx += 1) { ... }
```

For reverse loops we use this idiom:

```javascript
var idx = xs.length;
while (idx--) { ... }
```

I dislike this approach for two reasons. First, it uses the tricksy postfix `--` operator. Secondly, it relies on implicit Number → Boolean conversion in evaluating the condition. A `for` loop can avoid these sources of incidental complexity:

```javascript
for (var idx = xs.length - 1; idx >= 0; idx -= 1) { ... }
```

I propose that we use JSHint's [`plusplus`][2] option to forbid `++` and `--` in both prefix and postfix forms, and that our conditional expressions evaluate to `true` or `false`.

[1]: http://www.jslint.com/lint.html#inc
[2]: http://jshint.com/docs/options/#plusplus

davidchambers:
Closes #886

### DIFF
--- a/test/unfold.js
+++ b/test/unfold.js
@@ -9,7 +9,15 @@ describe('unfold', function() {
     });
 
     it('is cool!', function() {
-        var fib = function(n) {var count = 0; return function(pair) {if (count++ < n) {return [pair[0], [pair[1], pair[0] + pair[1]]];}};};
+        var fib = function(n) {
+            var count = 0;
+            return function(pair) {
+                count += 1;
+                if (count <= n) {
+                    return [pair[0], [pair[1], pair[0] + pair[1]]];
+                }
+            };
+        };
         assert.deepEqual(R.unfold(fib(10), [0, 1]), [0, 1, 1, 2, 3, 5, 8, 13, 21, 34]);
     });
 
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,6 +1,7 @@
 {
   "evil": true,
   "eqnull": true,
+  "plusplus": true,
   "predef": ["beforeEach", "console", "define", "describe", "it", "module", "process", "require"],
   "undef": true,
   "unused": true
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -14,8 +14,7 @@
     };
 
     var _all = function _all(fn, list) {
-        var idx = -1;
-        while (++idx < list.length) {
+        for (var idx = 0, len = list.length; idx < len; idx += 1) {
             if (!fn(list[idx])) {
                 return false;
             }
@@ -24,8 +23,7 @@
     };
 
     var _any = function _any(fn, list) {
-        var idx = -1;
-        while (++idx < list.length) {
+        for (var idx = 0, len = list.length; idx < len; idx += 1) {
             if (fn(list[idx])) {
                 return true;
             }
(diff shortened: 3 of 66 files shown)


buzzdecafe:
https://github.com/ramda/ramda/pull/535

davidchambers:
#886 makes an argument for avoiding the `++` and `--` operators (both prefix and postfix). #535 makes no argument at all.

We should choose a position from the following matrix:

|                                | allow ++ and -- | disallow ++ and -- |
| ------------------------------ | --------------- | ------------------ |
| __allow implicit coercion__    | Option 1        | Option 2           |
| __disallow implicit coercion__ | Option 3        | Option 4           |

__Option 1:__

```javascript
var idx = -1;
var len = xs.length;
while (++idx < len) {…}
```

```javascript
var idx = xs.length;
while (idx--) {…}
```

__Option 2:__

There's no reason to use implicit coercion if one is not using the `while (idx--)` idiom.

__Option 3:__

```javascript
var idx = -1;
var len = xs.length;
while (++idx < len) {…}
```

```javascript
var idx = xs.length;
while (--idx >= 0) {…}
```

__Option 4:__

```javascript
// while                      // for
var idx = -1;                 for (var idx = 0, len = xs.length; idx < len; idx += 1) {…}
var len = xs.length;
while ((idx += 1) < len) {…}
```

```javascript
// while                      // for
var idx = xs.length;          for (var idx = xs.length - 1; idx >= 0; idx -= 1) {…}
while ((idx -= 1) >= 0) {…}
```

Note that it's possible to satisfy the requirements of option 4 using a `while` loop. I favour a `for` loop for three reasons:

  - `idx` is initially assigned the first (or last) index, not the one less (or more) than this value;
  - the condition does not contain an assignment expression; and
  - it indicates to the reader that `idx` and `len` are loop variables (`let` can one day replace `var`).

I favour option 4. Where do you lie on the matrix, @buzzdecafe?

CrossEye:
I'm quite happy with choice 3.  I've been pretty disengaged and didn't get a chance to respond to #886, but I absolutely prefer the readability of `while` loops.  (Once upon a time there was some performance benefit too, but I'm guessing that has washed away over time.)  I never found Crockford's arguments against `++`/`--` to be convincing.  Especially in prefix form, they seem to be a useful and benign construct in Javascript.

I think it would be a good idea to get away from `while (idx--) {…}`, although I don't think I share some of your abhorrence of implicit conversions in general.  But I would much rather keep to the `while` loops.

davidchambers:
You don't find `var idx = -1;` distasteful, @CrossEye?

> I think it would be a good idea to get away from `while (idx--) {…}`

I'm pleased to learn our views are not entirely at odds! Would you accept a pull request which replaces occurrences of `while (idx--)` with `while (--idx >= 0)`?

buzzdecafe:
I've never had an Amana Radar Range, so i guess I'll choose ... Door Number 3?

CrossEye:
> You don't find var idx = -1; distasteful, @CrossEye?

I find it much less distasteful than a `for`-loop.... which is not a "no".  :smile:

> Would you accept a pull request which replaces occurrences of `while (idx--)` with `while (--idx >= 0)`?

Absolutely.  I think it's significantly more readable.

test/unfold.js
```diff
@@ -9,7 +9,15 @@ describe('unfold', function() {
     });
 
     it('is cool!', function() {
```
buzzdecafe: hmmm. subjective?

davidchambers:
> > Would you accept a pull request which replaces occurrences of `while (idx--)` with `while (--idx >= 0)`?
>
> Absolutely. I think it's significantly more readable.

#947

PR #899, opened 2015-03-08
stop updating dist/ramda.js with each commit

### LINKED ISSUE #849: stop updating dist/ramda.js with each commit
Updating __dist/ramda.js__ with each commit is problematic for several reasons:

  - it makes contributing more difficult, particularly for documentation tweaks;
  - it adds a lot of noise, making pull requests more difficult to review; and
  - it increases the likelihood of merge conflicts.

While there are some benefits to having __dist/ramda.js__ up to date at all times, they do not justify the costs. I suggest this file only be built at release time (as is the case with __dist/ramda.min.js__).

davidchambers:
Closes #849

We could add __dist/ramda.js__ to __.gitignore__ but this would add complexity to the prepublish script.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,18 +6,13 @@
 
         $ git checkout -b <branch>
 
-3.  Add new functions as separate source files in src/. If the function is
-    used by ramda functions elsewhere in uncurried form, then place the
-    implementation in src/internal (see src/internal/_map.js as an example).
-
-    NOTE: The file `dist/ramda.js` is **generated**.  Any changes to it will
-    be overridden.
+3.  If adding a function `R.foo`, define it in __src/foo.js__ and include unit
+    tests in __test/foo.js__. If adding an internal function `_foo`, define it
+    in __src/internal/_foo.js__.
 
 4.  Make one or more atomic commits. Each commit should have a descriptive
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -7,17 +7,12 @@
         $ git checkout -b <branch>
 
 3.  Add new functions as separate source files in src/. If the function is
```
buzzdecafe: should we add a line/subsection something like "Test files should be included for all functions. Include your test files in the `test` directory."

davidchambers: We have "Include unit tests if adding a function or fixing a bug." in the next list item, but I think it'd be better here. I'll move it.

davidchambers: Updated.

buzzdecafe:
:cow2:

davidchambers:
I'm keen to hear @CrossEye's thoughts on this, as I believe he has something different in mind.

CrossEye:
I think this is fine for now.  What I really would like is to check this on Travis after every successful build, and have Travis check it back in, so that users get it as part of cloning the repo.  It would be `.gitignore`d so that users wouldn't accidentally check it in.  But there are real issues trying to do that only once across the mutliple Travis builds, so IMHO this is fine for now.  If I can ever figure out those Travis issues, I'd like to take another swing at it.

PR #901, opened 2015-03-10
add R.test

davidchambers:
I find myself defining this function quite frequently. It's a nice companion to `R.match`.

Note that `RegExp.prototype.test` is not referentially transparent:

```javascript
> var pattern = /x/g
undefined
> pattern.lastIndex
0
> pattern.test('xyz')
true
> pattern.lastIndex
1
> pattern.test('xyz')
false
```

It's thus necessary to invoke the method on a clone of `pattern`.

The implementation of `_cloneRegExp` comes from purescript/purescript-strings#6.

We should also use `_cloneRegExp` in `clone`, but I've leave that for a separate pull request.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -1,0 +1,22 @@
+var assert = require('assert');
+
+var R = require('..');
+
+
+describe('test', function() {
+    it('returns true if string matches pattern', function() {
+        assert.strictEqual(R.test(/^x/, 'xyz'), true);
+    });
+
+    it('returns false if string does not match pattern', function() {
+        assert.strictEqual(R.test(/^y/, 'xyz'), false);
+    });
+
+    it('is referentially transparent', function() {
+        var pattern = /x/g;
+        assert.strictEqual(pattern.lastIndex, 0);
+        assert.strictEqual(R.test(pattern, 'xyz'), true);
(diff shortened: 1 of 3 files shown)


test/test.js
```diff
@@ -0,0 +1,22 @@
...
+        assert.strictEqual(R.test(/^y/, 'xyz'), false);
+    });
+
+    it('is referentially transparent', function() {
+        var pattern = /x/g;
+        assert.strictEqual(pattern.lastIndex, 0);
+        assert.strictEqual(R.test(pattern, 'xyz'), true);
+        assert.strictEqual(pattern.lastIndex, 0);
+        assert.strictEqual(R.test(pattern, 'xyz'), true);
+    });
```
buzzdecafe: worthwhile to show how this fails for native `test`?

buzzdecafe:
LGTM

megawac:
:+1:

test/test.js
davidchambers: The assertions on lines 19 and 20 fail if one removes the call to `_cloneRegExp`, so I think what we have is sufficient.

davidchambers:
What say you, @CrossEye?

CrossEye:
:herb:

PR #902, opened 2015-03-10
Add logical && and ||. Rename R.and -> R.both, and R.or -> R.either

### LINKED ISSUE #843: Add logical (|| , &&, !) and rename functional versions
As of 0.10 `R.or`, `R.and`, and `R.not` all operate on functions. While this is certainly useful, the names sound more like functional implementations of the operators. The matching operator functions are not present in the library. We offer many other operator functions (`R.add`, `R.eq`, `R.gt,` etc),  and I think `||`, `&&`, and `!` are valid additions to complete the suite. 

I propose:
* `R.or` renamed to `R.either`
* `R.and` renamed to `R.both`
* `R.or = (a,b) => a || b`
* `R.and = (a, b) => a && b`

! is a bit trickier. I'm having difficulty coming up with an intuitive name for the functional "not".
Logical "not" should certainly be `R.not = (a) => ! a` but best I have for functional not is `R.opposite` or `R.notBy`.

Thoughts?

jethrolarson:
Fixes #843
Decided to skip ! for now.

### DIFF
--- a/src/and.js
+++ b/src/and.js
@@ -1,30 +1,23 @@
 var _curry2 = require('./internal/_curry2');
 
-
 /**
- *
- * A function wrapping calls to the two functions in an `&&` operation, returning the result of the first
- * function if it is false-y and the result of the second function otherwise.  Note that this is
- * short-circuited, meaning that the second function will not be invoked if the first returns a false-y
- * value.
+ * A function that returns the first argument if it's falsy otherwise the second
+ * argument. Note that this is NOT short-circuited, meaning that if expressions
+ * are passed they are both evaluated.
  *
  * @func
  * @memberOf R
  * @category Logic
- * @sig (*... -> Boolean) -> (*... -> Boolean) -> (*... -> Boolean)
- * @param {Function} f a predicate
- * @param {Function} g another predicate
- * @return {Function} a function that applies its arguments to `f` and `g` and ANDs their outputs together.
+ * @sig * -> * -> *
+ * @param {*} a any value
+ * @param {*} b any other value
+ * @return {*} the first argument if falsy otherwise the second argument.
  * @example
  *
- *      var gt10 = function(x) { return x > 10; };
- *      var even = function(x) { return x % 2 === 0 };
- *      var f = R.and(gt10, even);
- *      f(100); //=> true
- *      f(101); //=> false
+ *      R.and(false, true); //=> true
+ *      R.and(0, []); //=> []
+ *      R.and(null, ''); => false
  */
-module.exports = _curry2(function and(f, g) {
-    return function _and() {
-        return f.apply(this, arguments) && g.apply(this, arguments);
-    };
+module.exports = _curry2(function and(a, b) {
+    return a && b;
 });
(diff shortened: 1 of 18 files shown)


davidchambers:
:thumbsdown: from me for the reasons given [here][1] and [here][2].

If we *are* to make this change I strongly believe `R.not` should change as well.


[1]: https://github.com/ramda/ramda/issues/843#issuecomment-76621221
[2]: https://github.com/ramda/ramda/issues/843#issuecomment-77297120

src/and.js
```diff
@@ -1,30 +1,23 @@
...
+ * are passed they are both evaluated.
  *
  * @func
  * @memberOf R
  * @category Logic
- * @sig (*... -> Boolean) -> (*... -> Boolean) -> (*... -> Boolean)
- * @param {Function} f a predicate
- * @param {Function} g another predicate
- * @return {Function} a function that applies its arguments to `f` and `g` and ANDs their outputs together.
+ * @sig * -> * -> Boolean
```
buzzdecafe: may not return a boolean

buzzdecafe:
> If we *are* to make this change I strongly believe `R.not` should change as well.

:+1: to that. If not `nix`, maybe `anti` will do. 

Otherwise, I am fine with this; I understand @davidchambers 's objections to the JS-iness of it. But it seems essential to me to be able to compose with these kinds of truthy comparisons. If you must force to Boolean, you could always `compose(Boolean, or)('moo', 'cow')`

src/and.js
jethrolarson: Despite the dumb diff that's unchanged from what it was before

jethrolarson:
Yeah. It seemed off to reject it just it just because it has all the problems that the JS one does. We don't reject R.add because it works on strings

src/and.js
jethrolarson: Nm, I was wrong. Will fix when once at keyboard

CrossEye:
> Yeah. It seemed off to reject it just it just because it has all the problems that the JS one does. We don't reject R.add because it works on strings

:+1:

This library will always exist in some tension between being familiar to JS developers and being comfortable for developers more familiar with more strongly-FP languages.  I don't know that there are any overall guidelines, but to me this decision is pretty easy: these functions are meant to work like `add`, `modulo`, etc., producing functions that act like language operators.  That these ones can't quite do that is the only thing that makes me hesitate (https://github.com/ramda/ramda/issues/843#issuecomment-76114415) but on balance they still seem worth having.

That they can be used to create bad code bothers me very little.  Heck, we include `forEach`, which I would much rather not have, as I think that's much more likely to lead to bad code, but certain exigencies forced it on us.

buzzdecafe:
@jethrolarson can you please squash to one commit? then i think we are good to go. 

Sorry @davidchambers you can't win them all -- and you have been on a winning streak!

buzzdecafe:
aw, bugger, it's out of date. @jethrolarson can you please rebase & amend? i will merge promptly

jethrolarson:
@buzzdecafe Done.

PR #903, opened 2015-03-11
clarify `always` documentation

raine:
`always` doesn't internally clone non-primitives like objects so this might
be good to note.

Discussion in gitter:

https://gitter.im/ramda/ramda?at=55001e958233c93368d0a1e8


CrossEye:
:herb:

Thanks

PR #906, opened 2015-03-11
rename R.func

### LINKED ISSUE #905: rename R.func
`func` has never struck me as a good name. It's not a verb (assuming "func" is short for "function").

How about renaming it `invoke`? This *is* a verb, has an association with OOP (I think of function *application* but method *invocation*), and would result in the function appearing just before `invoker` in the documentation.

davidchambers:
Closes #905

I did not make `R.func` an :alien: for `R.invoke` as I'd like to propose a change to `R.invoke` in a subsequent pull request and we'll want to preserve the current `R.func` behaviour until we remove that function.

### DIFF
--- a/src/func.js
+++ b/src/func.js
@@ -7,6 +7,7 @@ var curry = require('./curry');
  * after `fn` and `obj` are passed in to `fn`. If no additional arguments are passed to `func`,
  * `fn` is invoked with no arguments.
  *
+ * @deprecated since v0.12.0
  * @func
  * @memberOf R
  * @category Object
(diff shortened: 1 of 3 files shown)


src/func.js
```diff
@@ -7,6 +7,7 @@ var curry = require('./curry');
  * after `fn` and `obj` are passed in to `fn`. If no additional arguments are passed to `func`,
  * `fn` is invoked with no arguments.
  *
+ * @deprecated since v0.12.0
```
buzzdecafe: should probably add see `invoke`

davidchambers: Updated.

PR #907, opened 2015-03-11
(WIP) Meta implementation

megawac:
For discussion before this goes any further

Currently the obvious 15 tests fail.

This implements all the internal methods that rely on the `arity` of a function through metadata. It also tracks the source function and name of a function.

An obvious extension of this is to do native unwrapping (for example doing currying and partial and other decorators in a single method as lodash does [big benefit!])

The main motivation of this is (for me) is a) a big perf benefit by avoiding arity and b) enables us to later implement `restParams` which will be a substantial improvement in several places in the code base (it didn't make sense to implement this through arity). Overall this enables a couple follow up optimizations

Points of interest:

- Should it be added to an object publicly   or use a symbol
- What should be included
- Should repl stuff be added

Benchmarks to follow (in case its not obvious the benefit): http://jsperf.com/ramda-currying-cost/4

### DIFF
--- a/src/internal/_meta.js
+++ b/src/internal/_meta.js
@@ -1,0 +1,32 @@
+module.exports = (function() {
+    // Should we use a Symbol or let users have free access to metadata
+    // which could be useful for extensions.
+    var metaSymbol = typeof Symbol === 'function' ? Symbol('metadata') : '@@metadata';
+
+    return {
+        set: function(wrapper, child, arity) {
+            wrapper[metaSymbol] = {
+                arity: arity != null ? arity : child.length,
+                name: child.name,
+                wrapped: child
+            };
+            return wrapper;
+        },
+        get: function(wrapper) {
+            return wrapper[metaSymbol];
+        },
+        arity: function(wrapper) {
+            return wrapper[metaSymbol] ?
+                    wrapper[metaSymbol].arity :
+                    wrapper.length;
+        },
+        // Find the original source of the wrapped function
+        // (for debugging)
+        source: function(wrapper) {
+            while (wrapper[metaSymbol]) {
+                wrapper = wrapper[metaSymbol].child;
+            }
+            return wrapper;
+        }
+    };
+})();
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -33,6 +33,18 @@
         return false;
     };
 
+    var _arrayReduce = function _arrayReduce(xf, acc, list) {
+        var idx = -1, len = list.length;
+        while (++idx < len) {
+            acc = xf.step(acc, list[idx]);
+            if (acc && acc.__transducers_reduced__) {
+                acc = acc.value;
+                break;
+            }
+        }
+        return xf.result(acc);
+    };
+
     var _assoc = function _assoc(prop, val, obj) {
         var result = {};
         for (var p in obj) {
@@ -42,32 +54,8 @@
         return result;
     };
 
-    /**
-     * Basic, right-associative composition function. Accepts two functions and returns the
-     * composite function; this composite function represents the operation `var h = f(g(x))`,
-     * where `f` is the first argument, `g` is the second argument, and `x` is whatever
-     * argument(s) are passed to `h`.
-     *
-     * This function's main use is to build the more general `compose` function, which accepts
-     * any number of functions.
-     *
-     * @private
-     * @category Function
-     * @param {Function} f A function.
-     * @param {Function} g A function.
-     * @return {Function} A new function that is the equivalent of `f(g(x))`.
-     * @example
-     *
-     *      var double = function(x) { return x * 2; };
-     *      var square = function(x) { return x * x; };
-     *      var squareThenDouble = _compose(double, square);
-     *
-     *      squareThenDouble(5); //≅ double(square(5)) => 50
-     */
-    var _compose = function _compose(f, g) {
-        return function () {
-            return f.call(this, g.apply(this, arguments));
-        };
+    var _cloneRegExp = function _cloneRegExp(pattern) {
+        return new RegExp(pattern.source, (pattern.global ? 'g' : '') + (pattern.ignoreCase ? 'i' : '') + (pattern.multiline ? 'm' : '') + (pattern.sticky ? 'y' : '') + (pattern.unicode ? 'u' : ''));
     };
 
     /**
@@ -109,6 +97,21 @@
         return false;
     };
 
+    /*
+     * Returns a function that makes a multi-argument version of compose from
+     * either _compose or _composeP.
+     */
+    var _createComposer = function _createComposer(composeFunction) {
+        return function () {
+            var idx = arguments.length - 1;
+            var fn = arguments[idx];
+            while (idx--) {
+                fn = composeFunction(arguments[idx], fn);
+            }
+            return fn;
+        };
+    };
+
     /**
      * Create a function which takes a comparator function and a list
      * and determines the winning value by a compatator. Used internally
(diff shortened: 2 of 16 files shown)


src/internal/_meta.js
```diff
@@ -0,0 +1,32 @@
+module.exports = (function() {
+    // Should we use a Symbol or let users have free access to metadata
+    // which could be useful for extensions.
```
CrossEye: We can always use a symbol and later decide to publicly expose the symbol if we choose.

CrossEye:
There have been several things lately that made us talk about including metadata, although off-hand I can't recall what they were.  I don't have any strong objection to it.

But I'm bothered by this change.  We discussed a while ago (#140) some of the rationale for properly reporting the arity of our functions.  To my mind, nothing has changed.  It's a matter of being good citizens.  We don't know how consumers of Ramda functions or of functions generated via Ramda functions will then use them.  If we break the contract the user has with some other tool that needs to know the arity of functions passed to it, it's entirely our fault, not the user's, and not that of the other tool.  I'd rather not be in that circumstance.

While we're not turning Javascript into Haskell, Ramda does seem to be pushing for a somewhat stronger typing of functions than is typical in Javascript.  I believe that this helps people work with Javascript as a functional language.  And properly reporting arity is part of how we we do this.

-----

I looked over the jsperf, and it seems to me to prove a few things:

  * standard lodash is quite a bit faster than Ramda for this case
  * lodash-fp is suprisingly slower than standard lodash
  * there seems to be little difference between Ramda and Ramda meta

-----

I didn't understand what you meant about including "repl stuff".

megawac:
> I looked over the jsperf, and it seems to me to prove a few things:
- standard lodash is quite a bit faster than Ramda for this case
- lodash-fp is suprisingly slower than standard lodash
- there seems to be little difference between Ramda and Ramda meta

This is the first step of closing that gap as I mention in the issue comments; this will enable various new optimization strategies

buzzdecafe:
i think we may have to pay the price of `arity`. but i still want to pursue this line and see where it leads.

megawac:
Very few libraries rely on function.length that don't allow the user to
override the property (e.g. curryN). Thinking about it the metadata
properties should be exposed publicly.

Based on a quick & dirty grep closing this door will make it hard to
improve the performance on the following functions any further without a
ton of boiler plate (methods that would benefit heavily from restParams)

- curry and curryN
- partial*
- converge
- flip
- func
- All invoker results
- lift
- useWith
- Any function using checkForMethod

CrossEye:
> Very few libraries rely on function.length that don't allow the user to override the property (e.g. curryN).

But the point is not what other current libraries do or don't do.  It's a matter of what we do.  If another library altered a user's functions in such a way, she could not comfortably pass them to Ramda without having to manually choose their arities.  That would burden her code.  That's why I wouldn't want Ramda to do it to her for her interoperation with another library.

Sure, I'd love to improve our performance.  But that will always be secondary to me to having an API I prefer.

buzzdecafe:
it seems as though `function.length` is getting less and less respect. `rest` params do not play well with `function.length`. Are we going to ignore them, or come to some accommodation?

megawac:
We can also add a function that returns a function that corresponds to
metadata.arity. Maybe rename arity arityN?

CrossEye:
> Are we going to ignore [`rest` params], or come to some accommodation?

As far as I can tell, although they are very useful, they are useful for creating variadic functions.  And Ramda will probably never play particularly nicely with variadic functions.  At the very least, we might well auto-curry them to the wrong lengths.

I don't think there's any good way around this.  With good reason, Ramda gently pushes toward slightly more strongly-typed functions than are the norm in JS.  It is more opinionated in this than some of the bigger libraries, because its goals are different.  It's not meant to be a multi-paradigm toolkit that will work with whatever style you want.  It's meant to support a very specific style of coding.  It's hard to see where most variadic functions would fit with that style.

CrossEye:
> We can also add a function that returns a function that corresponds to metadata.arity. Maybe rename arity arityN?

That sounds like locking the barn after the horse has escaped.  The problem is how Ramda is being a good citizen, not how it would offer ways for a user to clean up after it.

There's something bothering me about the approach here, and it encompasses all of the metadata ideas, even though I've had at least somewhat positive feelings about them.  The problem is that at the moment we deal with all functions on the same basis: whether they are native functions, Host object functions, user functions, third-party library functions, Ramda functions, or functions created by calls to Ramda functions, they get treated the same.  If we start tagging functions with metadata, and dealing with them on that basis, we make everything but the Ramda, and Ramda-created functions in some way second-class citizens.  Maybe that won't ever have a deleterious effect, but it makes me worry.

buzzdecafe:
@megawac -- ok to close this?

megawac:
up to you guys - still valuable iin my opinion if you were asking

buzzdecafe:
i think it's worth exploring, subject to @CrossEye 's constraints. not sure how difficult that makes it to thread the needle

CrossEye:
@megawac: Yours have been some of the most valuable contributions to the library.  Of course we're asking!  Of course I disagree with many of your suggestions.  But I disagree with many of those from @buzzdecafe and @davidchambers, too.  Hashing these things out and coming up with useful compromises is what makes for a good codebase.

buzzdecafe:
first question: can we use function metadata to "multi-dispatch" e.g.: At present we can only dispatch to a function with the same name: `R.chain(f, Maybe.Just(1))` works because `Just` has a `chain` method. But what if we want to dispatch from `chain` to an object with a `flatMap` method? A lookup in the metadata might be a good way to go?

buzzdecafe:
i think we should revive this. I've been playing with this idea and like it. I propose we work around the arity thing by providing a public function to set correct function.length, but internally we use something like this. then we can also set other meta-properties e.g. to support debugging.

@megawac lmk if you want to/have time to pick this up, or if you want to pass it off.

megawac:
> I propose we work around the arity thing by providing a public function
to set correct function.length, but internally we use something like this.

Great! I think lodash is planning to do the same (<issue>)

Want to move this to a ramda branch to allow more contributors? I'm pretty
busy for the next couple days but can spend some more time on it next week

buzzdecafe:
created `meta` branch

buzzdecafe:
((:bell:))

paldepind:
Excuse me but can anybody explain what this meta stuff actually does? I'm not learning much by looking at the code in this PR and I have not found explanations elsewhere.

We store some data about our functions and somehow use it? How is that different from today where the arity of our functions is implicit information based on which curry function is called? Am I correct that these ideas are inspired from Lodash?

CrossEye:
@paldepind: I'm not sure that these ideas were exactly inspired by lodash, but @megawac, who, though not as active now, was in the past one of our most prolific contributors, and helped us figure out a number of thorny issues, including some serious performance ones.  He has long-suggested this approach to us.  @megawac is a committer on Underscore, I believe, and has also been active in lodash.  Lodash has definitely adopted this approach.  I don't think Underscore has; it would make their source code significantly more complicated, something I know that @jashkenas is loathe to do.

The basic idea is that instead of wrapping and returning new functions, we could often decorate existing functions with a `ramdaProperties` (or some such) object that details information about such things as whether the function is curried, what parameters have been partially applied with which values,  Instead of wrapping functions in our arity-controlling wrappers, we could just keep an `arity` property in this expando.  Then when we run, we could interpret this metadata, thereby skipping many of the nested calls to which Ramda is prone.

There are many reasons that this might be worth considering.  Such notions as debug-only settings or declarative type-checking might be significantly easier to do if each one was simply an application of some meta-data to be interpreted later.  I can easily handling some of our dispatching through such a mechanism.

Of course this means that the actual application of our functions would be significantly more complex.  But there is a very good chance that we would gain significant performance.  This is most telling when you run performance comparisons between Ramda and lodash:  for many (most?) of our functions, Ramda runs just as fast as lodash, often faster, in Chrome.  But in Firefox, where there is a much greater cost to function invocations, Ramda can be substantially slower.  I expect a great deal of this has to do with the number of fewer function calls needed, partly because of this metadata.

To my mind, performance will only drive our decisions when there are egregious problems.  We are building a library designed to support a certain style of work.  It's all about the API.  Some time after 1.0, we might want to take this on, but it's not a driving factor.  So that's not enough of a reason for me to consider this.  But some of the other reasons sound like good ones.

However, it scares me a fair bit, too.  Most of our functions are _simple_, both in conception and in implementation.  Creating a function runner that works universally for every conceivable combination of curries, placeholders, dispatch configurations, debug settings, and type validators, strikes me as a nasty, bug-prone operation that might not be easily decomposeable into smaller ones.  Obviously we can takes some inspiration from the lodash version, but their internal style is so different from our own that it might not offer us much help, and it would probably offer none for our carrying this notion to things like dispatching and type validation.  I'm also worried but what I suggested earlier, that this might somehow privilege Ramda functions over user ones.  But that might be something we end up not having to worry about; it's a problem I can imagine creating, depending on how this is implemented.

In any case, that's a description.  And as seems to be happening lately, there's two sides to this, and I seem to be squarely on both of them!  :frowning:

buzzdecafe:
i agree this may cut both ways. but i would *much* rather explore interesting approaches and understand their implications rather than foreclose them.

CrossEye:
Absolutely.   I want to see where this might go.   I want to see what might happen with your `compose` changes.   I want to see if we can return curried functions everywhere, at least by default.

Metadata interests me less than some of the others, because in the end its mostly about implementation.   And while I recognize how important a good implementation is, I much more interested at first in the API that's being built.

paldepind:
Thank you for the write up @CrossEye. I still don't fully get it I'm afraid :(

My best guess right now is that this means we should have mega-functions that can cover a bunch of different cases by taking a small specification of what they should do.

But, how for instance would this allow us to avoid the `arity` function? Every time a user partially applies a function he needs to get a new function back with the correct arity? How can one avoid that?

buzzdecafe:
i think what the `_meta` proposal is driving at is instead of decorating the function we care about with more and more wrapper functions, we assign decorations to a property on the function object.

paldepind:
@buzzdecafe Then why not just create on single wrapper function that can do all the wrapping? Like:

```javascript
var someFunction = doEverythingWrapper({
  dispatchesOn: 'someProperty',
  arity: 3,
  // Lots of stuff
}, function() {
  /* actual function */
});
```

megawac:
Definitely possible but it'd probably be hard for others to understand (see
lodash baseWrapper) and difficult to tune for jit optimizations. The
advantage being less loc

paldepind:
@megawac How is it different from attaching the data directly to the function? In both cases the meta handling function will receive the function and the meta data?

buzzdecafe:
https://github.com/ramda/ramda/commit/976c7cf15c91713a6db0d8cafa822c2dfa4fbd44

megawac:
@simon the result is roughly equivalent (jack of all wrappers vs current
focussed wrappers) but I feel like this way is more explicit, at least for
now. I would prefer to now store this on an expando but it wouldn't be
feasible for legacy environments (requires weakmap as in lodashs
implementation)

paldepind:
@megawac Ok. The other way seems more intuitive to me. But there are probably things I'm not seeing. In any case I think the do everything wrapper

jdalton:
I'm not a fan of adding enumerable properties for meta data. This means in some environments it'll show up for all kinds of methods that iterate over the function. The nice thing about lodash's approach is it has no side effects on the function. In fact, you can remove it (there's builds with it removed) and things still just work.

megawac:
That won't work for ramdas case where tracking some metadata is paramount
(such as arity)

jdalton:
Ramda has introduced a fair amount of inconsistency with dispatching so I guess this just adds to it then.

buzzdecafe:
concern troll much?

PR #908, opened 2015-03-12
style: fix indentation

davidchambers:


buzzdecafe:
sure. FWIW I strongly prefer 2-space indentation, but i think that ship has sailed

davidchambers:
> I strongly prefer 2-space indentation

I do too, aesthetically. It also reduces line length, offering a practical benefit for those of use who have our text editors display 80 columns. :)

Would you be open to switching from four spaces to two, @CrossEye?

CrossEye:
Not at all.   I've been used to either over the years.   The only reason I like the larger one is to discourage deep nesting.

PR #909, opened 2015-03-12
simplify implementation of _stepCat

davidchambers:
I noticed that the body of `_stepCatObject.step` can be written as an expression if we use `_createMapEntry`. :)


buzzdecafe:
much cleaner. :cow2:

PR #910, opened 2015-03-12
remove R.I

davidchambers:
See #859


buzzdecafe:
:sob:

this one hurts as bad as when `cons`, `car`, and `cdr` got whacked

PR #911, opened 2015-03-12
remove deedpoll

davidchambers:
I'm now of the opinion that deedpoll needn't be part of the linting process, though I imagine I'll continue to use it from time to time to keep this project's identifiers consistent.


CrossEye:
While removing this wouldn't bother me much, I think it's done a fair bit of good and believe it should be replaced at least by a style guide.

PR #912, opened 2015-03-12
Remove _symTransformer until official spec is worked out.

kevinbeaty:
Prompted by cognitect-labs/transducers-js#20 from @tgriesser, removing
the internal transformer symbol until issues are worked out. This symbol
is currently only supported in jlongster/transducers.js and
interoperability is questionable as (Symbol('transformer') !==
Symbol('transformer')).

This feature was not documented, nor do the tests depend on it, and
only effects an undocumented feature of into. Better to remove it now
until details are worked out.  If an "official spec" is agreed upon,
I will open a new PR.

### DIFF
--- a/src/internal/_isTransformer.js
+++ b/src/internal/_isTransformer.js
@@ -1,7 +1,3 @@
-var _symTransformer = require('./_symTransformer');
-
-
 module.exports = function _isTransformer(obj) {
-    return (obj[_symTransformer] != null) ||
-        (typeof obj.step === 'function' && typeof obj.result === 'function');
+    return typeof obj.step === 'function' && typeof obj.result === 'function';
 };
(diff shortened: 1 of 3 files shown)


buzzdecafe:
:cow2:

src/internal/_isTransformer.js
```diff
@@ -1,7 +1,3 @@
-var _symTransformer = require('./_symTransformer');
-
-
 module.exports = function _isTransformer(obj) {
-    return (obj[_symTransformer] != null) ||
-        (typeof obj.step === 'function' && typeof obj.result === 'function');
+    return (typeof obj.step === 'function' && typeof obj.result === 'function');
```
davidchambers: Nit: No need for parens here.

kevinbeaty:
Removed parens in `_isTransformer` per @davidchambers

davidchambers:
:deciduous_tree:

PR #914, opened 2015-03-12
remove short-circuiting tests for R.any and R.all

davidchambers:
This was discussed in #881. These tests assert left-to-right enumeration order, but nothing in the definitions of these functions requires a particular enumeration order. Rather than *test* for a specific implementation, let's consider short-circuiting part of the *specification* for each of these functions.

### DIFF
--- a/test/all.js
+++ b/test/all.js
@@ -36,14 +36,6 @@ describe('all', function() {
         assert.deepEqual(intoArray(R.all(T), []), [true]);
     });
 
-    it('short-circuits on first false value', function() {
-        var count = 0;
-        var test = function(n) {count++; return even(n);};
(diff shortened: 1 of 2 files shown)


buzzdecafe:
:cow2:

CrossEye:
:+1:

@buzzdecafe and I discussed this today.  It really seemed like an important point at the time I wrote them, but the more I think about it, the less I like these tests.

I'm not sure about your distinction between _test_ and _specification_ here.  I think of unit tests as one way to write a decent (partial, to be sure, but useful) specification.

davidchambers:
All I was saying is that the specification of a function can live anywhere. For example, we could include:

_`R.all` stops performing tests as soon as a value disproves the hypothesis._

in the documentation.

PR #915, opened 2015-03-13
Combine internal reduce functions into one file.

kevinbeaty:
Merges _arrayReduce, _iterableReduce, _isIterable, _symIterator into
_reduce.


kevinbeaty:
I think this makes the implementation of `_reduce` easier to follow, and removes several `_internal` files only used within `_reduce`.   But it's really a matter of taste, so feel free to shoot this one down :)

CrossEye:
:herb:

This looks much easier to follow.

buzzdecafe:
thanks!

PR #916, opened 2015-03-13
Fix typo in pathEq docs

michaelckelly:
`R.pathEq` now accepts a list rather than a period delimited string.


davidchambers:
:deciduous_tree: Thanks, Michael!

PR #917, opened 2015-03-13
n-wise list splitting

buzzdecafe:


megawac:
:+1: for the function -1 for the name. I would definitely call this chunk

test/nwise.js
```diff
@@ -0,0 +1,22 @@
...
+
+describe('nwise', function() {
+    var sevenLs = [1, 2, 3, 4, 5, 6, 7];
+    it('creates a list of n-tuples from a list', function() {
+        assert.deepEqual(R.nwise(3, sevenLs), [[1, 2, 3], [2, 3, 4], [3, 4, 5], [4, 5, 6], [5, 6, 7]]);
+        assert.deepEqual(R.nwise(4, [1, 2, 3, 4]), [[1, 2, 3, 4]]);
+    });
+
+    it('returns an empty list when `n` > `list.length`', function() {
+        assert.deepEqual(R.nwise(6, [1, 2, 3]), []);
```
megawac: I'm not a big fan of omitting the ones outside of the chunk size. Its easier for the user to filter by length (i.e. remove the ones that don't fit the chunk size perfectly) than reconstruct the last set

buzzdecafe:
i think of `chunk` as different, viz:

```
chunk(2, [1, 2, 3, 4, 5]); //=> [[1, 2], [3, 4], [5]]
nwise(2, [1, 2, 3, 4, 5]); //=> [1, 2], [2, 3], [3, 4], [4, 5]]
```

test/nwise.js
buzzdecafe: for `chunk` I agree. for `nwise` that could only be an issue when the list length is less than the tuple size. in that case reconstructing the missing tuple is trivial, since it is the input list

kevinbeaty:
I immediately think of a moving average that could be implemented as a mean aggregate of each of these chunks... so maybe `movingWindow` or something like that?  There has to be a name for this...  maybe it's nwise :)

buzzdecafe:
i am willing to rename. `nwise` is probably better thought of as *all* `n`-tuples, rather than just consecutive ones. so open to pitches for new names.

buzzdecafe:
`consecutive`?
`sequent`?

kevinbeaty:
I like something like `(moving|sliding|rolling)(Window|Buffer)`  

It's almost the outer loop of a convolution (sliding a 1d kernel), I'm surprised I can't find a name for this...

src/nwise.js
```diff
@@ -0,0 +1,29 @@
...
+ * @param {Array} list The list to split into `n`-tuples
+ * @return {Array} The new list.
+ * @example
+ *
+ *      R.nwise(2, [1, 2, 3, 4, 5]); //=> [[1, 2], [2, 3], [3, 4], [4, 5]]
+ *      R.nwise(3, [1, 2, 3, 4, 5]); //=> [[1, 2, 3], [2, 3, 4], [3, 4, 5]]
+ *      R.nwise(7, [1, 2, 3, 4, 5]); //=> []
+ */
+module.exports = _curry2(function nwise(n, list) {
+    var i = -1;
```
davidchambers: s/i/idx/

Perhaps deedpoll was catching more style violations than I realized. ;)

buzzdecafe: :laughing: 

_mea culpa_

buzzdecafe:
`overlap`?

megawac:
Whats a use case for this - I thought it was chunk when I saw it

buzzdecafe:
this was inspired by the discussion on #872

buzzdecafe:
`aperture`?

CrossEye:
`pairwise` seems so obvious:

    R.pairwise(['a', 'b', 'c', 'd']); //=> [['a', 'b'], ['b', 'c'], ['c', 'd']]

that it really would be nice if the general name was an abstraction of that.  But I can't think of one.  `apeture` would be the best so far except that, like most suggestions here, it's a noun.  If we can't have a verb, I'd really rather have an adjective.  :smile:

kevinbeaty:
`dilate`, `regurgitate`, `exfoliate`? ;)

You could drop the `ing` to make a verb  `(move|slide|roll)(Window|Buffer)` but that's not nearly as fun.

Eh. Whatever works.

buzzdecafe:
`pairwise` should really be `[1, 2, 3, 4] => [[1,2], [2,3], [3,4], [1,3], [1,4], [2,4]]` so anything-`wise` ia not accurate.

`lens` also not a verb. let's not overconstrain ourselves. I'm for `aperture`.

CrossEye:
:+1: for aperture

buzzdecafe:
*ping*

CrossEye:
:ram:

PR #918, opened 2015-03-14
Functor method dispatch on reduce.

kevinbeaty:
`into` and `transduce` also allow dispatch on `reduce`. This
behavior is similar to Clojure with `IReduceInit` dispatching
and allows chaining applications of transducers with `eduction`.
Also allows using `into` and `transduce` with any object that
has a `reduce` method.

### DIFF
--- a/src/internal/_reduce.js
+++ b/src/internal/_reduce.js
@@ -1,4 +1,5 @@
 var _xwrap = require('./_xwrap');
+var bind = require('../bind');
 var isArrayLike = require('../isArrayLike');
 
 
@@ -28,6 +29,10 @@ module.exports = (function() {
         return xf.result(acc);
     }
 
+    function _methodReduce(xf, acc, obj) {
+        return xf.result(obj.reduce(bind(xf.step, xf), acc));
+    }
+
     var symIterator = (typeof Symbol !== 'undefined') ? Symbol.iterator : '@@iterator';
     return function _reduce(fn, acc, list) {
         if (typeof fn === 'function') {
@@ -35,6 +40,9 @@ module.exports = (function() {
         }
         if (isArrayLike(list)) {
             return _arrayReduce(fn, acc, list);
+        }
+        if (typeof list.reduce === 'function') {
+            return _methodReduce(fn, acc, list);
         }
         if (list[symIterator] != null) {
             return _iterableReduce(fn, acc, list[symIterator]());
--- a/src/transduce.js
+++ b/src/transduce.js
@@ -44,6 +44,5 @@ var curryN = require('./curryN');
  *      R.transduce(transducer, R.flip(R.append), [], numbers); //=> [2, 3]
  */
 module.exports = curryN(4, function(xf, fn, acc, list) {
-    return (typeof fn === 'function') ? _reduce(xf(_xwrap(fn)), acc, list)
-                                      : _reduce(xf(fn), acc, list);
+    return _reduce(xf(typeof fn === 'function' ? _xwrap(fn) : fn), acc, list);
 });
--- a/test/into.js
+++ b/test/into.js
@@ -1,4 +1,5 @@
 var assert = require('assert');
+var lodash = require('lodash');
 
 var R = require('..');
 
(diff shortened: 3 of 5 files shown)


kevinbeaty:
I realize I need to add tests for this, but I wanted to get your thoughts whether it was desired first.  It's a little strange since `transduce` and `into` actually dispatch to `reduce`, but, as I noted this is similar to Clojure and works with [eduction](http://clojure.org/transducers#toc5).  There is an implementation in JS [here](https://github.com/transduce/transduce/blob/master/core/eduction.js) which was discussed in [this PR](https://github.com/transduce/transduce/issues/6) from @tgriesser

kevinbeaty:
Note that `eduction` also depends on `sequence` and I can submit another PR for that if desired.  Example [here](https://github.com/transduce/transduce/blob/master/core/sequence.js)

src/transduce.js
```diff
@@ -44,6 +44,6 @@ var curryN = require('./curryN');
  *      R.transduce(transducer, R.flip(R.append), [], numbers); //=> [2, 3]
  */
 module.exports = curryN(4, function(xf, fn, acc, list) {
-    return (typeof fn === 'function') ? _reduce(xf(_xwrap(fn)), acc, list)
-                                      : _reduce(xf(fn), acc, list);
+    return (typeof fn === 'function') ? reduce(xf(_xwrap(fn)), acc, list)
+                                      : reduce(xf(fn), acc, list);
```
davidchambers: Slight simplification:

```javascript
return reduce(xf(typeof fn === 'function' ? _xwrap(fn) : fn), acc, list);
```

buzzdecafe:
:+1: for dispatching on `reduce`, i think we are going to want to work with Foldable/Traversable types

kevinbeaty:
Added tests and tweaked `transduce` per @davidchambers

test/into.js
```diff
@@ -29,6 +29,12 @@ describe('into', function() {
         assert.deepEqual(R.into({}, R.identity, [{a: 1}, {b: 2, c: 3}]), {a: 1, b: 2, c: 3});
     });
 
+    it('dispatches to objects that implement `reduce`', function() {
+        var obj = {x: [1, 2, 3], reduce: function(f, acc) { return R.reduce(f, acc, this.x); }};
+        assert.deepEqual(R.into([], R.map(add(1)), obj), [2, 3, 4]);
+        assert.deepEqual(R.into([], R.map(add(1)), obj), [2, 3, 4]);
+    });
+
```
buzzdecafe: would probably be good to test one that does not call `R.reduce` but some other `reduce`

test/reduce.js
```diff
@@ -12,6 +12,12 @@ describe('reduce', function() {
         assert.strictEqual(R.reduce(mult, 1, [1, 2, 3, 4]), 24);
     });
 
+    it('dispatches to objects that implement `reduce`', function() {
+        var obj = {x: [1, 2, 3], reduce: function(f, acc) { return R.reduce(f, acc, this.x); }};
+        assert.strictEqual(R.reduce(add, 0, obj), 6);
+        assert.strictEqual(R.reduce(add, 10, obj), 16);
+    });
+
```
buzzdecafe: likewise here

test/into.js
kevinbeaty: Do you mean adding a `reduceCalled` flag in the `obj.reduce` method or using a different `reduce` or something else?  The method itself must be called here, (how else would it get access to the list `[1, 2, 3]`?)

kevinbeaty: Noticed `lodash` is a `devDependency`, could call that within the `method`.

kevinbeaty: So I did this and it surfaced a problem with this implementation.  Good call. I'll expand below in a normal comment.

buzzdecafe: thinking something along the lines of:

`R.into([], R.identity, {reduce: function() { return 'dispatched'; }}) //=> ???`

kevinbeaty:
I think we may want to use a different dispatch then `_checkForMethod` to force a reducing function instead of a transformer when calling the method.

 `transduce` will wrap a reducing function with a transformer, initialize the transformer and pass to `reduce`.  If third party libraries have a `reduce` method, I would doubt they would handle a transformer and expect a reducing function instead.  I do this in my `transduce` library to handle this case:

```javascript
function methodReduce(xf, init, coll){
  var result = coll.reduce(function(result, value){
    return xf.step(result, value)
  }, init)
  return xf.result(result)
}
```
Could do something similar here.

kevinbeaty:
Pushed a new version that dispatches to methods within internal `_reduce`

buzzdecafe:
:cow2:

src/internal/_reduce.js
```diff
@@ -28,6 +29,11 @@ module.exports = (function() {
         return xf.result(acc);
     }
 
+    function _methodReduce(xf, acc, obj) {
+        acc = obj.reduce(bind(xf.step, xf), acc);
+        return xf.result(acc);
```
davidchambers: Nit: `acc` could be inlined.

kevinbeaty:
Inline `acc` in `_methodReduce

davidchambers:
:deciduous_tree: Lovely work, @kevinbeaty.

PR #919, opened 2015-03-14
make R.invoke a fixed-arity function

davidchambers:
I take exception to this expression:

```javascript
R.invoke('substring', 'Ramda', 1, 4)
```

We're providing three things: a method name, an object, and zero or more arguments (two in this case). Though we're providing three things, we're providing *four* arguments to `R.invoke`. The expression would be clearer if the arguments we wish to provide to `'Ramda'.substring` were grouped:

```javascript
R.invoke('substring', 'Ramda', [1, 4])
```

This would make the function easier to describe:

<p><del>Returns the result of invoking <code>obj[methodName]</code> with the zero or more positional arguments following <code>methodName</code> and <code>obj</code>.</del></p>

<p><ins>Returns the result of applying <code>obj[methodName]</code> to <code>args</code>.</ins></p>

It would also simplify the function's `@sig`. Perhaps most significantly, it would mean we could partially apply any combination of arguments. For example:

```javascript
//  toBinary :: Number -> String
var toBinary = R.invoke('toString', R.__, [2])

toBinary(42);  // => '101010'
toBinary(63);  // => '111111'
```

I can also imagine contexts in which partially applying the first and second arguments is useful.

This is not a breaking change since `R.invoke` was added in #906 and we have not published since that pull request was merged.

### DIFF
--- a/src/invoke.js
+++ b/src/invoke.js
@@ -1,24 +1,25 @@
-var _slice = require('./internal/_slice');
 var curry = require('./curry');
 
 
 /**
- * Returns the result of invoking `obj[methodName]` with the zero or more
- * positional arguments following `methodName` and `obj`.
+ * Returns the result of applying `obj[methodName]` to `args`.
  *
  * @func
  * @memberOf R
  * @category Object
- * @sig String -> Object -> *
+ * @sig String -> Object -> [*] -> *
  * @param {String} methodName
  * @param {Object} obj
+ * @param {Array} args
  * @return {*}
  * @example
  *
- *      R.invoke('add', R, 1, 2); //=> 3
+ *      //  toBinary :: Number -> String
+ *      var toBinary = R.invoke('toString', R.__, [2])
  *
- *      R.invoke('f', {f: function() { return 'f called'; }}); //=> 'f called'
+ *      toBinary(42); //=> '101010'
+ *      toBinary(63); //=> '111111'
  */
-module.exports = curry(function invoke(methodName, obj) {
-    return obj[methodName].apply(obj, _slice(arguments, 2));
+module.exports = curry(function invoke(methodName, obj, args) {
+    return obj[methodName].apply(obj, args);
 });
--- a/test/invoke.js
+++ b/test/invoke.js
@@ -13,17 +13,17 @@ describe('invoke', function() {
         }};
         var gName = R.invoke('getName');
         assert.strictEqual(typeof gName, 'function');
-        assert.strictEqual(gName(fred), 'Fred Flintstone');
-        assert.strictEqual(gName(barney), 'Barney Rubble');
+        assert.strictEqual(gName(fred, []), 'Fred Flintstone');
+        assert.strictEqual(gName(barney, []), 'Barney Rubble');
     });
 
     it('passes arguments appropriately when not curried', function() {
-        assert.strictEqual(R.invoke('add', R, 3, 6), 9);
+        assert.strictEqual(R.invoke('add', R, [3, 6]), 9);
     });
 
     it('invokes the function with no arguments when no extra params are supplied', function() {
         var obj = {f: function() { return 'called f'; }};
-        assert.strictEqual(R.invoke('f', obj), 'called f');
+        assert.strictEqual(R.invoke('f', obj, []), 'called f');
     });
 
     it('applies additional arguments to the function', function() {
@@ -39,7 +39,7 @@ describe('invoke', function() {
 
 
         var moveBy = R.invoke('moveBy');
-        moveBy(p1, 5, 7);
+        moveBy(p1, [5, 7]);
         assert.strictEqual(p1.x, 15);
         assert.strictEqual(p1.y, 27);
     });


CrossEye:
While this is not technically a breaking change, I guess, it is in spirit, as this function is meant to be the successor to `func`.

I like the change, though.  I don't expect this will cause too much pain, and if we can remove another variadic function from Ramda, I'm all for it.  :herb:

buzzdecafe:
well that was easy. this *does require* a list of args to get passed in. that makes me a *bit* apprehensive that this may start creeping in other places that are at present accepting varargs. but *maybe* that is *good* -- so :cow2:

davidchambers:
The second and third arguments should be switched. I've been upgrading various projects to `ramda@0.12.x` and so far every occurrence of `R.func` uses `R.__` as the second argument as in:

> ```javascript
> //  toBinary :: Number -> String
> var toBinary = R.invoke('toString', R.__, [2])
>
> toBinary(42);  // => '101010'
> toBinary(63);  // => '111111'
> ```

It seems according to our "data last" philosophy the object should come last.

buzzdecafe:
> It seems according to our "data last" philosophy the object should come last.

IMO "data last" is really a side-effect of arranging the arguments to ease composition. So if this order is more composable, i'm for it

CrossEye:
> IMO "data last" is really a side-effect of arranging the arguments to ease composition. So if this order is more composable, i'm for it

Agreed.  I like to describe it as ordering the parameters from those least likely to change to those most likely to change.  And again that fits with this suggestion.  I'll add a warning to the 0.12 upgrade guide too.

paldepind:
I thought one should use `R.invoker` if one wanted to supply the target object last.

Doesn't this argument order change make `R.invoke` awfully similar to `R.invoker`? Now both `invoke` and `invoker` has the same order: method name, arguments and target object. The only difference is that `invoker` also needs an arity and that it's variadic.

```javascript
var toBinary1 = R.invoker(1, 'toString', 2)
var toBinary2 = R.invoke('toString', [2])

var sliceFrom1 = R.invoker(1, 'slice');
sliceFrom1(6, 'abcdefghijklm'); //=> 'ghijklm'
var sliceFrom2 = R.invoke('slice')
sliceFrom2([6], 'abcdefghijklm') //=> 'ghijklm'
```

If `invoker` was made non-variadic they'd be even more alike.

davidchambers:
Is there a compelling reason to have both `R.invoke` and `R.invoker`? Perhaps we should deprecate the latter.

buzzdecafe:
no reason i know of. :+1: for deprecating `invoker`

CrossEye:
Agreed.  This has been a  long, strange journey, though.

This was the original version of `invoker`:

    var invoker = R.invoker = function _invoker(name, obj, len) {
        var method = obj[name];
        var length = len === void 0 ? method.length : len;
        return method && curryN(length + 1, function() {
            if (arguments.length) {
                var target = Array.prototype.pop.call(arguments);
                var targetMethod = target[name];
                if (targetMethod == method) {
                    return targetMethod.apply(target, arguments);
                }
            }
        });
    };

This is the current version of `invoke`: 

    curry(function invoke(methodName, args, obj) {
      return obj[methodName].apply(obj, args);
    });

I'm not really commenting on their relative complexities, although that's interesting, but on _how much less_ the new function does.  And that's a very good thing.

:+1:

paldepind:
To me `invoker` seems like the most useful of the two since it can be used to create curried functions. It could easily be made invariant by removing the support for supplying initial arguments initially (which is not a great loss since the resulting function is curried anyway).

CrossEye:
@paldepind: That's a very good point.  I've been uncomfortable with something that's been happening for a while; the move to avoid all variadic signatures is the latest sign of it.  We seem to be moving away from part of our mantra of being a library "for Javascript developers."

While `invoke` certainly adds some consistency, `invoker` creates functions more familiar to -- and arguably more useful to -- Javascript developers.

So I guess I'm switching to a :-1: .

paldepind:
Where is this going? I think we could:

1. Make `invoker` invariant by removing the support for supplying arguments initially (which is not a great loss since the resulting function is curried anyway).
2. Preserve `invoker` because creating curried functions is more powerful. Deprecate and remove `invoke`.

Or:

1. Preserve `invoke` because `invoker` does not work well with variadic methods.
2. Users will cumbersomely get the power of `invoker` with `curryN` and `invoke`.

Or:

1. Deprecate `invoker`
2. Implement a new function `invokeN` which could be a one-liner based on `invoke` and `curryN`.

I guess this is another instance of the the "should functions created by Ramda be curried, should they not be curried or should we do both"-issue.

CrossEye:
Sorry, I missed this a few days ago.  And I don't have an answer now.  I want to think about it a little more now that it's been brought back to our attention.

You're right, though.  This is part of a larger question about currying functions returned from Ramda.

buzzdecafe:
> "should functions created by Ramda be curried, should they not be curried or should we do both"

"functions created by Ramda should be curried" :heavy_check_mark:

CrossEye:
... and those who have functions with named optional parameters should just lump it. :interrobang:

buzzdecafe:
"and those who have functions with named optional parameters should just lump it." :heavy_check_mark:

davidchambers:
> [T]hose who have functions with named optional parameters should just lump it.

Yes, I think so. Rather than:

```javascript
function euclideanDistance(p1, p2) {
  return arguments.length < 2 ?
    euclideanDistance(p1, new Point(0, 0)) :
    Math.sqrt(square(Math.abs(p1.x - p2.x)) +
              square(Math.abs(p1.y - p2.y)));
}
```

we should encourage:

```javascript
var euclideanDistance = R.curry(function(p1, p2) {
  return Math.sqrt(square(Math.abs(p1.x - p2.x)) +
                   square(Math.abs(p1.y - p2.y)));
});

var euclideanDistanceFromOrigin = euclideanDistance(new Point(0, 0));
```

In other words, the philosophy of Ramda should extend beyond the bounds of Ramda itself. Those who wish to use Ramda in conjunction with libraries with different philosophies may need to write wrapper functions to bridge the two worlds. I believe this price worth paying to make Ramda even better when used as suggested.

CrossEye:
I think I'll probably get there.  But when I think about the kinds of things that could fail and how difficult they might be to track down, I get a bit worried.  ("Wait, `addEventListener` has **three** parameters?"  "What do you mean `parseInt` needs a radix?  What the hell is a radix, and why should I care?"  And who are you to tell me that my function is broken because I switched from `parseFloat` to `parseInt` when I don't give a damn about the stupid radix?")  I'm wondering if we need some sort of `uncurry` capability to help out in these cases, although I don't really know what it would look like.  Perhaps @megawac's ideas on metadata would be useful.

-----

(BTW, `Math.abs` is no needed in your `euclideanDistance` examples.)

PR #920, opened 2015-03-14
Reduce over object [key, value] as last option

kevinbeaty:
Iterates over own property `[key, value]` pairs if not array, method
or iterable. Used as last option instead of throwing error. Dual to
`_stepCat` merging of pairs with transducers.


kevinbeaty:
This one may be controversial based on caveat from @buzzdecafe on original transducers PR, but I think it is useful to iterate own properties as pairs as a last resort in `reduce`.  It is a nice dual to the `[key, value]` merging in `stepCat` when using `into({})`.  When merging into objects, the unspecified iteration order isn't an issue.

buzzdecafe:
right. the probem is that iteration order is not guaranteed, so logically equivalent objects could have different outputs. I pitched sort-and-document for `foldObj` but it didn't make it in. The sort would have to happen *up front*, blowing away the notion of laziness. Anyways, it's always a conversation starter! here we go again ...

algesten:
This really needs a sort on the keys, or the behaviour will be undefined for non-commutative functions.

algesten:
```javascript
function _objectReduce(xf, acc, obj) {
    Object.keys(obj).sort().forEach(function(k) {
...
```
or equivalent with a `for`-loop instead of `forEach`

kevinbeaty:
> Anyways, it's always a conversation starter! here we go again ...
Nice.  I'll let it play out.   I can certainly sort the keys up front if everyone agrees that is better.  My motivation is to support `into({}, transducer, {...})`, so I'm impartial.

src/internal/_reduce.js
```diff
@@ -33,6 +34,20 @@ module.exports = (function() {
         return xf.result(obj.reduce(bind(xf.step, xf), acc));
     }
 
+    function _objectReduce(xf, acc, obj) {
+        var key;
+        for (key in obj) {
```
buzzdecafe: sort keys here?

buzzdecafe: and there should be a test for `obj1 = {a: 1, b: 2}, obj2 = {b: 2, a: 1}` and a non-commutative function and prove that the outputs are the same, e.g. `function(a, b) { return a - b; }`

kevinbeaty:
Previous discussions for reference:  #257 #625 #656

kevinbeaty:
I don't have incredibly strong feeling about this one, so if we think it's been talked to death already, I have no heartache closing it.

kevinbeaty:
Another option would be to convert the Object to an iterable and just reuse `_iterableReduce`.  This iterable could also be reused for `sequence(transducer, list)`, a lazy iterable transformation function.  I considered doing that for this PR but thought it would cost too much in terms of performance and be controversial from that regard. But if we need to sort the keys anyway, that argument is moot.

It would look something like this:

```javascript
function ObjectIterable(obj){
  this.obj = obj
  this.keys = R.keys(obj).sort()
}
ObjectIterable.prototype[symbol] = function(){
  var obj = this.obj,
    keys = this.keys,
    idx = 0
  return {
    next: function(){
      if(idx >= keys.length){
        return {done: true}
      }
      var key = keys[idx++]
      return {done: false, value: [key, obj[key]]}
    }
  }
}
```

buzzdecafe:
`iterator` impl LGTM

kevinbeaty:
Pushed a version using `_iterable` to iterate on sorted object keys.

kevinbeaty:
I still need to add a test for non-commutative functions...

test/reduce.js
```diff
@@ -13,6 +14,16 @@ describe('reduce', function() {
...
+    it('folds functions over iterables with the supplied accumulator', function() {
+        assert.strictEqual(R.reduce(add, 0, iterdone.range(1, 5)), 10);
+        assert.strictEqual(R.reduce(mult, 1, iterdone.range(1, 5)), 24);
+    });
+
+    it('folds functions over duck typed iterators with the supplied accumulator', function() {
+        assert.strictEqual(R.reduce(add, 0, iterdone.range(1, 5)[iterdone.symbol]()), 10);
+        assert.strictEqual(R.reduce(mult, 1, iterdone.range(1, 5)[iterdone.symbol]()), 24);
+    });
+
```
buzzdecafe: needs a test with logically equivalent objects and non-commutative function, but results should be equal.

otherwise i like this -- more general than my `foldObj`, should have consistent behavior.

kevinbeaty:
@buzzdecafe Stole your tests from your previous `foldObj` proposal

test/reduce.js
```diff
@@ -7,18 +8,45 @@ var R = require('..');
...
         var obj = {x: [1, 2, 3], reduce: function(f, acc) { return lodash.reduce(this.x, f, acc); }};
         assert.strictEqual(R.reduce(add, 0, obj), 6);
         assert.strictEqual(R.reduce(add, 10, obj), 16);
     });
 
+    it('folds an object in sorted-key order', function() {
+        assert.strictEqual(R.reduce(cat, '', objA), ' uc X uc Y uc Z lc x lc y lc z');
+        assert.strictEqual(R.reduce(cat, '', objA), R.reduce(cat, '', objB));
+        assert.strictEqual(R.reduce(cat, '', objA), R.reduce(cat, '', objC));
+    });
```
buzzdecafe: :smile:

buzzdecafe:
> Stole your tests from your previous `foldObj` proposal

expect a letter from my attorney.

this LGTM. awaiting others input

megawac:
What's the justification of implementing object iterations through transducers, as objects tend to be significantly smaller (and a less common case) than arrays. I feel like the extra code may not be worthwhile

src/internal/_symbolIterator.js
```diff
@@ -0,0 +1 @@
+module.exports = (typeof Symbol !== 'undefined') ? Symbol.iterator : '@@iterator';
```
megawac: do all symbol implementation s expose iterator? I know a lot of them don't include the es6 symbol.match yet

kevinbeaty: I believe so, see [here](http://kangax.github.io/compat-table/es6/#Symbol).  Although it probably wouldn't hurt to add another check:

```javascript
module.exports = typeof Symbol !== 'undefined' && typeof Symbol.iterator !== 'undefined' ? Symbol.iterator : '@@iterator';
```

kevinbeaty:
> What's the justification of implementing object iterations

Convenience mostly.  Really the same argument as the justification for `foldObj`, `mapObj` and friends but without the separation.

> through transducers

Once you have the object iteration in `reduce`, you get any operation implemented as a transducer.  I think @buzzdecafe is interesting in expanding the number of functions that dispatch to transducers (I plan on making additional PRs later). So you pay the cost once in `reduce` and get the rest for free.  Initially even the array versions delegated to `reduce` using transducers, but that was removed do to performance considerations.

>  I feel like the extra code may not be worthwhile

It may not be.  Again, I don't think this PR is critical.  I rarely use object iteration, but it is nice to have when I do.  The extra code is a one time cost and the addition of `_iterable` is a step to some interesting higher level functions that I would like to discuss in future PRs (including `sequence` and `eduction`).

CrossEye:
I still have the same objections as from all the linked issues.  (We also discussed this today in the [Gitter room][gr].)  Basically the only functions that work consistently are commutative, associative ones.  Anything else can fall prey to key order issues.  And that worries me.  jdalton insists that I'm putting "imaginary things like this over user experience", but I would like to stick with the requirement that Ramda functions do things in clear ways, in consistent ways, and as simply as we can.

  [gr]: https://gitter.im/ramda/ramda

buzzdecafe:
> Anything else can fall prey to key order issues.

Did you look at the PR?

kevinbeaty:
Well, so I finally reviewed the previous threads and the recent chatter in the Gitter room.   I think I my opinions fall closer to @jdalton on this one, but I do understand the argument for association/commutativity (not sure if I agree, but that's beside the point).  This PR does now sort the keys to address that concern.

I can close the PR if we are just going to rehash the same arguments. I'm still new to Ramda, I wasn't trying to rock the boat :)

 My initial use case is just sugar for:

```javascript
var result = R.into({}, transducer, R.toPairs({....}))

// or with iterable dispatch in reduce
var iterable  = require('iterdone').iterable // or require('transduce/core/iterable');
var result = R.into({}, transducer, iterable({...}))
```

As far as eduction/sequence/iterable, that may be better to keep in a [separate library](https://github.com/transduce/transduce) anyway.  Everything in Ramda is compatible with everything in transduce, transducer.js, transducer-js so things can be used interchangeably.

buzzdecafe:
at the risk of going around the mulberry bush again, sorting the keys is IMO a reasonable solution.

CrossEye:
It might be a mulberry bush.  I don't remember why we dropped the sorting discussions earlier.  It could give us at least _an_ answer to those who look for `foldObj`: "Yes, but you may not like it."

```javascript
reduce(difference, {start: 10, end: 50}); //=> -40 // WTF? It's 40 in Underscore!
```

This would still be better in my mind than

```javascript
reduce(difference, {start: 10, end: 50}); //=> -40
reduce(difference, {end: 50, start: 10}); //=> 40
```

But either is likely confusing to users.

buzzdecafe:
i think what it boils down to is: Is *a* solution better than *no* solution? Because a *comprehensive and consistent* solution appears to be impossible. IMO a consistent solution where we document the behavior is better than nothing. If what the user wants is beyond the scope of that solution, they are on their own.

CrossEye:
>> Anything else can fall prey to key order issues.

> Did you look at the PR?

Yes, the problem is not that we cannot define a consistent key order.  Obviously we can do that, but there's a very good chance that this won't meet users' expectations.

It seems to me that there is a fundamental contradiction here in adding a function -- not because of underlying necessity but because a number of users are looking for it -- that we can only make lawful by choosing to sidestep what the users expect.

But there's also a fundamental disconnect in my mind between the notion of `reduce`, which dictates an ordered set of operations, and an object, which has no inherent order to it.  I feel as though we have to shave off too many corners of that square peg.

I guess it still feels to me as though this belongs in user-land.  It's not as though it's difficult, especially for the non-indexed version:

```javascript
var foldObj = R.useWith(R.reduce, R.identity, R.identity, R.values);
foldObj(R.add, 0, {a: 1, b: 2, c: 3}); //=> 6
```

(Man, I do miss the `R.I` alias here!)

PR #922, opened 2015-03-15
anti-sauce

buzzdecafe:
sauce labs is running after each platform build and if it errors -- not *fails* -- on any one, the whole build gets marked as errored. I propose running sauce either pre-publish or at regular intervals.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,15 +3,7 @@ node_js:
 - '0.10'
 - '0.12'
 - 'io.js'
-env:
-  global:
-  - secure: dgshxCQaY6olKOZnD+BT8R1Vp8uJhwCIUlSnoFji1m+7KIy3zwbbgi/2jkf/SWBCxzlpzDBWUuWbaMJOrJHWfjUaX3JfUY+2gbAPunrofD+42+brfo506rIxr20yaoomsnLxXFLgleLW5FuaezP6KP4lW5IOlWZXOEv0vTvoxQk=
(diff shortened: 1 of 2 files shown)


CrossEye:
:herb:

I wonder if Travis allows any other triggers than check-in.  A daily build for this sounds perfect.

buzzdecafe:
i think this would have to be outside the context of travis, since travis will always do what is configured in .travis.yml. but the idea of adding more infrastructure is making me feel sick

.travis.yml
```diff
@@ -14,4 +14,4 @@ script:
 - make test lint
 after_success:
 - grunt bench
-- grunt sauce
+#- grunt sauce
```
davidchambers: Let's delete this line. We can always consult the file's history to see earlier versions.

davidchambers:
:deciduous_tree: No more 20-minute builds? Sounds good to me!

buzzdecafe:
should probably do the same for bench tests

buzzdecafe:
1798.1 	1 min 36 sec 	less than a minute ago 	0.10 	linux
1798.2 	1 min 31 sec 	less than a minute ago 	0.12 	linux
1798.3 	1 min 45 sec 	less than a minute ago 	io.js 	linux

megawac:
its not too hard to run sauce on only one step of the matrix (see underscore).an alternative is to use twitters tool to run the sauce tests and comment on the relevant PR or commit

CrossEye:
@megawac: That bit from Underscore looks very useful, and might also help with [another issue][r899c].  I will look into that soon.

  [r899c]: https://github.com/ramda/ramda/pull/899#issuecomment-77824113

PR #923, opened 2015-03-15
avoid recursing once more than necessary in XTake.prototype.step

### LINKED ISSUE #921: take allow to do one extra step
Hi,

I noticed strange behavior of the `take` function. It returns right result, but it doesn't end reduce immediately, so one extra step is executed:

```
var range = function(from, to) {
  return function*() {
    for (var i = from; i <= to; i++) { console.log('yield', i); yield i }; 
  }();
};

console.log(R.transduce(R.take(2), R.flip(R.append), [])(range(10, 2000000)));
```

```
yield 10
yield 11
yield 12 <--- this yield shouldn't be executed
[10, 11]
```

davidchambers:
Fixes #921

We need to decide how to include tests involving ES6 features such as generators. Suggestions as to other ways to write a regression test for #921 much appreciated.

### DIFF
--- a/src/internal/_xtake.js
+++ b/src/internal/_xtake.js
@@ -15,7 +15,8 @@ module.exports = (function() {
     };
     XTake.prototype.step = function(result, input) {
         this.n -= 1;
-        return this.n >= 0 ? this.xf.step(result, input) : _reduced(result);
+        return this.n === 0 ? _reduced(this.xf.step(result, input))
+                            : this.xf.step(result, input);
     };
 
     return _curry2(function _xtake(n, xf) { return new XTake(n, xf); });


src/internal/_xtake.js
```diff
@@ -14,8 +14,16 @@ module.exports = (function() {
         return this.xf.result(result);
     };
     XTake.prototype.step = function(result, input) {
-        this.n -= 1;
-        return this.n >= 0 ? this.xf.step(result, input) : _reduced(result);
+        switch (this.n) {
+            case 0:
+                return _reduced(result);
```
davidchambers: Is this case necessary, @kevinbeaty? Will the transducer machinery ever invoke `step` if `this.n` is zero?

kevinbeaty: It should not be.  Calling `step` after a `reduced` would be a bug in the [transducible process](http://clojure.org/transducers#toc11)

davidchambers: Excellent! I'll update the pull request. :)

kevinbeaty:
> We need to decide how to include tests involving ES6 features such as generators. 

Generators are just iterators, so you can use add a dependency on a library that implements iterators without generators as a `devDependency`.   I noticed we didn't have any tests for iterators or iterables so I added a few in (now dead) #920 based on [iterdone](https://github.com/kevinbeaty/iterdone). 

 If you really want generators you can probably come up with a way to separate those tests in the Travis matrix to run only those tests with a `node.js --harmony` and io.js

src/internal/_xtake.js
davidchambers: Updated. The diff is now easier to follow. The final step is now:

```javascript
_reduced(this.xf.step(result, input))
```

whereas previously it was:

```javascript
_reduced(result)
```

which ignored the final `input`.

svozza:
It's easy enough to run a specific test file only on a particular version of Node using Travis, we looked into it for Highland:

https://github.com/caolan/highland/issues/131#issuecomment-75250994

But actually all we ended up doing was using feature detection to run specific tests.

buzzdecafe:
:cow2:

CrossEye:
:herb:

PR #924, opened 2015-03-15
add R.toString

### LINKED ISSUE #875: "show" function
I'd like Ramda to provide a function from value to string representation. Ideally the output could be pasted into the REPL to produce a value equivalent to the input value.

```javascript
> R.show(42)
'42'
> R.show('42')
'"42"'
> R.show([1, 2, 3])
'[1, 2, 3]'
> R.show({x: 0, y: 0, 'foo-bar': true})
'{"foo-bar": true, x: 0, y: 0}'
> R.show(null)
'null'
```

Types could define `show`:

```javascript
function Nothing() {
  if (!(this instanceof Nothing)) {
    return new Nothing();
  }
}
Nothing.prototype.show = R.always('Nothing()');

function Just(value) {
  if (!(this instanceof Just)) {
    return new Just(value);
  }
  this.value = value;
}
Just.prototype.show = function() { return 'Just(' + R.show(this.value) + ')'; };

R.show([Nothing(), Just([1, 2, 3])]);  // => '[Nothing(), Just([1, 2, 3])]'
```

davidchambers:
Closes #875

I'm very excited about this function! I've written similar functions several times but never with such a clear goal in mind:

> `eval`'ing the output should result in a value equivalent to the input value.

Without this it wouldn't have been clear what to return for `-0`, `new Boolean(true)`, and many other values.

The implementation is straightforward, which also pleases me. :)

### DIFF
--- a/src/internal/_toString.js
+++ b/src/internal/_toString.js
@@ -1,0 +1,37 @@
+var _indexOf = require('./_indexOf');
+var _map = require('./_map');
+var _quote = require('./_quote');
+var _toISOString = require('./_toISOString');
+var keys = require('../keys');
+
+
+module.exports = function _toString(x, seen) {
+  var recur = function recur(y) {
+    var xs = seen.concat([x]);
+    return _indexOf(xs, y) >= 0 ? '<Circular>' : _toString(y, xs);
+  };
+
+  switch (Object.prototype.toString.call(x)) {
+    case '[object Arguments]':
+      return '(function() { return arguments; }(' + _map(recur, x).join(', ') + '))';
+    case '[object Array]':
+      return '[' + _map(recur, x).join(', ') + ']';
+    case '[object Boolean]':
+      return typeof x === 'object' ? 'new Boolean(' + recur(x.valueOf()) + ')' : x.toString();
+    case '[object Date]':
+      return 'new Date(' + _quote(_toISOString(x)) + ')';
+    case '[object Null]':
+      return 'null';
+    case '[object Number]':
+      return typeof x === 'object' ? 'new Number(' + recur(x.valueOf()) + ')' : 1 / x === -Infinity ? '-0' : x.toString(10);
+    case '[object String]':
+      return typeof x === 'object' ? 'new String(' + recur(x.valueOf()) + ')' : _quote(x);
+    case '[object Undefined]':
+      return 'undefined';
+    default:
+      return (typeof x.constructor === 'function' && x.constructor.name !== 'Object' &&
+              typeof x.toString === 'function' && x.toString() !== '[object Object]') ?
+             x.toString() :  // Function, RegExp, user-defined types
+             '{' + _map(function(k) { return _quote(k) + ': ' + recur(x[k]); }, keys(x).sort()).join(', ') + '}';
+  }
+};
--- a/test/toString.js
+++ b/test/toString.js
@@ -1,0 +1,166 @@
+var assert = require('assert');
+
+var R = require('..');
+
+
+describe('toString', function() {
+
+  it('returns the string representation of null', function() {
+    assert.strictEqual(R.toString(null), 'null');
+  });
+
+  it('returns the string representation of undefined', function() {
+    assert.strictEqual(R.toString(undefined), 'undefined');
+  });
+
+  it('returns the string representation of a Boolean primitive', function() {
+    assert.strictEqual(R.toString(true), 'true');
+    assert.strictEqual(R.toString(false), 'false');
+  });
+
+  it('returns the string representation of a number primitive', function() {
+    assert.strictEqual(R.toString(0), '0');
+    assert.strictEqual(R.toString(-0), '-0');
+    assert.strictEqual(R.toString(1.23), '1.23');
+    assert.strictEqual(R.toString(-1.23), '-1.23');
+    assert.strictEqual(R.toString(1e+23), '1e+23');
+    assert.strictEqual(R.toString(-1e+23), '-1e+23');
+    assert.strictEqual(R.toString(1e-23), '1e-23');
+    assert.strictEqual(R.toString(-1e-23), '-1e-23');
+    assert.strictEqual(R.toString(Infinity), 'Infinity');
+    assert.strictEqual(R.toString(-Infinity), '-Infinity');
+    assert.strictEqual(R.toString(NaN), 'NaN');
+  });
+
+  it('returns the string representation of a string primitive', function() {
+    assert.strictEqual(R.toString('abc'), '"abc"');
+    assert.strictEqual(R.toString('x "y" z'), '"x \\"y\\" z"');
+  });
+
+  it('returns the string representation of a Boolean object', function() {
+    /* jshint -W053 */
+    assert.strictEqual(R.toString(new Boolean(true)), 'new Boolean(true)');
+    assert.strictEqual(R.toString(new Boolean(false)), 'new Boolean(false)');
+    /* jshint +W053 */
+  });
+
+  it('returns the string representation of a Number object', function() {
+    /* jshint -W053 */
+    assert.strictEqual(R.toString(new Number(0)), 'new Number(0)');
+    assert.strictEqual(R.toString(new Number(-0)), 'new Number(-0)');
+    /* jshint +W053 */
+  });
+
+  it('returns the string representation of a String object', function() {
+    /* jshint -W053 */
+    assert.strictEqual(R.toString(new String('abc')), 'new String("abc")');
+    assert.strictEqual(R.toString(new String('x "y" z')), 'new String("x \\"y\\" z")');
+    /* jshint +W053 */
+  });
+
+  it('returns the string representation of a Date object', function() {
+    assert.strictEqual(R.toString(new Date('2001-02-03T04:05:06.000Z')), 'new Date("2001-02-03T04:05:06.000Z")');
+  });
+
+  it('returns the string representation of a RegExp object', function() {
+    assert.strictEqual(R.toString(/(?:)/), '/(?:)/');
+    assert.strictEqual(R.toString(/\//g), '/\\//g');
+  });
+
+  it('returns the string representation of a function', function() {
+    assert.strictEqual(R.toString(function add(a, b) { return a + b; }), 'function add(a, b) { return a + b; }');
+  });
+
+  it('returns the string representation of an array', function() {
+    assert.strictEqual(R.toString([]), '[]');
+    assert.strictEqual(R.toString([1, 2, 3]), '[1, 2, 3]');
+    assert.strictEqual(R.toString([1, [2, [3]]]), '[1, [2, [3]]]');
+    assert.strictEqual(R.toString(['x', 'y']), '["x", "y"]');
+  });
+
+  it('returns the string representation of an arguments object', function() {
+    assert.strictEqual(R.toString((function() { return arguments; }())), '(function() { return arguments; }())');
+    assert.strictEqual(R.toString((function() { return arguments; }(1, 2, 3))), '(function() { return arguments; }(1, 2, 3))');
+    assert.strictEqual(R.toString((function() { return arguments; }(['x', 'y']))), '(function() { return arguments; }(["x", "y"]))');
+  });
+
+  it('returns the string representation of a plain object', function() {
+    assert.strictEqual(R.toString({}), '{}');
+    assert.strictEqual(R.toString({foo: 1, bar: 2, baz: 3}), '{"bar": 2, "baz": 3, "foo": 1}');
+    assert.strictEqual(R.toString({'"quoted"': true}), '{"\\"quoted\\"": true}');
+    assert.strictEqual(R.toString({a: {b: {c: {}}}}), '{"a": {"b": {"c": {}}}}');
+  });
+
+  it('treats instance without custom `toString` method as plain object', function() {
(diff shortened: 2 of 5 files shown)


test/toString.js
```diff
@@ -0,0 +1,155 @@
...
+
+    it('returns the string representation of a String object', function() {
+        /* jshint -W053 */
+        assert.strictEqual(R.toString(new String('abc')), 'new String("abc")');
+        assert.strictEqual(R.toString(new String('x "y" z')), 'new String("x \\"y\\" z")');
+        /* jshint +W053 */
+    });
+
+    it('returns the string representation of a Date object', function() {
+        assert.strictEqual(R.toString(new Date('2001-02-03T04:05:06.000Z')), 'new Date("2001-02-03T04:05:06.000Z")');
```
buzzdecafe: only IE >= 9?

megawac:
-1 - I feel this belongs better in third party packages such as https://github.com/thejameskyle/pretty-format

test/toString.js
davidchambers: I've added a polyfill for `Date.prototype.toISOString`.

davidchambers:
I agree that this could be a stand-alone package. I don't think pretty-format is for me, though:

```
> var prettyFormat = require('pretty-format')
undefined
> prettyFormat({})
TypeError: Object function Object() { [native code] } has no method 'getOwnPropertySymbols'
    at Type.Object.Type.print (/private/tmp/node_modules/pretty-format/dist/pretty-format.js:257:33)
    at Type.print (/private/tmp/node_modules/pretty-format/dist/pretty-format.js:245:23)
    at print (/private/tmp/node_modules/pretty-format/dist/pretty-format.js:412:10)
    at repl:1:2
    at REPLServer.self.eval (repl.js:110:21)
    at Interface.<anonymous> (repl.js:239:12)
    at Interface.emit (events.js:95:17)
    at Interface._onLine (readline.js:202:10)
    at Interface._line (readline.js:531:8)
    at Interface._ttyWrite (readline.js:760:14)
> prettyFormat(-0)
'0'
> prettyFormat(new Date('2001-02-03T04:05:06Z'))
TypeError: Cannot call method 'print' of undefined
    at print (/private/tmp/node_modules/pretty-format/dist/pretty-format.js:412:10)
    at repl:1:2
    at REPLServer.self.eval (repl.js:110:21)
    at Interface.<anonymous> (repl.js:239:12)
    at Interface.emit (events.js:95:17)
    at Interface._onLine (readline.js:202:10)
    at Interface._line (readline.js:531:8)
    at Interface._ttyWrite (readline.js:760:14)
    at ReadStream.onkeypress (readline.js:99:10)
    at ReadStream.emit (events.js:98:17)
```

Perhaps it's ES6+?

megawac:
I really don't think this belongs in ramda core. Easy enough to fix in other implementations
https://github.com/thejameskyle/pretty-format/pull/1
https://github.com/thejameskyle/pretty-format/pull/2

thejameskyle:
Fixed the Symbol issue too. pretty-format was primarily built for the [Babel Online REPL](http://babeljs.io/repl/) so it's designed to work well with ES6+ both natively and via polyfills.

CrossEye:
Although this is not very central to Ramda's core functionality, it is one of those things I reach for all the times when debugging.  I like it.

:herb:

src/internal/_toString.js
```diff
@@ -0,0 +1,37 @@
...
+    } else if (x.constructor === String) {
+        return _quote(x);
+    } else if (Object.prototype.toString.call(x) === '[object Arguments]') {
+        return '(function() { return arguments; }(' + _map(recur, x).join(', ') + '))';
+    } else if (x.toString !== Object.prototype.toString) {  // Boolean, Function, RegExp, user-defined types
+        return x.toString();
+    } else {
+        return '{' + _map(function(k) { return _quote(k) + ': ' + recur(x[k]); }, keys(x).sort()).join(', ') + '}';
+    }
+};
```
CrossEye: Third try at this comment...

These conditionals have [cross-frame issues].  Should we care about this?

My gut feeling is that since it's straightforward enough to make this more robust, we probably should do so.  Others?

  [cf]: http://perfectionkills.com/instanceof-considered-harmful-or-how-to-write-a-robust-isarray/

davidchambers: I've addressed all but one of the cross-frame issues, I believe.

src/internal/_toString.js
```diff
@@ -0,0 +1,36 @@
...
+        case '[object Null]':
+            return 'null';
+        case '[object Number]':
+            return typeof x === 'object' ? 'new Number(' + recur(x.valueOf()) + ')' : 1 / x === -Infinity ? '-0' : x.toString(10);
+        case '[object String]':
+            return typeof x === 'object' ? 'new String(' + recur(x.valueOf()) + ')' : _quote(x);
+        case '[object Undefined]':
+            return 'undefined';
+        default:
+            return x.toString !== Object.prototype.toString ?  // Boolean, Function, RegExp, user-defined types
```
davidchambers: Can anyone think of an analogous check that will work across frames?

CrossEye: I'm not certain, but would something like

    `Object.prototype.toString.call(x)  !== '[object Object]'` 

do it?  It's a little awkward placed like that there, but that could just be one more `case` statement and your `default` could just be `x.toString()`.

davidchambers: That won't work for user-defined types such as:

```javascript
function Point(x, y) {
    this.x = x;
    this.y = y;
}

Point.prototype.toString = function() {
    return 'new Point(' + this.x + ', ' + this.y + ')';
};
```

CrossEye: I'm not sure if that was clear.  I'm suggesting:

```javascript
    case '[object Undefined]':
        return 'undefined';
    case '[object Object]':
        return '{' + _map(function(k) { return _quote(k) + ': ' + recur(x[k]); }, keys(x).sort()).join(', ') + '}';
    default:
        return x.toString() ;
```

CrossEye: Right, of course not... Hmmm.

davidchambers: I haven't yet worked out how to *test* cross-frame compatibility. I served a document from `http://localhost:4000` containing an iframe served from `http://localhost:4001`, and used [`window.postMessage`][1] to send values to the iframe, but `instanceof` checks worked correctly.


[1]: https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage

CrossEye: I think the structured clone algorithm in `window.postMessage` will recreate in your current context.  I haven't re-read kangax's article in years, but I think it doesn't take any cross-domain stuff, just, for instance, passing an Array created in one frame to a function created in another.

CrossEye: This is not perfect either:

```javascript
    case '[object Undefined]':
        return 'undefined';
    default:
        var result = x.toString();
        return result === '[object Object]' ? 
            '{' + _map(function(k) { return _quote(k) + ': ' + recur(x[k]); }, keys(x).sort()).join(', ') + '}'; :
            result;
```

But the only failure case I can see is if the user overrides `toString` in order to return `[object Object]`.  We might be able to live with that.

CrossEye: After that, my other solutions involve walking the prototype chain, which is also suspect in old IE...

davidchambers: Good idea.

`x.toString()` will throw if `x` is an object created via `Object.create(null)`, so we first need to check whether `typeof x.toString === 'function'`. I added a test case for this.

As well as checking that `x.toString() !== '[object Object]'` we can check that `x.constructor.name !== 'Object'`. I've updated the pull request to include both checks.

megawac: I think this should be really added as a ramda extension as it doesn't have
a clear purpose for general programs and its debugging functionality is
mostly covered by console.

davidchambers: > its debugging functionality is mostly covered by console

Yes and no. I'd prefer to see `Nothing()` rather than `{}`; `Maybe(42)` rather than `{value: 42}`.

I agree, though, that this function could live in a stand-alone package. The issues being discussed are relevant in that case too, so let's resolve the issues then decide whether to include this function in Ramda.

buzzdecafe:
:+1: @megawac this is simple enough to bolt on and needn't be included in core

davidchambers:
Two reasons to reconsider including this function:

  - it could be used to make `R.memoize` *much* more useful; and
  - it could be used to make the [REPL][1] output useful object representations.

The REPL currently displays `[object Object]` for any object, which is not very helpful at all. The REPL could of course take advantage of a third-party package. The question is whether we should include this function in Ramda or whether I should publish it as a stand-alone package. I don't mind either way. If we decide not to include this function in Ramda I suggest we rewrite `R.memoize` to take a serializer as an argument so users can provide the batteries.


[1]: http://ramdajs.com/repl/?v=0.13.0#(%7Bx%3A%201%2C%20y%3A%202%7D)

CrossEye:
I'm for including it.

buzzdecafe:
memoization argument is persuasive. :cow2:

megawac:
Still :-1: for including it - you guys would be better off to allow the users to customize the implementation of memoize so that people could use `WeakMap`s there (which are pretty much built for memoize). This would also be a crazy perf hit on memoize

The REPL argument isn't persuasive either because users can setup middleware to use pretty-format or the module you create

buzzdecafe:
I agree that possible REPL UX improvements are not germane.

OTOH, as it stands, current `R.memoize` is really only useful for primitive arguments. Which means it's not very useful at all. I'll be interested to see what it costs to evaluate and compare reference types for equality, and maybe it is not worth it. `WeakMap` would definitely be better, and perhaps we can define `memoize` in terms of that where available.

CrossEye:
Agree also that the REPL is not a motivator.  I think there's a good chance that the best solution for that would be a tree viewer anyway...

I agree that a WeakSet implementation of memoize would be better than our current one; although I don't know that Ramda wants to go that way yet or not, given that it's still supporting ES3 and ES5 as well as trying to be ready for es-next/ES6/ES2015/es-CAPEtheRENAMING, but to me the benefit of this function is not in that either.  It's simply a very useful tool for debugging/logging that I would often want to use in conjunction with my other tools.  I agree that it's an oddball, and I have no serious objection to suggesting that it belongs in a stand-alone package.  But Ramda is all about handling data, transforming it, manipulating it.  This is simply a tool to help display data; it doesn't seem much of a stretch to include it.

buzzdecafe:
on reflection, `WeakMap` not a good solution either, since it will still look up keys based on reference equality:

```javascript
var b2 = {b: 2};
var w = new WeakMap();
w.set(b2, 'B2');
w.get(b2); //=> 'B2' ... hooray, it works
w.get({b: 2}); //=> undefined ... bugger
```

CrossEye:
> on reflection, WeakMap not a good solution either

d'oh!  Of course!  That's the whole point rationale of a Stringified version...

CrossEye:
The other point to note is that

> allow the users to customize the implementation of memoize

is really quite tricky in Ramda, 

With its insistence on referential transparency, Ramda generally offers no way for users to customize the implementation of _anything_, except in the parameters supplied.  Memoization might be the one interesting exception as the whole point of that one is that since the functions are referentially transparent, it doesn't matter whether we recalculate them or not, but it's a strange concept.

davidchambers:
> With its insistence on referential transparency, Ramda generally offers no way for users to customize the implementation of _anything_, except in the parameters supplied.

I can't speak for others, but what I was suggesting is changing `R.memoize` from:

    (** -> a) -> (** -> a)

to:

    (* -> String) -> (** -> a) -> (** -> a)

A user could then define her own memoize function via partial application:

```javascript
const memoize = R.memoize(CANON.stringify);
```

CrossEye:
Right, we can do this through parameters passed.  What I was suggesting we couldn't do is something like [this suggestions][jdd] from a similar discussion on Underscore:

> For exampe devs could do `_.memoize.Cache = WeakMap;`

That's side-effect-y and out-of-bounds for Ramda.  Passed through a parameter to the function, it's fair game.

  [jdd]: https://github.com/jashkenas/underscore/issues/1862#issuecomment-56233877

buzzdecafe:
that makes the existing `memoize` essentially equivalent to `R.memoize(R.I)`. 

(p.s. I miss `R.I`)

buzzdecafe:
ping. in or no? i say let's give it a whirl and see if we can make `memoize` more robust (not a high bar, i grant you)

CrossEye:
:herb:

megawac:
> on reflection, WeakMap not a good solution either, since it will still look up keys based on reference equality

Are different objects with the same values really equal? For instance

```
function Foo() {}
function Bar() {}

eq(new Foo, new Bar) // ???????
```

I'm not sure I would actually want or expect this as it differs directly with JS paradigms

buzzdecafe:
it's a good question. are you suggesting some `instanceof` equality check as well? or just scrapping the whole thing?

davidchambers:
*Ideally* each constructor function would define an appropriate `toString` method on its prototype:

```javascript
function Foo() {}
Foo.prototype.toString = R.always('new Foo()');

function Bar() {}
Bar.prototype.toString = R.always('new Bar()');
```

megawac:
> or just scrapping the whole thing?

I'm very sceptical of the decision for its inclusion

PR #925, opened 2015-03-16
refer to partial in bind docs

raine:
It says

> Note: R.bind does not provide the additional argument-binding capabilities of Function.prototype.bind.

but doesn't tell you where to go.


CrossEye:
Good catch.  Thank you.

PR #927, opened 2015-03-16
Added example for take

bobiblazeski:
Added documentation example for take.


buzzdecafe:
this is great.

buzzdecafe:
should i add an example for `take(theATrain)`?

CrossEye:
Beautiful example!

bobiblazeski:
:blush:

CrossEye:
So long as no one define `var meOutTo = 9` and calls `take(meOutTo, theBallgame)` I think we're all safe.

PR #932, opened 2015-03-17
change mapObj category to object

taylor1791:
I don't if this was intentional, but mapObj is in the List category. I think it belongs in the Object category. If there is strong opposition, then feel free to close the request. 

I apologize if I didn't build some auxiliary files. `make` told me `'dist/ramda.js' is up to date.`


CrossEye:
Definitely not intentional.  Thanks for the fix.

We no longer commit `dist/ramda.js`, so you should be all set.

Issue #835, opened 2015-02-24
Document R.__ placeholder


michaelckelly:
It'd be great to see `R.__` broken out specifically in the docs.  It's fairly useful and only mentioned incidentally from what I have seen.

davidchambers:
:+1:

CrossEye:
Good idea.  Do I hear a PR coming?  :smile:

lwassermann:
+1, though it did not work in most of the cases were I tried using it. Could there be a restriction about using it as placeholder for a function argument or such?

CrossEye:
@lwassermann: It will probably work in many more places since #819 has been included, but that is not yet in any published version.

buzzdecafe:
closed by #935

PR #935, opened 2015-03-18
Add separate documentation for R.__

mcamac:
Addressing #835.

One question I had - would it be worthwhile to remove the mostly identical sections within the docs for `R.curry` and `R.curryN`?

Also slightly unsure about the `@constant` tag.

### DIFF
--- a/src/__.js
+++ b/src/__.js
@@ -1,1 +1,25 @@
+/**
+ * A special placeholder value used to specify "gaps" within curried functions,
+ * allowing partial application of any combination of arguments,
+ * regardless of their positions.
+ *
+ * If `g` is a curried ternary function and `_` is `R.__`, the following are equivalent:
+ *
+ *   - `g(1, 2, 3)`
+ *   - `g(_, 2, 3)(1)`
+ *   - `g(_, _, 3)(1)(2)`
+ *   - `g(_, _, 3)(1, 2)`
+ *   - `g(_, 2, _)(1, 3)`
+ *   - `g(_, 2)(1)(3)`
+ *   - `g(_, 2)(1, 3)`
+ *   - `g(_, 2)(_, 3)(1)`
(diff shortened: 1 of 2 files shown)


src/__.js
```diff
@@ -1 +1,26 @@
+/**
+ * A special placeholder value used to specify "gaps" wtihin curried functions,
```
davidchambers: s/wtihin/within/

src/__.js
```diff
@@ -1 +1,26 @@
...
+ *   - `g(_, 2)(_, 3)(1)`
+ *
+ * @constant
+ * @memberOf R
+ * @category Function
+ * @example
+ *
+ *      var powThenMod = R.curry(function(base, exp, mod) { return Math.pow(base, exp) % mod; });
+ *      var cubeThenMod = powThenMod(R.__, 3, R.__);
+ *      cubeThenMod(4, 10); //=> 4
```
davidchambers: The example above is a bit contrived. I'd prefer something like this:

```javascript
var greet = R.replace('{name}', R.__, 'Hello, {name}!');

greet('Alice'); //=> 'Hello, Alice!'
```

Ramda's build script uses this approach [here][1].


[1]: https://github.com/ramda/ramda/blob/a6b5af388741b6a36a786bd39dde78e8dc321fdc/scripts/build#L176

CrossEye:
:herb:

davidchambers:
:deciduous_tree: Very nice. Thank you, @mcamac!

mihaisavezi:
This is still not found in the documentation page on the homepage.

CrossEye:
Yes, another thing on which I dropped the ball.  This was working until a recent docs build/deploy.

I'll try again to get to this tonight.  See https://github.com/ramda/ramda.github.io/issues/214

PR #936, opened 2015-03-18
docs: improve usability

raine:
- focus filter input automatically
- go to the first entry on Enter

### DIFF
--- a/lib/doc/jsdoc-template/docs.html.handlebars
+++ b/lib/doc/jsdoc-template/docs.html.handlebars
@@ -31,6 +31,7 @@
                    placeholder="Filter"
                    type="text"
                    data-bind="textInput: filter"
+                   autofocus
             >
             <span class="form-control-feedback">
                 <span class="glyphicon glyphicon-search"></span>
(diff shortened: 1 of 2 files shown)


CrossEye:
:herb:

raine:
there we go

CrossEye:
LGTM. Anyone else have comments?

buzzdecafe:
:cow2:

A related question: rather than including `dist/ramda.js` should we point at version on CDN?

raine:
Ramda is included also for the purpose of making it available for users in the dev tools REPL.

It's a bit of a corner case, but there might be a situation where the page documents certain API that isn't available in the version we are using from cdn.

buzzdecafe:
I think the only times referencing the latest tag from a CDN would be a problem are:

* when generating docs from HEAD; or
* in the interval between publishing a new version & its docs and the CDN getting updated.

as i said, this is a separate issue. This PR is good to go as is AFAICT

CrossEye:
The docs are still built against `dist/ramda.js`, not the `src` dir, correct?  So we need to build it as part of creating the docs.  But we definitely shouldn't be _using_ it.

PR #937, opened 2015-03-19
replace erroneous occurrences of R.nix with R.not

davidchambers:
These files `require('ramda')` as in the dev dependency `ramda@0.11.0` specified in __package.json__. They were mistakenly updated in #902.


CrossEye:
Thanks.  I had to leave as the conversation about this was winding down, or I would have done it right away.

PR #940, opened 2015-03-19
Type signature fix for groupBy

santialbo:
`groupBy` returns an object where the values are arrays, the type signature in the docs should then be 

    (a -> s) -> [a] -> {s: [a]}`

### DIFF
--- a/src/groupBy.js
+++ b/src/groupBy.js
@@ -15,7 +15,7 @@ var _xgroupBy = require('./internal/_xgroupBy');
  * @func
  * @memberOf R
  * @category List
- * @sig (a -> s) -> [a] -> {s: [a]}
+ * @sig (a -> String) -> [a] -> {String: [a]}
  * @param {Function} fn Function :: a -> String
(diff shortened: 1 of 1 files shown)


src/groupBy.js
```diff
@@ -15,7 +15,7 @@ var _xgroupBy = require('./internal/_xgroupBy');
  * @func
  * @memberOf R
  * @category List
- * @sig (a -> s) -> [a] -> {s: a}
+ * @sig (a -> s) -> [a] -> {s: [a]}
```
davidchambers: We should use `String` in place of `s`.

CrossEye:
Nice catch.  Thank you.

megawac:
Or just s->str

santialbo:
Renamed the type parameter to `String`

buzzdecafe:
please rebase to pick up my change to .travis.yml. io.js made a change recently that broke some stuff, so we're using the prior version for now. thanks

PR #944, opened 2015-03-20
lock down io version io.js/path fixes

buzzdecafe:


buzzdecafe:
see also: http://docs.travis-ci.com/user/build-environment-updates/2015-02-03/#iojs_support

PR #946, opened 2015-03-20
Long-overdue credit

CrossEye:
Should have done this ages ago.

I think that when we add any further core members, we'll have to change the phrasing.

### DIFF
--- a/scripts/header
+++ b/scripts/header
@@ -4,7 +4,7 @@ set -e
 cat <<END
 //  Ramda v$(node -p 'require("envvar").string("VERSION", require("./package.json").version)')
 //  https://github.com/ramda/ramda
-//  (c) 2013-$(git show -s --format=%ai | cut -d - -f 1) Scott Sauyet and Michael Hurley
+//  (c) 2013-$(git show -s --format=%ai | cut -d - -f 1) Scott Sauyet, Michael Hurley, and David Chambers
 //  Ramda may be freely distributed under the MIT license.
 
 END


davidchambers:
I'm listed as a contributor in __package.json__. I don't mind being left out of the header, as you and Michael are Ramda's authors. I appreciate the recognition, of course. :)

CrossEye:
There are plenty of contributors, but pretty much since you came aboard, you're our most active one.  What Mike and I had been writing at this point seems merely the germ of the idea that this library has become.  And that's due to help from any number of people, but especially you.

I think it's appropriate.

megawac:
Definitely, done some (a lot of) good work @davidchambers

scripts/header
```diff
@@ -4,7 +4,7 @@ set -e
 cat <<END
 //  Ramda v$(node -p 'require("envvar").string("VERSION", require("./package.json").version)')
 //  https://github.com/ramda/ramda
-//  (c) 2013-$(git show -s --format=%ai | cut -d - -f 1) Scott Sauyet and Michael Hurley
+//  (c) 2013-$(git show -s --format=%ai | cut -d - -f 1) Scott Sauyet, Michael Hurley, and David Chambers
```
TheLudd: One comma too much =)

CrossEye: I've always been a fan of the [Oxford comma][oc].

  [oc]: http://en.m.wikipedia.org/wiki/Serial_comma

TheLudd: Never even heard of it before. But as long as it is intentional....
And I noticed that it wasn't used before. Perhaps this will be a breaking change then? ;)

CrossEye:
Of course if we were to add more names, the next one on the list would be Graeme Yeates!

Thank you too, @megawac, for all your contributions.

PR #948, opened 2015-03-20
accept negative offsets in R.substring{,From,To}

davidchambers:
It's slightly embarrassing that despite the existence of three substring functions, taking the last N characters of a string currently requires `R.slice` (which does not officially support strings). This pull request rectifies the situation by making `R.substring` an :alien: for `R.slice`.

### DIFF
--- a/src/substring.js
+++ b/src/substring.js
@@ -1,20 +1,21 @@
-var invoker = require('./invoker');
+var slice = require('./slice');
 
 
 /**
- * returns a subset of a string between one index and another.
+ * Returns a string containing the characters of `str` from `fromIndex`
+ * (inclusive) to `toIndex` (exclusive).
  *
  * @func
  * @memberOf R
  * @category String
  * @sig Number -> Number -> String -> String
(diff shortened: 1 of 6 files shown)


megawac:
:+1:

CrossEye:
:herbç

kevinbeaty:
This one surprises me.  It's not that I think it's a bad idea, just that @davidchambers is exporting an :alien: .   

Real question:  Would using `Array.prototype.slice` on a String lead to some of the string iteration issues for older environments that @CrossEye is trying to avoid?

megawac:
@kevinbeaty invoker's not using `Array.prototype.slice` but `String.prototype.slice`

kevinbeaty:
@megawac Ah. Makes sense. I completely glossed over the '_checkForMethod` in `slice`.  Thanks.

buzzdecafe:
> It's not that I think it's a bad idea, just that @davidchambers is exporting an :alien: .

zing!

PR #949, opened 2015-03-20
add R.nthChar and R.nthCharCode

davidchambers:
We're making a family: `R.nth`, `R.nthArg`, `R.nthChar`, `R.nthCharCode`.

### DIFF
--- a/src/charAt.js
+++ b/src/charAt.js
@@ -4,6 +4,7 @@ var invoker = require('./invoker');
 /**
  * The character at the nth position in a String:
  *
+ * @deprecated since v0.12.0
  * @func
  * @memberOf R
  * @category String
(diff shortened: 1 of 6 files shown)


CrossEye:
I think since changing from `toUpperCase`/`toLowerCase` to `toUpper`/`toLower`, we've slowly been moving away from an idea that I once had been contemplating, but had barely formulated, the idea of copying the entire native API into functional form.  There really was no good reason to do this, it did seem to be on of the things I was aiming for nonetheless; all the `invoker`-related code seemed to be headed that way.  The issues with doing `splice` should have demonstrated the futility once and for all, but I think it's still been in the back of my mind.

I think this puts the nail in that coffin.  And that's a good thing.  I can't stop feeling little pulls of regret that really make no sense at all, when we do something that's clearly an API improvement!

:+1:

buzzdecafe:
:cow2: :moon:

PR #951, opened 2015-03-21
docs: fix indent

### LINKED ISSUE #715: Clean up JSDoc @example whitespace
`publish.js` currently is a bit too aggressive stripping extra whitespace from JSDoc @example blocks. It should only trim leading/trailing newlines and the common amount of indentation on the whole block.

Alternatively, we could reformat all the @example tags, but that'd probably be more effort.

raine:
Closes #715

5 is a magic number for the amount of indent ramda has in its examples

```
 @example

      R.F(); //=> false
 └─5─┘
```

[before](https://cloud.githubusercontent.com/assets/11027/6765625/86039db8-cff1-11e4-8e9d-db2fa36030b6.png)
[after](https://cloud.githubusercontent.com/assets/11027/6765624/8421d6ae-cff1-11e4-8f38-93654c4e0fe6.png)

### DIFF
--- a/lib/doc/jsdoc-template/publish.js
+++ b/lib/doc/jsdoc-template/publish.js
@@ -20,7 +20,7 @@ var unprefix = R.pipe(
 
 var trimCode = R.pipe(
     R.split(/\n/),
-    R.map(R.trim),
+    R.map(R.replace(/^[ ]{5}/, '')),
     R.join('\n')
 );
 


buzzdecafe:
great, wanna try and publish 0.12 today, so i'd like to get this in. 5??? seems strange, but if it works ...

raine:
@buzzdecafe 

I checked jsdoc and their examples of `@example` have no indent, but ramda's does. No idea where this convention in ramda comes from.

lib/doc/jsdoc-template/publish.js
```diff
@@ -20,7 +20,7 @@ var unprefix = R.pipe(
 
 var trimCode = R.pipe(
     R.split(/\n/),
-    R.map(R.trim),
+    R.map(R.replace(/^\s{5}/, '')),
```
davidchambers: Shouldn't this be `[ ]` rather than `\s`?

raine: What's the difference?

davidchambers: [Shorthand character classes][1]:

> `\s` stands for "whitespace character". Again, which characters this actually includes, depends on the regex flavor. In all flavors discussed in this tutorial, it includes `[ \t\r\n\f]`. That is: `\s` matches a space, a tab, a line break, or a form feed. Most flavors also include the vertical tab, with Perl (prior to version 5.18) and PCRE (prior to version 8.34) being notable exceptions. In flavors that support Unicode, `\s` normally includes all characters from the Unicode "separator" category. Java and PCRE are exceptions once again. But JavaScript does match all Unicode whitespace with `\s`.


[1]: http://www.regular-expressions.info/shorthand.html

raine: You're right. `[ ]` makes more sense here.

davidchambers:
:deciduous_tree: Thanks!

PR #955, opened 2015-03-21
define R.partition without _reduce

davidchambers:


megawac:
Worth doing something like https://github.com/jashkenas/underscore/issues/1594?

davidchambers:
I'm keen to remove an internal dependency, not replace one with another. ;)

megawac:
Eh, in my opinion composition makes sense for groupBy, partition, countBy,
and index?By as they use the same code (this would also allow them to
potentially be implemented via a transducer)

buzzdecafe:
i should compile a list of remaining functions that should be transducerified.

davidchambers:
((:bell:)) @CrossEye

CrossEye:
Guess I missed this one.   I have the same feeling about this as about the other related ones.   I mostly take issue with the motivation behind it.  There's nothing wrong with the code, or with making the change.   But as you pointed out in another thread, we need to determine how we weigh the competing priorities.

PR #956, opened 2015-03-22
replace occurrences of assert(...) and assert.ok(...)

davidchambers:
`assert` and its :alien: `assert.ok` are problematic for two reasons:

  - They're imprecise. Consider this buggy `isOdd` function:

    ```javascript
    var isOdd = function(n) { return n % 2; };

    assert(isOdd(7));
    assert.strictEqual(isOdd(7), true);  // AssertionError: 1 === true
    ```

  - They don't provide useful feedback in case of failure:

    ```javascript
    assert(2 + 2, 5);  // AssertionError: false == true
    assert.strictEqual(2 + 2, 5);  // AssertionError: 4 === 5
    ```

### DIFF
--- a/test/all.js
+++ b/test/all.js
@@ -12,7 +12,7 @@ describe('all', function() {
 
   it('returns true if all elements satisfy the predicate', function() {
     assert.strictEqual(R.all(even, [2, 4, 6, 8, 10, 12]), true);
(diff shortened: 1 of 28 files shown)


buzzdecafe:
:cow2:

PR #958, opened 2015-03-22
define R.mapObj and R.mapObjIndexed without _reduce

davidchambers:
See #955

### DIFF
--- a/src/mapObj.js
+++ b/src/mapObj.js
@@ -1,5 +1,4 @@
 var _curry2 = require('./internal/_curry2');
-var _reduce = require('./internal/_reduce');
 var keys = require('./keys');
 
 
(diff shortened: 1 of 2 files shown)


CrossEye:
I have the same objection I've expressed several times to the general program of making it an important goal to drastically reduce internal depenencies.

In this case, we had two five-line functions that were duplicates except in one key line.  This change makes them two eleven-line functions that are duplicates except in one key line.  I don't see how the benefit of removing this dependency is worth the cost of the increased code duplication.

I do realize that this is a balancing act, and I'd like to hear the arguments for it.

honzabrecka:
Now, when Ramda supports transducers, isn't `mapObj` slightly redundant?

CrossEye:
Does transducers give us as simple an API as:@

    mapObj (square, {a: 1, b: 2, c: 3});
    //=> {a: 1, b: 4, c: 9})

Certainly you can achieve this, but the smoke API for it is useful.

CrossEye:
I wasn't trying to close off debate.  Unless I've convinced you, I would rather hear the arguments in favor.

davidchambers:
It's a question of priorities. There are trade-offs involving:

  - performance;
  - code duplication;
  - the file size of __ramda.js__ when minified and gzipped; and
  - the file size added by `require('ramda/src/<name>')` when using Browserify or a similar tool.

It would be useful for the contributing guidelines to cover the relative importance of the factors above along with correctness, consistency, and compatibility.

PR #961, opened 2015-03-22
make R.cond a fixed-arity function

davidchambers:
See #919

### DIFF
--- a/src/cond.js
+++ b/src/cond.js
@@ -1,7 +1,10 @@
+var _curry1 = require('./internal/_curry1');
+
+
 /**
  * Returns a function, `fn`, which encapsulates if/else-if/else logic.
- * Each argument to `R.cond` is a [predicate, transform] pair. All of
- * the arguments to `fn` are applied to each of the predicates in turn
+ * `R.cond` takes a list of [predicate, transform] pairs. All of the
+ * arguments to `fn` are applied to each of the predicates in turn
  * until one returns a "truthy" value, at which point `fn` returns the
  * result of applying its arguments to the corresponding transformer.
  * If none of the predicates matches, `fn` returns undefined.
@@ -9,22 +12,21 @@
  * @func
  * @memberOf R
  * @category Logic
- * @sig [(*... -> Boolean),(*... -> *)]... -> (*... -> *)
- * @param {...Function} functions
+ * @sig [[(*... -> Boolean),(*... -> *)]] -> (*... -> *)
+ * @param {Array} pairs
  * @return {Function}
  * @example
  *
- *      var fn = R.cond(
+ *      var fn = R.cond([
  *        [R.equals(0),   R.always('water freezes at 0°C')],
  *        [R.equals(100), R.always('water boils at 100°C')],
  *        [R.T,           function(temp) { return 'nothing special happens at ' + temp + '°C'; }]
- *      );
+ *      ]);
  *      fn(0); //=> 'water freezes at 0°C'
  *      fn(50); //=> 'nothing special happens at 50°C'
  *      fn(100); //=> 'water boils at 100°C'
  */
-module.exports = function cond() {
-  var pairs = arguments;
+module.exports = _curry1(function cond(pairs) {
   return function() {
     var idx = 0;
     while (idx < pairs.length) {
@@ -34,4 +36,4 @@ module.exports = function cond() {
       idx += 1;
     }
   };
-};
+});
--- a/test/cond.js
+++ b/test/cond.js
@@ -5,41 +5,41 @@ var R = require('..');
 
 describe('cond', function() {
   it('returns a function', function() {
-    assert.strictEqual(typeof R.cond(), 'function');
+    assert.strictEqual(typeof R.cond([]), 'function');
   });
 
   it('returns a conditional function', function() {
-    var fn = R.cond(
+    var fn = R.cond([
       [R.equals(0),   R.always('water freezes at 0°C')],
       [R.equals(100), R.always('water boils at 100°C')],
       [R.T,           function(temp) { return 'nothing special happens at ' + temp + '°C'; }]
-    );
+    ]);
     assert.strictEqual(fn(0), 'water freezes at 0°C');
     assert.strictEqual(fn(50), 'nothing special happens at 50°C');
     assert.strictEqual(fn(100), 'water boils at 100°C');
   });
 
   it('returns a function which returns undefined if none of the predicates matches', function() {
-    var fn = R.cond(
+    var fn = R.cond([
       [R.equals('foo'), R.always(1)],
       [R.equals('bar'), R.always(2)]
-    );
+    ]);
     assert.strictEqual(fn('quux'), undefined);
   });
 
   it('predicates are tested in order', function() {
-    var fn = R.cond(
+    var fn = R.cond([
       [R.T, R.always('foo')],
       [R.T, R.always('bar')],
       [R.T, R.always('baz')]
-    );
+    ]);
     assert.strictEqual(fn(), 'foo');
   });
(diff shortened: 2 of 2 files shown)


buzzdecafe:
i don't see the reason for this.

davidchambers:
I would like Ramda to export no variadic functions. If a function operates on an arbitrary number of values, `[` and `]` are all one needs. :)

I prefer `R.max` to `Math.max`: working with lists is nicer than dealing with `Function.prototype.apply` (or `R.apply`).

buzzdecafe:
not presuaded yet. what are the concrete advantages of wrapping `cond` in an extra pair of square brackets?

davidchambers:
> what are the concrete advantages of wrapping `cond` in an extra pair of square brackets?

Let's not focus on syntax. Let's focus on the underlying data structures.

In Lisp, where `if` is a function-like special form rather than a statement, we can *traverse* the data structures underlying our conditionals and even *transform* them.

JavaScript is not a Lisp, but `R.cond` does allow conditional logic to be represented as a list of pairs. One can transform a list of pairs (via `R.map`, for example) before applying `R.cond`. (The function returned by `R.cond` is a black box, preventing further transformations.)

Here's an example of input transformation prior to application of `R.cond`:

```javascript
//  grade :: GradeMap -> Number -> String
var grade =
R.pipe(R.toPairs,
       R.map(R.reverse),
       R.sort(R.unapply(R.pipe(R.map(R.head), R.apply(R.lt)))),
       R.map(R.converge(R.unapply(R.identity),
                        R.pipe(R.head, R.flip(R.gte)),
                        R.pipe(R.last, R.always))),
       R.cond);


//  highSchoolGrade :: Number -> String
var highSchoolGrade = grade({
  'A' : 90,
  'B' : 80,
  'C' : 70,
  'D' : 60,
  'F' :  0,
});


//  collegeGrade :: Number -> String
var collegeGrade = grade({
  'A' : 93,
  'A−': 90,
  'B+': 87,
  'B' : 83,
  'B−': 80,
  'C+': 77,
  'C' : 70,
  'D' : 60,
  'F' :  0,
});
```

```javascript
> highSchoolGrade(90)
'A'
> collegeGrade(90)
'A−'
```

CrossEye:
I think I'm finally coming around to the no-variadic functions point of view.  (`func` vs `invoker` and the discussions about changing `compose` and `pipe` make this clear.)

But this example is not particularly compelling.  It uses `unapply` twice and `apply` once already.  Just wrapping the final `cond` in an `apply` and it would work with the current version of `cond`.

I would like this change as part of a campaign to remove the final few variadic functions and make Ramda more consistent.  That's a win in itself.  But I don't see it as a huge win.  And with the existence of `apply` and `unapply`, I doubt we're going to find too many compelling arguments of the see-how-much-easier-this-is variety.

davidchambers:
> I would like this change as part of a campaign to remove the final few variadic functions and make Ramda more consistent.

I view this pull request of part of the campaign. :)

As you say, `R.apply` and `R.unapply` make variadic functions easy to include in functional pipelines. Why, though, should `R.cond` be any different from `R.max`? Why is `R.max` not a variadic function? After all, `R.apply(R.max)` could be used wherever a max function of type `[Number] -> Number` is required.

I wonder whether you and/or @buzzdecafe are drawing a distinction between lists which appear in the code itself and lists which are passed around as arguments. Perhaps you see arguments to `R.cond` as code and arguments to `R.max` as data. I don't draw this distinction, and therefore believe `R.cond` and `R.max` should agree on whether values are taken as positional arguments or as a list.

CrossEye:
Sorry, I seem to have lost several threads some days ago.

I am agreeing here.  This is worth doing.  I would like to remove the final few variadic functions from Ramda.  I was simply pointing out that this is good mostly because consistency is itself a good thing, and not because anything will become much simpler or more obvious as the result of the change.

I do draw some distinctions between source code and data, but it's not a hard distinction.  It's just that how I like to work with source code in JS is often at odds with the more flexible way I would deal with general data.  This is the reason that I regretted the change from `path('a.b.c')` to `path(['a', 'b', 'c'])`.  While the latter is clearly more flexible, more generic, etc., the former better expresses in source code what I mean.  I accept the inevitable, but such acceptance isn't always wholehearted endorsement.  :smiling_imp:

megawac:
I'm not a fan of this. No varidity for the sake of no varidity (at the sake of api simplicity) doesn't make sense to me

Further, this change deviates from the semantics of this fn in functional languages (scheme&racket)

buzzdecafe:
i remain unconvinced. I have yet to see what the benefit is, or the harm, depending on how you frame the question.

incidentally, you left out `gradSchoolGrade = grade({A: 50, B: 0})`

CrossEye:
> Further, this change deviates from the semantics of this fn in functional languages (scheme&racket&clojure)

Those languages are very different, though.  The auto-currying in Ramda is closer in spirit to ML or Haskell.

The pros and cons are simple:

**Pros**
  - You don't have to wrap things up with `apply` or `unapply` to use this comfortably in source
  - It adds additional consistency to the library

**Cons**
  - It requires an additional pair of `[` - `]` in source code to use.
  - It's a breaking change

To me those pros outweigh the cons, but only because we're still trying to get the API in shape for 1.0.  Once we hit there, this would be much harder to change.

(BTW, in my grad school, it was  `grade({A: 50, 'A-': 0})` :smile:.)

buzzdecafe:
> Those languages are very different, though. The auto-currying in Ramda is closer in spirit to ML or Haskell.

sure, but `cond` is _clearly_ taken from Scheme. so this would break expectation of its behavior for anyone who has ever used `cond` anywhere else, ever.

buzzdecafe:
hmmmm

https://hackage.haskell.org/package/cond-0.1/docs/Control-Cond.html

CrossEye:
I don't know.  I try to keep to the point that we take inspiration from other languages, but make our own API.  For instance, if we were following LISP, 

    R.map(R.multiply, [3, 4, 5], [4, 5, 6]); //=> [12, 20, 30]

But we haven't bothered really considering this, although I have a vague memory of a request for it.

buzzdecafe:
> `R.map(R.multiply, [3, 4, 5], [4, 5, 6]); //=> [12, 20, 30]` 

I don't see how you could do that and curry. 

Anyways, I am still mulling this. Do we want *no variadic functions ever*? Why? We still have to cope with variadic functions in the outside world.

scott-christopher:
My 2 cents: A global constraint of "no variadic functions" does makes it easier to design and ensure consistency of the API.

Users of the API will still have access to `apply` for interacting with external variadic functions or `unapply` for converting Ramda functions to become variadic, if desired.

buzzdecafe:
i am willing to accede to this, although it feels pretty weird to me. let's hold off on this though, in order to get #982 out there

buzzdecafe:
status here?

CrossEye:
I'm still weakly in favor, but haven't given it any further thought.

davidchambers:
I'm strongly in favour of reducing the number of variadic functions in the library to zero. If we're to do so we should do so all at once. Let's discuss this after publishing v0.14.0.

davidchambers:
I've revived this pull request. I'm still very much in favour of this change.

CrossEye:
:herb:

davidchambers:
It seems we have three in favour (@CrossEye, @davidchambers, @scott-christopher) and two against (@buzzdecafe, @megawac). Apparently this issue is quite divisive.

davidchambers:
@buzzdecafe pointed out that while he isn't thrilled with this change he did [consent][1] to it.


[1]: https://github.com/ramda/ramda/pull/961#issuecomment-87659943

PR #963, opened 2015-03-22
add R.isNaN

davidchambers:
Useful given the questionable semantics of [`isNaN`][1].

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN

### DIFF
--- a/test/isNaN.js
+++ b/test/isNaN.js
@@ -1,0 +1,28 @@
+var assert = require('assert');
+
+var R = require('..');
+
+
+describe('isNaN', function() {
+
+  it('returns true for `NaN`', function() {
+    assert.strictEqual(R.isNaN(NaN), true);
+
+    assert.strictEqual(isNaN(NaN), true);
+  });
+
+  it('returns false for `new Number(NaN)`', function() {
+    /* jshint -W053 */
+    assert.strictEqual(R.isNaN(new Number(NaN)), false);
+    /* jshint +W053 */
+  });
+
+  it('returns false for any other value', function() {
+    assert.strictEqual(R.isNaN(void 0), false);
+    assert.strictEqual(R.isNaN({}), false);
+
+    assert.strictEqual(isNaN(void 0), true);
+    assert.strictEqual(isNaN({}), true);
+  });
+
+});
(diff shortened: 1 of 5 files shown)


megawac:
This misses `isNaN(new Number(NaN))`

test/isNaN.js
```diff
@@ -0,0 +1,28 @@
...
+    assert.strictEqual(R.isNaN(new Number(NaN)), false);
+    /* jshint +W053 */
+  });
+
+  it('returns false for any other value', function() {
+    assert.strictEqual(R.isNaN(void 0), false);
+    assert.strictEqual(R.isNaN({}), false);
+
+    assert.strictEqual(isNaN(void 0), true);
+    assert.strictEqual(isNaN({}), true);
```
megawac: Are these tests necessary?

davidchambers:
> This misses `isNaN(new Number(NaN))`

Yes, to mirror [`Number.isNaN`][1]. There's even a test case which asserts that `R.isNaN(new Number(NaN))` evaluates false.


[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN

test/isNaN.js
davidchambers: No, although they may be instructive for users thinking _Doesn't JavaScript provide a global `isNaN` function?_

CrossEye:
:herb:

megawac:
Fair enough, but does it not make more sense to implement this in `R.is`?

davidchambers:
> Fair enough, but does it not make more sense to implement this in `R.is`?

I don't think so. `R.is` checks a value's *type*. It's confusing if the first argument is a *value* in one case.

CrossEye:
Yeah, good old JS:

    R.is(Number, NaN); //=> true,

which matches

    typeof NaN; //=> "number"

davidchambers:
I assumed @megawac was suggesting `R.is(NaN, x)`, but perhaps I misunderstood.

megawac:
Yea

CrossEye:
> I assumed @megawac was suggesting `R.is(NaN, x)`, but perhaps I misunderstood.

Right, but it would also not do to change the current behavior of `.is(Number)`, so imagine:

    var x = 0 / 0;
    R.is(Number, x); //=> true
    R.is(NaN, x); //=> true
    // WTF?

That's all.

PR #964, opened 2015-03-22
remove unnecessary adverbs

davidchambers:


CrossEye:
Another one I missed.  What have you got against adverbs added unnecessarily?  :smile:

PR #971, opened 2015-03-24
Adds lensIndex, lensProp, composeL & pipeL

scott-christopher:
This change introduces 2 new helpers for generating lenses,
`lensIndex` and `lensProp`, where `lensIndex` is used to
create a new lens with a focus on an array index, while
`lensProp` is used to create a new lens with a focus on an
object property.

```js
var phraseLens = R.lensProp('phrase');
var obj1 = { phrase: 'Absolute filth . . . and I LOVED it!'};
phraseLens(obj1); // => 'Absolute filth . . . and I LOVED it!'
phraseLens.set('Ooh Betty', obj1); //=> { phrase: 'Ooh Betty'}

var headLens = R.lensIndex(0);
headLens([10, 20, 30, 40]); //=> 10
headLens.set('mu', [10, 20, 30, 40]); //=> ['mu', 20, 30, 40]
```

This change also introduces `composeL` and `pipeL` which
are used to compose lenses together, resulting in a new lens
which can be used to access and update nested properties
of a target object.

```js
var headLens = R.lensIndex(0);
var secondLens = R.lensIndex(1);
var xLens = R.lensProp('x');
var secondOfXOfHeadLens = R.composeL(secondLens, xLens, headLens);

var source = [{x: [0, 1], y: [2, 3]}, {x: [4, 5], y: [6, 7]}];
secondOfXOfHeadLens(source); //=> 1
secondOfXOfHeadLens.set(123, source); //=> [{x: [0, 123], y: [2, 3]}, {x: [4, 5], y: [6, 7]}]
```

### DIFF
--- a/src/composeL.js
+++ b/src/composeL.js
@@ -1,0 +1,37 @@
+var _composeL = require('./internal/_composeL');
+
+/**
+ * Creates a new lens that allows getting and setting values of nested properties, by
+ * following each given lens in succession.
+ *
+ * Note that `composeL` is a right-associative function, which means the lenses provided
+ * will be invoked in order from right to left.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @see R.lens
+ * @sig ((y -> z), (x -> y), ..., (b -> c), (a -> b)) -> (a -> z)
+ * @param {...Function} lenses A variable number of lenses.
+ * @return {Function} A new lens which represents the result of calling each of the
+ *         input `lenses`, passing the result of each getter/setter as the source
+ *         to the next, from right to left.
+ * @example
+ *
+ *      var headLens = R.lensIndex(0);
+ *      var secondLens = R.lensIndex(1);
+ *      var xLens = R.lensProp('x');
+ *      var secondOfXOfHeadLens = R.composeL(secondLens, xLens, headLens);
+ *
+ *      var source = [{x: [0, 1], y: [2, 3]}, {x: [4, 5], y: [6, 7]}];
+ *      secondOfXOfHeadLens(source); //=> 1
+ *      secondOfXOfHeadLens.set(123, source); //=> [{x: [0, 123], y: [2, 3]}, {x: [4, 5], y: [6, 7]}]
+ */
+module.exports = function() {
+  var idx = arguments.length - 1;
+  var fn = arguments[idx];
+  while (--idx >= 0) {
+    fn = _composeL(arguments[idx], fn);
+  }
+  return fn;
+};
--- a/src/lensIndex.js
+++ b/src/lensIndex.js
@@ -1,0 +1,27 @@
+var _slice = require('./internal/_slice');
+var lens = require('./lens');
+var nth = require('./nth');
+
+/**
+ * Creates a lens that will focus on index `n` of the source array.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @see R.lens
+ * @sig Number -> (a -> b)
+ * @param {Number} n The index of the array that the returned lens will focus on.
+ * @return {Function} the returned function has `set` and `map` properties that are
+ *         also curried functions.
+ * @example
+ *
+ *     var headLens = R.lensIndex(0);
+ *     headLens([10, 20, 30, 40]); //=> 10
+ *     headLens.set('mu', [10, 20, 30, 40]); //=> ['mu', 20, 30, 40]
+ *     headLens.map(function(x) { return x + 1; }, [10, 20, 30, 40]); //=> [11, 20, 30, 40]
+ */
+module.exports = function lensIndex(n) {
+  return lens(nth(n), function(x, xs) {
+    return _slice(xs, 0, n).concat([x], _slice(xs, n + 1));
+  });
+};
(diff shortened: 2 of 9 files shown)


src/lensIndex.js
```diff
@@ -0,0 +1,30 @@
...
+ * @example
+ *
+ *     var headLens = R.lensIndex(0);
+ *     headLens([10, 20, 30, 40]); //=> 10
+ *     headLens.set('mu', [10, 20, 30, 40]); //=> ['mu', 20, 30, 40]
+ *     headLens.map(function(x) { return x + 1; }, [10, 20, 30, 40]); //=> [11, 20, 30, 40]
+ */
+module.exports = function lensIndex(n) {
+  return lens(nth(n), function(x, xs) {
+    return _concat(
```
megawac: probably faster to do `slice(xs, 0, n].concat([x], _slice(xs, n + 1))`

scott-christopher: As suggested, I've updated this to use `_slice(xs, 0, n).concat([x], _slice(xs, n + 1))` instead.

src/lensProp.js
```diff
@@ -0,0 +1,28 @@
...
+ *     var obj1 = { phrase: 'Absolute filth . . . and I LOVED it!'};
+ *     var obj2 = { phrase: "What's all this, then?"};
+ *     phraseLens(obj1); // => 'Absolute filth . . . and I LOVED it!'
+ *     phraseLens(obj2); // => "What's all this, then?"
+ *     phraseLens.set('Ooh Betty', obj1); //=> { phrase: 'Ooh Betty'}
+ *     phraseLens.map(R.toUpper, obj2); //=> { phrase: "WHAT'S ALL THIS, THEN?"}
+ */
+module.exports = function(k) {
+  return lens(prop(k), assoc(k));
+};
```
buzzdecafe: no doubt, this is a super useful pattern

src/composeL.js
```diff
@@ -0,0 +1,36 @@
...
+ * following each given lens in succession.
+ *
+ * Note that `composeL` is a right-associative function, which means the lenses provided
+ * will be invoked in order from right to left.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig ((y -> z), (x -> y), ..., (b -> c), (a -> b)) -> (a -> z)
+ * @param {...Function} lenses A variable number of lenses.
```
buzzdecafe: should probably include `@see lens` on these docs

CrossEye:
:herb:

Very nice stuff!  I really like `composeL`/`pipeL`.

I wonder if we also want `lensPath` similar to `lensProp` using `path` and `assocPath`

src/lensIndex.js
```diff
@@ -0,0 +1,26 @@
...
+ * @example
+ *
+ *     var headLens = R.lensIndex(0);
+ *     headLens([10, 20, 30, 40]); //=> 10
+ *     headLens.set('mu', [10, 20, 30, 40]); //=> ['mu', 20, 30, 40]
+ *     headLens.map(function(x) { return x + 1; }, [10, 20, 30, 40]); //=> [11, 20, 30, 40]
+ */
+module.exports = function lensIndex(n) {
+  return lens(nth(n), function(x, xs) {
+    return _slice(xs, 0, n).concat([x], _slice(xs, n + 1));
```
scott-christopher: Is it worth pulling this out to a separate function, equivalent to `assoc` for arrays?

src/composeL.js
scott-christopher: I've updated the docs to include `@see R.lens`.

src/lensIndex.js
buzzdecafe: discussion about whether `assoc` should work on arrays as well: #773

buzzdecafe:
:cow2:

this looks good. any input @davidchambers ?

src/lensProp.js
```diff
@@ -0,0 +1,29 @@
...
+
```
davidchambers: :space_invader:

davidchambers:
I haven't yet used lenses, but these functions make sense to me.

kevinbeaty:
Love this.  Was actually just looking for something like `R.lensIndex` and the other functions look just as useful.  Thanks @scott-christopher !

niant:
Looking forward to this! Hopefully next release is soon :+1:

jethrolarson:
Should lensIndex this use R.update for the setter?

CrossEye:
It would not necessarily be too much dogfood.  But I would want to compare that to the existing implementation, especially as `converge` is one of our fill-in-the-gaps functions.  (`useWith` is the other main one.)  These help us do things that we would really like to be able just to do with `compose` but cannot for various reasons.  So as useful as `converge` is, I would rather not promote it overmuch.  And the other important point is that the current implementation is already very clean.  Just making it points-free because we can doesn't add much.

scott-christopher:
The reason why `update` wasn't initially used to implement the setter in `lensIndex` was because `update` didn't exist at the time (well, for 4 days) :)

Using `update` would provide a nice symmetry with `lensProp`.

`lensProp`:
```js
function(k) {
  return lens(prop(k), assoc(k));
}
```

`lensIndex`:
```js
function(n) {
  return lens(nth(n), update(n));
}
```

CrossEye:
I would definitely be in favor of that change.

PR #972, opened 2015-03-24
uncurryN

### LINKED ISSUE #970: Higher order functions and other ways of currying.
For example:

http://ramdajs.com/repl/#R.reduce(function(x)%20%7B%0A%20%20return%20function(y)%20%7B%0A%20%20%20%20return%20x%20%2B%20y%3B%0A%20%20%7D%3B%0A%7D%2C%200%2C%20%5B1%2C2%2C3%5D)%3B

```javascript
R.reduce(function(x) {
  return function(y) {
    return x + y;
  };
}, 0, [1,2,3]);
```

Gives back 
```javascript
function (y) {
    return x + y;
  }
```

whereas 
http://ramdajs.com/repl/#R.reduce(R.add%2C%200%2C%20%5B1%2C2%2C3%5D)%3B
```javascript
R.reduce(R.add, 0, [1,2,3]);
```

Gives back
```javascript
6
```
 as expected.

buzzdecafe:
closes #970

### DIFF
--- a/src/uncurryN.js
+++ b/src/uncurryN.js
@@ -1,0 +1,46 @@
+var _curry2 = require('./internal/_curry2');
+var _slice = require('./internal/_slice');
+var curryN = require('./curryN');
+
+
+/**
+ * Returns a function of arity `n` from a (manually) curried function.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig Number -> (a -> b) -> (a -> c)
+ * @param {Number} length The arity for the returned function.
+ * @param {Function} fn The function to uncurry.
+ * @return {Function} A new function.
+ * @see R.curry
+ * @example
+ *
+ *      var addFour = function(a) {
+ *        return function(b) {
+ *          return function(c) {
+ *            return function(d) {
+ *              return a + b + c + d;
+ *            };
+ *          };
+ *        };
+ *      };
+ *
+ *      var uncurriedAddFour = R.uncurryN(4, addFour);
+ *      curriedAddFour(1, 2, 3, 4); //=> 10
+ */
+module.exports = _curry2(function uncurryN(depth, fn) {
+  return curryN(depth, function() {
+    var currentDepth = 1;
+    var value = fn;
+    var idx = 0;
+    var endIdx;
+    while (currentDepth <= depth && typeof value === 'function') {
+      endIdx = currentDepth === depth ? arguments.length : idx + value.length;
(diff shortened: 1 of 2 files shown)


src/uncurryN.js
```diff
@@ -0,0 +1,43 @@
...
+ *      };
+ *
+ *      var uncurriedAddFour = R.uncurryN(4, addFour);
+ *      curriedAddFour(1, 2, 3, 4); //=> 10
+ */
+module.exports = _curry2(function uncurryN(length, fn) {
+  return arity(length, function() {
+    var idx = 0;
+    var value = fn;
+    while (idx < length && is(Function, value)) {
```
davidchambers: You could use `typeof value === 'function'` here.

buzzdecafe: ok, will do; alson considering whether it might be better to wrap the returned function in `curryN` rather than just `arity`.

CrossEye: On [further consideration][ss], I agree that we should curry the output, but that probably means that this function needs a different name.  We discussed `recurry`.  Any other possibilities?

  [ss]: https://github.com/ramda/ramda/issues/970#issuecomment-85625668

davidchambers: `recurry` is fine, though I think `uncurry` is fine too. `curry` returns a function which accepts all its arguments at once, so I have no problem with `uncurry` returning a function which accepts its arguments one by one.

algesten: given the names i sort of expect `uncurry` to undo what `curry` did. 

```javascript
var fn1 = curry(fn);
var fn2 = uncurry(fn1);
fn === fn2 // true would be cool
```

buzzdecafe: wow, that *would* be cool, but maybe a tall order.

CrossEye: This probably could be done with metadata.  But it would also work the other way, and that might get a bit more crazy:

```javascript
var fn1 = uncurry(fn);
var fn2 = curry(fn1);
fn === fn2; //=> true
```

But I don't think such invariants are actually all that important.  The point of `uncurry`/`recurry`, or whatever we end up calling it, whether or not the result actually ends up curried, is that it could take a function that was manually curried to depth `n` and return an arity-n function with the same behaviour.  This implementation  has the nice advantage that it would also perform the same service for an already arity-n function or one that was curried through a `curry` function like Ramda's, meaning that it could be used fairly widely, but that's not central.  And actual function identity is still further from the core.

buzzdecafe:
we good here?

CrossEye:
:herb:

PR #973, opened 2015-03-25
Adds `adjust` and `update`

scott-christopher:
**This is a proposal / stawman for #773**

This change introduces 2 new list functions `adjust` and `update`, providing the ability to modify values at a given index of a list. `adjust` applies a given function to replace the value, while `update` is the specialised form, updating the array with a new explicit value.

```js
R.adjust(R.add(10))(1)([0, 1, 2]);  //=> [0, 11, 2]
R.update(1)(11)([0, 1, 2]);         //=> [0, 11, 2]
```

### DIFF
--- a/src/adjust.js
+++ b/src/adjust.js
@@ -1,0 +1,34 @@
+var _concat = require('./internal/_concat');
+var _curry3 = require('./internal/_curry3');
+
+/**
+ * Applies a function to the value at the given index of an array,
+ * returning a new copy of the array with the element at the given
+ * index replaced with the result of the function application.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig (a -> a) -> Number -> [a] -> [a]
+ * @param {Function} fn The function to apply.
+ * @param {Number} idx The index.
+ * @param {Array|Arguments} list An array-like object whose value
+ *        at the supplied index will be replaced.
+ * @return {Array} A copy of the supplied array-like object with
+ *         the element at index `idx` replaced with the value
+ *         returned by applying `fn` to the existing element.
+ * @example
+ *
+ *      R.adjust(R.add(10), 1, [0, 1, 2]);     //=> [0, 11, 2]
+ *      R.adjust(R.add(10))(1)([0, 1, 2]);     //=> [0, 11, 2]
+ */
+module.exports = _curry3(function(fn, idx, list) {
+  if (idx >= list.length || idx < -list.length) {
+    return list;
+  }
+  var start = idx < 0 ? list.length : 0;
+  var _idx = start + idx;
+  var _list = _concat(list);
+  _list[_idx] = fn(list[_idx]);
+  return _list;
+});
--- a/src/update.js
+++ b/src/update.js
@@ -1,0 +1,24 @@
+var _curry3 = require('./internal/_curry3');
+var adjust = require('./adjust');
+var always = require('./always');
+
+/**
+ * Returns a new copy of the array with the element at the
+ * provided index replaced with the given value.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig Number -> a -> [a] -> [a]
+ * @param {Number} idx The index to update.
+ * @param {*} x The value to exist at the given index of the returned array.
+ * @param {Array|Arguments} list The source array-like object to be updated.
+ * @return {Array} A copy of `list` with the value at index `idx` replaced with `x`.
+ * @example
+ *
+ *      R.update(1, 11, [0, 1, 2]);     //=> [0, 11, 2]
+ *      R.update(1)(11)([0, 1, 2]);     //=> [0, 11, 2]
+ */
+module.exports = _curry3(function(idx, x, list) {
+  return adjust(always(x), idx, list);
+});
--- a/test/adjust.js
+++ b/test/adjust.js
@@ -1,0 +1,36 @@
+var assert = require('assert');
+
+var R = require('..');
+
+describe('adjust', function() {
+  it('applies the given function to the value at the given index of the supplied array', function() {
+    assert.deepEqual(R.adjust(R.add(1), 2, [0, 1, 2, 3]), [0, 1, 3, 3]);
+  });
+
+  it('offsets negative indexes from the end of the array', function() {
+    assert.deepEqual(R.adjust(R.add(1), -3, [0, 1, 2, 3]), [0, 2, 2, 3]);
+  });
+
+  it('returns the original array if the supplied index is out of bounds', function() {
+    var list = [0, 1, 2, 3];
+    assert.deepEqual(R.adjust(R.add(1), 4, list), list);
+    assert.deepEqual(R.adjust(R.add(1), -5, list), list);
+  });
+
+  it('does not mutate the original array', function() {
+    var list = [0, 1, 2, 3];
+    assert.deepEqual(R.adjust(R.add(1), 2, list), [0, 1, 3, 3]);
+    assert.deepEqual(list, [0, 1, 2, 3]);
+  });
+
+  it('curries the arguments', function() {
+    assert.deepEqual(R.adjust(R.add(1))(2)([0, 1, 2, 3]), [0, 1, 3, 3]);
+  });
+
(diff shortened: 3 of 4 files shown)


src/adjust.js
```diff
@@ -0,0 +1,30 @@
...
+ * @param {Function} f The function to apply.
+ * @param {Number} i The index.
+ * @param {Array} xs The array whose value at the supplied index will be replaced.
+ * @return {Array} A copy of the supplied array with the adjusted element at index `i`.
+ * @example
+ *
+ *      R.adjust(R.add(10), 1, [0, 1, 2]);     //=> [0, 11, 2]
+ *      R.adjust(R.add(10))(1)([0, 1, 2]);     //=> [0, 11, 2]
+ */
+module.exports = _curry3(function(f, i, xs) {
```
buzzdecafe: house style would be `fn, idx, list`

scott-christopher: I've updated the param names as suggested.

CrossEye:
Since there was clear (and understandable) resistance to folding this into `assoc`, I'm all for this.  It's clearly a useful feature.  :herb:

test/adjust.js
```diff
@@ -0,0 +1,29 @@
...
+  it('should not mutate the original array', function() {
+    var list = [0, 1, 2, 3];
+    assert.deepEqual(R.adjust(R.add(1), 2, list), [0, 1, 3, 3]);
+    assert.deepEqual(list, [0, 1, 2, 3]);
+  });
+
+  it('should curry the arguments', function() {
+    assert.deepEqual(R.adjust(R.add(1))(2)([0, 1, 2, 3]), [0, 1, 3, 3]);
+  });
+});
```
CrossEye: We are trying to get away from the `it('should apply...')` style and use the more direct `it('applies ...')`.  While we haven't necessarily gone back and fixed all the old tests, it would be nice not to introduce new ones like this.  (I may have missed this in the Lenses commit.)

scott-christopher: As suggested, I've updated both test file descriptions to remove 'should'. I'll raise a PR to update the Lens tests as well.

CrossEye:
LGTM

TheLudd:
Very nice addition imho

src/update.js
```diff
@@ -0,0 +1,24 @@
...
+ * @param {*} x The value to exist at the given index of the returned array.
+ * @param {Array} list The source array to be updated.
+ * @return {Array} A copy of `list` with the value at index `idx` replaced with `x`.
+ * @example
+ *
+ *      R.update(1, 11, [0, 1, 2]);     //=> [0, 11, 2]
+ *      R.update(1)(11)([0, 1, 2]);     //=> [0, 11, 2]
+ */
+module.exports = _curry3(function(idx, x, list) {
+  return adjust(always(x), idx, list);
```
megawac: I would do it the other way around probably

buzzdecafe: you mean define `adjust` in terms of `update`?

test/update.js
```diff
@@ -0,0 +1,29 @@
...
+describe('update', function() {
+  it('updates the value at the given index of the supplied array', function() {
+    assert.deepEqual(R.update(2, 4, [0, 1, 2, 3]), [0, 1, 4, 3]);
+  });
+
+  it('offsets negative indexes from the end of the array', function() {
+    assert.deepEqual(R.update(-3, 4, [0, 1, 2, 3]), [0, 4, 2, 3]);
+  });
+
+  it('returns the original array if the supplied index is out of bounds', function() {
```
buzzdecafe: is this the behavior we want? i'm concerned about this creating obscure bugs. I don't know what a good alternative is though; throw when out of bounds?

test/update.js
```diff
@@ -0,0 +1,29 @@
...
+var R = require('..');
+
+describe('update', function() {
+  it('updates the value at the given index of the supplied array', function() {
+    assert.deepEqual(R.update(2, 4, [0, 1, 2, 3]), [0, 1, 4, 3]);
+  });
+
+  it('offsets negative indexes from the end of the array', function() {
+    assert.deepEqual(R.update(-3, 4, [0, 1, 2, 3]), [0, 4, 2, 3]);
+  });
```
buzzdecafe: what if the abs value of the negative index is greater than the list length?

scott-christopher: This is tested in the "out of bounds" test below.

test/update.js
scott-christopher: If other partial functions in Ramda throw when given arguments outside their domain, then I'm happy to modify it for consistency.

buzzdecafe: i suppose original list makes sense

buzzdecafe:
:cow2:

CrossEye:
@davidchambers: any feedback?

src/adjust.js
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Applies a function to the value at the given index of an array,
+ * returning a new copy of the array with the element at the given
+ * index replaced with the result of the function application.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig (a -> a) -> Int -> [a] -> [a]
```
davidchambers: We use `Number` rather than `Int` elsewhere.

src/adjust.js
```diff
@@ -0,0 +1,30 @@
...
+ *      R.adjust(R.add(10), 1, [0, 1, 2]);     //=> [0, 11, 2]
+ *      R.adjust(R.add(10))(1)([0, 1, 2]);     //=> [0, 11, 2]
+ */
+module.exports = _curry3(function(fn, idx, list) {
+  if (idx >= list.length || idx < -list.length) {
+    return list;
+  }
+  var start = idx < 0 ? list.length : 0;
+  var _idx = start + idx;
+  var _list = list.concat();
```
davidchambers: This will fail for array-like objects which don't provide a `concat` method (such as arguments objects).

test/update.js
davidchambers: We should use `deepEqual` rather than `strictEqual` for the two assertions below. We needn't guarantee that the argument is returned in these cases.

CrossEye: Agreed.  Reference equality shouldn't be part of the contract.

src/adjust.js
scott-christopher: If you want to support array-like objects, is there an existing function I should use for shallow cloning, or should I just use `Array.prototype.concat.call(list)` here?

buzzdecafe: try internal/_concat

scott-christopher: I've updated both `adjust` and `update` to accept array-like objects and added some tests to validate this behaviour.

src/adjust.js
scott-christopher: Updated as suggested.

test/update.js
scott-christopher: Good catch. I've switched the assertion to deepEqual in both "out of bounds" tests.

buzzdecafe:
:cow2:

CrossEye:
Thank you very much.  This is great stuff!

scott-christopher:
You're welcome, @CrossEye. Glad I could contribute to a great project. :)

kedashoe:
Why does haskell put the index in different argument positions for these two functions? Do we want to mimic that here?

scott-christopher:
My reasoning behind following the `Data.Sequence` parameter order was that when using key/value pairs such as in `update`, the key typically comes before the value. The reason for reversing this with `adjust` was to allow it to be curried with a function as the first parameter.

I appreciate that it could be a point of confusion if users are using both functions interchangeably, so I'm happy to hear what others think about this, though if it's to be changed, it needs to be changed soon before a new version is released.

kedashoe:
> The reason for reversing this with adjust was to allow it to be curried with a function as the first parameter.

Makes sense. I guess I end up more often writing things like

    pipe(
        map(foo),
        map(adjust(bar, 2))
    )(listOfLists);

than needing to create one off functions that will be resused ala `var barAdjuster = adjust(bar);` Also if that is needed and the params are switched, still very easy in ramda: `var barAdjuster = adjust(__, bar);` In any event, struck me as odd, thanks for the reply :)

PR #974, opened 2015-03-26
add R.mean and R.median

davidchambers:
I'd also like to add a [mode][1] function, but since it should work with values other than numbers it's more of a cousin than a sibling of `R.mean` and `R.median`. I'll open a separate pull request for that function if and when this pull request is merged.

[1]: http://en.wikipedia.org/wiki/Mode_%28statistics%29

### DIFF
--- a/src/mean.js
+++ b/src/mean.js
@@ -1,0 +1,21 @@
+var _curry1 = require('./internal/_curry1');
+var sum = require('./sum');
+
+
+/**
+ * Returns the mean of the given list of numbers.
+ *
+ * @func
+ * @memberOf R
+ * @category Math
+ * @sig [Number] -> Number
+ * @param {Array} list
+ * @return {Number}
+ * @example
+ *
+ *      R.mean([2, 7, 9]); //=> 6
+ *      R.mean([]); //=> NaN
+ */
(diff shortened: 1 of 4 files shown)


src/mean.js
```diff
@@ -0,0 +1,21 @@
...
+ * @category Math
+ * @sig [Number] -> Number
+ * @param {Array} xs
+ * @return {Number}
+ * @example
+ *
+ *      R.mean([2, 7, 9]); //=> 6
+ *      R.mean([]); //=> NaN
+ */
+module.exports = _curry1(function mean(xs) {
```
CrossEye: Usually we'd use the name `list`.  Did you want `xs` to distinguish a list of numeric values

CrossEye:
These look good.  `mode` is to my mind significantly more challenging.  How do you capture equality for non-primitives?  What do you do when there is more than one cohort of maximal value?

But that would be for another PR... :herb:

buzzdecafe:
:cow2: :accept:

src/mean.js
davidchambers: :zap:

davidchambers:
> But that would be for another PR...

#981

PR #978, opened 2015-03-28
Changing isNaN to eqNaN

### LINKED ISSUE #977: isNaN('blue') is false
isNaN seems broken: http://bit.ly/1xirZyx
`R.isNaN('blue')` returns false

The fun thing is that if you "install" all functions in ramda to `window` to avoid having to type `R.` all the time, `R.isNaN` will replace window.isNaN and all libraries that uses isNaN might behave a bit irrationally :)

CrossEye:
Closes #977


dist/ramda.js
```diff
@@ -932,9 +932,9 @@
      * @return {*} the first argument if falsy otherwise the second argument.
      * @example
      *
-     *      R.and(false, true); //=> true
-     *      R.and(0, []); //=> []
-     *      R.and(null, ''); => false
+     *      R.and(false, true); //=> false
+     *      R.and(0, []); //=> 0
+     *      R.and(null, ''); => null
```
davidchambers: Don't commit changes to this file.

davidchambers:
I like the current name. JavaScript has `isNaN` and `Number.isNaN`, and `R.isNaN` is clearly a member of the family. I don't find the argument for renaming the function compelling, but if we do decide to avoid naming collisions I suggest removing the function and suggesting people use `R.eq(NaN)` instead.

dist/ramda.js
CrossEye: How the hell did that get back in there?

buzzdecafe:
removing and using `eq(NaN)` looks good to me

CrossEye:
That would be fine with me too.  This function is a little specialized in any case.  Since `eq` is actually built to handle this, I think that may be best.

CrossEye:
Will open another one soon to remove the function and suggest `eq(NaN)`.  I suppose with a deprecation first.

PR #979, opened 2015-03-29
Path should return entire object for empty path

paldepind:
Today `_.path([], someObj)` returns `undefined`. This seems inconsistent to me. If you take zero steps down, you're still at the top.

This modifies the internal `_path` function so that `_.path([], obj)` returns the passed in object and so that `_.pathEq([], obj, obj) === true`.

This PR simply removes a special case in the code. This, I think, goes to show that the current behavior is inconsistent, i.e. it is not how a user would guess the function would be working.

If the current behavior was implemented with a good reason I'm sorry for this PR :)

### DIFF
--- a/test/path.js
+++ b/test/path.js
@@ -22,7 +22,7 @@ describe('path', function() {
       j: ['J']
     };
     assert.strictEqual(R.path(['a', 'b', 'c'], obj), 100);
-    assert.strictEqual(R.path([], obj), undefined);
+    assert.strictEqual(R.path([], obj), obj);
     assert.strictEqual(R.path(['a', 'e', 'f', '1'], obj), 101);
     assert.strictEqual(R.path(['j', '0'], obj), 'J');
     assert.strictEqual(R.path(['j', '1'], obj), undefined);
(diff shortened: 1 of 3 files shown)


buzzdecafe:
makes sense.

:cow2:

CrossEye:
:herb:

test/path.js
```diff
@@ -5,7 +5,7 @@ var R = require('..');
 
 describe('path', function() {
   var deepObject = {a: {b: {c: 'c'}}, falseVal: false, nullVal: null, undefinedVal: undefined, arrayVal: ['arr']};
-  it('takes a path and an object and returns the value at the path or undefined', function() {
+  it('takes a path and an object and returns the value at the path or the object', function() {
```
davidchambers: This wording change should be reverted. Your argument is that the object *is* the value at the path when the path is `[]`. Also, the function does still return undefined if the path does not exist.

paldepind:
You're right David. I didn't think about that when changing the text.

davidchambers:
:deciduous_tree:

PR #980, opened 2015-03-29
replace === with R.eq in equality functions

davidchambers:


buzzdecafe:
i think this will go a long way to ensuring consistent behavior.

buzzdecafe:
... and :cow2:

CrossEye:
:herb:

PR #981, opened 2015-03-30
add R.modes

davidchambers:
One of the test cases will fail until #980 is merged.

### DIFF
--- a/test/modes.js
+++ b/test/modes.js
@@ -1,0 +1,30 @@
+var assert = require('assert');
+
+var R = require('..');
+
+
+describe('modes', function() {
+
+  it('returns the most common values', function() {
+    assert.deepEqual(R.modes([]), []);
+    assert.deepEqual(R.modes([1, 2, 2, 2, 1]), [2]);
+    assert.deepEqual(R.modes([1, 2, 3, 2, 1]), [1, 2]);
+  });
+
+  it('determines object equality by identity', function() {
+    var a = {};
+    var m = {};
+    var d = {};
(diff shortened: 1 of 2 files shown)


test/modes.js
```diff
@@ -0,0 +1,30 @@
...
+    assert.deepEqual(R.modes([]), []);
+    assert.deepEqual(R.modes([1, 2, 2, 2, 1]), [2]);
+    assert.deepEqual(R.modes([1, 2, 3, 2, 1]), [1, 2]);
+  });
+
+  it('determines object equality by identity', function() {
+    var a = {};
+    var m = {};
+    var d = {};
+    var result = R.modes([R, a, m, d, a]);
```
buzzdecafe: nice

buzzdecafe:
this is fine, but it makes me wonder if for completeness we need a `modeWith(eqFn, list)` where the user can supply a comparator that defines equality.

CrossEye:
While I have no major objection, is this being done out of any real need?  Or is it simply a case of, "We have `mean` and `median`, obviously we should also have `mode`?"  I ask because I have degrees in mathematics (though not specifically statistics) and have used mean and median often.  I don't think I've ever in my life had an actual need for the mode of a data set.

For the same reason, if we do include this, I would likely hold off on `modeWith` unless and until we get a call for it.  I agree that there's a nice consistency there, but I just don't see it likely to get used much.

davidchambers:
> While I have no major objection, is this being done out of any real need?

No. Let's close this pull request and revisit it if and when a user requests such a function.

PR #982, opened 2015-03-30
DifferenceWith - Fixing the array position that output items are placed in

andyleggett:
The previous behaviour placed items one after the other.  The newer version created a sparse array with items placed at their original locations.


buzzdecafe:
oops. looks like this got introduced in b7a8f67b7237d10f09caf1a4c899668c7458f96d. thanks for catching this.

:cow2:

andyleggett:
No worries - had me scratching my head for a bit this morning. Glad I could help out.

buzzdecafe:
just waiting to see if there is any other input, but this looks pretty cut-n-dried to me. we should get this in and publish a version.

CrossEye:
:herb:

Aourin:
I have been fighting this all day and just found it too. I am glad you guys got a fix in already though.

PR #985, opened 2015-03-30
have R.converge return a function of appropriate length

davidchambers:


src/converge.js
```diff
@@ -31,10 +34,10 @@ var curryN = require('./curryN');
  */
 module.exports = curryN(3, function(after) {
   var fns = _slice(arguments, 1);
-  return function() {
+  return arity(max(pluck('length', arguments)), function() {
```
buzzdecafe: as with `uncurryN`, should this use `curryN`?

CrossEye: Probably yes for `curryN`, but one other thing:

```javascript
return arity(max(pluck('length', _slice(arguments, 1))), function() {  // or
return curryN(max(pluck('length', _slice(arguments, 1))), function() {
```

The first function -- the join -- should not count when calculating the length of the returned function.

buzzdecafe: simpler:

`return curryN(max(pluck('length', fns)), function() {`

CrossEye: Yes.   A bit myopic here.

davidchambers: > as with `uncurryN`, should this use `curryN`?

Yes, it should. Good catch. :)

> The first function -- the join -- should not count when calculating the length of the returned function.

Thank you for catching this error, @CrossEye! I've updated the code and the relevant test case.

buzzdecafe:
:coffee:

CrossEye:
:herb:

anton-rudeshko:
:+1:

PR #880, opened 2015-03-06
deprecate R.I

davidchambers:
As discussed in #859. We were 2–1 in favour of `R.identity`; this pull request deprecates `R.I`. Those who'll miss `R.I` can console themselves by defining an :alien:: `var I = R.identity;`.

PR #988, opened 2015-03-30
deprecate isNaN

buzzdecafe:
in favor of `R.eq(NaN)`


src/isNaN.js
```diff
@@ -6,6 +6,7 @@ var _curry1 = require('./internal/_curry1');
  *
  * Equivalent to ES6's [`Number.isNaN`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN).
  *
+ * @deprecated since 0.14.0
```
davidchambers: I've included the `v` prefix in other cases. See #880, for example.

CrossEye:
On my agenda for this evening.  Thanks.

PR #991, opened 2015-03-31
docs: Improve merge documentation

rpwjanzen:
Add example indicating behaviour with values set to 'undefined'

### DIFF
--- a/src/merge.js
+++ b/src/merge.js
@@ -19,6 +19,9 @@ var _extend = require('./internal/_extend');
  *      R.merge({ 'name': 'fred', 'age': 10 }, { 'age': 40 });
  *      //=> { 'name': 'fred', 'age': 40 }
  *
+ *      R.merge({ 'name': 'fred', 'age': 10 }, { 'age': undefined });
+ *      //=> { 'name': 'fred', 'age': undefined }
+ *
  *      var resetToDefault = R.merge(R.__, {x: 0});
  *      resetToDefault({x: 5, y: 2}); //=> {x: 0, y: 2}
  */


src/merge.js
```diff
@@ -19,6 +19,9 @@ var _extend = require('./internal/_extend');
  *      R.merge({ 'name': 'fred', 'age': 10 }, { 'age': 40 });
  *      //=> { 'name': 'fred', 'age': 40 }
  *
+ *      R.merge({ 'name': 'fred', 'age': 10 }, { 'age': undefined });
+ *      //=> { 'name': 'fred', 'age': undefined }
```
davidchambers: Perhaps [`R.dissoc`][1] is more appropriate in this case.


[1]: http://ramdajs.com/docs/#dissoc

rpwjanzen: I'm not sure what you mean. Are you suggesting that the added docs should be moved (and reworked if applicable) to `R.dissoc`?  The purpose of this commit is to clarify the behavior of `R.merge` with respect to any items of higher precedence that contain fields set to `undefined`.  Should this behavior be left undocumented?  Or do you mean something else?

davidchambers: > The purpose of this commit is to clarify the behavior of `R.merge` with respect to any items of higher precedence that contain fields set to `undefined`.

I'm curious as to why one might expect `undefined` to be treated differently from other values. Perhaps some other libraries have set a bad precedent. ;)

CrossEye: > I'm curious as to why one might expect `undefined` to be treated differently from other values. Perhaps some other libraries have set a bad precedent. ;)

That, and JS itself is extremely schizophrenic about `undefined`.  It's the value that says, "there's no value, damn it, never has been and perhaps never will be."  Its cousin, `null`, is the more sane one that simply says, "this space intentionally left blank".

But I agree with the overall point that there's no compelling reason to document that `{age: undefined}` will override an `age` field with another value.  Why would one expect otherwise?

CrossEye: Adding an `@see`  link to `assoc` and especially `dissoc` does sound appropriate, though.

rpwjanzen:
@davidchambers I've seen "other libraries set a bad precedent".  In light of your's and @CrossEye's  comments above, I'm closing this PR.

davidchambers:
I appreciate your effort, @rpwjanzen. Please continue to draw attention to parts of the documentation which could be improved. :)

CrossEye:
And to anything else you notice!

Thank you for your attention to detail.

PR #998, opened 2015-04-01
`adjust` and `assoc` now returns the same instance if the value does not change

scott-christopher:
This prevents an unnecessary clone of the array and allows users to perform strict equality checks of the resulting array to easily determine whether the values of the array have changed.

### DIFF
--- a/src/adjust.js
+++ b/src/adjust.js
@@ -1,5 +1,7 @@
 var _concat = require('./internal/_concat');
 var _curry3 = require('./internal/_curry3');
+var _eq = require('./internal/_eq');
+
 
 /**
  * Applies a function to the value at the given index of an array,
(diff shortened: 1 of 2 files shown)


CrossEye:
:herb:

davidchambers:
We should discourage dependence on object identity. I don't mind if we happen to return an argument rather than an equivalent object in some cases, but I certainly don't think we should document this.

scott-christopher:
If this behaviour isn't documented, couldn't users be lead to think that a new array instance is always returned, causing issues if the returned array is passed off to some external function that mutates arrays?

davidchambers:
The question is whether Ramda should defensively clone objects. My view is that it should not, as Ramda is philosophically opposed to functions which mutate their arguments and does not provide any. Cloning is the responsibility of the programmer who chooses to use such functions.

Perhaps I'm being overly idealistic. :)

CrossEye:
This change is about avoiding a clone operation when it's not necessary.  It's an interesting point about the documentation.  `update` can be seen as a parallel to `assoc`, but `assoc` so far always returns a new object.  And `assoc` is specifically documented beginning with "makes a shallow clone of an object".  Perhaps the same optimization could be applied there, and we should drop that part of the documentation.  That would probably be the most logically consistent way to go.  And the documentation would simply remain silent about this, assuming it's an implementation detail.

scott-christopher:
I've updated the change to also include the same shortcut for `assoc` and remove the docs describing the implementation behaviour as suggested.

CrossEye:
I'm fairly happy with this idea as a slight memory optimization and as an extension of the idea that Ramda should only ever care about value equality and never reference equality.

But I'm not sure we want to commit ourselves to it with unit test or other forms of documentation.  If we treat this only as an internal optimization that we could later choose to alter, we are less constrained, and we force fewer assumptions onto our users.

What would others think about keeping these changes, but leaving out the unit tests that mention them?

davidchambers:
> What would others think about keeping these changes, but leaving out the unit tests that mention them?

I like this idea.

scott-christopher:
I've removed all docs and tests mentioning the behaviour.

davidchambers:
:deciduous_tree:

Issue #993, opened 2015-04-01
filterObj?


donnut:
I have the following object:
```javascript
let storage = {
  11: {continue: true},
  21: {continue: true},
  31: {continue: false},
  41: {continue: true}
}
```
The number if occurrences of truth values of continue are of interest. 
I was thinking to use 
```javascript
return R.length(
    R.filter(function(v) {
        return v.continue === true;
    }, R.values(storage))
);
```
But what if I want to have all key/value pairs that equal the condition? So:
```js
{
  11: {continue: true},
  21: {continue: true},
  41: {continue: true}
}
```
Would this be an option?
```javascript
filterObj(function(val, key, obj) {
    return val.continue === true;
}, storage):
```
Or is there already a (efficient) function combination that does this?

edit: updated `return val` to `return val.continue`

buzzdecafe:
`filterObj` seems reasonable to me. I don't see a simple solution for this, and key order is immaterial.

TheLudd:
:horse:

CrossEye:
I might write your first example differently.  I would definitely pass this to `filter`:

    R.where({continue: true})

And I might go all out and use this:

    var fn = R.pipe(R.values, R.filter(R.where({continue: true})), R.length);

For the more general problem (I assume you want `return val.continue === true`), I don't see any simple combination that generates what you're looking for.  Obviously something could be done involving a cycle of `toPairs` -> `filter` -> `fromPairs`, but that filter step would be a fairly ugly wrapper on the filter predicate supplied.

This does not fall prey to the issues described whenever we bring up a `foldObj` implementation.  Since the result is an object, it shouldn't matter in what order the keys are added.  I wouldn't mind seeing a PR for this, but I would also like to see some argument for its general usefulness.  Do you have a reasonably easy-to-describe real-world scenario?

CrossEye:
I start typing a response, go away for a few minutes, come back to finish it up and find two others in ahead of me!  And as always, much more concise than me.  Especially the horse!

buzzdecafe:
yes, i think :horse: is a "Godfather" reference, so we better include this function ....

TheLudd:
Just trying to adapt to the current jargon =)

donnut:
I'll make a PR. That's the easy part. The difficult part is to squash is into one commit :(

TheLudd:
@donnut I am happy to teach. Mastering git rebase is a joy :)

TheLudd:
[...]

PR #1004, opened 2015-04-04
filter objects

donnut:
A proposal for filtering objects.

### DIFF
--- a/src/filterObj.js
+++ b/src/filterObj.js
@@ -1,0 +1,35 @@
+var _curry2 = require('./internal/_curry2');
+var keys = require('./keys');
+
+
+/**
+ * Returns a new object containing only those key-value pairs of which the
+ * values match a given predicate function.
+ * The predicate function is passed one argument: *(value)*.
+ *
+ * @func
+ * @memberOf R
+ * @category Object
+ * @sig (a -> Boolean) -> {String, a} -> {String, a}
+ * @param {Function} fn The function called per key.
+ * @param {Object} obj The object to iterate over the keys.
+ * @return {Object} The new filtered object.
+ * @example
+ *
+ *      var isPositive = function(n) {
+ *        return n > 0;
+ *      };
+ *      R.filterObj(isPositive, {a: 1, b: 2, c: -1, d: 0, e: 5}); //=> {a: 1, b: 2, e: 5}
+ *
+ *      var colors = {1: {color: 'red'}, 2: {color: 'black', bgcolor: 'yellow'}};
+ *      R.filterObj(has('bgcolor'), colors); //=> {2: {color: 'black', bgcolor: 'yellow'}}
+ */
+module.exports = _curry2(function filterObj(fn, obj) {
+  var result = {}, ks = keys(obj), idx = ks.length;
+  while (--idx >= 0) {
+    if (fn(obj[ks[idx]])) {
+      result[ks[idx]] = obj[ks[idx]];
+    }
+  }
+  return result;
+});
--- a/test/filterObj.js
+++ b/test/filterObj.js
@@ -1,0 +1,37 @@
+var assert = require('assert');
+
+var R = require('..');
+
+
+describe('filterObj', function() {
+  var even = function(x) { return x % 2 === 0; };
+  var proceed = function(x) { return x.proceed; };
+
+  it('reduces an object with simple values to those matching a filter', function() {
+    assert.deepEqual(R.filterObj(even, {a:1, b:2, c:3, d:4, e:5}), {b:2, d:4});
+  });
+
+  it('reduces an object with object values to those matching a filter', function() {
+    var obj = {1: {proceed: true}, 2: {proceed: true, name: 'solist'}, 3: {proceed: false}, 4: {proceed: true}};
+    var reducedObj = R.filterObj(proceed, obj);
+    assert.deepEqual(reducedObj, {1: {proceed: true}, 2: {proceed: true, name: 'solist'}, 4: {proceed: true}});
+  });
+
+  it('reduces an object with null and undefined values to those matching a filter', function() {
+    assert.deepEqual(R.filterObj(proceed, {1: {proceed: null}, 2: {proceed: void 0}, 3: {proceed: false}, 4: {proceed: true}}),
+      {4: {proceed: true}});
+  });
+
+  it('returns an empty object if no item matches', function() {
+    assert.deepEqual(R.filterObj(function(x) { return x > 100; }, {a:1, b:9, c:99}), {});
+  });
+
+  it('returns an empty object if asked to filter an empty object', function() {
+    assert.deepEqual(R.filterObj(function(x) { return x > 100; }, {}), {});
+  });
+
+  it('is curried', function() {
+    var onlyEven = R.filterObj(even);
+    assert.deepEqual(onlyEven({a:1, b:2, c:3, d:4, e:5}), {b:2, d:4});
+  });
+});
--- a/src/filterObjIndexed.js
+++ b/src/filterObjIndexed.js
@@ -1,0 +1,32 @@
+var _curry2 = require('./internal/_curry2');
+var keys = require('./keys');
+
+
+/**
+ * Like `filterObj`, but passes additional arguments to the predicate function.
+ * The predicate function is passed three arguments: *(value, key, obj)*.
+ *
+ * @func
+ * @memberOf R
+ * @category Object
+ * @sig (a, String, {String, a} -> Boolean) -> {String, a} -> {String, a}
+ * @param {Function} fn The function called per key.
+ * @param {Object} obj The object to iterate over the keys.
+ * @return {Object} The new filtered object.
+ * @example
+ *
+ *      var isUpperCase = function(val, key) { return key.toUpperCase() === key; };
+ *      var isUpperCasePositive = function(value, key) {
+ *        return value > 0 && isUpperCase(key);
+ *      };
+ *      R.filterObjIndexed(isUpperCasePositive, {a: 1, B: 2, c: -1, D: 0, e: 5}); //=> {B: 2}
+ */
+module.exports = _curry2(function filterObjIndexed(fn, obj) {
+  var result = {}, ks = keys(obj), idx = ks.length;
+  while (--idx >= 0) {
+    if (fn(obj[ks[idx]], ks[idx], obj)) {
+      result[ks[idx]] = obj[ks[idx]];
+    }
+  }
+  return result;
+});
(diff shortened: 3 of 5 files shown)


src/filterObj.js
```diff
@@ -0,0 +1,29 @@
...
+ *      };
+ *      R.filterObj(isPositive, {a: 1, b: 2, c: -1, d: 0, e: 5}); //=> {a: 1, b: 2, e: 5}
+ *
+ *      var containsBackground = function(x) {
+ *        return x.bgcolor;
+ *      };
+ *      var colors = {1: {color: 'read'}, 2: {color: 'black', bgcolor: 'yellow'}};
+ *      R.filterObj(containsBackground, colors); //=> {2: {color: 'black', bgcolor: 'yellow'}}
+ */
+module.exports = _curry2(_filterObj);
```
buzzdecafe: unless `_filterObj` is being reused somewhere else, might as well inline the definition here

src/internal/_filterObj.js
```diff
@@ -0,0 +1,11 @@
+var _baseCopy = require('./_baseCopy');
+
+module.exports = function _filterObj(fn, obj) {
+  var result = {};
+  for (var key in obj) {
```
buzzdecafe: this will descend the whole proto chain, yes? maybe use `R.keys` here instead, as in `mapObj`

src/filterObj.js
donnut: I will inline it.

src/internal/_filterObj.js
donnut: Yes, which is not intended.

src/filterObj.js
```diff
@@ -0,0 +1,29 @@
+var _curry2 = require('./internal/_curry2');
+var _filterObj = require('./internal/_filterObj');
+
+/**
```
davidchambers: Nit: Two empty lines after imports, please. :)

src/filterObj.js
```diff
@@ -0,0 +1,29 @@
...
+ * Returns a new object containing only those key-value pairs of which the
+ * values match a given predicate function.
+ * The predicate function is passed one argument: *(value)*.
+ *
+ * @func
+ * @memberOf R
+ * @category Object
+ * @sig (v -> Boolean) -> {k, v} -> {k, v}
```
davidchambers:
```diff
- * @sig (v -> Boolean) -> {k, v} -> {k, v}
+ * @sig (a -> Boolean) -> {String: a} -> {String: a}
```

src/filterObj.js
```diff
@@ -0,0 +1,29 @@
...
+ *      var isPositive = function(n) {
+ *        return n > 0;
+ *      };
+ *      R.filterObj(isPositive, {a: 1, b: 2, c: -1, d: 0, e: 5}); //=> {a: 1, b: 2, e: 5}
+ *
+ *      var containsBackground = function(x) {
+ *        return x.bgcolor;
+ *      };
+ *      var colors = {1: {color: 'read'}, 2: {color: 'black', bgcolor: 'yellow'}};
+ *      R.filterObj(containsBackground, colors); //=> {2: {color: 'black', bgcolor: 'yellow'}}
```
davidchambers: This is more precise and, in my view, clearer:

```diff
- *      R.filterObj(containsBackground, colors); //=> {2: {color: 'black', bgcolor: 'yellow'}}
+ *      R.filterObj(R.has('bgcolor'), colors); //=> {2: {color: 'black', bgcolor: 'yellow'}}
```

src/filterObj.js
```diff
@@ -0,0 +1,29 @@
...
+ *
+ *      var isPositive = function(n) {
+ *        return n > 0;
+ *      };
+ *      R.filterObj(isPositive, {a: 1, b: 2, c: -1, d: 0, e: 5}); //=> {a: 1, b: 2, e: 5}
+ *
+ *      var containsBackground = function(x) {
+ *        return x.bgcolor;
+ *      };
+ *      var colors = {1: {color: 'read'}, 2: {color: 'black', bgcolor: 'yellow'}};
```
davidchambers: Presumably this should be `'red'` rather than `'read'`.

test/filterObj.js
```diff
@@ -0,0 +1,39 @@
+var assert = require('assert');
+
+var R = require('..');
+
+
+describe('filterObj', function() {
+  var even = function(x) {return x % 2 === 0;};
+  var proceed = function(x) {return x.proceed;};
```
davidchambers: Nit: Please include spaces inside the curlies.

src/filterObj.js
CrossEye:
```javascript
 var colors = {1: {color: 'read'}, 2: {color: 'bleak', bgcolor: 'mellow'}};
```

??

src/filterObj.js
CrossEye: :+1:

CrossEye:
I like this overall, but wonder if the callback should accept `(value, key, obj)`

buzzdecafe:
> I like this overall, but wonder if the callback should accept `(value, key, obj)`

No.

we have `mapObj` and `mapObjIndexed` for those separate concerns

davidchambers:
I'm with @buzzdecafe on this.

scott-christopher:
I don't think the Object is needed to make a decision in the predicate function, though it would make it potentially useful to include both the Key and the Value in the parameter list:
```
(a -> String -> Boolean) -> {String: a} -> {String: a}
```
Otherwise to filter an object based on the keys, you'd have to something a little clunkier such as:
```js
var secondEq42 = R.compose(R.eq('fortytwo'), R.nthArg(1))
R.compose(R.filterObj(R.identity), R.mapObjIndexed(secondEq42))
```
Rather than:
```js
var secondEq42 = R.compose(R.eq('fortytwo'), R.nthArg(1))
R.filterObj(secondEq42)
```

CrossEye:
That's probably enough reason not to do this.  But I wonder if we carried an analogy too far when we created `mapObj` and `mapObjIndexed`.  There are good reasons for `map`, `filter`, etc. not to carry the extra parameters.  Such functions have real meanings that extend beyond finite lists, to data structures for which there are no indices and possibly extend to systems for which there aren't even obvious objects to pass as a third parameter.  But `mapObj` doesn't have these concerns.  At least as far as I can see, there are no obvious extensions to other structures.

I suppose the old point about things like `parseInt` is reason enough for this, along with a slightly worn argument about consistency.  But I still feel as though if we're adding `filterObj`, we would probably also want the API of `filterObjIndexed`.  I guess I was just hoping for a two-for-one.

test/filterObj.js
donnut: @davidchambers you mean `{ return x.proceed }`?

donnut:
From the discussion and issue #993 I conclude it is a good thing to create filterObjIndexed too. I'm curious for insights in the remark in #993 why the results of e.g. `filter` and `propBy` still have a reference to the original array/object.

CrossEye:
@donnut If you're asking why `filter` and `pickBy` don't do some sort of deep cloning, then my answer is that we shouldn't ever be doing that unless specifically asked to do so.  We should care about value equality and not reference equality and act as though our users are working with immutable data as much as possible.  Obviously if we're asked to do something that is mutation-like (`assoc`, `dissoc`, `merge`, etc.) then we need to do some sort of cloning in order to not mutate user data, but otherwise we can use references wherever possible.

Or did I misunderstand your question?

buzzdecafe:
i don't see any compelling reason to deep clone when filtering.

donnut:
If deep cloning had no performance/memory penalty I suppose you would prefer functions like `filter`, `pickBy` and `filterObj` to create a new arrays or objects without any reference to the original. Am I correct?

CrossEye:
@donnut: No, the point is that we should never care.  If, as a user, I'm working with immutable data, then I see no reason to clone it when passing it around and don't want to pay the memory price for two copies of the same thing.  If I'm not and expect to mutate my objects, then I might object if a library instead passes clones of them where I think I'm getting references.

I simply see no advantage to doing a clone.  Am I missing something?

TheLudd:
Cloning should really be treated the same way as regular filter right? If I have an array of objects and do filter on them, are they / should they be cloned? I see no reason for it.

donnut:
@CrossEye I agree that cloning when filtering an object or array prevents users from working with mutable data and that is something that javascript developers won't expect and thus not wanted. When working with an immutable object I assume that it, well, doesn't mutate, whatever happens with it during filter, mapping, or whatever. And the good thing is that it isn't! I guess the 'problem' I have is that immutable objects and arrays in javascript just don't exist and save guards agains the effects of mutation (a referenced object is mutated too) are to costly, are not expected and are not wanted in many cases.

@TheLudd 
> Cloning should really be treated the same way as regular filter right? 

When filtering an array, the array is not cloned, so it is not handled the same. Is that what you mean?

In the current proposal of filterObj I included baseCopy and a test to 'prove' the filtered object was not referencing the original object. I'll remove that and send in a new pr. Shall I also make filterObjIndexed while I'm at it?

davidchambers:
`R.pickBy` should be deprecated as part of this change. It would be confusing to have two very similar functions with very different names.

donnut:
@davidchambers Yes, `filterObj` and `filterObjIndexed` makes `pickBy` redundant. Should I add

`* @deprecated since v0.14.0` 

after the two new functions are accepted?

buzzdecafe:
or `pickBy` makes `filterObjIndexed` unnecessary. `pickBy` is a much better name

davidchambers:
> Should I add
>
> `* @deprecated since v0.14.0`
>
> after the two new functions are accepted?

You could update this pull request to include this note.

> or `pickBy` makes `filterObjIndexed` unnecessary. `pickBy` is a much better name

Are you suggesting we have both `R.pickBy` and `R.filterObj`?

CrossEye:
As much as I agree that `pickBy` is the better name, and it does contribute to certain consistencies within the library  (`xxx` /  `xxxBy`), I think there is a stronger sense of the `[map|filter|reduce|mapObj]` / `[map|filter|reduce|mapObj]Indexed` pairing.  Moreover, `filterObjIndexed` adds a feature not found in the current `pickBy`, namely the full object parameter to the callback function.  As usual, I don't think that's often useful, but we supply it in the other similar functions, so I think it belongs here.

So I think it's time to deprecate `pickBy`.

donnut:
I added the deprecated warning in `pickBy`, although I'm not sure everybody agrees. Please let me know if this is ok, otherwise I'm happy to change it.

davidchambers:
Please `git rebase master`. :)

donnut:
done, with help of @TheLudd!

buzzdecafe:
in light of #1061 what is the status of this one? FWIW, I am still *strongly* :-1: on the name `filterObjIndexed`. Much better is `pickBy`. Much, *much* better.

donnut:
I'm fine with `pickBy`. The whole idea of `filterObj` started off with me overseeing `'pickBy`. Closing this PR.

PR #1006, opened 2015-04-04
update to transducer protocol. fixes #1003

### LINKED ISSUE #1003: Change to Official Transformer Protocol
Change transducers implementation to support "Official Transformer Protocol" discussed in cognitect-labs/transducers-js#20 .  Both transducers.js and transducers-js have been updated to support it (my libraries to follow soon).

Summary of changes:
```javascript
// anywhere you use a transformer
// (including transducer implementations)
var xf = {
   init: function()
   step: function(acc, item)
   result: function(acc)
}
// becomes
var xf = {
   '@@transducer/init': function()
   '@@transducer/step': function(acc, item)
   '@@transducer/result': function(acc)
}
// reduced wrapper object changes from 
{
   value: acc,
   __transducers_reduced__: true
}
// to
{
   '@@transducer/value': acc,
   '@@transducer/reduced':true
}
```
Obviously, there is room to generalize this.  This may be a good opportunity to reduce the redundancy and consider other transducer implementation strategies (other than `.prototype`) as @CrossEye and @megawac brought up in the initial transducers PR

buzzdecafe:
ping @kevinbeaty 

please review! thanks

### DIFF
--- a/src/internal/_reduced.js
+++ b/src/internal/_reduced.js
@@ -1,3 +1,7 @@
 module.exports = function(x) {
-  return x && x.__transducers_reduced__ ? x : {value: x, __transducers_reduced__: true};
+  return x && x['@@transducer/reduced'] ? x :
+    {
+      '@@transducer/value': x,
+      '@@transducer/reduced': true
+    };
 };
--- a/src/internal/_isTransformer.js
+++ b/src/internal/_isTransformer.js
@@ -1,3 +1,3 @@
 module.exports = function _isTransformer(obj) {
-  return typeof obj.step === 'function' && typeof obj.result === 'function';
+  return typeof obj['@@transducer/step'] === 'function';
 };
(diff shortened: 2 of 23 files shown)


src/internal/_reduced.js
```diff
@@ -1,3 +1,7 @@
 module.exports = function(x) {
-  return x && x.__transducers_reduced__ ? x : {value: x, __transducers_reduced__: true};
+  return x && x['@@transducer/reduced'] ? x :
```
davidchambers: For clarity, let's use:

```javascript
return x != null && ...
```

buzzdecafe: ok. i'll know that i have arrived when i get *no* review comments from you :smile:

davidchambers: This one doesn't really count, as the comment relates to code not modified in this pull request.

kevinbeaty:
@buzzdecafe Looks good to me.  Thanks

CrossEye:
:herb:

davidchambers:
:deciduous_tree:

Issue #890, opened 2015-03-08
replace all uses of Array.prototype.push with index assignment


davidchambers:
Added to the Google JavaScript Style Guide in [July 2010][1]:

> You can also use an array as a stringbuilder, and convert it into a string with `myArray.join('')`.  Note that since assigning values to an array is faster than using `push()` you should use assignment where possible.

It's worth noting that this commit was made almost five years ago. The section above was removed in [November 2012][2].

What's the story in 2015? It looks like [much of a muchness][3] to me.

/cc @megawac

[1]: https://code.google.com/p/google-styleguide/source/detail?r=52
[2]: https://code.google.com/p/google-styleguide/source/detail?r=97
[3]: http://jsperf.com/index-vs-push/6

davidchambers:
@AutoSponge, in #592:

> Approximately 30% speed gain on filter benchmarks.

Before #887:

    Running suite filter [lib/bench/filter.bench.js]...
    >> _.filter(nums, isEven) x 2,735,544 ops/sec ±0.38% (102 runs sampled)
    >> filter(isEven, nums) x 2,483,532 ops/sec ±0.42% (97 runs sampled)
    >> filter(isEven)(nums) x 1,822,110 ops/sec ±3.81% (94 runs sampled)
    >> filterEven(nums) x 2,300,862 ops/sec ±5.15% (93 runs sampled)
    Fastest test is _.filter(nums, isEven) at 1.10x faster than filter(isEven, nums)

After #887:

    Running suite filter [lib/bench/filter.bench.js]...
    >> _.filter(nums, isEven) x 2,670,441 ops/sec ±0.32% (102 runs sampled)
    >> filter(isEven, nums) x 648,427 ops/sec ±0.45% (100 runs sampled)
    >> filter(isEven)(nums) x 602,343 ops/sec ±0.51% (98 runs sampled)
    >> filterEven(nums) x 618,233 ops/sec ±2.66% (93 runs sampled)
    Fastest test is _.filter(nums, isEven) at 4.1x faster than filter(isEven, nums)

That's significant. I'll open a pull request to replace all uses of `Array.prototype.push`.

megawac:
Usually push is faster from my experience - at least it was 3 months ago in
v8 and ff

PR #1007, opened 2015-04-05
Add R.dropRepeats and R.dropRepeatsWith

### LINKED ISSUE #1005: Add dedupe/dropRepeats function
I think Ramda needs a function that drops consecutive repeated values from an array.

```javascript
R.dropRepeats([1, 1, 2, 3, 4, 4, 4, 3, 1]); //=> [1, 2, 3, 4, 3, 1]
```

This is not `uniq`. This function only removes _consecutive_ repeated values. This function can be implemented more efficiently and is useful for reducing a dataset only to its changes. For instance, if you have an array with some value sampled every second and you want to reduce the dataset to only the points at which the sampled value actually changed.

Furthermore this function can be implemented as a transducer (transducers.js actually includes a function like this).

My own use case for a function like this is to achieve something similar to Elms [dropRepeats](http://package.elm-lang.org/packages/elm-lang/core/1.1.1/Signal#dropRepeats) using [Flyd](https://github.com/paldepind/flyd)s transducer support.

If there is support for this I'll be happy to create a pull request with the suggested function.

paldepind:
This closes #1005.

I've tried to conform to the existing code patterns, but there a few things to note:

* I've added a test for the transducer version by using `R.into`. I could hardly find any tests for the existing transducer compatible functions. Maybe I've missed something.
* The non-transducer implementation is not based on the transducer implementation. This seems to be in line with the existing transducer compatible functions.
* The non-transducer implementation is meant to be performant and I've used `Array#push`. There is however no other Ramda function that uses push. Is there a reason for that? It is a fast way to build an array and it is supported by all browsers.
* The non-`With` version is simply based on the `With`-version. I couldn't find any other functions implemented on top of their `With` companion in this way but it seems logical so I did it anyway.

What do you think?

### DIFF
--- a/src/dropRepeats.js
+++ b/src/dropRepeats.js
@@ -1,0 +1,24 @@
+var _curry1 = require('./internal/_curry1');
+var _dispatchable = require('./internal/_dispatchable');
+var _eq = require('./internal/_eq');
+var _xdropRepeatsWith = require('./internal/_xdropRepeatsWith');
+var dropRepeatsWith = require('./dropRepeatsWith');
+
+
+/**
+ * Returns a new list without any consecutively repeating elements.
+ *
+ * Acts as a transducer if a transformer is given in list position.
+ * @see R.transduce
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig [a] -> [a]
+ * @param {Array} list The array to consider.
+ * @return {Array} `list` without repeating elements.
+ * @example
+ *
+ *     R.dropRepeats([1, 1, 1, 2, 3, 4, 4, 2, 2]); //=> [1, 2, 3, 4, 2]
+ */
+module.exports = _curry1(_dispatchable('dropRepeats', _xdropRepeatsWith(_eq), dropRepeatsWith(_eq)));
--- a/src/dropRepeatsWith.js
+++ b/src/dropRepeatsWith.js
@@ -1,0 +1,42 @@
+var _curry2 = require('./internal/_curry2');
+var _dispatchable = require('./internal/_dispatchable');
+var _xdropRepeatsWith = require('./internal/_xdropRepeatsWith');
+var last = require('./last');
+
+
+/**
+ * Returns a new list without any consecutively repeating elements. Equality is
+ * determined by applying the supplied predicate two consecutive elements.
+ * The first element in a series of equal element is the one being preserved.
+ *
+ * Acts as a transducer if a transformer is given in list position.
+ * @see R.transduce
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig (a, a -> Boolean) -> [a] -> [a]
+ * @param {Function} pred A predicate used to test whether two items are equal.
+ * @param {Array} list The array to consider.
+ * @return {Array} `list` without repeating elements.
+ * @example
+ *
+ *      function lengthEq(x, y) { return Math.abs(x) === Math.abs(y); };
+ *      var l = [1, -1, 1, 3, 4, -4, -4, -5, 5, 3, 3];
+ *      R.dropRepeatsWith(lengthEq, l); //=> [1, 3, 4, -5, 3]
+ */
+module.exports = _curry2(_dispatchable('dropRepeatsWith', _xdropRepeatsWith, function dropRepeatsWith(pred, list) {
+  var result = [];
+  var idx = 0;
+  var len = list.length;
+  if (len !== 0) {
+    result[0] = list[0];
+    while (++idx < len) {
+      if (!pred(last(result), list[idx])) {
+        result[result.length] = list[idx];
+      }
+    }
+  }
+  return result;
+}));
+
--- a/src/internal/_xdropRepeatsWith.js
+++ b/src/internal/_xdropRepeatsWith.js
@@ -1,0 +1,29 @@
+var _curry2 = require('./_curry2');
+
+
+module.exports = (function() {
+  function XDropRepeatsWith(pred, xf) {
+    this.xf = xf;
+    this.pred = pred;
+    this.lastValue = undefined;
+    this.seenFirstValue = false;
+  }
+  XDropRepeatsWith.prototype['@@transducer/init'] = function() {
+    return this.xf['@@transducer/init']();
+  };
+  XDropRepeatsWith.prototype['@@transducer/result'] = function(result) {
+    return this.xf['@@transducer/result'](result);
+  };
+  XDropRepeatsWith.prototype['@@transducer/step'] = function(result, input) {
+    var sameAsLast = false;
+    if (!this.seenFirstValue) {
+      this.seenFirstValue = true;
+    } else if (this.pred(this.lastValue, input)) {
+      sameAsLast = true;
+    }
+    this.lastValue = input;
+    return sameAsLast ? result : this.xf['@@transducer/step'](result, input);
+  };
+
+  return _curry2(function _xdropRepeatsWith(pred, xf) { return new XDropRepeatsWith(pred, xf); });
+})();
--- a/test/dropRepeatsWith.js
+++ b/test/dropRepeatsWith.js
@@ -1,0 +1,41 @@
+var assert = require('assert');
+
+var R = require('..');
+
+
+describe('dropRepeatsWith', function() {
+  var objs = [
+    {i: 1}, {i: 2}, {i: 3}, {i: 4}, {i: 5}, {i: 3}
+  ];
+  var objs2 = [
+    {i: 1}, {i: 1}, {i: 1}, {i: 2}, {i: 3},
+    {i: 3}, {i: 4}, {i: 4}, {i: 5}, {i: 3},
+  ];
+  var eqI = R.eqProps('i');
+
+  it('removes repeated elements based on predicate', function() {
+    assert.deepEqual(R.dropRepeatsWith(eqI, objs2), objs);
+    assert.deepEqual(R.dropRepeatsWith(eqI, objs), objs);
+  });
+
+  it('keeps elements from the left', function() {
+    assert.deepEqual(
+      R.dropRepeatsWith(eqI, [{i: 1, n: 1}, {i: 1, n: 2}, {i: 1, n: 3}, {i: 4, n: 1}, {i: 4, n: 2}]),
+      [{i: 1, n: 1}, {i: 4, n: 1}]
+    );
+  });
+
+  it('returns an empty array for an empty array', function() {
+    assert.deepEqual(R.dropRepeatsWith(eqI, []), []);
+  });
+
+  it('is curried', function() {
+    assert.strictEqual(typeof R.dropRepeatsWith(eqI), 'function');
+    assert.deepEqual(R.dropRepeatsWith(eqI)(objs), objs);
+    assert.deepEqual(R.dropRepeatsWith(eqI)(objs2), objs);
+  });
+
+  it('can act as a transducer', function() {
+    assert.deepEqual(R.into([], R.dropRepeatsWith(eqI), objs2), objs);
+  });
+});
--- a/test/dropRepeats.js
+++ b/test/dropRepeats.js
@@ -1,0 +1,22 @@
+var assert = require('assert');
+
+var R = require('..');
+
+
+describe('dropRepeats', function() {
+  var objs = [1, 2, 3, 4, 5, 3, 2];
+  var objs2 = [1, 2, 2, 2, 3, 4, 4, 5, 5, 3, 2, 2];
+
+  it('removes repeated elements', function() {
+    assert.deepEqual(R.dropRepeats(objs2), objs);
+    assert.deepEqual(R.dropRepeats(objs), objs);
+  });
+
+  it('returns an empty array for an empty array', function() {
+    assert.deepEqual(R.dropRepeats([]), []);
+  });
+
+  it('can act as a transducer', function() {
+    assert.deepEqual(R.into([], R.dropRepeats, objs2), objs);
+  });
+});


src/internal/_xdropRepeatsWith.js
```diff
@@ -0,0 +1,22 @@
...
+    this.pred = pred;
+  }
+  XDropRepeatsWith.prototype['@@transducer/init'] = function() {
+    return this.xf.init();
+  };
+  XDropRepeatsWith.prototype['@@transducer/result'] = function(result) {
+    return this.xf['@@transducer/result'](result);
+  };
+  XDropRepeatsWith.prototype['@@transducer/step'] = function(result, input) {
+    var sameAsLast = result.length !== 0 && this.pred(last(result), input);
```
kevinbeaty: You should not make any assumptions of the accumulator type when implementing transducers.  The `result` here could be anything from an array, to a stream, channel, observable, etc.  The only thing you should do with `result` is optionally pass it through `@@transducer/step` and return.

src/internal/_xdropRepeatsWith.js
```diff
@@ -0,0 +1,22 @@
...
+var last = require('../last');
+
+
+module.exports = (function() {
+  function XDropRepeatsWith(pred, xf) {
+    this.xf = xf;
+    this.pred = pred;
+  }
+  XDropRepeatsWith.prototype['@@transducer/init'] = function() {
+    return this.xf.init();
```
kevinbeaty: This should be `return this.xf['@@transducer/init']()`

buzzdecafe: note to self: since the `ini` functions are almost always the same, we should should just make them all point to the same function

kevinbeaty: As are the `result` functions ... certainly opportunity to reduce redundancy.  I added [`transducer`](https://github.com/transduce/transduce#transducerstep-result-init) to alleviate it in my `transduce` project, but it's not very ramda-y with the optional args...

paldepind:
Thanks @kevinbeaty. How do you like it now?

kevinbeaty:
@paldepind Looks good. Thanks

CrossEye:
I'll look through the rest of it soon, but in answer to the following, see #890:

> The non-transducer implementation is meant to be performant and I've used Array#push. There is however no other Ramda function that uses push. Is there a reason for that? It is a fast way to build an array and it is supported by all browsers.

There's no really compelling reason to choose one form over the other, but there is a very compelling reason to remain consistent throughout the library.

src/dropRepeatsWith.js
```diff
@@ -0,0 +1,42 @@
...
+ */
+module.exports = _curry2(_dispatchable('dropRepeatsWith', _xdropRepeatsWith, function dropRepeatsWith(pred, list) {
+  var result = [];
+  var idx = 0;
+  if (list.length === 0) {
+    return result;
+  }
+  result.push(list[0]);
+  while (++idx < list.length) {
+    if (!pred(last(result), list[idx])) {
```
CrossEye: Would it be cleaner to cache the previous call rather than call `last` on each iteration?

test/dropRepeatsWith.js
```diff
@@ -0,0 +1,41 @@
...
+
+describe('dropRepeatsWith', function() {
+  var objs = [
+    {i: 1}, {i: 2}, {i: 3}, {i: 4}, {i: 5}, {i: 3}
+  ];
+  var objs2 = [
+    {i: 1}, {i: 1}, {i: 1}, {i: 2}, {i: 3},
+    {i: 3}, {i: 4}, {i: 4}, {i: 5}, {i: 3},
+  ];
+  function eqI(x, accX) { return x.i === accX.i; }
```
CrossEye: This is fine.  But you could also just use `var eqI = R.eqProps('i');`

src/dropRepeatsWith.js
paldepind: It would be more code. But one could argue that it would be more performant (in theory).

paldepind:
> There's no really compelling reason to choose one form over the other, but there is a very compelling reason to remain consistent throughout the library.

I think `push` is cleaner and more obvious. But it doesn't really matter. I've changed it know. I hadn't figured out that length assignment was the pattern you where using.

src/dropRepeatsWith.js
```diff
@@ -0,0 +1,42 @@
...
+ *      var l = [1, -1, 1, 3, 4, -4, -4, -5, 5, 3, 3];
+ *      R.dropRepeatsWith(lengthEq, l); //=> [1, 3, 4, -5, 3]
+ */
+module.exports = _curry2(_dispatchable('dropRepeatsWith', _xdropRepeatsWith, function dropRepeatsWith(pred, list) {
+  var result = [];
+  var idx = 0;
+  if (list.length === 0) {
+    return result;
+  }
+  result.push(list[0]);
```
CrossEye: We've recently (#890) consolidated on using `result[result.length] = list[0]` rather than using `push`.  This is mostly just for consistency.

src/dropRepeatsWith.js
```diff
@@ -0,0 +1,42 @@
...
+  while (++idx < list.length) {
+    if (!pred(last(result), list[idx])) {
+      result.push(list[idx]);
```
CrossEye: Ditto here on the `push`.

test/dropRepeatsWith.js
paldepind: Yes. That is true. I will change it.

I wrote the tests based on the `uniqWith` tests. [They do the same](https://github.com/ramda/ramda/blob/master/test/uniqWith.js#L15).

CrossEye:
I have a few minor caveats noted inline.  But overall, this looks very nice!

test/dropRepeatsWith.js
CrossEye: I don't recall the history.  But it's quite possible that `uniqWith` predates `eqProps`.  It would be good to change that one too, I suppose.

src/dropRepeatsWith.js
```diff
@@ -0,0 +1,42 @@
...
+ *      R.dropRepeatsWith(lengthEq, l); //=> [1, 3, 4, -5, 3]
+ */
+module.exports = _curry2(_dispatchable('dropRepeatsWith', _xdropRepeatsWith, function dropRepeatsWith(pred, list) {
+  var result = [];
+  var idx = 0;
+  if (list.length === 0) {
+    return result;
+  }
+  result[0] = list[0];
+  while (++idx < list.length) {
```
davidchambers: Slight optimization:

```diff
-  var result = [];
-  var idx = 0;
-  if (list.length === 0) {
-    return result;
-  }
-  result[0] = list[0];
-  while (++idx < list.length) {
+  var len = list.length;
+  if (len === 0) {
+    return [];
+  }
+  var result = [list[0]];
+  var idx = 0;
+  while (++idx < len) {
```

paldepind:
I don't understand why you practice this kind of micro optimizations that reduce readability? For loops instead of while loops, length assignment instead of `push` and caching length instead of direct access.

[Microbenchmarks are silly](http://mrale.ph/blog/2014/12/24/array-length-caching.html) and modern JIT [cache the length anyway](http://mrale.ph/blog/2014/12/24/array-length-caching.html) so that our code can get rid of that nastyness.

When I said the function was meant to be performant I simply meant not doing anything out of the ordinary. Like [slicing inside a loop](https://github.com/ramda/ramda/blob/master/src/curryN.js#L70).

src/dropRepeatsWith.js
paldepind: I've cached the length and rearranged the loop to get rid of the second return statement.

davidchambers: > I've […] rearranged the loop to get rid of the second return statement.

Nice!

davidchambers:
:deciduous_tree:

---

Some of the suggestions in this thread were motivated by a desire for consistency. `xs[xs.length] = x` is no better than `xs.push(x)`, but it's the style currently used throughout the library. Caching `xs.length` may be unnecessary but we do so elsewhere. If we're to change these things we should do so throughout the codebase in a separate pull request.

paldepind:
Yes, I understand that. I agree that consistency is important and I have no problem with adhering to the style set by the existing code base. It was the chosen style that I, slightly, objected to :)

davidchambers:
> It was the chosen style that I, slightly, objected to :)

For the record, I also dislike `xs[xs.length] = x`. :)

CrossEye:
I'm not quite sure how that got chosen.  I prefer `push` as well.

@buzzdecafe, can we blame this on you?

`while` loops over `for` you can definitely blame on me, though.  I started the code-base, and one of the (minor and silly, but very real) goals was to never have to write any more `for` loops.  Although people are very used to them, C-style `for`-loops are to my mind very unreadable.  The [first iteration][ew] of what eventually became Ramda tried to code its internals in a very functional manner.  When that didn't scale well, we moved to more imperative internals, but we were not interested in `for`-loops anywhere.


paldepind:
How are for loops unreadable? Thee have clearly defined places for everything and are more compact than while loops?

megawac:
All right, lets clear a couple of these up/move them to more relevant threads

> I'm not quite sure how that got chosen. I prefer push as well.

> @buzzdecafe, can we blame this on you?

I blame Chambers! Lets move these talks to the relevant issue https://github.com/ramda/ramda/issues/890

> I don't understand why you practice this kind of micro optimizations that reduce readability? For loops instead of while loops, length assignment instead of push and caching length instead of direct access.

Caching length can be important in certain cases (when a users function is called) in case they modify the length attribute of the collection

> How are for loops unreadable? Thee have clearly defined places for everything and are more compact than while loops?

Nitty style preferences. I'd suggest ramda stick to what it does

megawac:
> What do you think?

Very :-1:

1) I don't think `drop` is the right turn its a `filter`

2) Its pretty easy to implement via `filterIndexed` or whatever its called now

```js
let filterRepeats = R.filterIndexed(function(item, index, xs) {
    return index < xs.length - 1 || item !== xs[index + 1];
});

```

paldepind:
> I don't think drop is the right term its a filter

No, this is not a filter. Both in Haskell and in Ramda a filter is a function that drops elements from a list based on a unary predicate. A filter looks at a single element in isolation and decides on whether or not to drop it. `dropRepeats` considers the preceding element when deciding what to do. Naming it something with filter, as you suggests, would be inconsistent and confusing. `dropRepeats` is a short descriptive name that intentionally doesn't imply it being a filter since it's not.

I didn't invent the name. Elm uses the same name for a identical function. If you just don't like the current name, then `dedupe` is an alternative used by transducers.js. I don't care much (as long as the name does not include "filter") but I choose `dropRepeats` because I think it is move descriptive than `dedupe`.

> Its pretty easy to implement via filterIndexed or whatever its called now which I believe is also transducered

`filterIndexed` is not transduced. I think it's sloppy of you to not take the 30 seconds it would have taken to validate your claim. Also a transduced `filterIndexed` makes no sense since transducers can apply to things without indexes (channels, functional reactive streams, tree data structures, etc).

Sure, `dropRepeats` could be defined in terms of `filterIndexed`. But I consider `filterIndexed` filter indexed to be an ugly function with rare use as anything but a stepping stone for JavaScript programmers getting used to programming in a functional style. A better an more natural implementation would be based on `reduce`. But as mentioned performance and transducer compatibility were priorities in this implementation.

> A better alternative would be to add a sorted unique function (ala _.uniq) which I suppose would be pretty much the same thing just better termonology

This is certainly not a sorted unique function. The result of this function is neither sorted nor are elements guaranteed to be unique.

megawac:
> I didn't invent the name. Elm uses the same name for a identical function. If you just don't like the current name, then dedupe is an alternative used by transducers.js. I don't care much (as long as the name does not include "filter") but I choose dropRepeats because I think it is move descriptive than dedupe.

Yeah but it was dropped https://github.com/elm-lang/core/issues/200 ;p

> > A better alternative would be to add a sorted unique function (ala _.uniq) which I suppose would be pretty much the same thing just better termonology

> This is certainly not a sorted unique function. The result of this function is neither sorted nor are elements guaranteed to be unique.

I meant that through implicit context

```js
_.uniq([1,1,1,1,1,2,2,1,1,1, 1,1], true)
// => [1, 2, 1]
```

---------

Apologies about the error about `filterIndexed` I wrote that up on a bus ride

I still don't think this belongs in ramda (see the elm issue linked)

paldepind:
> I still don't think this belongs in ramda (see the elm issue linked)

I don't see any arguments in the Elm issue that applies to this function and implementation.

CrossEye:
> I meant that through implicit context, as far as the implementation is concerned it's sorted
> ```javascript
> _.uniq([1,1,1,1,1,2,2,1,1,1, 1,1], true)
> // => [1, 2, 1]
> ```

Ramda's `uniq` does not try to offer the sort of algorithm hint that Underscore does here.  Our `uniq` simply accepts a list and returns a list containing the uniq elements.  Whether the list is sorted or not makes no difference.  And to my mind using that `sorted` hint in this manner would still smack of abuse.

megawac:
> Ramda's uniq does not try to offer the sort of algorithm hint that
Underscore does here. Our uniq simply accepts a list and returns a list
containing the uniq elements. Whether the list is sorted or not makes no
difference. And to my mind using that sorted hint in this manner would
still smack of abuse.

Definitely abusive of the implementation :)

paldepind:
Underscore and Lodash offers two algorithms in `uniq` in a non-transparent way. One could say that this commit recognizes how the `dropRepeats` algorithm is useful for things besides guaranteeing uniqueness in a sorted list.

I.e. this just adds to the usefulness of the function: It gives Ramda users a O(n) algorithm to create a unique list out of a sorted list. It is a more powerful and simpler way to provide what Underscore and Lodash throws together into one function.

davidchambers:
I'm very much in favour of merging this pull request.

CrossEye:
I am as well.  I don't really understand the objection from @megawac.

src/dropRepeatsWith.js
```diff
@@ -0,0 +1,42 @@
...
+ * determined by applying the supplied predicate two consecutive elements.
+ * The first element in a series of equal element is the one being preserved.
+ *
+ * Acts as a transducer if a transformer is given in list position.
+ * @see R.transduce
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig [a] -> [a]
```
davidchambers: I think this `@sig` should be:

    (a,a -> Boolean) -> [a] -> [a]

src/dropRepeats.js
```diff
@@ -0,0 +1,23 @@
+var _curry1 = require('./internal/_curry1');
+var _dispatchable = require('./internal/_dispatchable');
+var _eq = require('./internal/_eq');
+var _xdropRepeatsWith = require('./internal/_xdropRepeatsWith');
+var dropRepeatsWith = require('./dropRepeatsWith');
+
+/**
```
davidchambers: Nit: Two empty lines after imports, please. :)

davidchambers:
@buzzdecafe, @CrossEye: Shall we merge this? LGTM.

buzzdecafe:
i have no strong feelings in either direction.

PR #1009, opened 2015-04-06
Don't clone in sort

paldepind:
Sort currently makes a clone (deep copy) of the list to sort. That is unnecessary and undocumented.

This PR changes `clone` to a `_slice` which is consistent with the rest of the code ([sortBy for instance](https://github.com/ramda/ramda/blob/master/src/sortBy.js#L37)).


davidchambers:
> Sort currently makes a clone (deep copy) of the list to sort. That is unnecessary and undocumented.

Good catch! I accidentally introduced this regression in #673.

:deciduous_tree:

PR #1010, opened 2015-04-06
package.json: attempt to fix erroneous `npm test` failures

davidchambers:
See, for example, <https://travis-ci.org/ramda/ramda/jobs/57303350>.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
   "main": "dist/ramda.js",
   "scripts": {
     "jshint": "grunt jshint",
-    "test": "grunt test"
+    "test": "grunt jshint && grunt jscs && grunt browserify:client && grunt mochaTest:unit"
   },
   "dependencies": {},
   "devDependencies": {


CrossEye:
I don't see how this would help anything.  The Grunt script [already does this][gr] internally.


  [gr]: https://github.com/ramda/ramda/blob/master/Gruntfile.js#L213

davidchambers:
> I don't see how this would help anything.

Running four commands in series is different from running one command which does four things. I don't know *why* this change fixes the problem for me, but the problem was causing lots of pull requests to fail. I'm happy to rely on empirical evidence in this case.

CrossEye:
I have no problem using it if it does fix things.  I just didn't know if it was a magical incantation or if you'd run down the source of the problem.

davidchambers:
Definitely a magical incantation. :frog:

CrossEye:
The really odd thing is that as far as I can tell, this should be making us run jshint, jscs, and browserify each twice, but that doesn't seem to be happening...  I wish I understood magic.

TheLudd:
I have noticed one thing and I don't know if it is related but grunt-cli is installed globally. I think it would be better if it was added as a devDependency. Then anyone could use grunt in the npm scripts after doing npm install. I have previously seen conflicts between global and local modules. But agan, not sure that is the case here.

buzzdecafe:
in that case the solution is to drop grunt

Issue #848, opened 2015-02-28
R.propOr not as useful as lodash's _.result; should it throw if object is undefined?


Andrew-Webb:
This is nice in lodash:-

        var people = [
            {
                name: "Dave",
                age: 30
            },
            {
                name: "Anne",
                age: 28
            },
        ];

        _.result(_.find(people, function (person) {
            return person.age === 40;
        }), "name", undefined);

Depending on the result of the `person.age === N` test, either the name of the person or undefined will be returned.

The Ramda equivalent using propOr (is there a better way?) isn't useful:-

        R.propOr(undefined, "name", R.find(R.propEq('age', 40))(people));

This throws with "Uncaught TypeError: Cannot convert undefined or null to object".

What makes the lodash function work is the fact that _.result can cope with the result of the find being undefined (or null).  R.propOr can't cope with this.  I think that it (and similar functions) should be able to cope with undefineds and nulls - especially since the result of many list functions (returning undefined) might well be piped into them.

(Yes, I've got the first two args to R.propOr the right way round.  The documentation has it wrong; I've raised a separate issue about that.)

CrossEye:
This is a very interesting question.  Ramda generally tries to do little hand-holding.  But throwing an exception here doesn't seem particularly useful.

I think I would need more convincing, though, that we could comfortably know the right result of

    propOr(value, name, refToNothing)

Should it be `value`?  `undefined`? `null`?  Or should it actually throw an exception?  I don't see a great argument for any of the alternatives.  Can you convince me?

Andrew-Webb:
One of the fundamental tenets of functional programming and in Ramda is the composition of functions in a pipeline.  Therefore the behavior of a function like propOr wrt an undefined object should not be considered in isolation, but rather as part of Ramda's whole notion of function pipelining.  After all, as an API user I'm _not_ going to write this:-

    var result = R.propOr('n/a', 'name', undefined);

Since many Ramda functions can return undefined, it then follows that, for the sake of pipelining, every function that a) takes a data arg (or args) and b) might reasonably be part of a pipeline, needs to have a story in the code and in the documentation for what happens if that arg gets piped in as undefined or null.

The worst thing is to have no story.  Then the API user has to experiment to see what happens (or get bitten unexpectedly later on), and the likelihood of a good thing happening is low.  And if a good thing happens, can it be relied upon since there's no documentation?

[...]

PR #1014, opened 2015-04-07
Returning default value from propOr on nil object

CrossEye:
#848

### DIFF
--- a/src/propOr.js
+++ b/src/propOr.js
@@ -1,5 +1,6 @@
 var _curry3 = require('./internal/_curry3');
 var _has = require('./internal/_has');
+var isNil = require('./isNil');
 
 
 /**
@@ -28,5 +29,5 @@ var _has = require('./internal/_has');
  *      favoriteWithDefault(alice);  //=> 'Ramda'
  */
 module.exports = _curry3(function propOr(val, p, obj) {
-  return _has(p, obj) ? obj[p] : val;
+  return (!isNil(obj) && _has(p, obj)) ? obj[p] : val;
 });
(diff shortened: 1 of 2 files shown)


davidchambers:
This change seems reasonable to me.

src/propOr.js
```diff
@@ -28,5 +29,5 @@ var _has = require('./internal/_has');
  *      favoriteWithDefault(alice);  //=> 'Ramda'
  */
 module.exports = _curry3(function propOr(val, p, obj) {
-  return _has(p, obj) ? obj[p] : val;
+  return (!isNil(obj) && _has(p, obj)) ? obj[p] : val;
```
davidchambers: I'd prefer to avoid the import:

```diff
-  return (!isNil(obj) && _has(p, obj)) ? obj[p] : val;
+  return (obj != null && _has(p, obj)) ? obj[p] : val;
```

I'm sure we'll disagree on this one. ;)

buzzdecafe:
:cow: :art:

src/propOr.js
CrossEye: No, I do agree here.  `isNil` is useful to have on the public interface, but it really gains us nothing in readability, performance, or reuse when we use it internally.  I'll update soon.

buzzdecafe: "soon" is a relative term, of course

CrossEye: All things are relative.

Uncle Albert (Einstein) was always muttering something about that.

Andrew-Webb:
Nice change; thanks guys.  I look forward to using it.

PR #1017, opened 2015-04-10
dispatch in R.and and R.or

### LINKED ISSUE #1015: dispatch in R.and and R.or
Sanctuary's [`or`][1] function is simple, intuitive, and useful. Example usage:

```javascript
S.or(S.Nothing(), S.Nothing());  // => S.Nothing()
S.or(S.Nothing(), S.Just('bar'));  // => S.Just('bar')
S.or(S.Just('foo'), S.Nothing());  // => S.Just('foo')
S.or(S.Just('foo'), S.Just('bar'));  // => S.Just('foo')
```

Now that Ramda's [`or`][2] function operates on values rather than functions we could consider dispatching to an `or` method if present. We could also dispatch to an `and` method in `R.and`.

[1]: https://github.com/plaid/sanctuary/blob/v0.4.0/index.js#L210-L221
[2]: http://ramdajs.com/docs/#or

davidchambers:
Closes #1015

### DIFF
--- a/src/and.js
+++ b/src/and.js
@@ -1,9 +1,13 @@
 var _curry2 = require('./internal/_curry2');
+var _hasMethod = require('./internal/_hasMethod');
+
 
 /**
  * A function that returns the first argument if it's falsy otherwise the second
  * argument. Note that this is NOT short-circuited, meaning that if expressions
  * are passed they are both evaluated.
+ *
+ * Dispatches to the `and` method of the first argument if applicable.
  *
  * @func
  * @memberOf R
@@ -19,5 +23,5 @@ var _curry2 = require('./internal/_curry2');
  *      R.and(null, ''); => null
  */
 module.exports = _curry2(function and(a, b) {
-  return a && b;
+  return _hasMethod('and', a) ? a.and(b) : a && b;
 });
--- a/src/or.js
+++ b/src/or.js
@@ -1,10 +1,13 @@
 var _curry2 = require('./internal/_curry2');
+var _hasMethod = require('./internal/_hasMethod');
 
 
 /**
  * A function that returns the first truthy of two arguments otherwise the
  * last argument. Note that this is NOT short-circuited, meaning that if
  * expressions are passed they are both evaluated.
+ *
+ * Dispatches to the `or` method of the first argument if applicable.
  *
  * @func
  * @memberOf R
@@ -20,5 +23,5 @@ var _curry2 = require('./internal/_curry2');
  *      R.or(null, ''); => ''
  */
 module.exports = _curry2(function or(a, b) {
-  return a || b;
+  return _hasMethod('or', a) ? a.or(b) : a || b;
 });
--- a/test/and.js
+++ b/test/and.js
@@ -11,6 +11,26 @@ describe('and', function() {
     assert.strictEqual(R.and(true, someAr), someAr);
   });
 
+  it('dispatches to `and` method', function() {
+    function Nothing() {}
+    Nothing.prototype.and = function() { return this; };
+
+    function Just(x) { this.value = x; }
+    Just.prototype.and = R.identity;
+
+    var n1 = new Nothing();
+    var n2 = new Nothing();
+    var j1 = new Just(1);
+    var j2 = new Just(2);
+
+    assert.strictEqual(R.and(n1, n2), n1);
+    assert.strictEqual(R.and(n2, n1), n2);
+    assert.strictEqual(R.and(n1, j2), n1);
+    assert.strictEqual(R.and(j2, n1), n1);
+    assert.strictEqual(R.and(j1, j2), j2);
+    assert.strictEqual(R.and(j2, j1), j1);
+  });
+
   it('is curried', function() {
     var halfTruth = R.and(true);
     assert.strictEqual(halfTruth(false), false);
--- a/test/or.js
+++ b/test/or.js
@@ -12,6 +12,26 @@ describe('or', function() {
     assert.strictEqual(R.or('', 0), 0);
   });
 
+  it('dispatches to `or` method', function() {
+    function Nothing() {}
+    Nothing.prototype.or = R.identity;
+
+    function Just(x) { this.value = x; }
+    Just.prototype.or = function() { return this; };
+
+    var n1 = new Nothing();
+    var n2 = new Nothing();
+    var j1 = new Just(1);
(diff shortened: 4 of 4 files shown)


src/and.js
```diff
@@ -19,5 +23,5 @@ var _curry2 = require('./internal/_curry2');
  *      R.and(null, ''); => null
  */
 module.exports = _curry2(function and(a, b) {
-  return a && b;
+  return _hasMethod('and', a) ? a.and(b) : a && b;
```
buzzdecafe: interesting. `&&` is commutative, so it may not matter. But typically we dispatch on the last argument, except in cases where it's an operator, e.g. `ap`. Why did you opt to dispatch on `a` instead of `b`?

davidchambers: `&&` is not commutative for types with multiple "true" values:

```javascript
1 && 2;  // => 2
2 && 1;  // => 1
```

It's important that `R.and` behave similarly:

```javascript
Just(1).and(Just(2));  // => Just(2)
Just(2).and(Just(1));  // => Just(1)
```

buzzdecafe: > && is not commutative for types with multiple "true" values:

good :point_right: 

so you're essentially treating `and` as an operator. The rationale was not immediately clear to me, but that makes perfect sense

davidchambers: Cool. :)

davidchambers:
@CrossEye?

CrossEye:
Sorry, there were a few issues I thought I responded to some days ago that I never seem to have sent.

This one I'm absolutely in favor of.  Easy :+1:

But I do want to note that I notice an oddity in the implementation of `Just`/`Nothing` you use for examples.  Is this really what you would want?

    R.and(Just(false), Just(true)); //=> Just(true)

Something seems very off here.

davidchambers:
> Is this really what you would want?
>
>     R.and(Just(false), Just(true)); //=> Just(true)

This is the behaviour I expect, yes.

If you want to `&&` the `a`s in two `Maybe a`s, there's [`R.lift`][1]:

```javascript
R.lift(R.and)(Just(false), Just(true)); //=> Just(false)
```


[1]: http://ramdajs.com/docs/#lift

CrossEye:
Yes, of course that makes sense.

I still have one other concern: this should extend also to `not` for completeness, should it not?

davidchambers:
For `not` to apply to a type the type must have exactly one "true" value and exactly one "false" value. I can't think of a type for which this is the case except for Boolean. `not(Just(42))` could be `Nothing()`, for example, but what would `not(Nothing())` be?

CrossEye:
The fact that we can't do that makes me suspicious of the whole endeavor, though.  While your current implementation of `Just` and `Nothing` with their `and` and `or`, alongside Ramda's `not`, follow [De Morgan's Laws][dm], they only do so because `R.not(Just(whatever))` and `R.not(Nothing)` both yield `false`.

It still strikes me that somehow, if you're going to define `and` and `or` without defining `not`, you're not capturing something essential.  But I'm still struggling to figure out exactly what it is.

  [dm]: http://en.wikipedia.org/wiki/De_Morgan%27s_laws

davidchambers:
I suppose since `R.not('foo')` evaluates false and `R.not('')` evaluates true, `R.not(Just(42))` could evaluate false and `R.not(Nothing())` could evaluate true.

CrossEye:
Don't let me hold anything up here.   I'm very confused by something, but I can't really tell if it's something fundamental or an oddity of my own warped mind.

davidchambers:
To me this thread indicates that `R.not` doesn't make sense for types other than Boolean.

If for some reason one wished to get a Boolean from another type, it would make more sense to define `toBoolean` rather than `not`:

```javascript
Nothing.prototype.toBoolean = R.F;
Just.prototype.toBoolean = R.T;
```

This isn't at all useful, though, so let's not define `R.toBoolean`. ;)

scott-christopher:
I think the potential confusion here is that dispatching `and` and `or` can work if the type is a monoid, where the functions return the same type as the two parameters, while `not` will only ever results in a Boolean value.

> To me this thread indicates that R.not doesn't make sense for types other than Boolean.
Pretty much this, though it can still make sense for types that can be coerced to a boolean value, if there is a desire to support it.

CrossEye:
> I think the potential confusion here is that dispatching and and or can work if the type is a monoid, 

I don't think it requires a full monoid, just the existence of an `empty` value... but that's probably not it either.  With numbers, we have +0 and -0, so there are two empty values.  What I've been trying to figure out is reasonable laws for a type that overrides `or` and `and`.  It does seem to have to do with having a collection of designated empty values for the type (e.g. `Nothing`) and rules something like

    or(A, B) returns A, unless A is an empty value, in which case it returns B
    and(A, B) returns B, if A is an empty value, or A otherwise

But I've always thought of `or`, `and`, and `not` as a trinity, and clearly there are no such laws related for `not`, some I'm still confused.

CrossEye:
So without dealing with the notion of empty values, perhaps the laws could be phrased like this:

    for values, A, B of a given type
    if and(A, B) returns A, then or(A, B) returns B
    if or(A, B) returns B, then and(A, B) returns A

Does this make sense?

buzzdecafe:
extending this a bit:

```
for values, A, B of a given type
if and(A, B) returns B, then or(A, B) returns A
if or(A, B) returns A, then and(A, B) returns B
```

davidchambers:
I'll merge this as is. We can update `R.not` in a separate pull request if we decide it makes sense to do so.

CrossEye:
This is fine.  What I think we want to do is somehow delineate the responsibilities of types overriding `and` and `or`.  Something like
 
    eq(and(A, B), A) <=> eq(or(A, B), B)
    eq(and(A, B), B) <=> eq(or(A, B), A)

Would we also want to insist that either `eq(and(A, B), A)` or `eq(and(A, B), B)`, and similarly for `or`, or can we think of any reasons that an implementation might reasonably want to do something different?

CrossEye:
Obviously my ideas on `not` were misguided.  I was expecting a lawfulness contract for these would have to include it, but clearly that's not the case.

davidchambers:
In plaid/sanctuary#26 we dispatch to `toBoolean` rather than `and` or `or`. We could use this approach in Ramda too, though it would be the first example of dispatching to a method of a different name. If there's interest, I'll put together a pull request.

Issue #1002, opened 2015-04-03
drop array items starting from the tail


bobiblazeski:
How about adding negative index to drop that returns the array items dropping the ones from the tail.
R.drop(-2,[1,2,3,4,5]) //=>  [1,2,3]
Current implementation of drop accepts negative arguments but returns weird result:
R.drop(-2,[1,2,3,4,5]) //=>[,,1,2,3,4,5]

CrossEye:
Well we should definitely fix `drop` so that it doesn't have the current odd behavior when passed a negative number. 

But I'd have to think a little more about the main point.   It feels odd.   In `drop (n)`, the `n` does not to my mind represent a _position_, for which a negative number had a fairly obvious meaning by extension.   Instead, it represents a _count_.  There's simply no clear meaning to switching it to negative.   The current semantics of `drop` easily extend to streams, for instance.  But a negative value would be inchoate.

So my initial reaction is negative.   But people have been able to convince me before.   :smile:

bobiblazeski:
What I people I need tips :) ?
Joke aside what do you recommend instead for dropping  elements from the tail.

buzzdecafe:
if you know the length of the list you can just `slice` it down to the desired size.

bobiblazeski:
I'm doing it but its ugly, might as well call array slice directly. On the other hand Ramda drop, head & last add value by dealing  with low level stuff so I don't have to.

buzzdecafe:
can you post the code in question?

bobiblazeski:
I have a list of elements [1,2,3,4,5] and a remove button that pushes that last element into removing items where removal animation is played. In this case 5 would be put into removing array while [1,2,3,4] will stay at entries array. If I removed items from the top head and tail would fit perfectly.

https://github.com/bobiblazeski/ramda-animate/blob/master/client/views/entries.jsx#L15
http://ramda-animate.meteor.com/

CrossEye:
The difficulty is that Ramda's default type is logically the _list_.  There is no real list type in JS, so we settle for arrays.  And occasionally we'll compromise and include some functions that are specific to this implementation (`last`, `init`, and the implementation details of `slice`.)  It's harder to do this with `drop`, which already has a clear meaning for streams.

davidchambers:
`R.drop` should not accept a negative count. [`R.slice(0, -2, [1, 2, 3, 4, 5])`][1] works as expected.


[...]

PR #1018, opened 2015-04-10
treat negative number as zero in R.drop

davidchambers:
See #1002

Before:

```javascript
R.drop(-2, [1, 2, 3, 4, 5]);  // => [, , 1, 2, 3, 4, 5]
```

After:

```javascript
R.drop(-2, [1, 2, 3, 4, 5]);  // => [1, 2, 3, 4, 5]
```

### DIFF
--- a/src/drop.js
+++ b/src/drop.js
@@ -5,7 +5,7 @@ var _xdrop = require('./internal/_xdrop');
 
 
 /**
- * Returns a new list containing all but the first `n` elements of the given `list`.
+ * Returns a list containing all but the first `n` elements of the given `list`.
  *
  * Acts as a transducer if a transformer is given in list position.
  * @see R.transduce
@@ -22,5 +22,5 @@ var _xdrop = require('./internal/_xdrop');
  *      R.drop(3, [1,2,3,4,5,6,7]); //=> [4,5,6,7]
  */
 module.exports = _curry2(_dispatchable('drop', _xdrop, function drop(n, list) {
-  return n < list.length ? _slice(list, n) : [];
+  return n <= 0 ? list : _slice(list, n);
 }));
--- a/test/drop.js
+++ b/test/drop.js
@@ -12,6 +12,12 @@ describe('drop', function() {
     assert.deepEqual(R.drop(20, ['a', 'b', 'c', 'd', 'e', 'f', 'g']), []);
   });
 
+  it('returns an equivalent list if `n` is <= 0', function() {
+    assert.deepEqual(R.drop(0, [1, 2, 3]), [1, 2, 3]);
+    assert.deepEqual(R.drop(-1, [1, 2, 3]), [1, 2, 3]);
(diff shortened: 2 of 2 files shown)


megawac:
:+1:

src/drop.js
```diff
@@ -22,5 +22,5 @@ var _xdrop = require('./internal/_xdrop');
  *      R.drop(3, [1,2,3,4,5,6,7]); //=> [4,5,6,7]
  */
 module.exports = _curry2(_dispatchable('drop', _xdrop, function drop(n, list) {
-  return n < list.length ? _slice(list, n) : [];
+  return n <= 0 ? list : _slice(list, n);
```
megawac: You should create a new list

davidchambers: We've discussed this in other threads recently, and the decision was to avoid defensive cloning.

megawac: I saw that discussion but figured the decision would end up on the other side of the fence. I'm against that decision as it can lead to some confusing side effects and breaches the API contract/expectation of not returning a new collection in some cases

davidchambers: I've removed the word "new" from the docstring so it now reads:

> Returns a list containing all but the first `n` elements of the given `list`.

buzzdecafe:
:cow2: :zap:

src/drop.js
megawac: It still seems like surprising behaviour and against ramdas principles

CrossEye: @megawac: If our API is meant to deal mostly with immutable data, why should Ramda take the hit of cloning?  If one user needs to worry about reference equality, then she will need to do defensive cloning as necessary, but other users shouldn't have to pay the perf costs of that.

I guess I'm not particularly worried about 

```javascript
var a = [4,5,6];
var n = random(-5, 5);

R.take(n, a).push(1,2,3)
```

when 

```javascript
R.concat(R.take(n, a), [1, 2, 3])
```

actually _is_ consistent.

megawac: Again, I'm suggesting performance on a minor edge case doesn't justify the
surprise a user may experience from results of this - users may be using
ramda as a step leading up to a non immutable process

CrossEye: Below was the argument I was going to make before I started to reconsider:

> I guess my feeling is that if we're going to clone, then we should go all the way:
> 
> ```javascript
> var list = [{x: 1}, {x: 2}, {x: 3}];
> var a = R.head(list);
> var b = R.head(R.take(1, list));
> R.deepEq(a, b); //=> true
> a == b; //=> false, because we've cloned.
> ```
> 
> Perf is only one minor concern.  The major one has to do with what sort of API we're presenting....

But now I can't get my head around it.  I'm not really sure what sounds correct.  I'll think on it a little more.

megawac:
Ping @CrossEye second thoughts. I'm still strongly against these of the API decisions you guys have been making over the past month or so

davidchambers:
Let's move the discussion to #1025.

PR #1019, opened 2015-04-10
Update _xfind to transducer protocol

paldepind:
Is seems like [this commit](https://github.com/ramda/ramda/commit/fd3c99b8ac0d07a16019e902663a6e1d27f8cfa6) missed this `init` function in `_xfind`.

### DIFF
--- a/src/internal/_xfind.js
+++ b/src/internal/_xfind.js
@@ -9,7 +9,7 @@ module.exports = (function() {
     this.found = false;
   }
   XFind.prototype['@@transducer/init'] = function() {
-    return this.xf.init();
+    return this.xf['@@transducer/init']();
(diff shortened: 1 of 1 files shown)


davidchambers:
Does this mean this function is not invoked by the test suite?

paldepind:
I don't think so. As I noted in #1007 it appears to me as if most of the transducer functions are untested.

If I'm right there should probably be a discussion about the general approach to testing these functions (I used `R.into` in the before mentioned PR).

And note that the commit by @buzzdecafe I linked to above doesn't change any tests either.

buzzdecafe:
whoops. yes, i broke it, and yes, we need coverage

Issue #1020, opened 2015-04-10
Duplicated transducer `init` and `result` functions


paldepind:
This has been mentioned elsewhere, but there doesn't seem to be an open issue about it.

The plain reducer init function is apparently used 13 places in Ramda:

```
grep -r "return this.xf\['@@transducer/init']()"
_xtake.js:    return this.xf['@@transducer/init']();
_xall.js:    return this.xf['@@transducer/init']();
_xmap.js:    return this.xf['@@transducer/init']();
_xfind.js:    return this.xf['@@transducer/init']();
_xfilter.js:    return this.xf['@@transducer/init']();
_xgroupBy.js:    return this.xf['@@transducer/init']();
_xtakeWhile.js:    return this.xf['@@transducer/init']();
_xdrop.js:    return this.xf['@@transducer/init']();
_xany.js:    return this.xf['@@transducer/init']();
_xdropWhile.js:    return this.xf['@@transducer/init']();
_xfindLastIndex.js:    return this.xf['@@transducer/init']();
_xfindLast.js:    return this.xf['@@transducer/init']();
_xfindIndex.js:    return this.xf['@@transducer/init']();
```

Maybe default `init` and `result` functions should be defined in a helper file to avoid the repetition?

buzzdecafe:
absolutely yes.

While we're at it, there are a number of redundant `@@transducer/result` functions as well:

https://github.com/ramda/ramda/search?utf8=%E2%9C%93&q=return+this.xf%5B%27%40%40transducer%2Fresult%27%5D%28result%5C%29%3B&type=Code

paldepind:
But, where should the helper functions be located? The current scheme seems to be one file in `src/internal/` per helper function.

But separating these two functions across two files might be overkill?

buzzdecafe:
i think something like this would be fine:

```javascript
// file: src/internal/_xfBase.js
module.exports = {
   init: function() { return this.xf['@@transducer/init'](); },
   result: function(result) { return this.xf['@@transducer/result'](result); }
};
```

then:

```
var _xfBase = require('./_xfBase');

// .... :
XThing.prototype.init = _xfBase.init
XThing.prototype.result = _xfBase.result
// etc.
```

paldepind:
Yes. I agree. I'll create a pull request if you don't mind.

buzzdecafe:
sounds good, thank you

buzzdecafe:
fixed in #1022

PR #1022, opened 2015-04-10
Shared definition for common transducer init and result functions

paldepind:
As per the discussion in #1020.

I don't think I've missed anything :)

### DIFF
--- a/src/internal/_xfBase.js
+++ b/src/internal/_xfBase.js
@@ -1,0 +1,8 @@
+module.exports = {
+  init: function() {
+    return this.xf['@@transducer/init']();
+  },
+  result: function(result) {
+    return this.xf['@@transducer/result'](result);
+  }
+};
--- a/src/internal/_xall.js
+++ b/src/internal/_xall.js
@@ -1,5 +1,6 @@
 var _curry2 = require('./_curry2');
 var _reduced = require('./_reduced');
+var _xfBase = require('./_xfBase');
 
 
 module.exports = (function() {
@@ -8,9 +9,7 @@ module.exports = (function() {
     this.f = f;
     this.all = true;
   }
-  XAll.prototype['@@transducer/init'] = function() {
-    return this.xf['@@transducer/init']();
-  };
+  XAll.prototype['@@transducer/init'] = _xfBase.init;
   XAll.prototype['@@transducer/result'] = function(result) {
     if (this.all) {
       result = this.xf['@@transducer/step'](result, true);
--- a/src/internal/_xany.js
+++ b/src/internal/_xany.js
@@ -1,5 +1,6 @@
 var _curry2 = require('./_curry2');
 var _reduced = require('./_reduced');
+var _xfBase = require('./_xfBase');
 
 
 module.exports = (function() {
@@ -8,9 +9,7 @@ module.exports = (function() {
     this.f = f;
     this.any = false;
   }
-  XAny.prototype['@@transducer/init'] = function() {
-    return this.xf['@@transducer/init']();
-  };
+  XAny.prototype['@@transducer/init'] = _xfBase.init;
   XAny.prototype['@@transducer/result'] = function(result) {
     if (!this.any) {
       result = this.xf['@@transducer/step'](result, false);
--- a/src/internal/_xdrop.js
+++ b/src/internal/_xdrop.js
@@ -1,4 +1,5 @@
 var _curry2 = require('./_curry2');
+var _xfBase = require('./_xfBase');
 
 
 module.exports = (function() {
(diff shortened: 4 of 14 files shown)


CrossEye:
While I'm probably :+1: on this change, I've got to say that seeing all those `prototypes` and constructor functions in our FP library is not a good feeling!

But if we're going to do it, are we better off going just a little further and adding something like `inheritFrom`?:

```javascript
XAll.prototype = _inheritFrom(_xfBase);
XAll.prototype['@@transducer/result'] = function(result) {...}
```

buzzdecafe:
inheritance seems like overkill to me, since all we're doing is sharing function references.

davidchambers:
> inheritance seems like overkill to me, since all we're doing is sharing function references.

:+1:

buzzdecafe:
we all good then? I say :cow2:

CrossEye:
> inheritance seems like overkill to me, since all we're doing is sharing function references.

What else would you mean by inheritance?   Especially in JS.

But yes, I'm good to go.

src/internal/_xfBase.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+   init: function() {
```
davidchambers: :triangular_ruler: You have three spaces here rather than two.

davidchambers:
> What else would you mean by inheritance?

An approach which enabled `dog instanceof Animal` to evaluate true.

CrossEye:
I almost never use `instanceof` in JS, preferring to use duck-typing.  But yes, you do want to know something about object types.   I would argue, though that you only need that information as a proxy for the types of messages you can send to that object: in JS those are the function properties it has along its prototype chain.

kevinbeaty:
What about something like this (not sure on the naming):

```javascript
function _xfDefaults(xfImpl){
   xfImpl.prototype['@@transducer/init'] = _xfInit
   xfImpl.prototype['@@transducer/result'] = _xfStep
   // ? return xfImpl
}

// later
function XTake(n){}
_xfDefaults(XTake)
XTake.prototype['@@transducer/step'] = ...

// or even
function _xfWithStep(xfImpl, xfStep){
   xfImpl.prototype['@@transducer/init'] = _xfInit
   xfImpl.prototype['@@transducer/step'] = _xfStep
   xfImpl.prototype['@@transducer/result'] = _xfStep
   // ? return xfImpl
}


_xfWithStep(XTake, function(acc, value){})
```

Just a  thought.  May not be worth it.  I'm not opposed to the current PR either.

CrossEye:
@kevinbeaty: Your first suggestion is similar to what I was thinking about.  I'm perfectly happy for the moment with this PR.  Whatever unhappiness I have is with the usage of so many constructor functions and prototypes to implement this part of our functional library.  Something really seems off.

paldepind:
@kevinbeaty: That's not a bad idea. But it may not be worth since were only adding 1 or 2 functions (as far as I know there are no step function shared across the transducers). If this is what people prefer I can change this PR.

@carlosrocha: Getting rid of the the constructor functions and prototype usage is certainly do able. `_xfind` for instance could be changed from this:

```javascript
  function XFind(f, xf) {
    this.xf = xf;
    this.f = f;
    this.found = false;
  }
  XFind.prototype['@@transducer/init'] = _xfBase.init;
  XFind.prototype['@@transducer/result'] = function(result) {
    if (!this.found) {
      result = this.xf['@@transducer/step'](result, void 0);
    }
    return this.xf['@@transducer/result'](result);
  };
  XFind.prototype['@@transducer/step'] = function(result, input) {
    if (this.f(input)) {
      this.found = true;
      result = _reduced(this.xf['@@transducer/step'](result, input));
    }
    return result;
  };

  return _curry2(function _xfind(f, xf) { return new XFind(f, xf); });
})();
```

To something like this:

```javascript
var resFn = function(result) {
  if (!this.found) {
    result = this.xf['@@transducer/step'](result, void 0);
  }
  return this.xf['@@transducer/result'](result);
};
var stepFn = function(result, input) {
  if (this.f(input)) {
    this.found = true;
    result = _reduced(this.xf['@@transducer/step'](result, input));
  }
  return result;
};
module.exports = curry2(function(f, xf) {
  return {
    xf: xf,
    f: f,
    found: false,
    '@@transducer/init': _xfBase.init,
    '@@transducer/result': resFn,
    '@@transducer/step': stepFn,
  }
});
```

Do you like the second version better?

buzzdecafe:
unless there is some demonstrable benefit from this rewrite, i don't think it's worth it. the interface is the point, not the implementation.

CrossEye:
> Do you like the second version better?

Assuming that Carlos was some weird typo for me (:smile:), then yes I do, but not enough to make it worth going back and changing the existing implementations unless we're in there doing other things.  I agree with @buzzdecafe here.

paldepind:
> Assuming that Carlos was some weird typo for me (:smile:)

Yacks! I apparently can't handle the GitHub autocompletion :(

I'm indifferent about this – and happy as long as the duplication goes away.

CrossEye:
So is everyone on-board with this?

PR #1023, opened 2015-04-11
Fix pathEq test

scott-christopher:
Noticed this while playing around with the tests. Interesting that these tests only passed because `['value'].toString() === 'value'`, which means `({value: NaN})[['value']] === NaN`


CrossEye:
:herb:

davidchambers:
:deciduous_tree:

Issue #183, opened 2014-07-20
`uniq` could be defined in terms of `uniqWith`


wavebeem:
Would it be reasonable to simply implement `uniq` like follows?

``` javascript
var uniq = R.uniq = uniqWith(eq);
```

I dunno if this would have undesirable performance implications, but it's definitely a lot DRY-er.

PR #1026, opened 2015-04-14
Implement uniq in terms of uniqWith

paldepind:
Removes repetition.

With equivalent benefits this could easily be done for other functions with `*With` equivalents as well.

### DIFF
--- a/src/uniq.js
+++ b/src/uniq.js
@@ -1,5 +1,5 @@
-var _contains = require('./internal/_contains');
-var _curry1 = require('./internal/_curry1');
+var eq = require('./eq');
+var uniqWith = require('./uniqWith');
 
 
 /**
@@ -19,14 +19,4 @@ var _curry1 = require('./internal/_curry1');
  *      R.uniq([{}, {}]);     //=> [{}, {}]
  *      R.uniq([1, '1']);     //=> [1, '1']
  */
-module.exports = _curry1(function uniq(list) {
-  var idx = -1, len = list.length;
-  var result = [], item;
-  while (++idx < len) {
-    item = list[idx];
-    if (!_contains(item, result)) {
-      result[result.length] = item;
-    }
-  }
-  return result;
-});
+module.exports = uniqWith(eq);


buzzdecafe:
i think this was a performance optimization way back when. did you bench this to see what effect this change may have?

paldepind:
All of the functions with `*With` variants are implemented in this way. Why
would that give  better performance?

buzzdecafe:
https://github.com/ramda/ramda/issues/183

paldepind:
I truly don't understand why the performance difference is so significant. But I can't argue with the benchmarks.

[Here is an updated benchmark](http://jsperf.com/ramda-uniqwitheq/3). There's and added test that shows that `R.eq` is part of the reason. Also note that in Firefox `R.uniqWith(function(a, b) { return a === b; });` is twice as fast as `R.uniq`!

davidchambers:
:+1: from me. :)

buzzdecafe:
let's do it, if it needs optimizing later we can circle back
:cow2:

paldepind:
I agree. `uniqWith` is exposed to users. If its not good enough for us then I think the issue is in `uniqWith`  and a stab should be taken at optimizing it instead of not using internally.

buzzdecafe:
POITROAE. Should it ever arise that we need to improve perf here, we have some possible optimizations in our back pocket.

CrossEye:
I guess I'm on board.  It would certainly be nice to have simpler code.  And we know that if we want to speed this up, it's not `uniqWith` that's the issue.  It would be easy enough to do this with a lambda.

buzzdecafe:
problem is our `eq` is different than simple `===`, since it handles `-0` and `NaN`

CrossEye:
Yes, but if we needed to change from we could inline that and save the function call and probably not take too bad a performance hit:

    var uniqWith = uniq((a, b) =>   (a === 0 ? (1 / a === 1 / b) : (a === b || (a != a && b != b)));

or some such, although this is clearly not as DRY as we would like.

buzzdecafe:
agreed, i was just noting that it should not be a simple `a === b` lambda

PR #1029, opened 2015-04-16
docs/filter: only filter categories with a full match

raine:
i feel this filtering logic serves the majority of use cases better

before:

screenshot 2015-04-16 21 07 55

after:

screenshot 2015-04-16 21 07 31


buzzdecafe:
i agree

:cow2:

davidchambers:
:deciduous_tree:

CrossEye:
Yup.  There was talk about adding the ability to search by something like 'cat:Function', and maybe we'll get there some day, but this is definitely an improvement.

Issue #1032, opened 2015-04-18
simplify R.where


davidchambers:
One of the things we often cite as a difference between Ramda and certain other utility libraries is its [simple][1] API. [`R.where`][2] is not simple (emphasis mine):

> Takes a spec object and a test object and returns true if the test satisfies the spec. Any property on the spec *that is not a function* is interpreted as an equality relation.
>
> *If the spec has a property mapped to a function, then `where` evaluates the function, passing in the test object's value for the property in question, as well as the whole test object.*

Making the behaviour of `R.where` dependent on the types of the spec's values increases the complexity of this function. The ability to provide a function is very useful, though. I propose that each of the spec's values *must* be a function. This offers several advantages:

  - simpler implementation;
  - simpler type signature;
  - simpler documentation;
  - it exposes users to the strictly more powerful functionality; and
  - it removes an edge case.

This demonstrates the edge case:

```javascript
> Array.prototype.foo = {}
{}
> Array.prototype.bar = function() {}
[Function]
> R.where({foo: Array.prototype.foo}, [])
true
> R.where({bar: Array.prototype.bar}, [])
false
```

Existing uses of `R.where` could then be trivially transformed thanks to [`R.eq`][3]:

```diff
-var spec = {x: 2};
+var spec = {x: R.eq(2)};
 R.where(spec, {w: 10, x: 2, y: 300}); //=> true
 R.where(spec, {x: 1, y: 'moo', z: true}); //=> false

 var spec2 = {x: function(val, obj) { return  val + obj.y > 10; }};
 R.where(spec2, {x: 2, y: 7}); //=> false
 R.where(spec2, {x: 3, y: 8}); //=> true

 var xs = [{x: 2, y: 1}, {x: 10, y: 2}, {x: 8, y: 3}, {x: 10, y: 4}];
-R.filter(R.where({x: 10}), xs); // ==> [{x: 10, y: 2}, {x: 10, y: 4}]
+R.filter(R.where({x: R.eq(10)}), xs); // ==> [{x: 10, y: 2}, {x: 10, y: 4}]
```

Actually, a trivial transformation is *not* possible due to another source of complexity: the function is passed the value *and the object in question* (as can be seen in the `spec2` example above). We'd need to stop passing this second argument as part of this change to make `R.eq` viable. This would lose us the ability to make decisions based on multiple fields, though this is a different problem for which `R.where` is not a good fit. The `spec2` example is better written *without* `R.where`:

```javascript
[...]

PR #1034, opened 2015-04-20
remove special handling of functions from R.where

davidchambers:
In #1032 I proposed going in the other direction: requiring that every value in the spec be a predicate. I'm now of the opinion that since `R.where` is just sugar, it's more useful to support the common case nicely.

It's possible that we should have both versions. If so, perhaps this version should be named `R.whereEq` and the function version `R.where`.


CrossEye:
I'm pretty sure I prefer #1032.

Noting that something is just syntactic sugar doesn't say much.  All our functions can be seen as syntactic sugar on the underlying Javascript language.  LISP shows us that most syntax can be seen as simple syntactic sugar on  a few primitives.  Lambda calculus shows us that we can build anything computable as sugar for the `S`, `K`, and `I` combinators, and even `I` is derivable.  But I don't want to express `where` directly in terms of `S` and `K`.  Ever.  So we always need to choose the right level of sugar to express our intent in code the way we find comfortable.

I would have no major objection to splitting the capabilities of `where`.  You correctly point out that the API is less simple than it reasonably could be.  But I'd like to at least retain the more powerful version:

    var predicate = where({
        a: eq('foo'),
        b: complement(eq('bar')),
        x: gt(_, 10),
        y: lt(_, 20)
    });

If this is separate from the `eq` version, I could live with it, but I do like its capabilities.

davidchambers:
Superseded by #1036

PR #1036, opened 2015-04-20
separate R.where into two simpler functions: R.where and R.whereEq

### LINKED ISSUE #1032: simplify R.where

davidchambers:
Closes #1032

This is the alternative to #1034. If we decide we want *both* functions we need to decide what to name them. :)


test/where.js
```diff
@@ -50,17 +27,13 @@ describe('where', function() {
   });
 
   it('is false if the test object is null-ish', function() {
```
davidchambers: I'd love to remove this special case.

src/where.js
```diff
@@ -1,44 +1,46 @@
...
  */
 module.exports = _curry2(function where(spec, testObj) {
-  var parsedSpec = groupBy(function(key) {
-    return typeof spec[key] === 'function' ? 'fn' : 'obj';
-  }, keys(spec));
-
-  return _satisfiesSpec(spec, parsedSpec, testObj);
+  testObj = Object(testObj);
+  for (var prop in spec) {
+    if (_has(prop, spec) && !spec[prop](testObj[prop])) {
```
CrossEye: I would still rather pass the object as well.  We clearly need to hash this out.  The triangle inequality example deleted below is a reasonable example of why.

test/where.js
CrossEye: :+1:

src/where.js
davidchambers: My view is that multiple-field constraints are better expressed in terms of functions which take the *whole* object. The `(value, object)` parameters strike me as awkward: if one has access to the whole object, `value` isn't necessary. The counterargument is likely that `value` is convenient in the common case. To this I argue that `R.where` should focus on the common case and let multiple-field constraints be expressed in different ways.

With `R.where`:

```javascript
// valid :: { a :: Number, b :: Number, c :: Number } -> Boolean
var valid = R.where({
  a: function(a, obj) {
    return a < obj.b + obj.c;
  },
  b: function(b, obj) {
    return b < obj.a + obj.c;
  },
  c: function(c, obj) {
    return c < obj.a + obj.b;
  }
});
```

Without `R.where`:

```javascript
// valid :: { a :: Number, b :: Number, c :: Number } -> Boolean
var valid = function(obj) {
  return (obj.a < obj.b + obj.c &&
          obj.b < obj.a + obj.c &&
          obj.c < obj.a + obj.b);
};
```

Or even:

```javascript
// valid :: { a :: Number, b :: Number, c :: Number } -> Boolean
var valid = R.pipe(R.props(['a', 'b', 'c']),
                   R.sortBy(R.identity),
                   R.converge(R.lt,
                              R.last,
                              R.pipe(R.init, R.sum)));
```

test/where.js
davidchambers: :zap:

buzzdecafe:
IMO `where` is about specifying an interface that the input _object_ must satisfy, not that the object's individual _properties_ must satisfy _in isolation_.

megawac:
I'm fine with this as long as `whereEq` is included

```js
whereEq = compose(where, mapObject(eq))
```

CrossEye:
Damn, it hadn't hit me how straightforward that conversion was:

>     whereEq = compose(where, mapObject(eq))

Nice.

davidchambers:
Shall I add `R.whereEq`?

CrossEye:
I could be happy with this version of `where` alongside a `whereEq`.  But I believe @buzzdecafe still might have some reservations about it.

buzzdecafe:
not reservations. just not seeing the point particularly.

CrossEye:
I say yes, then.  Two simple functions beat one complex function hands-down.

davidchambers:
I've added `R.whereEq`.

src/whereEq.js
```diff
@@ -0,0 +1,35 @@
...
+var where = require('./where');
+
+
+/**
+ * Takes a spec object and a test object; returns true if the test satisfies
+ * the spec, false otherwise. An object satisfies the spec if, for each of the
+ * spec's own properties, accessing that property of the object gives the same
+ * value (in `R.eq` terms) as accessing that property of the spec.
+ *
+ * `whereEq` is a specialization of [`where`](#where).
```
davidchambers: Is this valid use of the word specialization?

scott-christopher: `whereEq` is the generalisation of `where`, while `where` is a specialisation of `whereEq`.

CrossEye: I would say so.

davidchambers:
((:bell:)) @buzzdecafe, @CrossEye

buzzdecafe:
i don't mind the complexity of the original. i don't really object to this, but it doesn't excite me either. i expect this may break a lot of stuff that uses `where` at present.

this is nice: `return where(mapObj(eq, spec), testObj);`
but i wonder how it performs? that was the reason for the nastiness of the original.

davidchambers:
> i expect this may break a lot of stuff that uses `where` at present.

Certainly. In most cases, users will need to replace `R.where` with `R.whereEq`:

```diff
-R.where({x: 1, y: 2})
+R.whereEq({x: 1, y: 2})
```

In each of the remaining cases, each non-function value in the spec must be wrapped with `R.eq`:

```diff
-R.where({x: 1, y: 2, z: R.gte(R.__, 0)})
+R.where({x: R.eq(1), y: R.eq(2), z: R.gte(R.__, 0)})
```

CrossEye:
Sorry, I thought I'd made it clear that I was in favor of this.

I do think that sometime soon, we'll have to look at the basic implementation.  If I'm not mistaken, in the Great Curry Debacle of 2014, we reset this one in such a way that it might have lost some of the benefits that had been gained by the initial manual currying we had done.  This one was the function that showed us the need for changes to our old currying, and it had some interesting behavior...

megawac:
Let's just make sure this is bold, highlighted, underlined in change log as
it will break a good amount of code

buzzdecafe:
yep, totally agree

buzzdecafe:
here is some evidence suggesting we can get a boost simply by defining `mapObjEq` in closure and using that: http://jsperf.com/mapobj-test

buzzdecafe:
not sure what to make of this: http://jsperf.com/mapobj-test

davidchambers:
I could certainly get behind a more efficient implementation. It was API complexity this pull request sought to reduce.

src/where.js
```diff
@@ -1,44 +1,45 @@
...
+ *      pred({a: 'foo', b: 'xxx', x: 11, y: 20}); //=> false
  */
 module.exports = _curry2(function where(spec, testObj) {
-  var parsedSpec = groupBy(function(key) {
-    return typeof spec[key] === 'function' ? 'fn' : 'obj';
-  }, keys(spec));
-
-  return _satisfiesSpec(spec, parsedSpec, testObj);
+  for (var prop in spec) {
+    if (_has(prop, spec) && !spec[prop](testObj[prop])) {
```
megawac: Any reason you're not using `keys` here (it'd be faster)?

davidchambers: Sounds like a good idea to me!

PR #1039, opened 2015-04-21
use R.toString to generate cache keys in R.memoize

davidchambers:
As mentioned in https://github.com/ramda/ramda/pull/924#issuecomment-88779624, using `R.toString` to generate cache keys makes `R.memoize` *much* more useful as it now works correctly for objects (!) and for dissimilar arrays which have the same string representation according to `Array.prototype.toString` (e.g. `[0]` and `['0']`).

The implementation is simpler as `repr` and `serialize` are no longer necessary.

### DIFF
--- a/test/memoize.js
+++ b/test/memoize.js
@@ -67,8 +67,22 @@ describe('memoize', function() {
   });
 
   it('differentiates values with same string representation', function() {
-    var f = R.memoize(function(x) { return x; });
-    assert.strictEqual(f(42), 42);
-    assert.strictEqual(f('42'), '42');
+    var f = R.memoize(R.toString);
+    assert.strictEqual(f(42), '42');
+    assert.strictEqual(f('42'), '"42"');
+    assert.strictEqual(f([[42]]), '[[42]]');
+    assert.strictEqual(f([['42']]), '[["42"]]');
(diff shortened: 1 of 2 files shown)


test/memoize.js
```diff
@@ -67,8 +67,22 @@ describe('memoize', function() {
...
+
+  it('respects object equivalence', function() {
+    var count = 0;
+    var f = R.memoize(function(x) {
+      count += 1;
+      return R.toString(x);
+    });
+    assert.strictEqual(f({x: 1, y: 2}), '{"x": 1, "y": 2}');
+    assert.strictEqual(f({x: 1, y: 2}), '{"x": 1, "y": 2}');
+    assert.strictEqual(f({y: 2, x: 1}), '{"x": 1, "y": 2}');
```
buzzdecafe: this is the case that interests me. :smiley:

buzzdecafe:
:cow2:

test/memoize.js
davidchambers: Object equivalence generally, or ignoring key enumeration order specifically?

buzzdecafe: > key enumeration order specifically

Issue #774, opened 2015-02-04
Is there a way to group an array into arrays of (maximum) given size?


mfirry:
I need something like:
`([a], n) -> [array1, array2]` (or an equivalent curried form) where `array1` and `array2` have maximum `length` of `n`?

davidchambers:
Do you want to transform something like `['a', 'b', 'c', 'd', 'e', 'f', 'g']` into `[['a', 'b', 'c'], ['d', 'e', 'f'], ['g']]`?

mfirry:
Yeah, with n = 3, that'd be it.

CrossEye:
There's nothing built in.  This is probably a job for recursion:

    var groupsOf = R.curry(function group(n, list) {
        return R.isEmpty(list) ? [] : R.prepend(R.take(n, list), group(n, R.drop(n, list)));
    });

I would almost certainly choose to put the `n` first, but obviously you could choose to organize it as you see fit.

mfirry:
I'll try it out. Thanks! And yes, silly me, n first makes perfect sense. 
It'd be in fact some sort of groupOfN kind of thing.
Thanks!

buzzdecafe:
you can also do it as a fold if you are worried about recursion stack:

```javascript
var subdivide = R.curry(function subdivide(n, ls) {
    function f(acc, elem, idx) {
        if (idx % n === 0) {
            return R.append([elem], acc);
        } else {
            var lastArr = R.last(acc);
            return R.append(R.append(elem, lastArr), R.init(acc));
        }
    }
    return R.foldlIndexed(f, [], ls);
});
```

but @CrossEye 's solution is a lot nicer

CrossEye:
Well, I did mine recursively because I couldn't think of a clean way to do it without recursion, so bravo!

tolmasky:
If there was an indexed version of groupBy, then you could just do:

groupByIndexed(array, R.unapply(R.pipe(R.prop(1), R.flip(R.divide)(3), parseInt)))

(that would give you { 0: ['a', 'b', 'c'], 1: ['d', 'e', 'f'], 2:[7] } , as opposed to [['a','b','c'], ['d'... )

mfirry:
I might be completely off but what I need seems what Clojure calls [partition-all](https://clojuredocs.org/clojure.core/partition-all)

stoeffel:
@mfirry Did just implement a [partition-all](https://github.com/stoeffel/partition-all) module out of fun.

[...]

PR #1042, opened 2015-04-21
batch for highland integration

buzzdecafe:


src/batch.js
```diff
@@ -0,0 +1,33 @@
...
+ */
+module.exports = _curry2(_checkForMethod('batch', function batch(n, list) {
+  var idx = 0;
+  var length = list.length;
+  var acc = [];
+  if (n < 1) {
+    return length === 0 ? [] : [list];
+  }
+  while (idx < length) {
+    acc[acc.length] = _slice(list, idx, Math.min(idx + n, length));
```
CrossEye: We really should move this check into `_slice`.  There's no reason that callers need to check the bounds before calling `_slice`.

CrossEye:
:herb:  Better than either of our solutions in #774.

davidchambers:
Of these names `splitEvery` is my clear favourite. It reads very well when partially applied: `R.splitEvery(2)`. Compare with `R.batch(2)` and `R.chunk(2)`, which are less descriptive.

src/batch.js
```diff
@@ -0,0 +1,33 @@
...
+ * @example
+ *
+ *      R.batch(2, [1, 2, 3, 4, 5]); //=> [[1, 2], [3, 4], [5]]
+ */
+module.exports = _curry2(_checkForMethod('batch', function batch(n, list) {
+  var idx = 0;
+  var length = list.length;
+  var acc = [];
+  if (n < 1) {
+    return length === 0 ? [] : [list];
```
davidchambers: This seems like two base cases. Can this be simplified?

buzzdecafe: well, we have to decide what to do based on both the length of `list` and the value of `n`. so i don't know if there is a simplification here, but I am open to suggestions.

src/batch.js
buzzdecafe: i was very surprised to discover that i had to do that. so i agree, internal `_slice` should be smart enough to handle the bounds.

buzzdecafe:
we good here?

davidchambers:
I'm a strong :-1: on the name.

CrossEye:
I'd rather take the opportunity while we're thinking about it to fix `_splice` and not put this call here.

Also I agree with @davidchambers on the name.  I'd be fine with either `chunk` or `splitEvery`.

src/batch.js
davidchambers: #1047

PR #1043, opened 2015-04-22
intersperse for highland integration

buzzdecafe:


src/intersperse.js
```diff
@@ -0,0 +1,31 @@
...
+ */
+module.exports = _curry2(_checkForMethod('intersperse', function intersperse(separator, list) {
+  var out = [];
+  var idx = -1;
+  var length = list.length;
+  while (++idx < length) {
+    if (idx === length - 1) {
+      out = out.concat(list[idx]);
+    } else {
+      out = out.concat(list[idx], separator);
```
megawac: Use push instead of concat

CrossEye:
Agree with @megawac, but other than that, LGTM.

src/intersperse.js
davidchambers: Didn't we decide to use `xs[xs.length] = x`? See #891.

I'd be happy for us to reverse that decision, but were we to do so it would be orthogonal to this change.

src/intersperse.js
```diff
@@ -0,0 +1,31 @@
...
+var _curry2 = require('./internal/_curry2');
+
+
+/**
+ * Creates a new list with the separator interposed between elements.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig * -> [a] -> [a]
```
davidchambers: Given that the return value has type `[a]` the separator must have type `a`.

src/intersperse.js
megawac: Push is just better in this case David

davidchambers: > Push is just better in this case

Because we're pushing two values? Fair enough.

I'd like to use `push` in all cases, for the record. :)

CrossEye: So would I.  I'm really not sure how we ended up in our current state.

src/intersperse.js
buzzdecafe: hmmm. really, the input array could be `[*]` and the return could be `[*]`, so the whole sig probably should be `* -> [*] -> [*]`.

CrossEye:
Do we describe these with the most generic signature that fits or the expected one?  Clearly `a -> [a] -> [a]` is nicer and captures the _likely_ use of the function.   But there is nothing to prevent, say, 

    intersperse (new Date, ['a', 1, /\d+/, {}, []]):

I don't know that we have any choice but to document the most generic version, but it certainly doesn't make me happy.

davidchambers:
I prefer `a -> [a] -> [a]` here. I prefer to capture the intended use of the function. If we're completely honest we can end up with types such as `* -> * -> *` which aren't particularly helpful.

buzzdecafe:
> I prefer a -> [a] -> [a] here. 

i'm fine with that too. will revise when able

CrossEye:
Go ahead, twist my arm.

Ok, ok, I'm convinced!

src/intersperse.js
```diff
@@ -0,0 +1,31 @@
...
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig a -> [a] -> [a]
+ * @param {*} separator The element to add to the list.
+ * @param {Array} list The list to be interposed.
+ * @return {Array} The new list.
+ * @example
+ *
+ *      R.interpserse('n', ['ba', 'a', 'a']); //=> ['ba', 'n', 'a', 'n', 'a']
```
mpajunen: Typo, inter**sp**erse

buzzdecafe: nice catch, thanks

buzzdecafe:
ping

:baby: :arrow_right: :older_man:

davidchambers:
:deciduous_tree:

anton-rudeshko:
Great and long awaited addition, thanks! I implemented same extension in lodash, but it wasn't accepted.

PR #1045, opened 2015-04-23
Update mapObjIndexed.js

lwassermann:
Category error. mapObjIndexed should be in the same category as mapObj.


buzzdecafe:
well spotted, thanks

Issue #1028, opened 2015-04-15
paritionIndexed?


jorda0mega:
Is there a way to pass in the index to the partition function to be able to partition by number of items much in the same way that clojure's partition works? Or would this have to be implemented through a separate function such as partitionIndexed?

davidchambers:
What's your objective? Do you wish to transform something like `['r', 'a', 'm', 'd', 'a']` into something like `[['r', 'a', 'm'], ['d', 'a']]`? It might be simpler to do something like `[R.take(3, xs), R.drop(3, xs)]` in this case.

jorda0mega:
given an array I want to partition it in equal parts of x items each. So something like this (say x is 4):
``` javascript
var a = [0,1,2,3,4,5,6,7,8,9,10,11];
R.partition(4, a) => [[0,1,2,3],[4,5,6,7],[8,9,10,11]]
```

davidchambers:
This was proposed in #774. Everyone seemed keen on the idea, so we'd welcome a pull request. :)

jorda0mega:
fair enough. I'll get on it :)

davidchambers:
> fair enough. I'll get on it :)

Terrific! Thank you!

tipdbmp:
I don't know how to curry... =)

        // http://clojuredocs.org/clojure.core/partition
        // (partition n coll) (partition n step coll) (partition n step pad coll)
        // doesn't curry
        function partition() {
            var args_len = arguments.length;
            var n, step, pad, coll;

            if (args_len === 2) {
                n    = arguments[0];
                coll = arguments[1];
                step = n;
                pad  = undefined;
            }
            else if (args_len === 3) {
                n    = arguments[0];
                step = arguments[1];
                coll = arguments[2];
                pad  = undefined;
            }
            else if (args_len === 4) {
                n    = arguments[0];
                step = arguments[1];
                pad  = arguments[2];
                coll = arguments[3];
            }
            else {
                throw "partition takes: 2, 3 or 4 positional arguments: " 
                    + arguments.length + " given";
            }

            var result = [];
            var coll_len = coll.length;

            var i = 0;
            while (i + n <= coll_len) {
                result.push(coll.slice(i, i + n));
                i += step;
            }

            if (i < coll_len && Array.isArray(pad)) {
                var rest_plus_pad = [];
                rest_plus_pad.push.apply(rest_plus_pad, coll.slice(i, coll_len));
[...]

PR #1050, opened 2015-04-26
adding chunk to ramda

jorda0mega:
adding chunk function which given an n and a list will split the list into n equal parts.


CrossEye:
See an alternative implementation and bikeshedding over the name in #1042.

CrossEye:
This description is a little off, as it doesn't split the list into `n` equal parts, but rather into parts of length `n`.

CrossEye:
Also see #1028 for a more general function (although not likely to be included as described.)

buzzdecafe:
the point of #1042 was to dispatch for highland stream. got spiked for its name. if we're not going to care about dispatch/integration, then i would also prefer the name `splitEvery`.

davidchambers:
Regardless of what we name *this* function, Highland integration will require some mapping of Ramda names to Highland names. This being the case, we should choose the name that is best for Ramda. `splitEvery` gets my vote. You could update the pull request to use this name, @jorda0mega.

src/chunk.js
```diff
@@ -0,0 +1,37 @@
...
+ * @example
+ *
+ *      R.chunk(3, [5,3,8,1,2,9,7]);
+ *      //=> [ [ 5, 3, 8 ],  [ 1, 2, 9 ], [ 7 ] ]
+ */
+
+module.exports = _curry2(function(n, list) {
+  var idx = 0;
+  return _reduce(function(acc, elt) {
+    var pos = Math.floor(idx / n);
```
buzzdecafe: what if `n` is zero?

CrossEye:
:+1: for `splitEvery`

src/chunk.js
CrossEye: It seems like it would work, although this implementation and yours have different behavior in the edge cases: http://bit.ly/1FrWRir

buzzdecafe: we should settle on what we want that behavior to be. the `batch` impl was compatible with highland semantics

jorda0mega:
sounds good. I can modify my code to change the name of the function and comply with batch edge cases

buzzdecafe:
> comply with batch edge cases

i think that is the way to go. thanks

luminarious:
I ran into a use-case where I need precisely this functionality and for a beginner it was quite unclear how to write a composable version that does this. So I'm happy I'm not the only one who needs this and I hope it'll be merged soon!

src/splitEvery.js
```diff
@@ -0,0 +1,39 @@
...
+ * @example
+ *
+ *      R.chunk(3, [5,3,8,1,2,9,7]);
+ *      //=> [ [ 5, 3, 8 ],  [ 1, 2, 9 ], [ 7 ] ]
+ */
+
+module.exports = _curry2(function(n, list) {
+  var idx = 0;
+  if (n < 1) {
+    return list.length === 0 ? [] : [list];
```
paldepind: How does wrapping the list in a list make sense when `n = 0`?

CrossEye: The only reasonable alternative I can think of is to throw an error, which I'm loathe to do.  A list consisting of an infinite number of copies of the empty list would be quite appropriate, but probably not practical.

paldepind: That is not what I meant. I don't understand how wrapping the list in a list makes sense. I'd rather just do `if (n < 1) return []`. That doesn't do anything meaningful either but at least it doesn't pretend to do something that makes sense.

In any cases users are not going to get what they expect so an error _would_ be ideal. Such a thing could be expressed in a fancy argument validation system! ;)

buzzdecafe: imagine if `list` is a stream. would you want to get back an empty stream? incidentally, as written would make `splitEvery` have the same semantics as `highland.batch`. that is one reason to do it this way.

CrossEye: Fair point.  I'm not sure which is better: give the user all her data in one chunk, or no data at all?  Neither makes a great deal of sense to me.  Your suggestion might be slightly better.

It's too bad no one's proposing a fancy argument validation system.  Surely any reasonable person would be persuaded.  :wink:

paldepind: I don't see how streams are different than lists in this regard. If I chunk a stream into parts of size 0 getting the entire stream wrapped in an array doesn't help me. The current behavior treats 0 like i meant "chunk of entire array". That doesn't make sense. If anything should be treated like that it should be Infinity.

My point is: if we can't return anything sensible, so we should throw an error do as little as possible as to signal "nothing I can do here, I'll bail out with this empty list". Otherwise the code pretends to be meaningful when it is not.

We could also return `[[]]`. That would be closer to the requested behavior because there would actually be a chunk of size 0. But again this doesn't _actually_ make sense so the code would again be pretending to make sense and trying to hide a user error.

If we can't justify other libraries behavior I see not reason to duplicate it. That would just be repeating other peoples mistakes.

But if anyone can explain the current behavior in a meaningful way I will rest my case!

paldepind: @CrossEye I didn't see your comment before my latest. What about returning `undefined`? That clearly communicates that what the user requested doesn't make sense.

buzzdecafe: >  If I chunk a stream into parts of size 0 getting the entire stream wrapped in an array doesn't help me. 

you would get a stream not an array.

paldepind: > you would get a stream not an array. 

Yes, of course. My mistake.

wires:
Ha, just needing this ;-) :+1:

src/splitEvery.js
CrossEye: I agree.  The current code is not robust enough.  I'm not a big fan of throwing errors, but that might be better than returning `null`, `undefined` or `[]`.  What do people think?

davidchambers: I'm in favour of throwing here. It's impossible to divide a :cake: into pieces of size zero. It's also impossible to divide a cake into pieces of negative size.

src/splitEvery.js
```diff
@@ -0,0 +1,39 @@
...
+  } else {
+    return _reduce(function(acc, elt) {
+      var pos = Math.floor(idx / n);
+      if (acc[pos]) {
+        var xs = acc[pos];
+        acc[pos][xs.length] = elt;
+      } else {
+        acc[acc.length] = [elt];
+      }
+      idx++;
```
davidchambers: Please `git rebase master`. You'll then need to update this to `idx += 1` to get `make test` to pass. :)

src/splitEvery.js
```diff
@@ -0,0 +1,39 @@
...
+module.exports = _curry2(function(n, list) {
+  var idx = 0;
+  if (n < 1) {
+    return list.length === 0 ? [] : [list];
+  } else {
+    return _reduce(function(acc, elt) {
+      var pos = Math.floor(idx / n);
+      if (acc[pos]) {
+        var xs = acc[pos];
+        acc[pos][xs.length] = elt;
```
davidchambers: @buzzdecafe and @CrossEye: Could we *please* use `Array.prototype.push` here?

```javascript
acc[pos].push(elt);
```

is a lot clearer than

```javascript
acc[pos][acc[pos].length] = elt;
```

src/splitEvery.js
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+module.exports = _curry2(function(n, list) {
+  var idx = 0;
+  if (n < 1) {
+    return list.length === 0 ? [] : [list];
+  } else {
+    return _reduce(function(acc, elt) {
+      var pos = Math.floor(idx / n);
+      if (acc[pos]) {
```
davidchambers: I very much like to avoid relying on type coercion, as we are doing here. Could we replace this with an expression which evaluates to a Boolean value? Perhaps:

```javascript
pos >= 0 && pos < acc.length
```

src/splitEvery.js
paldepind: :+1:

buzzdecafe: fine with me

CrossEye: Absolutely.  Somehow I thought we'd already swept through the code for that.  If not, I'd like to.  Whatever micro-optimization that used that awful style is long forgotten, and I'd like the code based on it to follow it into the sunset.

davidchambers: #1174

davidchambers:
Superseded by #1211

PR #1051, opened 2015-04-26
Patch bin build

cyberglot:
I've been using ramda in a good amount of projects. Most of them, it's only a build of some functions, since I usually don't need the lib as a whole. However, it's been problematic to use the version from github, instead of the one I usually grab from npm, to make this build. Then, I've made this PR in order to add the build script as a bin, so everybody would be able to make their own build right away from the npm version.
I know this PR may not be welcomed and may not be a good fit to your current workflow, but I'd like to give it a try. :smile: 

And thanks for your amazing work in this lib.
Cheers.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -12,7 +12,7 @@ var TRAVIS_NODE_VERSION = envvar.string('TRAVIS_NODE_VERSION', process.versions.
 var TRAVIS_TAG          = envvar.string('TRAVIS_TAG', '');
 
 var jsFiles = [
-  'scripts/build',
+  'bin/build',
   '**/*.js',
   '!{dist,lib/test,node_modules,bower_components}/**'
 ];
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ XYZ = node_modules/.bin/xyz --repo git@github.com:ramda/ramda.git --script scrip
 SRC = $(shell find src -name '*.js')
 
 
-dist/ramda.js: scripts/build scripts/header scripts/template.js $(SRC)
+dist/ramda.js: bin/build scripts/header bin/template.js $(SRC)
 	git checkout -- '$@'
 	scripts/header >ramda.js.tmp
 	'$<' --complete >>ramda.js.tmp


CrossEye:
While I would have no problem including the build script in our NPM output if there's a desire to do so, I'm missing the significance of moving it from `scripts` to `bin`.  Is this some standard location for NPM packages?

buzzdecafe:
https://docs.npmjs.com/files/folders#executables

I think this is a nice idea to simplify creating custom builds

CrossEye:
@buzzdecafe: I thought that link was specific to NPM's own files, not user files.  Am I just confused?

I'm not trying to raise objections.  I really don't care where this is.  I just don't understand why moving it makes it any easier to run.  (And of course we'd still have to include it in our NPM packaging, right?)

buzzdecafe:
> (And of course we'd still have to include it in our NPM packaging, right?)

indeed. but i am :+1: on that

CrossEye:
>>   (And of course we'd still have to include it in our NPM packaging, right?)

> indeed. but i am :+1: on that

Me too.  I have no objections here.  I'm just trying to understand how these changes help with the goal.  If we didn't make the change but simply started including the build file in the `scripts` directory, does that change anything?

jugoncalves:
I moved it to `bin` in order to keep the other scripts ignored to npm. But it would solve the problem to just "un-ignore" the script folder. I also thought I would be good to have the command-line to do it right from a global instalation or by using `node_modules/.bin/ramda`, but I think it should be renamed from `build` to `ramda` (I should've done it, sorry, my bad). But yes, feel free to close it and do it on your own way. :smile:

buzzdecafe:
i'd rather build on what you've started than close and start from scratch, if you're willing to stick with it.

CrossEye:
I'm fine with any of this.  One goal would be to move as much of the build process away from Grunt and Make and toward NPM as possible.  I don't know if this would help at all with that.

davidchambers:
Thanks for this pull request, @jugoncalves. :)

Some projects follow the convention, @CrossEye, of putting public executables in __bin__ and internal executables in __scripts__. @michaelficarra first suggested this to me, and I appreciate the separation it affords.

I agree, @jugoncalves, that we should rename the executable if we "export" it. `ramda` is a good option, as is `ramda-build`.

This pull request should be updated to specify the [`bin`][1] option in __package.json__. This way npm will create an alias in __node_modules/.bin__.

__template.js__ doesn't belong in __bin__, as it's not an executable. I believe it should live at the project root, as it did prior to #816.


CrossEye:
> Thanks for this pull request, @jugoncalves. :)

Yes, I should have said that, first thing.  I was just a bit confused!

Thanks for the explanation @davidchambers.  I think I'd prefer the name `ramda-build` to just `ramda`, but it's not a big deal.

> __template.js__ doesn't belong in __bin__, as it's not an executable. I believe it should live at the project root, as it did prior to #816.

I'd really love for it to have a home not in the root, honestly.  Although that was a good clean-up, to me the root is still too full of ancillary material.  But I'm not going to fight hard for this.

jugoncalves:
I think there's no problem on having `template.js` at `bin`, [mocha does it too](https://github.com/mochajs/mocha/tree/master/bin).  I think the name is the same one is going to use in the command-line, like:
```
$ ramda-build -- src/compose.js src/reduce.js src/filter.js > dist/ramda.custom.js
```
I am not 100% sure, but I think that's how it works.

yordis:
@CrossEye is this still an issue? Can we close this?

Issue #1055, opened 2015-04-29
support Map


davidchambers:
For 20 years JavaScript programmers have been using objects as maps. In some environments we now have the option of using [Map][1] which saves us from worrying about inherited properties and allows us to use values other than strings as keys. We could even think of objects as "map-like", in that they're an acceptable substitute for a limited set of use cases.

It would be nice to support real maps in Ramda. We could have a second version of each "object" function, or we could switch based on argument type. This is an option we've avoided in the past, but it may be justified in this case. We're not attempting to unify String and Array or Array and Object.

The switch would look something like this:

```javascript
if (typeof Map === 'function' && x instanceof Map) {
  // Map logic
} else {
  // Object logic
}
```

`R.assoc`, for example, would then have the following types:

    R.assoc :: String -> a -> {String: a} -> {String: a}
            :: a -> b -> Map a b -> Map a b

This issue was prompted by discussion in plaid/sanctuary#24 and plaid/sanctuary#25.

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

CrossEye:
This scares the heck out of me.

I certainly understand the desire, and agree.

But I'm scared of unmaintainable code.  I'm scared of losing the simplicity of our API.  I'm really scared of the other types we'd want to support.  If `Map`, why not `Set`?  Why not `WeakMap`?  And we've already discussed the idea that `assoc` should really support lists.

I don't know of any way to do this without such a `switch` statement.  If we could find something like our dispatch mechanism to deal with this, I'd be much happier.

buzzdecafe:
sacry sure, but it's gotta happen. not sure if we can just fold it into existing object functions, that may be a bit of a trick.

CrossEye:
Seeing the initial implementation, I'm much less concerned.  This is clearly reasonable.

If we get to the point where we have to check equality of keys, we might need  a new version of `eq`, or want to revisit our current `eq` a bit, because according to [ES6  § 23.1][es23.1], keys are compared according to the [SameValueZero][es7.2.10] comparison, for which -0 and 0 are identical.




  [es23.1]: https://people.mozilla.org/~jorendorff/es6-draft.html#sec-map-objects
  [es7.2.10]: https://people.mozilla.org/~jorendorff/es6-draft.html#sec-samevaluezero

krainboltgreene:
I would love to see this come to fruition. Is there anything I can do to help this happen?

[...]

PR #1056, opened 2015-04-30
partial support for Map

davidchambers:
As proposed in #1055. I'm opening this pull request for discussion. I've implemented a handful of "object" functions for Map. The remaining "object" functions could be implemented fairly easily, I imagine.

### DIFF
--- a/src/internal/_dispatchToMapMethod.js
+++ b/src/internal/_dispatchToMapMethod.js
@@ -1,0 +1,16 @@
+var _mapFunctions = require('./_mapFunctions');
+var _objectFunctions = require('./_objectFunctions');
+
+
+module.exports = (function() {
+  return typeof Map === 'function' ?
+    function _dispatchToMapMethod(methodname) {
+      return function() {
+        var isMap = arguments[arguments.length - 1] instanceof Map;
+        return (isMap ? _mapFunctions : _objectFunctions)[methodname].apply(null, arguments);
+      };
+    } :
+    function _callObjectMethod(methodname) {
+      return _objectFunctions[methodname];
+    };
+}());
--- a/src/internal/_assoc.js
+++ b/src/internal/_assoc.js
@@ -1,14 +1,4 @@
-var _eq = require('./_eq');
+var _dispatchToMapMethod = require('./_dispatchToMapMethod');
 
 
-module.exports = function _assoc(prop, val, obj) {
-  if (_eq(obj[prop], val)) {
-    return obj;
-  }
-  var result = {};
-  for (var p in obj) {
-    result[p] = obj[p];
-  }
-  result[prop] = val;
-  return result;
-};
+module.exports = _dispatchToMapMethod('assoc');
--- a/src/internal/_dissoc.js
+++ b/src/internal/_dissoc.js
@@ -1,9 +1,4 @@
-module.exports = function _dissoc(prop, obj) {
-  var result = {};
-  for (var p in obj) {
-    if (p !== prop) {
-      result[p] = obj[p];
-    }
-  }
-  return result;
-};
+var _dispatchToMapMethod = require('./_dispatchToMapMethod');
+
+
+module.exports = _dispatchToMapMethod('dissoc');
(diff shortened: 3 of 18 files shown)


CrossEye:
I definitely like this better than the idea of internal "switch" operations.  It's starting to scare me a bit less.

The interesting next take of course would be to see where our list functions could work on any arbitrary iterable...

src/internal/_dispatchToMapMethod.js
```diff
@@ -0,0 +1,10 @@
+var _mapFunctions = require('./_mapFunctions');
+var _objectFunctions = require('./_objectFunctions');
+
+
+module.exports = function _dispatchToMapMethod(methodname) {
+  return function() {
+    var isMap = typeof Map === 'function' && arguments[arguments.length - 1] instanceof Map;
+    return (isMap ? _mapFunctions : _objectFunctions)[methodname].apply(null, arguments);
+  };
+};
```
CrossEye: Or possibly the untested

```js
    return typeof Map === 'function' ? function _dispatchToMapMethod(methodname) {
        return function () {
            var isMap = arguments[arguments.length - 1] instanceof Map;
            return (isMap ? _mapFunctions : _objectFunctions)[methodname].apply(null, arguments);
        };
    } : function _callObjectMethod(methodname) {
        return _objectFunctions[methodname].apply(null, arguments);
    }
}());
```

As I'm not sure we want to check `typeof Map === 'function'` on every call.

buzzdecafe: :+1:

buzzdecafe:
> The interesting next take of course would be to see where our list functions could work on any arbitrary iterable...

we are already supporting iterables via transducers

src/internal/_dispatchToMapMethod.js
davidchambers: :zap:

benperez:
:-1:

benperez:
:100: This looks great!

src/internal/_dispatchToMapMethod.js
```diff
@@ -0,0 +1,18 @@
...
+  return typeof Map === 'function' ?
+    function _dispatchToMapMethod(methodname) {
+      return function() {
+        var isMap = arguments[arguments.length - 1] instanceof Map;
+        return (isMap ? _mapFunctions : _objectFunctions)[methodname].apply(null, arguments);
+      };
+    } :
+    function _callObjectMethod(methodname) {
+      return function() {
+        return _objectFunctions[methodname].apply(null, arguments);
```
buzzdecafe: can't we just return a reference to the appropriate function, rather than wrapping in a lambda? this would also preserve arity which would be nice

davidchambers: :zap:

wayneseymour:
+1

src/internal/_dispatchToMapMethod.js
```diff
@@ -0,0 +1,16 @@
...
+var _objectFunctions = require('./_objectFunctions');
+
+
+module.exports = (function() {
+  return typeof Map === 'function' ?
+    function _dispatchToMapMethod(methodname) {
+      return function() {
+        var isMap = arguments[arguments.length - 1] instanceof Map;
+        return (isMap ? _mapFunctions : _objectFunctions)[methodname].apply(null, arguments);
+      };
```
buzzdecafe: this lambda troubles me because it destroys arity.

buzzdecafe: oh i see, then you have to pass thru `curry`. yeesh. ok

buzzdecafe:
let's pull this post 0.14 or start an ES6 branch

davidchambers:
Post v0.14.0 I'll update this pull request to support Map objects in the remaining Object functions. On a related note, functions such as `R.intersection` and `R.union` should support Set objects. :)

buzzdecafe:
> Post v0.14.0 I'll update this pull request to support Map objects in the remaining Object functions. 

we are in the post-0.14 era :smile:

davidchambers:
I should revive this pull request.

CrossEye:
:herb:

davidchambers:
This pull request is squatting on a useful branch name, so I'll close it for now. I do hope to come back to this one day.

Issue #1011, opened 2015-04-06
Build pipeline suggestions.


scott-christopher:
There are a few tweaks that I would like to propose with the current build process and would love to get some feedback before I embark on anything further.

What I suggest:
- **Remove Grunt and the Makefile in favour of NPM scripts.**
  - There is currently quite a bit of duplication across all three that could be reduced down to a single source of truth.
  - Anything that requires more than a simple one-liner can be added as a shell or JS script in the `scripts` folder.
- **Use browserify for generating `dist/ramda.js`.**
  - I'm not aware of the history behind the `scripts/build` script and as such, unaware of anything additional that it is doing beyond resolving modules. If its only purpose is to resolve module dependencies, that logic could be taken care of by something such as the following shell script, which will build out a common-js and browser friendly `ramda.js`:
```sh
( cd src;
  find . -name '*.js' -not -path './internal/*' \
  | sed 's:\\./\\(.*\\)[.]js:exports.\\1 = require(\"./\\1\");:' \
  | browserify - -s R \
  > ../dist/ramda.js )
```
- **Customise the NPM publish step by copying all the contents of `src` and `package.json` into `dist`, alongside a freshly built `ramda.js` and `ramda.min.js`, then running `npm publish dist`.**
  - The tagging of a release should probably continue in its current form (i.e. with only `ramda.js` and `ramda.min.js` in dist) to prevent duplicated code in both `src` and `dist`.
  - This would require either updating `x.y.z` to support passing a path to `npm publish`, or creating a bespoke script to capture the version bump and release process.
  - This change would allow for use in a common-js environment such as:
```js
// load an individual function
var add = require('ramda/add');
// or alternatively load everything
var R = require('ramda');
```
- **Update the tests to require the individual function modules being tested, rather than requiring a built `R` module.**
  - Makes it simpler/quicker to run the tests in isolation if a build isn't required after every change.
  - Watching files for quicker test feedback becomes straight forward.

I appreciate these requests can be potentially controversial, so I am happy to accept a "Go home, there is no reason to change this" response, though I just wanted to raise them as issues I have noticed since contributing a little recently. And to be honest, the main motivator behind this for me is to try to get nicer common-js module support in before a v1.0 release.

buzzdecafe:
> Remove Grunt and the Makefile in favour of NPM scripts.

:+1: i have already started doing this in buzzdecafe/ramda "degrunt" branch. Any help gratefully accepted :smile:

[...]

PR #1057, opened 2015-05-01
hint, jscs, and test without grunt

buzzdecafe:
begins addressing #1011


package.json
```diff
@@ -32,9 +32,17 @@
...
+  "bin": {
+    "jscs": "node_modules/jscs/bin/jscs",
+    "jshint": "node_modules/jshint/bin/jshint",
+    "mocha": "node_modules/mocha/bin/mocha"
+  },
   "scripts": {
-    "jshint": "grunt jshint",
-    "test": "grunt jshint && grunt jscs && grunt browserify:client && grunt mochaTest:unit"
+    "clean": "rimraf dist",
+    "jscs": "jscs --config .jscsrc scripts/build **/*.js",
```
davidchambers: __.jscsrc__ is the default filename, so you could drop `--config .jscsrc`.

package.json
```diff
@@ -32,9 +32,17 @@
     "url": "git://github.com/ramda/ramda.git"
   },
   "main": "dist/ramda.js",
+  "bin": {
+    "jscs": "node_modules/jscs/bin/jscs",
+    "jshint": "node_modules/jshint/bin/jshint",
+    "mocha": "node_modules/mocha/bin/mocha"
+  },
```
davidchambers: Is this standard use of the `"bin"` field?

buzzdecafe: idk -- but it worked for aliasing things in `scripts`

package.json
buzzdecafe: nice, will repair

package.json
davidchambers: This is an abuse, I believe. It will add symlinks for these to a users __node_modules/.bin__ directory when she runs `npm install ramda`. Let's just specify the full paths.

By the way, you can use __node_modules/.bin/jscs__ rather than __node_modules/jscs/bin/jscs__ because JSCS specifies [`"bin"`][1] in its [__package.json__][2]. The same applies for the other two.


[2]: https://github.com/jscs-dev/node-jscs/blob/v1.13.1/package.json#L99-L101

scripts/testfiles
```diff
@@ -0,0 +1,23 @@
...
+
+'use strict';
+
+var fs = require('fs');
+var path = require('path');
+var p = path.resolve('./test/');
+
+fs.readdir(p, function(err, files) {
+  var filenames;
+  if (err) {
```
davidchambers: I'd prefer `err != null`. I hate to rely on implicit type coercion.

scripts/testfiles
```diff
@@ -0,0 +1,23 @@
...
+    throw err;
+  }
+
+  filenames = files.map(function(file) {
+    return path.join(p, file);
+  }).filter(function(file) {
+    return fs.statSync(file).isFile() &&
+           file.match(/\.js$/) &&
```
davidchambers: `/[.]js$/.test(file)` is better as it evaluates to a Boolean.

TheLudd:
Nice. npm as a build tool ftw!

buzzdecafe:
:smile: 
completely purging grunt may take a bit of work, but excising it from the primary build flow will be a big win IMO

package.json
```diff
@@ -66,12 +69,15 @@
     "handlebars": "^2.0.0",
     "highlight.js": "^8.4.0",
     "js-yaml": "^3.2.5",
+    "jscs": "^1.12.0",                                                                                                                           
+    "jshint": "^2.6.3",
```
davidchambers: Let's update jscs to `"1.13.x"` and jshint to `"2.7.x"` while we're at it. :)

scripts/testfiles
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env node
```
davidchambers: This file is not currently executable. You should run `chmod +x scripts/testfiles.`

scripts/testfiles
```diff
@@ -0,0 +1,23 @@
...
+
+  filenames = files.map(function(file) {
+    return path.join(p, file);
+  }).filter(function(file) {
+    return fs.statSync(file).isFile() &&
+           file.match(/\.js$/) &&
+           !(file.match('examplesRunner') || file.match('bundle.js'));
+  }).join(' ');
+  process.stdout.write(filenames);
+});
```
davidchambers: This script could be replaced with

    find test -type f -depth 1 -name '*.js' -not -name bundle.js -not -name test.examplesRunner.js

were in not for Windows. :\

davidchambers: Here's a nicer version:

```javascript
R.pipe(
  fs.readdirSync,
  R.filter(R.test(/[.]js$/)),
  R.reject(R.eq('bundle.js')),
  R.reject(R.eq('test.examplesRunner.js')),
  R.map(R.partial(path.resolve, 'test')),
  R.join('\n'),
  console.log
)('test');
```

test/keys.js
```diff
@@ -23,6 +23,7 @@ describe('keys', function() {
   });
 
   it('works for primitives', function() {
+    /* jshint elision: true */
```
davidchambers: What does this do?

buzzdecafe: jshint would warn about have `undefined` in an array

davidchambers: Oh I see, we have `[..., Infinity, , []]`. Thanks for the response. :)

buzzdecafe: my pleasure, thanks for the review

buzzdecafe: (this also suggests we are not be linting test files at present -- 'cuz this does not get flagged by jshint without this change)

buzzdecafe:
i'll make those fixes, but i should integrate browserify before this is merged. so please stand by

CrossEye:
Great stuff!

buzzdecafe:
let's wait post 0.14 to merge this one

package.json
```diff
@@ -33,45 +33,46 @@
   },
   "main": "dist/ramda.js",
   "scripts": {
-    "jshint": "grunt jshint",
-    "test": "grunt jshint && grunt jscs && grunt browserify:client && grunt mochaTest:unit"
+    "browserify": "node_modules/browserify/bin/cmd.js test/*.js --outfile lib/test/bundle.js",
```
scott-christopher: NPM scripts automatically add `node_modules/.bin` to `$PATH` when run, so these should be able to be trimmed to:
```json
"browserify": "browserify test/*.js --outfile lib/test/bundle.js",
"clean": "rimraf dist",
"jscs": "jscs scripts/build **/*.js",
"jshint": "jshint scripts/build **/*.js",
```

davidchambers: Very nice!

scripts/testfiles
davidchambers: Any `:heart:` for this version?

buzzdecafe: :+1:

buzzdecafe:
question: I've moved test/bundle.js and test/index.html to lib/test/. Should I check in bundle.js? You only need to create bundle.js if you want to run the test suite in a browser. (this will become mandatory once we spin up saucelabs testing again)

davidchambers:
> question: I've moved test/bundle.js and test/index.html to lib/test/. Should I check in bundle.js?

I don't think so. If it's derived from other files, why commit it?

buzzdecafe:
in that case, i should probably add it to .gitignore?

davidchambers:
Sounds good to me!

buzzdecafe:
if i had bothered to look, i would have seen test/bundle.js was in .gitignore. so there you go.

.travis.yml
```diff
@@ -5,5 +5,4 @@ node_js:
 - 'iojs-v1.5.1'
 before_script:
 - npm install -g grunt-cli
```
buzzdecafe: probably don't need this anymore either

buzzdecafe:
ok, everybody cool to do this?

davidchambers:
:deciduous_tree:

buzzdecafe:
sweet, more to come

PR #1058, opened 2015-05-02
remove lift and liftN

buzzdecafe:
while `lift` is certainly interesting, it is better suited to ramda-fantasy. In the context of ramda and lists, its most common uses are redundant:

* `liftN(1)` is `map`
* `liftN(2)` is `zipWith`

So `lift` doesn't start to contribute something new until `liftN(3)`, which might be better done in ramda as `zipAllWith` or some such. 

cf. https://github.com/ramda/ramda-fantasy/pull/31


davidchambers:
You should also open a pull request to deprecate `R.lift` and `R.liftN`. This pull request should not be merged until we've published a version with these functions marked as deprecated.

buzzdecafe:
right, forgot about that

CrossEye:
I'm don't think I like the idea of moving such functions into ramda-fantasy.

The reason for the dynamic dispatch is that we can work with many different data types, and that we don't have to think solely in terms of lists.  I believe our `lift` (especially with the slight change I suggested in https://github.com/ramda/ramda-fantasy/pull/31#issuecomment-97617498) is a useful abstraction that allows us to work with any compliant [Monad][mo], [Applicative][av], or even [Apply][ap] instance.  If we move it to r-f, then someone who wanted to use it with another [FantasyLand][fl] implementation would also have to import r-f.

To my mind, the only things that belong in ramda-fantasy are the actual specific algebraic types and the tests and other infrastructure needed to support them.  Any generic tools that operate on their interface belongs in Ramda itself, even if it's not particularly list-based.


  [av]: https://github.com/fantasyland/fantasy-land#apply
  [ap]: https://github.com/fantasyland/fantasy-land#applicative
  [mo]: https://github.com/fantasyland/fantasy-land#monad

buzzdecafe:
closing for now. not sure i agree, but will circle back to it later

ivan-kleshnin:
> liftN(2) is zipWith

Is this a correct statement?!

```js
let X = ["A", "B", "C", "D"];      // possible letters for first character
let Y = ["1", "2", "3", "4", "5"]; // possible numbers for second character

let lifter1 = lift((x, y) => x + y);
let lifter2 = liftN(2, (x, y) => x + y);
let lifter3 = zipWith((x, y) => x + y);

console.log(lifter1(X, Y)); // ['A1', 'A2'...'D5']
console.log(lifter2(X, Y)); // ['A1', 'A2'...'D5']
console.log(lifter3(X, Y)); // ['A1', 'B2', 'C3', 'D4' ]

```

Issue #1035, opened 2015-04-20
Function Definition Syntax


ccorcos:
I'm confused about what this means:

Converge Function
 ((*… → c) → (((* → a), (* → b), …) → c)

After staring at the example, I can figure out what converge does, but I still don't get what this means...

CrossEye:
... with good reason.   That's far off-base.   When I'm in front of a computer, I'll try to write up an accurate version.

CrossEye:
The best I can come up with is the pretty awful:

    (x1 -> x2 -> ... -> xn -> z) -> (a -> b -> ... -> x1) -> (a -> b -> ... -> x2) -> 
        ... (a -> b -> ... -> xn) -> (a -> b -> ... -> z)

Anyone have something better?

bergus:
Can we have that tuple back maybe? It made sense to me as the function is surely not curryed (or is it?)

    (x1->x2->…->xn -> z) -> ((a->b->… -> x1), (a->b->… -> x2), …, (a->b->… -> xn)) -> (a->b->… -> z)

I think we even can omit the `n` case for brevity

    (x1->x2->… -> z) -> ((a->b->… -> x1), (a->b->… -> x2), …) -> (a->b->… -> z)

CrossEye:
There is a (mostly-irrelevant) currying to three places there.  But you're right; that tuple makes it clearer.  And you're also right that the `n` is not too important.  Very nice changes.

lucasmreis:
Whenever I use `converge` I ident it this way:
```
converge(
  doSomething,
    transformA,
    transformB,
    transformC)
```
So, I can visualize better what happens when it's also idented:
```
(x1 -> x2 -> ... -> xn -> z) -> 
  (a -> b -> ... -> x1) -> 
  (a -> b -> ... -> x2) -> 
  ... 
  (a -> b -> ... -> xn) -> 
  (a -> b -> ... -> z)
```
Does it help?

CrossEye:
That's a nice visualization, although I would outdent the response.  Unfortunately, it doesn't really fit inside our documentation.  Still, it's a very helpful way to explain the function.  Thanks.

ccorcos:
Ahh. That make much more sense.

CrossEye:
We need to do something similar to `useWith`.  It should be

    (x1->x2->… -> z) -> ((a->x1), (b->x2), …) -> (a->b->… -> z)

ccorcos:
[...]

PR #1061, opened 2015-05-03
Creates a common 'addIndex' function.  Closes #484.

### LINKED ISSUE #484: Replace all the .idx functions with a common addIdx

CrossEye:
This has been a long time coming.  It addresses only list function, not object ones.

### DIFF
--- a/src/addIndex.js
+++ b/src/addIndex.js
@@ -1,0 +1,39 @@
+var _concat = require('./internal/_concat');
+var _curry1 = require('./internal/_curry1');
+var _prepend = require('./internal/_prepend');
+var _slice = require('./internal/_slice');
+var curryN = require('./curryN');
+
+
+/**
+ * Creates a new list iteration function from an existing one by adding two new parameters
+ * to its callback function: the current index, and the entire list.
+ *
+ * This would turn, for instance, Ramda's simple `map` function into one that more closely
+ * resembles `Array.prototype.map`.  Note that this will only work for functions in which
+ * the iteration callback function is the first parameter, and where the list is the last
+ * parameter.  (This latter might be unimportant if the list parameter is not used.)
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @category List
+ * @sig ((a ... -> b) ... -> [a] -> *) -> (a ..., Int, [a] -> b) ... -> [a] -> *)
+ * @param {Function} fn A list iteration function that does not pass index/list to its callback
+ * @return An altered list iteration function thats passes index/list to its callback
+ * @example
+ *
+ *      var mapIndexed = R.addIndex(R.map);
+ *      mapIndexed(function(val, idx) {return idx + '-' + val;}, ['f', 'o', 'o', 'b', 'a', 'r']);
+ *      //=> ['0-f', '1-o', '2-o', '3-b', '4-a', '5-r']
+ */
+module.exports = _curry1(function(fn) {
+  return curryN(fn.length, function() {
+    var idx = -1;
+    var origFn = arguments[0];
+    var list = arguments[arguments.length - 1];
+    var indexedFn = function() {return origFn.apply(this, _concat(arguments, [++idx, list]));};
+
+    return fn.apply(this, _prepend(indexedFn, _slice(arguments, 1)));
+  });
+});
--- a/src/mapIndexed.js
+++ b/src/mapIndexed.js
@@ -16,6 +16,8 @@ var _curry2 = require('./internal/_curry2');
  * @param {Function} fn The function to be called on every element of the input `list`.
  * @param {Array} list The list to be iterated over.
  * @return {Array} The new list.
+ * @deprecated since v0.14.0
+ * @see R.addIndex
  * @example
  *
  *      var squareEnds = function(elt, idx, list) {
--- a/test/addIndex.js
+++ b/test/addIndex.js
@@ -1,0 +1,74 @@
+var assert = require('assert');
+
+var R = require('..');
+
+
+describe('addIndex', function() {
+  describe('works with unary functions like `map`', function() {
+    var times2 = function(x) {return x * 2;};
+    var addIndexParam = function(x, idx) {return x + idx;};
+    var squareEnds = function(x, idx, list) {
+      return (idx === 0 || idx === list.length - 1) ? x * x : x;
+    };
+    var mapIndexed = R.addIndex(R.map);
+
+    it('working just like a normal map', function() {
+      assert.deepEqual(mapIndexed(times2, [1, 2, 3, 4]), [2, 4, 6, 8]);
+    });
+
+    it('passing the index as a second parameter to the callback', function() {
+      assert.deepEqual(mapIndexed(addIndexParam, [8, 6, 7, 5, 3, 0, 9]), [8, 7, 9, 8, 7, 5, 15]); // [8 + 0, 6 + 1...]
+    });
+
+    it('passing the entire list as a third parameter to the callback', function() {
+      assert.deepEqual(mapIndexed(squareEnds, [8, 6, 7, 5, 3, 0, 9]), [64, 6, 7, 5, 3, 0, 81]);
+    });
+
+    it('acting as a curried function', function() {
+      var makeSquareEnds = mapIndexed(squareEnds);
+      assert.deepEqual(makeSquareEnds([8, 6, 7, 5, 3, 0, 9]), [64, 6, 7, 5, 3, 0, 81]);
+    });
+  });
+
+  describe('works with binary functions like `reduce`', function() {
+    var reduceIndexed = R.addIndex(R.reduce);
+    var timesIndexed = function(tot, num, idx) {return tot + (num * idx);};
+    var objectify = function(acc, elem, idx) { acc[elem] = idx; return acc;};
+
+    it('passing the index as a third parameter to the predicate', function() {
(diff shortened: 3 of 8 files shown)


src/addIndex.js
```diff
@@ -0,0 +1,40 @@
...
+var curryN = require('./curryN');
+
+
+/**
+ * Creates a new list iteration function from an existing one by adding two new parameters
+ * to its callback function: the current index, and the entire list.
+ *
+ * This would turn, for instance, Ramda's simple `map` function into one that more closely
+ * resembles `Array.prototype.map`.  Note that this will only work for functions in which
+ * the iteration callback function is first the first parameter, and where the list is the
```
davidchambers: s/is first the first parameter/is the first parameter/

src/addIndex.js
```diff
@@ -0,0 +1,40 @@
...
+ * @memberOf R
+ * @category Function
+ * @category List
+ * @sig ((a -> b) -> ... -> [a] -> *) -> (a -> Int -> [a] -> b) -> ... -> [a] -> *)
+ * @param {Function} fn A list iteration function that does not pass index/list to its callback
+ * @return An altered list iteration function thats passes index/list to its callback
+ * @example
+ *
+ *      var mapIndexed = R.addIndex(R.map);
+ *      mapIndexed(function(val, idx) {return idx + '-" + val;}, ['f', 'o', 'o', 'b', 'a', 'r']);
```
davidchambers: s/"/'/

src/mapIndexed.js
```diff
@@ -16,6 +16,8 @@ var _curry2 = require('./internal/_curry2');
  * @param {Function} fn The function to be called on every element of the input `list`.
  * @param {Array} list The list to be iterated over.
  * @return {Array} The new list.
+ * @deprecated since v0.14.0
+ * @see R.addIndex
```
davidchambers: :grinning:

test/addIndex.js
```diff
@@ -0,0 +1,74 @@
...
+      return (idx === 0 || idx === list.length - 1) ? x * x : x;
+    };
+    var mapIndexed = R.addIndex(R.map);
+
+    it('working just like a normal map', function() {
+      assert.deepEqual(mapIndexed(times2, [1, 2, 3, 4]), [2, 4, 6, 8]);
+    });
+
+    it('passing the index as a second parameter to the callback', function() {
+      assert.deepEqual(mapIndexed(addIndexParam, [8, 6, 7, 5, 3, 0, 9]), [8 + 0, 6 + 1, 7 + 2, 5 + 3, 3 + 4, 0 + 5, 9 + 6]);
```
davidchambers: This example would be clearer with a shorter input list.

davidchambers:
This is very exciting to see!

test/addIndex.js
CrossEye: Yes, but would get [stuck in your head][tt] much less often.

  [tt]: https://www.youtube.com/watch?v=6WTdTwcmxyo

davidchambers: I remember having a landline with a seven-digit phone number, and that was only after a digit was added to our six-digit number. :)

test/addIndex.js
```diff
@@ -0,0 +1,74 @@
...
+describe('addIndex', function() {
+  describe('works with unary functions like `map`', function() {
+    var times2 = function(x) {return x * 2;};
+    var addIndexParam = function(x, idx) {return x + idx;};
+    var squareEnds = function(x, idx, list) {
+      return (idx === 0 || idx === list.length - 1) ? x * x : x;
+    };
+    var mapIndexed = R.addIndex(R.map);
+
+    it('working just like a normal map', function() {
```
buzzdecafe: caveman test?

src/mapIndexed.js
buzzdecafe: :tada:

test/addIndex.js
CrossEye: It's hard to decide if it's the source code or the output that we want to look correct.

In other words, ooga ooga.

buzzdecafe: :fire: bad

davidchambers:
Squash and merge, I say. :)

buzzdecafe:
ya this is a BFD. well done

CrossEye:
One other question before merging this.  The signature I've given this seems overly complex, but I can't think of anything simpler that's also fairly accurate.  Is there a better version?

buzzdecafe:
` @sig ((a... -> b) -> x? -> [a] -> *) -> ((a..., Int, [a] -> b) -> x? -> [a] -> x)`

I think this is only slightly less ugly. the issue is that this function is doing a lot of magic work to handle `map` and `reduce` at one go.

CrossEye:
I like dropping the initial arrows, and switching to commas is fine.  But I don't buy ` -> x? ->` to represent an unknown number of parameters.  The function would work fine on

    (a -> b) -> c -> d -> e -> [a] -> x

transforming it into

    (a -> Int -> [a] -> b)  -> c -> d -> e -> [a] -> x

I don't feel that `x?` is a good representation of that `c -> d -> e`.

wayneseymour:
I'm sorry I've very little to contribute, but where best to go learn the syntax for your signatures?

buzzdecafe:
@wayneseymour our sigs are like pidgin Haskell sigs. You are better off looking at Haskell (or ML) sigs and then seeing how we are polluting them

wayneseymour:
@buzzdecafe Hahahaha, thanks man!  :)

buzzdecafe:
@wayneseymour you can only graft so much strong functional typing onto js before the cracks begin to show. The sig we're discussing here is a good (bad?) example of that ...

wayneseymour:
@buzzdecafe 
Ah, makes sense. I'll be looking into Haskell's types very soon.

CrossEye:
@wayneseymour: I gave a partial description of it recently: https://github.com/ramda/ramda/issues/1035#issuecomment-95327214

davidchambers:
> I don't feel that `x?` is a good representation of that `c -> d -> e`.

I've been using `?` to mean "or null/undefined". So `String -> String?`, for example, is a function from String to String/null/undefined.

wayneseymour:
@CrossEye
  Ah, thanks

buzzdecafe:
> I've been using ? to mean "or null/undefined". 

That seems reasonable. The problem with the function here is that it really needs *two* sigs, one for `map` type functions, and one for `reduce` type functions, since the `x?` may not be there at all.

CrossEye:
But it's much more generic than that.  It doesn't care _what_ your signature is like as long as the iteration callback comes first and the list comes last.  That's what I want to capture, in any case.

buzzdecafe:
are we just waiting on sig for this? let's merge the code and fix the docs later :smile:

CrossEye:
Somehow I thought I'd already merged this one!

Brainfart.

CrossEye:
I guess I need to update it now, though, since all the deprecation warnings say "since 0.14"...

buzzdecafe:
merging. will fix deprecated notices in a separate pr

PR #1063, opened 2015-05-04
fixes #1059

### LINKED ISSUE #1059: `R.converge` does not pass `this` context to its arguments
I hope this is not intended behavior.

https://github.com/ramda/ramda/blob/master/src/converge.js#L38-L40

There should be `_map(function () { ... }.bind(this), fns)`.

buzzdecafe:


test/converge.js
```diff
@@ -26,6 +26,17 @@ describe('converge', function() {
...
+  it('passes context to its functions', function() {
+    var a = function(x) { return this.f1(x); };
+    var b = function(x) { return this.f2(x); };
+    var c = function(x, y) { return this.f3(x, y); };
+    var d = R.converge(c, a, b);
+    var context = {f1: R.add(1), f2: R.add(2), f3: R.add};
+    assert.equal(a.call(context, 1), 2);
+    assert.equal(b.call(context, 1), 3);
+    assert.equal(d.call(context, 1), 5);
+  });
```
davidchambers: For some reason I find this test case difficult to follow.

CrossEye: I'm not sure there's much help for that.

We could take out the first two (passing) asserts, which don't show the current issue -- or possibly just comment them out, leaving them in place as documentation of where the current regression is.

But other than this, it seems pretty well a minimal case to demonstrate the issue.

I suppose we could add an annotation, but it would probably end up fairly cumbersome:

    d.call(context, 1); //=> (R.converge(c, a, b)).call(context, 1); //=>
    (R.converge(function(x, y) {return this.f3(x, y);}, 
                         function(x) {return this.f1(x);},
                         function(x) {return this.f2(x);}
    )).call(context, 1); //=>
    R.converge(function(x, y) {return context.f3(x, y);}, 
                         (function(x) {return context.f1(x);})(1),
                         (function(x) {return context.f2(x);})(1)
    ); //=>
    R.converge(function(x, y) {return R.add(x, y);}, 
                         (function(x) {return R.add(1)(x);})(1),
                         (function(x) {return R.add(2)(x);})(1)
    ); //=>
    R.converge(function(x, y) {return R.add(x, y);}, 
                         (function(x) {return R.add(1)(x);})(1),
                         (function(x) {return R.add(2)(x);})(1)
    ); //=>
    (function(x, y) {return R.add(x, y);})( 
                         (function(x) {return R.add(1)(x);})(1),
                         (function(x) {return R.add(2)(x);})(1)
    ); //=>
    (function(x, y) {return R.add(x, y);})(2, 3); //=> 5

buzzdecafe:
?

CrossEye:
:herb:

davidchambers:
:deciduous_tree:

PR #1066, opened 2015-05-06
Adds reduceRightLazy

scott-christopher:
Not sure if there is any interest for inclusion of this, or whether transducers should be promoted to cover this use case instead. (I'm not familiar enough with transducers yet to say one way or another)

This introduces `R.reduceRightLazy`, which is a slight variation of `R.reduceRight` whereby the accumulated value is evaluated lazily, allowing for short-circuiting to prevent iterating over the entire list.

e.g.
```js
function exists(p, list) {
  return R.reduceRightLazy(function(acc, a) {
    return p(a) || acc();
  }, false, list);
}
```

### DIFF
--- a/src/reduceRightLazy.js
+++ b/src/reduceRightLazy.js
@@ -1,0 +1,44 @@
+var _curry3 = require('./internal/_curry3');
+var _slice = require('./internal/_slice');
+
+
+/**
+ * Reduces a list by passing each element to the provided iterator
+ * function, along with a thunk (zero arity function) that evaluates
+ * to the accumulated result of reducing all elements to the right of
+ * the current element.
+ *
+ * This function operates in a similar manner to `R.reduceRight`, with
+ * the exception that the accumulated value is evaluated lazily, such
+ * that the iterator function must execute the thunk to get access to
+ * the value. If the iterator chooses to not evaluate the thunk, this
+ * function will short circuit and return the current value without
+ * iterating over the entire list.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig ((() -> b, a) -> b) -> b -> List[a] -> b
+ * @param {Function} fn The iterator function. Receives two values,
+ *        a thunk (zero arity function) that evaluates the accumulated
+ *        value and the current element from the array.
+ * @param {*} acc The initial accumulated value.
+ * @param {Array} list The list to reduce
+ * @return {*} The final accumulated value
+ * @see R.reduceRight
+ * @example
+ *
+ *      function exists(p, list) {
+ *        return R.reduceRightLazy(function(acc, a) {
+ *          return p(a) || acc();
+ *        }, false, list);
+ *      }
+ *
+ *      exists(R.eq(4), [2,4,6,8]); //=> true
+ */
+module.exports = _curry3(function reduceRightLazy(fn, acc, list) {
+  return list.length === 0 ? acc : fn(
+    function() { return reduceRightLazy(fn, acc, _slice(list, 1)); },
+    list[0]
+  );
+});
--- a/test/reduceRightLazy.js
+++ b/test/reduceRightLazy.js
@@ -1,0 +1,43 @@
+var assert = require('assert');
+
+var R = require('..');
+
+
+describe('reduceRightLazy', function() {
+  var plusThunk = function(a, b) { return a() + b; };
+
+  it('folds lists with right-associativity', function() {
+    assert.strictEqual(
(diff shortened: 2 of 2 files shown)


CrossEye:
This looks quite interesting.  And I was intrigued by the gist mentioned on Gitter.

I'm trying to figure out why the callback is not also supplied the accumulated value.  It strikes me that such might often be used to decide whether to continue or not.  Is there some good reason not to include it?

scott-christopher:
If I understand your question correctly, it's largely because the accumulated value doesn't yet exist until its wrapping function is called. This highlights the point that right folds don't necessarily just iterate over a function from right to left applying a function to each value, but are instead right-associative.

So in the diagram below borrowed from Wikipedia, you can see that `[1, 2, 3, 4, 5]` applied to `foldr f z`, where `f` is a reducing function of type `b -> a -> b`<sup>*</sup> and `z` is the initial accumulator, the result is essentially `f(foldr(f, z, [2, 3, 4, 5]), 1)`.

Right-fold-transformation

One way right folds have been described that clicked with me (for lists at least) is you can think of it as swapping out the cons function of a list with the reducing function, and the empty list with the initial accumulator value.

So the desugared cons form of `[1, 2, 3, 4, 5]` can be represented as:
```
1 : (2 : (3 : (4 : (5 : [])))
```
and `foldr (+) 0 [1, 2, 3, 4, 5]` can be represented as:
```
1 + (2 + (3 + (4 + (5 + 0))))
```
Note that `foldr (:) [] [1, 2, 3, 4, 5]` is equivalent to the cons form of the same list:
```
1 : (2 : (3 : (4 : (5 : []))) -- cons form of list
1 : (2 : (3 : (4 : (5 : []))) -- right-associative cons
```

So you can think of `reduceRightLazy` as producing the above as `f(() => foldr(f, z, [2, 3, 4, 5]), 1)`, where the accumulated value isn't evaluated until it is called, as needed.

I hope that actually answers your question. If not I apologise for the lesson in right folds :)

<sup>*</sup> *right folds typically have an argument order of `a -> b -> b` though I have kept with the conventions of the existing reduceRight function in Ramda*

CrossEye:
I do understand the difference between right and left folds, but this is a particularly well-done explanation.  Thanks.

What I was suggesting makes no sense:

```js
module.exports = _curry3(function reduceRightLazy(fn, acc, list) {
  return list.length === 0 ? acc : fn(
    function() { return reduceRightLazy(fn, acc, _slice(list, 1)); },
    list[0],
    acc  // <-- This is what's new
  );
});
```

That will do nothing useful (with left- or right-folds) if the fold is lazy.

Sorry about that.  What I really want has to do with early escape from folds, but not with laziness.

scott-christopher:
One approach to address that use case is what I believe @joneshf was referring to in gitter by being similar to [`unfold`](https://hackage.haskell.org/package/base-4.8.0.0/docs/Data-List.html#v:unfoldr), which is used for building a list using a builder function that returns `Just a` where `a` is the next value or `Nothing` to signal that `unfold` should return.

Following that convention, something like the following could do the trick (don't mind the mutable vars :D):
```js
function partialReduceRight(fn, initAcc, list) {
  var idx = list.length, acc = initAcc, maybeAcc;
  while (--idx >= 0) {
    maybeAcc = fn(acc, list[idx]);
    if (maybeAcc === void 0) {
      break;
    } else {
      acc = maybeAcc;
    }
  }
  return acc;
}
```

This breaks out of the reduce upon receiving an `undefined` value. The use of `undefined` could be swapped with a `Maybe` type of another sentinel value like `R.__` if `undefined` should ever be considered a valid result of a reduce.

Again, I don't know if this is on par with what you're thinking of, but it could also be useful.

CrossEye:
I'm hoping others will have strong opinions on this, because I don't have one.

While I can see some potential, I don't see it as anything close to critical.  Perhaps I'm looking for a stronger use-case than `exists`, which is a more anemic version of `find` or `findIndex`, which also short-circuit, although clearly not in as generic a manner.

scott-christopher:
I think I've talked myself out of it too.

The shortcutting behaviour can really only return based on the value of a single element, such as `exists`, or `first` which would return the first element satisfying a given predicate.

The downside here is that compared to the bespoke `while` loop variations of these use cases, this isn't stack friendly for large lists.

I'll close this for now.

scott-christopher:
... And what I was describing as `first` already exists as `R.find` and `exists` as `R.any`.

CrossEye:
Thanks for the interesting ideas, in any case!

PR #1075, opened 2015-05-09
Noting issue with empty lists.  Fixes #1071

### LINKED ISSUE #1071: Consider the API of maxBy/MinBy
https://github.com/ramda/ramda/pull/456#discussion_r29909778

If passed an empty list, these functions don't return anything.  Should we accept a default/initial value?  Is there any other possibility?  Or is this simply a matter of _caveat emptor_?

CrossEye:


buzzdecafe:
:cow2:

Issue #1025, opened 2015-04-14
returning a reference to input object


davidchambers:
This has been discussed in #1018 and elsewhere. Let's continue the discussion here.

The question is whether it's okay for a function which usually returns a *new* object to return the input object in certain cases. Is it okay for `R.slice(0, Infinity, xs)` to return `xs`, for example, or should it return a shallow clone of `xs`?

paldepind:
I agree with @megawac. A function that sometimes returns a new object and sometimes returns the passed object seems inconsistent to me. Then I suddenly can't rely on it to return a new object anymore.

In the cases where the difference matters I'll have a harder time reasoning about what might happen. I'll have to ask myself "could this code in any circumstances call `slice` with the specific arguments for which it won't clone it's input?" and if I conclude that the answer is "yes" I'll have to clone manually. And if I mistakenly conclude that the answer is "no" I've just introduced a bug in my program.

This is slightly similar to how functions that are sometimes synchronous and sometimes asynchronous are a disaster. Imagine if `Array.prototype.slice` had returned the exact array when passed with 0. As in `arr === arr.slice(0)`. That would have been bad right?

I think this introduces a lot of uncertainty to the API. Does `R.sort` return the passed in list if it is already sorted? I no longer know.

buzzdecafe:
still mulling it over, but i'm pretty sure i'm on the other side. IMO we shouldn't be in the business of _enforcing_ behavior. If you want to mutate your objects, then _caveat emptor_.

> A function that sometimes returns a new object and sometimes returns the passed object seems inconsistent to me.

only if you are relying on reference equality, and that seems a bit weird. As far as I can see, what you are recommending is:

```javascript
var p = {x: 1}; 
var q = R.identity(p);
p === q; //=> false, since q is a clone
```

and that would be *very* surprising to me.

megawac:
I'm done arguing this but it seems like a new user trap; consider someone
who decides to convert a codebase. This one may be minor but I hope no
others follow suit

paldepind:
> only if you are relying on reference equality, and that seems a bit weird.

No. The function is inconsistent because is sometimes does one thing and sometimes does another thing. A function that returns a new object provides two guarantees:

1/ You can modify the returned object without affecting anybody.
2/ Nobody except you can modify this object.

[...]

PR #1076, opened 2015-05-10
Revert "`adjust` and `assoc` now returns the same instance if the val…

buzzdecafe:
Reverts ramda/ramda#998

see discussion in https://github.com/ramda/ramda/issues/1025

### DIFF
--- a/src/adjust.js
+++ b/src/adjust.js
@@ -1,7 +1,5 @@
 var _concat = require('./internal/_concat');
 var _curry3 = require('./internal/_curry3');
-var _eq = require('./internal/_eq');
-
 
 /**
  * Applies a function to the value at the given index of an array,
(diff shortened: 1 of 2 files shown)


davidchambers:
:deciduous_tree:

CrossEye:
:herb:

etherealite:
Ramda V0.21.0 adjust has a confusing behavior where its return type is incongruent with its input. When given an empty string, it returns a string, but when given a non-empty string, it returns an array of single character strings.

```
R.adjust(R.identity,0, "")  //=> ""
R.adjust(R.identity,0, "abc")  //=> ["a", "b", "c"]
```

To me, this behavior feels like a confusing 'gotcha'. I would expect the type of the return to be an array in both of the above cases.

buzzdecafe:
@etherealite i agree that is bad behavior. Do you think it would be better to simply return `[ ]` or throw since String is an invalid input type? Or some other behavior?

I'm leaning more and more in the direction of :boom: rather than try to paoper over it.

etherealite:
Plus one for the Michael Bay option.

PR #1083, opened 2015-05-13
lint: allow unused parameters

davidchambers:
See plaid/sanctuary#31

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -7,5 +7,5 @@
   "newcap": false,
   "predef": ["beforeEach", "console", "define", "describe", "it", "module", "process", "require"],
   "undef": true,
-  "unused": true
+  "unused": "vars"
 }
--- a/src/arity.js
+++ b/src/arity.js
@@ -30,16 +30,16 @@ var _curry2 = require('./internal/_curry2');
 module.exports = _curry2(function(n, fn) {
   switch (n) {
     case 0: return function() {return fn.apply(this, arguments);};
-    case 1: return function(a0) {void a0; return fn.apply(this, arguments);};
-    case 2: return function(a0, a1) {void a1; return fn.apply(this, arguments);};
-    case 3: return function(a0, a1, a2) {void a2; return fn.apply(this, arguments);};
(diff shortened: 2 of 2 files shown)


buzzdecafe:
hmmm. since it's just for that one function, seems to me it would be neater to comment that one function.

davidchambers:
It's a philosophical question. I believe I have the right to give each function I define a `length` of my choosing (without using `void` or `R.arity`). JSHint currently prevents me from doing something valid.

CrossEye:
:herb:

JSHint always prevents us from doing valid things; that's its job.  If it weren't valid, the interpreter would catch it.  We'll take out unused parameters where we see them, and if a few sneak by, I don't think it's the end of the world.

buzzdecafe:
this means it is up to us to detect when unused params/vars occur. it's nice to have the tool catch that, especially since it is not just the three of us contributing. Let's follow the principle of least privilege. Clearly, we all agree jshint is a PITA in `arity`. I argue it should be bypassed there, and left in force everywhere else.

CrossEye:
I don't have strong feelings on this.  I hate cluttering my source code with comments meant for the tools, but sometimes its a necessary evil.

davidchambers:
> this means it is up to us to detect when unused params/vars occur.

Not so. There are four possible values for JSHint's [`unused`][1] option: `true`, `false`, `"strict"`, and `"vars"`. The last of these options disallows unused variables but allows unused parameters. Does this change your stance, @buzzdecafe? I littered Sanctuary's source code with `// jshint ignore:line` comments because I was unwilling to give up unreferenced variable detection. Yesterday I discovered I can have my :cake: and eat it too. :)


[1]: http://jshint.com/docs/options/#unused

buzzdecafe:
yes, i did see that. i agree that is much better. i still think in a lib like this that is so concerned with arity, it's nice to have the security blanket.

CrossEye:
mu

davidchambers:
What does "mu" mean, @CrossEye?

buzzdecafe:
https://en.wikipedia.org/wiki/Mu_%28negative%29

Issue #1073, opened 2015-05-08
Deep merge


ivan-kleshnin:
```js
import {merge} from "ramda";

console.log(
  merge(
    {age: 30, tags: ["a"] },
    {tags: ["b"] }
  )
); // {age: 30, tags: ["b"]} -- ok

console.log(
  merge(
    {model: {age: 30, tags: ["a"] }},
    {model: {tags: ["b"] }}
  )
); // {model: { tags: ["b"] }} -- ??? where is age ???
```

So now merge behaves like `Object.assign` and does not merge nested objects.

```js
console.log(
  Object.assign(
    {},
    {model: {age: 30, tags: ["a"] }},
    {model: {tags: ["b"] }}
  )
); // {model: { tags: ["b"] }} -- this is still lame but expectable from Object.assign
```

The note from docs 

> This function will not mutate passed-in objects

is not clear. Do we talk about being immutable or do we talk about mimicking `Object.assign` behavior?

If it's expected behavior I'd recommend to reconsider it. 
To my experience, the business cases where merges of nested objects are required are
much more common. And, even more important, you may emulate "shallow merge" behavior with "deep merge" but not vice versa.

```js 
// emulate SHALLOW merge for specific field
result = assoc("model", data2.model, merge(data1, data2));
```

buzzdecafe:
just because you want `mergeDeep` does not mean "`merge` is broken". 

For the short-term, I recommend we explicitly document that `merge` is shallow, not recursive. 

For the longer term, we should decide if we want to tackle this problem. I think it is thornier than it may appear on the surface. Have to handle cases for constructed objects, undefined properties, cycle detection, clobbering with different types. 

Who wants to take a whack at it? (not it!)

CrossEye:
Not it!

CrossEye:
There are several requests that have come up multiple times.  Some, like `foldObj` are shot down for solid API reasons.  Some, like a deep `clone`, were delayed because of implementation difficulties. This one is a little trickier.  @paldepind has mostly convinced me that there are no technical issues with `mergeDeep` that don't also plague our current shallow `merge`.  And yet the uses and limitations of our current `merge` seem clear in my mind, but I can't figure them out for a deep `merge`.

What happens with cyclic references?  Copying the references at the root doesn't seem to be an issue, but how about if we want to walk the tree?  

What happens with constructed object properties and their prototypal members?  Do we duplicate all prototypal values in the merged object?  Ignore them?

[...]

Issue #1038, opened 2015-04-21
Mention "constant" in the docs


bergus:
I was searching for a function in the http://ramdajs.com/docs/ that created a *constant* function, i.e constantly returns the given value regardless of its input. In Haskell, `const` does this.

Only by conincidence I noticed that such function already exists, only going by the name `R.always`. Can we add a synonym, rename, or at least mention the term "constant" in the docs so that it can be found via `Ctrl`+`F`?

TheLudd:
Just a note here. It has been discussed on gitter that `always` is not *truly* a constant function depending on how it is viewed. Example:

```
var arr = [0];
var alwaysArrayWithZero = R.always(arr);
arr[0] = 1;
alwaysArrayWithZero() // => [1]
```

It could be argued that a `constant` function should guard against this.

buzzdecafe:
sounds like another issue @TheLudd !

in the meantime, we should certainly add an `@aka` or some such to the docs as suggested here: #792

CrossEye:
@bergus: We've moved entirely away from aliases for reasons described best by [this parable][pa].

We really would like to find a good way to add a list of AKAs from other libraries and other languages to the docs.  `constant` and `const` would be good such tags for this one.


@TheLudd: I don't find that too worrying, I guess.  We could, I suppose, do a deep clone of the object passed to `always`, once received and before returning, but that might break other people's expectations.  I think that reference identity here is probably enough.  Of course, I've mostly used it for primitive values, so it doesn't affect me either way.  Has the current behavior caused problems for you?



  [pa]: https://github.com/ramda/ramda/pull/574#issuecomment-64124850

TheLudd:
@CrossEye No It hasn't. I think mainly because I don't mutate data since I use ramda to modify objects and arrays. I have no issue with `always` remaining the way it is. I am just not sure if I consider it quite the same thing as `constant`.

buzzdecafe:
for the purposes of searching the docs, i'd say they are close enough

sebinsua:
Agreed that it would be nice if methods were tagged with other terms that people might search for when looking for them.

The aim being not that the method is aliased itself, but that it is easier to find without knowing its name.

davidchambers:
We needn't do anything fancy here. We could just update the description to contain the word "constant". :)

buzzdecafe:
that's certainly easier than a custom tag :+1:

CrossEye:
[...]

PR #1088, opened 2015-05-15
Add `deepMerge`/`set`/`extend`/`mixin`

paldepind:
This pull request adds a function that recursively combines two supplied objects into one. This has previously been discussed at #1073. This PR uses the name `set` as a non-final placeholder.

This implementation follows the rules I outlined in the above mentioned thread with a few changes based on the discussion. These rules leads to a very simple implementation and behavior that is IMO easy to understand:

1/ The function recurses into ~~arrays and~~ non-atomic objects. Non-atomic objects are objects that have keys in the eyes of `Object.keys` (i.e. they have own enumerable properties). It will for instance recurse into an object literal but not into a date object.
2/ Values from objects are always copied into a new plain object.
3/ ~~Value from arrays are copied into a new array. Sparse arrays are not handled.~~ Arrays are always copied by reference.

__Use cases__

As I've [mentioned](https://gitter.im/ramda/ramda?at=555229f11817239c37e4e6c4) I think the most interesting way to think about such a function is as [a way to set properties](https://gitter.im/ramda/ramda?at=5552359df853e7f14c2b84c5) on objects.

This way of thinking has changed my opinions on the implementation in two ways:

1. The overriding object should be first instead of last to be consistent with `assoc` and `assocPath`. `R.assoc('c', 3, {a: 1, b: 2})` matches `R.set({c: 3}, {a: 1, b: 2})`.
2. I now think of the returned object as "`b` with these changes from `a`" instead of "a brand new thing made up of `a` and `b`." With the first view I'd expect as little deep copying as possible but with the second view I'd expect everything to be deep copied. This PR implements the former.

I see this function as a more powerful and useful version of `assoc` and `deepAssoc`.

```javascript
// Compare this
var o = {a: {b: 1, c: 2, d: 3}};
var o2 = R.set({a: {b: 4, c: 5}}, 0);
// To this
var o3 = R.compose(
  R.assocPath(['a', 'b'], 4)
  R.assocPath(['a', 'c'], 5)
)(o);
```

I think this is a very welcome addition to Ramda. Currently Ramda has a lot of great functions for working with arrays without mutating them but it feels a bit weaker with regards to objects.

__About arrays__

@ivan-kleshnin had some [great reasons](https://github.com/ramda/ramda/issues/1073#issuecomment-100600208) why we should not (unlike Lodash) attempt to merge arrays.

While being able to do something like this is neat:

```javascript
var users = {
  'data': [{ 'user': 'barney' }, { 'user': 'fred' }]
};
var ages = {
  'data': [{ 'age': 36 }, { 'age': 40 }]
};
_.merge(users, ages);
// → { 'data': [{ 'user': 'barney', 'age': 36 }, { 'user': 'fred', 'age': 40 }] }
```

But the same behavior can also be a disaster:

```javascript
var users = {
  'data': [{ 'user': 'barney', 'age': 36 }, { 'user': 'fred', 'age': 40 }]
};
var anotherUser = {
  'data': [{ 'user': 'kim', 'age': 42 }] // new user, not an update
};
_.merge(users, anotherUser);
// → { 'data': [{ 'user': 'kim', 'age': 42 }, { 'user': 'fred', 'age': 40 }] }
// Barney disappeared :(
```

__About naming__

I really don't know what to call this thing. In my head the name `merge` implies merging two conceptually equal objects. Not applying a few changes to an existing object. On the other hand @joneshf seemed to have been thinking about `merge` like this all along. I really don't know.

But I definitely think that having both this function _and_ `merge` would be too much. I choose `set` as the name in this PR just because I had to call it something (and `set` might not be a bad name).

__Symmetry with `R.evolve`__

When/if this gets merged I think `R.evolve` should be made recursive in the same way as well. I think these two functions together provides a very solid experience for updating object without mutations.

Handling the before mentioned Lodash example is very easy and, importantly, free from bad surprises (the example below can be implemented with todays `evolve` and `merge` but this power does not extend to nested objects).

```javascript
var users = {
  'data': [{ 'user': 'barney' }, { 'user': 'fred' }]
};
var ages = {
  'data': [{ 'age': 36 }, { 'age': 40 }]
};
R.evolve({data: R.zipWith(R.set, ages.data)}, users);
// { 'data': [{ 'user': 'barney', 'age': 36 }, { 'user': 'fred', 'age': 40 }] }

var anotherUser = {
  'data': [{ 'user': 'kim', 'age': 42 }] // new user, not an update
};
R.evolve({data: R.concat(anotherUser.data)}, users);
// { 'data': [{ 'user': 'barney', 'age': 36 }, { 'user': 'fred', 'age': 40 }, { 'user': 'kim', 'age': 42 }] }
// Barney did not disappear :)
```

__Cycles__

This implementation does not detect cycles. I agrees with @ivan-kleshnin's [opinion on this](https://github.com/ramda/ramda/issues/1073#issuecomment-100601526). Also as far as I can think the only way to create data with cycles it with mutations (I could be wrong)? So not supporting them should be a minor issue. If there was a very simple way to detect cycles that would not be too inefficient I think adding it would be great though. But I can't think of one.

Consider this PR to be nothing but my current suggestion for an implementation. Everything is up for discussion :)

### DIFF
--- a/src/set.js
+++ b/src/set.js
@@ -1,0 +1,47 @@
+var _curry2 = require('./internal/_curry2');
+var _extend = require('./internal/_extend');
+var _has = require('./internal/_has');
+var _isArray = require('./internal/_isArray');
+var isEmpty = require('./isEmpty');
+var keys = require('./keys');
+
+
+/**
+ * Creates a new object with the values from `b` merged with the
+ * values in `a`. Only own enumerable properties are considered. This function
+ * recursively handles nested objects, but not arrays, with own enumerable
+ * properties. All non-primitive properties are copied by reference if
+ * possible. This function * will *not* mutate passed-in objects.
+ *
+ * @func
+ * @memberOf R
+ * @category Object
+ * @sig {k: v} -> {k: v} -> {k: v}
+ * @param {Object} a object with higher precedence in output
+ * @param {Object} b source object
+ * @return {Object} The merged object.
+ * @example
+ *
+ *      R.set({age: 40, {name: 'fred', age: 10});
+ *      //=> {'name': 'fred', 'age': 40}
+ *
+ *      R.set({parent: {age: 40}}, {name: 'fred', age: 10, parent: {age: 39}});
+ *      //=> {'name': 'fred', 'age': 10, {parent: {age: 40}}}
+ */
+module.exports = _curry2(function set(update, original) {
+  if (_isArray(update) || _isArray(original)) {
+    return update;
+  }
+  var updateKeys = keys(update);
+  var originalKeys = keys(original);
+  if (isEmpty(updateKeys) || isEmpty(originalKeys)) {
+    return update;
+  }
+  var result = _extend({}, original);
+  var key, idx = -1, length = updateKeys.length;
+  while (++idx < length) {
+    key = updateKeys[idx];
+    result[key] = _has(key, original) ? set(update[key], original[key]) : update[key];
+  }
+  return result;
+});
--- a/test/set.js
+++ b/test/set.js
@@ -1,0 +1,38 @@
+var assert = require('assert');
+
+var R = require('..');
+
+
+describe('set', function() {
+  it('handles flat object', function() {
+    var a = {w: 1, x: 2, y: 3, z: 4};
+    var b = R.set({y: 5}, a);
+    assert.deepEqual(b, {w: 1, x: 2, y: 5, z: 4});
+  });
+
+  it('add new properties to original object', function() {
+    var a = {w: 1, x: 2, y: 3};
+    var b = R.set({z: 4}, a);
+    assert.deepEqual(b, {w: 1, x: 2, y: 3, z: 4});
+  });
+
+  it('handles nested objects', function() {
+    var a = {w: 1, x: {a: 1, b: 2, c: 3}, z: 4};
+    var b = R.set({x: {a: 5}}, a);
+    assert.deepEqual(b, {w: 1, x: {a: 5, b: 2, c: 3}, z: 4});
+  });
+
+  it('swallow copies when possible', function() {
+    var a = {w: 1, x: {a: 1, b: 2, c: 3}, z: 4};
+    var b = R.set({z: 5}, a);
+    assert.strictEqual(a.x, b.x);
+    var c = R.set({x: {a: 5}}, a);
+    assert.notStrictEqual(a.x, c.x);
+  });
+
+  it('overrides arrays', function() {
+    var a = {w: 1, x: [1, 2, 3], z: 4};
+    var b = R.set({x: [4]}, a);
+    assert.deepEqual(b.x, [4]);
+  });
+});


ivan-kleshnin:
> Was proposed: Value from arrays are copied into a new array.
> Now proposed: Arrays are always copied by reference.

Not that I'm disagree (defensive copying are of little help anyway), but I think this need explanation. 
If I missed the place it was already discussed, please point.

paldepind:
@ivan-kleshnin I tried to answer that question in the numbered points following "This way of thinking has changed my opinions on the implementation in two ways". Also see the linked Gitter messages just above that sentence. If that does not properly explain my reasoning let me know. The gist of it is that this function is applying changes to a pretended immutable data structure and thus as many shared parts of the original data structure should be preserved.

ivan-kleshnin:
> The gist of it is that this function is applying changes to a pretended immutable data structure and thus as many shared parts of the original data structure should be preserved.

Thank you. I agree here. My own view on this is next. There are two possible strategies to overcome mutability in JS without involving 3-rd party libs or enforcing immutability with `freeze()`. 

First: defensive copying. Second: just never call mutating functions. Defensive copying or cloning failed very often for me. You can't just `clone` everything and expect it to work because there may be:
1) Circular references. By passing original objects where it's possible you avoid this bad case. 
2) Some keys can link to very big objects in memory. Like `component.props` in React (never try to clone them).

So the only requirement I put for Ramda is to never ever mutate input arguments.

buzzdecafe:
> So the only requirement I put for Ramda is to never ever mutate input arguments.

100% agree with that.

CrossEye:
I definitely like this idea.  I might nitpick about the implementation a little and I agree that we should think about the name some more, but I think this is definitely worth doing.

I don't know if some variant on `assoc` would work for a name, perhaps '`assocObj`'?  That's a mouthful, and maybe not as good as '`set`'.  We can bikeshed, though, after its clear that others are interested too.  @buzzdecafe, @davidchambers, what say you?

src/set.js
```diff
@@ -0,0 +1,27 @@
...
+ * @return {Object} The merged object.
+ * @example
+ *
+ *      R.set({age: 40, {name: 'fred', age: 10});
+ *      //=> {'name': 'fred', 'age': 40}
+ *
+ *      R.set({parent: {age: 40}}, {name: 'fred', age: 10, parent: {age: 39}});
+ *      //=> {'name': 'fred', 'age': 10, {parent: {age: 40}}}
+ */
+module.exports = _curry2(_set);
```
CrossEye: I don't see any reason for the internal/external distinction here.  We need to do that if there are internal consumers of the uncurried/undispatched version.  (Perhaps.  Even then, I would often prefer usually to dogfood our stuff.  But here the only use of the internal function is the recursive use.  That will be captured by the named function expression.

src/internal/_set.js
```diff
@@ -0,0 +1,23 @@
+var _extend = require('./_extend');
+var _has = require('./_has');
+var _isArray = require('./_isArray');
+var isEmpty = require('../isEmpty');
+var keys = require('../keys');
+
+module.exports = function _set(update, original) {
+  if (_isArray(original) || _isArray(update)) {
```
CrossEye: Nit: I would prefer

```js
  if (_isArray(update) || _isArray(original))
```

just in order to keep a parallel with the parameters.

src/internal/_set.js
```diff
@@ -0,0 +1,23 @@
...
+    return update;
+  }
+  var originalKey = keys(original);
+  var updateKeys = keys(update);
+  if (isEmpty(originalKey) || isEmpty(updateKeys)) {
```
CrossEye: ditto here

src/internal/_set.js
```diff
@@ -0,0 +1,23 @@
...
+    return update;
+  }
+  var result = _extend({}, original);
```
CrossEye: I don't know if there are any ways to make this more efficient by folding this into the iteration below.  It might not be possible, but it might be worth investigating.

src/set.js
paldepind: I noticed that `assocPath` was implemented in this way even though the internal `_assocPath` isn't used by anyone but `assocPath`. So I assumed you wanted this for recursive functions? But I don't think it makes much sense either.

src/internal/_set.js
paldepind: Yes! Good point.

src/internal/_set.js
paldepind: What inefficiency? Do you mean avoid a recursive function calls by doing the checks before recursing instead of after?

davidchambers:
I'm very much in favour of this proposal. The pull request description is exceptional, @paldepind.

src/set.js
CrossEye: No.   I think we should fix that one too.

buzzdecafe:
i think we should detect cycles. that is how `clone` works today, this should behave in a similar fashion. and in js circular objects are everywhere.

src/internal/_set.js
CrossEye: Just noting that `extend` should be doing much the same job as this function, less the recursive bit.   I didn't know if there would be a useful way to combine the looping inside `extend` with what's going on here.   But you might be able to chalk that up to random thoughts on a bus-ride.

CrossEye:
Thinking about dealing with cycles makes my head hurt:

```js
    var x = {a: 1, b: 2};
    x.c = x;
    var y = set({c: {b: 20}, x);
```

Now what should `y.b` be?  Since `x.c` is a reference to `x` and we're updating a property of `x.c`, we should be also update `x`, so `y.b` should be 20, right?  On the other hand, we're specifically updating only a nested property of `x`, the `c` property, and so we shouldn't be changing other ones, which means that `b` should remain as `2`, right?

I don't see any reason to privilege one of these points of view.

In lodash, they've made the call that it's still `2`, although the recursively nested objects all have `20`.  And maybe there's a logical reason for this choice that I'm not seeing.  Obviously just the native `x.c.b = 20` would do the opposite.  Either way simply feels arbitrary to me.

src/internal/_set.js
paldepind: The only thing I can think of if to create an `extend` that takes a callback that handles the copying or recursing.

paldepind:
Thank you @davidchambers.

This function somewhat handles cycles as long as they don't appear in the setting object.

This is what it does in your example @CrossEye:

```javascript
> var x = {a: 1, b: 2};
undefined
> x.c = x;
{ a: 1, b: 2, c: [Circular] }
> var y = R.set({c: {b: 20}}, x)
undefined
> y
{ a: 1,
  b: 2,
  c: { a: 1, b: 20, c: { a: 1, b: 2, c: [Circular] } } }
```

So `x.b` is still `2`, `x.c.b` is 20 and `x.c.c.b` is still `2`. This can both make a lot of sense or make no sense depending on how you look at it :weary: But to implement the behavior you get when you mutate `x` the function would have to do a lot of extra work. It wouldn't no longer be enough to only look at the setter object (my guess is that lodash is the way it is for the same reason).

@buzzdecafe the shared reference detection thing in `clone` is currently O(n^2) where `n` is the number of references in the cloned object. That's not very fast but I don't think it can be implemented faster (at least not without native support for the new Map). I think doing that is worth it for `clone` but not for this function.

I still think cyclic data is bad practice just like @ivan-kleshnin. What does this even mean:

```javascript
var obj  = {a: 1, b: 2, c: {a: 1, b: 2}};
var setter = {a: 2};
setter.c = setter; // again – we have to mutate to create the cycle
var newObj = R.set(setter, obj);
// What on earth is `newObj` now?
```

For the above code a stack overflow might be an adequate wake up call ;)

buzzdecafe:
ok, i am not willing to hold up progress on this. let's give it a go.

CrossEye:
Fixed by #1088

paldepind:
Nice :+1: 

Did you notice this:

> But I definitely think that having both this function and `merge` would be too much.

I think `merge` should be deprecated. Or `set` should be renamed to `merge` to override it. Otherwise it will be another confusing API instance of "I can't tell the difference between `x` and `y`. Why would I ever want to use `x` instead of `y`?"

davidchambers:
> I think `merge` should be deprecated. Or `set` should be renamed to merge to override it.

I absolutely agree that we should do one or other of these things.

The problem with renaming this function is that `R.merge({x: 1}, {x: 2})` currently gives `{x: 2}`, but after renaming `R.set` it would give `{x: 1}`. From an ease of upgrading perspective, keeping `R.set` and deprecating `R.merge` is preferable.

Setting aside the fact that we already have a function named "merge", which name is better? I'm not sure, but unless there's a compelling reason to favour "merge" we should stick with "set" for the reason given above.

paldepind:
@davidchambers The only advantage I saw for the name `merge` was backwards compatibility. But as you pointed out that would only subtly break things instead of loudly breake things. I had not thought about that. I like `set` since it's short.

davidchambers:
@paldepind, could you open a pull request to add the `@deprecated` tag to the docstring for `R.merge`?

CrossEye:
Agreed.  Let's deprecate `merge` in favor of `set`.

I probably won't have much time to work on anything before the weekend, but as I [seem to have volunteered][co] to add some sort of aka mechanism to the docs, I will try to do that then.  And `set` can be listed with `merge`, `extend`, `mixin`, and perhaps others.

  [co]: https://github.com/ramda/ramda/issues/1038#issuecomment-102142713

nthtran:
`set` doesn't handle falsy values (e.g. `null`) and `{}` the same way `merge` does. For example:

```js
merge({ a: 1 }, null) // { a: 1 }
set(null, { a: 1 }) // null

merge({ a: 1 }, {}) // { a: 1 }
set({}, { a: 1 }) // {}
```

This is unexpected behaviour IMO. `set` should work the same way as `merge` in these cases.

paldepind:
`merge` is deprecated so being consistent with it is not a priority.

But `set({}, { a: 1 }) // {}` is certainly a bug.

ivan-kleshnin:
@nthtran  agree

```set(null, {a: 1}) // null``` -- loses type of original value.

```set({}, {a: 1}) // {}``` -- loses keys 

Both are "side-effects" (nothing about **setting** values).

**Comment by paldepind** [comment]
Should be fixed in #1170.

**Comment by ivan-kleshnin** [comment]
**As a remark.** I think it's not a good practice to use reserved words as function names.
I always import Ramda functions to a global namespace because I think of them as JS polyfills.

```js
import {set} from "ramda";
```

Which "works" but will be highlighted as a reserved word or even a syntax error, depending on your editor. I dunno if this can be considered at this history stage, though.

---

**Considering the topic.**
I tend to think we need both behaviors with clearly defined semanthics.

```
set -> merge // "merging" with a respect to keys
++ assign // "merging" without a respect to keys
assoc -> set 
prop -> get 
```

### Expectations

#### Merge

`merge` merges objects as dicts, keys can't be lost in a result

```js
merge(null, {x: "X"}); // {x: "X"}
merge({}, {x: "X"}); // {x: "X"} 
merge({tags: []}, {tags: ["foo", "bar"]}); // {tags: []}
merge({models: {}}, {models: {...}}); // {models: {...}} 
```

#### Assign

`assign` merges objects as dicts, keys can be lost in a result

```js
assign(null, {x: "X"}); // {x: "X"}
assign({x: "Y"}, {x: "X"}); // {x: "Y"}
assign({}, {x: "X"}); // {} -- different from *merge* and *Object.assign*
assign({tags: []}, {tags: ["foo", "bar"]}); // {tags: []}
assign({models: {}}, {models: {...}}); // {models: {}} -- different from *merge*
```

### This is worthless 

If `f(x, y) == x` it's just `takeFirstArg` behavior...

```js
replace(null, {x: "X"}); // null
replace({}, {x: "X"}); // {}
```

buzzdecafe:
even `R.set` by itself would break ES3 compat, which so far we have been able to maintain without too much difficulty. I think the proposed naming is sensible, and I would like to see `assign` added.

CrossEye:
> I always import Ramda functions to a global namespace because I think of them as JS polyfills.
> 
> ```js
> import {set} from "ramda";
> ```
> 
> Which "works" but will be highlighted as a reserved word or even a syntax error, depending on your  editor. I dunno if this can be considered at this history stage, though.

Why would this be tagged?  `set` is not a reseved word in JS, at least not in all lowercase, right?

ivan-kleshnin:
> Why would this be tagged? set is not a reseved word in JS, at least not in all lowercase, right?

Both are reserved. There is a `new Set()` (constructor) and there is a `set [expression]` syntax.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get 
WebStorm IDE which I use highlights lowercased `set` as a reserved word.

davidchambers:
That's one of the bad parts of JavaScript hindering our use of one of the good parts! >.<

CrossEye:
Ah, you know I guess I'd seen `set` be tagged by my IDE (I use WebStorm at work) as well, and simply ignored it.  It's not actually a syntax error to use it, only a highlighting error in such tools, but it should at least give us pause.

texastoland:
@davidchambers You changed nixed @paldepind's `set` in d815f826010118a87c30add1f0abe30044334948 in favor of a lensy implementation in 2d0b02701c138ea2c8a31c77e9f927f2923e81a9. Am I missing a way to achieve `deepMerge` or should I just recurse `mergeWith` like Immutable implements it?

davidchambers:
> Am I missing a way to achieve `deepMerge` […]?

You're not missing anything. Ramda doesn't currently provide such a function.

texastoland:
Can you make this prettier?
```js
R.deepMerge = R.mergeWith((l, r) =>
  l.concat && r.concat
    ? R.concat(l, r)
  : !(R.is(Object, l) && R.is(Object, r))
    ? r
  : R.deepMerge(l, r)
)

// => {"bar": {"bizz": [2, 3, 5, 6], "buzz": 7}, "foo": 4}
R.deepMerge(
  {foo: 1, bar: {bizz: [2, 3]}},
  {foo: 4, bar: {bizz: [5, 6], buzz: 7}}
)
```

CrossEye:
While I would definitely like adding a deep merge function, and I love the simplicity of this one, I don't think I like the API of this one:

```js
mergeConcat(
  {foo: 1, bar: {bizz: "abc"}},
  {foo: 4, bar: {bizz: [5, 6], buzz: 7}}
); //=> {"bar": {"bizz": "abc5,6", "buzz": 7}, "foo": 4}

mergeConcat(
  {foo: 1, bar: {bizz: "abc"}},
  {foo: 4, bar: {bizz: "xyz", buzz: 7}}
); //=> {"bar": {"bizz": "abcxyz", "buzz": 7}, "foo": 4}

mergeConcat(
  {foo: 1, bar: {bizz: [2, 3]}},
  {foo: 4, bar: {bizz: "xyz", buzz: 7}}
); //=> {"bar": {"bizz": [2, 3, "xyz"], "buzz": 7}, "foo": 4}
```

Issues like this were why we eventually decided not to try to merge lists, only to replace them.

texastoland:
Hmm thanks for catching a bug in my code. I guess it should just be Arrays. FWIW I used Immutable and lodash-fp previously and each had issues requesting the same basic thing. Maybe not a solved problem.

CrossEye:
I'm not sure there is a solution to that problem.  It seems to me that what we really want is to merge plain objects by (recursively) replacing overlapping keys, but not to try any further trickiness with lists, strings or any other type.

texastoland:
> but not to try any further trickiness with lists, strings or any other type.

It may not be the domain of Ramda but the problem is users have to reimplement it every time regardless e.g. facebook/immutable-js#406. My specific use case is merging user options objects e.g. [webpack-config](https://github.com/mdreizin/webpack-config/blob/6adc77b911b882debc375f480fa0a7694fc6910f/src/Config.js#L116-L119). I believe there is some solution. The problem with testing `concat` is it basically checks whether both arguments are semigroups which is irrelevant. Immutable got closer with `isIndexed` except it necessitates a separate type hierarchy. The problem reminds me of Snoyberg's nominal classes. I want a flag other than `concat` to represent a type can be combined with itself.

paldepind:
> It seems to me that what we really want is to merge plain objects by (recursively) replacing overlapping keys, but not to try any further trickiness with lists, strings or any other type.

Agreed. That is what my deep merge implementation does.

texastoland:
To quote @leebyron:

> merging by index isn't really useful behavior anyway

I'm not saying Ramda is doing it wrong. I'm saying a common use case is being left to users to reinvent. And if it's trick for you it's trickier for us.

CrossEye:
But do we create separate functions to handle this case:

```js
mergeDeepConcat(
  {todos: {list: [1, 2, 3]}}, 
  {todos: {list: [3, 4, 5]}}
); //=> {todos: {list: [1, 2, 3, 4, 5, 6]}}
```

and this one:

```js
mergeDeepPlain(
  {color: 'red', sides: ['top']},
  {sides: ['right', 'left']}
); //=> {color: 'red', sides: ['right', 'left']}
```

Because we clearly can't make that call for the user.  Why the assumption that the user wants to replace strings, booleans, and numbers, but wants to concat lists?  Why not concat strings as well?

CrossEye:
@paldepind: Any interest in reviving this?  I think we really could use some version of a deep merge.

texastoland:
> But do we create separate functions to handle this case

Perhaps implementing each in terms of `mergeDeepWith` for custom logic? Immutable does [something similar](https://github.com/facebook/immutable-js/blob/137a9ccfff6fcc72e56d21b11d82e987e6754eaf/src/Map.js#L135-L137).

> Why the assumption that the user wants to replace strings, booleans, and numbers, but wants to concat lists?

The use common case I've seen in several build systems is merging user options. The motive behind `concat` is not combining arbitrary same typed things (which is what [semigroups](https://github.com/fantasyland/fantasy-land#semigroup) do) but arbitrary same typed containers of things (e.g. an immutable set). Strings are used as primitives rather than containers of characters. By the same intuition I wouldn't expect numbers to be added or multiplied together (both valid semigroups).

paldepind:
> @paldepind: Any interest in reviving this? I think we really could use some version of a deep merge.

Yes. I am going on a small vacation tomorrow. But when I get home I could re-submit the original PR>

> The motive behind using concat is not combining arbitrary same typed things (which is what semigroups do) but arbitrary same typed containers of things (e.g. an immutable set). Strings are universally used as primitive values rather than as queues of characters. By the same intuition I wouldn't expect numbers to be added or multiplied together (both valid semigroups). I'm updating my example to account for strings and same types.

If I understand you correctly you think a string is different from something like a `set` because a set is more of a general container than a string is? I can see your point but I do think the notion is rather blurry. We can also consider a string to be a list of characters and then it should be treated like any list. In some cases you could argue that you'd expect numbers to be added together. For instance like this:

```js
const account1 = {amount: 100};
const account2 = {amount: 200}:
const combinerAccount = merge(account1, account2); // I want this to be {amount: 300}
```

texastoland:
> We can also consider a string to be a list of characters and then it should be treated like any list.

I've never seen them used that way.

> In some cases you could argue that you'd expect numbers to be added together.

Not for the common use case implemented in a number of config libraries.

paldepind:
> I don't see how.

You're making a distinction about different kinds of things even though they are all semigroups (as you mention). Your want sets to be merged but not strings to be merged even though they are both equally capable of being merged. What is the precise distinction between the two?

> I've never seen them used that way.

I don't know what you mean. A list of characters is isomorphic to a string. JavaScript allows strings to be indexed exactly as if they where arrays which is based exactly on that perspective on strings. You can take the first character of a string, you can concat two strings together, etc. There are many similarities with lists.

texastoland:
Because I'm addressing a specific common use case and not making a mathematical argument.

paldepind:
> Because I'm addressing a specific common use case and not making a mathematical argument.

I'm not requesting "mathematical arguments". I am simply pointing out that to me your distinction seems arbitrary. I do not know what is the most common case. But in my use cases your `mergeConcat` is _less_ useful because it makes it impossible to overwrite an array with another. And you may think that merging by index isn't useful but Lodash and it's users apparently think otherwise.

I simply want a non-mutating `Object.assign` that recurses into objects but not arrays. `Object.assign` is very often used in JavaScript and I believe a simple recursive version of it would be suprise-free and apply in many use cases.

texastoland:
> I do not know what is the most common case.

Merging user options in build libraries.

> But in my use cases your `mergeConcat` is less useful because it makes it impossible to overwrite an array with another.

Not asking exclusively.

> And you may think that merging by index isn't useful but Lodash and it's users apparently think otherwise.

If that were true the only `mergeWith` example in the docs wouldn't show to concat arrays.

> that recurses into objects but not arrays.

Which sounds more arbitrary since they refer to specific types rather than behaviors of types.

texastoland:
Perhaps just a `mergeDeepWith` function that handles recursion like lodash and Immutable instead of divergent array behavior?

paldepind:
> If that were true the only `mergeWith` example in his docs wouldn't show how to concat arrays. I suspect it's implemented that way for simplicity.

Take a look at the documentation for `_.merge`. It has an example for index based merging. Btw, I agree with you that index merging is weird behavior but that apparently is not a census.

> Which also sounds arbitrary since they refer to specific types rather than behaviors of types.

It is not ideal. But I think it is less arbitrary. The description is simply that it recurses into objects. That's it. It handles a single type. The alternative recursively merges objects and arrays but not strings, numbers or other custom things (like Immutable.js datstructures). That sounds more arbitrary because one specifically has to name the subset of merge-able things that are merged. I agree with you that it has a valid use case but I think it is a behavior that is more likely to confuse users.

texastoland:
> The alternative recursively merges objects and arrays but not strings, numbers or other custom things (like Immutable.js datstructures).

I'll adapt [my implementation](https://github.com/ramda/ramda/pull/1088#issuecomment-231102816) to work with Immutable then since I'm just checking methods 😆

CrossEye:
>> What is the precise distinction between the two?

> Whether they're used as containers.

I find that pretty arbitrary for a general-purpose function.  When you merge, I have no idea if you expect a given container to represent **the** collection of values at some node or some subset of them that needs to be combined with another subset.

But I do see a reason to offer both, and as @texastoland suggests, they might build on the same implementation.

kurtmilam:
For my purposes, primarily merging POJOs that would withstand a `JSON.stringify` -> `JSON.parse` round trip, a stack overflow is a perfectly acceptable cycle check.

It's not really important to me whether arrays are treated as ordered lists, where a value at a given index in one array would be overwritten by a different value at the same index in a second array, or as collections that are simply concatenated on merging. I can see logic and usefulness in both approaches.

I wrote an underscore deep extend mixin (strictly for my own use, initially) back in early 2012, after jashkenas made it clear that he was against adding such a method to underscore and before lodash added `.merge` (AFAICT from the lodash changelogs). I chose to treat arrays as lists in which order mattered, and favor that approach, but concatenation wouldn't be a deal-breaker for me.

Occasional difficulties balancing between practicality and purity are understandable. Does one simply make an executive decision and add a method, purity be darned, or does one hold one's horses in the absence of an unambiguously pure solution?

Some of the purity questions could likely be pushed to the side with well-documented design choices. A `deepMerge` method that is meant to operate on arguments that are POJOs doesn't need to worry about cycles or objects with properties that shouldn't be in a POJO. The choice on how to treat arrays can also be a design decision that's well-documented. Just make the calls and document them clearly, et voilà, ramda will have a new, useful method that I'm sure a number of developers would appreciate.

CrossEye:
I don't think the questions here are about purity so much as coherence.  No fairly obvious API has leaped to the front of the line.  I'm bothered by arbitrary decisions here.  And this topic seems to have more than its share whenever we visit it.

texastoland:
Based on @paldepind's feedback I think lodash made the right decision in that regard. Basically JD abstracted recursion without dictating any default beyond  `Object.assign`. I initially disliked the `return undefined` semantic but it's effectively `Nothing` and any other value is `Just` but nicer in colloquial JS.

expelledboy:
Just started using Ramda, very exciting to see javascript libraries of this nature coming from an Erlang background!

From my perspective I think  `mergeDeep` should recursively merge plain objects by replacing overlapping keys (as suggested by @CrossEye), allowing further customization with  `mergeDeepWith` on non-object values, or (just for laughs)  `mergeDeepEvolve`:

``` javascript
mergeDeepEvolve(
  { a: { b: [1], c: true }, d: true },
  { a: { b: [2], c: false }, e: true },
  { a: { b: R.concat } }
);
// => { a: { b: [1,2], c: false }, d: true, e: true }
```

Given the comprehensive API already available for arrays I don't see why a basic implementation is being delayed by discussion of how to concat? I have only been using Ramda for about two weeks I have needed `mergeDeep` twice, and recursive merging based on just the value sounds like hell.

FYI If anyone on the interwebs is looking for a basic implementation:

``` javascript
const isObject = maybeObj => Object.prototype.toString.call(maybeObj) === '[object Object]';

function mergeDeep(...vars) {
  return R.ifElse(
    R.all(isObject),
    R.apply(R.mergeWith(mergeDeep)),
    R.nth(1)
  )(vars);
}
```

buzzdecafe:
is `merge` deprecated? I see that [it *was*](https://github.com/ramda/ramda/commit/ec8f5d02ec1154dd8c503b1d60f59f533ae6cd33), but it looks like [the next commit undid that](https://github.com/ramda/ramda/commit/d815f826010118a87c30add1f0abe30044334948)

CrossEye:
@paldepind?  ... Interested?

paldepind:
Yes :smile: The original PR made `merge` deep/recursive. Do we want that or a seperate `mergeDeep`?

expelledboy:
`merge` isn't deep?

CrossEye:
My initial thought would be a separate `mergreDeep`, but I don't have very strong feelings on it.   If we remove the shallow version, I'm sure I could make one easily enough when needed by folding `assoc` over the result of `toPairs`.

scott-christopher:
I've created #1867 to introduce `mergeDeep`, `mergeDeepWith` and `mergeDeepWithKeys`.

PR #1093, opened 2015-05-20
pluck: fix signature

raine:


buzzdecafe:
LGTM. not sure how we got it that wrong. probably because it was done in a wicked rush.

PR #1095, opened 2015-05-20
fix empty docs

buzzdecafe:


src/empty.js
```diff
@@ -3,7 +3,7 @@ var _hasMethod = require('./internal/_hasMethod');
 
 
 /**
- * `empty` wraps any object in an array. This implementation is compatible with the
+ * `empty` returns an empty list for any argument. This implementation is compatible with the
```
davidchambers: I would say _`empty` returns the empty value of its argument's type_, since the empty value needn't be a list.

buzzdecafe: that's definitely better, but not quite right I think. If you call `R.empty("abc")`you get back `[ ]`

buzzdecafe: looking over the docs, we are documenting the list case, and leaving the dispatch case as a delightful easter egg for users to discover.

src/empty.js
```diff
@@ -3,8 +3,8 @@ var _hasMethod = require('./internal/_hasMethod');
 
 
 /**
- * `empty` wraps any object in an array. This implementation is compatible with the
- * Fantasy-land Monoid spec, and will work with types that implement that spec.
+ * `empty` returns an empty list for any argument, except when  the argument satisfies the
+ * Fantasy-land Monoid spec. In that case, this function will invoke `empty` on that Monoid.
```
davidchambers:
```diff
-In that case, this function will invoke `empty` on that Monoid.
+In that case, this function will return the result of invoking `empty` on that Monoid.
```

buzzdecafe: yep. i think that's it.

src/empty.js
```diff
@@ -3,8 +3,9 @@ var _hasMethod = require('./internal/_hasMethod');
 
 
 /**
- * `empty` wraps any object in an array. This implementation is compatible with the
- * Fantasy-land Monoid spec, and will work with types that implement that spec.
+ * `empty` returns an empty list for any argument, except when  the argument satisfies the
```
davidchambers: Nit: You have two spaces between "when" and "the".

buzzdecafe: your copy editing skills continue to astound

davidchambers:
:deciduous_tree:

CrossEye:
Definitely an improvement.  But is it oversold?  `empty` will dispatch to the object if it has any `empty` function, regardless if that object implements `Monoid`, correct?

buzzdecafe:
that's why i used "satisifies" instead of "implements"

PR #1096, opened 2015-05-21
value-based equality

### LINKED ISSUE #1094: equality functions
Currently we have [`R.eq`][1] and [`R.eqDeep`][2]. What we do not have is function to determine whether `Just(42)` and `Just(42)` are equivalent. Presumably `Just.prototype.equals` is defined, allowing `Just(42).equals(Just(42))`. But `Just([1, 2, 3]).equals(Just([1, 2, 3]))` will evaluate false if the method's implementation is similar to the following:

```javascript
Just.prototype.equals = function(x) {
  return x instanceof Just && x.value === this.value;
};
```

To resolve this, an author might use `R.eqDeep` to compare the values:

```javascript
Just.prototype.equals = function(x) {
  return x instanceof Just && R.eqDeep(x.value, this.value);
};
```

But `R.eqDeep` does not dispatch to an object's `equals` method if present, which creates another issue. `Just(Left(42)).equals(Just(Right(42)))`, for example, will evaluate true if `Left(42)` and `Right(42)` both appear as `{value: 42}` in the eyes of `R.eqDeep`.

My proposal is as follows:

1.  We change the name of `R.eq` to `R.identical`. The semantics will not change: `R.identical(0, -0)` will evaluate false, as will `R.identical([], [])`; `R.identical(NaN, NaN)` will evaluate true.

2.  We remove `R.eqDeep`.

3.  We define a new function, `R.eq`, with deep equality semantics which recursively dispatches to the `equals` method, if present, of each objects it encounters.

As a result of these changes the other equality functions (such as [`R.propEq`][3]), which are all defined in terms of `R.eq`, will acquire the new equivalence semantics.

The author of the Maybe type could then update her implementation of `Just.prototype.equals`:

```javascript
Just.prototype.equals = function(x) {
  return x instanceof Just && R.eq(x.value, this.value);
};
```

This will give the correct behaviour:

```javascript
> Just([1, 2, 3]).equals(Just([1, 2, 3]))
true
> Just(Left(42)).equals(Just(Right(42)))
false
```

[1]: http://ramdajs.com/docs/#eq
[2]: http://ramdajs.com/docs/#eqDeep
[3]: http://ramdajs.com/docs/#propEq

davidchambers:
Closes #1094

This pull request:

  - changes the name of `R.eq` to `R.identical`;
  - removes `R.eqDeep`;
  - adds `R.equals`, an equivalence function which offers dynamic dispatch; and
  - replaces internal references to `R.eq` with references to `R.equals`.

All Ramda functions (with the obvious exception of `R.identical`) will use value-based equality. So `R.uniq([[42], [42]])`, for example, will give `[[42]]` rather than `[[42], [42]]`.


buzzdecafe:
yep. :cow2:

test/equals.js
```diff
@@ -0,0 +1,110 @@
...
+  it('dispatches to `equals` method', function() {
+    function Just(x) { this.value = x; }
+    Just.prototype.equals = function(x) {
+      return x instanceof Just && R.equals(x.value, this.value);
+    };
+
+    assert.strictEqual(R.equals(new Just([42]), new Just([42])), true);
+    assert.strictEqual(R.equals(new Just([42]), new Just([43])), false);
+  });
+
```
CrossEye: I think we need some other asserts here:

```js
    assert.strictEqual(R.equals(new Just([42]), {value: 42}), false);

    function Foo(x) { this.value = x; }
      Foo.prototype.equals = function(x) {
      return x instanceof Foo && R.equals(x.value, this.value);
    };
    assert.strictEqual(R.equals(new Just([42]), new Foo([42])), false);
```

I didn't see the code that would make this work.  But I have house-guests and have little time to examine it closely.

buzzdecafe:
> `assert.strictEqual(R.equals(new Just([42]), {value: 42}), false);`

should be: `assert.strictEqual(R.equals(new Just(42), {value: 42}), false);`

src/internal/_toString.js
```diff
@@ -17,15 +17,15 @@ module.exports = function _toString(x, seen) {
...
       return 'null';
     case '[object Number]':
-      return typeof x === 'object' ? 'new Number(' + recur(x.valueOf()) + ')' : 1 / x === -Infinity ? '-0' : x.toString(10);
+      return typeof x === 'object' ? 'new Number(' + _toString(x.valueOf(), []) + ')' : 1 / x === -Infinity ? '-0' : x.toString(10);
     case '[object String]':
-      return typeof x === 'object' ? 'new String(' + recur(x.valueOf()) + ')' : _quote(x);
+      return typeof x === 'object' ? 'new String(' + _toString(x.valueOf(), []) + ')' : _quote(x);
     case '[object Undefined]':
       return 'undefined';
     default:
```
CrossEye: Was this file part of an unrelated change?  I'm not seeing the connection.

Not a big deal, just curiosity.

CrossEye:
> should be: `assert.strictEqual(R.equals(new Just(42), {value: 42}), false);`

Absolutely.  Good catch.

paldepind:
I like this. Defaulting to value equality seems very sensible for a functional programming library. @ivan-kleshnin should be happy about this as well.

I'm not a fan of the naming though.

* Not using `R.eq` at all seems like a waste to me. It's a short name and yet totally easy to understand.
* The difference between `R.equals` of `R.identical` is unclear. Try asking ten 10 random JS programmers what the difference is between `R.equals` and `R.identical`?

My current proposal:

* Use `R.eq` instead of `R.equals` because it's a short name and having a short name for an often used function is very convenient.
* Use `R.strictEq` instead of `R.identical` because it's slightly shorter and because it's meaning will be obvious to most JS programmers (as you know Node for instance uses `strictEqual` in `assert`).

These names would also be more consistent with all the existing functions having `eq` in their name.

My guess is that most JS programmers would initially think that the difference between these two functions is `==` vs `===`. But with my suggestion they would at least be right about one of the functions. With the current naming in this PR is think many would map `identical` to `==` and `R.equals` to `===` because _identical_ can be though of meaning "having identical value" (in the sense that "12" and 12 has the same value).

I understand that this might make upgrading a bit trickier. But going for the best API long term is better than a short term gain.

ivan-kleshnin:
> I like this. Defaulting to value equality seems very sensible for a functional programming library. @ivan-kleshnin should be happy about this as well.

I am. I would go even further and drop `identical` completely. 

If 
> All Ramda functions (with the obvious exception of R.identical) will use value-based equality. So R.uniq([[42], [42]]), for example, will give [[42]] rather than [[42], [42]].

Then keeping `identical` because user *may need* violates the YAGNI principle.
I never need half-baked operators. Even if I do I would rather fallback to native JS than remember all the differences between similar Ramda functions with similar names. Just give us the best behavior we can ask for – JS already has the *other* part. Then no arguing about name, just use the shortest possible: `clone`, `merge`, `eq`, `gt`. I am assured noone will ever ask you for their weaker counterparts.

There comes bonus – by dropping *shallow twins* and by relying on value semanthics everywhere you make the library cleaner, lighter and just plain obvious.

The one warning sign of caution for me is the common phrase pattern:
> Never use metaclasses in app code, it's for library code
> Never use DSL in app code, it's for library code
> Never use macros...
> Never use state...
> Never use dependency injection...
> ...
> Never use *foo* in ..., it's for ...

In my experience, every time I heard this... it turned out wrong.
**Both libs and apps should be assembled from the same primitives.**
If you need something to build a big lib – it will be required to build a big app as well.
If you don't need something to build a big lib – it will probably be worthless for any app.

This very statement is applicable even here despite this lib has a simplest structure, being a collection of helpers. 

**Whenever you use `eq` instead of `deepEq` you create implicit, hidden and unspoken rule: "this value should never be replaced with a nested object". Are you ready for such time-bombs in your code?**

paldepind:
> I am. I would go even further and drop identical completely. 

If one operates on data in a purely functional way (i.e. uses immutable data structures or uses mutable data structures without ever mutating them) then comparing values by reference can be a _hugely_ optimized way of achieving something _equivalent_ to comparing them _by value_ because `obj1 === obj2` will never be true without `deepRecursiveEq(obj1, obj2)` also being true. In purely functional languages the compiler knows that data is never modified so it can use this fact to optimize. Data structures in JS are not immutable so users will have to do this optimization themselves when they know they're not modifying data (unless using a library with purely functional data structures which can sometimes do these optimizations automatically).

Thus reference equality is not a "half-baked operator" it is instead a highly important operator for people who do functional programming with mutable data structures.

ivan-kleshnin:
> Thus reference equality is not a "half-baked operator" it is instead a highly important operator for people who do functional programming with mutable data structures.

Pls, provide an expected difference between native JS comparison operators and this `identical` function.

paldepind:
@ivan-kleshnin You can look at the [current implementation](https://github.com/ramda/ramda/blob/master/src/internal/_eq.js). The main reason for it's existence is to be able to use `===` where a function is expected. That's also why there is `R.add`, `R.subtract`, etc.

ivan-kleshnin:
> You can look at the current implementation. The main reason for it's existence is to be able to use === where a function is expected. That's also why there is R.add, R.subtract, etc.

Well, if someone really needs em – don't listen to me :smile: I'm just thinking that **optimization** means much more radical things. Like rewriting loops, tail calls etc. And such things are much better to leave for compiler than pile them up in a library.

paldepind:
> Well, if someone really needs em – don't listen to me :smile:
They're nice to have and easy to implement.

> I'm just thinking that optimization means much more radical things. Like rewriting loops, tail calls etc. And such things are much better to leave for precompiler than pile them up in a library.
I was talking about the optimization of replacing `deepRecursiveEq` with `===`. It can certainly be a significant optimization.

ivan-kleshnin:
And it can be easily done with `(a, b) => a === b` in case you want to apply it in a critical place. 
No need to reserve a function for it.

src/internal/_toString.js
davidchambers: Thanks for raising this issue, @CrossEye. There is a connection. Here's the definition of `recur`:

```javascript
var recur = function recur(y) {
  var xs = seen.concat([x]);
  return _indexOf(xs, y) >= 0 ? '<Circular>' : _toString(y, xs);
};
```

To evaluate `R.toString(new Number(0))`, we end up evaluating `_indexOf([new Number(0)], 0)` which *should* give `-1` but actually gives `0`. As a result we return `'new Number(<Circular>)'` rather than `'new Number(0)'`. Once #1099 has been merged I'll rebase and revert this workaround.

davidchambers: I've reverted the changes to this file.

test/equals.js
davidchambers: Good catch! I'd forgotten to add dynamic dispatch. I've updated the pull request.

buzzdecafe:
LGTM. any reason *not* to do this?

CrossEye:
Still not thrilled with the name `identical`, but I'd rather get it an and bikeshed over names later if necessary.  Let's do it!

paldepind:
Did you guys miss my comment about the names?

CrossEye:
I did see it, and thought I responded to it, and to several comments from @ivan-kleshnin as well.  Not sure what happened to that comment.

In any case, what I said was that I liked `equals` for the FantasyLand integration, and that I did prefer `strictEq` / `strictEq` to `identical`.

Since both @davidchambers and @buzzdecafe were expressing agreement with the current version, and my only caveat is a very weak preference for something other than `identical`, it seemed ready to merge.

@buzzdcafe, @davidchambers, do you want to revert this, and re-open discussion on the names?

paldepind:
> In any case, what I said was that I liked equals for the FantasyLand integration, and that I did prefer strictEq / strictEq to identical.

There's no reason why Ramda could not have an `R.eq` that dispatched to a method named `equals`. I don't see why having a strict one to one mapping between function names and what methods they dispatch to would be a goal. Especially not when it means giving up on the name of a core function whose name is a included in 5 other functions and when that new name is 4 characters longer. It's not going to work anyway when someone someday wants `R.reduce` to dispatch to a `foldl` method (or something else like that).

buzzdecafe:
> There's no reason why Ramda could not have an R.eq that dispatched to a method named equals. I don't see why having a strict one to one mapping between function names and what methods they dispatch to would be a goal.

Yes it can be done. It probably will eventually be done (e.g. for highland compat). But I like the 1-1 correspondence with FL names so we can say: "You can write FL compatible stuff point free" instead of  "You can write FL compatible stuff point free, except for `equals` which you have to use `eq` for".

buzzdecafe:
i am also not thrilled with `identical`, so let the bike-shedding continue.

davidchambers:
> i am also not thrilled with `identical`, so let the bike-shedding continue.

Clojure's equivalent function is named [`identical?`][1]. I'm not attached to `identical`: it makes sense to me, but to @benperez (and possibly @paldepind) identical is synonymous with equal, which is a problem.


[1]: https://clojuredocs.org/clojure.core/identical_q

paldepind:
> Yes it can be done. It probably will eventually be done (e.g. for highland compat). But I like the 1-1 correspondence with FL names so we can say: "You can write FL compatible stuff point free" instead of "You can write FL compatible stuff point free, except for equals which you have to use eq for".

If I heard the last sentence I'd probably just be like "nice, a shorter name for `equals`". But in addition to that I also think the fantasy land spec is mostly for implementers and I think that they should've done [something like this](https://github.com/fantasyland/fantasy-land/issues/92). I.e. I don't think choosing names based on what fantasy land calls things makes much sense.

Question: Should `eqProps`, `pathEq`, `propEq` and `whereEq` be renamed?

> Clojure's equivalent function is named identical?. I'm not attached to identical: it makes sense to me, but to @benperez (and possibly @paldepind) identical is synonymous with equal, which is a problem.

I think the name makes sense when you know what it does and think about it. It's not bad. But it's not as crystal clear as my suggestion (which was `strictEq`, but that's not going to work without `eq`).

davidchambers:
Having functions named `R.eq` and `R.strictEq` would be a source of confusion. Since `===` is known as the strict equality operator, many people would guess—wrongly—that the functions correspond to `==` and `===` respectively.

I agree that having `R.equals` and a number of functions with `eq` or `Eq` in their names is not ideal. I would find it acceptable for `R.eq` to dispatch to `equals`.

paldepind:
@davidchambers My initial argument was that people would also wrongly guess that `R.equals` and `R.identical` is `==` vs `===`. But I can certainly see what you mean. This could be an even more likely conclusion if one of the names included `strict`.

Without putting `deep` in the name I don't think any JavaScript would initially expect a function named `R.eq` nor `R.equals` to do a deep comparison. So not matter what this could potentially be a source of confusion (or a positive surprise).

CrossEye:
> I agree that having `R.equals` and a number of functions with `eq` or `Eq` in their names is not ideal. I would find it acceptable for `R.eq` to dispatch to `equals`.

Nothing seems great to me.  But I would rather deal with the `equals` / `eq` inconsistency than the dispatching one.

jethrolarson:
`R.threequals`

PR #1097, opened 2015-05-21
lint: increase search depth by one level

davidchambers:
I miss `find`. :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,8 +36,8 @@
     "browserify": "node_modules/browserify/bin/cmd.js test/*.js --outfile lib/test/bundle.js",
     "bookmarklet": "scripts/bookmarklet",
     "clean": "rimraf dist",
-    "jscs": "jscs scripts/bookmarklet scripts/build **/*.js",
-    "jshint": "jshint scripts/bookmarklet scripts/build **/*.js",
+    "jscs": "jscs scripts/bookmarklet scripts/build **/*.js **/**/*.js",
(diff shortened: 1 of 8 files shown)


buzzdecafe:
:cow2:

package.json
```diff
@@ -36,8 +36,8 @@
     "browserify": "node_modules/browserify/bin/cmd.js test/*.js --outfile lib/test/bundle.js",
     "bookmarklet": "scripts/bookmarklet",
     "clean": "rimraf dist",
-    "jscs": "jscs scripts/bookmarklet scripts/build **/*.js",
-    "jshint": "jshint scripts/bookmarklet scripts/build **/*.js",
+    "jscs": "jscs scripts/bookmarklet scripts/build **/*.js **/**/*.js",
```
CrossEye: Too late to the party once again.  I would rather just:

    "jscs": "jscs scripts/bookmarklet scripts/build .",

which will allow it to work in Windoze too.

package.json
```diff
@@ -36,8 +36,8 @@
...
+    "jshint": "jshint scripts/bookmarklet scripts/build **/*.js **/**/*.js",
```
CrossEye: and here as well.

package.json
davidchambers: Wouldn't that match files without the __.js__ extension?

CrossEye: I don't think so.  I mean to test it last night at home.  But I think that means for both jscs and jshint: "Search starting here for any files ending in '.js`."

I did test yesterday on a Windows 7 machine and it worked fine.

PR #1098, opened 2015-05-21
lint: enable JSHint's eqeqeq option

davidchambers:


buzzdecafe:
surprised this isn't set already. as long as `eqnull` is still good i'm all :cow2:

CrossEye:
While I don't object, I want to note that all of the non-test code that had to be changed was non-coercing.  I only buy this because I want a tool to be able to check this for us, and the tool isn't going to do the analysis to know that "since `typeof` always returns a string, `==` is perfectly appropriate here."  Otherwise I might object.  :smiling_imp:

davidchambers:
> the tool isn't going to do the analysis to know that "since `typeof` always returns a string, `==` is perfectly appropriate here."

Even if JSHint could be configured as you would like I would still favour `===`. It keeps the rule simple: use `===` and `!==` except when comparing to null/undefined, in which case `==` and `!=` are appropriate.

PR #1099, opened 2015-05-21
fix handling of `new Number(0)` in R.eq

davidchambers:
Currently `R.eq(0, new Number(0))` evaluates true, which is incorrect.


buzzdecafe:
:cow2:

PR #1100, opened 2015-05-21
fix bug causing R.eq(NaN, x) to evaluate true for every x

davidchambers:
I've made the same change to the MDN [polyfill][1].

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is#Polyfill


buzzdecafe:
yikes! :cow2:

PR #1111, opened 2015-05-26
deprecate R.arity

### LINKED ISSUE #1107: Deprecate `R.arity` and/or `R.nAry`
I think at least one of these functions should be deprecated.

`R.nAry` is just `R.curryN` that doesn't curry. But when you have a fixed arity function there's no reason not to curry. It creates confusion. The first time (and only) time I used `R.nAry` it I expected it to curry and thus introduced a bug into my program.

`R.arity` seems really funky to me. It pretends to create a fixed arity function. But it's not really fixed since it can be abused by passing extra arguments. How is that going to help with anything but writing contrived code?

davidchambers:
Closes #1107

### DIFF
--- a/src/arity.js
+++ b/src/arity.js
@@ -14,6 +14,7 @@ var _curry2 = require('./internal/_curry2');
  * @param {Function} fn The function to wrap.
  * @return {Function} A new function wrapping `fn`. The new function is
  *         guaranteed to be of arity `n`.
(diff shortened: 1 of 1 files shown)


buzzdecafe:
:cow2:

CrossEye:
:herb:

I wonder if we can remove the half-dozen or so internal uses of it as well and get rid of the damn function altogether...

PR #1113, opened 2015-05-26
dispatch in R.indexOf and R.lastIndexOf

davidchambers:
This pull request was prompted by plaid/sanctuary#38. It'll be nice to remove `R.strIndexOf` and `R.strLastIndexOf`, which have always felt awkward to me. :)


CrossEye:
:herb:

These were written before the dispatching mechanism, I believe.  Changing to use it just makes sense.

Issue #1109, opened 2015-05-25
Expose _reduced?


garyposter:
What are your thoughts on exposing _reduced to the API?  Not important to me right now, but it would scratch an itch at least, and I could imagine it being more important to me or others in the future.

Thanks for a cool library!

CrossEye:
I think it's a fine idea, and I would welcome a PR.

garyposter:
Cool, thanks. :+1:

CrossEye:
Closed by #1118

PR #1118, opened 2015-05-27
expose "reduced" to public API

garyposter:
Simple first approach to #1109, taking a stab at various questions along the way.

### DIFF
--- a/test/reduced.js
+++ b/test/reduced.js
@@ -1,0 +1,31 @@
+var assert = require('assert');
+
+var R = require('..');
+
+
+describe('reduced', function() {
+  it('wraps a value', function() {
+    // white box test.
+    var v = {};
+    assert.strictEqual(R.reduced(v)['@@transducer/value'], v);
+  });
+
+  it('flags value as reduced', function() {
+    // white box test.
+    assert.strictEqual(R.reduced({})['@@transducer/reduced'], true);
+  });
+
+  it('short-circuits reduce', function() {
+    // black box test.
+    assert.strictEqual(
+      R.reduce(
+        function(acc, v) {
+          var result = acc + v;
+          if (result >= 10) {result = R.reduced(result);}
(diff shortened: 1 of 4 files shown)


test/reduced.js
```diff
@@ -0,0 +1,27 @@
...
+    var v = {};
+    assert.strictEqual(R.reduced(v)['@@transducer/value'], v);
+  });
+
+  it('flags value as reduced', function() {
+    // white box test.
+    assert.strictEqual(R.reduced({})['@@transducer/reduced'], true);
+  });
+
+  it('shortcircuits reduce', function() {
```
buzzdecafe: probably want to hyphenate (i.e. "short-circuits") when verbing this noun

test/reduced.js
```diff
@@ -0,0 +1,27 @@
...
+    // black box test.
+    assert.strictEqual(
+      R.reduce(
+        R.pipe(R.add, R.ifElse(R.lte(10), R.reduced, R.identity)),
```
buzzdecafe: for a unit test i think we're better off writing out a lambda here, rather than having internal dependencies

test/reduced.js
garyposter: cool, can do.

test/reduced.js
garyposter: Makes sense, assuming we keep R.reduce itself.

buzzdecafe: yes, since the point of the test is to show how it integrates with `reduce`, let's keep that

garyposter:
Changes made & ready for re-review.  Thanks!

buzzdecafe:
LGTM. :cow2: 

anybody else have a word to put in?

CrossEye:
:herb:

garyposter:
Thanks! :-)

PR #1122, opened 2015-05-27
Add support to `R.pick` for numbers

### LINKED ISSUE #1121: R.pick doesn't support numeric object keys.
Assume we create an entry in an object as follows:

```javascript
  var obj = {};
  obj[1] = 'foo';
  console.log(obj[1]);
  console.log(obj['1']);
```

Logs 'foo' twice.

However,

```javascript
  var obj = {};
  obj[1] = 'foo';
  console.log(R.pick([1], obj));
  console.log(R.pick(['1'], obj));
```

Logs {}, and then {1: hello}.

paldepind:
Closes #1121.

### DIFF
--- a/src/pick.js
+++ b/src/pick.js
@@ -1,5 +1,4 @@
 var _curry2 = require('./internal/_curry2');
-var _indexOf = require('./internal/_indexOf');
 
 
 /**
@@ -9,7 +8,7 @@ var _indexOf = require('./internal/_indexOf');
  * @func
  * @memberOf R
  * @category Object
- * @sig [String] -> {String: *} -> {String: *}
+ * @sig [k] -> {k: v} -> {k: v}
  * @param {Array} names an array of String property names to copy onto a new object
  * @param {Object} obj The object to copy from
  * @return {Object} A new object with only properties from `names` on it.
@@ -20,9 +19,10 @@ var _indexOf = require('./internal/_indexOf');
  */
 module.exports = _curry2(function pick(names, obj) {
   var result = {};
-  for (var prop in obj) {
-    if (_indexOf(names, prop) >= 0) {
-      result[prop] = obj[prop];
+  var idx = -1;
+  while (++idx < names.length) {
+    if (names[idx] in obj) {
+      result[names[idx]] = obj[names[idx]];
     }
   }
   return result;
--- a/test/pick.js
+++ b/test/pick.js
@@ -4,10 +4,14 @@ var R = require('..');
 
 
 describe('pick', function() {
-  var obj = {a: 1, b: 2, c: 3, d: 4, e: 5, f: 6};
+  var obj = {a: 1, b: 2, c: 3, d: 4, e: 5, f: 6, 1: 7};
 
   it('copies the named properties of an object to the new object', function() {
     assert.deepEqual(R.pick(['a', 'c', 'f'], obj), {a: 1, c: 3, f: 6});
+  });
+
+  it('handles numbers as properties', function() {
+    assert.deepEqual(R.pick([1], obj), {1: 7});
   });
 
   it('ignores properties not included', function() {


paldepind:
Should have better performance as well. The old algorithm was O(n*m) where `n` is number of names and `m` is keys on target object. This is faster since `name in obj` probably has better complexity than linear.

buzzdecafe:
> The old algorithm was O(n*m)

yes, when i looked at the difference between `pick` and `pickAll` in current code, I was surpised to see it iterated all the object properties. This implementation LGTM, thanks @paldepind 

:cow2:

src/pick.js
```diff
@@ -9,7 +8,7 @@ var _indexOf = require('./internal/_indexOf');
  * @func
  * @memberOf R
  * @category Object
- * @sig [String] -> {String: *} -> {String: *}
+ * @sig [k] -> {k: v} -> {k: v}
```
davidchambers: `String` seems more appropriate than `k`. The latter suggests the list can contain values of any type.

Also, `{k: v}` suggests all the values in the object must be of the same type, which needn't be the case.

paldepind: I simply reused the signature from `pickAll`. But I agree with you. Should I change it back to the old signature as well as changing `pickAll`?

davidchambers: In my view, yes, but perhaps @buzzdecafe has a different view.

buzzdecafe: i don't have strong feelings on the subject. I will point out, however, that the whole point is that this implementation will support numeric keys. So `k` may be more accurate than `String`

davidchambers: We document *intended* use in the type signatures. We often use `[a]`, for example, though the elements of an array needn't be of the same type.

CrossEye:
We also should really note that the idea of numeric indices is a little slippery:

```js
var obj = {a: 1, b: 2, '1': 3};
obj[2] = 4;
pick(['a', 1, '2'], obj);  //=> {"1": 3, "2": 4, "a": 1}
```

The output object does not have a `1` property, but a `'1'` property, which maps to the input `'1'` property.  The output has a `'2'` property, which was found from what we set as the `2` property on the input object.  This is simply how JS works, but I don't know that we actually want to mention numbers here.

I agree that this is a better algorithm, and we should definitely unify `pick` and `pickAll`, but I think we should probably not document this as having anything to do with numeric keys.

buzzdecafe:
> I think we should probably not document this as having anything to do with numeric keys.

ok, then let's remove that test and use the sig as recommended by @davidchambers

paldepind:
> [...] from what we set as the `2` property on the input object [...]

We did not set any `2` property. `2` is a number and a property key cannot be a number.

> we should probably not document this as having anything to do with numeric keys.

There is no such thing as numeric keys in JavaScript. Whenever you supply a primitive that is not a string it is automatically converted to a string.

```javascript
Object.keys({undefined: 1, 12: 2, null: 3}); //=> ["12", "undefined", "null"]
```

This implementation of `R.pick` has the same behavior. I.e. you can use any primitive as a key.

> ok, then let's remove that test and use the sig as recommended by @davidchambers

I do not want to remove the test. @josephong requested this behavior. When it is released him and others will/might write code that depends on it! If the test is removed a regression could happen that would break his code and the nothing would be there to show it!

I think accepting the same things that can be used to access a property in normal JavaScript in a property access function is very sensible. But the behavior should be documented somehow so that people know that it is intentional and that they can rely on it!

CrossEye:
> There is no such thing as numeric keys in JavaScript. 

That's the point I was trying to make.  I was just suggesting that between the name of the issue, test code like this

```js
it('handles numbers as properties', function() {
```

and suggestions to use `{k: v}` instead of `String`, we seemed to be losing sight of that basic fact in our discussion here.

test/pick.js
```diff
@@ -4,12 +4,16 @@ var R = require('..');
 
 
 describe('pick', function() {
-  var obj = {a: 1, b: 2, c: 3, d: 4, e: 5, f: 6};
+  var obj = {a: 1, b: 2, c: 3, d: 4, e: 5, f: 6, 1: 7};
```
davidchambers: I believe that property names which are not valid identifiers must be quoted in old environments. Perhaps we should do so here.

paldepind: I actually did so initially. But JSHint complained!

paldepind:
@CrossEye I don't see anything wrong with the title, the test description or the signature. Only mentioning strings is losing sight of the fact that using numbers to access properties is second nature to any JavaScript developer (especially when using arrays).

paldepind:
There are a lot of Ramda functions that access properties of objects with a passed in key. Here are a few of them (there are probably some I missed) `pickAll`, `lensProp`, `dissoc`, `dissocPath`, `has`, `omit`, `props` and `project`. Today the type signatures of these functions are very inconsistent. Some specify that the key is of type `String` and some use `k`.

So it would probably be a great idea to make decision now and then update the signature of other functions accordingly to achieve consistency.

I think Ramda should be intentional about inheriting the property access semantics from JavaScript and in that case it should be documented _somehow_ so users know they can rely on it. Using `String` in the type signature would not express that numbers are part of the intended use case.

One idea I just had (I'm not sure if it's a good idea) would be to invent a `Key` type and specify somewhere that it's defined as all the values a user can typically use to access the properties of an object. I.e. anything that can appear inside square brackets when accessing properties. The advantage would be that the type would be easy to understand, not too specific (like `String`) and not too vague (like just `k`).

buzzdecafe:
i agree we should standardize the object sigs. But I don't think that should hold up this PR. is this one good to go?

paldepind:
Yes. This should be good to go. The function signature matches `pickAll` which is at least something

CrossEye:
Yes, we need to make another pass at signatures for all our functions I think, and make sure we've got a standard convention across the library.  But until we figure that out, the _ad hoc_ method we've got will have to do.

PR #1129, opened 2015-05-29
Implement mapKeys and test it

### LINKED ISSUE #1124: R.objFromKeys(valueFunction, keys)
Similar to mapObj, a utility that generates an object given a value-for-key function and an array of keys seems useful.

Something that works like this: http://bit.ly/1AwyGiv

asaf-romano:
Closes #1124


asaf-romano:
Some alternative names suggested in the ticket.

test/mapKeys.js
```diff
@@ -0,0 +1,9 @@
+var assert = require('assert');
+
+var R = require('..');
+
+describe('mapKeys', function() {
+  it('returns an object with each of the keys passed set to value returned for it by the given function', function() {
+    assert.deepEqual(R.mapKeys(R.toUpper, ['a', 'b', 'c']), {a: 'A', b: 'B', c: 'C'});
+  });
+});
```
buzzdecafe: definitely needs more tests. what happens for empty list? is the function curried? what if the input list is a multiset? which one wins?

CrossEye: I think multiset would only matter for a non-deterministic function, not a case I'm particularly worried about, and would happily leave to a last-in-wins scenario.  But I agree that the tests should demonstrate a little more of its behavior, especially around the edges.

asaf-romano: Honestly, I just modeled this after the test after the test for mapObj. I'll try to elaborate it a little bit.

By the way, thinking about the ongoing ES6-Map support I'm now also against the mapKeys naming, and would rather go with objFromKey proposed in the other bug (for similar reason, I think createMapEntry should be renamed).

asaf-romano:
Sigh, I renamed the branch and github got all confused. Opened new PR #1137

test/mapKeys.js
buzzdecafe: > Honestly, I just modeled this after the test after the test for mapObj. 

thanks for the heads up -- we should fix that

PR #1137, opened 2015-05-30
objFromKeys

### LINKED ISSUE #1124: R.objFromKeys(valueFunction, keys)

asaf-romano:
Closes #1124


asaf-romano:
Friendly ping.

buzzdecafe:
LGTM :cow2:

paldepind:
Hm. I think implementing this with a `for` or a `while` loop would be better. The code would be just as short and performance should be better.

Most importantly, I generally think code that modifies the accumulator in a `reduce` is a bit ugly. If one is doing mutations I think using a construct that scream "imperative" is better than one that pretends to be functional.

But that is just my opinion. This function seems very useful so I'm happy no matter what :)

buzzdecafe:
> The code would be just as short and performance should be better.

otoh, this one will be transducible, which is a win IMO

> I generally think code that modifies the accumulator in a reduce is a bit ugly.

i would agree for "public" code, but not for internal implementation detail

paldepind:
> otoh, this one will be transducible, which is a win IMO

Indeed! That is a big win in functionality! Certainly more important than other concerns.

> i would agree for "public" code, but not for internal implementation detail

The way I see it you don't win anything in elegance but you lose something in performance.

asaf-romano:
The current implementation simply follows mapObj, and since they do a similar job I believe they should be in sync unless there's a good reason not do so. FWIW, back when this was discussed in the gitter channel, numerous implementations were proposed (mine was something like (es6-synatax) curry((valueFunction, keys) => fromPairs(map(k => [k, valueFunction(k)], keys))) ), and IIRC @CrossEye was for a reduce-based implementation as done here (but probably using assoc in each iteration).

davidchambers:
This need is quite easily satisfied without a new function:

```javascript
R.zipObj(keys, R.map(R.toUpper, keys));
```

buzzdecafe:
i like that @davidchambers . The only difference is that won't transduce, but maybe we could make `zipObj` a transducer function.

CrossEye:
I think my implementation when this was discussed was much uglier.  As we talked this out, @buzzdecafe drew out much the same solution as is presented here on a whiteboard.

@paldepind: If you have a loop-based alternative (ideally, `while` :smiling_imp:) I'd be interested.  I don't exactly share your objection to the mutation here, but that's partly because Ramda does all sorts of mutation with its own local variables.  That's how it works.  But I do see some reason not to do it, along the lines of not wanting to promote bad practices.

@davidchambers: That's nicest of all, and maybe this is simply not needed.  Of course if there's enough of a need for a function like this, then it might be worth considering in any case.

asaf-romano:
Hehe, my own use case that brought this up was something like this
```
objFromKeys([...], always(0))
```
And my very first (in my own code) implementation of that was using zipObj, and it looked like this:

```
var keys = ["a", "b"];
R.zipObj(keys, R.repeat(0, keys.length));
```

I didn't like that for two reasons:
1) I cannot inline the keys (that is, if I do I have to hardcode the length)
2) It's not as declarative as it could be.

davidchambers:
You could use `R.converge` to avoid referencing `keys` in two places:

```javascript
R.converge(R.zipObj, R.identity, R.map(R.always(0)))(keys);
```

Perhaps this is more appropriate as a [recipe][1] rather than as a specific Ramda function.


davidchambers:
Will anyone champion this pull request? I'd prefer not to add this function to the library.

megawac:
Agree with David here, this is better as a recipe

raine:
We could have this in the upcoming ramda-contrib lib.

CrossEye:
@asaf-romana?  I think we're mostly looking for your arguments in favor of this.  No one want to simply cut off debate.  It's an interesting function.  But no one else has seen it as quite important enough to make it into Ramda itself?  Do you want to convince us?

asaf-romano:
I've had a few use cases for this function in my projects, admittedly some of them are very border line for FP-style code (e.g. initializing a key-to-count mapping object). Another repeating pattern is constructing maps like { backgroundColor: "background-color" } as demonstrated in the chat room some time ago.

I do have to admit though that one of the reasons I view this as a utility function is because I'm trying to avoid ```converge``` usage as much as possible outside my own "utility belts": for my Conservative OOP Colleagues, it was much easier to grasp currying and composing than useWith and converge usage. Obviously if you do see converge/useWith as elegant and clear, converge+zipObject is more than enough.

CrossEye:
Ok, this does sound like a recipe or contrib.  It's been interesting.  And there are occasions when I might use such a function.

As to `converge` / `useWith`, I agree that they _should_ somehow feel like secondary solutions.  There is something dirty about them that doesn't taint `compose` / `pipe`.  I don't know why.

asaf-romano:
For the record, I forgot the most obvious use case when I wrote the previous comment:

objectsMap = objectFromKeys(findOneById, ids);

Anyway, I'm good with closing this for now. If this turns out useful I guess we can reopen this later.

asaf-romano:
Regarding converge dirty-ness... Unlike other cases discussed elsewhere, converge is the only API in which the variadic signature actually hurts readability, as it puts the first function at the same level with all the others. The API would have been much easier to follow if it took the converged functions as an array, imho.

davidchambers:
> The API would have been much easier to follow if it took the converged functions as an array, imho.

:thumbsup: 100% agree!

CrossEye:
Agreed:  `converge` is one of the few remaining variadic functions.  I'd welcome a PR to fix this.

davidchambers:
#1229

PR #1138, opened 2015-05-30
Dispatch forEach

### LINKED ISSUE #1134: Dispatch forEach for anything but arrays
Symbol.iterator / @@iterator aside, both ES6 and third-party libs follow the pattern of implementing forEach for all iterables, as the "functional analogy" for for-of loops.

Now, I know that for arrays ramda's forEach semantics are intentionally different from Array.prototype.forEach, but for anything else I believe dispatching forEach makes sense.

asaf-romano:
Closes #1134

### DIFF
--- a/src/forEach.js
+++ b/src/forEach.js
@@ -1,5 +1,6 @@
 var _curry2 = require('./internal/_curry2');
 var _forEach = require('./internal/_forEach');
+var _hasMethod = require('./internal/_hasMethod');
 
 
 /**
@@ -30,4 +31,6 @@ var _forEach = require('./internal/_forEach');
  *      //-> 7
  *      //-> 8
  */
-module.exports = _curry2(_forEach);
+module.exports = _curry2(function forEach(fn, list) {
+  return _hasMethod('forEach', list) ? list.forEach(fn) : _forEach(fn, list);
+});
--- a/test/forEach.js
+++ b/test/forEach.js
@@ -22,6 +22,18 @@ describe('forEach', function() {
     assert.deepEqual(R.forEach(function(x) { return x * x; }, []), []);
   });
 
+  it('dispatches to `forEach` method', function() {
+    var dispatched = false;
+    var fn = function() {};
+    function DummyList() {}
+    DummyList.prototype.forEach = function(callback) {
+      dispatched = true;
+      assert.strictEqual(callback, fn);
+    };
+    R.forEach(fn, new DummyList());
+    assert.strictEqual(dispatched, true);
+  });
+
   it('is curried', function() {
     var xStr = '';
     var xe = R.forEach(function(x) { xStr += (x + ' '); });


buzzdecafe:
failing jshint. did you run `npm test`?

asaf-romano:
Indeed not :-/ I just run mocha.

buzzdecafe:
please squash to 1 commit, then LGTM. :cow2:

asaf-romano:
Done, I think.

asaf-romano:
Well, it does test the dispatching, because DummyList is not an array, and thus fn wouldn't be called if forEach wasn't dispatched. Indeed, I could set called in the dispatched forEach, but then I'm testing one thing less: that the callback was passed properly in the dispatching case.

davidchambers:
:deciduous_tree:

test/forEach.js
```diff
@@ -22,6 +22,14 @@ describe('forEach', function() {
     assert.deepEqual(R.forEach(function(x) { return x * x; }, []), []);
   });
 
+  it('dispatches to `forEach` method', function() {
+    function DummyList() {}
+    DummyList.prototype.forEach = function(fn) { fn(); };
+    var called = false;
+    R.forEach(function() { called = true; }, new DummyList());
+    assert(called);
```
buzzdecafe: house style is `assertStrictEqual(x, y)` here.

CrossEye:
> Well, it does test the dispatching, because DummyList is not an array, 

I think if you passed an array instead of a DummyList, the assertion would still pass.  That's the problem.  If you did this just inside the dispatched `forEach` then the assertion would only pass for the correct case.

asaf-romano:
Let's test both then.

asaf-romano:
Fixed both issues.

CrossEye:
:herb:

src/forEach.js
```diff
@@ -30,4 +31,6 @@ var _forEach = require('./internal/_forEach');
  *      //-> 7
  *      //-> 8
  */
-module.exports = _curry2(_forEach);
+module.exports = _curry2(function forEach(fn, list) {
+  return _hasMethod('forEach', list) ? list.forEach(fn) : _forEach(fn, list);
```
benjycui: Here is a problem.
> Also note that, unlike `Array.prototype.forEach`, Ramda's `forEach` returns the original
> array. In some libraries this function is named `each`.

But if we dispatch `forEach` to `Map`/`Set`, `undefined` will be returned.
See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/forEach#Description
> `forEach` executes the callback function once for each element in the Map object; it does not return a value.

So does `Set`.

So, we should implement `forEach` in this way, right?
```js
module.exports = _curry2(function forEach(fn, list) {
  _hasMethod('forEach', list) ? list.forEach(fn) : _forEach(fn, list);
  return list;
});
```

CrossEye: > But if we dispatch `forEach` to `Map`/`Set`, `undefined` will be returned.

That's a problem with all dispatching.  We can't guarantee that the return of the dispatched-to function matches our own.  The user is really responsible for that.

In this case, it's really not a big deal, to my eyes, as `forEach` has no logical return.  Ramda simply returns the original list in order to return _something_, not because there's any great virtue in that particular return.

benjycui: OK, I am just confused for the returned value is unstable. THX :-)

PR #1139, opened 2015-06-01
replace R.curryN and R.nAry with R.curryMinMax

davidchambers:
As suggested in https://github.com/ramda/ramda/issues/1107#issuecomment-106011338

Should we deprecate `R.curryN` rather than removing it? It would be confusing to have both `R.curryN` and `R.curryMinMax` for one release, but perhaps it would make upgrades smoother for some users.

If we decide to merge this pull request, we should consider renaming `R.uncurryN`. `R.uncurry` may be a better name.

### DIFF
--- a/src/addIndex.js
+++ b/src/addIndex.js
@@ -2,7 +2,7 @@ var _concat = require('./internal/_concat');
 var _curry1 = require('./internal/_curry1');
 var _prepend = require('./internal/_prepend');
 var _slice = require('./internal/_slice');
-var curryN = require('./curryN');
+var curryMinMax = require('./curryMinMax');
 
 
 /**
@@ -28,7 +28,7 @@ var curryN = require('./curryN');
  *      //=> ['0-f', '1-o', '2-o', '3-b', '4-a', '5-r']
  */
 module.exports = _curry1(function(fn) {
-  return curryN(fn.length, function() {
+  return curryMinMax(fn.length, fn.length, function() {
     var idx = -1;
     var origFn = arguments[0];
     var list = arguments[arguments.length - 1];
--- a/src/allPass.js
+++ b/src/allPass.js
@@ -1,6 +1,6 @@
 var _all = require('./internal/_all');
 var _predicateWrap = require('./internal/_predicateWrap');
-var curry = require('./curry');
+var curryMinMax = require('./curryMinMax');
 
 
 /**
@@ -22,4 +22,4 @@ var curry = require('./curry');
  *      f(11); //=> false
  *      f(12); //=> true
  */
-module.exports = curry(_predicateWrap(_all));
+module.exports = curryMinMax(1, Infinity, _predicateWrap(_all));
--- a/src/anyPass.js
+++ b/src/anyPass.js
@@ -1,6 +1,6 @@
 var _any = require('./internal/_any');
 var _predicateWrap = require('./internal/_predicateWrap');
-var curry = require('./curry');
+var curryMinMax = require('./curryMinMax');
 
 
 /**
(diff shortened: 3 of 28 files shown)


CrossEye:
I think we do want the deprecation.  And I agree that `uncurry` would be better.

I have one question: why did you move `arity` to an internal function?  What is the advantage to it?  I ask because I'm all in favor of the other direction, removing such internal helpers anywhere they're not strictly necessary.  I'd like to be eating our own dogfood where we can.

davidchambers:
> I think we do want the deprecation.

Okay. I'll update the pull request.

> I have one question: why did you move `arity` to an internal function?

`R.arity` is a deprecated function. We'll need to keep it around after we remove it from the public API, so we need to make it an internal function at some point. This pull request struck me as a good opportunity to do so.

CrossEye:
> `R.arity` is a deprecated function. 

Ok, that's fine.  I didn't notice the deprecation warning.  I have no trouble with truly internal functions.

Other than that, this LGTM.

paldepind:
I must admit that I fail to see the practical benefit that this arguably more complex function offers compared to the current `R.curryN`? The way I understand it `R.curryMinMax(n, n, fn)` would create a well behaved curried function identical to current `R.arity` if curried. And `R.curryMinMax(n, Infinity, fn)` would create a curried function identical to current `R.curryN`  based on this slightly ad-hoc but practical notion of currying to a limiting arity. It seems to me like these two functions would cover all use cases while having reasonable semantics?

davidchambers:
> I must admit that I fail to see the practical benefit that this arguably more complex function offers compared to the current `R.curryN`?

The goal is to reduce the number of functions with similar roles. #1111 deprecated `R.arity`, and we can deprecate `R.nAry` (and `R.unary` and `R.binary`) if we give `R.curryN` the ability to limit the number of arguments forwarded to the wrapped function.

v0.14.0:

|                   | sets arity   | curries      | limits args  |
| ----------------- |:------------:|:------------:|:------------:|
| `R.arity`         |   &#x2714;   |              |              |
| `R.curryN`        |   &#x2714;   |   &#x2714;   |              |
| `R.nAry`          |   &#x2714;   |              |   &#x2714;   |

Proposed:

|                   | sets arity   | curries      | limits args  |
| ----------------- |:------------:|:------------:|:------------:|
| `R.curryMinMax`   |   &#x2714;   |   &#x2714;   |   &#x2714;   |

paldepind:
`R.arity`. is already deprecated independently of this PR.

Well, I think having two simple functions with (somewhat) well defined understandable semantics is simpler than having one function that can do a lot a lot of things – some of which no one knows a use case for.

This function is more powerful and complex than `R.nAry` and `R.curryN` combined. And I currently see it as unnecessary power that pulls away from the otherwise simple concepts of currying and arity. This function also seems to invite users to create variadic functions and it even more embraces currying variadic functions.

Also let me point out that this function cannot replace `R.nAry`. It actually does something quite different from `R.nAry`. Now, I still don't think they where very convincing, but the examples that @CrossEye posted where all about _allowing_ but _ignoring_ additional arguments passed to a function. This function however is about _disallowing_ additional arguments completely by throwing an error.
This also means that this function goes against the established idea in JS that passing too many arguments to a function is completely fine. The curent `R.nAry` is on the other hand based on the recognition that functions will sometimes be given too many arguments.

davidchambers:
> Also let me point out that this function cannot replace `R.nAry`. […] This function […] is about _disallowing_ additional arguments completely by throwing an error.

Ah yes, good point! I set out to ignore additional arguments, but then realized additional arguments are confusing in conjunction with placeholders. If `f` is a binary curried function, what does `f(R.__, 'foo', 'bar')` mean?

Another option is to deprecate `R.nAry`, `R.unary`, and `R.binary` without changing `R.curryN`. The confusion these functions create outweighs the utility they provide, in my view.

paldepind:
> Another option is to deprecate R.nAry, R.unary, and R.binary without changing R.curryN. The confusion these functions create outweighs the utility they provide, in my view.

Yes! That is what I think would be ideal.

Issue #1130, opened 2015-05-29
recursive deep merge / evolve


ccorcos:
[`merge`](https://github.com/ramda/ramda/blob/master/src/merge.js) and [`evolve`](https://github.com/ramda/ramda/blob/master/src/evolve.js) both use [`_extend`](https://github.com/ramda/ramda/blob/master/src/internal/_extend.js) to do a shallow merge / evolve. 

I think it would be really convenient if these were both deep:

```js
function merge(a, b) {
  return _.extend(clone(a), b);
});

function isPlainObject(value) {
  return Object.prototype.toString.call(value) == '[object Object]';
}

function _extend(destination, other) {
  var props = keys(other);
  var idx = -1, length = props.length;
  while (++idx < length) {
    if (isPlainObject(other[props[idx])) {
      destination[props[idx]] = _extend(destination[props[idx]], other[props[idx]])
    } else {
      destination[props[idx]] = other[props[idx]];
    }
  }
  return destination;
};

function evolve(a, b) {
  return _.evolve(clone(a), b);
});

function isFunction(value) {
  return Object.prototype.toString.call(value) == '[object Function]';
}

function _evolve(destination, other) {
  var props = keys(other);
  var idx = -1, length = props.length;
  while (++idx < length) {
    if (isPlainObject(other[props[idx])) {
      destination[props[idx]] = _evolve(destination[props[idx]], other[props[idx]])
    } else if (isFunction(other[props[idx])) {
      destination[props[idx]] = other[props[idx]](destination[props[idx]]);
    } else {
      destination[props[idx]] = other[props[idx]];
    }
  }
  return destination;
};
```

paldepind:
`R.merge

ccorcos:
Awesome. I like `R.set`. How about making `R.evolve` deep as well?

paldepind:
> How about making R.evolve deep as well?

See:

> I plan to soon do a PR that makes R.evolve recursive/deep just like R.set is.

ccorcos:
sweet :+1: I cant wait.

ccorcos:
I'd also like to suggest that `R.evolve` defaults to the `R.set` functionality when a value isnt a function. For example:

```js
R.evolve({name: 'joe', info: {age: R.add(1)}}, {name:'chet', info:{age:24}})
// => {name: 'joe', info: {age: 25}}
```

CrossEye:
:+1: on updating `evolve` to work deeply.

[...]

PR #1141, opened 2015-06-01
Make `R.evolve` recursive/deep

paldepind:
This PR makes `R.evolve` recurse into nested objects if one is present in both the object to evolve and the object of transformation functions.

Short example:

```javascript
R.evolve({nested: {object: R.add(1)}}, {nested: {object: 1}}); //=> {nested: {object: 2}}
```

This PR also changes the behavior of `R.evolve` when a transformation function exists for a property that does not exist on the object to evolve. Currently the transformation function is invoked with `undefined` and a new key is created on the returned object. New behavior is simply to not invoke the transformation function. I've decided to do that for the following reasons:

1. `evolve` is a function for modifying properties. To me it seems unexpected that it in some cases will also add new properties. I.e. not only modify existing properties but also change the shape of the object.
2. Passing `undefined` to a function otherwise used to update an existing property will rarely result in something desirable. Most often it will give broken data or runtime errors.
`R.evolve({amount: R.add(1)}, {}); //=> {amount: NaN}`
`R.evolve({name: R.toUpper}, {}); //=> Throws "TypeError: n is undefined"`
The transformation function would pretty much have to special case on `undefined`.
3. It makes `evolve` much more useful when dealing with objects of different shapes. I can do stuff like `R.map(R.evolve({name: R.toUpper}), list)` even if some items in `list` does not have a `name` property.
4. Ramda has other functions if one wants to add new properties to an object (`R.set` can be used for things like supplying defaults).
4. As an added benefit it makes the implementation simpler and more efficient since we'll only have to loop over the object to transform and not over the object of functions.

This is a breaking change. But I'd say it's a very low risk one. The old behavior was undocumented and quite odd. I doubt that many people depends on it. This PR documents the new behavior both in the description of `evolve` and in the provided example. The recursiveness is of course documented as well.

### DIFF
--- a/src/evolve.js
+++ b/src/evolve.js
@@ -1,11 +1,12 @@
 var _curry2 = require('./internal/_curry2');
-var _extend = require('./internal/_extend');
-var mapObjIndexed = require('./mapObjIndexed');
 
 
 /**
- * Creates a new object by evolving a shallow copy of `object`, according to the
- * `transformation` functions.  All non-primitive properties are copied by reference.
+ * Creates a new object by recursively evolving a shallow copy of `object`, according to the
+ * `transformation` functions. All non-primitive properties are copied by reference.
+ *
+ * A `tranformation` function will not be invoked if its corresponding key does not exist in
+ * the evolved object.
  *
  * @func
  * @memberOf R
@@ -17,10 +18,22 @@ var mapObjIndexed = require('./mapObjIndexed');
  * @return {Object} The transformed object.
  * @example
  *
- *      R.evolve({ elapsed: R.add(1), remaining: R.add(-1) }, { name: 'Tomato', elapsed: 100, remaining: 1400 }); //=> { name: 'Tomato', elapsed: 101, remaining: 1399 }
+ *      var tomato  = {firstName: '  Tomato ', elapsed: 100, remaining: 1400};
+ *      var transformations = {
+ *        firstName: R.trim,
+ *        lastName: R.trim, // Will not get invoked.
+ *        data: {elapsed: R.add(1), remaining: R.add(-1)}
+ *      };
+ *      R.evolve(transformations, tomato); //=> {firstName: 'Tomato', data: {elapsed: 101, remaining: 1399}}
  */
 module.exports = _curry2(function evolve(transformations, object) {
-  return _extend(_extend({}, object), mapObjIndexed(function(fn, key) {
-    return fn(object[key]);
-  }, transformations));
+  var transformation, key, type, result = {};
+  for (key in object) {
+    transformation = transformations[key];
+    type = typeof transformation;
+    result[key] = type === 'function' ? transformation(object[key])
+                : type === 'object'   ? evolve(transformations[key], object[key])
+                                      : object[key];
+  }
+  return result;
 });
--- a/test/evolve.js
+++ b/test/evolve.js
@@ -11,18 +11,26 @@ describe('evolve', function() {
     assert.deepEqual(R.evolve(transf, object), expected);
   });
 
-  it('invokes function with undefined if object does not contain the key', function() {
-    var transf   = {wow: String};
-    var object   = {};
-    var expected = {wow: 'undefined'};
+  it('does not invoke function if object does not contain the key', function() {
+    var transf   = {n: R.add(1), m: R.add(1)};
+    var object   = {m: 3};
+    var expected = {m: 4};
     assert.deepEqual(R.evolve(transf, object), expected);
   });
 
   it('is not destructive', function() {
     var transf   = {elapsed: R.add(1), remaining: R.add(-1)};
     var object   = {name: 'Tomato', elapsed: 100, remaining: 1400};
+    var expected = {name: 'Tomato', elapsed: 100, remaining: 1400};
     R.evolve(transf, object);
-    assert.deepEqual(object, {name: 'Tomato', elapsed: 100, remaining: 1400});
+    assert.deepEqual(object, expected);
+  });
+
+  it('is recursive', function() {
+    var transf   = {nested: {second: R.add(-1), third: R.add(1)}};
+    var object   = {first: 1, nested: {second: 2, third: 3}};
+    var expected = {first: 1, nested: {second: 1, third: 4}};
+    assert.deepEqual(R.evolve(transf, object), expected);
   });
 
   it('is curried', function() {


src/evolve.js
```diff
@@ -17,10 +18,22 @@ var mapObjIndexed = require('./mapObjIndexed');
...
-  return _extend(_extend({}, object), mapObjIndexed(function(fn, key) {
-    return fn(object[key]);
-  }, transformations));
+  var transformation, key, type, result = {};
+  for (key in object) {
+    transformation = transformations[key];
+    type = typeof transformation;
+    result[key] = type === 'function' ? transformation(object[key])
+                : type === 'object'   ? evolve(transformations[key], object[key])
+                                      : object[key];
```
davidchambers: Why do we support this case? It seems to me every leaf node of `transformations` should be a function. One should use `{foo: R.always(42)}` rather than `{foo: 42}` if one wishes to transform the value of the `foo` property to `42` in every case.

paldepind: The code says "if it's a function, apply it. If it's an object recurse into it. If it's anything else ignore it and copy the old value". It sounds like that is the behavior you want?

paldepind:
This closes #1130.

src/evolve.js
CrossEye: I don't see a good alternative to this behavior, except to throw an exception.  @davidchambers, would you suggest that, or do you see another alternative?

CrossEye:
Absolutely agree on the rationale for the change to `evolve`.  And I agree that this is a low-risk change.

CrossEye:
:herb:

src/evolve.js
davidchambers: I was reading `: object[key]` as `: transformations[key]`. Now that I'm interpreting the code correctly, it LGTM. :)

test/evolve.js
```diff
@@ -11,18 +11,26 @@ describe('evolve', function() {
...
-    assert.deepEqual(object, {name: 'Tomato', elapsed: 100, remaining: 1400});
+    assert.deepEqual(object, expected);
+  });
+
+  it('is recursive', function() {
+    var transf   = {nested: {second: R.add(-1), third: R.add(1)}};
+    var object   = {first: 1, nested: {second: 2, third: 3}};
+    var expected = {first: 1, nested: {second: 1, third: 4}};
+    assert.deepEqual(R.evolve(transf, object), expected);
   });
```
davidchambers: Should we add a test to verify that `R.evolve` also recurses when it encounters an array?

CrossEye: Ugh.  JS strikes again!

Obviously the implementation will do so.  But is it really what we want?  So we have to then check that the object and transformation both have an array, and both have the same length?  Meh!

paldepind: Recursing into arrays was not intentional. But with the current implementation you could in theory modify an object with numbered keys by supplying an array. Modifying an array though doesn't really work since the result is always copied into an object.

CrossEye:
Seeing all this discussion makes me interested in [another version][e2] of `evolve`:

```js
evolve2 :: {k: (v -> v -> v)} -> v -> {k: v} -> {k: v}

var transforms = {
  min: (val, curr) => val < curr ? val : curr,
  max: (val, curr) => val > curr ? val : curr,
  count: (val, curr) => curr + 1,
  mean: (val, curr, obj) => (obj.mean * obj.count + val) / (obj.count + 1)
};

var update = evolve2(transforms);

var obj = {
  min: Infinity,
  max: -Infinity,
  count: 0,
  mean: 0
};

var o1 = update(6, obj); //=> {"count": 1, "max": 6, "mean": 6, "min": 6}
var o2 = update(11, o1); //=> {"count": 2, "max": 11, "mean": 8.5, "min": 6}
var o3 = update(5, o2);  //=> {"count": 3, "max": 11, "mean": 7.333333333333333, "min": 5}
var o4 = update(15, o3); //=> {"count": 4, "max": 15, "mean": 9.25, "min": 5}
var o5 = update(8, o4);  //=> {"count": 5, "max": 15, "mean": 9, "min": 5}
```

This is not meant as a real proposal at the moment, just some musings.  I haven't really even used the original `evolve` much.  But this version looks quite useful.

  [e2]: http://bit.ly/1QnpmPJ

paldepind:
@CrossEye 

One could do it like this:

```javascript
var update = function(val, obj) {
  return R.evolve({
    min: (curr) => val < curr ? val : curr,
    max: (curr) => val > curr ? val : curr,
    count: (curr) => curr + 1,
    mean: (curr) => (obj.mean * obj.count + val) / (obj.count + 1)
  }, obj);
};
```

CrossEye:
@paledpind, yes, but that wouldn't allow one to do

```js
var update = evolve2(transformers, val);
var newObjs = map(update, oldObjs);
```

As I said, this is not a serious proposal for Ramda right now.  But it's an interesting extension to the notion of `evolve`.

paldepind:
Why wouldn't it allow that? You could curry `update` and do:

```javascript
var newObjs = map(update(val), oldObjs);
```

CrossEye:
> Why wouldn't it allow that? You could curry `update` and do:
> 
> ```javascript
> var newObjs = map(update(val), oldObjs);
> ```

But then if you wanted to use a new set of transformers, in my API, you would just do

```js
var update2 = evolve2(transformers2, val);
var newObj = map(update2, oldObj);
```

With your suggestion, you would have to write

```js
var update2 = function(val, obj) {
  return R.evolve(/*
    inline version of transformers2 here
  */, obj);
};
var newObj = map(update2, oldObj);
```

It just seems to me a matter of a cleaner API.

paldepind:
Indeed. You'd have to wrap each invocation of `evolve` in function. But I really don't think that is much to do. That method also extends nicely to more arguments. With `evolve2` you're out of luck if you want to pass two arguments to the transformer functions.

buzzdecafe:
> you're out of luck if you want to pass two arguments to the transformer functions.

`lift`?

paldepind:
@buzzdecafe How? Truth be told I really don't get `lift`. It's certainly not the type of lift I'm used to.

CrossEye:
> Indeed. You'd have to wrap each invocation of `evolve` in function.

It's not just that.  Because the functions inside there have free variables bound to the `update` scope, you cannot use function references.  Mine was just hacked together, and not particularly thought out.  Perhaps the internal signatures should look more like `reduce` callback ones and get `curr` first, which would allow, for instance:

```js
var transforms = {
  min: R.binary(Math.min),
  max: R.binary(Math.max),
  count: R.inc,
  mean: (curr, val, obj) => (obj.mean * obj.count + val) / (obj.count + 1)
};
```

The question about multiple arguments is interesting, but that can be asked about all sorts of functions.  Adding a first new parameter to a function has much greater consequence to my mind than adding additional ones, especially when that first parameter can always be a structured object.

-----

I'm not sure where @buzzdecafe was going with `lift`.  I don't see where it would help.  But my understanding of `lift` is growing.  If you're curious, I think of it as a generalization of `map`.  `map` in its general sense takes a function operating on values, and then a wrapper for those values, and creates a wrapper for the result of applying that function to those values.  So `map(square, Maybe(5); //=> Maybe(25)` in much the same way `map(square, [1, 2, 3]; //=> [1, 4, 9]`. 

But `map` takes a function from one value to another.  What if you had a two-parameter function, like `add`?  This is `lift2`:  `lift2(add)(Maybe(3), Maybe(5)); //=> Maybe(8)`.  It lifts the function that operates on the plain number values to one that operates on containers.  `lift3` would work for three parameters.  `lift1` is just `map`.  At the moment, we just do `lift` and get the length from the function, and offer `liftN` as we do with `curryN`.

That's my understanding, anyway.

paldepind:
I agree with you on what you said about `evolve2`. Especially the relation to `reduce`. One could see the relationship between `evolve` and `evolve2` a bit like the relationship between `map` and `reduce`.

Everything you're describing sounds like `lift` as I understand it. Lift and map are the same. The main thing that confused me was how it works on arrays. I'd have expected something like this:

```javascript
R.lift(R.add)([1,2,3], [4,5,6]); // => [5,7,9]
```

But I actually do remember now that lift on lists generates all the permutations of the function on the list. It probably makes sense somehow. It's just not that intuitive to me. Forget what I said about lift not being the type of lift I'm used to.

Issue #1077, opened 2015-05-11
Consider implementing uniq (and union) on top of native Set, if available


asaf-romano:
ES6 includes a native Set object, that has the proper "has" complexity you'd want for uniq implementation.

Here's a quick & dirty performance compression between ramda's loop-based uniq to a uniq implementation on top of Set - it only works in Firefox because I'm using Array.from.

http://bit.ly/1F8awKZ

For 350 unique objects, each repeated 1000 times, I get 0.15 native : non-native ratio.

Note that Array.from isn't essential for this proposal, and that Set (unlike Array.from) is available in most environments.

asaf-romano:
The ratio depends mostly on the number of unique objects. If you try it with 4 objects (no matter how many times they're repeated) the ratio is at about 0.8.

CrossEye:
This is much more convincing than the recent example on Gitter.

It might be made a little better with `R.of`: http://bit.ly/1H9rfNC.  This now works in Chrome as well.

I don't think we want to abandon non-es6 environments, so if we choose to do this, it would have to use some sort of testing for the existence of `Set`.  We've done it before (`keys` comes to mind) but I'm loathe to make a habit out of it.  These numbers are pretty convincing, though.

asaf-romano:
Yes, it took me a moment to realize that if all four elements appear at the top of the array, it's surprising that there was any difference to begin with.

Testing for Set existence is one option. Another is some sort of polyfill that is used unconditionally.

buzzdecafe:
:+1: for using Set and backfilling for older environments. 

http://kangax.github.io/compat-table/es6/#Set

asaf-romano:
@CrossEye Regarding R.of, it doesn't do what Array.from does; it simply wraps it's 1st arg in an array. Array.from creates an array out of an iterable.

asaf-romano:
R.into([], R.map(R.identity), new Set([1,1,3,4])) works though.

CrossEye:
I'm definitely convinced that this is something we should do.  I would love to see a PR for it.

asaf-romano:
It was only after I implemented this that I realized this cannot be supported (from tests/uniq.js):
assert.strictEqual(R.uniq([new Just([42]), new Just([42])]).length, 1);

asaf-romano:
So for further discussion, here's what I did:
https://github.com/asaf-romano/ramda/commit/44c011e76b34a785d2e9cd2443ed23c1bc21c385

[...]

PR #1142, opened 2015-06-01
R.uniq implementation to use Set if available

scott-christopher:
Implementation for https://github.com/ramda/ramda/issues/1077

Below is a simple benchmark (take with a grain of salt) of the time taken to run `R.uniq` 100 times, comparing master branch with the branch of this PR.

When called with `R.flatten(R.repeat([{}, 0, new Just(5)], size))` (many dupes):

branch | time (ms) for 1000 items | time (ms) for 10000 items
--------- | -------------- | ---------------
master | 700 | 7000
uniq-set | 600 | 5900

When called with `R.map(function (x) { return x(); }, R.repeat(function () { return {}; }, size))` (no dupes):

branch | time (ms) for 10000 items | time (ms) for 100000 items
--------- | -------------- | ---------------
master | 750 | 7500
uniq-set | 75 | 1700

ping @asaf-romano

### DIFF
--- a/src/uniq.js
+++ b/src/uniq.js
@@ -1,3 +1,5 @@
+var _contains = require('./internal/_contains');
+var _curry1 = require('./internal/_curry1');
 var equals = require('./equals');
 var uniqWith = require('./uniqWith');
 
@@ -18,4 +20,39 @@ var uniqWith = require('./uniqWith');
  *      R.uniq([1, '1']);     //=> [1, '1']
  *      R.uniq([[42], [42]]); //=> [[42]]
  */
-module.exports = uniqWith(equals);
+module.exports = (function() {
+  /**
+   * Uses a native `Set` instance where possible for
+   * removing duplicate items. Items that implement
+   * the fantasy-land Setoid spec fallback to using
+   * `_contains` to support custom equality behaviour.
+   */
+  function uniq(list) {
+    /* global Set */
+    var item,
+        set = new Set(),
+        idx = -1,
+        len = list.length,
+        items = [],
+        uniqs = [];
+
+    while (++idx < len) {
+      item = list[idx];
+      // `_contains` is also used to differentiate between
+      // +0 and -0, as the native Set does not.
+      if (item === 0 || (item != null && typeof item.equals === 'function')) {
+        if (!_contains(item, items)) {
+          items.push(item);
+          uniqs.push(item);
+        }
+      } else {
+        if (set.size !== set.add(item).size) {
+          uniqs.push(item);
+        }
+      }
+    }
+    return uniqs;
+  }
+
+  return typeof Set !== 'function' ? uniqWith(equals) : _curry1(uniq);
+})();
--- a/test/uniq.js
+++ b/test/uniq.js
@@ -27,4 +27,8 @@ describe('uniq', function() {
     assert.strictEqual(R.uniq([NaN, NaN]).length, 1);
     assert.strictEqual(R.uniq([new Just([42]), new Just([42])]).length, 1);
   });
+
+  it('handles null and undefined elements', function() {
+    assert.deepEqual(R.uniq([void 0, null, void 0, null]), [void 0, null]);
+  });
 });


paldepind:
This is a much simpler implementation:

```javascript
function uniq(list) {
  return Array.from(new Set(list));
}
```

This basically converts the array to a set and then the set to an array. It doesn't handle the fantasy land `equals` though.

I think a better idea would be to implement `uniqWith` on top of Map and then keep the current definition of `uniq`. That should give performance improvements to both `uniq` and `uniqWith`. It would also provide an implementation that doesn't go O(n^2) on arrays of fantasy land Setoids.

A Map only implementation could look like this:

```javascript
function uniqWith(fn, list) {
  var map = new Map(_map(function(item) { return [fn(item), item]; }, list));
  return Array.from(map.values());
}
```

Which is still much shorter!

scott-christopher:
Your `uniq` example is essentially what @asaf-romano came up with in #1077, though hit turbulence with the tests that compare dispatching behaviour with `R.equals`.

The `uniqWith` implementation using `Map` looks nice and succinct, though if I understand correctly it requires a unary hashing function as its first argument rather than a binary comparison function, so `R.uniqWith(R.equals)` won't work as expected.

paldepind:
Arg! You're right. I forgot that `uniqWith` takes a comparison function. I was thinking that it took a function that returned a value on which to determine uniqueness. Well, then I don't see a better solution that the current.

src/uniq.js
```diff
@@ -18,4 +20,39 @@ var uniqWith = require('./uniqWith');
...
+  /* global Set */
+  var item, set, idx, len, items, uniqs;
+
+  /**
+   * Uses a native `Set` instance where possible for
+   * removing duplicate items. Items that implement
+   * the fantasy-land Setoid spec fallback to using
+   * `_contains` to support custom equality behaviour.
+   */
+  if (typeof Set === 'function') {
```
buzzdecafe: we shouldn't test for `Set` on every invocation. Let's detect it at initialization and construct the appropriate function instead.

scott-christopher: I originally tried that with something like the following, though there is something whacky with the build tool, where it doesn't include the separate `uniq` function in `dist/ramda.js` (and also pulls all inline comments out above).
```js
function uniq(list) {/*...*/}
module.exports = typeof Set === 'function' ? _curry1(uniq) : uniqWith(equals);
```
results in `dist/ramda.js` only containing:
```js
    /* global Set */
    /**
       * Uses a native `Set` instance where possible for
       * removing duplicate items. Items that implement
       * the fantasy-land Setoid spec fallback to using
       * `_contains` to support custom equality behaviour.
       */
    // -0 must be handled separately, as Set treats 0 and -0 as equal.
    /**
     * Returns a new list containing only one copy of each element in the original list.
     * `R.equals` is used to determine equality.
     *
     * @func
     * @memberOf R
     * @category List
     * @sig [a] -> [a]
     * @param {Array} list The array to consider.
     * @return {Array} The list of unique items.
     * @example
     *
     *      R.uniq([1, 1, 2, 1]); //=> [1, 2]
     *      R.uniq([1, '1']);     //=> [1, '1']
     *      R.uniq([[42], [42]]); //=> [[42]]
     */
    var uniq = typeof Set === 'function' ? _curry1(uniq) : uniqWith(equals);
```

Is there a "correct" way for handling this?

davidchambers: You'll need an IIFE:

```javascript
  var uniq = function(...) { ... };
  return typeof Set === 'function' ? _curry1(uniq) : uniqWith(equals);
}());
```

See [__trim.js__][1] for an example of a module which uses this approach.


[1]: https://github.com/ramda/ramda/blob/v0.14.0/src/trim.js

scott-christopher: Thanks, that did the trick and I've updated the PR.

On a side note, is that something where there would be interest in improving in the build script, or is the behaviour intentional? I'd be happy to take a look.

davidchambers: The behaviour is intentional, though there may be a better approach. The reason we disallow top-level assignments (aside from the module imports) is that they could conflict. If two modules were both to define a top-level `x` variable, we could end up with the following output:

```javascript
var x = ...;  // from src/f.js

var f = function(...) { ... };

var x = ...;  // from src/g.js

var g = function(...) { ... };
```

We could wrap each module in an IIFE in __dist/ramda.js__, but at least at the time I wrote the build script the readability of the generated file was considered important.

CrossEye: I was writing a similar build script to modularize our files at the same time David did his, and my output was much uglier because I ended up with those IIFEs.  I took forever to build mine because I spent all sorts of effort trying to determine the cases where I could skip these IIFEs.  David's approach has been much simpler.  And there have only been a handful of cases where it's caused us any difficulty at all.

scott-christopher: Yup, that all makes sense. Would there be any objection to me updating the build script to emit a warning if it detects "detached" top-level variables/functions?

buzzdecafe: sounds good to me

buzzdecafe:
:cow2:

src/uniq.js
```diff
@@ -18,4 +20,37 @@ var uniqWith = require('./uniqWith');
...
+    var item,
+        set = new Set(),
+        idx = -1,
+        len = list.length,
+        items = [],
+        uniqs = [];
+
+    while (++idx < len) {
+      item = list[idx];
+      if (item === -0 || typeof item.equals === 'function') {
```
davidchambers: Why `=== -0` rather than `=== 0`? The `===` considers zero and negative zero equal.

src/uniq.js
```diff
@@ -18,4 +20,37 @@ var uniqWith = require('./uniqWith');
...
+  function uniq(list) {
+    /* global Set */
+    var item,
+        set = new Set(),
+        idx = -1,
+        len = list.length,
+        items = [],
+        uniqs = [];
+
+    while (++idx < len) {
```
thefourtheye: This will not work if the `list` is a sparse array. What is the desired behavior in that case?

paldepind: Ramda generally treats sparse arrays as dense.

src/uniq.js
thefourtheye: What if rest of the items are of different type and one of the elements is zero? The result will not include zero then.

src/uniq.js
thefourtheye: Then in this particular case we will introduce `undefined` in the result. This will become inconsisten with `indexOf` result. Will that be okay?

CrossEye: Ramda's general attitude is that if you feed us sparse arrays where lists are expected, all bets are off.  Things may work fine, and you'll be happy with the results.  On the other hand, Ramda may corrupt your database, crash your car, steal your boyfriend, or insult your _latke_ recipe.  We take no responsibility whatsoever for its behavior if you can't be bothered to pass it a well-behaved list.

Ramda would really like to be dealing with lists, but since the languages doesn't support true lists, we make do with arrays as substitutes.  However,we simply assume that these are list-like dense arrays.

thefourtheye: Cool then. If we capture this information clearly in the official documentation, then it would be good.

src/uniq.js
scott-christopher: @davidchambers: Good catch. The intent was to allow `+0` in the Set, while kicking the `-0` over to the `items` array where they can treated as unique numbers. I've updated it to actually test for `0` and `-0` now. It was co-incidentally still passing the tests because it was falling through and delegating to `R.equals` for comparing `0` and `-0`.

@thefourtheye: I'm not sure I understand. Both paths end up building up the unique items in `uniqs`, regardless of whether `Set` or `_contains` is used.
```
> var x = {}
> uniq([x, 'a', x, 'b', x, 'a', x, 'b', 0])
[ {}, 'a', 'b', 0 ]
```

src/uniq.js
```diff
@@ -18,4 +20,37 @@ var uniqWith = require('./uniqWith');
...
+      item = list[idx];
+      if ((item === 0 && 1 / item === -Infinity) || typeof item.equals === 'function') {
```
davidchambers: Could we simplify this check to `item === 0`? It seems fine to take the slow path for both `0` and `-0`.

scott-christopher: I don't mind either way. I've pushed up a change that performs the `item === 0` check instead and added some comments along with it.

davidchambers: :thumbsup: The comment is very useful.

src/uniq.js
```diff
@@ -18,4 +20,39 @@ var uniqWith = require('./uniqWith');
...
+        idx = -1,
+        len = list.length,
+        items = [],
+        uniqs = [];
+
+    while (++idx < len) {
+      item = list[idx];
+      // `_contains` is also used to differentiate between
+      // +0 and -0, as the native Set does not.
+      if (item === 0 || typeof item.equals === 'function') {
```
davidchambers: Do we need to replace

```javascript
typeof item.equals === 'function'
```

with

```javascript
item != null && typeof item.equals === 'function'
```

to avoid throwing a type error when `item` is null or undefined? If so, we should include a test for this.

scott-christopher: Good catch. :fishing_pole_and_fish: I've updated the check and added a test case to support it.

davidchambers:
:deciduous_tree:

PR #1143, opened 2015-06-01
put build instructions lost in CONTRIBUTING.md to README.md

trueproof:
In fresh master `mocha` throwed, `R.addIndex` was missing because __dist/ramda.js__ was outdated. So i was searching for info on updating the build and only could find it here: https://github.com/ramda/ramda/pull/899

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,12 @@ These script tags add the variable `ramda` on the browser's global scope.
 
 Or you can inject ramda into virtually any unsuspecting website using [the bookmarklet](BOOKMARKLET.md).
 
-### Partial Builds
+### Build
+
+* on Unix-based platforms, `npm run build` updates __dist/ramda.js__ and __dist/ramda.min.js__
+* on Windows, write the output of `scripts/build --complete` to a temporary file, then rename the temporary file __dist/ramda.js__.
+
+#### Partial Builds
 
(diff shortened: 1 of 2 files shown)


buzzdecafe:
yes, this makes sense. we should go even further and have an `npm build` task

trueproof:
Included "npm run build" for UNIX-like systems into commit. Don't know exactly, if it's possible to do it os-independently.

README.md
```diff
@@ -110,7 +110,12 @@ These script tags add the variable `ramda` on the browser's global scope.
 
 Or you can inject ramda into virtually any unsuspecting website using [the bookmarklet](BOOKMARKLET.md).
 
-### Partial Builds
+### Build
+
+* on Unix-based platforms, `npm run build` updates __dist/ramda.js__ and __dist/ramda.min.js__
+* on Windows, write the output of `scripts/build --complete` to a temporary file, then rename the file as __dist/ramda.js__.
```
davidchambers: The word "as" doesn't seem right to me. Rather than

> […] rename the file as __dist/ramda.js__.

let's write

> […] rename the temporary file __dist/ramda.js__.

trueproof:
I had doubts, found this, they say both are correct: http://forum.wordreference.com/threads/renamed-as-renamed-to-or-just-renamed.2177764/#post-10904616. Updated according to your preference.

davidchambers:
:deciduous_tree: Thanks!

buzzdecafe:
:cow2:

PR #1144, opened 2015-06-01
deprecate R.invoker

davidchambers:
We needn't provide both [`R.invoke`][1] and [`R.invoker`][2]: these functions have similar roles (despite having very different types).

[1]: http://ramdajs.com/docs/#invoke
[2]: http://ramdajs.com/docs/#invoker

### DIFF
--- a/src/invoker.js
+++ b/src/invoker.js
@@ -18,6 +18,7 @@ var curryN = require('./curryN');
  *        before the target object.
  * @param {Function} method Name of the method to call.
  * @return {Function} A new curried function.
+ * @deprecated since v0.15.0
  * @example
  *
  *      var sliceFrom = R.invoker(1, 'slice');


paldepind:
Why not deprecate `R.invoke` instead? `R.invoker` create much nicer and curried functions.

Say you want to work with `Array.prototype.slice` as a function:

```javascript
var slice = R.invoker(2, 'slice');
var sliceFromStart = slice(0);
var sliceFrom2 = slice(2);
```
`R.invoke` can't do the above.

asaf-romano:
I'm using invoker this way too for "functionalizing" *.prototytype.* methods (e.g. Map.prototype.get)

davidchambers:
Here's an example of switching from `R.invoke` to `R.invoker`:

```diff
 const getLargeCities = R.pipe(
   cheerio.load,
   R.partialRight(R.call, '#United_States'),
-  R.invoke('parent', []),
-  R.invoke('nextAll', ['table.sortable']),
-  R.invoke('first', []),
-  R.invoke('find', ['tr']),
-  R.invoke('toArray', []),
+  R.invoker(0, 'parent'),
+  R.invoker(1, 'nextAll')('table.sortable'),
+  R.invoker(0, 'first'),
+  R.invoker(1, 'find')('tr'),
+  R.invoker(0, 'toArray'),
   R.tail,
   R.map(cheerio),
-  R.map(R.invoke('children', [])),
+  R.map(R.invoker(0, 'children')),
   R.map(R.converge(R.concat,
                    R.pipe(R.converge(R.merge,
-                                     R.pipe(R.invoke('eq', ['1']),
-                                            R.invoke('text', []),
+                                     R.pipe(R.invoker(1, 'eq')('1'),
+                                            R.invoker(0, 'text'),
                                             R.replace(/\[\d+\]$/, ''),
                                             R.toUpper,
                                             R.createMapEntry('city')),
-                                     R.pipe(R.invoke('eq', ['2']),
-                                            R.invoke('text', []),
+                                     R.pipe(R.invoker(1, 'eq')('2'),
+                                            R.invoker(0, 'text'),
                                             R.createMapEntry('state'))),
                           R.of),
                    R.pipe(R.nth(4),
```

Is it correct to state that everything that can be expressed in terms of `R.invoke` can be expressed in terms of `R.invoker`, but that the reverse is not true? If so, let's deprecate `R.invoke` instead. :)

paldepind:
> Is it correct to state that everything that can be expressed in terms of R.invoke can be expressed in terms of R.invoker, but that the reverse is not true?

It seems so to me.

Btw. `R.invoker` actually has an undocumented feature. Instead of `R.invoker(1, 'find')('tr')` one can do `R.invoker(1, 'find', 'tr')`. I.e. initial arguments can be supplied initially in an invariadic way. I kinda like the feature but I've also previously suggested removing it. I think your example proved that one can do just fine without it.

CrossEye:
> Btw. `R.invoker` actually has an undocumented feature. 

I might call that "underdocumented" instead, as one of the examples uses this behavior.

I like the simplicity of `invoke`.  What can `invoker` do that can't be done with `invoke`?  I know that older versions of `invoker` had significantly different behavior, but since we switched to a name-based API, I don't know what differences there are.

paldepind:
> I might call that "underdocumented" instead, as one of the examples uses this behavior.

Yes. I had forgot about that. I was looking at the function signature only.

>  What can invoker do that can't be done with invoke?

`invoker` creates curried functions with proper arity. `invoke` creates odd array taking functions. Continuing my last example. Compare:

```javascript
var slice = R.invoker(2, 'slice');
slice(2, 4, list);
```

Versus:

```javascript
var slice = R.invoke('slice');
slice([2, 4], list);
```

Which `slice` seems nicest? The one with arity 3 that is curried? Or the one with arity 2 that takes two arguments in an array and thus is both unnatural to use and very tricky to partially apply?

CrossEye:
Sold!

I kept seeing `invoke` more fully applied.

I do prefer `invoker` now that it's put this way.

davidchambers:
Superseded by #1146

PR #1147, opened 2015-06-02
Remove support for supplying initial arguments to `R.invoker`

paldepind:
This makes `R.invoker` invariant. Which simplifies the implementation and makes the signature comprehensive/correct.

This is a breaking change. But the feature was only half documented before and the upgrade path is straight forward: `R.invoker(2, 'slice', 6)` to `R.invoker(2, 'slice')(6)`.

I think the current behavior has it's benefits. But in the end I think it does more harm than good.

### DIFF
--- a/src/invoker.js
+++ b/src/invoker.js
@@ -1,5 +1,5 @@
+var _curry2 = require('./internal/_curry2');
 var _slice = require('./internal/_slice');
-var curry = require('./curry');
 var curryN = require('./curryN');
 
 
@@ -7,14 +7,14 @@ var curryN = require('./curryN');
  * Turns a named method with a specified arity into a function
  * that can be called directly supplied with arguments and a target object.
  *
- * The returned function is curried and accepts `len + 1` parameters where
+ * The returned function is curried and accepts `arity + 1` parameters where
  * the final parameter is the target object.
  *
  * @func
  * @memberOf R
  * @category Function
- * @sig (Number, String) -> (a... -> c -> b)
- * @param {Number} len Number of arguments the returned function should take
+ * @sig Number -> String -> (a -> b -> ... -> n -> Object -> *)
+ * @param {Number} arity Number of arguments the returned function should take
  *        before the target object.
  * @param {Function} method Name of the method to call.
  * @return {Function} A new curried function.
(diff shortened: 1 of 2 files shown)


src/invoker.js
```diff
@@ -7,14 +7,14 @@ var curryN = require('./curryN');
...
  * that can be called directly supplied with arguments and a target object.
  *
- * The returned function is curried and accepts `len + 1` parameters where
+ * The returned function is curried and accepts `arity + 1` parameters where
  * the final parameter is the target object.
  *
  * @func
  * @memberOf R
  * @category Function
  * @sig (Number, String) -> (a... -> c -> b)
```
davidchambers: I don't understand this signature. Shouldn't it be something like `Number -> String -> (a -> b -> ... -> n -> Object -> *)`?

src/invoker.js
```diff
@@ -22,15 +22,13 @@ var curryN = require('./curryN');
...
-  var initialArgs = _slice(arguments, 2);
-  var len = arity - initialArgs.length;
-  return curryN(len + 1, function() {
-    var target = arguments[len];
-    var args = initialArgs.concat(_slice(arguments, 0, len));
+module.exports = _curry2(function invoker(arity, method) {
+  return curryN(arity + 1, function() {
+    var target = arguments[arity];
+    var args = _slice(arguments, 0, arity);
     return target[method].apply(target, args);
```
davidchambers: We could remove the `args` variable.

src/invoker.js
CrossEye: Yes, that's probably the best we can do, although even '`Object`' might be replaced with '`*`'.  This is as weakly typed as I think we get in Ramda.  Give us a number and a string, and we'll give you back a function from that many parameters plus some specific something, returning something else.

... Sure, dude.

However, it's also quite useful.

CrossEye:
:+1:

src/invoker.js
paldepind: @davidchambers I agree. That is a much better signature.
@CrossEye We'll invoke a method on the last parameter so it has to be an object.

src/invoker.js
paldepind: I almost did it ;) But you're right. Readability is probably not harmed by doing so.

paldepind:
I've updated the signature and removed the `args` variable.

CrossEye:
:herb:

paldepind:
First merge :cold_sweat:

buzzdecafe:
:tada:

CrossEye:
:balloon:

Issue #1133, opened 2015-05-29
evolveWhere


ccorcos:
Here's a useful function I've been using a bunch when a changes is made to a collection:

```coffee
evolveWhere = R.curry (where, evolution, list) ->
  R.map(R.cond([
    [where, evolve(evolution)]
    [R.T, R.identity]
  ]), list)
```

You'd use it like this:

```coffee
data = [{id:1, name:'chet', score:10}, {id:2, name:'joe', score:50}]
# change the name where id:1
evolveWhere(R.propEq('id', 1), {name:'chet'}, data)
# increment the score where name:'joe'
evolveWhere(R.propEq('name', 'joe'), {score: R.add(10)}, data)
```

Do you think this is a worthy candidate?

paldepind:
I've been thinking of something similar myself. But I think `evolveWhere` is two specific. It's also confusing since evolve is for objects and `evolveWhere` would be for arrays.

My suggestion is `updateWhere`. Kinda like this:

```javascript
function updateWhere(pred, fn, list) {
  return R.map(function(elm) { return pred(elm) === true ? fn(elm) : elm; });
}
```

I'm not sure about the name though since `R.update` doesn't take a modification function but a constant. But actually `R.update` would be move powerful if it took a modification function. Then the current behavior could be achieved with `R.always`.

ccorcos:
I entirely agree :+1:

buzzdecafe:
> But actually `R.update` would be move powerful if it took a modification function. 

:+1:

paldepind:
Oops. That is `R.adjust`. Then the function I suggested above should be `R.adjustWhere`.

ccorcos:
Seems to me like `R.update` and `R.adjust` should take a predicate function and an evolve function.

davidchambers:
This problem could be solved using [lenses][1]:

```javascript
var data = [{id: 1, name: 'amy', score: 10},
            {id: 2, name: 'joe', score: 50}];

var transformIf = R.ifElse(_, _, R.identity);

var f = transformIf(R.propEq('id', 1),
                    R.set(R.lensProp('name'), 'chet'));
R.map(f, data);
// => [{id: 1, name: 'chet', score: 10}, {id: 2, name: 'joe', score: 50}]

var g = transformIf(R.propEq('name', 'joe'),
                    R.over(R.lensProp('score'), R.add(10)));
R.map(g, data);
[...]

PR #1153, opened 2015-06-04
Deprecate `update`

paldepind:
`R.update` is just `R.adjust` with an `R.always` on the first parameter. It seems quite unnecessary and [confusing](https://github.com/ramda/ramda/issues/1133#issuecomment-106910074).

I originally found `update` and didn't think there'd be another function this similar to it so I wrongly assumed only the more limited `update` existed. Probably because it seems to me like Ramda generally does not supply functions that are just an `always` away.

### DIFF
--- a/src/update.js
+++ b/src/update.js
@@ -14,6 +14,7 @@ var always = require('./always');
  * @param {*} x The value to exist at the given index of the returned array.
  * @param {Array|Arguments} list The source array-like object to be updated.
  * @return {Array} A copy of `list` with the value at index `idx` replaced with `x`.
+ * @deprecated since v0.15.0
  * @example
  *
  *      R.update(1, 11, [0, 1, 2]);     //=> [0, 11, 2]


davidchambers:
:deciduous_tree:

CrossEye:
Ping @scott-christopher.

scott-christopher:
The inclusion of `update` was to provide an analog of `assoc` for lists. Perhaps the confusion could be better managed by simply including an `@see` in the docs for both `update` and `adjust`?

If there is a goal to begin trimming back the number of functions offered in Ramda, then sure, it is superfluous given that someone could easily create it themselves, though I would have thought a functional equivalent of `a[i] = b` was a pattern common enough to be worthy of a helper. I guess the same question could also be asked of all the other functions built on top of their more general `*With` form.

That said, I don't mind if this is included or removed as it is simple enough to implement myself where needed, so I'm happy to go with the consensus here.

paldepind:
I think those are great reasons. Both regarding `assoc` and `a[i] = b`. Keeping it and adding a `@see` would be fine by me.

But there is also value in just having a single function that is powerful enough to express the other with `always`. Is there a variant of `assoc` that takes a function?

davidchambers:
> Is there a variant of `assoc` that takes a function?

Not as far as I know. `assoc` can be used to update an existing key–value pair or to add a new key–value pair. If it were to take a function the function would need to accept `undefined`, which feels dirty to me.

paldepind:
I'm no longer of the opinion that this is necessary.

PR #1155, opened 2015-06-05
Simplify curryN

paldepind:
I've always felt that there was something off with `R.curryN` but my last stab at it resulted in only minor modifications.

This PR introduces an internal `_curryN` as a base for the external `curryN`. The internal function has a slightly different signature. The new internal function is half the size of the old `curryN` and completely avoids using `_slice`.

I've also used `for` loops just to provocate. I think the idea of always using `while` loops is silly. It's like always using `reduce` and forbidding `map`. And the kind of "consistency" that is achieved by doing so is just as harmful. Code becomes less expressive and harder to read and understand because you're using a more general construct when a more specific one would capture the desired control flow better. 

Here is the new `_curryN` written with while loops:

```javascript
module.exports = function _curryN(length, recieved, fn) {
  return function() {
    var combined = [];
    var combinedIdx = -1; // What! Minus one? But that's not really where we begin indexing
    var argsIdx = 0;
    var left = length;
    while (++combinedIdx < recieved.length) { // Stuffing increment and condition check into one statement :( :( :(
      var val = recieved[combinedIdx];
      combined[combinedIdx] = val === __ && argsIdx < arguments.length ? arguments[argsIdx++] : val;
      if (combined[combinedIdx] !== __) { --left; }
    }
    while (argsIdx < arguments.length) {
      combined[combinedIdx] = arguments[argsIdx];
      if (combined[combinedIdx] !== __) { --left; }
      combinedIdx++; argsIdx++; // Huh!? For-loops have a well defined location for such increments
    }
    return left <= 0 ? fn.apply(this, combined) : arity(left, _curryN(length, combined, fn));
  };
});
```

I will change this PR to use `while` loops if requested to do so :)

### DIFF
--- a/src/internal/_curryN.js
+++ b/src/internal/_curryN.js
@@ -1,0 +1,38 @@
+var arity = require('../arity');
+
+
+/**
+ * Internal curryN function.
+ *
+ * @private
+ * @category Function
+ * @param {Number} length The arity of the curried function.
+ * @return {array} An array of arguments received thus far.
+ * @param {Function} fn The function to curry.
+ */
+module.exports = function _curryN(length, received, fn) {
+  return function() {
+    var combined = [];
+    var argsIdx = 0;
+    var left = length;
+    var combinedIdx = 0;
+    while (combinedIdx < received.length || argsIdx < arguments.length) {
+      var result;
+      if (combinedIdx < received.length &&
+          (received[combinedIdx] == null ||
+           received[combinedIdx]['@@functional/placeholder'] !== true ||
+           argsIdx >= arguments.length)) {
+        result = received[combinedIdx];
+      } else {
+        result = arguments[argsIdx];
+        argsIdx += 1;
+      }
+      combined[combinedIdx] = result;
+      if (result == null || result['@@functional/placeholder'] !== true) {
+        left -= 1;
+      }
+      combinedIdx += 1;
+    }
+    return left <= 0 ? fn.apply(this, combined) : arity(left, _curryN(length, combined, fn));
+  };
+};
--- a/src/curryN.js
+++ b/src/curryN.js
@@ -1,5 +1,5 @@
 var _curry2 = require('./internal/_curry2');
-var _slice = require('./internal/_slice');
+var _curryN = require('./internal/_curryN');
 var arity = require('./arity');
 
 
@@ -47,43 +47,5 @@ var arity = require('./arity');
  *      g(4); //=> 10
  */
 module.exports = _curry2(function curryN(length, fn) {
-  return arity(length, function() {
-    var n = arguments.length;
-    var shortfall = length - n;
-    var idx = n - 1;
-    while (idx >= 0) {
-      if (arguments[idx] != null &&
-          arguments[idx]['@@functional/placeholder'] === true) {
-        shortfall += 1;
-      }
-      idx -= 1;
-    }
-    if (shortfall <= 0) {
-      return fn.apply(this, arguments);
-    } else {
-      var initialArgs = _slice(arguments);
-      return curryN(shortfall, function() {
-        var currentArgs = _slice(arguments);
-        var combinedArgs = [];
-        var idx = 0;
-        var currentArgsIdx = 0;
-        while (idx < n) {
-          var val = initialArgs[idx];
-          if (val != null && val['@@functional/placeholder'] === true) {
-            combinedArgs[idx] = currentArgs[currentArgsIdx];
-            currentArgsIdx += 1;
-          } else {
-            combinedArgs[idx] = val;
-          }
-          idx += 1;
-        }
-        while (currentArgsIdx < currentArgs.length) {
-          combinedArgs[idx] = currentArgs[currentArgsIdx];
-          idx += 1;
-          currentArgsIdx += 1;
-        }
-        return fn.apply(this, combinedArgs);
-      });
-    }
-  });
+  return arity(length, _curryN(length, [], fn));
 });


src/internal/_curryN.js
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Internal curryN function.
+ *
+ * @private
+ * @category Function
+ * @param {Number} length The arity of the curried function.
+ * @return {array} An array of arguments recieved thus far.
+ * @param {Function} fn The function to curry.
+ */
+module.exports = function _curryN(length, recieved, fn) {
```
davidchambers: s/recieved/received/

davidchambers:
I prefer `for` loops rather than `while` loops in many cases because they group the boilerplate for managing the loop index. I don't like our `var idx = -1; while (++idx < list.length) { ... }` pattern as it requires setting the index variable to -1 initially rather than 0.

On a similar note, I'm opposed to the use of `++` and `--` in most cases. The existence of both prefix and postfix forms is a source of confusion, and the prefix forms are syntactic sugar for `+= 1` and `-= 1` respectively. Were we permitted to use `for` loops and `Array.prototype.push` we could remove most or all occurrences of `++` and `--` (both prefix and postfix).

buzzdecafe:
part of the reason i started collaborating on eweda way back when was because i hate `for` loops so much. this lib is no longer my playground, and hasn't been for a while, so i won't waste energy fighting `for`.  it's an implementation detail. I'll just throw up in my mouth a little bit every time i see one in the codebase.

paldepind:
> part of the reason i started collaborating on eweda way back when was because i hate for loops so much.

I don't understand this when the alternative is `while` loops. The way I see it almost all iterations are captured more cleanly with a `for` loop.

buzzdecafe:
_De gustatibus non est disputandum._

paldepind:
Are you ok with the `for` loops @CrossEye? I know you don't like them as well.

I'm still willing to take them out.

CrossEye:
I'm more than willing to _disputandum_ about _gustatibus_.  :-)

I actually have less objection, I think, to the `for`-loops themselves than @buzzdecafe does.  But I would prefer to keep our source code consistent.  (Yes, I know the Emerson quote, thank you.)  I was the one who started us down the `while`-loop path, based both on very ancient memories of when `while`-loops were at least marginally faster than `for`-loops and on a mild distaste for the `for`-loop syntax.

I still have a distaste for that syntax.  It's somewhat mild, but quite real.  I feel we've conditioned ourselves to accept an ergonomically unfit syntax rather than build a syntax designed to help us.  Compare any of these to the C-style `for`-loops the Javascript has inherited:

```python
for x in range(1, 11):
    print x
```
```ruby
for x in 1..10
   puts x
end
```
```lisp
(loop for x from 1 to 10
    do (print x))
```

All these feel clean and readable.  Of course they're not as powerful and flexible (well, LISP is, but only by becoming more convoluted) but they express the common case so much more nicely.

I do appreciate the concern @davidchambers has about starting the loop iteration at `-1`.  I don't really like that either.  And, while I don't share his objection to `++` and/or `--`, I would have no problems with giving up that construct.  But I am loathe to switch to `for`-loops.

I would have no problem at all, though changing universally from this style:

```js
function map(fn, list) {
  var idx = -1, len = list.length, result = [];
  while (++idx < len) {
    result[idx] = fn(list[idx]);
  }
  return result;
};
```

to this one:

```js
function map(fn, list) {
  var idx = 0, len = list.length, result = [];
  while (idx < len) {
    result[idx] = fn(list[idx]);
    idx+= 1;
  }
  return result;
};
```

So, yes, I would rather do this with `while`-loops.  But I wouldn't at all mind switching the style of while loops if that would make others (anyone besides David?) happier.

davidchambers:
> And, while I don't share [David's] objection to `++` and/or `--`, I would have no problems with giving up that construct.

#1164

paldepind:
@CrossEye

>  But I would prefer to keep our source code consistent.

As I wrote before: "It's like always using reduce and forbidding map. And the kind of "consistency" that is achieved by doing so is just as harmful." It's like _always_ using the hammer in the name of consistency.

I certainly agree with you that the `for` loops you posted from other languages are more readable. But you made no argument for why JavaScript `while` loops are preferable to JavaScript `for` loops. I have a few reasons for the opposite:

* `for` loops are more compact. They take fewer lines. And, less importantly, fewer characters.
* `for` loops have _well defined_ locations for initialization, end condition and increment. This makes them more straightforward to write and easier to read (because there's a fixed location for these things). If we where using `for` loops there would be no discussion about which style to use because the `for` loops themselves dictates a common style.

paldepind:
I'll just repost this diagram that @davidchambers posted because I think it demonstrates very clearly the problem with how we use `while` loops.

This is a while loop:

```javascript
<x>;
while (<y>) {
  <a>;
  <z>;
}
```

And this is the equivalent for loop:

```javascript
for (<x>; <y>; <z>) {
  <a>;
}
```

Notice first of all that the for loop is 2 lines shorter! That in itself is nice. But the important difference is that in the for loop the meaning and the location of the various `<*>`s is dictated by the for loop! Almost _every_ programmer will be immediately familiar with what the various parts do and where they are.

On the other hand the location of the various things in the while loop is arbitrary and invented by us.

The strictness that the for loop enforces makes it easier to parse for the programmer! There are less moving parts.

CrossEye:
Although I will buy arguments from brevity, only for equally clear code.  And my take is that `for` is simply not as clear.  What you see as something equally familiar to all developers, I see as something similar to http://underscorejs.org/#sortedIndex: too much complexity masquerading as flexibility.

First of all, you only mention one of the three `for` syntaxes the language now has:

```js
    for (initializers; tests; incrementers) { ...}
    for (name in obj) { ... }
    for (prop of iterator) { ... }
```

Then you have to consider whether to use `var` or `let` inside the statement, or whether to do it beforehand:

```js
    for (var i = 0; i < arr.length; ++i) { ... }
    for (let i in arr) { ... }

    var ctr, arr = getData(), len = arr.length;
    for (ctr = 0; ctr < len; ctr++) { ... }
```

Then you need to deal with the fact that not only are the three sections of the `;`-separated syntax version optional, two of them allow for multiple phrases:

```js
    for (var idx = 0, matches = 0; 
         idx < arr.length && matches <20; 
         idx++, matches += isMatch(idx) ? 1 : 0) { ... }
    for (;val; val--) { ... }
    for (; --current; ) { ... }
```

This is not simply a matter of crazy remote possibilities:  All these were found in the recent projects from a developer we all know:

```js
    for (var i = 0; i < arguments.length; ++i) { ... }
    for (; orderNextIdx >= 0; --orderNextIdx) {...}
    for (operation in delta) {
        for (key in delta[operation]) { ... }
    }
    for(;j=list[i++];) { ... }
```

So yes, I'm still mostly opposed to `for`-loops, although my objection might not be as strong as the foregoing makes it sound.

paldepind:
> First of all, you only mention one of the three for syntaxes the language now has:

Even though they begin with the same keyword those are all different constructs that are easy to tell apart.

> Although I will buy arguments from brevity, only for equally clear code.

That we agree on! But I think `for` is clearer ;)

> Then you have to consider whether to use var or let inside the statement, or whether to do it beforehand:

Where to declare the variable. That is an issue JavaScript developers face all the time independent of whether or not they use `for`-loops. I personally prefer to only put the `var` inside the for loop if it's inside a function where the index is the only one being declared.

> Then you need to deal with the fact that not only are the three sections of the ;-separated syntax version optional, two of them allow for multiple phrases:

I don't accept that argument. In JavaScript the comma operator can be used _everywhere_ a statement is expected. `for`-loops and `while`-loops are on equal grounds here.

```javascript
while (idx < len, 'what', 15, 'going on', Math.min(idx, 0), true) {
  // Wow! Aren't while loops bad?
}
```

> This is not simply a matter of crazy remote possibilities.

Yes it is. You can write horrible code with both `for` and `while` loops. It is unfair to compare worst-possible-for-loop-usage with best-possible-while-loop-usage. If, on the other hand, you compare `for`- and `while`-loops on equal terms the conclusion is that _for loops are harder to screw up_. Why? Because the construct itself dictates where you should put `<x>`, `<y>`, `<z>` and `<a>`. This guarantees consistency. Not only across Ramda – but across all of JavaScript. With `while` loops the individual programmer will have to decide where to put `<x>`, `<y>`, `<z>` and `<a>`. Then you're certain that you will not get something consistent across code bases.

The above is also why the `for` loop is clearer! Since the for loop is consistent across JavaScript (and even other languages) it is very easy and mentally untaxing to parse. You immediately know where to look for `<x>`, `<y>`, `<z>` and `<a>`. With `while` loops you first have to figure out where this particularly project choose to put those things. A splendid example of this issue is the recent reordering of `<x>`, `<y>`, `<z>` and `<a>` in Ramda. With `for`-loops that wouldn't even have been possible.

And FWIW, in the example code you posted the only usage of `for` I don't like is `for(;j=list[i++];) { ... }` – that is essentially a while loop and thus a while loop would indicate the intend more clearly. But I guess the for loop was used as a size optimization (a `while` loop can always be replaced with a smaller `for` variant).

buzzdecafe:
Arguments in favor of one or the other terrible loop syntax strike me as a waste of energy. `for` and `while` are really quite close. *That said,* the _reason_ `for` loops are inferior is not just that they are _a vulgar display of explicit iteration_. A `for` loop is how a computer "thinks". There is nothing human about it whatsoever. It is pure, technical *prose*, that bends the person to the tool. A `while` loop is not much better. But any amount better is still better.

To my mind the arguments is purely *aesthetic*.

Now I'm done. Carry on.

CrossEye:
@paldepin: I think you're misunderstanding what I'm suggesting here.  I'm not arguing for `while`.  I'm arguing against `for`.  I believe `for` is overcomplicated, and does not do a good job of simply describing developer intent.

I do think `while` is simpler, but that's not the point I had been making, only that `for` is pretty horrible.  Here I'll try to defend `while` as a better alternative to `for`.

When you make this argument,

```js
for (<x>; <y>; <z>) {
  <a>;
}
```

```js
<x>;
while (<y>) {
  <a>;
  <z>;
}
``` 


You are presupposing the structure of `for` (or of one of the `for` syntaxes.)  If I were to start from [the definition of `while`][wh]

```js
while (<c>) {
  <s>
}
```

And then map that back onto `for`, I would get the bizarre version of `for` that you didn't like:

```js
for ( ; <c> ; ) {
  <s>
}
```

I don't think that really proves much, either way.  Nor do particular individual examples, such as 

```js
while (getNextToken() !== null) { ... }
```

But they do serve to show that there are things more easily done with one than the other.

But speaking of the definitions, the MDN definitions of the [two forms][wh] [of while][dw] are both substantially simpler than any of the definitions of the [three][fr] [forms][fi] [of for][fo].  I think that says something important about their relative complexities.

People have been quoting a great deal here lately from the Zen of Python, and I think three of its lessons are quite important:

>     Explicit is better than implicit.

If we explicitly perform any index initialization in a very clear and open fashion without burying it inside some looping construct, I think we gain something.

>     Simple is better than complex.

`while` tests a condition and repeated processes a statement/block so long as the conditions is truthy.  `for` initializes your loop counters, tests a conditions, processes a statement/block so long as its truthy, and then updates the loop counters.  (And by specifying loop counters, I'm actually oversimplifying this a bit.)  Clearly `while` -- _as a statement_ -- is simpler.  We can also discuss the code we write that uses `while`, realizing that adding back initialization and incrementing will change the balance somewhat.  But I will argue that generally, building on simpler pieces is simpler.

>     There should be one-- and preferably only one --obvious way to do it.

`while` is clearly the more general of the two constructs, as you can model any `for` as a `while`, but the reverse is substantially more difficult, perhaps impossible.  And since we might need these features, we will need to allow `while` statements.  If we disallow `for` statements, it helps make more obvious how to code your functions.
  

>And FWIW, in the example code you posted the only usage of for I don't like is ...

I would _hope_ you wouldn't mind that code too much!  It's all taken from top projects on your Github page.  My point was simply that `for` really does get used in all these ways.  And I really have seen many examples in the real world of multiple initialization statements in a single block.  I've never that I can recall seen a comma operator in a `while` condition.

  [dw]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/do...while
  [fi]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in
  [fo]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of
  [fr]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for
  [wh]:  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/while

paldepind:
> I think you're misunderstanding what I'm suggesting here. I'm not arguing for while. I'm arguing against for. I believe for is overcomplicated, and does not do a good job of simply describing developer intent.

Yes. I was misunderstanding you. I thought you where arguing for why `for` is worse than `while`.

> You are presupposing the structure of for (or of one of the for syntaxes.)

No. I'm showing how `while` loops would look in our code base compared to how `for` loops would look. I'm arguing specifically that `for` loops are better at iterating lists.

> But they do serve to show that there are things more easily done with one than the other.

Yes. That is a great example of when `while` is preferable.

> But speaking of the definitions, the MDN definitions of the two forms of while are both substantially simpler than any of the definitions of the three forms of for. I think that says something important about their relative complexities.

`while` loops are simpler. No doubt about that. But not by much. The description of the `for` loop on MDN is still quite brief.

> If we explicitly perform any index initialization in a very clear and open fashion without burying it inside some looping construct, I think we gain something.

No no no! Nothing can be more explicit than putting index initialization inside a syntax construct that dictates where to put the index initializations. When I read the before mentioned MDN article about `for` I know where to find the index initialization in all `for` loops. How can putting something at a location where all developers will expect it and immediately understand it be "burying"?

What you're suggesting is putting the initialization in some (to the language) arbitrary position where it only has a meaning because you assign it one (and that location was a different one yesterday).

> We can also discuss the code we write that uses while, realizing that adding back initialization and incrementing will change the balance somewhat. But I will argue that generally, building on simpler pieces is simpler.

Why is building on a simpler piece necessarily simpler? When you at _all_ the additional complexity of the `for` loop to the `while` loop you are left with identical complexity. You have the exact same number of statements. _If_ the `while` loop actually let us iterate a list with _fewer_ statement I would be all over it. But it doesn't.

> while is clearly the more general of the two constructs, as you can model any for as a while, but the reverse is substantially more difficult, perhaps impossible.

No. As I've mentioned before a `for` loop can express the same (and always in fewer characters). But, let's rephrase your statement into using `reduce` and `map`:

> `reduce` is clearly the more general of the two constructs, as you can model any `map` as a `reduce`, but the reverse is substantially more difficult, perhaps impossible. And since we might need these features, we will need to allow `reduce` statements. If we disallow `map` statements, it helps make more obvious how to code your functions.

Do you still agree? I think having a rich (but not too rich vocabulary) is very important when programming. It makes it possible to communicate ones intend better. The `for` loop is specifically designed for iterating over an array and thus communicates that intend better than a `while` loop.

I don't buy the "`while` is simpler" argument. True, while _is_ simpler. That is why one should use it when only the limited power of `while` is required. `for` captures more complexity. But for iterating over a list with an index (the case we are discussing) it captures the _right_ amount of complexity. `for` loop is a tool that fits the job perfectly. Sure, you can add a bunch of statements to `while` and achieve a solution as well. But don't insist on using the hammer when we're not dealing with a nail. The problem at hand is iterating an index and `while` does not let us arrive at a simpler solution.

> I would hope you wouldn't mind that code too much! It's all taken from top projects on your Github page. My point was simply that for really does get used in all these ways.

Haha! I didn't recognize the code. Where is the last line from? That is surely a mistake.

> I've never that I can recall seen a comma operator in a while condition.

That because `while` is to simple that you instead add extra statements before or inside the `while` loop.

CrossEye:
> Yes. I was misunderstanding you. I thought you where arguing for why `for` is worse than `while`.

I do think that, but only because I believe the JS `for`-loop is so horrid.  That's what I was arguing.

Where functions like `map` abstract away iterator initialization, bounds-checking, and incrementing, and where the Python, Ruby, and LISP syntaxes above make these concerns declarative, the C-style loop makes you implement all these blocks, structure your code around its notion of how they should be organized, and then ensure that you don't muck about with those iterator indices it has claimed as its own.  In short, to understand code in a `for`-loop, I have to play JS interpreter.

Here's a simple example:  a loop that iterates three times, with {1, 2, 3, 4, ...} representing the order in which each step is run.  In each of the samples, mentally (or even better, by pointing) walk through the sequence of events:

```js
for (<x> {1}; <y> {2, 5, 8, 11}; <z> {4, 7, 10}) {
  <a>; {3, 6, 9}
}
{12}
```

```js
<x>; {1}
while (<y>) {2, 5, 8, 11} {
  <a>; {3, 6, 9}
  <z>; {4, 7, 10}
}
{12}
```

To me, the latter one flows much more smoothly.  The steps of what is supposed to be an imperative block of code runs top to bottom, the flow interrupted only by jumping back to the beginning of the loop, and one final jump out of the loop.  (The `do-while` structure wouldn't even have that last jump.)

In the `for`-loop version, you can build up the right rhythm to follow the steps, but it certainly doesn't feel as natural.  It's worse if some of the `;`-separated sections are missing, because of course the work has to be done, and therefore it's being done somewhere else (as part of `<a>`, obviously), but the mental model becomes more problematic.

And there is further complexity in the fact that the same keyword is used for two other unrelated -- although similar to one another -- structures.  `for-in`-loops and `for-of` loops overload the symbol `for` with very little benefit.  The reason that Ramda signatures are not overloaded is because we find such to be an unwarranted source of complexity.  `for`-loops come across to me the same way.


> I'm arguing specifically that `for` loops are better at iterating lists.

Right, and I'm arguing that `for`-loops are so generally bad that we wouldn't want them at all, even if in using them for iterating lists we might reduce line-count.


>> If we explicitly perform any index initialization in a very clear and open fashion without burying it inside some looping construct, I think we gain something.

> No no no! Nothing can be more explicit than putting index initialization inside a syntax construct that dictates where to put the index initializations. 

Well yes, we could become more explicit still, with some imaginary syntax:

```js
    var i = 0;
    loopBegin:
        if (i >= arr.length) goto loopEnd
        // do something useful
        goto loopBegin  
    loopEnd:
```

But we've long-since grown accustomed to seeing this style of coding as wrong.  We're trying to play compiler/interpreter in our own source code here, not abstracting up to a useful level.  To me `for`-loops in Javascript are just like that.

>When I read the before mentioned MDN article about `for` I know where to find the index initialization in all `for` loops. How can putting something at a location where all developers will expect it and immediately understand it be "burying"?

I'm really trying to make the distinction between [simple and easy][rh].  Simple is "less intertwined", easy is "more familiar".  Where `while` loops ignore any considerations besides repetition and bounds testing and  where the Python, Ruby, and LISP implementations abstract them so that the user doesn't need to consider them, one Javascript `for`-loop syntax intertwines the notion of iterator initialization and iterator incrementing with what might be seen as the more  fundamental parts of looping.  Worse, the other two formats, which are almost more like puns on the name `for` than shared ideas, actually act much more like their Python, _et. al._, peers, so Javascript can't even make up its mind about what the important parts of looping really are.  The fact that these loops are familiar to developers is not much of a selling point.  Mutation is familiar to developers, OO is familiar to developers.  All sorts of familiar things are pretty awful.



> What you're suggesting is putting the initialization in some (to the language) arbitrary position where it only has a meaning because you assign it one (and that location was a different one yesterday).

No, I'm suggesting that we don't use `for`-loops.  Since `while` is the only other game in town, I feel as though we're stuck with it.  I'm not particularly concerned with where we put initialization, or even whether it's always required.  I like the change made to stop intializing our counters to `-1`, but if we found some other way to do the initialization when it's needed, it wouldn't bother me at all.  The fact that it's not always the more familiar way is way, way down on my list of concerns.


> But, let's rephrase your statement into using `reduce` and `map`:

>> `reduce` is clearly the more general of the two constructs, as you can model any `map` as a `reduce`, but the reverse is substantially more difficult, perhaps impossible. And since we might need these features, we will need to allow `reduce` statements. If we disallow `map` statements, it helps make more obvious how to code your functions.

> Do you still agree? 

I absolutely would if `map` were significantly more complex than `reduce`

> I think having a rich (but not too rich vocabulary) is very important when programming. It makes it possible to communicate ones intend better. The `for` loop is specifically designed for iterating over an array and thus communicates that intend better than a `while` loop.

No, no, no!  The `for` loop is not specifically designed for iterating over an array, or it would be be better designed.  LISP, Ruby, and Python proved that it's quite possible.  The `for` loop is designed to be a general-purpose looping construct, a close-to-the-metal gloss on a looping pattern used in assembly programming.

To me, this is way past the edge, not exactly of too rich, but definitely of too complex.  

> I don't buy the "`while` is simpler" argument. True, while _is_ simpler. That is why one should use it when only the limited power of `while` is required. `for` captures more complexity. But for iterating over a list with an index (the case we are discussing) it captures the _right_ amount of complexity. 

Well, I'm not arguing particularly that `while`-loops are good, only that `for`-loops are bad enough to banish them altogether.  Since we still have `while`-loops available, that's what we're stuck using.  If the language offered a better construct, we would use that.  If, for instance, the `for-of`-loop was the only one available in JS, I would unhesitatingly prefer that, as it captures the correct level of abstraction.


>> I would hope you wouldn't mind that code too much! It's all taken from top projects on your Github page. My point was simply that for really does get used in all these ways.

> Haha! I didn't recognize the code. Where is the last line from? That is surely a mistake.

[synceddb/client/synceddb.js][sd]

I had really been hoping to find an example with comma operators inside the `for`-loops, which is part of what make them so awful to me.  But from the little investigating I did, you don't use that construct.  It is one I've seen far too often, though.


>> I've never that I can recall seen a comma operator in a while condition.

> That because `while` is to simple that you instead add extra statements before or inside the `while` loop.

To me that's a big plus!

----------

Okay, that's it from me for now.  Feel free to continue to try to convince me.  But this message was already repetitive, and I'm just likely to get more and more and more and more and more repetitive (see?) if I continue this.  I will respond to this sub-thread only if I find that you or someone else is starting to change my mind.

We do need to put this question to bed soon.  I know that @davidchambers would prefer to at least ease the ban on `for`-loops, and I know that @buzzdecafe would prefer never to have to look at a `for`-loop again, and would probably rather to never have to discuss them again either.  I'm not sure what @raine thinks.  (Pinging @raine: any thoughts on this?)  Anyone else have input here?

My feeling for the specific case which brought this long discussion on is that for the moment, consistency should still trump.  But I'm more than willing to listen to arguments about this too.


  [sd]: https://github.com/paldepind/synceddb/blob/c9e801abcc61bbc55f6a7715d792d0115f377ff5/client/synceddb.js#L24

paldepind:
> To me, the latter one flows much more smoothly.

While interesting that is not how I read `for` nor `while` loops. I don't execute them in my head in that way. I read them from top to bottom once.

> No, no, no! The for loop is not specifically designed for iterating over an array, or it would be be better designed. LISP, Ruby, and Python proved that it's quite possible. The for loop is designed to be a general-purpose looping construct, a close-to-the-metal gloss on a looping pattern used in assembly programming.

It is specifically made to capture a looping pattern where one initializes a starting value, declares an end condition and an incrementation.

> To me, this is way past the edge, not exactly of too rich, but definitely of too complex. 

Instead of focusing so much about whether `for` loops or `while` loop I think you should look more at the complexity of the resulting code.

> To me that's a big plus!

To me it's a big disadvantage. Having these things contained inside the syntax construct adds structure.

> synceddb/client/synceddb.js

I didn't write that :) It's a tiny dependency that I decided to copy directly inside the source code.

> My feeling for the specific case which brought this long discussion on is that for the moment, consistency should still trump. But I'm more than willing to listen to arguments about this too.

Consistency in this regard only reduces our vocabulary and thus how precisely we can express ourself in code.

Anyway, when I get the time I'll rebase this PR and switch to `while` loops.

CrossEye:
>> synceddb/client/synceddb.js

> I didn't write that :) It's a tiny dependency that I decided to copy directly inside the source code.

Ahh, I only searched through some of your repos for `for`.  No wonder you didn't recognize it!

paldepind:
I've now taken the `for`-loops out of this PR.

CrossEye:
:herb:

src/internal/_curryN.js
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Internal curryN function.
+ *
+ * @private
+ * @category Function
+ * @param {Number} length The arity of the curried function.
+ * @return {array} An array of arguments received thus far.
+ * @param {Function} fn The function to curry.
+ */
+module.exports = function _curryN(length, recieved, fn) {
```
davidchambers: s/recieved/received/

src/internal/_curryN.js
```diff
@@ -0,0 +1,38 @@
...
+    var argsIdx = 0;
+    var left = length;
+    var combinedIdx = 0;
+    while (combinedIdx < recieved.length || argsIdx < arguments.length) {
+      var result;
+      var rec = recieved[combinedIdx];
+      var recNotPlaceholder = rec == null || rec['@@functional/placeholder'] !== true;
+      if (combinedIdx < recieved.length &&
+          (recNotPlaceholder || argsIdx >= arguments.length)) {
+        result = rec;
```
davidchambers: I think this would be clearer without `recNotPlaceholder`. Also, it seems wrong to access `recieved[combinedIdx]` before checking the bounds. I suggest the above become:

```javascript
if (combinedIdx < received.length &&
    (received[combinedIdx] == null ||
     received[combinedIdx]['@@functional/placeholder'] !== true ||
     argsIdx >= arguments.length)) {
  result = received[combinedIdx];
```

src/internal/_curryN.js
```diff
@@ -0,0 +1,38 @@
...
+module.exports = function _curryN(length, recieved, fn) {
+  return function() {
+    var combined = [];
+    var argsIdx = 0;
+    var left = length;
+    var combinedIdx = 0;
+    while (combinedIdx < recieved.length || argsIdx < arguments.length) {
+      var result;
+      var rec = recieved[combinedIdx];
+      var recNotPlaceholder = rec == null || rec['@@functional/placeholder'] !== true;
```
CrossEye: Before we go much further, should we make 

```js
var placeholder = require('_placeholder');
//...
    var recNotPlaceholder = rec == null || rec[placeholder] !== true;
```

and stop using that string all over the source code?  (I can't believe I'm suggesting a String constant after railing against them over and over at work, but I think it makes sense here.)

src/internal/_curryN.js
paldepind: Yes. That is probably better.

src/internal/_curryN.js
buzzdecafe: > I can't believe I'm suggesting a String constant after railing against them over and over at work ...

:laughing: ya i don't know how many times i have contradicted myself over the years on this lib. more than i can count.

paldepind: What about a helper function?

```javascript
var isPlaceholder = require('_isPlaceholder');
//...
var recNotPlaceholder = !isPlaceholder(req);
```

davidchambers: > What about a helper function?

I assumed a helper function was off the table for performance reasons, but I'm far from a performance guru. :)

> (I can't believe I'm suggesting a String constant after railing against them over and over at work, but I think it makes sense here.)

I've been on your side of this debate many times myself. My argument has been that using `someModule.FOOBAR` (whose value is `'foobar'`) provides no benefit over using `'foobar'` directly. In this case, though, there would be a benefit: JSHint would tell us if we mistyped `placeholder`.

paldepind: It is not my impression that function calls are expensive. At least not in V8. Small functions will in general be optimized better and are subject to inlining. But if this is a concern we should benchmark it. I think the code cleanness advantage would justify it.

davidchambers: I'm absolutely in favour of a helper function in that case. I'd have written one myself had I thought I were "allowed". :)

buzzdecafe: do it

paldepind: I'd also like to point out that the helper could easily test explicitly for `undefined` and `null` instead of relying on type coercion like we currently do. I've experienced performance gains from that in the past.

davidchambers: Interesting. You could also go in the other direction to provide another point of comparison:

```javascript
module.exports = function _isPlaceholder(x) {
  return Object(x)['@@functional/placeholder'] === true;
};
```

CrossEye:
Well, there is a performance penalty, especially in Firefox.  I don't know how large.  At some point we need to measure this stuff.  Or get input from @megawac.  :smile:

davidchambers:
:tada:

buzzdecafe:
> At some point we need to measure this stuff.

sure, we can always circle back and optimize certain paths. for now we should worry about the :horse: and the cart can wait. (disappointed there is no `:cart:` emoji)

src/internal/_curryN.js
```diff
@@ -0,0 +1,38 @@
...
+      if (combinedIdx < received.length &&
+          (received[combinedIdx] == null ||
+           received[combinedIdx]['@@functional/placeholder'] !== true ||
+           argsIdx >= arguments.length)) {
+        result = received[combinedIdx];
+      } else {
+        result = arguments[argsIdx];
+        argsIdx += 1;
+      }
+      combined[combinedIdx] = result;
```
megawac: Isn't the idiom you guys have been following `combined[++combinedIdx] = result;` or did I miss something

src/internal/_curryN.js
```diff
@@ -0,0 +1,38 @@
...
+      if (result == null || result['@@functional/placeholder'] !== true) {
+        left -= 1;
+      }
+      combinedIdx += 1;
+    }
+    return left <= 0 ? fn.apply(this, combined) : arity(left, _curryN(length, combined, fn));
```
megawac: left can't be less than 0

src/internal/_curryN.js
paldepind: It's `push` now :) And `++` has been banished :(

megawac:
> Well, there is a performance penalty, especially in Firefox. I don't know how large. At some point we need to measure this stuff. Or get input from @megawac.  :smile:

Arity unwrapping (probably through metadata) would provide a huge boost

buzzdecafe:
hey @megawac did you want me to merge your meta pr into meta branch? or can you re-pr vs meta branch instead?

and i think `meta` could have a lot of other benefits, including reduced stack size, which means more comprehensible debugging, a space to store function sigs, and possibly even source code. I think there is a lot of potential there. 

Sorry for hijacking the thread. carry on.

megawac:
Can you rebase the meta branch on the pr? I currently don't have proper
internet access (coffee shop life)

buzzdecafe:
i feel your pain. i'll give it a shot. 

:coffee:

PR #1156, opened 2015-06-05
identify R.__ by special property rather than by identity

### LINKED ISSUE #1052: Curry placeholder should play nice with other Ramda's and other libs
Currently the Ramda `__`-placeholder is defined like this:

```javascript
module.exports = {ramda: 'placeholder'};
```

And it is checked for at various places with strict equality:

```javascript
if (arguments[idx] === __) {
  // foo
}
```

Now, consider what happens if I'm using `npm` to manage dependencies and I'm using a library, `someFnLibDepOnOldRamda`, that depends on a slightly different version of Ramda than the one I'm using myself. Then `npm` will pull in two different versions of Ramda and with them two different instances of `__`. Thus the strict equality check that Ramda uses breaks.

If pass my instance of `__` to a function from the library, curried with it's Ramda's curry it will not be recognized as a placeholder. Bam! I now have an obscure confusing bug :(

One solution: Define this placeholder like this:

```javascript
module.exports = {someSpecialPropertyName: true};
```

And check for it like this:

```javascript
function isPlaceholder(o) {
  return o && o.someSpecialPropertyName === true;
}
```

In particular the name `someSpecialPropertyName` should _not_ include the word "ramda". This would make it more natural for potential other curry implementations to adopt the same placeholder "specification".

davidchambers:
Closes #1052

The `'@@functional/placeholder'` property name was proposed by @paldepind in https://github.com/ramda/ramda/issues/1052#issuecomment-97822071.

---

If we decide to merge this pull request, I wonder how likely it is that the Underscore contributors would agree to add this property to the `_` object.

### DIFF
--- a/test/curry.js
+++ b/test/curry.js
@@ -76,6 +76,36 @@ describe('curry', function() {
     assert.deepEqual(g(_, _, _)(1, _, _)(_, _)(2, _)(_)(3), [1, 2, 3]);
   });
 
+  it('supports @@functional/placeholder', function() {
+    var f = function(a, b, c) { return [a, b, c]; };
+    var g = R.curry(f);
+    var _ = {'@@functional/placeholder': true, x: Math.random()};
+
+    assert.deepEqual(g(1)(2)(3), [1, 2, 3]);
+    assert.deepEqual(g(1)(2, 3), [1, 2, 3]);
+    assert.deepEqual(g(1, 2)(3), [1, 2, 3]);
+    assert.deepEqual(g(1, 2, 3), [1, 2, 3]);
+
+    assert.deepEqual(g(_, 2, 3)(1), [1, 2, 3]);
+    assert.deepEqual(g(1, _, 3)(2), [1, 2, 3]);
+    assert.deepEqual(g(1, 2, _)(3), [1, 2, 3]);
+
+    assert.deepEqual(g(1, _, _)(2)(3), [1, 2, 3]);
+    assert.deepEqual(g(_, 2, _)(1)(3), [1, 2, 3]);
+    assert.deepEqual(g(_, _, 3)(1)(2), [1, 2, 3]);
+
+    assert.deepEqual(g(1, _, _)(2, 3), [1, 2, 3]);
+    assert.deepEqual(g(_, 2, _)(1, 3), [1, 2, 3]);
+    assert.deepEqual(g(_, _, 3)(1, 2), [1, 2, 3]);
+
+    assert.deepEqual(g(1, _, _)(_, 3)(2), [1, 2, 3]);
+    assert.deepEqual(g(_, 2, _)(_, 3)(1), [1, 2, 3]);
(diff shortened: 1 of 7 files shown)


buzzdecafe:
:cow2: 
let's do it, and adjust to reactions of others later

megawac:
Heads up, there used to be a deopt associated with property names that cannot be accessed by dot notation

CrossEye:
I think this is a good pass at this.  Later, we may want to consider a Symbol instead, for those environments supporting them, but that's not urgent.

@megawac: although I doubt that would stop us (as this style of property name for such interfaces if getting to be a defacto standard, ISTM), I am curious to hear more about that de-opt.  Do you have any references for it?

test/curry.js
```diff
@@ -76,6 +76,36 @@ describe('curry', function() {
...
+    assert.deepEqual(g(_, _, 3)(1, 2), [1, 2, 3]);
+
+    assert.deepEqual(g(1, _, _)(_, 3)(2), [1, 2, 3]);
+    assert.deepEqual(g(_, 2, _)(_, 3)(1), [1, 2, 3]);
+    assert.deepEqual(g(_, _, 3)(_, 2)(1), [1, 2, 3]);
+
+    assert.deepEqual(g(_, _, _)(_, _)(_)(1, 2, 3), [1, 2, 3]);
+    assert.deepEqual(g(_, _, _)(1, _, _)(_, _)(2, _)(_)(3), [1, 2, 3]);
+  });
+
```
paldepind: All this testing looks superfluous to me. `curry` is based on `curryN` and `curryN` is tested for the exact same behavior

davidchambers: I don't follow. We need these tests to ensure we're not comparing to `R.__` by identity, don't we?

paldepind: My point is that these tests should be for `curryN`. You've made no change to `curry` – yet you've changed the tests for `curry` but not for `curryN`.

CrossEye: Well, since `curry` is public and its behavior has changed, it's tests should probably reflect that.  The fact that it is implemented in terms of `curryN` is not a public guarantee, so our tests shouldn't assume this.

OTOH, we probably also need similar tests for `curryN`, as that one is also public.

paldepind: That is not an unreasonable way to see it. I agree.

davidchambers: I've added tests for `R.curryN`. Thanks for pointing out this omission, @paldepind.

paldepind: Hehe. It was not really my initial intent. But I'm glad it lead to something useful.

CrossEye:
:herb:

jdalton:
> If we decide to merge this pull request, I wonder how likely it is that the Underscore contributors would agree to add this property to the _ object.

Not likely from my end. Lodash/Underscore support customizable placholders in a different way, based on the `.placeholder` property on the methods with support, e.g. `_.partial.placeholder`.

Issue #1089, opened 2015-05-15
assert.notStrictEqual


buzzdecafe:
in light of the discussion in #1025 , add tests to show that list functions do not return a reference to the input list

PR #1159, opened 2015-06-06
derive R.take and R.drop from R.slice

### LINKED ISSUE #1158: derive R.take and R.drop from R.slice
`R.slice` works on strings as well as arrays, since we dispatch to a `slice` method if present:

```javascript
> R.slice(0, 3, [1, 2, 3, 4, 5])
[ 1, 2, 3 ]
> R.slice(0, 3, 'Ramda')
'Ram'
```

One can define `take` in terms of `slice`. If one does so `take` also works on strings:

```javascript
> var take = R.slice(0)
undefined
> take(3, [1, 2, 3, 4, 5])
[ 1, 2, 3 ]
> take(3, 'Ramda')
'Ram'
```

`drop` can be defined similarly:

```javascript
> var drop = R.slice(R.__, Infinity)
undefined
> drop(3, [1, 2, 3, 4, 5])
[ 4, 5 ]
> drop(3, 'Ramda')
'da'
```

`take` and `drop` are derivable from `slice`. I suggest we update the implementations of `R.take` and `R.drop` to use this approach. We could then deprecate `R.substringTo` and `R.substringFrom`.

davidchambers:
Closes #1158

The simplicity of the implementation reinforces my belief that this is an entirely reasonable change. :)

### DIFF
--- a/src/drop.js
+++ b/src/drop.js
@@ -1,7 +1,7 @@
 var _curry2 = require('./internal/_curry2');
 var _dispatchable = require('./internal/_dispatchable');
-var _slice = require('./internal/_slice');
 var _xdrop = require('./internal/_xdrop');
+var slice = require('./slice');
 
 
 /**
@@ -14,13 +14,13 @@ var _xdrop = require('./internal/_xdrop');
  * @memberOf R
  * @category List
  * @sig Number -> [a] -> [a]
- * @param {Number} n The number of elements of `list` to skip.
- * @param {Array} list The array to consider.
- * @return {Array} The last `n` elements of `list`.
+ * @param {Number} n The number of elements of `xs` to skip.
+ * @param {Array} xs The collection to consider.
+ * @return {Array}
  * @example
  *
  *      R.drop(3, [1,2,3,4,5,6,7]); //=> [4,5,6,7]
  */
-module.exports = _curry2(_dispatchable('drop', _xdrop, function drop(n, list) {
-  return n <= 0 ? list : _slice(list, n);
+module.exports = _curry2(_dispatchable('drop', _xdrop, function drop(n, xs) {
+  return slice(Math.max(0, n), Infinity, xs);
 }));
--- a/src/take.js
+++ b/src/take.js
@@ -1,12 +1,12 @@
 var _curry2 = require('./internal/_curry2');
 var _dispatchable = require('./internal/_dispatchable');
-var _slice = require('./internal/_slice');
 var _xtake = require('./internal/_xtake');
+var slice = require('./slice');
 
 
 /**
- * Returns a new list containing the first `n` elements of the given list.  If
- * `n > * list.length`, returns a list of `list.length` elements.
+ * Returns a new list containing the first `n` elements of the given list.
+ * If `n > list.length`, returns a list of `list.length` elements.
  *
  * Acts as a transducer if a transformer is given in list position.
  * @see R.transduce
@@ -16,8 +16,8 @@ var _xtake = require('./internal/_xtake');
  * @category List
  * @sig Number -> [a] -> [a]
  * @param {Number} n The number of elements to return.
- * @param {Array} list The array to query.
- * @return {Array} A new array containing the first elements of `list`.
+ * @param {Array} xs The collection to consider.
+ * @return {Array}
  * @example
  *
  *      R.take(3,[1,2,3,4,5]); //=> [1,2,3]
@@ -28,6 +28,6 @@ var _xtake = require('./internal/_xtake');
  *      var takeFive = R.take(5);
  *      takeFive(members); //=> ["Paul Desmond","Bob Bates","Joe Dodge","Ron Crotty","Lloyd Davis"]
  */
-module.exports = _curry2(_dispatchable('take', _xtake, function take(n, list) {
-  return _slice(list, 0, n);
+module.exports = _curry2(_dispatchable('take', _xtake, function take(n, xs) {
+  return slice(0, n < 0 ? Infinity : n, xs);
 }));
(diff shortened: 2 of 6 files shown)


src/drop.js
```diff
@@ -14,13 +14,13 @@ var _xdrop = require('./internal/_xdrop');
  * @memberOf R
  * @category List
  * @sig Number -> [a] -> [a]
- * @param {Number} n The number of elements of `list` to skip.
- * @param {Array} list The array to consider.
- * @return {Array} The last `n` elements of `list`.
```
davidchambers: This was erroneous.

src/drop.js
```diff
@@ -14,13 +14,13 @@ var _xdrop = require('./internal/_xdrop');
...
+ * @param {Array} xs The collection to consider.
+ * @return {Array}
  * @example
  *
  *      R.drop(3, [1,2,3,4,5,6,7]); //=> [4,5,6,7]
  */
-module.exports = _curry2(_dispatchable('drop', _xdrop, function drop(n, list) {
-  return n <= 0 ? list : _slice(list, n);
+module.exports = _curry2(_dispatchable('drop', _xdrop, function drop(n, xs) {
+  return n <= 0 ? xs : slice(n, Infinity, xs);
```
buzzdecafe: didn't we agree somewhere always to clone? in which case `n < 0` might be good enough

src/take.js
```diff
@@ -28,6 +28,6 @@ var _xtake = require('./internal/_xtake');
  *      var takeFive = R.take(5);
  *      takeFive(members); //=> ["Paul Desmond","Bob Bates","Joe Dodge","Ron Crotty","Lloyd Davis"]
  */
-module.exports = _curry2(_dispatchable('take', _xtake, function take(n, list) {
-  return _slice(list, 0, n);
+module.exports = _curry2(_dispatchable('take', _xtake, function take(n, xs) {
+  return n < 0 ? xs : slice(0, n, xs);
```
buzzdecafe: (likewise here)

test/drop.js
```diff
@@ -18,6 +18,13 @@ describe('drop', function() {
     assert.deepEqual(R.drop(-Infinity, [1, 2, 3]), [1, 2, 3]);
   });
 
+  it('dispatches to `slice` method', function() {
+    var o = {slice: function(a, b) { return '[' + a + ':' + b + ']'; }};
+
+    assert.strictEqual(R.drop(3, 'Ramda'), 'da');
+    assert.strictEqual(R.drop(3, o), '[3:Infinity]');
+  });
+
```
buzzdecafe: to satisfy https://github.com/ramda/ramda/issues/1089 should probably have a test of that sort

buzzdecafe:
this is simple in implementation, and that is nice. and reducing the footprint of the api is also a plus. as noted in #1158 i think we should tread carefully when deliberately treating strings as lists. then shouldn't all list functions work on strings? i certainly wouldn't want to support that. So I'm a bit torn on this. Curious to hear other opinions.

src/drop.js
davidchambers: :zap:

src/take.js
davidchambers: :zap:

test/drop.js
davidchambers: :zap:

megawac:
> when deliberately treating strings as lists

Could be done pretty easily with a wrapper I agree, its a double edge sword. On one hand, I would love to be able to treat strings as list of chars without thinking of it; on the other hand its significant complexity

CrossEye:
I agree that we don't want to make the wholesale transition to Strings as lists, but I see much less harm in doing that where we can than in the notion some libraries have of treating Objects in the same way.  Many of the list functions make sense when applied to Strings.  In fact, I think pretty much any ` ? -> [a] -> ?` would comfortably apply to Strings.  I just don't see any great reason to go out of our way to do this.  When it can fall out of what we're doing, I have no issue, and this -- building two of our functions on top of an existing internal function, thereby removing the need for two of our less intuitive ones -- seems like a no-brainer.

So a :herb: from me, with the understanding that I agree with @buzzdecafe about not making this a precedent for turning all possible list functions into functions on Strings as well.

buzzdecafe:
thanks, :cat: :cat:. let's keep an eye out for string-creep. but in the meantime :cow2:

davidchambers:
> then shouldn't all list functions work on strings?

Due to dynamic dispatch several Ramda functions already work on strings: `R.concat`, `R.indexOf`, and `R.slice` come to mind. If one day the language defines `String.prototype.contains`, I'll expect `R.contains` to work on strings too.

I agree with you both, though, that we should not have any string-specific logic in our list functions. This is a nice clear line to draw.

megawac:
FYI, String.prototype.includes exists in the ES6 spec

davidchambers:
> FYI, String.prototype.includes exists in the ES6 spec

Thanks, @megawac. I've just realized that `contains` is derivable from `indexOf`, so `R.contains` *can* work on strings if we define it appropriately.

paldepind:
I'm happy to see `substringTo` and `substringFrom` deprecated. But I'm a little unhappy too see that this deliberately adds supports for strings to a function without it being documented. The type signature still says `Array` and it is completely opaque that this function will work on strings because they have a `slice` method.

davidchambers:
The lack of documentation was an oversight on my part. I'll rectify the situation with another pull request this weekend.

paldepind:
Great! No problem :)

asaf-romano:
Hrm, am I right that the string-based API supported negative offset, but drop doesn't?

davidchambers:
`R.slice` accepts negative offsets:

```javascript
> R.slice(-3, Infinity, 'foobar')
'bar'
```

PR #1160, opened 2015-06-06
derive R.contains from R.indexOf

davidchambers:
Related to #1159

### DIFF
--- a/src/contains.js
+++ b/src/contains.js
@@ -1,23 +1,34 @@
-var _contains = require('./internal/_contains');
 var _curry2 = require('./internal/_curry2');
+var _hasMethod = require('./internal/_hasMethod');
+var indexOf = require('./indexOf');
 
 
 /**
  * Returns `true` if the specified value is equal, in `R.equals` terms,
  * to at least one element of the given list; `false` otherwise.
  *
+ * Dispatches to the collection's `contains` method if applicable. If the
+ * collection does not have a `contains` method but does have an `indexOf`
+ * method, this function dispatches to `indexOf` and returns `true` if the
+ * return value is greater than or equal to zero; `false` otherwise. As a
+ * result, this function can be used to determine whether a given string
+ * contains a particular substring.
+ *
  * @func
  * @memberOf R
  * @category List
  * @sig a -> [a] -> Boolean
  * @param {Object} a The item to compare against.
- * @param {Array} list The array to consider.
- * @return {Boolean} `true` if the item is in the list, `false` otherwise.
+ * @param {Array} xs The collection to consider.
+ * @return {Boolean} `true` if the item is in the collection, `false` otherwise.
  *
  * @example
  *
  *      R.contains(3, [1, 2, 3]); //=> true
  *      R.contains(4, [1, 2, 3]); //=> false
  *      R.contains([42], [[42]]); //=> true
+ *      R.contains('am', 'Ramda'); //=> true
  */
-module.exports = _curry2(_contains);
+module.exports = _curry2(function contains(a, xs) {
+  return _hasMethod('contains', xs) ? xs.contains(a) : indexOf(a, xs) >= 0;
+});
--- a/test/contains.js
+++ b/test/contains.js
@@ -28,6 +28,64 @@ describe('contains', function() {
     assert.strictEqual(R.contains(new Just([42]), [new Just([42])]), true);
   });
 
+  it('dispatches to `contains` method', function() {
+    function Empty() {}
+    Empty.prototype.contains = R.always(false);
+
+    function List(head, tail) {
+      this.head = head;
+      this.tail = tail;
+    }
+    List.prototype.contains = function(x) {
+      return this.head === x || this.tail.contains(x);
+    };
+
+    var list = new List('foo', new List('bar', new List('baz', new Empty())));
+
+    assert.strictEqual(R.contains('foo', list), true);
+    assert.strictEqual(R.contains('bar', list), true);
+    assert.strictEqual(R.contains('baz', list), true);
+    assert.strictEqual(R.contains('quux', list), false);
+  });
+
+  it('dispatches to `indexOf` method', function() {
+    function Empty() {}
+    Empty.prototype.indexOf = R.always(-1);
+
+    function List(head, tail) {
+      this.head = head;
+      this.tail = tail;
+    }
+    List.prototype.indexOf = function(x) {
+      var idx = this.tail.indexOf(x);
+      return this.head === x ? 0 : idx >= 0 ? 1 + idx : -1;
+    };
+
+    var list = new List('b',
+               new List('a',
+               new List('n',
+               new List('a',
+               new List('n',
+               new List('a',
+               new Empty()))))));
+
+    assert.strictEqual(R.contains('a', list), true);
+    assert.strictEqual(R.contains('b', list), true);
+    assert.strictEqual(R.contains('c', list), false);
+    assert.strictEqual(R.contains('a', 'banana'), true);
+    assert.strictEqual(R.contains('b', 'banana'), true);
+    assert.strictEqual(R.contains('c', 'banana'), false);
+    assert.strictEqual(R.contains('nana', 'banana'), true);
+  });
+
+  it('does not dispatch to `indexOf` if `contains` is defined', function() {
+    function Empty() {}
+    Empty.prototype.contains = R.always(false);
+    Empty.prototype.indexOf = R.always(42);
+
+    assert.strictEqual(R.contains('foo', new Empty()), false);
+  });
+
   it('is curried', function() {
     assert.strictEqual(typeof R.contains(7), 'function');
     assert.strictEqual(R.contains(7)([1, 2, 3]), false);


test/contains.js
```diff
@@ -28,6 +28,36 @@ describe('contains', function() {
     assert.strictEqual(R.contains(new Just([42]), [new Just([42])]), true);
   });
 
+  it('dispatches to `indexOf` method', function() {
```
buzzdecafe: is this name mismatch a first? it strikes me as odd in this case. The dispatching behavior simply falls out of the implementation. But `contains` returns a `Boolean` while `indexOf` returns a `Number`. It's not a match.

davidchambers: I don't see this as being different from `R.take` dispatching to `slice`.

davidchambers: The fact that `indexOf` returns a number is not a problem. If the return value is greater than or equal to zero, the list/string contains the value/substring.

buzzdecafe: yes, i understand that. what i am squeamish about is exposing the implementation of `contains` in a test like this (same problem for `take`). until now dispatch has been of similar function types. e.g. anything that implements functor interface (`(a -> b) -> f a -> f b`), we explicity support.

CrossEye: Squeamish is a good word for it.  I feel this way too.  It's saying, "Hey, we have a function `foo:: a -> b`, and if you want us to dispatch to your type, then your type needs to implement a method `bar:: obj -> c`."  Something doesn't feel right.

buzzdecafe: i think this test should be removed. i'm fine with this implementation of `contains`, but this test lifts the skirt too much.

davidchambers: Are you suggesting this be an undocumented feature which we're free to remove at any time?

Derivability is mentioned in the Fantasy Land specification in several places. [Applicative][1], for example, does not require `map` as it can be derived from `of` and `ap`:

> A value which satisfies the specification of an Applicative does not need to implement:
>
>   - Functor's `map`; derivable as `function(f) { return this.of(f).ap(this); })}`

We could *also* dispatch to `contains`, but it seems completely reasonable to me to state that a type which provides `indexOf` but not `contains` is compatible with `R.contains`.


[1]: https://github.com/fantasyland/fantasy-land#applicative

CrossEye: You're right.  This is a good point.  (Of course, I've had my qualms about the derivable stuff in FL too, wondering if their API is simply too granular, but that's besides the point at the moment.)

> We could *also* dispatch to `contains`

I think I would prefer that.  If a type provide both `contains` and `indexOf`, we should definitely prefer to dispatch to the `contains` version.

davidchambers: I've updated this pull request to dispatch to `contains` if applicable. Note also the updated docstring.

src/contains.js
```diff
@@ -1,23 +1,34 @@
...
  * Returns `true` if the specified value is equal, in `R.equals` terms,
  * to at least one element of the given list; `false` otherwise.
  *
+ * Dispatches to the collection's `contains` method if applicable. If the
+ * collection does not have a `contains` method but does have an `indexOf`
+ * method, this function dispatches to `indexOf` and returns `true` if the
+ * return value is greater than or equal to zero; `false` otherwise. As a
+ * result, this function can be used to determine whether a given string
+ * contains a particular substring.
+ *
```
paldepind: Splendidly documented :)

kedashoe:
While I don't know that I really mind all the dispatch stuff, and the docs are very nice, still seems a bit un-ramda to me. Why is this simple, but checking if a string is passed to `R.path` and then splitting on `.`s to create an array complicated? To me the latter is simpler (and can be very clearly documented as well). Seems like dispatch gets a free pass for some reason?

src/contains.js
davidchambers: Thanks!

davidchambers:
> Seems like dispatch gets a free pass for some reason?

I'm starting to see that dynamic dispatch is not a nice extra: it's the very core of Ramda (or should be).

Ideally, every Ramda function would simply dispatch: there would be no code to speak of (except perhaps code to derive functions from other functions).

Unfortunately `R.map` cannot simply dispatch to `Array.prototype.map` because this method passes *three* arguments to the transformer. So, Ramda provides a compatible implementation.

Unfortunately `R.chain` cannot simply dispatch to `Array.prototype.chain` because there's no such method. So, Ramda provides an implementation for Array.

One can view Ramda as a thin layer that exposes methods as curried functions, with the majority of the code devoted to implementing Array methods to make these functions useful.

When viewed this way, dispatch is not a free pass: dispatch is *all there is*.

test/contains.js
```diff
@@ -28,6 +28,64 @@ describe('contains', function() {
...
+
+    var list = new List('foo', new List('bar', new List('baz', new Empty())));
+
+    assert.strictEqual(R.contains('foo', list), true);
+    assert.strictEqual(R.contains('bar', list), true);
+    assert.strictEqual(R.contains('baz', list), true);
+    assert.strictEqual(R.contains('quux', list), false);
+  });
+
+  it('dispatches to `indexOf` method', function() {
```
buzzdecafe: This test does not belong. this locks in the implementation of `contains` to forever be dependent on `indexOf`. The definition of `contains` is "returns true if the element is inside the container," _not_ "returns true if the index of the element in the container is greater than zero."

paldepind: If this PR gets merged this test _has_ to be there. It is written in the documentation that the function dispatches to `indexOf`. It is in the title of the PR. It is the entire point of the PR. It changes the definition of _contains_ to expose what would otherwise be an implementation detail.

We can not deliberately add a certain behavior to a function _and_ document it without also testing it. Otherwise someone might change this in half a year for some reason and break both the documentation and a lot of peoples code without any tests failing.

buzzdecafe:
> If this PR gets merged this test has to be there. It is written in the documentation that the function dispatches to indexOf. 

that should _also_ be removed.

> Otherwise someone might change this in half a year for some reason and break both the documentation and a lot of peoples code without any tests failing.

i'm very surprised to be meeting resistance to hiding implementation details! Isn't that fundamental?

This is _dispatch abuse_: Making use of implementation details for dispatch is a very dangerous precedent. The point of dispatch is to use functions on similar data types, so `Array.map` can dispatch comfortably to some other `Functor.map` because the behavior is the same at some high level. In this test we are saying `contains :: x -> xs -> Boolean` will dispatch to your object's `indexOf :: x -> xs -> Number`. That is just plain wrong. This test is relying on `contains` being implemented in a particular way that is separate from the *actual contract* of `contains`.

paldepind:
> that should also be removed.

Then it sounds to me like you're against the whole idea of this PR.

> This is dispatch abuse: Making use of implementation details for dispatch is a very dangerous precedent.

I agree.

I'd like to mention that I really don't like huge parts of the dispatch mechanism at all. It seems like a complex novelty feature to me. And the whole idea is very fragile since it currently builds on an idea of one to one mapping. It seems like we're departing from that idea which in turn makes the dispatch mechanism more complex and intervened.

Dispatching on the the fantasy land specification is fine with me. But the way I see it we're focusing too much on dispatching and too little on actually making any use of fantasy land. We're not utilizing fantasy land the way it was intended. `R.lift` for instance should work on any fantasy land applicative – but it doesn't.

Maybe @davidchambers can enlighten me on this but I don't see the practical benefit of all this dispatching beyond the fantasy land specification. Has the development of this been driven by a desire from actual use cases? It looks to me like it's currently there just because it _might_ be useful. I mean, how many interfaces do we know of that implements `indexOf`? If `String` is the only one then we should just add string support to `contains`, `take`, `drop`, etc.

All this dispatching would be more useful if it was more controlled and there was an actual specification of it. Something like "implement these 12 methods and Ramda guarantees that all these 25 functions will work on you data type". Without such a specification all this just look way too ad-hoc to me.

buzzdecafe:
> Then it sounds to me like you're against the whole idea of this PR.

No, changing the implementation to something cleaner/simpler is fine with me. Forever coupling `contains` to this particular implementation is what I am objecting to.

> All this dispatching would be more useful if it was more controlled and there was an actual specification of it. Something like "implement these 12 methods and Ramda guarantees that all these 25 functions will work on you data type". Without such a specification all this just look way too ad-hoc to me.

i agree with that. We originally included dynamic dispatch just to work with FL types. Then we extended it to integrate transducers. Now some cases have slipped in that are _ad hoc_. Before it gets too hairy we should lay down some groundrules.

paldepind:
> No, changing the implementation to something cleaner/simpler is fine with me. Forever coupling contains to this particular implementation is what I am objecting to.

This is not a cleaner nor simpler implementation. This is a slightly more complex implementation with the sole purpose of dispatching to `indexOf`.

> i agree with that. We originally included dynamic dispatch just to work with FL types. Then we extended it to integrate transducers. Now some cases have slipped in that are ad hoc. Before it gets too hairy we should lay down some groundrules.

Yes!

buzzdecafe:
> This is not a cleaner nor simpler implementation. This is a slightly more complex implementation with the sole purpose of dispatching to indexOf.

You may be right. This PR has the benefit of removing a couple of awkward String functions, but at the cost of potentially conveying the impression that `R` funtions treat Strings as lists.

we should probably start a new thread to discuss dispatch. But at present, all I am sure of is I want to support FL and transducers point-free.

paldepind:
> This PR has the benefit of removing a couple of awkward String functions, but at the cost of potentially conveying the impression that R funtions treat Strings as lists.

Either I'm missing something or you're confusing this PR with #1159.

CrossEye:
> Has the development of [dispatch] been driven by a desire from actual use cases? It looks to me like it's currently there just because it _might_ be useful.

I think this started with me.

I've long thought about how Ramda might be able to offer a functional programming gloss on other styles of API.  The way I want to work is through composing functions.  But that doesn't work well with an `object.method(...params)` API.  Adding a dispatch mechanism that maps `method => func` and offers `func(...params, object)` in its place gives us a great way to use other toolkits inside Ramda.  `invoker` was designed for this.  Our dispatching is designed for this.

Before I'd heard of Highland, perhaps before it came out, I was considering the idea that Ramda should not just work with lists, but also with streams.  But I knew there were already perfectly good tools for dealing with streams; I didn't want to rewrite them, only to be able to integrate.  Pairing that with the desire to integrate such things as `map` and FantasyLand's Functor interface, the notion of our dispatch became fairly important.

David's point above, that FantasyLand does something akin to its own internal implicit dispatching by optionally deriving certain of the required methods from others, adds a serious wrinkle.  Doing it here is interesting. I'm not thrilled with it, as I certainly buy Mike's statement that "The definition of `contains` is 'returns true if the element is inside the container,' _not_ 'returns true if the index of the element in the container is zero or greater.'"  But on the other had, it makes sense as an internal implementation for Ramda's array-based lists: why in the world wouldn't we offer the same capability to those who want to use it?  The only objection I can imagine would be a case where some type offers an `indexOf` that doesn't in fact imply a `contains` relationship.  It's conceivable, of course, something to do with book indices, for instance, but I'm having a hard time imagining details.

So, I guess my preference would be to defer to Mike's strong objection to the `indexOf` delegation.  If we want to do the FantasyLand-style "you can implement `foo` by implementing `bar`", we should do so in a coordinated fashion and not as part of simply changing an internal implementation.

buzzdecafe:
> Either I'm missing something or you're confusing this PR with #1159.

yep. _meum et confusionem meam_

davidchambers:
> All this dispatching would be more useful if it was more planned and there was an actual specification of it. Something like "implement these 12 methods and Ramda guarantees that all these 25 functions will work on your data type". Without such a specification all this just look way too ad-hoc to me. With such a specification however I could be a way more useful feature.

I agree. I believe dynamic dispatch is one of Ramda's two key features (along with curried functions). It deserves a lot more attention in the documentation. I have some ideas:

  - We should include metadata in a function's docstring which indicates whether it dispatches and whether, if dispatching is not possible, it is implemented in terms of one or more other functions which *themselves* dispatch. For example:

    ```
     * @dispatches to `contains`
     * @derivable from `indexOf`
    ```

  - The metadata above should be prominently exposed in the rendered documentation.

  - Ideally it would be possible to view all these metadata at once. When implementing a List type, for example, one should be able to see all the methods one should implement:

      * `R.contains` requires:
          - `contains :: List a ~> a -> Boolean`; or
          - `R.indexOf` compatibility.

      * `R.filter` requires:
          - `filter :: List a ~> (a -> Boolean) -> List a`.

      * `R.indexOf` requires:
          - `indexOf :: List a ~> a -> Number`.

      * `R.map` requires:
          - `map :: List a ~> (a -> b) -> List b`.

      * `R.reject` requires:
          - `R.filter` compatibility.

> Maybe @davidchambers can enlighten me on this but I don't see the practical benefit of all this dispatching beyond the fantasy land specification. Has the development of this been driven by a desire from actual use cases?

Absolutely. Thanks to dynamic dispatch I can now do this:

```javascript
> R.filter(n => n >= 0, S.Just(42))
Just(42)
> R.filter(n => n >= 0, S.Just(-1))
Nothing()
> R.reject(n => n < 0, S.Just(42))
Just(42)
> R.reject(n => n < 0, S.Just(-1))
Nothing()
> S.or(S.Nothing(), S.Just('foo'))
Just('foo')
> S.or(S.Just('bar'), S.Just('baz'))
Just('bar')
```

In plaid/sanctuary#45 I'm considering adding an Int type to Sanctuary, at which point functions for the mathematical operators will need to provide dynamic dispatch as well. I'd like to leverage Ramda's functions, but this is only an option if they allow the Int type to define its own `add`, `divide`, et cetera.

kedashoe:
> I agree. I believe dynamic dispatch is one of Ramda's two key features (along with curried functions).

Interesting. At a quick glance I don't see dispatching (or fantasy-land) mentioned anywhere in the README or even one step away in any of the README links (why ramda, etc). It certainly looks like currying and parameter order were an inspiration from the beginning while dispatching was a "cool, this is easy enough to do and we can become FL compliant". I was actually thinking of the other direction, providing a ramda with no dispatching at all and creating a new repo built on top of ramda that codifies all the dispatching semantics. Could get creative with aliases and all that without worrying about interfering with code that doesn't want to dispatch (the book class the defines indexOf, etc). Hard to see it happening at this point, just thinking out loud.

CrossEye:
>  I was actually thinking of the other direction, providing a ramda with no dispatching at all and creating a new repo built on top of ramda that codifies all the dispatching semantics.

I was speaking the other day with @buzzdecafe about the possibilities of metadata, and one of the things that I think would be phenomenal would be if some features like the dispatching could be codified in JSDoc-style comments, much like @davidchambers suggested above, but that this drives, at build time, the actual implementation of these features.  So if there is a wrapper to add dispatching, it's added on top of our basic implementation.  But that would allow us to do as you suggest, directly from inside the build system, offer a build which simply doesn't include the dispatching and another that does.

There's a flaw in this I don't know how to get around yet, which is that the one-off functions are included directly without going through the sort of build our main distribution goes through, but I think this might be worth pursuing.

davidchambers:
Where do we stand on this? It's worth noting that `R.reject` is defined in terms of `R.filter`, so we can already do this:

```javascript
> R.reject(n => n < 0, S.Just(42))
Just(42)
> R.reject(n => n < 0, S.Just(-1))
Nothing()
```

paldepind:
> Absolutely. Thanks to dynamic dispatch I can now do this:

That is certainly nice! But all of that is mostly possible because you're very familiar with the internals of Ramda. If you weren't figuring out how to make your data types work with Ramda would require you to manually go through all the implementations. And there would still be no big picture.

I certainly see much bigger potential in all of this after @davidchambers explanations in this thread. I've never been giving this much thought so I didn't see the potential.

But I would like to stress that these specifications should at least roughly be worked out before we start dispatching like crazy and before we start deriving a ton of functions in terms of other dispatching functions. But _if_ we do that I think this could be quite nice.

I don't really like fantasy land. It appears that there is no desire at all to make the specification understandable and meaningful to a wider audience. When I implemented the fantasy land specification for Flyd my primary question was simply "what's in it for me?". How does my library benefit from implementing this? I asked the question on the fantasy land Gitter room and was linked directly to the source code of a `lift` implementation in a mostly undocumented library. Today I still see implementing fantasy land as mostly useless.

If Ramda had some kind of specification then the advantage of implementing it would be much more obvious. Your data type would work with a comprehensive, approachable and high quality functional library. But defining a nice specification for something like a "collection" requires consideration (and probably trial and error as well)!

buzzdecafe:
> these specifications should at least roughly be worked out before we start dispatching like crazy and before we start deriving a ton of functions in terms of other dispatching functions.

:100:

davidchambers:
This pull request was based on the misconception that we're unwilling to add any String-specific logic to the definitions of our list functions. In fact, we're willing to do so, so we needn't define everything in terms of methods which happen to exist on both `Array.prototype` and `String.prototype`.

PR #1161, opened 2015-06-06
Improve the performance of curryN/curry

benjycui:
We can improve the performance of `curryN`/`curry` with `_curry1`. Let's look at the performance.

Before:

curry | Count | Hertz
-------|----------|---------
_x4(100) | 592421 | 9868550.23950353 
mult4(100) | 152955 | 2609371.9307046887
manual(100) | 4406550 | 76401537.32172365

After:

curry | Count | Hertz
-------|----------|---------
_x4(100) | 628122 | 11911507.368938653 
mult4(100) | 226094 | 4277420.856223587 
manual(100) | 4400142 | 82213774.10192372

Also, I try to use `_curry2` and `_curry3` to improve the performance. But I found that we have to modify `_curry2` and `curry3` in order to pass test case, and those modifications will slow down `_curry2` and `_curry3` a lot.

### DIFF
--- a/src/curryN.js
+++ b/src/curryN.js
@@ -1,4 +1,5 @@
 var _arity = require('./internal/_arity');
+var _curry1 = require('./internal/_curry1');
 var _curry2 = require('./internal/_curry2');
 var _curryN = require('./internal/_curryN');
 
@@ -47,5 +48,8 @@ var _curryN = require('./internal/_curryN');
  *      g(4); //=> 10
  */
 module.exports = _curry2(function curryN(length, fn) {
+  if (length === 1) {
+    return _curry1(fn);
+  }
   return _arity(length, _curryN(length, [], fn));
 });
--- a/src/internal/_curry1.js
+++ b/src/internal/_curry1.js
@@ -13,7 +13,7 @@ module.exports = function _curry1(fn) {
     } else if (a != null && a['@@functional/placeholder'] === true) {
       return f1;
     } else {
-      return fn(a);
+      return fn.apply(this, arguments);
     }
   };
 };


CrossEye:
I'd like to know a little more about the actual test cases, especially as the change is performance varies so widely between them.  Are these tests comparing only the calls to `curry` itself or the calls to the curried function?

I'm also curious about why you're going about this.  Do you think `curry` is a bottleneck in code that uses Ramda?  I certainly don't mind performance-focused PRs, but I always want to know **why**, and I always want to balance them against other  important factors in our system, such as API cleanliness and stability (no change here) and code simplicity (only a very minor issue with this PR.)  But first and foremost, what is driving this?

benjycui:
I just run `npm run bench`. Actually, I have not met performance problems in Ramda, but this idea pop up while reading the source code of Ramda. Sorry to trouble you :-)

paldepind:
I have thought of the exact same thing myself.

What changes did you have to make to `_curry2` and `_curry3` for it to work?

CrossEye:
> I just run `npm run bench`. 

Oh, I simply didn't recognize our test case names!  :-)

> Actually, I have not met performance problems in Ramda, but this idea pop up while reading the source code of Ramda.  Sorry to trouble you :-)

Not at all.  Not only is it no bother, it's absolutely worth considering.  I was wondering exactly this, whether it was the code that was troubling you or the actual performance.  It's clearly time that we take on `curry` and try to get it in as good a shape as we can.

benjycui:
@CrossEye all of you are so kind :-) Actually, I am considering to recommend FP to my teammate, and Ramda is a cool library to start from.So I am trying to make it clear.

@paldepind take `_curry2` for an example. My solution will modify 4 lines of codes in order to forward extra arguments(To pass test case, 'forwards extra arguments'. See: `path-to/ramda/test/curryN.js:68`).
```js
var _slice = Array.prototype.slice;
function _curry2(fn) {
  return function f2(a, b) {
    var n = arguments.length;
    if (n === 0) {
      return f2;
    } else if (n === 1 && a === __) {
      return f2;
    } else if (n === 1) {
      return _curry1(function(b) { return fn.apply(this, [a].concat(_slice.call(arguments))); });  // (1)
    } else if (n === 2 && a === __ && b === __) {
      return f2;
    } else if (n === 2 && a === __) {
      return _curry1(function(a) { return fn.apply(this, [a].concat(b).concat(_slice.call(arguments, 1))); });  // (2)
    } else if (n === 2 && b === __) {
      return _curry1(function(b) { return fn.apply(this, [a].concat(_slice.call(arguments))); });  // (3)
    } else {
      return fn.apply(this, arguments);  // (4)
    }
  };
};
```
I have not test the performance of this solution, but my instinct told me that it is dirty and slow XD. Maybe you have better solution.

davidchambers:
What's the plan here?

CrossEye:
I have no problem merging this.  It looks like a nice little optimization.  Are there reasons we should not do it?

davidchambers:
I don't object. @benjycui, could you `git rebase master`?

benjycui:
Sorry for the delay, and it is OK now.

PR #1162, opened 2015-06-06
simplify primary  path for keys

buzzdecafe:


CrossEye:
Yes!

Nice fix.

:herb:

PR #1164, opened 2015-06-08
lint: replace uses of ++ and -- operators

davidchambers:
I believe these operators should be avoided.

The existence of both prefix and postfix operators is a source of confusion. One might write `x++` when one means `++x`, or vice versa. Understanding code which uses both prefix and postfix forms requires more mental effort than understanding code which uses just one form.

Postfix `++` and postfix `--` are particularly tricky: they have side effects but are only useful in "expressions", since in "statements" the prefix forms may be used instead. Avoiding the postfix forms makes side effects more apparent.

Prefix `++` and prefix `--` are sugar for `+= 1` and `-= 1`, but promote unnecessarily tricky code (such as initializing a loop variable to `-1` rather than `0`). Banning these operators means there is only one way to increment a variable and only one way to decrement a variable.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -6,6 +6,7 @@
     "Symbol": true
   },
   "newcap": false,
+  "plusplus": true,
   "predef": ["beforeEach", "console", "define", "describe", "it", "module", "process", "require"],
   "undef": true,
   "unused": true
--- a/lib/sauce/ie8-shim.js
+++ b/lib/sauce/ie8-shim.js
@@ -26,7 +26,7 @@ if (typeof document.getElementsByClassName !== 'function') {
     } else {
       var rightClass = new RegExp('(^| )' + className + '( |$)');
       seek = document.getElementsByTagName(tag);
-      for (i = 0; i < seek.length; i++) {
+      for (i = 0; i < seek.length; i += 1) {
         if (rightClass.test((node = seek[i]).className)) {
           result.push(seek[i]);
         }
--- a/src/addIndex.js
+++ b/src/addIndex.js
@@ -29,10 +29,14 @@ var curryN = require('./curryN');
  */
 module.exports = _curry1(function(fn) {
   return curryN(fn.length, function() {
-    var idx = -1;
+    var idx = 0;
     var origFn = arguments[0];
     var list = arguments[arguments.length - 1];
-    var indexedFn = function() {return origFn.apply(this, _concat(arguments, [++idx, list]));};
+    var indexedFn = function() {
+      var result = origFn.apply(this, _concat(arguments, [idx, list]));
+      idx += 1;
+      return result;
+    };
 
     return fn.apply(this, _prepend(indexedFn, _slice(arguments, 1)));
   });
--- a/src/aperture.js
+++ b/src/aperture.js
@@ -19,11 +19,12 @@ var _slice = require('./internal/_slice');
  *      R.aperture(7, [1, 2, 3, 4, 5]); //=> []
  */
 module.exports = _curry2(function aperture(n, list) {
-  var idx = -1;
+  var idx = 0;
   var limit = list.length - (n - 1);
   var acc = new Array(limit >= 0 ? limit : 0);
-  while (++idx < limit) {
+  while (idx < limit) {
     acc[idx] = _slice(list, idx, idx + n);
+    idx += 1;
   }
   return acc;
 });
--- a/src/composeL.js
+++ b/src/composeL.js
@@ -28,10 +28,11 @@ var _composeL = require('./internal/_composeL');
  *      secondOfXOfHeadLens.set(123, source); //=> [{x: [0, 123], y: [2, 3]}, {x: [4, 5], y: [6, 7]}]
  */
 module.exports = function() {
-  var idx = arguments.length - 1;
-  var fn = arguments[idx];
-  while (--idx >= 0) {
+  var fn = arguments[arguments.length - 1];
+  var idx = arguments.length - 2;
+  while (idx >= 0) {
     fn = _composeL(arguments[idx], fn);
+    idx -= 1;
   }
   return fn;
 };
--- a/src/cond.js
+++ b/src/cond.js
@@ -26,11 +26,12 @@
 module.exports = function cond() {
   var pairs = arguments;
   return function() {
-    var idx = -1;
-    while (++idx < pairs.length) {
+    var idx = 0;
+    while (idx < pairs.length) {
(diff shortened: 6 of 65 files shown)


CrossEye:
28 minutes.  You beat me- to this by 28 minutes!

This looks good to me at a quick once-over.  I'll look again in the morning.

thefourtheye:
Why can't we do `(idx = idx + 1)`? That would save a lot of lines and still do the same thing, right?

paldepind:
I don't see the advantage of this. An increase in lines of count by over 80 lines clearly shows that this is a more bulky and less compact style.

`++`/`--` are splendid operators. They are compact and captures the idea of _incrementing_ more clearly than `+= 1`. I think they can be misused but they can also make coder shorter, more to the point and easier to read.

I am especially opposed to using `+= 1` in `for` loops where using `++` is very established.

> such as initializing a loop variable to -1 rather than 0

`++` is not the reason for that. Ignoring `for` loops is!

davidchambers:
> Why can't we do `(idx = idx + 1)`? That would save a lot of lines and still do the same thing, right?

Removing side effects from predicates makes the code clearer. The increase in the number of lines of code is a cost I'm willing to pay. (Were we to switch to `for` loops we could have our :cake: and eat it too.)

> An increase in lines of count by over 80 lines clearly shows that this is a more bulky and less compact style.

That is true by definition. :P

> `++`/`--` are splendid operators.

Which ones? Prefix, postfix, or both?

> such as initializing a loop variable to -1 rather than 0
>
> > `++` is not the reason for that. Ignoring `for` loops is!

Let's not conflate the two issues. (For the record, I agree with you that `for` loops offer objective benefits.)

Assume we're operating in a world in which `while` is the only loop construct. Then, assume the world lacks all four of the `++` and `--` operators. How would one access each of the elements of an array in turn? Perhaps:

```javascript
var idx = 0;
while (idx < xs.length) {
  xs[idx];
  idx += 1;
}
```

Of course, one *could* write:

```javascript
var idx = -1;
while ((idx += 1) < xs.length) {
  xs[idx];
}
```

The fact that `idx += 1` must be wrapped in parens in this context makes it more natural to separate the side effect from the predicate: a very good thing in my view.

Now imagine we have access to the four unary increment and decrement operators, but are still limited to `while` loops. One might write:

```javascript
var idx = -1;
while (++idx < xs.length) {
  xs[idx];
}
```

The fact that prefix `++` allows us to increment the loop variable in the predicate without requiring parens makes it feel natural to do so. But what has happened? Our predicate now contains a side effect, and the side effect requires us to initialize the loop variable with a misleading value.

I agree, @paldepind, that we should use `for` loops where appropriate, but in a world without `for` loops I favour deconstructed `for` loops:

```javascript
<x>;
while (<y>) {
  <a>;
  <z>;
}
```

The above is equivalent to:

```javascript
for (<x>; <y>; <z>) {
  <a>;
}
```

Our current approach is to combine `<y>` and `<z>`, but this is indefensible. If compactness is the goal, a `for` loop would serve us better. If clarity is the goal, evaluating the predicate should not have side effects. (Note that `for` loops also satisfy this requirement, since `<y>` and `<z>` are separate.)

paldepind:
> Which ones? Prefix, postfix, or both?

I prefer the prefix version since they're simpler. The postfix variants are more convoluted but they do have their places.

> Let's not conflate the two issues.

Exactly my point!

I do see the advantage of getting rid of the `idx = -1`. That is a huge plus in readability!

It's banishing `++` and `--` that I'm opposed to. Can you please explain what harm you see in them _when they appear in a statement of their own_? I.e. not inside a bigger statement like they currently do in our `while` loops.

davidchambers:
> I do see the advantage of getting rid of the `idx = -1`. That is a huge plus in readability!

Excellent. I believe you, @CrossEye, and I are all in agreement on this point.

> Can you please explain what harm you see in them *when they appear in a statement of their own*?

There's <del>no</del> <ins>not much</ins><sup>†</sup> harm in this (which I consider responsible use). I don't mind the `<z>` part of a `for` loop being `++idx`, but there's no way to configure JSHint to only allow these operators in "statements". I'd prefer to disallow responsible use than to allow irresponsible use, especially given that `idx += 1` is itself rather convenient.

---

† Although I'm from New Zealand, I like to think I'm half Dutch:

```console
$ python -c 'import this' | ack 'only one' --after-context 1
There should be one-- and preferably only one --obvious way to do it.
Although that way may not be obvious at first unless you're Dutch.
```

paldepind:
> I'd prefer to disallow responsible use than to allow irresponsible use, especially given that idx += 1 is itself rather convenient.

If a feature can be both potentially useful and harmful then I prefer allowing it to disallowing it. Allowing it makes reaching the optimal solution possible and likely to happen while disallowing it makes it impossible. Also, this is Ramda and all PRs are peer reviewed. Crazy programmers are not running loose in the code base.

Regarding the "one obvious way" argument I see not conflict. `++` is the increment operator. It is the obvious choice when incrementing an index. To me `++` is more readable because it is more specific and thus communicates the intend better. In the same way that `map` is more specific than `reduce` and thus communicates the intend better even though `reduce` can always replace `map`.

I believe the reason why Python don't have a `++` operator has more to do with the fact that they have a much neater `for` loop and thus have a much much smaller use for such an operator. At which point it becomes an unnecessary complexity.

CrossEye:
> I believe the reason why Python don't have a `++` operator has more to do with the fact that they have a much neater `for` loop and thus have a much much smaller use for such an operator. At which point it becomes an unnecessary complexity.

So that's settled then?  We'll revisit this when Javascript gets a better `for` loop?

:smiling_imp:

paldepind:
> So that's settled then? We'll revisit this when Javascript gets a better for loop?

Yes! Let's adopt some Python wisdom. Like this:

> As the `for` loop in Python is so powerful, `while` is rarely used

– [From Pythons documentation on while loops](https://wiki.python.org/moin/WhileLoop)

The `for` loop in JavaScript is just as powerful so it applies equally well here.

davidchambers:
I'd like to open a pull request to resolve #1165, but don't want to do so until this is merged (or closed) to avoid merge conflicts.

paldepind:
If it wasn't for the global disallowance of `++` I would have nothing against this PR.

This `while` style is preferable to the current so if `++` is still open for discussion in the future I'm :+1: for merging.

CrossEye:
There is no clear-cut consensus that `++` is wrong, so I think we're still open to using it where available.  So with that said, I think we can merge now.

PR #1169, opened 2015-06-09
Remove uneeded voids from `arity`

paldepind:
All those `voids` are there to make JSHint happy. This PR just makes it relax a little while inside `arity`.


buzzdecafe:
:cow2:

CrossEye:
Damn, I would _swear_ we'd tried that and it didn't work for some reason!

:+1"

PR #1170, opened 2015-06-09
Handle empty objects and falsey values in `set`

paldepind:
[See this comment](https://github.com/ramda/ramda/pull/1088#issuecomment-110279037).

### DIFF
--- a/src/set.js
+++ b/src/set.js
@@ -22,7 +22,7 @@ var keys = require('./keys');
  * @return {Object} The merged object.
  * @example
  *
- *      R.set({age: 40, {name: 'fred', age: 10});
+ *      R.set({age: 40}, {name: 'fred', age: 10});
  *      //=> {'name': 'fred', 'age': 40}
  *
  *      R.set({parent: {age: 40}}, {name: 'fred', age: 10, parent: {age: 39}});
@@ -34,7 +34,7 @@ module.exports = _curry2(function set(update, original) {
   }
   var updateKeys = keys(update);
   var originalKeys = keys(original);
-  if (isEmpty(updateKeys) || isEmpty(originalKeys)) {
+  if (isEmpty(updateKeys) && isEmpty(originalKeys)) {
     return update;
   }
   var result = _extend({}, original);
--- a/test/set.js
+++ b/test/set.js
@@ -22,6 +22,22 @@ describe('set', function() {
     assert.deepEqual(b, {w: 1, x: {a: 5, b: 2, c: 3}, z: 4});
   });
 
+  it('handles empty object', function() {
+    var a = {w: 1, x: {a: 1, b: 2, c: 3}, z: 4};
+    var b = R.set({}, a);
+    assert.deepEqual(b, a);
+  });
+
+  it('handles nested empty object', function() {
+    var a = {w: 1, x: {a: 1, b: 2, c: 3}, z: 4};
+    var b = R.set({w: 0, x: {}, z: 3}, a);
+    assert.deepEqual(b, {w: 0, x: {a: 1, b: 2, c: 3}, z: 3});
+  });
+
+  it('ignores `null`', function() {
+    assert.deepEqual(R.set(null, {a: 1}), {a: 1});
+  });
+
   it('swallow copies when possible', function() {
     var a = {w: 1, x: {a: 1, b: 2, c: 3}, z: 4};
     var b = R.set({z: 5}, a);


test/set.js
```diff
@@ -22,6 +22,23 @@ describe('set', function() {
...
+  it('handles empty object', function() {
+    var a = {w: 1, x: {a: 1, b: 2, c: 3}, z: 4};
+    var b = R.set({}, a);
+    assert.deepEqual(b, a);
+  });
+
+  it('handles nested empty object', function() {
+    var a = {w: 1, x: {a: 1, b: 2, c: 3}, z: 4};
+    var b = R.set({w: 0, x: {}, z: 3}, a);
+    console.log(R.set(null, {a: 1}));
```
svozza: Stray debugging statement there.

test/set.js
```diff
@@ -22,6 +22,22 @@ describe('set', function() {
...
+    var b = R.set({}, a);
+    assert.deepEqual(b, a);
+  });
+
+  it('handles nested empty object', function() {
+    var a = {w: 1, x: {a: 1, b: 2, c: 3}, z: 4};
+    var b = R.set({w: 0, x: {}, z: 3}, a);
+    assert.deepEqual(b, {w: 0, x: {a: 1, b: 2, c: 3}, z: 3});
+  });
+
```
CrossEye: This becomes tricky here.  It's not so obvious to me that this is the desired result.  I think there's a good case to be made for

    {w: 0, x: {}, z: 3}

Or put another way, if the user want to use `set` to transform `{w: 1, x: {a: 1, b: 2, c: 3}, z: 4}` to ` {w: 0, x: {}, z: 3}`, what should she pass as the first parameter?

paldepind: I was unsure at first as well. But this looks consistent to me.

If you are fine with the previous test you accept this behaviour

```javascript
set({}, {a: 1}) //=> {}
```

Then you've already chosen to interpret `{}` as "change no keys". That meaning should be the same when `{}` appears nested to be consistent.

Another angle: The current `set` can _never_ remove properties. Missing properties generally means "no change". If this behavior was what you suggest `set` would get a whole new power. In a way that would be a very drastic deviation.

CrossEye: Well, I wasn't so sure I bought the previous test either.  But this one was easier to discuss.

This change will mean that you can't even do the following:

```js
    set({x: 42, z: 100}, {w: 1, x: {a: 2, b: 3}, y: 4}); 
    //=> want {w: 1, x: 42, y: 4, z: 100},
    //   get  {w: 1, x: {a: 2, b: 3}, y, 4, z: 100}
```

Before this change, that worked as expected.  You could replace whole objects with a new value.

But even if we fix that, we're left with

```js
    var a =  {w: 1, x: {a: 2, b: 3}, y: 4};
    var b = set({x: 42}, a)
    var c = set({x: {}}, b); //=> {w: 1, x: {}, y: 4}
    
    var d = set({x: {}}, a); //=> {w: 1, x: {a: 2, b: 3}, y: 4};
```

Which seems rather bizarre.

I did warn you that I thought there were prickers in deep merging!  :smiling_imp:

src/set.js
```diff
@@ -22,7 +22,7 @@ var keys = require('./keys');
  * @return {Object} The merged object.
  * @example
  *
- *      R.set({age: 40, {name: 'fred', age: 10});
+ *      R.set({age: 40, {name: 'fred', age: 10}});
```
buzzdecafe: missing `}` after 40

buzzdecafe: missing `}` after 40

test/set.js
paldepind: The current behavior is basically "you can't override an object with something that is not an object". This is consistent with the my previous remark that you can not remove properties. The old behavior did also make sense to me though. It understood `{}` as a "flat" value that could override other values.

kedashoe: This feels like a naming issue. What I think of as `set` is called `assoc` and `assocPath`, and what I think of as `merge` is now being called `set`? What was wrong with the name `merge`? I would definitely expect an operation called `set` to result in `{w: 0, x: {}, z: 3}`. And `merge` resulting in `{w: 0, x: {a: 1, b: 2, c: 3}, z: 3}` seems reasonable, especially with docs/examples/tests so everyone is aware of that behaviour.

paldepind: @kedashoe [The foregoing discussion is here](https://github.com/ramda/ramda/pull/1088). I'd like to hear what you think.

kedashoe: Not sure I have much to add beyond my previous comment :) If we're talking perfect world, I'd like to see

`set` -> `merge`
`assoc` -> `set`
`prop` -> `get`

but even if we're not at 1.0 yet having `prop` renamed to `get` might be a bit much to ask. So dropping the name `set` and un-deprecating/changing the current `merge` to this deep merge implementation of `set` seems reasonable to me.

CrossEye: @kedashoe: I agree that something seems to be going awry here with the naming.  `assoc` has a useful cognate in other languages, so seemed like a good choice, especially because at the time `set` sounded to me very much like it mutated the object.  I'm really not clear enough on the API here, nor on the name.  I will have to go back over the recent history to determine why we deprecated `merge`, as it doesn't seem to be covered by anything else.  But I think mostly I'm just confused at the moment.

CrossEye: I don't know why the previous two messages didn't show up before I posted my comment.... sorry.

CrossEye: From that other discussion, @paldepind, you said,

> I see this function as a more powerful and useful version of assoc and deepAssoc.

I like that notion better than this change implies.

```js
    set({x: {}, z: 100}, {w: 1, x: {a: 2, b: 3}, y: 4}); 
    //=> {w: 1, x: {a: 2, b: 3}, y, 4, z: 100}

    R.pipe(R.assoc('x', {}), R.assoc('z', 100))({w: 1, x: {a: 2, b: 3}, y: 4});
    //=> {"w": 1, "x": {}, "y": 4, "z": 100}
```

I would simply expect that for flat first parameters, `set` would be consistent with something like this:

```js
    var set = R.useWith(
        R.flip(R.reduce(R.flip(R.call))), 
        R.pipe(R.toPairs, R.map(R.apply(R.assoc))), 
        R.identity
    );
```

(not that I would use such an implementation directly, but something with that behavior, namely, doing the equivalent of calling `assoc` on each value.)

But this would definitely conflict with the notion that you can't remove properties / override an object with anything other than an object.

I think we still have a significant API issue here.

paldepind: If we choose to interpret `{}` as a "flat" value that can override objects just like a primitive can then I don't see the issue. This is consistent with the rule "only recurse into things with keys". But the best thing would probably be to interpret empty objects as "deep" values and only primitives as flat values.

I don't see any of this as "significant API issues". The different behaviors are all explainable. They make sense. We just need to figure which is most expectable/useful.

src/set.js
davidchambers: Let's fix this line, @paldepind, so the pull request is in a mergeable state. :)

paldepind: Will do.

davidchambers:
Closing since #1181 removed `R.set` and #1205 repurposed the name (sorry, @paldepind)! I believe we're still in favour of the functionality this function provides, but we'll need to think of a different name.

PR #1172, opened 2015-06-12
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3

Also their maintainer says they probably won't ever use it: http://stackoverflow.com/questions/24844901/bowers-bower-json-file-version-property

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,7 +1,6 @@
 {
   "name": "ramda",
   "main": "dist/ramda.js",
-  "version": "0.14.0",
   "homepage": "https://github.com/ramda/ramda",
   "authors": [
     "(Scott Sauyet <scott@sauyet.com>)",


buzzdecafe:
i think this may mean adjusting xyz, or the next `npm publish` will just put the version right back. any comment, @davidchambers ?

davidchambers:
If we remove it, I think [xyz][1] will reinstate it next time we publish. If you're interested, @kkirsche, you could open an xyz issue suggesting that it only include `version` in __bower.json__ if already present.


[1]: https://github.com/davidchambers/xyz

kkirsche:
Issue made

kkirsche:
@CrossEye closed this because....?

CrossEye:
Perhaps I misunderstood.  I thought your "Issue made" meant that you'd made an issue over in [xyz][xy].  Since we'll be able to simply pick that up by upgrading our version of xyz, I thought that took care of it.  I'm more than willing to reopen it, if that's incorrect.  In fact I would just do so, but the submitting repo is gone.

  [xy]: https://github.com/davidchambers/xyz

davidchambers:
The sequence could have been:

1.  Merge davidchambers/xyz#19
2.  Publish xyz@0.6.0
3.  Update Ramda's xyz dependency
4.  Merge this pull request

Instead (now that we cannot merge this pull request) we can combine steps 3 and 4. There are actually two other changes I'd like to get into the next xyz release, though, so I don't expect to get to step 2 this weekend.

Issue #1151, opened 2015-06-03
compare


TGOlson:
Any thoughts on a `compare` function?

```
compare :: (a -> Number) -> a -> a -> Number
```

```js
var compare = R.curry(function(f, x, y) {
  return f(x) - f(y);
});
```

Basically, shorthand for converting `(a -> Number)` functions into comparators. I've found it helpful for things like this:

```js
var compareLength = compare(R.length);

R.sort(compareLength, ['hey', 'hi', 'hello', 'cya']);
// => ["hi", "hey", "cya", "hello"]
```

It could also be accompanied by a `sortByComparing` function.

```js
var sortByComparing = R.curry(function(f, xs) {
  return R.sort(compare(f), xs);
});

R.sortByComparing(R.length, ['hey', 'hi', 'hello', 'cya']);
// => ["hi", "hey", "cya", "hello"]
```

CrossEye:
`R.sortBy` does this in a more generic fashion:

```js
R.sortBy(R.length, ['hey', 'hi', 'hello', 'cya']);
// => ["hi", "hey", "cya", "hello"]
```

The function passed to it could generate String or numeric keys; either would work fine, as would Dates.  Basically, anything that you could compare with `a < b` would work.

TGOlson:
:+1: 

Awesome, I guess I would confused by the `(a -> String)` part of the type signature in `sortBy` - should have given it a poke first though. Thanks for the reply.

CrossEye:
Most of our type signatures were created in a heroic effort by @buzzdecafe.  He did this over the course of only a day or two.  He got a surprisingly high percentage of them right, but we really need to look through them more carefully and decide where some polishing is due.

This one is slightly more odd.  I'm pretty sure that function was mine (else it's very similar to one I'd written earlier.)  And I honestly expected it to be used with Strings.  I never really thought about Numbers (or Dates) until this issue was raised.  So it was documented for how we expected the function to be used.  I think this one should be redocumented, because the implementation is clearly right: as long as the keys your function generates are sortable by `<`, it should just work.  Often things like this that fall out of the implementation are intentionally not documented, since we want to feel free to change the implementation, but this feels more like a case of learning from the implementation.

Issue #1037, opened 2015-04-20
Highland.js Integration


svozza:
We had a bit of a discussion [here](https://github.com/plaid/sanctuary/issues/16) about possibly integrating Highland streams into Ramda so I said I'd create an issue here explore the issue more fully. @caolan has been very busy recently so I'm not sure if he has time to get involved but after him @vqvu probably knows the code better than anyone else - not to mention @jeromew, @LewisJEllis and @quarterto - and I'm sure we can all help in some way. What do you have in mind and what would you need from our side?

caolan:
Ramda looks fun, I'll have to take it for a spin :) Seems a good match for Highland. Perhaps it would be a good opportunity to produce a more 'skinny' Highland core for you to integrate too...

buzzdecafe:
some background: we already can dynamically dispatch to non-list objects in list position. That's how we support transducers, and how we are implementing the fantasy-land spec. so in some (many?) ways highland may just drop right in with ramda and work, e.g., `map(add(1), stream)` should be the same as `stream.map(add(1))`. But I am not familiar enough with highland api to know how well integrated we are. also, we will probably need to expand the number of functions that dispatch in this fashion.

So I'd like to firm this up s.t. you can just drop it in and start using it point-free. 

full disclosure: I'm working on a relational programming thing that i want to recast to use streams. so this is personally useful to me :smile_cat:

svozza:
Yeah, `map` works fine but unfortunately `reduce` doesn't because Highland's version has the arguments in a different order `_.reduce(memo, iterator)`. Really don't know how we can get around that one.

vqvu:
Interesting...so we'll need to figure out a list of Highland <-> Ramda operator equivalencies. We'd also have to make sure their semantics match.

One possible issue is that sometimes the names don't match between the two libraries. For instance, `R#forEach === hl#each`. Perhaps we can add aliases on our side for some of these inconsistencies, and Ramda can do the `_dispatchable` magic.

Plus, due to the nature of the two libraries, Ramda implements more operators than Highland does (e.g., `findFirst`, `findLast`). Maybe we can dispatch to the native HL transform if it exists and use transducers if it doesn't? Or just use transducers everywhere (the native transform is probably slightly faster in the sync case).

kevinbeaty:
> but unfortunately reduce doesn't because Highland's version has the arguments in a different order:

[...]

PR #1173, opened 2015-06-12
fix signature of R.sortBy

davidchambers:
See https://github.com/ramda/ramda/issues/1151#issuecomment-108713420

This is the signature of Haskell's equivalent function, [`sortWith`][1].

[1]: http://hackage.haskell.org/package/base-4.8.0.0/docs/GHC-Exts.html#v:sortWith


buzzdecafe:
does the typeclass help in javascript context?

paldepind:
I think we need a section in the documentation that defines our invented type classes. Here `Ord` means anything that can be compared with javascripts `<` and `>` if I'm not mistaken. I've previously mentioned that I think some types would be better if we had a `Key` class meaning anything that can appear between square brackets when accessing properties `obj[a]`.

That section could probably contain a short primer on the type specifications themselves as well.

CrossEye:
@paldepind: Something like this is very much on my agenda for this weekend. 

@davidchambers:  l like the new @ sig tag, but don't think the new definition is enough to go on. _How_ dies it sort by this function?  Should it be a comparator?  Users who read mostly the descriptions and not the sigs should have more information than is supplied here.

buzzdecafe:
does this imply we should change the sig of `map` to `Functor f => (a -> b) -> f a -> f b`? that strikes me as more accurate (because of dispatch) *and* more obscure to new users who may not be familiar with Haskell.

CrossEye:
@buzzdecafe: I don't think it does.  Our dispatch does not actually do any validation that what it dispatches to acts anything like a functor.  We can dispatch to anything at all:

```js    
R.map(shiftBy(3, 5), {
  map: function(convert) {
    return convert(polarToCartesian(this));
  },
  r: 10,
  theta: Math.PI / 4
});
```

(http://bit.ly/1L52xlF)

I think we should have further documentation about dispatching.  But once we dispatch, even if we mean it for such things as Functors, really all bets are off.  You've voided the warranty.

buzzdecafe:
sure, but the *intent* is to use `map` in a generic way with any functor. I'm not concerned about edge cases in sigs. One could argue that the typeclassed sig is more descriptive of the intent.

davidchambers:
> does this imply we should change the sig of `map` to `Functor f => (a -> b) -> f a -> f b`?

I believe this should be included. One of my colleague was very confused when reading some code that used [`R.map`][1] in conjunction with Maybe values. The signature of `R.map` suggests that it only works with Array values (and the description does not mention that it dispatches to `map` if available).


[1]: http://ramdajs.com/docs/#map

buzzdecafe:
hey @davidchambers , it's nice to agree! I think if we're in for a penny, we're should be in for a pound.

davidchambers:
It feels good to be on your side of an argument for a change, @buzzdecafe!

CrossEye:
This is very interesting.  I've been feeling that the dispatching is meant to be much more generic than, say FantasyLand.  That has always struck me as just one possible use-case among many.  To my mind, we dispatch to any function that has a similar input signature to our function.  I wouldn't mind documenting other known cases.  But I don't see that as high priority.  I do see documenting that it _does_ dispatch as quite important, though.

paldepind:
> To my mind, we dispatch to any function that has a similar input signature to our function

But we don't have any idea about that. We dispatch and hope for the best. I'm just mentioning it.

buzzdecafe:
> But we don't have any idea about that. We dispatch and hope for the best. I'm just mentioning it.

i don't see how we could know. it's up to the user not to be a dope

paldepind:
> i don't see how we could know. it's up to the user not to be a dope

[They could be namespaced](https://github.com/fantasyland/fantasy-land/issues/92). I'm actually a bit confused that @CrossEye is embracing the idea in said issue since Ramda is doing exactly the same.

> The problem is that FantasyLand wants to claim ownership of all of these names: [ ... long list ... ]

In a way Ramda is claiming ownership over a lot more method names. It is saying: "if your object has these methods and you want to work with our dispatching you _must_ follow this function signature".

I generally don't like the idea of using sensible names for something that is a "specification". I don't want to implement a nice object oriented API. Both fantasy land and Ramda dispatching forces one to do that. The transducer specification got this right IMO.

CrossEye:
@paldepind: I've been seriously considering what it would take to write a pure-function as opposed to object-based version of the FantasyLand specification.  There are some serious wrinkles to consider: how do you get from instance to the appropriate type-creation function, for instance?  How do you deal with hierarchies of types, e.g. `add(Just(Integer(5)), Just(Number(2.6)))`: does this throw an error, or return `Just(Number(7.6))` in the case that the `Integer` type is a subset of the `Number` type?  And is this decision commutative?  Might `add(Just(Number(2.6)), Just(Integer(5)))` return a value even when the first one threw an error?  Would the types allow the same sort of derivation of `ap` from `chain` and `map`?  So these ideas are very much on my mind.

But I'm not quite sure how they would apply to Ramda.  If the FantasyLand team adopted the [proposal][fl] discussed above, I would expect that Ramda would jump on it quickly and dispatch `map` to `@@alg-types/map` if it were available.  But I would not expect that to stop us from dispatching to `map` if the object didn't have  `@@alg-types/map`.

> I generally don't like the idea of using sensible names for something that is a "specification". I don't want to implement a nice object oriented API. Both fantasy land and Ramda dispatching forces one to do that. 

I don't think of Ramda as a specification.  Ramda defines a number of functions, a great many of which work on lists.  We know that the general idea of those functions is often duplicated by other libraries, meant to work on other data types.  Ramda does at the moment, the simplest thing it can: it checks to see if you are passing it something that it can't operate on (a non-list) and then looks to see if it defines the operation itself.

Maybe it's a subtle difference, but to me it's very real.  This is not really a matter of Ramda specifying, "If you want to work with us, here's what you must implement."  Instead, it's Ramda saying, "If you happen to implement this, then Ramda can transform your OO API into a more FP style."


  [fl]: https://github.com/fantasyland/fantasy-land/issues/92

paldepind:
> I've been seriously considering what it would take to write a pure-function as opposed to object-based version of the FantasyLand specification.

I don't think that is possible.

> But I'm not quite sure how they would apply to Ramda. If the FantasyLand team adopted the proposal discussed above, I would expect that Ramda would jump on it quickly and dispatch map to @@alg-types/map if it were available. But I would not expect that to stop us from dispatching to map if the object didn't have @@alg-types/map.

I think we should stop dispatching to `map`.

> I don't think of Ramda as a specification.

Me neither. But in my opinion there _should_ be a specification.

> Maybe it's a subtle difference, but to me it's very real. This is not really a matter of Ramda specifying, "If you want to work with us, here's what you must implement." Instead, it's Ramda saying, "If you happen to implement this, then Ramda can transform your OO API into a more FP style."

To me the latter is worth very little.

The above was just some short remarks. I've rambled about this [before](https://github.com/ramda/ramda/pull/1160#issuecomment-110668630). I expect that we will probably never get the dispatch mechanism to work in any sensible way with any third party libraries except with a specification (not counting Fantasy Land and libraries made by Ramda people). The chances are extremely slim that there will just pop up a library that by luck happens to implement all the list methods we dispatch to with the exact same names, function signatures and behaviors.

As was mentioned in the fantasy land issue Bacon implements a map that is not strictly compatible with fantasy land. Many libraries will implement a `map` that passes more than 1 argument to the mapper function. Some libraries will name `reduce` as `foldr`. Some libraries might invoke the reduce function with the accumulator last. Did you notice that those are the two most common list operations and we are already seeing _severe_ incompatibility issues? Even if there is just a single mismatch between Ramda and a certain third party library I would personally completely avoid dispatching on it at all because doing so would be entering a minefield of inconsistent code, tricky knowledge and easy to make mistakes. Knowledge like "I can dispatch on all the list functions expect `x`, `y` and `z`.

We have already seen issues with the whole Highland discussion, the renaming of `eq` to `equals` and with the problematic aspects of wanting to derive some function in terms of other dispatches.

Third party libraries will have a much harder time implementing something that is compatible with Ramda if we, just like Fantasy Land, demand certain things from sensible names like `map`, `reduce`, `filter`, `contains`, etc.

A specification helps users as well. If they know that FooLib implements the Ramda list specification they can look in the documentation and they will know that they can use all functions covered by the list specification with FooLib.

CrossEye:
I don't object to there being a specification, or to us working with it.  But I can see no harm in the simpler form of dispatching to commonly-named functions.  I really do want to find a way to deal with aliases for them on dispatch, and I think that metadata might  be it, but the basic point is that if a user's type has a `map` method, and that `map` is compatible with Ramda's `map` API, then she can _choose_ to call `map(fn, myTypeInstance)`.  The point is that Ramda will offer her a different API that she can use, one that is more convenient for functional composition.  It's not as though we're holding a basket of FP candies out and saying, "You can only have them if you're good and implement your functions our way."

But `map` is such a common idiom; there are functors everywhere.  If I were to implement a `Tree` type, I would likely implement a `map` for it, and that `map` would by its very nature be a functor map.  The fact that it could then would be able to work with Ramda is very convenient.  `reduce` too is extremely common, with a similar API.  There is still a real problem of naming/aliasing that I want to solve, since there's a real chance that someone would call it `fold` instead.  But there is a strong tradition here too in the design of a `fold` API, so there is a reasonable chance that a `reduce` someone made on her `QueryResults` data structure would be compatible as well.  If not, there is no penalty, no issues; it's simply not something that can be used so easily with Ramda.

I'm not quite as excited about dispatch as [David seems to be][dc], but they are a key part of how I want Ramda to work.  Lists are only a part of what we want to work with.  They're the only part we implement directly, but other algebraic types, and other somewhat-related types can be really important too.


  [dc]: https://github.com/ramda/ramda/pull/1160#issuecomment-109888577

buzzdecafe:
i think i am leaning toward @paldepind on this. dispatch was added to support fantasy-land style objects in a point-free fashion. This is fine, because those objects are well-specified (for the most part). Then it fell out that this would work with arbitrary objects. In some cases that sort-of-mostly-works. At the time, we considered this ok, because nobody was using the library but us anyways. 

> Ramda can transform your OO API into a more FP style.

To my mind, this was **never** the point. We have `invoker` for that. The point of dispatch was to be able to treat a Functor as a Functor, an Apply as an Apply, etc.

paldepind:
> To my mind, this was never the point. We have invoker for that. The point of dispatch was to be able to treat a Functor as a Functor, an Apply as an Apply, etc. 

I agree very much with this point. It seems to me that the idea is that all Ramda functions should be overloaded with invoker behavior. I think this very much goes against the idea that Ramda functions are _simple_.

> But I can see no harm in the simpler form of dispatching to commonly-named functions.

I consider the current dispatch mechanism harmful! If people use it it will lead to error prone inconsistent code because the libraries they use it with unavoidably will only be partly compatible! They will either have to fill in the holes manually with `invoker` or they will have to mix functional and object oriented styles. It could furthermore lead to code like this:

```javascript
R.map(function(element) { /* something */, plainArray);
// Somewhere else in the code base
R.map(function(index, element) { /* something */, jqueryCollection);
// Somewhere else again
R.map(function(element, index, list) { /* something */, immutableJsList);
```

The above is very disturbing to me. We are calling the same function but the mapper function has very different signatures. Both of those problems are _huge_ problems to me. We are promoting practices that will lead to horrible code for our users.

But there is another fundamental problem as well: _the dispatch behavior completely misses the fundamental point of fantasy land_. Fantasy land was not made so libraries could do dispatching. Dispatching in itself is worth very little. Fantasy land was made so that libraries could write code that would work for all instances implementing an interface of a certain type class. It was made so that one for instance could write a general `lift` that would work for _all_ applicatives.

The reason why one takes the effort to make a type an instance of type classes is to get functions for free! The promise is: implement `map` and `ap` and we will give you `lift` and more for free. That is an actual benefit that goes way beyond what we can currently offer: "if you implement `map` and `ap` methods we will give you `ap` and `map` functions (the library could easily just expose those functions directly (like Flyd does) or the functions could be created with `invoker`)". The latter is very moot compared to the huge benefits of the former.

Ramda completely missed the opportunity and instead only added the most basic fantasy land support imaginable: dispatching on the methods that are directly defined in fantasy land. To me the fact that fantasy land is implemented with methods is an _implementation detail_. Users should not care! All they should care about is "library x exposes an applicative, Ramda's `lift` supports all applicatives". Users should be thinking in terms of type classes (of course we don't have actual type classes but the specifications gives something similar) they should _not_ be thinking in terms of dispatching and methods!

> If not, there is no penalty, no issues; it's simply not something that can be used so easily with Ramda.

Yes. And if you take that thought a little further you would realize that at some point some Ramda user will open an issue on an unrelated library asking the maintainer of it to make a change so that the user can use the other library better with Ramda. With our current dispatching such a request would sound something like this "please rename your `foldr` method to `reduce`" or "please change the signature of your `map` method". I.e. our current dispatch mechanism would force the owner of the library to _change his API_ in order to be Ramda friendly. It is bad behavior to be the cause of such demands ending up at other party libraries! Fantasy land has met resistance because of this exact issue as well. That is what I mean when I say that we claim ownership of those names. Now _if_ instead we dispatched to a long namespaced method name other libraries could implement our specification and be Ramda compatible without interference with their existing API.

Dispatching to namespaced methods would also prevent all the paths to disaster that I mentioned in the beginning of this post.

buzzdecafe:
>  All they should care about is "library x exposes an applicative, Ramda's lift supports all applicatives". 
> Users should be thinking in terms of type classes (of course we don't have actual type classes but the specifications gives something similar) they should not be thinking in terms of dispatching and methods!

100% agree!

CrossEye:
> i think i am leaning toward @paldepind on this. dispatch was added to support fantasy-land style objects in a point-free fashion.

That's funny, I remember it differently.  I remember starting dispatch with the lazy-list extension and discussing how this might extend to things like streams.  The FantasyLand integration was always to me just one possibly use among many.

I would really like it if we could dispatch to Bacon or Highland.  Bacon is an especially interesting use-case, as @paldepind's offereing in this space, Flyd, already has a function-first API, but offers as well a FantasyLand-compatible method-based API.  It would be nice if Ramda could serve as a function-first API for Bacon.  Maybe it would only end up matching on `map` and `filter`, but even there it might still be useful.

But the general idea of a really simple integration does seem very useful to me.  This is an example from [Immutable.js][im]

```js
Immutable.Range(1, Infinity)
  .skip(1000)
  .map(n => -n)
  .filter(n => n % 2 === 0)
  .take(2)
  .reduce((r, n) => r * n, 1);
// 1006008
```

If we could figure out the `drop`-`skip` aliasing issue, this would simply work in Ramda today as:

```js
var fn = R.pipe(
  R.drop(1000),
  R.map(n => -n),
  R.filter(n => n % 2 === 0),
  R.take(2),
  R.reduce((r, n) => r * n, 1)
);
fn(Immutable.Range(1, Infinity)); //=> 1006008
```

And I think that would be a very good thing.  People who wanted both compositional goodness and true immutability would have a useful mashup of two libraries with little work.  But of those five functions `map` is specified by Functor, `reduce` by (the recently-added) Foldable, and the others are not FantasyLand specifications at all.  Restricting ourselves to just those would be a shame.


  [im]: https://facebook.github.io/immutable-js/

buzzdecafe:
> But the general idea of a really simple integration does seem very useful to me.

as @paldepind points out above, the integration is anything but simple. At present it is _ad hoc_.

paldepind:
> But the general idea of a really simple integration does seem very useful to me.

To me it is a minefield full of danger and horrible code. See the issues I pointed out before.

> This is an example from Immutable.js

That is a misguided example. What you're trying to do is exactly the problem that transducers solves. And using transducers would be a much better solution. The code you've written there is _fragile_! Hard to get right and easy to break. Throw something like `dropRepeats` into the mix and it _all falls apart_. Transducers on the other hand would not.

> If we could figure out the drop-skip aliasing issue

Then you have something even more complex and opaque that is still very brittle.

> People who wanted both compositional goodness and true immutability would have a useful mashup of two libraries with little work.

People who want that are using proper solutions already (transducers).

> Restricting ourselves to just those would be a shame.

We certainly shouldn't. But we should be specifying proper interfaces and prefixing the methods with namespaces.

CrossEye:
>> To my mind, this was **never** the point. We have `invoker` for that. The point of dispatch was to be able to treat a Functor as a Functor, an Apply as an Apply, etc.

> I agree very much with this point. 

And I disagree wholeheartedly.  FantasyLand is just one integration point among many, and to my mind has no privileged position among them.  The point instead was the treat things that need to `map` uniformly and those that need to `filter` uniformly.  We implement these for lists, but dispatch for other types.  For instance, I see no reason to skip `filter` in our dispatch simply because it's not specified by FantasyLand.  I can easily see it being used by other types with which we'd like to integrate: Bacon, Immutable, some Tree implementation, and others.

> It seems to me that the idea is that all Ramda functions should be overloaded with invoker behavior. I think this very much goes against the idea that Ramda functions are simple.

I don't see this.  There are, of course, some issues, most notably aliasing.  I don't think these are insurmountable.

> ```javascript
> R.map(function(element) { /* something */, plainArray);
> // Somewhere else in the code base
> R.map(function(index, element) { /* something */, jqueryCollection);
> // Somewhere else again
> R.map(function(element, index, list) { /* something */, immutableJsList);
> ```


> The above is very disturbing to me. We are calling the same function but the mapper function has very different signatures. Both of those problems are _huge_ problems to me. We are promoting practices that will lead to horrible code for our users.

I see these same blocks in a very different way.  Ramda allows the user to transform imperative code blocks like these:

```js
var results = plainArray.map(function(element) { /* something */ });
// ...
var results = $jqueryCollection.map(function(index, element) { /* something */});
// ...
var reults = immutableJsList.map(function(element, index, list) { /* something */ });
```

into pure functions like these

```js
var arrayFn = R.map(function(element) { /* something */ });;
// ...
var jqFn = R.map(function(index, element) { /* something */ });
// ...
var immutFn = R.map(function(element, index, list) { /* something */ });
```

for later use in building an applicaiton through functional composition.  I have no worries whatsoever that the same function isn't passed to each of them, or even that the signatures are different.  All our dispatching does here is to add this compositional layer atop the user's functions.


> Ramda completely missed the opportunity and instead only added the most basic fantasy land support imaginable: dispatching on the methods that are directly defined in fantasy land. 

Ramda, has not ever even tried to take on the challenge of working with Algebraic Data Types.  Of course there's a side project for that, but Ramda has not itself tried to ever understand and work with the roles of Monoids, Foldables, Applicatives, _et. al._.  I think there's a good reason for that.  Ramda has remained a fairly low-level library, and it's tried to stay independent.  Although through its dispatch, it does interact with FantasyLand and other libraries, it does not depend on them at all.  If we were to start to depend on certain FL laws in order for our own functions to operate, I think we'd lose a lot of what make Ramda the library it is.


> Now _if_ instead we dispatched to a long namespaced method name other libraries could implement our specification and be Ramda compatible without interference with their existing API.

I am really hoping this goes almost entirely in the other direction.  Ramda is trying to [find a way][is] to integrate with other libraries, and to do so unobtrusively.  If "integrate with Ramda" ever becomes an important goal of another library, then to my mind we will have failed miserably with this dispatch.  I know there is work to do to get there, but I'm thinking that the only significant issue is that of name aliases.  I'm really hoping that as we pursue metadata, we can find a solution to this problem using metadata.  If not, I have a few other, less appealing ideas.


  [is]: https://github.com/ramda/ramda/issues/1037

CrossEye:
>> It would be nice if Ramda could serve as a function-first API for Bacon.

> Users can easily create such a thing with invoker. And let me point out that Bacon specifically does not implement a proper fantasy land map...

Sure, but to me that doesn't matter at all.  Dispatching is **not about** FantasyLand.  That's just one example of the integrations that are available.

>> This is an example from Immutable.js

> That is a misguided example. What you're trying to do is exactly the problem that transducers solves. And using transducers would be a much better solution. 

This was the first example from the Immutable page that had more than three steps.  I realized immediately that it was a close fit for Ramda.  Why do you say transducers would be better?  Obviously we could not do the equivalent with arrays in Ramda, if you suggest an infinite collection, but that's not what I was suggesting, only that we could create a function that accept a collection (presumably of numbers), drop the first thousand, negate the remaining ones, remove the odd ones, take the next two, and multiply them together in Ramda in an almost identical manner to what was done in Immutable.  If we then apply this function to `Immutable.Range(1, Infinity)`, the dispatching works fine (modulo some missing aliasing for `skip`.)


> The code you've written there is _fragile_! Hard to get right and easy to break. Throw something like `dropRepeats` into the mix and it _all falls apart_.

Because then you would have to add a lamda instead of a Ramda function??


> But we should be specifying proper interfaces and prefixing the methods with namespaces.

So in #1193, @jdalton was noting the different callbacks used for `forEach` for `Sets`, `Maps`, and lists.  Do you want to give a stab at a proper interface definition for this one?  Would we have to somehow wrap `Sets`, `Maps`, etc. to use it, or would it be compatible?  Would it also cover jQuery's `each`, which serves the same role, but with a different API?  Would you expect that users would need to write interface wrapper layers in order to work with Ramda?

paldepind:
> into pure functions like these

Into a `map` that is a monster and something completely different from a semantically correct `map` on a semantically correct functor.

> Ramda, has not ever even tried to take on the challenge of working with Algebraic Data Types. Of course there's a side project for that, but Ramda has not itself tried to ever understand and work with the roles of Monoids, Foldables, Applicatives, et. al.. I think there's a good reason for that. Ramda has remained a fairly low-level library, and it's tried to stay independent. Although through its dispatch, it does interact with FantasyLand and other libraries, it does not depend on them at all. If we were to start to depend on certain FL laws in order for our own functions to operate, I think we'd lose a lot of what make Ramda the library it is.

I am not suggesting that we depend on fantasy land in any way. Also I took a look at the implementation of `lift` and it actually does work with all fantasy land applicatives. So what I said before and what you're saying now is not correct.

> I am really hoping this goes almost entirely in the other direction. Ramda is trying to find a way to integrate with other libraries, and to do so unobtrusively. If "integrate with Ramda" ever becomes an important goal of another library, then to my mind we will have failed miserably with this dispatch. I know there is work to do to get there, but I'm thinking that the only significant issue is that of name aliases. I'm really hoping that as we pursue metadata, we can find a solution to this problem using metadata. If not, I have a few other, less appealing ideas.

You can't get a meaningful integration with another library just by overloading every function with a bunch of dispatches.

> Why do you say transducers would be better?

I gave reasons for the specific example. But my [other arguments applies as well](https://github.com/ramda/ramda/pull/1173#issuecomment-112419278) in this case.

> Because then you would have to add a lamda instead of a Ramda function??

No. Because `dropRepeats` can's delegate to immuteable-js.

```javascript
var fn = R.pipe(
  R.drop(1000),
  R.map(n => -n),
  R.dropRepeats, // <-- Bam! It all falls apart. Fragile code sucks! You now have to rewrite with transducers
  R.filter(n => n % 2 === 0),
  R.take(2),
  R.reduce((r, n) => r * n, 1),
);
fn(Immutable.Range(1, Infinity)); //=> 1006008
```

>  Do you want to give a stab at a proper interface definition for this one?

Yes. It would be very easy. `forEach` is part of the SomethingSomething specification. All SomethingSomething's must implement a `@@SomethingSomething/forEach` method that must implement the obvious semantics.

paldepind:
We should create a new library "Damda" – for dispatch-Ramda. It would provide invokers for the 300 most common method names in JavaScript libraries. Then Damda would take care of "integrating" with third party libraries and Ramda would be free to only concern itself with well defined interfaces and specifications like fantasy land, transducers and other future specifications.

Damda of course could easily have both `foldr` and `reduce` and `drop` and `skip`. Then aliases would be uneeded and things would be slightly harder to get wrong and simpler. Furthemore it would be obvious when one is talking about an _actual_ `map` (that would be `R.map`) and when one is just talking about a method named map that can and will do all sorts of [crazy](http://api.jquery.com/map/) things (that would be `D.map`).

buzzdecafe:
not to disrupt the very amusing fight going on, but let's merge this?

buzzdecafe:
```js
var arrayFn = R.map(function(element) { /* something */ });;
// ...
var jqFn = R.map(function(index, element) { /* something */ });
// ...
var immutFn = R.map(function(element, index, list) { /* something */ });
```

this strikes me as a *huge* problem, since 2/3 of these break the signature: `Functor f => (a -> b) -> f a -> f b`. I'd *much* rather support the behavior specified in the signature than some sort of WildWest.js.

paldepind:
> this strikes me as a huge problem, since 2/3 of these break the signature: Functor f => (a -> b) -> f a -> f b. I'd much rather support the behavior specified in the signature than some sort of WildWest.js. 

Exactly. And in the Bacon example one doesn't even have to pass a function `R.map('.someProperty', baconStream)`.

But yes, :+1: for merging this.

CrossEye:
> I am not suggesting that we depend on fantasy land in any way. Also I took a look at the implementation of `lift` and it actually does work with all fantasy land applicatives. 

That's nice.  Do you want to insist that all functions are built only in a FantasyLand-compatible manner?  Or should we treat it as serendipity when it happens?

> You can't get a meaningful integration with another library just by overloading every function with a bunch of dispatches.

How much meaning we're trying to attach seems to be the main point of contention.  To my mind, it's very little; it's a syntactic convenience.  You seem to want significantly more.  It sounds as though you would like us to define general-purpose interfaces and possibly even guard against incorrect usage of them.  I'm not even _close_ to ready to go that route.  There's enough controversy over how many parameters one should pass to `map`, and whether the same function should work for lists and plain objects.  To try to take on a general-purpose specification, and try to get it right -- and worse to do this as part of what's essentially a utility library -- sounds just plain crazy.


> ```js
>   R.dropRepeats, // <-- Bam! It all falls apart. Fragile code sucks! You now have to rewrite with );
> ```

And look, it also fails if you insert `R.lt(42)` or `R.propEq('foo', 'bar')`.  Do those bother you just as much?  Does that seem equally fragile to you?

Somehow, a technique that is designed to trasform one style of code into another doesn't seem made bad when I can't further transform it into something else I might like.  I'm not trying to do arbitrary Ramda code on `Immutable` data structures.  I'm trying to offer whatever portions I can of a Ramda-style API to `Immutable` users.


>> Do you want to give a stab at a proper interface definition for this one?

> Yes. It would be very easy. `forEach` is part of the SomethingSomething specification. All SomethingSomething's must implement a `@@SomethingSomething/forEach` method that must implement the obvious semantics.

I'm sorry, but I had a little time parsing the details.  How many parameters again for the callback?  In what order?  Are any of them optional?  Are their types fixed, or can they vary?  What's the return type?  Can this just be delegated to `@@iterator` if that is present?  Is there any guarantee about iteration order?  What happens if an item is removed from the collection while the iteration is in progress?  What if one is added?  Does the callback receive references to the items in the original collection or copies of them?  Are infinite collections allowed?  What about cyclic ones?  Is the `forEach` from `Set` compliant?  How about the one from `Map`?  what about jQuery's `each`?...

CrossEye:
> ```js
> var arrayFn = R.map(function(element) { /* something */ });;
> // ...
> var jqFn = R.map(function(index, element) { /* something */ });
> // ...
> var immutFn = R.map(function(element, index, list) { /* something */ });
> ```

> this strikes me as a *huge* problem, since 2/3 of these break the signature: `Functor f => (a -> b) -> f a -> f b`. I'd *much* rather support the behavior specified in the signature than some sort of WildWest.js. 

Well, this is our current signature: 

```haskell
(a -> b) -> [a] -> [b]
```

So anything but lists are already out.

I'm simply not convinced that we want to be in the business at all of enforcing signatures for dispatching.  Do we do type checking of the parameters and the arities?  Do we wrap and check return types too?  It feels entirely the wrong direction to me.  

I'm not actually opposed to updating the documentation to point towards the _expected_ signature.  I just don't want to be in the business of trying to enforce it in any manner whatsoever.

buzzdecafe:
> I'm simply not convinced that we want to be in the business at all of enforcing signatures for dispatching. 

who said anything about enforcing? I'm concerned with what we want to *support*. If you break the contract, that is your problem.

paldepind:
> Do you want to insist that all functions are built only in a FantasyLand-compatible manner?

There must be a misunderstanding. I never suggested that _any_ function should only be built in a fantasy land compatible manner.

> To my mind, it's very little; it's a syntactic convenience.

Yes. I understand that. You want Ramda functions overloaded with invoker behavior. And I've argued several times now why I think that is a harmful thing. If one wants a huge bunch of predefined invoker functions I've suggested Damda (or some other name). There is no reason to include such complexity and overloaded semantics in Ramda. We are aiming for _simple_ functions.

> You seem to want significantly more. It sounds as though you would like us to define general-purpose interfaces and possibly even guard against incorrect usage of them. I'm not even close to ready to go that route. There's enough controversy over how many parameters one should pass to map, and whether the same function should work for lists and plain objects. To try to take on a general-purpose specification, and try to get it right -- and worse to do this as part of what's essentially a utility library -- sounds just plain crazy.

That is not at all what I want. Initially I want significantly _less_. I think we should dispatch only on fantasy land types. And I think we should work more with the fantasy land specification and consider implementing other of our functions so that they work with the fantasy land types. That is _less_ complexity. To me the ad-hoc dispatching to unknown methods in otherwise simple functions is the real craziness. And I only mentioned creating specifications because that is the sensible alternative to all this ad-hoc library "integration".

> And look, it also fails if you insert R.lt(42) or R.propEq('foo', 'bar'). Do those bother you just as much? Does that seem equally fragile to you?

No. Of course not. Because `lt` and `propEq` are not functions on lists. But inserting `dropRepeats` in that chain would be completely reasonable. But immuteable-js has no such method to dispatch to. You can never get a 100% correspondence and thus users have to constantly remember what 40% of list operations that are unmatched between Ramda and SomeLib. That leads to fragile code that is easy to break (just insert `dropRepeat`, etc).

> I'm simply not convinced that we want to be in the business at all of enforcing signatures for dispatching. Do we do type checking of the parameters and the arities? Do we wrap and check return types too? It feels entirely the wrong direction to me. 

As far as I've seen no one has suggested any of that.

paldepind:
> Or should we treat it as serendipity when it happens?

I dont understand that remark. Lift intentionally supports fantasy land applicatives. It's written in the documentation.

kedashoe:
> I think we should dispatch only on fantasy land types.

I think either 1. all dispatching in Ramspatch or 2. FL only makes sense. But, as you mentioned above, the latter is kinda impossible unless FL changes the spec by namespacing, in which case Sanctuary (and everyone else) would have to namespace (or alias) all its types' methods in order to work with Ramda. Thoughts on that @davidchambers ? Feels kinda gross, and yet at the same time feels cleaner than what Ramda does now. I guess that is javascript.

davidchambers:
If we decide to separate the dispatching layer from Ramda, we should name the new project Shepherd. :)

> Sanctuary (and everyone else) would have to namespace (or alias) all its types' methods in order to work with Ramda.

I like to use `R.map`, `R.filter`, etc. rather than use `Just#map`, `Maybe#filter`, etc. directly. As a user of Sanctuary, I shouldn't need to know the names of the underlying methods.

CrossEye:
>> To my mind, it's very little; it's a syntactic convenience.

> Yes. I understand that. You want Ramda functions overloaded with invoker behavior. 

I guess that's reasonably close.  I want almost no change from the current dispatch, except to figure out how to handle aliasing.  But I would describe it somewhat differently.  Ramda list functions, when passed a non-list object in the list position would dispatch to the appropriately named method of that object, if that object does in fact have such a method.  It would pass the remaining arguments to that method.  

> And I've argued several times now why I think that is a harmful thing. If one wants a huge bunch of predefined invoker functions I've suggested Damda (or some other name). There is no reason to include such complexity and overloaded semantics in Ramda. We are aiming for simple functions.

Here's where I simply don't understand the other side of the argument.  I feel like what I just described is extremely simple.  Combining this with some additional rules, whether it's to limit the API of any callback functions, to limit the types allowed in that list position, or to otherwise restrict what the dispatched-to methods are allowed to do, all strike me as adding significant additional complexity.

> I think we should dispatch only on fantasy land types. And I think we should work more with the fantasy land specification and consider implementing other of our functions so that they work with the fantasy land types.

First of all, why are the FantasyLand types so special?  If someone wants to create the `MonadPlus` or `Ord` specification, even though it's not listed in FantasyLand, why wouldn't we also be supporting those?

But even more fundamentally, how would you propose to do this?  How would you recognize an object as complying with a particular FantasyLand specification?  Basically the specification insists on the existence of some named methods, and asserts some laws about how they interact.  But Ramda obviously has no clear way to check the laws.  And as to the name, well, that's no different than my suggestion, as far as I can tell.  Can you suggest a way to choose to dispatch `map` to someone's `Maybe.Just` implementation but not the Bacon, Immutable, or jQuery ones you objected to above?

Is there a straightforward implementation that does this, @paldepind, @buzzdecafe?

paldepind:
> But, as you mentioned above, the latter is kinda impossible unless FL changes the spec by namespacing, in which case Sanctuary (and everyone else) would have to namespace (or alias) all its types' methods in order to work with Ramda. 

If fantasy land changes then all libraries that supports it Ramda, Sanctuary, etc. would have to change.

>  Ramda list functions, when passed a non-list object in the list position would dispatch to the appropriately named method of that object, if that object does in fact have such a method. It would pass the remaining arguments to that method. 

Yes. And you're adding that complexity to all list functions. Add aliasing as well and the complexity gets more unwieldy.

>  I feel like what I just described is extremely simple.

I am suggesting that we remove the dispatching. A function with no dispatching is certainly simpler than a function with dispatching? But it is not the complexity that bothers me the most (though that I don't fancy that as well). It is how it promotes fragile code! It is a serious problem. If you had your way with aliasing and you went out and wrote code using it like the immuteable-js example you suggested above. Then your new colleague could very reasonable assume that he could put in a `dropRepeats` in that compose chain. Then the house of cards fall together and her gets an error and he doesn't understand. What do you say to him when he asks why? You give him the previous description of how all Ramda functions are overloaded with invoker behavior and that some of them even has aliases. And you tell him that whenever he wants to program in that style he will have to remember to make sure that the Ramda function he wants to use as an invoker should have a mapping, potentially aliased, to a corresponding immuteable-js method. So he should always take great care in cross checking Ramda's documentation with immuteable-js's documentation otherwise things wont work. And, if there is no such correspondence he will have to fall back to using immuteable-js in an OO way. Do you think your colleague would think of that as a sane way to code? Do you think that is simple?

>  If someone wants to create the MonadPlus or Ord specification, even though it's not listed in FantasyLand, why wouldn't we also be supporting those?

Then we should support it! That is my whole point. We should support well specified interfaces but remove all the current ad-hoc complexity

> Combining this with some additional rules, whether it's to limit the API of any callback functions, to limit the types allowed in that list position, or to otherwise restrict what the dispatched-to methods are allowed to do, all strike me as adding significant additional complexity.

Where are you getting this from? I have not suggested any such things.

@davidchambers 
> If we decide to separate the dispatching layer from Ramda, we should name the new project Shepherd. :)

That is a much better name. Quick and dirty implementation:

```javascript
var keyArity = R.merge(R.mapObj(R.compose(R.dec, R.prop('length')), R), {foldr: 2, skip: 1});
var arity, invokers = {};
for (var key in keyArity) {
  arity = keyArity[key];
  if (typeof arity === 'number' && arity > 0) {
    invokers[key] = R.invoker(arity, key);
  }
}
module.exports = invokers;
```

Using this instead of dispatching Ramda functions has the following advantages:

* Ramda will be simpler
* Ramda functions will be reserved for their intended sensible use. I.e. when I see `R.map` in a code base I immediately know it is a proper map on a proper functor.
* When one wants to dispatch the dispatching is explicit instead of added inside another function.
* New dispatchers can easily be added. This avoids the complexity of aliases.

> But even more fundamentally, how would you propose to do this?

Remove all dispatchers except those for fantasy land, transducers and those we use to support strings in other functions (@davidchambers recent changes). Remove all mentions of dispatching in the documentation and stop suggesting to users that they do it (suggest Shepherd instead). Document when functions works on data types from fantasy land and/or transducers.

> Can you suggest a way to choose to dispatch map to someone's Maybe.Just implementation but not the Bacon, Immutable, or jQuery ones you objected to above?

The best solution IMO would be for fantasy land to add namespaces but that seems unlikely to happen. If we agree that dispatching is an implementation detail to support fantasy land then users wouldn't know that the dispatching happens and such the harm is vastly diminished.

CrossEye:
> If you had your way with aliasing and you went out and wrote code using it like the immuteable-js example you suggested above. Then your new colleague could very reasonable assume that he could put in a `dropRepeats` in that compose chain. Then the house of cards fall together and her gets an error and he doesn't understand.

Well, one would presumably decorate this with

    // :: Immutable.Range -> Number

I just can't tell what our issues are here, whether we're talking past each other a bit, or if the disagreement is really fundamental.  Obviously there is something fundamental in that I want to keep the dispatching, only slightly altered, and you want to remove it except for some very specific cases.

But I think we might be talking past each other here in how we would expect it to be used.  My example for a block of code that is already `:: Immutable.Range -> Number`, and is presumably annotated as such, but is more difficult to extract as a function because of the OO API of Immutable.  Our dispatch offers a way to turn that into a simple functional version.  It does not turn the user's code-base into a Ramda-centric system.  But if they are already using Immutable and already using Ramda, or want to use Ramda for it's more functional API, this is a nice gloss.  If my colleague acted as though it's trying to turn Immutable somehow into regular lists, I might give him a funny look or two before setting him straight, but that's about it.

> I am suggesting that we remove the dispatching. 

You say that, but then say:

> Remove all dispatchers except those for fantasy land, transducers and those we use to support strings in other functions 

which I see as two hugely different ideas.  I absolutely do not understand the privileged position you're giving to the FantasyLand specification.  Why would we dispatch `chain` and not `filter`, based on someone else deciding that one was important enough for their list of types and the other was not?  The Haskell folks have done some pretty brilliant things, and I get a lot out of rereading [Typeclassopedia][tc], but a translation of the ideas from there and from Scala and other similar places into Javascript is certainly not perfect; I certainly don't see it as one of the most important ways to drive Ramda's API.  And I do ask seriously about `chain` and `filter`.  I understand what I might get from `chain` in making my container type a Monad, but in my day-to-day programming, I'm still at least as likely to want to implement `filter` on it.  What makes the FantasyLand specification so important that it drives this API decision for Ramda and its users?


> [Using 'Shepherd'] Ramda functions will be reserved for their intended sensible use. I.e. when I see R.map in a code base I immediately know it is a proper map on a proper functor.

Proper according to what?

I keep harping on the FantasyLand spec here, and with some reason.  Although I think it's done a fair bit of good for the community, I also think as a project it's been a disaster.  The actual specification seems poorly written and contains some serious holes.  There is little appetite to maintain it or to consider reasonable proposals.  So, while I'd like to be able to work well with types that adhere to the FantasyLand specification, I don't want doing so to have too strong an influence Ramda's decisions.


>>Can you suggest a way to choose to dispatch `map` to someone's `Maybe.Just` implementation but not the Bacon, Immutable, or jQuery ones you objected to above?

> The best solution IMO would be for fantasy land to add namespaces but that seems unlikely to happen. If we agree that dispatching is an implementation detail to support fantasy land then users wouldn't know that the dispatching happens and such the harm is vastly diminished.

I don't understand this at all.  How do you propose to dispatch to types that conform to the FantasyLand specification but not to others that happen to use the same name?  How do you solve the multiple-`map` problem you describe above?



  [tc]: https://wiki.haskell.org/Typeclassopedia

scott-christopher:
I mentioned this in the Gitter room, but figured it was better to include here.

Upon reading the comments in this issue, I decided to have a play around with moving the conditional dispatching logic outside of the individual functions and into external objects, with the idea being that the existing dispatching could still be kept within Ramda’s reach. There is then a choice to either apply the existing dispatching to the distributed `R` object by default, or leave it up to the user to apply as needed.

This also allows others to provide their own dispatch conditions for things such as jQuery objects, etc, or in other creative ways such as debug wrapping.

https://gist.github.com/scott-christopher/a80af9be36f47901fb50

Thoughts?

joneshf:
> does this imply we should change the sig of map to Functor f => (a -> b) -> f a -> f b? that strikes me as more accurate (because of dispatch) and more obscure to new users who may not be familiar with Haskell.

> sure, but the *intent* is to use `map` in a generic way with any functor. I'm not concerned about edge cases in sigs. One could argue that the typeclassed sig is more descriptive of the intent.

If the unfamiliarity is a concern, why not say explicitly what is intended? The point is that the second argument satisfies some interface. Invent notation to represent that:

```javascript
R.map : f ~ {map, ...} => (a -> b) -> f a -> f b
```

Or look into something like what infernu is doing with [`this`](https://github.com/sinelaw/infernu#functions-and-this). It may not be the same, but the idea is what counts.

People have to learn the type notation anyway, might as well make it pertinent to this library. You don't have to try and shove type classes into js. Call things ducks if you want, or call them interfaces, or whatever. It doesn't really matter what, but making it explicit (even with special notation) is much better than not.

> > I've been seriously considering what it would take to write a pure-function as opposed to object-based version of the FantasyLand specification.

> I don't think that is possible.

Why not? That's where fantasy land came from. Why would it be impossible to go back that way? It would probably end up pretty terrible though.

> But there is another fundamental problem as well: the dispatch behavior completely misses the fundamental point of fantasy land. Fantasy land was not made so libraries could do dispatching. Dispatching in itself is worth very little. Fantasy land was made so that libraries could write code that would work for all instances implementing an interface of a certain type class. It was made so that one for instance could write a general lift that would work for all applicatives.
> 
> The reason why one takes the effort to make a type an instance of type classes is to get functions for free! The promise is: implement map and ap and we will give you lift and more for free. That is an actual benefit that goes way beyond what we can currently offer: "if you implement map and ap methods we will give you ap and map functions (the library could easily just expose those functions directly (like Flyd does) or the functions could be created with invoker)". The latter is very moot compared to the huge benefits of the former.
> 
> Ramda completely missed the opportunity and instead only added the most basic fantasy land support imaginable: dispatching on the methods that are directly defined in fantasy land. To me the fact that fantasy land is implemented with methods is an implementation detail. Users should not care! All they should care about is "library x exposes an applicative, Ramda's lift supports all applicatives". Users should be thinking in terms of type classes (of course we don't have actual type classes but the specifications gives something similar) they should not be thinking in terms of dispatching and methods!

Yes! That's exactly it! But I don't think the two are disparate ideas. I think it's actually intrinsic to js that fantasyland works as prototype methods and whatever. The alternative of passing around dictionaries goes far against the grain. Honestly it seems to me that ramda is following the spirit of fantasy-land. And I thought that ramda-fantasy was for all the derivations?

> > This is an example from Immutable.js

> That is a misguided example. What you're trying to do is exactly the problem that transducers solves. And using transducers would be a much better solution. The code you've written there is fragile! Hard to get right and easy to break. Throw something like dropRepeats into the mix and it all falls apart. Transducers on the other hand would not.
> 
> > If we could figure out the drop-skip aliasing issue

> Then you have something even more complex and opaque that is still very brittle.

Another option not involving transducers is `Traversable`. Though you might also need a `Monoid`. With these, you should be able to implement all those including `dropRepeats`, since from a `Traversable` you get `Foldable` and `Functor` for free. 

Or suggest another typeclass/interface/duck to fantasy-land.

Though, honestly, it's a type problem much more than it's a ramda/fantasy-land/whatever problem. There's nothing stating what the type of that function is. If the type of it is `[Int] -> Int`, then yes, you should expect that you're able to insert any old list function is the appropriate place, and go to town. If however the function is `Traversable f, Alternative f, Num a => f a -> a`, then you had better not expect to drop any random list or array or whatever function at any arbitrary point in the pipe and have things not mess up. There's nothing even talking about lists or arrays or whatever.

You're right that the best way is to design to an interface rather than a concrete type, but you can't expect to design to an interface and then arbitrarily stick a concrete function in the middle of things and have everything work out properly.

> I just can't tell what our issues are here, whether we're talking past each other a bit, or if the disagreement is really fundamental. Obviously there is something fundamental in that I want to keep the dispatching, only slightly altered, and you want to remove it except for some very specific cases.

If definitely seems like you're both wanting very similar things, but the words are getting jumbled.

paldepind:
> But I think we might be talking past each other here in how we would expect it to be used. My example for a block of code that is already :: Immutable.Range -> Number, and is presumably annotated as such, but is more difficult to extract as a function because of the OO API of Immutable. Our dispatch offers a way to turn that into a simple functional version. It does not turn the user's code-base into a Ramda-centric system. But if they are already using Immutable and already using Ramda, or want to use Ramda for it's more functional API, this is a nice gloss.

I understand what you want to achieve. And it makes sense but I think your approach it wrong. I've tried to explain why I think overloading Ramda functions with dispatch behavior is a very harmful way to do it. But I do not know how I can explain it better than what I've already tried.

> If my colleague acted as though it's trying to turn Immutable somehow into regular lists, I might give him a funny look or two before setting him straight, but that's about it.

And he would give you a _very_ sad look when you told him that inserting a reasonable function like `dropRepeats` makes it all fall together and that he has to cross check documentations to avoid breakage. What about code using Kefir like this:

```javascript
someStream.map(someFn).onEnd(someEndHandler);
```

This is _yet another_ example where the dispatch mechanism falls short. You can turn `map` into a function but not `onEnd`. So the current dispatching can only give you nothing or ugly code that is a mix of two styles:

```javascript
R.map(someFn, someStream).onEnd(someEndHandler);
```

These are all signs of a poor leaky abstraction.

> Proper according to what?

You know what a proper map is on a proper functor. In the wild west of JavaScript these are not the norm.

> What makes the FantasyLand specification so important that it drives this API decision for Ramda and its users?

I don't think fantasy land is should drive API decisions for Ramda. Actually I was very much against that we renamed our `eq` function just to match `equal` in fantasy land. But this is not about giving fantasy land special privilege. This something else, this is about ad-hoc dispatching being harmful.

> I keep harping on the FantasyLand spec here, and with some reason. Although I think it's done a fair bit of good for the community, I also think as a project it's been a disaster. The actual specification seems poorly written and contains some serious holes. There is little appetite to maintain it or to consider reasonable proposals. So, while I'd like to be able to work well with types that adhere to the FantasyLand specification, I don't want doing so to have too strong an influence Ramda's decisions.

We agree that fantasy land has issues. But it is a specification and supporting it is nice.

@scott-christopher Did you see the Shepherd implementation I posted before? It does not seem to me like your proposal gives the same benefits as what I listed after the code. It still makes it very easy to achieve the same dangerous coding style that Ramdas current dispatch mechanism does.

@joneshf 
> I think it's actually intrinsic to js that fantasyland works as prototype methods and whatever.

Yes. I certainly agree with you on that. It is that technique that makes it all possible. But to me it is an implementation detail that should not be exposed to end users. As I said before: "Users should be thinking in terms of type classes (of course we don't have actual type classes but the specifications gives something similar) they should not be thinking in terms of dispatching and methods!"

scott-christopher:
> `R.map(someFn, someStream).onEnd(someEndHandler);`
These are all signs of a poor leaky abstraction.

How would you handle this differently with a fantasy-land functor that also has an `onEnd` method that you needed to call, or `R.map(someFn, maybeSomething).getOrElse(defaultSomething)`?

> Did you see the Shepherd implementation I posted before? It does not seem to me like your proposal gives the same benefits as what I listed after the code. It still makes it very easy to achieve the same dangerous coding style that Ramdas current dispatch mechanism does.

I did see your example, though I don't see how that can support dispatching for transducers. The example I gave also ticks every one of the dot points you listed after your example when the dispatching is explicitly applied by the user.

> * Ramda will be simpler
> * Ramda functions will be reserved for their intended sensible use. I.e. when I see R.map in a code base I immediately know it is a proper map on a proper functor.
> * When one wants to dispatch the dispatching is explicit instead of added inside another function.
> * New dispatchers can easily be added. This avoids the complexity of aliases.

Dispatching logic is moved out of the individual Ramda functions and into external declarations. If you want transducer support in Ramda, include the transducer dispatches. If you want fantasy-land support, include the fantasy-land dispatches. And if you don't need either, then don't include them. I fail to see the *danger* here.

As for typeclasses, an approach similar to my example would also allow for providing fantasy-land implementations for native objects such as `Function`s without messing with their prototypes.

Having said that, I still don't particularly think this approach warrants the extra boilerplate.

paldepind:
> How would you handle this differently with a fantasy-land functor that also has an onEnd method that you needed to call, or R.map(someFn, maybeSomething).getOrElse(defaultSomething)?

Then you should create your own invoker or use a library that offers such an invoker. `invoker` was created exactly to handle things like that.

> I did see your example, though I don't see how that can support dispatching for transducers.

That is intentionally. It only dispatches on methods. That is simple and that is a good things. The library does one single thing and does it well.

>  The example I gave also ticks every one of the dot points you listed after your example when the dispatching is explicitly applied by the user.

Indeed. But the "when" here is my gripe. As far as I understand you still offer the invoker overloading even for things besides fantasy land?

PR #1174, opened 2015-06-12
lift ban on use of Array.prototype.push

davidchambers:
As proposed in [#1050][1]

[1]: https://github.com/ramda/ramda/pull/1050/files#r32039355


paldepind:
Yes! Certainly an advantage with regards to readability.

PR #1175, opened 2015-06-12
refactor R.curryN

davidchambers:
I realized while working on plaid/sanctuary#50 that `Array.prototype.shift` obviates the need for the awkward `currentArgsIdx` variable.

> Showing __1 changed file__ with __13 additions__ and __35 deletions__.

:smile:

I realize that compactness is not the end goal, but in this case the difference is significant: the function was 50 lines and is now 20 lines, making it significantly more approachable. I have MacVim configured to display 35 lines of code, so the function now easily fits on my screen.

I'll switch to `while` loops upon request. I first wanted to show just how compact this code can be.

### DIFF
--- a/src/curryN.js
+++ b/src/curryN.js
@@ -1,5 +1,4 @@
 var _curry2 = require('./internal/_curry2');
-var _slice = require('./internal/_slice');
 var arity = require('./arity');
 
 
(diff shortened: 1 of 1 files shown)


paldepind:
I have already refactored curryN #1155.

This refactor reintroduces `Array.prototype.concat` and `Array.prototype.shift` which I deliberately took out in a previous PR for performance reasons. My refactor doesn't and it completely avoids `_slice`. You're also slicing on `arguments` which should be avoided for performance reasons.

I'm currently cleaning up the before mentioned PR. And with cleaning up I actually mean taking out `for`-loops ;)

paldepind:
Are you fine with code like this:

```javascript
xs.push(x != null && x['@@functional/placeholder'] === true ? ys.shift() : x);
```

It's fine with me but I'm surprised since using `shift` like this is doing side effects in a very similar way to using `++` inside other statements and you strongly argued against that.

CrossEye:
I'm with Simon.  I was very glad to see `shift` go, although I think I was the one to introduce it originally. 

I'd rather see #1155 go forward.   The code seems clearer and less tricky.

davidchambers:
> Are you fine with code like this:
>
> ```javascript
> xs.push(x != null && x['@@functional/placeholder'] === true ? ys.shift() : x);
> ```
>
> It's fine with me but I'm surprised since using `shift` like this is doing side effects in a very similar way to using `++` inside other statements and you strongly argued against that.

I wouldn't say it makes me happy. :)

As I was refactoring, though, I was keen to remove `currentArgsIdx`. Using `shift` is clearer to me than maintaining the read position. I was not considering performance (though I began refactoring because I realized we currently use `_slice(arguments)` in two places where `arguments` would be equivalent).

Closing in favour of #1155. I didn't mean to open a competing pull request, @paldepind. :)

CrossEye:
Compteing PRs aren't a big deal, in my mind.  Sometimes that's the best way to test out ideas.

#1155 just took a while because I dragged out a debate on `for` versus `while` (with some able help from @paldepind!)

paldepind:
> I didn't mean to open a competing pull request, @paldepind. :)

No problem. I didn't interpret it like that

> #1155 just took a while because I dragged out a debate on for versus while (with some able help from @paldepind!)

I think we took equal part in that ;)

PR #1176, opened 2015-06-12
Add undefined as return value in docs where applicable

TheLudd:
This adds undefined as a possible return value in the docs for functions
that might return undefined. Examples has also been added showing how
the functions might return undefined.

This was previously done for `find` and `findLast` but the docs were not
consistent.

### DIFF
--- a/src/head.js
+++ b/src/head.js
@@ -8,11 +8,12 @@ var nth = require('./nth');
  * @func
  * @memberOf R
  * @category List
- * @sig [a] -> a
+ * @sig [a] -> a | Undefined
  * @param {Array} list The array to consider.
  * @return {*} The first element of the list, or `undefined` if the list is empty.
  * @example
  *
  *      R.head(['fi', 'fo', 'fum']); //=> 'fi'
+ *      R.head([]); //=> undefined
  */
 module.exports = nth(0);
--- a/src/last.js
+++ b/src/last.js
@@ -7,11 +7,12 @@ var nth = require('./nth');
  * @func
  * @memberOf R
  * @category List
- * @sig [a] -> a
+ * @sig [a] -> a | Undefined
  * @param {Array} list The array to consider.
  * @return {*} The last element of the list, or `undefined` if the list is empty.
  * @example
  *
  *      R.last(['fi', 'fo', 'fum']); //=> 'fum'
+ *      R.last([]); //=> undefined
  */
 module.exports = nth(-1);
--- a/src/nth.js
+++ b/src/nth.js
@@ -9,7 +9,7 @@ var _nth = require('./internal/_nth');
  * @func
  * @memberOf R
  * @category List
- * @sig Number -> [a] -> a
+ * @sig Number -> [a] -> a | Undefined
  * @param {Number} idx
  * @param {Array} list
  * @return {*} The nth element of the list.
--- a/src/path.js
+++ b/src/path.js
@@ -8,11 +8,12 @@ var _path = require('./internal/_path');
  * @func
  * @memberOf R
  * @category Object
- * @sig [String] -> {*} -> *
+ * @sig [String] -> {k: v} -> v | Undefined
  * @param {Array} path The path to use.
  * @return {*} The data at `path`.
  * @example
  *
  *      R.path(['a', 'b'], {a: {b: 2}}); //=> 2
+ *      R.path(['a', 'b'], {c: {b: 2}}); //=> undefined
  */
 module.exports = _curry2(_path);
--- a/src/prop.js
+++ b/src/prop.js
@@ -7,7 +7,7 @@ var _curry2 = require('./internal/_curry2');
  * @func
  * @memberOf R
  * @category Object
- * @sig s -> {s: a} -> a
+ * @sig s -> {s: a} -> a | Undefined
  * @param {String} p The property name
  * @param {Object} obj The object to query
  * @return {*} The value at `obj.p`.


CrossEye:
I like it except that I would capitalize `Undefined` to match things like `String`, `Bool`, and `Int`.

TheLudd:
@CrossEye Would you write it `Null` as well? To me that looks very strange, they are after all primitives are they not? This is what the documentation for `find` looks like today, undefined lower case. I prefer that. http://ramdajs.com/0.14/docs/#find

CrossEye:
Yes, I would write `Null`.  These are not primitives here but types.  I would like to follow the Haskell conventions that single lower-case letters represent specific arbitrary, but fixed-for-this-signature, types, and all other names use UpperCamelCase.

I haven't done it, but I am planning on making a new pass through the signatures based on what we've learned about them since the original pass.

buzzdecafe:
idk, `Undefined` and `Null` feel a little weird. But I don't have real strong fellings either way

TheLudd:
Regardless of first letter case, are there any functions missing after this?

CrossEye:
Almost certainly.   I'll try to take a look when I'm in front of a computer.   The first one that comes to mind is `reduce`.

buzzdecafe:
i took a pass thru the docs and didn't see any that have `undefined` as part of their contract (as opposed to pathological casses like `reduce -> undefined`)

CrossEye:
Cool.  Haven't had any time.  But should we document the -- pathological, really? -- edge cases, too?

TheLudd:
@CrossEye IMO no.

Reduce returns the same type that the accumulator function returns. If that is undefined, reduce will return undefined. And as I see it, that is already specified in the type signature.

davidchambers:
I agree with @CrossEye on the capitalization. The Undefined type has a single value. This hints at the existence of the Undefined constructor:

```javascript
> Object.prototype.toString.call(undefined)
'[object Undefined]'
```

I don't like to distinguish between null and undefined, however, so I would prefer to use `?`. For example:

```
head :: [a] -> a?
```

buzzdecafe:
i'm pretty sure we don't have any functions that return `null` by contract, so we can punt on that.

CrossEye:
> I don't like to distinguish between null and undefined, however, so I would prefer to use `?`. For example:
> 
> ```
> head :: [a] -> a?
> ```

Actually, I like that a lot.  I really don't like the `|` in our fairly-strongly typed signatures; it feels squishy.  The `?` annotates it more as an exception case.

davidchambers:
My point is that if a function may return null/undefined, callers should use `== null` rather than `=== null` or `=== undefined`. I don't believe in differentiating JavaScript's two nil values. This is just one opinion, of course. :)

buzzdecafe:
i prefer the bar, myself. It's more consistent with the ML/Haskell style, and more explicit -- I don't have to go look up what `?` means somewhere. 

On the other hand, how about grouping `null` and `undefined` under one pseudo-type `Nil` and have the sig be `head :: [a] -> a | Nil`? You would have to go lookup what `Nil` means, but we may have that problem anyways if @paldepind 's `Key` idea gets in

TheLudd:
The two nil values may be unpopular. But it is not as if `find` can return either of them. If no match is found, `undefined` is the value that is returned.

CrossEye:
> Reduce returns the same type that the accumulator function returns. If that is undefined, reduce will return undefined. And as I see it, that is already specified in the type signature.

But it could be a case that the function is never called, too, if you start with an empty list.  If you have an `undefined` accumulator too, then you will have an `undefined` result:

```js
var avg = function(ls) {
  var count = 0, total = 0;
  return R.reduce(function step(curr, next) {
    count++;
    total += (next || 0) 
    return total / count;
  }, undefined, ls);
};

avg([8, 6, 7, 5, 3, 0, 9]); //=> 5.428571428571429
avg([]); //=> undefined
```

Note that the `step` function will only return a number, but that `reduce` can still return `undefined`.  This is the sort of place where I think there is still ambiguity.

buzzdecafe:
you can do all sorts of crazy :hankey: in js to mess with type. i don't think we should strive to comprehensively cover them all in a signature. The purpose of the signature is to describe the intent, not all the crazy edge cases.

davidchambers:
> The purpose of the signature is to describe the intent, not all the crazy edge cases.

I'm with @buzzdecafe on this (for the second time today, which must be a record!).

buzzdecafe:
> I'm with @buzzdecafe on this (for the second time today, which must be a record!).

yeah, this could be trouble

CrossEye:
Right, and I am somewhat mixing up the two things presented by @TheLudd today:  The signatures probably should be simple.  But functions getting wrapped in `Maybe`s should probably include all those that could reasonably return `undefined`.

TheLudd:
@CrossEye You do present a valid argument that `reduce` might return `undefined`, but I am not at all enthusiastic about wrapping it in a `Maybe`. That was not an idea I had...

davidchambers:
`reduce` is only a partial function if it doesn't take an initial value. Since [`R.reduce`][1] requires an initial value, its output is defined for every input.


[1]: http://ramdajs.com/docs/#reduce

CrossEye:
@TheLudd:  That's the thing.  Once you set down that road, it seems you really have to commit to it.  That's why I was reluctant to do this same thing some time back.  And now David went off and started the excellent Sanctuary to prove me wrong!  :smiling_imp:

TheLudd:
@CrossEye But it seems like @davidchambers is right. `reduce` does not return `undefined` for an empty list. It returns the initial argument. And so I would not consider `reduce` to be a part of functions that need to specify that they might return `undefined`.

http://ramdajs.com/repl/?v=0.14.0#R.reduce(R.add%2C%200%2C%20%5B%5D)

davidchambers:
What's the plan? I'm fine with either `head :: [a] -> a?` or `head :: [a] -> a | Undefined`. I prefer the former.

TheLudd:
Come up with a decision and I will change accordingly.

buzzdecafe:
my vote is for `head :: [a] -> a | Undefined`

davidchambers:
Works for me. Update this pull request when you have a few minutes, @TheLudd. :)

TheLudd:
Done

davidchambers:
:deciduous_tree:

PR #1180, opened 2015-06-12
fix lint error

davidchambers:
It would be nice if GitHub indicated in a pull request whether the tests will pass if the pull request is merged, to catch this sorts of thing.

### DIFF
--- a/src/uniq.js
+++ b/src/uniq.js
@@ -31,12 +31,12 @@ module.exports = (function() {
     /* global Set */
     var item,
         set = new Set(),
-        idx = -1,
+        idx = 0,
         len = list.length,
         items = [],
         uniqs = [];
 
-    while (++idx < len) {
+    while (idx < len) {
       item = list[idx];
       // `_contains` is also used to differentiate between
       // +0 and -0, as the native Set does not.
@@ -50,6 +50,7 @@ module.exports = (function() {
           uniqs.push(item);
         }
       }
+      idx += 1;
     }
     return uniqs;
   }


buzzdecafe:
it does -- and at the time of that PR it did

davidchambers:
> it does

Where?

buzzdecafe:
"All is well — The Travis CI build passed"

it's travis that runs the tests. i ran into the same thing with the `keys` refactor, which overlapped with the `++` changes. so i rebased and amended and all was well.

davidchambers:
I'm suggesting that GitHub should *not* state "All is well" if merging a pull request will cause tests to fail. One shouldn't need to pull the branch, merge with master, and run the tests locally in order to know whether a pull request is fit to merge.

buzzdecafe:
hmmm. on the `keys` one, it did *not* say that, it instead said something like "this cannot be automatically merged."

davidchambers:
> hmmm. on the `keys` one, it did not say that, it instead said something like "this cannot be automatically merged."

Right, but that's something different. That's GitHub saying that there are merge conflicts which cannot be resolved automatically. I'm suggesting that Travis should simulate the merge and ensure that tests will also pass after merging.

buzzdecafe:
i see. might be a bit of a tall order, since there are many events that might kick off a test suite.

scott-christopher:
If you want to test post merge, you should just be able to update `.travis.yml` to something like:
```yaml
before_script:
- git pull origin master && make clean && make dist/ramda.js
```

PR #1182, opened 2015-06-13
Change mapped variable in construct examples into an array

jackjennings:
My assumption is that `allConfigs` should actually be an array, since that would be a more plausible object to map over.


buzzdecafe:
yes indeed.

buzzdecafe:
thanks -- this should show up after we publish 0.15, which i expect will be very soon

PR #1183, opened 2015-06-14
Add "See also" between `adjust` and `update`

paldepind:


davidchambers:
:deciduous_tree: Good idea!

paldepind:
Indeed. It was @scott-christopher's [here](https://github.com/ramda/ramda/pull/1153#issuecomment-109003002).

PR #1184, opened 2015-06-14
deprecate R.substring

davidchambers:
`R.slice` dispatches to `slice`, so `R.substring` should be deprecated just as `R.substringTo` and `R.substringFrom` were in #1159.

### DIFF
--- a/src/substring.js
+++ b/src/substring.js
@@ -14,6 +14,7 @@ var slice = require('./slice');
  * @param {String} str The string to slice.
  * @return {String}
  * @see R.slice
+ * @deprecated since v0.15.0
  * @example
  *
  *      R.substring(2, 5, 'abcdefghijklm'); //=> 'cde'


paldepind:
This is very nice!

CrossEye:
I'm still very wary of the API via implementation progression we seem to be on.

The main question should be whether we want to have a public `substring` function.  If we do, then we should support a function with that name, even if its functionality overlaps with that of another one, even it it's entirely subsumed by `slice`.

In other words, as I think I've written elsewhere.  the API should always remain king.

jethrolarson:
It'd be nice if our deprecations would say what the new solution is where applicable. I didn't assume that R.drop would work on strings.

davidchambers:
> It'd be nice if our deprecations would say what the new solution is where applicable.

Good idea. I'm sure we'd all be happy to see a pull request against [ramda/ramda.github.io][1]. :)


[1]: https://github.com/ramda/ramda.github.io

CrossEye:
But .in the future, we really should add a `@see` tag when we deprecate if the equivalent function is available, or some other similar mechanism in the comments

TheLudd:
Just putting a formal objection here. If I want to do a substring operation then calling `substring` just feels much more natural then `slice` to me. I find it a whole lot more expressive. In this case *especially* the names `substringFrom` and `substingTo` have just been much more clear than in any other lib or language. I value that much more than what could ever be the benefit of this.

A lot of these changes are going on and I just have a hard time understanding why. What is unclear to me is the actual intention of these changes. Is it to not have any two functions that can achieve the same goal? If that is correct then I suspect the ramda API could be reduced to about 10 functions depending on how you look at it.

davidchambers:
The goal is to avoid having two names for the same function, but you're free to define aliases if like:

```javascript
R.substring = R.slice;
R.substringTo = R.take;
R.substringFrom = R.drop;
```

TheLudd:
It might be possible to implement these functions in the same way but from the outside one set works on strings and the other on arrays. `slice`, `take`, `drop` belongs to the List category. This to me is not "API is king".

paldepind:
@TheLudd It's odd. I feel the exact opposite way. To me a string is just a list of characters. Using `slice`, `take` and `drop` on lists of characters feel very natural to me. On the other hand having to use functions that embed the word "string" in their name feels unnatural. Why do I have to use a string specific function when I'm doing a general list operation on a list of characters?

I'm not saying that Ramda should accept strings in all the list functions. But doing it in the most obvious candidates feels like a better API to me.

TheLudd:
@paldepind 
A string can be viewed as a list of characters. I can also be viewed as an object. I am pretty sure that a list can be viewed as an object as well. Claiming that this is how we view these types when we do programming feels a bit forced to me. Here is a list of characters: `[ 'f', 'o', 'o' ]`. Is that interchangeable with `'foo'` ?

While I don't share your opinion that a string is *just* a list of characters, I won't claim that you are wrong. But I doubt anyone can disagree with the claim that at string is a string.

Technically you are correct but I think the difference in concepts is important and and function names that clearly express their purpose just makes code more easy to read.

davidchambers:
In dynamically typed languages we sometimes use function names, parameter names, and variable names to provide hints as to the types of the values in our programs.

The following two examples produce equivalent functions:

```javascript
var getAccountPrefixes = function(accountNumbers) {
  return map(function(accountNumber) {
    return substringTo(4, accountNumber);
  }, accountNumbers);
};
```

```javascript
//  prefixes :: Functor f => f a -> f a
var prefixes = map(take(4));
```

The difference is the number of hints these functions provide as to the types we're expecting. The problem with using names such as `accountNumbers` and `substringTo` is they suggest that our function is more specialized than it really is. We *could* use `getAccountPrefixes` on a list of lists of Date values, but it would be confusing to do so. Giving things descriptive names limits code reuse.

paldepind:
@TheLudd 
I think @davidchambers said it very well. Functional programming is very much about identifying common properties in different type of things. In some languages a string really is just a list of characters. We also have `[]` notation for accessing characters in in strings which is also evidence that people like to think of them as lists of characters.

I think the only reason why a string in JavaScript is not just an array of characters is that arrays are mutable where as strings are immutable. But that is a technical difference and it is one that doesn't apply to Ramda.

PR #1187, opened 2015-06-14
Implements `lensIndex` setter using `update`

scott-christopher:
As per @jethrolarson's suggestion here: https://github.com/ramda/ramda/pull/971#issuecomment-111279241, `lensIndex` now uses `update` for its setter function, providing a nice symmetry with `lensProp`:

```js
function lensProp(k) {
  return lens(prop(k), assoc(k));
}
```
```js
function lensIndex(n) {
  return lens(nth(n), update(n));
}
```

### DIFF
--- a/src/lensIndex.js
+++ b/src/lensIndex.js
@@ -1,6 +1,8 @@
-var _slice = require('./internal/_slice');
+var _curry1 = require('./internal/_curry1');
 var lens = require('./lens');
 var nth = require('./nth');
+var update = require('./update');
+
 
 /**
  * Creates a lens that will focus on index `n` of the source array.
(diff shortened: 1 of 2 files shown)


CrossEye:
:herb:

davidchambers:
We could even implement this as `converge(lens, nth, update)`.

src/lensIndex.js
```diff
@@ -21,7 +21,5 @@ var nth = require('./nth');
  *     headLens.map(function(x) { return x + 1; }, [10, 20, 30, 40]); //=> [11, 20, 30, 40]
  */
 module.exports = function lensIndex(n) {
-  return lens(nth(n), function(x, xs) {
-    return _slice(xs, 0, n).concat([x], _slice(xs, n + 1));
-  });
+  return lens(nth(n), update(n));
 };
```
davidchambers: This function is missing its `_curry1` wrapper. I thought we had a test in place to catch this. Apparently not.

scott-christopher:
I've updated the change to include the `_curry1` wrapper. I also tried using `converge`, but it caused a bunch of the composition tests to fail.

src/lensProp.js
```diff
@@ -23,7 +25,6 @@ var prop = require('./prop');
  *     phraseLens.set('Ooh Betty', obj1); //=> { phrase: 'Ooh Betty'}
  *     phraseLens.map(R.toUpper, obj2); //=> { phrase: "WHAT'S ALL THIS, THEN?"}
  */
-module.exports = function(k) {
+module.exports = _curry1(function(k) {
```
davidchambers: Mind naming this function expression? We're not very consistent, but I know @CrossEye likes named function expressions.

scott-christopher: Sure thing. Updated.

davidchambers:
:deciduous_tree:

CrossEye:
:star2:

PR #1189, opened 2015-06-15
Revert "R.uniq implementation to use Set if available"

### LINKED ISSUE #1188: R.uniq bug with Set
At the risk of being labeled [a concern troll](https://github.com/ramda/ramda/pull/907#issuecomment-111716832) I found that enviros w/o `Set`, e.g. IE 10 or Node 0.10:
```js
R.uniq([[1],[1]])
// => [[1]]
```

While enviros w/ `Set`, e.g. IE 11 or Node 0.12:
```js
R.uniq([[1],[1]])
// => [[1],[1]]
```

davidchambers:
Fixes #1188

### DIFF
--- a/src/uniq.js
+++ b/src/uniq.js
@@ -1,5 +1,3 @@
-var _contains = require('./internal/_contains');
-var _curry1 = require('./internal/_curry1');
 var equals = require('./equals');
 var uniqWith = require('./uniqWith');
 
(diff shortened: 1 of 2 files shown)


paldepind:
Yes. Appears to be the only solution given our definition of equality.

scott-christopher:
It might be a good idea to add `[[1],[1]]` as a test for `uniq` here too.

CrossEye:
:+1:

davidchambers:
> It might be a good idea to add `[[1],[1]]` as a test for `uniq` here too.

Good idea. I have done so.

buzzdecafe:
LGTM
:cow2:

Issue #1186, opened 2015-06-14
0.15.0 upgrade guide


davidchambers:
Changes: https://github.com/ramda/ramda/compare/v0.14.0...v0.15.1

Removals:

  - :x: `R.isNaN` #1112

Deprecations:

  - :floppy_disk: `R.arity` #1111
  - :floppy_disk: `R.eq` #1096, #1190
  - :floppy_disk: `R.eqDeep` #1096, #1190
  - :floppy_disk: `R.filterIndexed` #1061
  - :floppy_disk: `R.forEachIndexed` #1061
  - :floppy_disk: `R.invoke` #1146
  - :floppy_disk: `R.mapIndexed` #1061
  - :floppy_disk: `R.reduceIndexed` #1061
  - :floppy_disk: `R.reduceRightIndexed` #1061
  - :floppy_disk: `R.rejectIndexed` #1061
  - :floppy_disk: `R.strIndexOf` #1113
  - :floppy_disk: `R.strLastIndexOf` #1113
  - :floppy_disk: `R.substring` #1184
  - :floppy_disk: `R.substringFrom` #1159
  - :floppy_disk: `R.substringTo` #1159

Changes:

  - all Ramda functions now accept any compliant placeholder, not just `R.__` #1156
  - all Ramda functions except `R.identical` now use value-based equality #1096
  - `R.chain` is now a transducer #1101
  - `R.evolve` is now recursive #1141
  - `R.forEach` now dispatch to `forEach` if available #1138
  - `R.identical` (née `R.eq`) now handles `new Number(0)` correctly #1099
  - `R.identical` (née `R.eq`) no longer always returns true when given NaN as its first argument #1100
  - `R.indexOf` and `R.lastIndexOf` now dispatch to their respective methods if available #1113
  - :warning: `R.invoker` no longer accepts initial arguments #1147
  - `R.pick` now supports numbers as property names #1122
  - :warning: `R.propOr` now returns the default value if the property is present but null/undefined #1014
  - `R.take` and `R.drop` can now operate on strings #1159

Additions:

  - :new: `R.addIndex` #1061
  - :new: `R.equals` #1096
  - :new: `R.identical` #1096
  - :new: `R.reduced` #1118

TheLudd:
Any reason `eq` and `eqDeep` were not deprecated?

paldepind:
> Any reason eq and eqDeep were not deprecated?

[...]

Issue #288, opened 2014-08-24
Documentation Format


CrossEye:
I'm trying to build a consistent documentation format, now that most of the code has been documented.  I've gone through, and normalized most of it to look like this sample:

``` javascript
/**
 * Returns a partial copy of an object containing only the keys that
 * satisfy the supplied predicate.
 *
 * @static .
 * @memberOf R
 * @category Object
 * @param {Function} pred A predicate to determine whether or not a key
 *        should be included on the output object.
 * @param {Object} obj The object to copy from
 * @return {Object} A new object with only properties that satisfy `pred`
 *         on it.
 * @see pick
 * @example .
 *
 *      function isUpperCase(x) { return x.toUpperCase() === x; }
 *      pickWith(isUpperCase, {a: 1, b: 2, A: 3, B: 4}) // => {A: 3, B: 4}
 */
R.pickWith = curry2(pickWith);
```

A few things to note: 
- To run these in [docker](https://github.com/jbt/docker), I had to add a character to the empty tags, `@private`, `@static`, and `@example`.  We can do a global search-and-replace as soon as we can get JSDoc working.
- Also for [docker](https://github.com/jbt/docker), I had to indent the examples five spaces beyond the normal one for the comment syntax.  Although this is one more than I think necessary, I believe this indentation really helps readability.  When we stop using docker we might reduce it a bit, but I'd like to keep some indentation here.  
- I strove really hard to maintain a consistent tag order, although I'm sure I got some of them wrong.  It would be nice if we can keep this universally.  The only ones we use missing from the above are `@private` which would be the very first, and `@alias` which would come after `@return` and before `@see`.  I don't think this particular order is in any way stronger than another, but it would be very nice to remain consistent.
- I have kept all the `@`-tags together, with no empty lines, but put one blank line after the `@example` tag and before the actual example code.  Again, I think this improves readability.
- For tags that take more than a single line, I've tried to indent to match the start of the first character after the tag.
- The `@static` tag is probably redundant since we use `@memberOf` but we can worry about that when we get JSDoc working.
- I haven't gone through yet to check where we're missing `@category` tags and whether our naming is consistent, because I didn't find such a tag on http://usejsdoc.org/ or any likely synonym, so I don't know how it'll be used.  But if we can use it, we will need to go through the docs for this.
[...]

PR #1190, opened 2015-06-15
Deprecate `eq` and `deepEq`

paldepind:
See comments by @TheLudd in #1186.

This commit deprecates `eq` and `eqDeep`. `eqDeep` is implemented as an alias to `equals`.

### DIFF
--- a/src/eq.js
+++ b/src/eq.js
@@ -1,0 +1,31 @@
+var _curry2 = require('./internal/_curry2');
+var _eq = require('./internal/_eq');
+
+
+/**
+ * Tests if two items are equal.  Equality is strict here, meaning reference equality for objects and
+ * non-coercing equality for primitives.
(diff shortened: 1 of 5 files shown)


davidchambers:
Let's include the following `@see` pointers:

  - `R.eq` → `R.identical`; and
  - `R.eqDeep` → `R.equals`.

paldepind:
Indeed a good idea.

buzzdecafe:
:cow2:

CrossEye:
I can't really complain that the tags are out of place, since I dropped the ball on #288.  At some point, we need to get a wiki page detailing the format for our JSDoc tags.

:star2:

TheLudd:
Thanks for the effort guys.
Is it possible to release this as `0.15.1`?

PR #1203, opened 2015-06-19
rework R.commute and R.commuteMap docstrings

davidchambers:
I believe we want to tell those reading about [`R.commute`][1] to look at [`R.commuteMap`][2] if they're dealing with functors other than Array. Currently *both* functions claim to be more useful for this purpose. ;)

[1]: http://ramdajs.com/docs/#commute
[2]: http://ramdajs.com/docs/#commuteMap

### DIFF
--- a/src/commuteMap.js
+++ b/src/commuteMap.js
@@ -9,29 +9,22 @@ var map = require('./map');
  * Turns a list of Functors into a Functor of a list, applying
  * a mapping function to the elements of the list along the way.
  *
- * Note: `commuteMap` may be more useful to convert a list of non-Array Functors (e.g.
- * Maybe, Either, etc.) to Functor of a list.
- *
  * @func
  * @memberOf R
  * @category List
  * @see R.commute
- * @sig (a -> (b -> c)) -> (x -> [x]) -> [[*]...]
+ * @sig Functor f => (f a -> f b) -> (x -> f x) -> [f a] -> f [b]
  * @param {Function} fn The transformation function
  * @param {Function} of A function that returns the data type to return
- * @param {Array} list An Array (or other Functor) of Arrays (or other Functors)
- * @return {Array}
+ * @param {Array} list An array of functors of the same type
+ * @return {*}
  * @example
  *
- *      var plus10map = R.map(function(x) { return x + 10; });
- *      var as = [[1], [3, 4]];
- *      R.commuteMap(R.map(function(x) { return x + 10; }), R.of, as); //=> [[11, 13], [11, 14]]
- *
- *      var bs = [[1, 2], [3]];
- *      R.commuteMap(plus10map, R.of, bs); //=> [[11, 13], [12, 13]]
- *
- *      var cs = [[1, 2], [3, 4]];
- *      R.commuteMap(plus10map, R.of, cs); //=> [[11, 13], [12, 13], [11, 14], [12, 14]]
+ *      R.commuteMap(R.map(R.add(10)), R.of, [[1], [2, 3]]);   //=> [[11, 12], [11, 13]]
+ *      R.commuteMap(R.map(R.add(10)), R.of, [[1, 2], [3]]);   //=> [[11, 13], [12, 13]]
+ *      R.commuteMap(R.map(R.add(10)), R.of, [[1], [2], [3]]); //=> [[11, 12, 13]]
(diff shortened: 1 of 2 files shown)


buzzdecafe:
both functions are more useful for other Functors. the only difference is `commuteMap` will apply a transformation on the way out, hence the `Map` in the name. if you just want to convert: `[Just(1), Just(2), Just(3)] => Just([1, 2, 3])` then `commute` is what you want. That is clearly more useful than the nested array case.

davidchambers:
In that case I'm misunderstanding the note: more useful than what?

buzzdecafe:
i'm suggesting that `[Just(1), Just(2), Just(3)] => Just([1, 2, 3])` is more useful than `[[1], [2], [3]] => [[1, 2, 3]]`. and the array cases with longer arrays are hard to see the usefulness: 

`R.commute(R.of, [[1, 2], [3, 4], [5, 6]]) //=> [[1, 3, 5], [2, 3, 5], [1, 4, 5], [2, 4, 5], [1, 3, 6], [2, 3, 6], [1, 4, 6], [2, 4, 6]]

davidchambers:
> the array cases with longer arrays are hard to see the usefulness

I agree, though this is currently the only behaviour the examples demonstrate.

> `[Just(1), Just(2), Just(3)] => Just([1, 2, 3])`

This is a wonderful example. We should include it.

Now that I understand these functions a little better I'll have another go at reworking their docstrings.

buzzdecafe:
>  this is currently the only behaviour the examples demonstrate.

yeah i only included examples that you could run with just `R`. But given that:

* I think we have some other examples that use ADTs, and
* the ADT example really illustrates what is useful about this function,

I think using it as an example is the appropriate thing to do

davidchambers:
I've made some changes. The type signatures are now:

```haskell
commute :: Functor f => (x -> f x) -> [f a] -> f [a]
commuteMap :: Functor f => (f a -> f b) -> (x -> f x) -> [f a] -> f [b]
```

What I'm not sure about is whether the `[f _] -> f [_]` portions are too specific. Do these functions specifically take a *list* of functors of the same type, or is any functor of functors acceptable? If the latter, please provide an example and suggest a more general signature. :)

buzzdecafe:
> Do these functions specifically take a list of functors of the same type, or is any functor of functors acceptable?

it has to be something Traversable. So the List implementation falls right out. Other ADTs may be a bit trickier.

davidchambers:
> it has to be something Traversable.

Something like this, then?

```haskell
commute :: (Functor f, Traversable t) => (x -> f x) -> t (f a) -> f (t a)
```

buzzdecafe:
... i think so. should be completely impenetrable to most js developers. :smiling_imp:

davidchambers:
> should be completely impenetrable to most js developers.

For these particular functions this is fine. Since they're unlikely to be useful with just the built-in types, these functions have a more limited audience.

I'd prefer to stick with

```haskell
commute :: Functor f => (x -> f x) -> [f a] -> f [a]
```

though, as it's easier to grok. It's fine to pretend that these functions are more limited than they really are.

src/commuteMap.js
```diff
@@ -9,29 +9,22 @@ var map = require('./map');
...
  * @memberOf R
  * @category List
  * @see R.commute
- * @sig (a -> (b -> c)) -> (x -> [x]) -> [[*]...]
+ * @sig Functor f => (f a -> f b) -> (x -> f x) -> [f a] -> f [b]
  * @param {Function} fn The transformation function
  * @param {Function} of A function that returns the data type to return
- * @param {Array} list An Array (or other Functor) of Arrays (or other Functors)
- * @return {Array}
+ * @param {Array} list An array of functors of the same type
```
buzzdecafe: this is definitely better

buzzdecafe:
LGTM. :cow2:

CrossEye:
Yes, this is great!

joneshf:
But these two functions don't work with just any `Functor`, both require `Applicative`s, since `Functor` isn't powerful enough to do what the signature implies it does. Shouldn't the docs state `Applicative` if they're going to state anything?

For instance, you can't use these functions with ramda-fantasy's `Tuple` (since it's not `Applicative`), but the type signature implies that you can.

davidchambers:
Good point, @joneshf. Fancy opening a pull request to correct this?

PR #1204, opened 2015-06-19
demonstrate that R.isEmpty({}) evaluates false

### LINKED ISSUE #1194: R.isEmpty({}) returns false
```
R.isEmpty('') returns true,
R.isEmpty([]) returns true
```

but

`R.isEmpty({}) returns false`

=> If an object is observed as a collection of (enumerable) properties, this can be considered inconsistent. If `Object.keys({})` returns `[]`, it may make sense to consider the object as empty

davidchambers:
Closes #1194

### DIFF
--- a/src/isEmpty.js
+++ b/src/isEmpty.js
@@ -12,10 +12,13 @@ var _curry1 = require('./internal/_curry1');
  * @return {Boolean}
  * @example
  *
- *      R.isEmpty([1, 2, 3]); //=> false
- *      R.isEmpty([]); //=> true
- *      R.isEmpty(''); //=> true
- *      R.isEmpty(null); //=> false
+ *      R.isEmpty([1, 2, 3]);   //=> false
+ *      R.isEmpty([]);          //=> true
+ *      R.isEmpty('');          //=> true
+ *      R.isEmpty(null);        //=> false
+ *      R.isEmpty(R.keys({}));  //=> true
+ *      R.isEmpty({});          //=> false ({} does not have a length property)
+ *      R.isEmpty({length: 0}); //=> true
  */
 module.exports = _curry1(function isEmpty(list) {
   return Object(list).length === 0;


src/isEmpty.js
```diff
@@ -16,6 +16,8 @@ var _curry1 = require('./internal/_curry1');
  *      R.isEmpty([]); //=> true
  *      R.isEmpty(''); //=> true
  *      R.isEmpty(null); //=> false
+ *      R.isEmpty({}); //=> false
```
buzzdecafe: should we annotate this possibly surprising example to point out, e.g.:

`R.isEmpty(null); //=> false; {} is not a list.`

davidchambers: For the `R.isEmpty({})` example, do you mean?

buzzdecafe: yes, for the commented line

davidchambers: You wrote `R.isEmpty(null)` which confused me. :stuck_out_tongue_winking_eye:

buzzdecafe: oops! fixed

davidchambers: I've updated the pull request. It now includes `R.isEmpty({length: 0})` as an example.

src/isEmpty.js
```diff
@@ -12,10 +12,13 @@ var _curry1 = require('./internal/_curry1');
...
- *      R.isEmpty([1, 2, 3]); //=> false
- *      R.isEmpty([]); //=> true
- *      R.isEmpty(''); //=> true
- *      R.isEmpty(null); //=> false
+ *      R.isEmpty([1, 2, 3]);   //=> false
+ *      R.isEmpty([]);          //=> true
+ *      R.isEmpty('');          //=> true
+ *      R.isEmpty(null);        //=> false
+ *      R.isEmpty(R.keys({}));  //=> true
+ *      R.isEmpty({});          //=> false ({} does not have a length property)
```
buzzdecafe: "`{}` is not list, i.e. not array-like" might be better here. i think specifying that it needs a `length` property is an implementation detail we should keep hidden.

src/isEmpty.js
```diff
@@ -12,10 +12,13 @@ var _curry1 = require('./internal/_curry1');
...
+ *      R.isEmpty({length: 0}); //=> true
```
buzzdecafe: and then here, maybe: "('{length: 0}` is array-like)"

davidchambers: Would we then state the following?

```javascript
R.isEmpty('');          //=> true ('' is array-like)
```

I don't think of strings as array-like.

For the record I remain in favour of removing this function. It's just sugar for `R.propEq('length', 0)`, but this fact is not obvious.

buzzdecafe: >  It's just sugar for `R.propEq('length', 0)`

put that way, i think i agree

buzzdecafe: > I don't think of strings as array-like.

now that more & more list functions are being used for strings, we may have to revisit that question. at present, String is excluded from `isArrayLike`

davidchambers:
I'll merge this as it's an improvement on what we have now and I'd like to include this change in v0.15.1 (which I'm about to publish).

PR #1205, opened 2015-06-20
composable lenses

davidchambers:
#### API changes

```
Before:                                     After:

- lens                                      - lens
- lensProp                                  - lensProp
- lensIndex                                 - lensIndex
- lensOn                                    - view
- composeL                                  - set
- pipeL                                     - over
```

#### Creating lenses

```javascript
var nameLens =                              var nameLens =
R.lens(R.prop('name'), R.assoc('name'));    R.lens(R.prop('name'), R.assoc('name'));
```

`R.lens` takes the same arguments as before, but returns a different value. In the new API one never interacts directly with the return value, one simply provides it as an argument to `R.view`, `R.set`, or `R.over`.

The helper functions `R.lensProp` and `R.lensIndex` remain unchanged:

```javascript
var addressLens = R.lensProp('address');    var addressLens = R.lensProp('address');
var headLens = R.lensIndex(0);              var headLens = R.lensIndex(0);
```

Lenses are now just functions, so `R.composeL` and `R.pipeL` are no longer necessary:

```javascript
var streetLens =                            var streetLens =
R.composeL(headLens, addressLens);          R.compose(addressLens, headLens);
```

Note that `R.compose` reads left-to-right here. It's way too late for me to reason about this. :)

The following examples will use this data structure:

```javascript
var alice = {
  name: 'Alice Jones',
  address: ['22 Walnut St', 'San Francisco', 'CA']
};
```

#### `view`

```javascript
nameLens(alice);                            R.view(nameLens, alice);
// => 'Alice Jones'                         // => 'Alice Jones'

streetLens(alice);                          R.view(streetLens, alice);
// => '22 Walnut St'                        // => '22 Walnut St'
```

#### `set`

```javascript
nameLens.set('Alice Smith', alice);         R.set(nameLens, 'Alice Smith', alice);
// => { name: 'Alice Smith'                 // => { name: 'Alice Smith'
//    , address: [ '22 Walnut St'           //    , address: [ '22 Walnut St'
//               , 'San Francisco'          //               , 'San Francisco'
//               , 'CA' ] }                 //               , 'CA' ] }
```

#### `over`

```javascript
streetLens.map(R.toUpper, alice);           R.over(streetLens, R.toUpper, alice);
// => { name: 'Alice Jones'                 // => { name: 'Alice Jones'
//    , address: [ '22 WALNUT ST'           //    , address: [ '22 WALNUT ST'
//               , 'San Francisco'          //               , 'San Francisco'
//               , 'CA' ] }                 //               , 'CA' ] }
```

This API feels more in keeping with the rest of the library (returning a function with two functions attached is rather odd). Also, removing the need for special composition functions is very nice indeed.

I'm opening this pull request for discussion. If we decide we like this change I'll spend some time writing documentation and figuring out the type signatures!

Many thanks to @tel for walking me through the various approaches to creating lenses. The existence of this pull request suggests I absorbed at least some of the information. ;)

/cc @joneshf, Master of Lenses


CrossEye:
Fascinating stuff. 

It's gong to take a while for all this to sink in.  The really obvious good side would be the ability to remove `composeL` and `pipeL`.  @buzzdecafe would have to be happy about reducing the number of `compose`s we have. 

But the downside is just as obvious: creating a lens currently is very straightforward   The new version would be much less clear unless we reintroduced some helper function for it. 

I agree that the API is more in keeping with the rest of Ramda, but only so long as we don't have to introduce some helper function.  But this means I'm torn in two directions. 

I will look at it later in the day when I'm in front of a computer.

buzzdecafe:
:cow2: :cow2: :cow2: :cow2: :cow2: :cow2: :cow2: :cow2:

scott-christopher:
:tophat: This looks great! :tophat: 

I don't think the complexity will be an issue, given basic building blocks like `lensIndex` and `lensProp` still exist. I do wonder whether it would benefit from moving into its own `ramda-lenses` repo though, allowing for more lens-related functions.

scott-christopher:
We could also include a helper function that just takes getter and setter functions to produce a lens, such as:

```js
var lens = curry(function (getter, setter, f, s) {
  return map(partialRight(setter, s), f(getter(s)));
});

var headLens = lens(nth(0), update(0));

view(headLens, ['a', 'b', 'c']); // 'a'
over(headLens, toUpper, ['alice', 'bob']); // [ 'ALICE', 'bob' ]
set(headLens, 'Alice', ['a', 'b', 'c']); // [ 'Alice', 'b', 'c' ]
```

joneshf:
:+1:

But I agree that `lens` still needs to exist, since writing these things by hand is hard enough with a type checker at your back. It's even harder without.

buzzdecafe:
> We could also include a helper function that just takes getter and setter functions to produce a lens,

:heart: this idea, although this made me chuckle to myself:

> `curry(function (getter, setter, f, s) {`

`_curry4` anyone? *I'm* not gonna do it ....

davidchambers:
Thanks for providing a definition of `lens`, @scott-christopher! I've updated this pull request to include it, along with updated `lensProp` and `lensIndex` implementations. I've also updated this pull request's description.

The next step is to work out the type signatures. I've included placeholders in the form `@sig $0 -> $1 -> $2 -> $3` to facilitate in-line suggestions. Any thoughts? :)

src/lens.js
```diff
@@ -1,51 +1,10 @@
...
- *          return out;
- *        }
- *      );
- *      var obj1 = { phrase: 'Absolute filth . . . and I LOVED it!'};
- *      var obj2 = { phrase: "What's all this, then?"};
- *      phraseLens(obj1); // => 'Absolute filth . . . and I LOVED it!'
- *      phraseLens(obj2); // => "What's all this, then?"
- *      phraseLens.set('Ooh Betty', obj1); //=> { phrase: 'Ooh Betty'}
- *      phraseLens.map(R.toUpper, obj2); //=> { phrase: "WHAT'S ALL THIS, THEN?"}
+ * @sig $0 -> $1 -> $2 -> $3 -> $4
```
davidchambers: Expanding `getter`, `setter`, and `f` gives:

    ($0 -> $1) -> ($2,$3 -> $4) -> ($5 -> $6) -> $7 -> $8

davidchambers: The argument to `getter` is the same value as the second argument to `setter`, so must be of the same type:

    ($0 -> $1) -> ($2,$0 -> $3) -> ($4 -> $5) -> $6 -> $7

CrossEye: Isn't $3 also the same as $0 ?

davidchambers: I don't think it *needs* to be. `R.assoc('x')(1, obj)` may or may not return a value of the same type as `obj`. `obj` may not have an `x` field, or it may have an `x` of a different type.

davidchambers: It depends on how granular one wants to be. Perhaps it would be helpful to say `R.assoc('x')(1) :: Object -> Object`, and to generalize to say that `$0` and `$3` are of the same type as you suggest.

davidchambers: The value returned by `getter` is fed to `f`, so we have:

    ($0 -> $1) -> ($2,$0 -> $3) -> ($1 -> $4) -> $5 -> $6

davidchambers: `f` must return a functor, since we `map` over the result:

    Functor f => ($0 -> $1) -> ($2,$0 -> $3) -> ($1 -> f $4) -> $5 -> $6

If we perform substitutions in `map :: (a -> b) -> f a -> f b` for our application of `map` we get:

    ($4 -> $3) -> f $4 -> f $3

So, we can replace `$6` with `f $3`:

    Functor f => ($0 -> $1) -> ($2,$0 -> $3) -> ($1 -> f $4) -> $5 -> f $3

Also, since `setter :: ($2,$0 -> $3)` and the first argument to `setter` is a `$4` (from `($4 -> $3) -> f $4 -> f $3` above), `$2` and `$4` are the same type:

    Functor f => ($0 -> $1) -> ($2,$0 -> $3) -> ($1 -> f $2) -> $4 -> f $3

The last argument to `lens` is now a `$4`. This is given as the first argument to `getter` (`$0`) and as the second argument to `setter` (`$0`). So, `$0` and `$4` are the same type:

    Functor f => ($0 -> $1) -> ($2,$0 -> $3) -> ($1 -> f $2) -> $0 -> f $3

Substituting the identifiers gives:

    Functor f => (a -> b) -> (s,a -> t) -> (b -> f s) -> a -> f t

Is this reasoning sound?

CrossEye: > Is this reasoning sound?

I think it is.  I don't know the literature to know how well established the `s`-`t`-`a`-`b` naming is.  I would really prefer something fairly evocative, if at all possible for `s` and `t`, or, failing that, to revert to `c` and `d`, but not at the expense of mismatching a very standard naming that's used everywhere else.

davidchambers: The `s t a b` naming is used in [Control.Lens.Lens][1]:

> ```haskell
> type Lens s t a b = forall f. Functor f => (a -> f b) -> s -> f t
> ```

I think `s t a b` is preferable to `a b c d` because it splits the type variables into two pairs: (`a`, `b`) and (`s`, `t`). I find it helpful to think about `a` and `b` going together in some way, and `s` and `t` going together. With such a complex type I'll take all the help I can get. :)


[1]: https://hackage.haskell.org/package/lens-4.11/docs/Control-Lens-Lens.html#t:Lens

scott-christopher: >
```
Functor f => (a -> b) -> (s,a -> t) -> (b -> f s) -> a -> f t
```
Is this reasoning sound?
>

It looks correct, though it differs from the identifiers used in `Lens s t a b`:
```
Functor f => (s -> a) -> ((b, s) -> t) -> (a -> f b) -> s -> f t
```

It may be clearer to define the type of `Lens` separately, then refer to it in the `lens` type sig?
```
@type Lens s t a b = Functor f => (a -> f b) -> s -> f t
      Where:
        s = Object type
        t = Updated object type
        a = Focus type
        b = Updated focus type

@sig lens :: (s -> a) -> ((b, s) -> t) -> Lens s t a b
```

tel: That sig looks good. The others were confusing me a bit though.

tel:
Any chance of getting prisms in here? Does Ramda have a position on sum types?

buzzdecafe:
like this? https://github.com/paldepind/union-type

src/lens.js
davidchambers: Thanks again, @scott-christopher! My mappings from `$0`, `$1`, `$2`, and `$3` to `s`, `t`, `a`, and `b` were incorrect.

> ```
> @sig lens :: (s -> a) -> ((b, s) -> t) -> Lens s t a b
> ```

This is very nice! The preceding type definition makes things much easier to understand. I'll update the pull request.

CrossEye:
> Any chance of getting prisms in here?

I'd like a chance to digest these ideas first, honestly.  This signature is far more complex than anything else we have in Ramda, and, while I can understand it while reading the code, it fade very quickly.  I think we need to decide just how complex we want our signatures to get.

Don't get me wrong.  I like this.  But I don't think we should do this in a hurry.

scott-christopher:
The types could be simplified a *little* for the purpose of documentation, if the ability to modify the types is ignored (e.g. `Lens'` in Kmett's library):
```
@type Lens s a = Functor f => (a -> f a) -> s -> f s

@sig lens :: (s -> a) -> ((a, s) -> s) -> Lens s a

@sig view :: Lens s a -> s -> a
@sig set  :: Lens s a -> a -> s -> s
@sig over :: Lens s a -> (a -> a) -> s -> s
```

CrossEye:
> The types could be simplified a _little_ for the purpose of documentation.

I actually think this is a great deal of simplification.  This puts them in the realm of the fairly easily comprehensible, at least to my foggy brain.  I like it a lot!  Especially because we are usually dealing with JS types that don't vary at the granular a level.

tel:
Of course, the difficulty with the "type-fixed" types is that they don't hint at the fact that `over((x) => x.length)` is valid. Since this is documentation instead of an actual type checker maybe we could include more specific and more general types? This is often what `lens` itself does to make the types more digestible.

tel:
For instance, this *very general* operator [`(%%~)`](https://hackage.haskell.org/package/lens-4.11/docs/Control-Lens-Lens.html#v:-37--37--126-) is decorated by 4 more specific types in its documentation.

tel:
@buzzdecafe That's one technique, yup. Here's another: https://github.com/puffnfresh/daggy.

In order to implement Prisms we'll need some kind of defined interface for sum types, though. This could be chosen carefully to be compatible with many different kinds of sum types, perhaps.

davidchambers:
[`R.update`][1] will become sugar for `R.set`, and [`R.adjust`][2] sugar for `R.over`:

```javascript
var lens = R.lensIndex(1);

R.update(1, 11, [0, 1, 2]);
R.set(lens, 11, [0, 1, 2]);

R.adjust(R.add(10), 1, [0, 1, 2]);
R.over(lens, R.add(10), [0, 1, 2]);
```

I like the symmetry between `R.set` and `R.over`; `R.update` and `R.adjust` take the index in different positions.

Lenses have the advantage of being composable. I suggest we deprecate `R.update` and `R.adjust` to avoid having similarly named functions (`R.set` and `R.update`) which serve the same role. (If we decide to keep these functions they should be sugar for their respective lens functions, and the argument order of `R.adjust` should be updated to match.)


[1]: http://ramdajs.com/docs/#update
[2]: http://ramdajs.com/docs/#adjust

tel:
Yeah, those can be seen as specializations of the use of lenses!

tel:
Also, to continue my train of thought from earlier about prisms, here's the best way I know to talk about how lens and prism relate.

You can see the (simple, non-type-changing) type `Lens s a` as containing values which are "proofs" that the types `s` and `(r, a)` are equivalent for some choice of `r`

    Lens s a ===> s ~ exists r . (r, a)

If it's alright, I'll actually write it in this way to denote that tuples are products

    Lens s a ===> s ~ exists r . r * a

On the other hand, this is only one way that a type `s` can be decomposed—into two pieces. The other way is to decompose `s` into two "options" via a sum.

    Prism s a ===> s ~ exists r . r + a

where `(+)` is more commonly called `Either`

    Prism s a ===> s ~ exists r . Either r a

So while Lenses abstract access into types across product boundaries, Prisms abstract access through "variants" or "options". This concept is not very well established in Javascript, though, so it's hard to say what we're talking about.

tel:
So as an example, we might have a prism, an interesting one, `odds :: Prism Int Int` which "pulls out" the option integers have to being either odd or even. Then `over odds (+1)` is a function which only "fires" when its target is odd. And, interestingly, turns all numbers even.

paldepind:
If this gets merged we need to invent a new name for the removed `set` https://github.com/ramda/ramda/pull/1181.

tel:
For the interested, here's an (untested) Javascript implementation of "pure profunctor" lenses which make the duality between lenses and prisms much, much more apparent.

https://gist.github.com/tel/9aca29952f3d1e9d9b04

davidchambers:
Yes, we'll need to come up with a new name for your `set` function, @paldepind. You should open a pull request to reinstate that function when you have some time. We can discuss name options on that thread.

@tel, I think I'll need some help on Gitter to understand your gist. I'm certainly intrigued.

paldepind:
@davidchambers For updating objects I think `evolve` is a much nicer solution than lenses. @CrossEye said it very well:

> This doesn't show that there's something wrong with lenses, but the isomorphism between the input structure and the change shows there's something really right with evolve.

Lenses are nice but not without problems. See [Lens is unidiomatic Haskell](https://ro-che.info/articles/2014-04-24-lens-unidiomatic.html) and [Why I don’t like the lens library](http://fvisser.nl/post/2013/okt/11/why-i-dont-like-the-lens-library.html). I'm sad to see that lenses are now (maybe accidentally) hijacking the name we had chosen for the recursive `merge`.

tel:
What's meant by 

>  the isomorphism between the input structure and the change

?

tel:
Also, I'd take those articles with grains of salt. In each case, the complaint is largely against the `lens` library itself and not lenses as a concept. A deeper complaint tends to be that `lens` likes to act like a rewrite of the `Prelude` core libraries. This is very true as lenses form a new mechanism for abstracting data access and nearly everything can be rewritten in this style. Often people dislike this "complete rewrite" for compatibility concerns (thus being "unidiomatic").

tel:
Huh, `evolve` is interesting but is a little tangential to lenses. It merges both an update and a focus forming a DSL for writing functions `(o -> o)`, but lenses pull this into two pieces being completely concerned with the "locations of focus" and not the transformation.

paldepind:
> Also, I'd take those articles with grains of salt. In each case, the complaint is largely against the lens library itself and not lenses as a concept.

Yes. Indeed some of the arguments are like that. But not all.

But with the old `set` and current `evolve` we get something that looks a bit more like Elms record syntax (and let me point out that Elm was created after Haskell, and lenses solves just one of Haskell several problems with records).

```elm
{ bill | name <- "Nye", age <- 27 }
```

> Huh, evolve is interesting but is a little tangential to lenses. It merges both an update and a focus forming a DSL for writing functions (o -> o), but lenses pull this into two pieces being completely concerned with the "locations of focus" and not the transformation.

You can put it like that. But in many cases `evolve` can express updates in a more concise form that better mirrors the object to transform. I also think most JavaScript programmers would find such a function more natural to work with.

This for instance would not be quite as readable with lenses:

```javascript
var fn = R.evolve({age: R.inc, name: R.toUpper, parent: {name: R.toUpper}};
fn({age: 12, name: 'Simon', parent: {name: 'Knud'}};
```

Don't get me wrong. This is certainly an improvement over our old lens API but they are not silver bullets.

tel:
Oh, to be clear, I'm not exactly arguing to promote any particular side. I'm happy talking about lens theory as I have been, but I'm hesitant to actually put a horse into the race. ;)

My concern with `evolve` is not whether or not it is more concise and readable than lenses—it may very well be—but simply that it is doing something slightly different. The Elm update syntax is also in this boat. In each case an "update" is a "location" along with a "transformation" and the two are inextricable. A lens is merely a location.

I'd be interested in seeing citations to arguments against lenses themselves. Again, not to say that I believe none exist nor to suggest that if none exist that the concept is therefore the best... but just because I honestly haven't seen very many!

tel:
As a side note, I also want to suggest that lenses should not be thought of as "solutions to Haskell's record issues". Haskell certainly has those, but lenses just solve it more or less on accident.

To me that statement comes most clear in seeing that `Lens`es and `Prism`s are essentially equally important and while `Lens`es apply directly to records, `Prism`s have almost nothing to do with them.

paldepind:
> As a side note, I also want to suggest that lenses should not be thought of as "solutions to Haskell's record issues". Haskell certainly has those, but lenses just solve it more or less on accident.

Can you elaborate on that? I've always thought fixing that was one of the primary motivations besides lenses (I'm not surprised to be wrong though, I really don't know that much on the subject).

tel:
I think the problems with records drove the development and popularity of lenses, but the concept that actually landed doesn't have too, too much to do with records. I'd even argue that lenses by themselves don't much solve the record problem (although lenses along with typeclasses get most of the way).

A lens just represents a "product-like" focus while a prism represents a "sum-like" focus. Since records are elaborate products, lenses are related to the fields of records.

Generally, however, lenslike things form an elaborate theory of "focus" and "selection" which is pretty unique all along. The nicest thing about `lens` is that it took the time to suss out much of the algebra of "focuses"—how they combine and separate and form a cogent language. This discovery (mirrored by similar "lens" discoveries in database theory) is pretty much the core of what I like about lenses generally. The `lens` library itself is just one implementation and the van Laarhoven style just one mechanism of implementation (which works very nicely in reflecting the natural subtyping hierarchy available).

buzzdecafe:
detail: `set` would fail in environments < ES5. So we can either ditch support for older environments or come up with a new name.

davidchambers:
I've added descriptions and the simplified type signatures suggested by @scott-christopher [above][1]. This pull request is now ready to be reviewed and—hopefully—merged!


[1]: https://github.com/ramda/ramda/pull/1205#issuecomment-113847633

src/lens.js
```diff
@@ -1,51 +1,31 @@
...
- * @param {Function} get A function that gets a value by property name
- * @param {Function} set A function that sets a value by property name
- * @return {Function} the returned function has `set` and `map` properties that are
- *         also curried functions.
+ * @typedef Lens s a = Functor f => (a -> f a) -> s -> f s
+ * @sig (s -> a) -> ((a, s) -> s) -> Lens s a
+ * @param {Function} getter
+ * @param {Function} setter
+ * @return {Lens}
+ * @see R.view, R.set, R.over, R.lensIndex, R.lensProp
```
davidchambers: I'm taking full advantage of our newly acquired ability to reference multiple other functions. :)

CrossEye:
I will look it over this weekend.

buzzdecafe:
i like this PR a lot. My only caveats: 

1. I don't like using the name `set`, which IMO is broader than just lens-y applications. I don't have any name much better, though. `tweak`? `alter`? 
2. Likewise, `view` might be better named with something that indicates its dependence on lens, e.g. `zoom`

davidchambers:
> I don't like using the name `set`, which IMO is broader than just lens-y applications.

Lens-y applications seem very broad to me. I struggle to think of something one could set without a lens one could not set with a lens.

> `view` might be better named with something that indicates its dependence on lens, e.g. `zoom`

Depending on one's background, `view` may in fact suggest a lens dependence, as this name is used in the Haskell [lens][1] package.

There's value in using the established names. Let's do so unless we have both a compelling reason not to and a satisfactory alternative.


[1]: https://hackage.haskell.org/package/lens

CrossEye:
:herb:

I do have one caveat.  I think we need to, sometime soon, figure out how to describe types.  We've had need for an Ordered type for sigs and we definitely need a Lens type that we can describe.  My guess is that our best bet will simply be to include another file parallel to `ReadMe.md` in the docs directory with type descriptions and figure out how to get JSDoc to link to it appropriately.

But however we do it, I think this is becoming critical.  The Lens type is too complex to just assume people are going to get it from some simple sigs.

tel:
It's fairly unlikely people will "get" lenses from types of any kind. This took the Haskell community years to internalize even with more than just types.

CrossEye:
@tel: My point, though, was that we have not had to describe types.  With the exception of the transducers details, the parameters to all of our functions, and their outputs have been easy to understand.  Although this PR include a `@typedef` tag to separate out the definition of Lens (is there already some way to render that?) we should now have some section of our documentation that gives additional details for these.

Lenses will be a big change for Ramda.  Generally there's no place where it's necessary to call one function to get the required input into another.  Here, while it's not technically required, it's damned close.  The only _raison d'être_ for the new `lens` function is to create input into other functions, and it's very unlikely to that users will create lenses except by calling `lens`.  So I think we really need decent documentation of what that `Lens` thingy is.

davidchambers:
:tada: Thank you all for your contributions to this pull request, especially @tel, for your explanations.

tel:
Excited to see it landed! And no problem—feel free to reach out any time as well.

Issue #1199, opened 2015-06-18
R.init works on strings, R.tail doesn't


asaf-romano:
This is very surprising.

http://bit.ly/1BpkpEk

davidchambers:
`R.tail` should be equivalent to `R.drop(1)`, so this is a bug as far as I'm concerned.

CrossEye:
So @davidchambers are you suggesting that `drop` be changed so that `drop(1, 'foo');`//=> ['o', 'o']` or that `tail` be changed so that `tail('foo'); //=> 'oo'`?

I agree that it seems logical that they be equivalent.  But I really don't like special-casing what should be list code to also deal with Strings, and I'm not sure we should be changing core functions like `tail` to match some implementation fall-out of a different function.  Neither of these is documented to work on Strings.  Should we spend much time worrying how they work if users pass Strings?

davidchambers:
I'm suggesting that the default implementation of `tail` should be `drop(1)`.

I find it helpful when writing—and particularly when refactoring code—to think in terms of equivalences. Functional programming increases our ability to do so (by avoiding side effects). I like knowing that two expressions are equivalent. I'm then free to use whichever expression better conveys my intent, or whichever is preferred by the maintainers of a given project.

`R.tail` and `R.drop(1)` could form such an equivalence. Unfortunately as its stands they are equivalent when applied to values of one type but *not* equivalent when applied to values of another type. I can imagine leaving comments on pull requests along the lines of _`R.drop(1)` is better written as `R.tail`_ only to have the pull request's author respond to say that she made the change but that tests failed as a result. She or I would then spend a few minutes in the REPL determining the actual behaviour of these functions, would discover the discrepancy, and possibly then raise the issue here.

One option is to provide String-specific versions of the various slice functions:

```haskell
strTail :: String -> String
strInit :: String -> String
strTake :: Number -> String -> String
strDrop :: Number -> String -> String
strTakeLast :: Number -> String -> String
strDropLast :: Number -> String -> String
```

All these functions can be derived from `slice` for both Array and String. I'd far prefer to do this than to have the String-specific functions above.

CrossEye:
> I'm suggesting that the default implementation of `tail` should be `drop(1)`.

While I could easily live with that solution, there are two issues to consider:

[...]

PR #1211, opened 2015-06-21
add R.splitEvery

davidchambers:
Supersedes #1050

This pull request differs from #1050 in several ways:

  - the implementation is significantly simpler;

  - this version of `splitEvery` throws if given a non-positive slice length; and

  - this version of `splitEvery` is defined in terms of `slice` so is defined for String as well as Array<sup>†</sup>.

---

† I'm picking up on a theme here: provide `slice`, get lots of things free.

### DIFF
--- a/src/splitEvery.js
+++ b/src/splitEvery.js
@@ -1,0 +1,31 @@
+var _curry2 = require('./internal/_curry2');
+var slice = require('./slice');
+
+
+/**
+ * Splits a collection into slices of the specified length.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig Number -> [a] -> [[a]]
+ * @sig Number -> String -> [String]
+ * @param {Number} n
+ * @param {Array} list
+ * @return {Array}
+ * @example
+ *
+ *      R.splitEvery(3, [1, 2, 3, 4, 5, 6, 7]); //=> [[1, 2, 3], [4, 5, 6], [7]]
+ *      R.splitEvery(3, 'foobarbaz'); //=> ['foo', 'bar', 'baz']
+ */
+module.exports = _curry2(function splitEvery(n, list) {
+  if (n <= 0) {
+    throw new Error('First argument to splitEvery must be a positive integer');
+  }
+  var result = [];
+  var idx = 0;
+  while (idx < list.length) {
+    result.push(slice(idx, idx += n, list));
+  }
+  return result;
+});
--- a/test/splitEvery.js
+++ b/test/splitEvery.js
@@ -1,0 +1,34 @@
+var assert = require('assert');
+
+var R = require('..');
+
+
+describe('splitEvery', function() {
+
+  it('splits a collection into slices of the specified length', function() {
+    assert.deepEqual(R.splitEvery(1, [1, 2, 3, 4]), [[1], [2], [3], [4]]);
+    assert.deepEqual(R.splitEvery(2, [1, 2, 3, 4]), [[1, 2], [3, 4]]);
+    assert.deepEqual(R.splitEvery(3, [1, 2, 3, 4]), [[1, 2, 3], [4]]);
+    assert.deepEqual(R.splitEvery(4, [1, 2, 3, 4]), [[1, 2, 3, 4]]);
+    assert.deepEqual(R.splitEvery(5, [1, 2, 3, 4]), [[1, 2, 3, 4]]);
+    assert.deepEqual(R.splitEvery(3, []), []);
+    assert.deepEqual(R.splitEvery(1, 'abcd'), ['a', 'b', 'c', 'd']);
+    assert.deepEqual(R.splitEvery(2, 'abcd'), ['ab', 'cd']);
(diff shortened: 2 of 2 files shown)


test/splitEvery.js
```diff
@@ -0,0 +1,34 @@
...
+  it('splits a collection into slices of the specified length', function() {
+    assert.deepEqual(R.splitEvery(1, [1, 2, 3, 4]), [[1], [2], [3], [4]]);
+    assert.deepEqual(R.splitEvery(2, [1, 2, 3, 4]), [[1, 2], [3, 4]]);
+    assert.deepEqual(R.splitEvery(3, [1, 2, 3, 4]), [[1, 2, 3], [4]]);
+    assert.deepEqual(R.splitEvery(4, [1, 2, 3, 4]), [[1, 2, 3, 4]]);
+    assert.deepEqual(R.splitEvery(5, [1, 2, 3, 4]), [[1, 2, 3, 4]]);
+    assert.deepEqual(R.splitEvery(3, []), []);
+  });
+
+  it('dispatches to `slice` method', function() {
```
buzzdecafe: again, i think this is a mistake

davidchambers: I like deriving functions from `slice` and `indexOf`, but perhaps the approach is better suited to Sanctuary than to Ramda.

buzzdecafe: how they are derived i don't mind. exposing how they are derived and cementing that implementation in place with tests that i find questionable at least.

CrossEye:
> † I'm picking up on a theme here: provide `slice`, get lots of things free.

Yes, but it also [makes me nervous][co].  The implementation in #1050, with a suitable use of generators, could reasonably be turned into something that works on streams or other lazy structures; it would fit well with Transducers.  This version is now limited to types that implement `slice` or that themselves define `splitEvery`.  This makes me edgy.

  [co]: https://github.com/ramda/ramda/issues/1199#issuecomment-113910880

src/splitEvery.js
```diff
@@ -0,0 +1,33 @@
...
+ * @param {Array} list
+ * @return {Array}
+ * @example
+ *
+ *      R.splitEvery(3, [1, 2, 3, 4, 5, 6, 7]); //=> [[1, 2, 3], [4, 5, 6], [7]]
+ *      R.splitEvery(3, 'foobarbaz'); //=> ['foo', 'bar', 'baz']
+ */
+module.exports = _curry2(function splitEvery(n, list) {
+  if (n <= 0) {
+    throw new Error('First argument to splitEvery must be a positive integer');
```
buzzdecafe: this is new. The only other functions where we deliberately throw are 

* `concat` because we can't figure out how to concatenate the arguments
* `constructN` because the arity `N > 10` 
* `nAry` 'First argument to nAry must be a non-negative integer no greater than ten'

I understand the rationale for throwing here; we really don't know what it means to split after every -2 elements. But i'm sure there are lots of other places in the code we are not doing such checking. i believe highland returns the whole stream/list in this situation, which may be reasonable; but may lead to difficult to trace bugs.

buzzdecafe:
> Yes, but it also makes me nervous.

me too. Feels much more "easy" than "simple" in the Hickeyan sense.

test/splitEvery.js
davidchambers: Okay. I think we should remove all the `slice`- and `indexOf`- based derivations. Shall we add String-specific versions of `splitEvery` and others?

CrossEye: The point to my mind is that unit tests are a form of contract as strong as our type annotations.  They should only document the public behavior of the system.  It's not -- or at least not always -- that we shouldn't implement things based on `slice`, but that we shouldn't commit our public API to that implementation.  It should be no one's business how we decide to do this, and we should feel free to change our minds later.

src/splitEvery.js
CrossEye: @paldepind started a discussion about this in https://github.com/ramda/ramda/pull/1050#discussion_r31990842, objecting exactly to the behavior Highland has.  I don't have a strong sense of the right thing to do.  There's a big part of me that wants to return an empty list; it may not be the most logically correct thing, but it seems least harmful.  However there are some reasonable invariants that we could use to define lawful behavior on this function; those would likely be broken by returning an empty list.  Something like

    R.equals(ls, R.reduce(R.concat, [])(R.splitEvery(n, ls)));

Although I suppose we could just say, "for any list ls, for any positive integers n", that still feels odd.

test/splitEvery.js
davidchambers: #1214

CrossEye: I think that misses the point that we -- or I at least -- was trying to make.

davidchambers:
I've updated this pull request.

CrossEye:
I can live with this.

:herb:

jethrolarson:
This would be useful to me. My main concern is whether the user would assume `splitEvery(2, [1,2,3]) === [[1,2],[3]] or [[1,2]]`. The behavior of zip makes me think the later, but I guess that'd be lossy.

PR #1212, opened 2015-06-21
add R.composeK and R.pipeK

### LINKED ISSUE #1197: function equivalent of "do" notation
This idea is being discussed in plaid/sanctuary#53. I won't duplicate the proposal, but I will open a Ramda pull request if we decide Ramda should provide this function.

@buzzdecafe is eager to *reduce* the number of composition functions, so may not want to add this one, though @joneshf has intimated that this behaviour could be provided by a more general `R.compose`.

davidchambers:
Closes #1197

### DIFF
--- a/src/composeK.js
+++ b/src/composeK.js
@@ -1,0 +1,40 @@
+var chain = require('./chain');
+var compose = require('./compose');
+var map = require('./map');
+
+
+/**
+ * Returns the right-to-left Kleisli composition of the provided functions,
+ * each of which must return a value of a type supported by [`chain`](#chain).
+ *
+ * `R.composeK(h, g, f)` is equivalent to `R.compose(R.chain(h), R.chain(g), R.chain(f))`.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @see R.pipeK
+ * @sig Chain m => ((y -> m z), (x -> m y), ..., (a -> m b)) -> (m a -> m z)
+ * @param {...Function}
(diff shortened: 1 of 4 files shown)


buzzdecafe:
LGTM. :cow2:

not sure what the deal is with the build. "npm license has expired"??? wtf?

CrossEye:
Wow, I thought @buzzdecafe might have apoplexy about yet another `compose` function.  And yet he's the first to sign off?  I have no direct objections.  I understand the desire.  But I agree with the goal @buzzdecafe has been discussing of reducing these things.  So it's only a tentative :+1:

buzzdecafe:
well, @davidchambers removed `composeL` so he is playing with house money. Also, I think we can probably unify `compose` and `composeP`. I'll do a PR for that. And then we can look at folding `K` in there as well, so we have one `compose` to rule them all.

buzzdecafe:
... now if this was derived from `slice` and there was a test exposing the implementation, then it might be a different story.

davidchambers:
> And then we can look at folding `K` in there as well, so we have one `compose` to rule them all.

It doesn't make sense to me to combine two fundamentally different composition functions. Having one function handle all these compositions (if it's even possible) is too much magic for my liking:

  - `b -> c` ∘ `a -> b`
  - `b -> c` ∘ `a -> m b`
  - `b -> m c` ∘ `a -> b`
  - `b -> m c` ∘ `a -> m b`

PR #1214, opened 2015-06-22
add R.str{Slice,Take,Drop,Init,Tail}

davidchambers:
`slice`-based derivations have met resistance, so we need to include String-specific functions if we want to officially support strings:

```haskell
R.strSlice :: Number -> Number -> String -> String
R.strTake  :: Number -> String -> String
R.strDrop  :: Number -> String -> String
R.strInit  :: String -> String
R.strTail  :: String -> String
```

I'd prefer to *specify* that the fallback definitions of `R.take`, `R.drop`, `R.init`, and `R.tail` dispatch to `slice`, which removes the need for String-specific functions. But if we don't want to formalize this, we must choose one of the following options:

  - support strings in `R.take` and friends as an *undocumented feature* which could be removed at any time;

  - add String-specific slice functions; or

  - suggest that users use `R.slice(1, Infinity)` rather than `R.tail` when dealing with strings, `R.slice(0, 3)` rather than `R.take(3)`, etc.

### DIFF
--- a/src/drop.js
+++ b/src/drop.js
@@ -1,7 +1,7 @@
 var _curry2 = require('./internal/_curry2');
 var _dispatchable = require('./internal/_dispatchable');
+var _slice = require('./internal/_slice');
 var _xdrop = require('./internal/_xdrop');
-var slice = require('./slice');
 
 
 /**
@@ -9,9 +9,6 @@ var slice = require('./slice');
  *
  * Acts as a transducer if a transformer is given in list position.
  * @see R.transduce
- *
- * Dispatches to its second argument's `slice` method if present. As a
- * result, one may replace `[a]` with `String` in the type signature.
  *
  * @func
  * @memberOf R
@@ -29,5 +26,5 @@ var slice = require('./slice');
  *      R.drop(3, 'ramda');               //=> 'da'
  */
 module.exports = _curry2(_dispatchable('drop', _xdrop, function drop(n, xs) {
-  return slice(Math.max(0, n), Infinity, xs);
+  return _slice(xs, Math.max(0, n));
 }));
--- a/src/slice.js
+++ b/src/slice.js
@@ -6,9 +6,6 @@ var _curry3 = require('./internal/_curry3');
  * Returns a list containing the elements of `xs` from `fromIndex` (inclusive)
  * to `toIndex` (exclusive).
  *
- * Dispatches to its third argument's `slice` method if present. As a
- * result, one may replace `[a]` with `String` in the type signature.
- *
  * @func
  * @memberOf R
  * @category List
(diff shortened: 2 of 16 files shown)


CrossEye:
Or option 4:

  - Document that these functions also work on strings, implement them just as they are, but do _not_ document that they are built on `slice`, leaving us free to change to other implementations when and if we choose.  

This would be my preference.  Granted these other implementations might be uglier, but if they gain us additional flexibility, the trade-off might well be worth it.

Now that ES6 has been accepted, I am certainly considering which of our functions might work on an arbitrary [Iterable][it].  These would all be good candidates, but `slice` won't cut it.  If we declare that `take` is implemented on `slice`, then, after 1.0, just changing that would have to be considered a breaking change, and a major version upgrade, to my mind.  If we only document that it works on lists and strings, then adding Iterables is not such a big deal.


  [it]: http://www.ecma-international.org/ecma-262/6.0/#sec-iteration

src/drop.js
```diff
@@ -29,5 +26,5 @@ var slice = require('./slice');
  *      R.drop(3, 'ramda');               //=> 'da'
  */
 module.exports = _curry2(_dispatchable('drop', _xdrop, function drop(n, xs) {
-  return slice(Math.max(0, n), Infinity, xs);
+  return _slice(xs, Math.max(0, n));
```
megawac: Seeing negative indexes aren't supported on `slice` the `Math.max` isn't necessary

megawac:
Actually `_slice` won't work on strings in older environments as it uses indexs to access characters instead of `charAt`

davidchambers:
> Or option 4:
>
>   - Document that these functions also work on strings, implement them just as they are, but do _not_ document that they are built on `slice`, leaving us free to change to other implementations when and if we choose.

Oh, I had no idea this option was even on the table. I've been operating under the assumption that we don't want to include any String-specific logic in any of the "list" functions. So in order to have a function like `R.tail` work with strings I thought our only option was to use the dispatch mechanism.

I'm absolutely in favour of having our slice functions officially support strings. I'm also in favour of not advertising the fact that they're built on `slice`.

I'm sorry if you've been telling me this for weeks and I've been stuck thinking that you and @buzzdecafe are both opposed to any String-specific logic in our "list" functions!

paldepind:
I prefer option 4 as well! Supporting strings in these functions makes a lot of sense. Strings are lists of characters. Having to switch to "str"-prefixed versions feels unnatural. I also very much agree that the dispatching here should be an implementation detail.

buzzdecafe:
i'm pretty sure option 4 is what i have been saying for weeks. :smile:

davidchambers:
I'm thrilled that we all like option 4!

PR #1216, opened 2015-06-22
Make pipeK mirror pipe implementation

megawac:
Makes the implementation of `pipeK` more or less identical to the impl of `pipe`


davidchambers:
:deciduous_tree: Works for me!

paldepind:
Quite nicer!

PR #1217, opened 2015-06-22
isThenable uses hasMethod

paldepind:
Just a tweak.

### DIFF
--- a/src/internal/_isThenable.js
+++ b/src/internal/_isThenable.js
@@ -1,6 +1,9 @@
+var _hasMethod = require('./_hasMethod');
+
+
 /**
  * Tests if a value is a thenable (promise).
  */
 module.exports = function _isThenable(value) {
-  return (value != null) && (value === Object(value)) && typeof value.then === 'function';
+  return _hasMethod('then', value);
 };


src/internal/_isThenable.js
```diff
@@ -1,6 +1,9 @@
+var _hasMethod = require('./_hasMethod');
+
+
 /**
  * Tests if a value is a thenable (promise).
  */
 module.exports = function _isThenable(value) {
-  return (value != null) && (value === Object(value)) && typeof value.then === 'function';
```
davidchambers: Does anyone remember why we included `value === Object(value)`? I recall someone saying this code was copied from another project.

megawac: Thats to ensure it's an `Object` and not a primitive; though it should be safe regardless to check `value.then` as long as `value != null`

davidchambers:
:deciduous_tree:

src/internal/_isThenable.js
CrossEye: > I think that check was copied from underscore?

Possibly, but it could have come from many places.  You asked about it in #573, and it's been in there since #524.  Hey, @yuri, any recollection of this?

CrossEye:
:herb:

src/internal/_isThenable.js
yuri: This wasn't copied from anywhere, but it was inspired by Q's implementation.

PR #1218, opened 2015-06-22
Make `and` and `or` sensible

paldepind:
After the last wonderful change to `equals` it appears that we are no longer fixed on implementing the flaws of JavaScript's operators in our equivalent functions.

This PR implements `and` and `or` functions with simple sensible semantics and without type coercion.

We _might_ consider throwing if non-boolean arguments are received since JavaScript programmers are somewhat used to the crazy semantics of `&&` and `||`.

I also took out dispatching. This case is actually a perfect example of how harmful dispatching is! Dispatching was originally added in #1015 since @davidchambers wanted to use it with Sanctuary. The reasoning was that it would make it possible to write code like this:

```javascript
var a = S.Just(true);
var b = S.Nothing();
R.or(a, b);  // => S.Just(true);
```

But that is the wrong way to achieve the goal. The correct way is this:

```javascript
var a = S.Just(true);
var b = S.Nothing();
R.lift(R.or)(a, b);
```

At first the difference might not look like much – but it is _huge_. The first code relies on a magic correspondence between dispatching in `R.or` and a method named `or` on `a` and `b`. It forces users into _thinking_ and _knowing_ about dispatching at a low level.

In the second code however the only thing that matters is that `a` and `b` are applicatives. We could swap out `a` and `b` with _any_ applicative and the code would still work!

In other words the first code is an example of brittle code that relies on implementation details and only pretends to be general but really isn't. The second code however is an example of actually general code where one only works on abstract types specified by well defined interfaces.

Once again dispatching deceives us away from proper abstractions.

Removing the dispatching also means that we and only us decides the semantics of `or` and `and`. Third party libraries can no longer overload it with additional meaning and complexity.

### DIFF
--- a/src/and.js
+++ b/src/and.js
@@ -1,27 +1,24 @@
 var _curry2 = require('./internal/_curry2');
-var _hasMethod = require('./internal/_hasMethod');
 
 
 /**
- * A function that returns the first argument if it's falsy otherwise the second
- * argument. Note that this is NOT short-circuited, meaning that if expressions
- * are passed they are both evaluated.
- *
- * Dispatches to the `and` method of the first argument if applicable.
+ * A function that returns `true` only if both its arguments are strictly equal
+ * to `true`.
  *
  * @func
  * @memberOf R
  * @category Logic
- * @sig * -> * -> *
- * @param {*} a any value
- * @param {*} b any other value
- * @return {*} the first argument if falsy otherwise the second argument.
+ * @sig Boolean -> Boolean -> Boolean
+ * @param {Boolean} a A boolean value
+ * @param {Boolean} b A boolean value
+ * @return {Boolean} `true` if both arguments are `true`, `false` otherwise
  * @example
  *
+ *      R.and(true, true); //=> true
+ *      R.and(true, false); //=> false
  *      R.and(false, true); //=> false
- *      R.and(0, []); //=> 0
- *      R.and(null, ''); => null
+ *      R.and(false, false); //=> false
  */
 module.exports = _curry2(function and(a, b) {
-  return _hasMethod('and', a) ? a.and(b) : a && b;
+  return a === true && b === true;
 });
--- a/src/or.js
+++ b/src/or.js
@@ -1,27 +1,24 @@
 var _curry2 = require('./internal/_curry2');
-var _hasMethod = require('./internal/_hasMethod');
 
 
 /**
- * A function that returns the first truthy of two arguments otherwise the
- * last argument. Note that this is NOT short-circuited, meaning that if
- * expressions are passed they are both evaluated.
- *
- * Dispatches to the `or` method of the first argument if applicable.
+ * A function that returns `true` if one or both of its arguments are strictly
+ * equal to `true` argument.
  *
  * @func
  * @memberOf R
  * @category Logic
- * @sig * -> * -> *
- * @param {*} a any value
- * @param {*} b any other value
- * @return {*} the first truthy argument, otherwise the last argument.
+ * @sig Boolean -> Boolean -> Boolean
+ * @param {Boolean} a A boolean value
+ * @param {Boolean} b A boolean value
+ * @return {Boolean} `true` if one or both arguments are `true`, `false` otherwise
  * @example
  *
+ *      R.or(true, true); //=> true
+ *      R.or(true, false); //=> true
  *      R.or(false, true); //=> true
- *      R.or(0, []); //=> []
- *      R.or(null, ''); => ''
+ *      R.or(false, false); //=> false
  */
 module.exports = _curry2(function or(a, b) {
-  return _hasMethod('or', a) ? a.or(b) : a || b;
+  return a === true || b === true;
 });
--- a/test/and.js
+++ b/test/and.js
@@ -4,36 +4,16 @@ var R = require('..');
 
 
 describe('and', function() {
-  it('compares two values with js &&', function() {
-    var someAr = [];
-    assert.strictEqual(R.and(1, 1), 1);
-    assert.strictEqual(R.and(1, 0), 0);
-    assert.strictEqual(R.and(true, someAr), someAr);
-  });
-
-  it('dispatches to `and` method', function() {
-    function Nothing() {}
-    Nothing.prototype.and = function() { return this; };
-
-    function Just(x) { this.value = x; }
-    Just.prototype.and = R.identity;
-
-    var n1 = new Nothing();
-    var n2 = new Nothing();
-    var j1 = new Just(1);
-    var j2 = new Just(2);
-
-    assert.strictEqual(R.and(n1, n2), n1);
-    assert.strictEqual(R.and(n2, n1), n2);
-    assert.strictEqual(R.and(n1, j2), n1);
-    assert.strictEqual(R.and(j2, n1), n1);
-    assert.strictEqual(R.and(j1, j2), j2);
-    assert.strictEqual(R.and(j2, j1), j1);
+  it('checks is both its values are true', function() {
+    assert.strictEqual(R.and(true, true), true);
+    assert.strictEqual(R.and(true, false), false);
+    assert.strictEqual(R.and(false, true), false);
+    assert.strictEqual(R.and(false, false), false);
   });
 
   it('is curried', function() {
     var halfTruth = R.and(true);
     assert.strictEqual(halfTruth(false), false);
-    assert.strictEqual(halfTruth('lie'), 'lie');
+    assert.strictEqual(halfTruth(true), true);
   });
 });
--- a/test/or.js
+++ b/test/or.js
@@ -5,36 +5,14 @@ var R = require('..');
 
 describe('or', function() {
   it('compares two values with js &&', function() {
-    var someAr = [];
-    assert.strictEqual(R.or(1, 0), 1);
-    assert.strictEqual(R.or(0, 1), 1);
-    assert.strictEqual(R.or(someAr, false), someAr);
-    assert.strictEqual(R.or('', 0), 0);
-  });
-
-  it('dispatches to `or` method', function() {
-    function Nothing() {}
-    Nothing.prototype.or = R.identity;
-
-    function Just(x) { this.value = x; }
-    Just.prototype.or = function() { return this; };
-
-    var n1 = new Nothing();
-    var n2 = new Nothing();
-    var j1 = new Just(1);
-    var j2 = new Just(2);
-
-    assert.strictEqual(R.or(n1, n2), n2);
-    assert.strictEqual(R.or(n2, n1), n1);
-    assert.strictEqual(R.or(n1, j2), j2);
-    assert.strictEqual(R.or(j2, n1), j2);
-    assert.strictEqual(R.or(j1, j2), j1);
-    assert.strictEqual(R.or(j2, j1), j2);
+    assert.strictEqual(R.or(true, true), true);
+    assert.strictEqual(R.or(true, false), true);
+    assert.strictEqual(R.or(false, true), true);
+    assert.strictEqual(R.or(false, false), false);
   });
 
   it('is curried', function() {
-    assert.strictEqual(R.or('lie')(false), 'lie');
+    assert.strictEqual(R.or(false)(false), false);
     assert.strictEqual(R.or(false)(true), true);
-    assert.strictEqual(R.or('')(0), 0);
   });
 });


buzzdecafe:
:cow2:

davidchambers:
> This PR implements `and` and `or` functions with simple sensible semantics and without type coercion.
>
> We _might_ consider throwing if non-boolean arguments are received since JavaScript programmers are somewhat used to the crazy semantics of `&&` and `||`.

If we change the meanings of these functions in this way we *must* throw if given a non-Boolean argument. `R.or(xs.length, 0)` evaluating false is almost certainly not the result the programmer expected, yet we will silently accept this invalid input. Essentially we're coercing everything except `true` to `false`. This is arguably just as crazy (or perhaps more so) than the existing semantics.

I'm in favour of making these Boolean-only functions. We should throw a TypeError if given one or more arguments of another type.

> I also took out dispatching. This case is actually a perfect example of how harmful dispatching is! Dispatching was originally added in #1015 since @davidchambers wanted to use it with Sanctuary. The reasoning was that it would make it possible to write code like this:
>
> ```javascript
> var a = S.Just(true);
> var b = S.Nothing();
> R.or(a, b);  // => S.Just(true);
> ```
>
> But that is the wrong way to achieve the goal. The correct way is this:
>
> ```javascript
> var a = S.Just(true);
> var b = S.Nothing();
> R.lift(R.or)(a, b);
> ```

You've misunderstood what I hoped to accomplish with #1015.

Some types have an empty or failure value (or multiple such values), and a non-empty or success value (or multiple such values):

| Type         | Empty/failure values          | Non-empty/success values      |
| ------------ | ----------------------------- | ----------------------------- |
| `Array`      | `[]`                          | every member other than `[]`  |
| `Boolean`    | `false`, `new Boolean(false)` | `true`, `new Boolean(true)`   |
| `Either a b` | `Left(a)` for every `a`       | `Right(b)` for every `b`      |
| `Maybe a`    | `Nothing()`                   | `Just(a)` for every `a`       |

Now, imagine each of these types provided a `toBoolean` method to indicate the subset (i.e. the column) in which a value resides. `[].toBoolean()` would evaluate false, as would `Nothing().toBoolean()` and `Left('Invalid JSON').toBoolean()`. `[1, 2, 3].toBoolean()` would evaluate true, as would `Just(false).toBoolean()` and `Right(false).toBoolean()`.

Given `toBoolean`, we can define `and` and `or` for all such types. If the type also provides an `empty` method (meaning that it has exactly one empty value), we can also define `xor`. So from just two methods we can derive the following:

  - `and :: [a] -> [a] -> [a]`
  - `and :: Boolean -> Boolean -> Boolean`
  - `and :: Either a b -> Either a b -> Either a b`
  - `and :: Maybe a -> Maybe a -> Maybe a`
  - `or  :: [a] -> [a] -> [a]`
  - `or  :: Boolean -> Boolean -> Boolean`
  - `or  :: Either a b -> Either a b -> Either a b`
  - `or  :: Maybe a -> Maybe a -> Maybe a`
  - `xor :: [a] -> [a] -> [a]`
  - `xor :: Boolean -> Boolean -> Boolean`
  - `xor :: Maybe a -> Maybe a -> Maybe a`

These signatures are nothing like what `R.lift(R.or)` gives us. ;)

I'm fine with `R.and` and `R.or` no longer dispatching; [`S.and`][1] and [`S.or`][2] provide the functionality I need. :)


[1]: https://github.com/plaid/sanctuary#and
[2]: https://github.com/plaid/sanctuary#or

megawac:
>> This PR implements and and or functions with simple sensible semantics and without type coercion.
We might consider throwing if non-boolean arguments are received since JavaScript programmers are somewhat used to the crazy semantics of && and ||.

>If we change the meanings of these functions in this way we must throw if given a non-Boolean argument. R.or(xs.length, 0) evaluating false is almost certainly not the result the programmer expected, yet we will silently accept this invalid input. Essentially we're coercing everything except true to false. This is arguably just as crazy (or perhaps more so) than the existing semantics.

>I'm in favour of making these Boolean-only functions. We should throw a TypeError if given one or more arguments of another type.

I 100% agree, you should a) leave the implentation as is and expect users to know a little a bit about how JS `&&` and `||` work (who is learning ramda before JS anyway) or b) throw an error for non booleans. I'm more in favour of a) as the JS semantics are pretty powerful but I can see how the signature isn't a great fit for ramda `* * -> *`

davidchambers:
> I can see how the signature isn't a great fit for ramda `* * -> *`

True, although if one writes it as `a -> a -> a` it seems reasonable. We wouldn't be enforcing that the arguments are of the same type, but we could at least suggest this sensible usage in our type signatures.

paldepind:
> Essentially we're coercing everything except true to false. This is arguably just as crazy (or perhaps more so) than the existing semantics.

I don't think that description is fair as this only happens for input that goes against the type declaration. Silently ignoring wrong input is common among Ramda functions.

> You've misunderstood what I hoped to accomplish with #1015.

Sorry. It mostly looked at the code you posted. It looks like you've creating an interface for a class of values that can answer the question "do you contain actual content? True or false?".

> These signatures are nothing like what R.lift(R.or) gives us. ;)

Are they not? The types you mention except for boolean are applicatives. And a lifted `and` or a lifted `or` works on all applicatives. So the signature actually looks quite the same to me. But what `S.and` is _doing_ is certainly something else. It doesn't answer "does both arguments contain `true`" it asks "does both arguments contain anything". And not overloading those semantics into `R.and` seems like a good thing to me.

> I'm more in favour of a) as the JS semantics are pretty powerful but I can see how the signature isn't a great fit for ramda * * -> *

I would say they are pretty awful, complex and error prone. To me Ramda is about creating a better way to write JavaScript. We cannot offer our users that if we insist on repeating JavaScripts mistakes.

davidchambers:
> > Essentially we're coercing everything except `true` to `false`. This is arguably just as crazy (or perhaps more so) than the existing semantics.
>
> I don't think that description is fair as this only happens for input that goes against the type declaration. Silently ignoring wrong input is common among Ramda functions.

I'm confused. If we're happy to ignore invalid input why is it necessary to switch from `a && b` to `a === true && b === true`?

CrossEye:
> I 100% agree, you should a) leave the implentation as is and expect users to know a little a bit about how JS `&&` and `||` work (who is learning ramda before JS anyway) or b) throw an error for non booleans. I'm more in favour of a) as the JS semantics are pretty powerful but I can see how the signature isn't a great fit for ramda `* * -> *` 

I agree.  I think the language already does for its known types something like David's table, and I'm perfectly happy with the semantics of `&&` and `||` in Javascript.  I would rather keep alive that part of our silly little mantra: "A practical functional library **for Javascript programmers**." 

As to this, though, "(who is learning ramda before JS anyway)", I'm getting the impression that some FP people are starting in Javascript with Ramda, because it's at least somewhat familiar.

davidchambers:
> It doesn't answer "does both arguments contain true" it asks "does both arguments contain anything".

This isn't correct either. The `and` and `or` functions I've described don't return a Boolean in response to a yes or no question: they return one of their arguments. They're closely related to JavaScript's `&&` and `||` operators, which evaluate to one of their operands.

[`S.and`][1]:

> __`and :: a -> a -> a`__
>
> Takes two values of the same type and returns the second value if the first is "true"; the first value otherwise.

It goes on to define "true".

I should add some more examples for these functions. I can see now that their purpose is unclear.

> [N]ot overloading those semantics into `R.and` seems like a good thing to me.

The semantics I have in mind are compatible. `&&` and `||` are for choosing one of two values. The problem is that when we define our own types these operators don't work as we would like, since the language considers every value of our type "true" (since they are objects).

I'm okay with Ramda not becoming the dispatches-to-everything library. Such a library could be built on top of Ramda.


[1]: https://github.com/plaid/sanctuary#and

CrossEye:
> I'm confused. If we're happy to ignore invalid input why is it necessary to switch from `a && b` to `a === true && b === true`?

The only thing I'm possibly interested in changing there is the return:

    return !! (a && b);

documented with 

    Bool* -> Bool* -> Bool

where, if we ever get around to documenting types, we can describe Bool* as the type of things that can be converted to Booleans (pretty much everything), along with the conversion rules.

But I don't have a strong feeling about this.

paldepind:
> I'm confused. If we're happy to ignore invalid input why is it necessary to switch from a && b to a === true && b === true?

This PR is also about removing dispatching, fixing the type declaration and removing the examples that suggests passing anything but boolean values to `and` and `or`. If we throw on non-boolean values just using `||` and `&&` is not an issue.

> I agree. I think the language already does for its known types something like David's table, and I'm perfectly happy with the semantics of && and || in Javascript. I would rather keep alive that part of our silly little mantra: "A practical functional library for Javascript programmers."

To me the semantics of `&&` and `||` are plain crazy. Fully explaining them is quite complicated. The semantics of `and` and `or` in this PR are as simple as it gets and they can completely be described by a simple truth table. To me this doesn't make Ramda any less practical or any less for JavaScript programmer (and remember that no one is prohibiting anyone from using `||` or `&&`). This is really all about making Ramda simpler.

> This isn't correct either. The and and or functions I've described don't return a Boolean in response to a yes or no question: they return one of their arguments. They're closely related to JavaScript's && and || operators, which evaluate to one of their operands.

Yes. You are right. I forgot about that aspect (probably because it is very different from how I think about `&&` and `||`). But still the signatures are the same as a lifted `and` or `or`.

> && and || are for choosing one of two values.

Do you really think so? To me they are logic operators and using them for anything else is misuse and potentially error prone (error prone as in `number = suppliedNumber || defaultNumber` will fail if the supplied number is 0).

davidchambers:
> But still the signatures are the same as a lifted `and` or `or` (not the suggested `and` and `or` though, they only work on booleans).

The signatures are the same, yes:

```javascript
S.and :: Maybe a -> Maybe a -> Maybe a
R.lift((a, b) => a && b) :: Maybe a -> Maybe a -> Maybe a
```

But the semantics are different:

```javascript
> S.and(S.Just(0), S.Just(1))
Just(1)
> R.lift((a, b) => a && b)(S.Just(0), S.Just(1))
Just(0)
```

> > `&&` and `||` are for choosing one of two values.
>
> Do you really think so?

Yes. Even with Boolean operands the operator selects one or other operand (as `new Boolean(true) && new Boolean(true)` demonstrates).

I've seen code like this in JavaScript many times:

```javascript
(groups[key] || (groups[key] = [])).push(value);
```

I very much consider this an abuse of `||`.

Haskell, I believe, has an Alternative typeclass which captures the idea I had in mind when I opened #1015. So the *idea* of `&&` and `||` choosing one of their operands is sound, I believe, I just don't like the JavaScript semantics and our inability to define how they operate on values of user-defined types.

paldepind:
> The signatures are the same, yes: [...] But the semantics are different:

Yes. Exactly what I've been saying ever since you wrote "These signatures are nothing like what R.lift(R.or) gives us. ;)"

> Yes. Even with Boolean operands the operator selects one or other operand (as new Boolean(true) && new Boolean(true) demonstrates).

You can view their behavior as such. But when you take away the notion that things besides booleans can be true or false it doesn't make much sense. `&&` and `||` are not for choosing one of two values – they are for implementing logic. Most languages that can't chase their boolean operators back to C would probably agree. Math, Haskell and SQL for instance would.

> Haskell, I believe, has an Alternative typeclass which captures the idea I had in mind when I opened #1015. So the idea of && and || choosing one of their operands is sound, I believe,

They might be sound _if_ you call them something besides `and` and `or`.

davidchambers:
> They might be sound *if* you call them something besides `and` and `or`.

Haskell has `(<|>)` for choosing one of two Alternatives. `R['(<|>)']`, anyone? ;)

CrossEye:
A big part of the reason for changing these name from their previous incarnation as binary functions on functions with similar APIs, was in order to make them [more like the JS operators][jl].  I've seen nothing that makes me want to switch that now.

-----

> I've seen code like this in JavaScript many times:
> 
> ```javascript
> (groups[key] || (groups[key] = [])).push(value);
> ```
> 
> I very much consider this an abuse of `||`.

If anything, that's an abuse of assignment, but I don't think I buy either.  There is a natural progression from one to another, and I don't see any cleat-cut dividing lines:

```js
groups[key].push(value);  // warning, possible failure!
if (groups[key]) {groups[key].push(value);}
if (!groups[key])) {groups[key] = []; groups[key].push(value);}
groups[key] = groups[key] || []; groups[key].push(value);
(groups[key] || (groups[key] = [])).push(value);
```

  [jl]: https://github.com/ramda/ramda/issues/843#issue-58981331

paldepind:
> A big part of the reason for changing these name from their previous incarnation as binary functions on functions with similar APIs, was in order to make them more like the JS operators. I've seen nothing that makes me want to switch that now.

Current proposal keeps all the good parts of the JS operators.

CrossEye:
> Current proposal keeps all the good parts of the JS operators.

Well, I'm afraid we have different opinions on the good parts of the operators.  :smile:

I think this is very much a good part of `||`:

```js
 var opts = mixin(defaults, options || {});
```

paldepind:
@CrossEye 

> I think this is very much a good part of `||`:

If you think of `||` as a logic operator then that makes no sense. And you would never use `R.or` for that anyway since Ramda has `R.defaultTo` which communicates the intend much better. But even if one likes to write code like that no one it taking plain `||` away.

Ramda is a functional library and one of the most important parts of functional programming is clear and precise semantics that are easy to understand and reason about. It is my impression that Ramda aims for simplicity. And the `and` and `or` I'm suggesting are clearly much simpler than `||` and `&&`. You can't claim otherwise.

As far as I can tell both @buzzdecafe and @davidchambers is in favor of simpler proper logic functions. Unless I'm misunderstanding something David makes an argument for the same case [here](https://github.com/ramda/ramda/issues/843#issuecomment-76621221). You are arguing _for complexity_ with hand waving. I need to see examples and reasons for why keeping this complexity is a good thing compared to having precise semantics that prevents misuse.

CrossEye:
> But even if one likes to write code like that no one it taking plain `||` away.

In some sense, that's all these functions should be, to my mind, a simple wrapper around the language operator, something like what David did in [nucleotides][nu].  When David asked that these also dispatch, I had no objection, although I felt no great need.

Clearly when trying to program with composeable functions, making functions out of syntactic constructs is fairly important.  If we didn't want this, we could live without `add` and `eq` and a number of others, since users could always pass lamdas for those.  I find these functions quite useful.  To my mind, the _only_ reason for giving up these useful names from their previous place as `and(f, g)(...params); //=>Bool` was to make way for these operator functions.  I really don't feel like going backward here.


> To me the semantics of `&&` and `||` are plain crazy. Fully explaining them is quite complicated.

Fully explaining `&&` in terms of truthy and falsey, even given multiple uses (as they are associative) is to my mind, just, "returns the first falsey value or the last value."  For `||` it's "returns the first truthy value or the last value".  It doesn't sound that complex.

I am especially bothered by the `a === true` / `b === true` blocks.

You say, 

> Silently ignoring wrong input is common among Ramda functions.

Are you equally happy with `a !== false` / `b !== false`?  Would you be happy if we changed from one to the other between versions without mentioning it?  As a user, I might be very upset by that.  "Sure it wasn't documented as working with anything but booleans, but it's Javascript, damn it!  When I used it as `R.and(user.firstName, user.lastName)`, it was just working.  Now suddently it's broken. WTF?"  While this is of course an argument against arbitrarily changing things on our users, it's also an argument against magic in our APIs.  If, on the other hand, you wouldn't be equally happy with `a !== false` / `b !== false`, then _why_?




  [nu]: https://github.com/davidchambers/nucleotides

paldepind:
> In some sense, that's all these functions should be, to my mind, a simple wrapper around the language operator, something like what David did in nucleotides. When David asked that these also dispatch, I had no objection, although I felt no great need.

Then you must be very bothered by `equals`. That is a much bigger deviation from `==`/`===` than anything in this PR.

> Clearly when trying to program with composeable functions, making functions out of syntactic constructs is fairly important. If we didn't want this, we could live without add and eq and a number of others, since users could always pass lamdas for those. I find these functions quite useful. To my mind, the only reason for giving up these useful names from their previous place as and(f, g)(...params); //=>Bool was to make way for these operator functions. I really don't feel like going backward here.

I don't know where you're coming from with this argument. I'm not suggesting we remove `and` and `or` and I'm not questioning the benefits of functions compared to operators.

> Fully explaining && in terms of truthy and falsey, even given multiple uses (as they are associative) is to my mind, just, "returns the first falsey value or the last value." For || it's "returns the first truthy value or the last value". It doesn't sound that complex. Explaining the full effect of coercion in the language is another issue, even if it's just for the fairly simple case of Booleans.

You're hiding all the complexity in the

CrossEye:
> Then you must be very bothered by `equals`. That is a much bigger deviation from `==/===` than anything in this PR.

Less so since we renamed it from `eq`, which felt much more like an operator.  And besides, "a foolish consistency is the [hobgoblin of little minds][re]."

> You're hiding all the complexity and all the ugliness inside the words "truthy" and "falsey". But it is still there, it is still massive and still full of ridiculous semantics.

I did say,

>> Explaining the full effect of coercion in the language is another issue, even if it's just for the fairly simple case of Booleans.

But here it is, if you want it, I don't think it's all that complex:  Javascript operators will coerce values of some types to other types in certain operations.  In `||` and `&&`, for testing, values may be coerced into _truthy_ or _falsey_ values according to this rule: `false`, `+0`, `-0`, `""` (the empty string), `NaN`, `null`, and `undefined` are _falsey_.  Everything else is _truthy_.

>> Are you equally happy with `a !== false` / `b !== false`? 

> I don't care between the two because they give the same results when used properly.

Then would you also be equally happy with `return !!(a || b)` / `return !!(a && b)` ?  This would remove what I think is actually a very good part of `||` and `&&`, the returning of one of the values rather than just a boolean, but it would also preserve the boolean semantics of the language for users who use `||` and `&&`  only for booleans, which works exactly as yours does, if "used properly."

> But it is a fine argument for throwing.

I'm really not happy with this option.  I think there are times to throw in Ramda, but those are chiefly when there is simply no way to provide some reasonable default behavior given the users data.  Other than that, I think Ramda should avoid throwing, as exception handling is very hard to compose with other functional constructs.

> Let me point out that you still haven't demonstrated why the additional complexity you want to have in and and or is desirable. As I wrote:

>> I need to see examples and reasons for why keeping this complexity is a good thing compared to having precise semantics that prevents misuse.

We're obviously coming from different ideas here.  You're assuming I agree with you on the goals of these functions.  I'm not.  I've stated what I want from them: functional equivalents of the language operators.  I don't see these operators as overly complex.  In fact I find them quite useful and quite easy to understand.  I have no problem at all understanding the fact that if not given booleans, they might not return one.  But I also know that these operators (and the current Ramda functions) work perfectly well as boolean operations.  If you pass them boolean arguments, they will return the correct boolean result for the conjunction / disjunction.  So these operations _already_ do everything you want, if "used properly".  Why do you want to spoil my fun?

And finally, we need to beware of arguing too hard about simplicity when we're discussing _what_ to do.  It's very, very important when we're deciding _how_ to accomplish something, but it has to play a much smaller role in the _what_.  The reason is clear: even if your suggestion is simpler than mine, I can offer a simpler one still: remove the functions altogether.  In fact, let's remove all the functions and the entire library.  That's the ultimate end of a process whose only goal is simplicity.  I don't suggest that simplicity has no role to play in those discussions; it's still important, but it has to be balanced against other criteria.


  [re]: http://www.brainyquote.com/quotes/quotes/r/ralphwaldo136909.html

paldepind:
@CrossEye 

Let me just repeat myself:

> You are arguing for complexity with hand waving. I need to see _examples_ and reasons for why keeping this complexity is a good thing compared to having precise semantics that prevents misuse.

And saying "the complexity I want is not hurting you – you can just ignore it" is never a good argument for adding or keeping complexity.

CrossEye:
@paldepind:

> Let me just repeat myself:

I'm afraid that's what we're doing here.  But I'll try one more time.

I am looking for a function which reifies the `||` operator of the language.  That's the specification I'm working under.  I find that operator useful and, just as we've done with `+`, with `%`, and with a number of other operators, I want us to make this one available for use in functional compositions.

When this function was created it was named `or`, which makes perfect sense, since that is how people pronounce "`||`".  And more than that, it precisely handles boolean logic "when used properly."  If you pass booleans, it functions exactly as would be expected from the boolean truth tables.

In this thread, you've already talked about the fact that silently ignoring wrong input is common among Ramda functions.  But this function already operates exactly as you would like when given what you claim is the only proper input.  So the only reason I can see that you want to change it is that you don't like how other people might decide to use it.  Is that it?

I see no extra complexity.  The implementation that's in place is pretty well the bare minimum necessary to meet the goal of creating a functional version of the operator `||`.  And, once again, if you're arguing that this goal is more complex than the goal of a plain boolean operator, you're certainly correct.  But having no operator at all is simpler still, and how do you argue against that?

paldepind:
@CrossEye 

> I'm afraid that's what we're doing here.

That is because you're not answering my question. I want _examples_. Use cases! Not hand waving.

>  I find that operator useful

You keep saying that. Then why can't you show examples of valid use cases of the extra complexity that you want to keep in `and` and `or`?

> So the only reason I can see that you want to change it is that you don't like how other people might decide to use it. Is that it?

Yes. That is why I call it misuse. But that is not the _only_ reason. Simplicity is as well.

> I see no extra complexity. [...] And, once again, if you're arguing that this goal is more complex than the goal of a plain boolean operator, you're certainly correct.

You start by saying that you don't see the complexity but then you say that I'm right about your goal being more complex? Semantically the `and` and `or` I'm suggesting is _much_ simpler than the `and` and `or` that we currently have. That is undeniable.

> But having no operator at all is simpler still, and how do you argue against that?

In the same way that all complexity is added to Ramda. Valid use cases must be demonstrated. Demonstrating valid use cases of correct simple logic functions as found in math is easy. Demonstrating a use case for additional complexity of matching the semantics of the monster logic operators that JavaScript inherited from Java that inherited them from C is pretty hard.

And just to make sure: I'm not requesting use cases for the operators themselves but for why our logic functions should implement their full complexity.

CrossEye:
Have you shown any use cases for a boolean operator?

No?  Why not?

CrossEye:
@paldepind:

Use cases abound.  I can't believe you have a hard time coming up with them.

```js
var person = {name: 'Fred', address: ''}
R.or(person.address, 'Unknown')
```

```js
var grade = function(score) {
  return (score < 65) ? 'F' : (score < 70) ? 'D' : (score < 80) ? 'C' : (score < 90) ? 'B' : 'A'
};

var students = [
   {name: 'fred', score: 78}, 
   {name: 'wilma', score: 'incomplete'}, 
   {name: 'betty', score: 96}
];

R.map(student => R.assoc('grade', R.or(parseInt(student.score), 0), student), students);
```

CrossEye:
I don't see _extra_ complexity.  I see some _inherent_ complexity, to be sure.  The question is whether that is justified.  I haven't seen your counterarguments.

paldepind:
@CrossEye 

Splendid examples. Exactly the kind of horrible code that `and` and `or` should not allow.

The first one could be written much more clearly like this:

```javascript
var person = {name: 'Fred', address: ''};
R.ifElse(R.isEmpty, R.identity, R.always('Unknown'));
```

Besides being easier to understand the rewrite is also easier to extend. You could also easily add a trim if necessary:

```javascript
var person = {name: 'Fred', address: ''};
R.ifElse(R.compose(R.isEmpty, R.trim), R.identity, R.always('Unknown'));
```

The root of the problem in the second code is that your objects has a field that can be of several different types. And you end up calling `parseInt` on numbers which is unnecessary. I would create a separate normalization function that you could pass the scores through with `assoc`.

If you feel like writing code like that then go ahead. But then I suggest that you use `||` instead and reserve `or` for sane cases.

paldepind:
Oh, and if the default grade becomes 10 tomorrow and only the `0` is changed a bug would be introduced. Such is the dangers when misusing logic operators.

CrossEye:
@paldepind:

So where are your use cases for a boolean operator?

CrossEye:
So to your mind this expression

```js
R.or(person.address, 'Unknown')
```

is more simply represented as

```js
R.ifElse(R.isEmpty, R.identity, R.always('Unknown'))(person.address)
```

?  No wonder we're having trouble agreeing!

paldepind:
@CrossEye I most certainly find that much cleaner than using logic operators as control structures.

buzzdecafe:
`ory` and `andy` anyone? :stuck_out_tongue_winking_eye:

CrossEye:
The question is, though, is it _simpler_?

CrossEye:
@paldepind:

> I most certainly find that much cleaner than using logic operators as control structures.

You're presupposing that these are 'logic operators'.  That's the whole point in contention.

paldepind:
> You're presupposing that these are 'logic operators'. That's the whole point in contention.

Am I _presupposing_(!) that `||` and `&&` are logic operators? No offense, but that is the most ridiculous claim I have heard in a long time. These operators are inherited from C and they have been called logic operators for _more than 40 years_. Just because you fancy misusing their semantics in your code doesn't make them any less logic operators.

I really thought that being for JavaScript programmers meant embracing the good parts of JavaScript and actually being _for_ JavaScript programmers. I guess I was wrong.

CrossEye:
What you're presupposing is clearly that they are operators that operate only on booleans.

The symbols have been inherited from C, if not before.  The semantics are Javascript's own.  Just because you fancy changing their semantics in your code does not make them any less Javascript operators.

I already told you that I believed this _is_ one of the good parts of Javascript.  You don't seem to be willing to debate at that level, preferring to assume that there is some settled policy that boolean-only is the preferred syntax for the translation of the `||`-opertator to a function.  I don't accept your presupposition.

And have you given your use-cases for a boolean operator?

paldepind:
> You don't seem to be willing to debate at that level, preferring to assume that there is some settled policy that boolean-only is the preferred syntax for the translation of the ||-opertator to a function.

That is not something I assume. I've provided many reasons.

CrossEye:
>>  [You] assume that there is some settled policy that boolean-only is the preferred syntax for the translation of the ||-opertator to a function.

> That is not something I assume. I've provided many reasons.

Not that I've seen.  All I've seen are circular arguments that say that one wouldn't want the behavior of the Javascript `||` because its more complex than a plain boolean operator.  That argument can be applied to pretty much any abstraction.

So, do you have use cases for a plain boolean operator?

davidchambers:
[Nucleotides][1] was an interesting experiment. I see no reason for Ramda to take this approach, though. In the past I've suggested replacing `R.mathMod` and `R.modulo` with `R.mod` and `R.rem`. If an operator does not behave the way we would like we should be free to make Ramda's "equivalent" behave differently. `R.identical` is a good example: `R.identical(a, b)` is not equivalent to `a === b`.

Were this discussion in the context of Sanctuary, I'd have no problem getting behind this definition:

```javascript
def('and', [Boolean, Boolean], function(x, y) { return x && y; });
```

Throwing given a non-Boolean argument is not in keeping with the approach Ramda generally takes, which is to avoid throwing unless we see no other option.

@paldepind, it seems to me you're looking to make Ramda more principled than it is already, in which case you could consider diverting some of your energies to Sanctuary. Sanctuary embraces correctness in a way that Ramda cannot, given Ramda's goal of remaining approachable to JavaScript programmers with no FP experience. This is a worthwhile goal, as we can't expect such programmers to jump directly to PureScript. This is a smoother upgrade path:

Vanilla JavaScript → Underscore/lodash → Ramda → Sanctuary → PureScript


[1]: https://github.com/davidchambers/nucleotides

kedashoe:
Some related but scattered thoughts:

1. I don't mind the idea of `||` and `&&`, but it bothers me that the empty string is false while empty lists and dicts are true.
2. I often use something like what `||` does in non-statically typed languages, but I never use `&&`
3a. I wouldn't mind seeing `and` and `or` work with booleans and throwing if they are given something
3b. changing `isEmpty` to the underscore definition (https://github.com/ramda/ramda/pull/539, except updated to current underscore) (@CrossEye, how can you argue so strongly for `||` js behavior but against that PR for isEmpty??)
3c. defining `unEmpty`, which takes a list of values and returns the first non-empty one, or the last value if all are empty

CrossEye:
@davidchambers: Well said.  And thank you for stepping in.

I'm not sure I'm ready to make the step toward PureScript, but I love the principled stand that Sanctuary is taking, and I am really glad to have both Ramda and Sanctuary available.

CrossEye:
@kedashoe:

>  I don't mind the idea of `||` and `&&`, but it bothers me that the empty string is false while empty lists and dicts are true.

I agree.  As a matter of language design, this seems an absurd choice.  But when duplicating the `||` and `&&` semantics, I wouldn't feel right trying to correct that.

> @CrossEye how for || js behavior but against PR for isEmpty?

I'm afraid I don't understand why you see that as a contradiction.  `isEmpty` doesn't have anything to do with built-in operators.  Is it just because #539 was about making one function accept a list, a String, or an object, and I objected to that?  This is quite different, to my mind, where we accept anything and use the language's ability to cast it to a truthy/falsey value.

kedashoe:
> I'm afraid I don't understand why you see that as a contradiction. isEmpty doesn't have anything to do with built-in operators.

Makes sense when viewed through the operator prism. I guess I thought of `isEmpty` as a better falsey, but now that I look at it, it is something else.

paldepind:
Superseded by #1335.

PR #1223, opened 2015-06-23
curry function returned from compose and pipe

buzzdecafe:


src/internal/_compose.js
```diff
@@ -21,7 +24,7 @@
  *      squareThenDouble(5); //≅ double(square(5)) => 50
  */
 module.exports = function _compose(f, g) {
-  return function() {
+  return curryN(g.length, function() {
```
megawac: Are you sure you want to do it at this level rather than on the `compose.js` level. This will be significantly slower as you'll be wrapping it at each step of the way in a curry

megawac:
I'm -1 for this change. I think we discussed this last July and and reached the ~agreement that user can accomplish this themselves pretty easy and may not always want a curried function from compose. 

The downside is composing a curried function currently doesn't work too well (something `metadata` can solve)!

src/internal/_compose.js
buzzdecafe: i see what you mean. i'll update the pr.

buzzdecafe:
> user can accomplish this themselves pretty easy and may not always want a curried function from compose. 

i think it's time for another go-round. i also think we should curry any function returned from `curry`

> The downside is composing a curried function currently doesn't work too well (something metadata can solve)!

anxious to see that

megawac:
> i also think we should curry any function returned from curry

100% agree, just don't think all composed functions need to be curried

src/internal/_compose.js
```diff
@@ -1,3 +1,6 @@
+var curryN = require('../curryN');
```
davidchambers: This import is not used.

buzzdecafe: _mea culpa_

test/compose.js
```diff
@@ -65,4 +65,14 @@ describe('compose', function() {
     assert.deepEqual(g(1, 2, 3), [1, 2, 3]);
   });
 
+  it('returns a curried function', function() {
+    function f(a) { return a + '_END'; }
+    function g(b, c, d) { return [b, c, d].join('_'); }
+    var c = R.compose(f, g);
```
davidchambers: It's typical to use `g ∘ f` in examples. Let's update the function names so this can read `R.compose(g, f)`.

buzzdecafe: i disagree. composition is almost always exemplified by _f &#9675; g_

megawac: Agreed

davidchambers: Interesting. I'd thought the alphabetical ordering of `f` and `g` was suggestive of application order, as in this [diagram][1]:

g ∘ f

It sounds strange to me to say _composition of `g` and `f`_, in the same way that it sounds strange to me to say _dad and mum_ rather than _mum and dad_. There's no logical reason for this, of course. :)


[1]: https://en.wikipedia.org/wiki/Function_composition#/media/File:Compfun.svg

buzzdecafe: "What's in a name? That which we call a rose by any other name would smell as sweet."

buzzdecafe:
it's worth observing that `useWith` and `converge` *already curry their output functions*. so this is partly a question of consistency.

CrossEye:
> The downside is composing a curried function currently doesn't work too well  - i.e. it doesn't stay curried so functions down the pipe get a function instead of data.

I don't follow this.  Do you have an example of what you mean?

megawac:
@CrossEye, note that `add` is curried

```js
R.compose(R.divide(R.__, 2), R.add)(5, 1)
R.compose(R.divide(R.__, 2), R.add)(5)(1)
```

CrossEye:
I would be :+1: on this change if we also added another version of `compose` / `pipe` -- sorry! -- under some less obvious names that give the users some way to opt out of the auto-currying here.  It's the same issue I've mentioned before: that of functions with named optional parameters.  If one of those is returned and then automatically curried, we know what can happen.

CrossEye:
I guess I don't understand this.

> ```js
>    R.compose(R.divide(R.__, 2), R.add)(5, 1)
>    R.compose(R.divide(R.__, 2), R.add)(5)(1)
> ```

The first one works now.  The second one would work with this PR.  The goal would be that Ramda's own functions _and functions returned by Ramda_ are all curried.  (I have a caveat in my previous message, but I like the general idea.)

It sounds as though this PR would _fix_ the issue you describe.  So I'm not quite sure what problem you're seeing.

megawac:
@CrossEye I'm suggesting with metadata we could leave it as is and be able to detect curried functions (read the rest of my original comment)

CrossEye:
I still haven't spent much time with your metadata implementation, partly because the basic ideas have inspired me to try my own, and I didn't want any cross-contamination while I developed it.  (I still haven't gone very far, and at this point with vacation coming up next week, I probably won't finish it soon.)  But does a composition store a reference to the first function to run, is that how you would track this?  Because obviously the composed function is a _new_ function, and the fact that `add` is curried doesn't mean squat about the new composed function unless you provide some sort of linkage.  Is that how this would work?

test/compose.js
davidchambers: It would be nice to update the identifiers in one or other example to emphasize the equivalence between `R.compose(f, g)` and `R.pipe(g, f)`.

davidchambers:
I like this change.

paldepind:
> I've mentioned before: that of functions with named optional parameters. If one of those is returned and then automatically curried, we know what can happen.

I've seen those posts but I can't find them right now? Where are they? I can't think of any harm at the top of my head since our curry plays fine with variadic functions and optional parameters.

test/compose.js
CrossEye: > Interesting. I'd thought the alphabetical ordering of `f` and `g` was suggestive of application order

Instead it's probably from the pure mathematical usage like this:

    h(x) = f(g(x)), for all x
    <==>
    h = f ∘ g

CrossEye:
> I've seen those posts but I can't find them right now? 

There are probably others, but #485, #736, #771.

I keep reusing variants on [the same example][ex]: 

```js
var fullName = function(first, last, suffix) {
    return first + ' ' + last + (suffix ? ', ' + suffix : '');
};

var f = R.compose(R.toUpper, fullName);
var g = R.curry(f);

f('Lon', 'Chaney', 'Jr.'); //=> LON CHANEY, JR
f('Boris', 'Karloff'); //=> BORIS KARLOFF
g('Lon', 'Chaney', 'Jr.'); //=> LON CHANEY, JR
g('Boris', 'Karloff'); :: String - > String
```


  [ex]: http://bit.ly/1BM4Irr

buzzdecafe:
@CrossEye :

> it's worth observing that `useWith` and `converge` already curry their output functions. 

do you argue that `compose/pipe` deserve special treatment? or that `useWith` and `converge` should not curry their output? 

Incidentally, I have seen no issues w/r/t this on `useWith` and `converge`, but I have seen several issues asking for `compose` to output a curried function.

CrossEye:
I think `useWith` is irrelevant.  It is the `useWith` function that's curried, and although it's curried to a length determined by the number of transformer functions, for better or (probably) worse, it passes through all parameters supplied.  Now if you supply a transformer for an optional parameter and then want the function to run without it, you will have an issue, but I think at that point it's a matter of _caveat emptor_.

`converge` is fairly similar.  If the target funciton has a named optional parameter, you really need to decide if you want to handle it when you structure your call to `converge`.  If you do handle it, it will be part of the arity used in the curry.  If not, then it won't.

The trouble I have with auto-currying here -- and as I have said, I do want to do it, preserving only an option for users to avoid it for these unusual cases -- is that it can actually break users' functions in an unrecoverable manner.  I don't think `useWith` or `converge` are in this category.  `curry` is in that category, but the chances there are much more remote.  I'm much less worried about that one.

buzzdecafe:
I'm talkking about the composed function that is returned from `converge` and `useWith`. The output function is *curried*. `compose` should behave the same way.

`converge`:

```js
module.exports = curryN(3, function(after) {
  var fns = _slice(arguments, 1);
  return curryN(max(pluck('length', fns)), function() {
    var args = arguments;
    var context = this;
    return after.apply(context, _map(function(fn) {
      return fn.apply(context, args);
    }, fns));
  });
});
```

`useWith`:

```js
module.exports = curry(function useWith(fn /*, transformers */) {
  var transformers = _slice(arguments, 1);
  var tlen = transformers.length;
  return curry(_arity(tlen, function() {
    var args = [], idx = 0;
    while (idx < tlen) {
      args[idx] = transformers[idx](arguments[idx]);
      idx += 1;
    }
    return fn.apply(this, args.concat(_slice(arguments, tlen)));
  }));
});
```

CrossEye:
I was trying to explain why the currying of those output functions wouldn't cause issues with the types of functions that might cause issues with `compose` and `pipe`.  Maybe when I get out of my interminable meeting I will try to put together actual examples.

paldepind:
> I keep reusing variants on the same example: 

Can you also think of an example this is not [trivially fixable](http://bit.ly/1IyPlzz)? Otherwise I'm afraid I'm not convinced.

```javascript
var fullName = R.curryN(2, function(first, last, suffix) {
    return first + ' ' + last + (suffix ? ', ' + suffix : '');
});

var f = R.compose(R.toUpper, fullName);
var g = R.curry(f);

f('Lon', 'Chaney', 'Jr.'); //=> LON CHANEY, JR
f('Boris', 'Karloff'); //=> BORIS KARLOFF
g('Lon', 'Chaney', 'Jr.'); //=> LON CHANEY, JR
g('Boris', 'Karloff'); //=> BORIS KARLOFF
```

src/internal/_createComposer.js
```diff
@@ -1,4 +1,4 @@
-var _arity = require('./_arity');
+var curryN = require('./curryN');
```
megawac: I would prefer we just document how to generate a curried compose, i.e.

```js
R.composeCurried = R.compose(R.curry, R.compose)
```

CrossEye:
@paldepind:

> Can you also think of an example this is not [trivially fixable](http://bit.ly/1IyPlzz)?

Absolutely not.  This is not something that the user cannot fix.  But that's not the question.  This is about the expectation of what Ramda does with our functions, and what part of it is automated..

But are you going to guarantee that every native and host function in every user environment will have exactly the same public signature?  It certainly hasn't always been that way.  If I have a function that has a named optional parameter in FF but not in IE do you expect me to start testing the environment before passing one of these functions to `compose`?  Or would it be better to simply have a verison of `compose` that does everything our current `compose` does but doesn't add in currying?

(There could also be a problem with variadic functions base on some of the other discussions about reducing these to a fixed-arity output, but that's a separate concern.)

davidchambers:
Superseded by #1260

PR #1225, opened 2015-06-24
Only include dist file when releasing

megawac:
/cc @davidchambers for sanity check

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@
 /tmp
 /bower_components/
 /dist/gh-pages/
+dist/
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,8 @@ lint:
 
 
 .PHONY: release-major release-minor release-patch
-release-major release-minor release-patch:
+release-major release-minor release-patch: dist/ramda.min.js
+	git add --force dist
 	@$(XYZ) --increment $(@:release-%=%)
 
 


.gitignore
```diff
@@ -8,3 +8,4 @@
 /tmp
 /bower_components/
 /dist/gh-pages/
+dist/
```
davidchambers: Couldn't we just remove `gh-pages/` from line 10?

davidchambers:
I don't think this works as you intend. After adding an ignored file with `--force`, Git starts tracking the file. If Git's going to track __dist/ramda.js__ and __dist/ramda.min.js__ there's no reason to include `/dist/` in __.gitignore__.

megawac:
Oh, I didn't think git would track the file after adding it forcefully.
Good to know thanks

.gitignore
buzzdecafe: yes indeed. all that stuff is moved out

CrossEye:
If this was followed by

    git rm --cached dist/\*

would that fix everything?

(I'm a git dunce; this may be a ridiculous question.)

davidchambers:
Now that I think about this again I'm not sure what we're hoping to achieve. We need __dist/ramda.min.js__ to be available somewhere on GitHub for the benefit of CDNs, don't we? We're already updating this file as infrequently as possible. @megawac, were you hoping to prevent __dist/ramda.js__ from appearing in the `git status` output? This would certainly be nice, but so far we haven't had problems with people updating this file in pull requests.

megawac:
Yeah, that was the idea. It shows up in my git status every now and then

CrossEye:
I'm pretty sure that we don't need it for the CDNs, which pull it from NPM.

We have had a number of issues with people committing this file in the past.  Not having it in `git status` would be nice.

OTOH, I would most like to have this file available with a `git clone`, but updated to a version built from HEAD.  Since I haven't done anything about figuring how to do that with Travis, I'm all for removing this from our Git distribution this way, if we can get it to work.

PR #1229, opened 2015-06-26
make R.converge a binary function

davidchambers:
Variadic functions with one or more "special" positions should be avoided. `R.converge` is currently in this category: the first argument is special, but the structure of `R.converge(f, g, h)` provides no indication of this. `R.converge(f, [g, h])` indicates that `g` and `h` are more closely related than `f` and `g`.

### DIFF
--- a/src/converge.js
+++ b/src/converge.js
@@ -1,24 +1,25 @@
+var _curry2 = require('./internal/_curry2');
 var _map = require('./internal/_map');
-var _slice = require('./internal/_slice');
 var curryN = require('./curryN');
 var max = require('./max');
 var pluck = require('./pluck');
 
 
 /**
- * Accepts at least three functions and returns a new function. When invoked, this new
- * function will invoke the first function, `after`, passing as its arguments the
- * results of invoking the subsequent functions with whatever arguments are passed to
- * the new function.
+ * Takes a combining function, `combine`, and a list of functions, `fns`.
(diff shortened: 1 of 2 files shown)


scott-christopher:
I guess it's worth asking, should `converge` be limited to only accept unary functions? If it is limited to unary functions, `converge` could accept the argument that is to be passed to the list of functions as a third argument, e.g.:
```
((b, c, ..., n) -> v) -> [(a -> b), (a -> c), ..., (a -> n)] -> a -> v
// or
converge(add, mul(2), sub(3), 5); // 12
```
I guess it could also support further arguments after a list of non-unary functions, if we're willing to reintroduce variadic args.

davidchambers:
I :heart: this idea, @scott-christopher. The type signature is complex enough when we pretend the list must contain unary functions. I frequently use `R.converge` to create unary functions; for binary functions a lambda is likely to be more appropriate.

buzzdecafe:
i'd rather see `converge` defined as ternary: `(f, g, h) -> f(g(x...), h(x...))`. that seems simplest to me. i wonder if there is much call for more than two functions passing to `f`.

kwijibo:
is `r.useWith` in this boat too? (ie, a variadic function with one or more special position)

buzzdecafe:
yes, `useWith` suffers from the same problem.

davidchambers:
Closing in favour of #1237

PR #1231, opened 2015-06-26
make R.min, R.max, R.minBy, and R.maxBy binary functions

### LINKED ISSUE #1230: R.min and R.max should be binary functions
There are (at least) two ways in which these functions may be used:

  - to find the min/max value in a list of arbitrary length; or
  - to limit a value.

The current functions are perfect for the first use case but a poor fit for the second. In fact, several times I've found myself using `R.partial(Math.max, 0)` or even defining my own `max` function: `var max = R.curry(Math.max);`.

It's easier to derive the list form from the binary form than vice versa.

Current API:

```javascript
// :: [a] -> a
var f = R.max;

// :: Number -> Number
var g = R.compose(R.max, R.concat([0]), R.of);
```

Proposed API:

```javascript
// :: [a] -> a
var f = R.reduce(R.max, -Infinity);

// :: Number -> Number
var g = R.max(0);
```

Were we to take the same approach with `R.minBy` and `R.maxBy` we'd avoid the difficult question of what to return given the empty list, as the user is *forced* by `R.reduce` to specify a default value! :)

Also, I expect these functions to work for any ordered type. Currently `R.max(['a', 'b', 'c'])` gives `-Infinity` rather than `'c'`.

davidchambers:
Closes #1230

This pull request also makes these functions work for any ordered type.

The fact that we're able to remove `_createMaxMin` and `_createMaxMinBy` is a nice bonus. `R.reduce` is a useful function!


src/maxBy.js
```diff
@@ -1,24 +1,23 @@
...
- * @sig (a -> Number) -> [a] -> a
- * @param {Function} keyFn A comparator function for elements in the list
- * @param {Array} list A list of comparable elements
- * @return {*} The greatest element in the list. `undefined` if the list is empty.
+ * @category Relation
+ * @sig Ord b => (a -> b) -> a -> a -> a
+ * @param {Function} f
+ * @param {*} a
+ * @param {*} b
+ * @return {*}
```
davidchambers: I should write a parser to derive the JSDoc cruft from the type signature. The four lines above are nothing but noise.

CrossEye: That's basically how I feel about JavaDoc-style docs in general.  It's why I love the idea of [transcribe][tr] and perhaps even more, [NaturalDocs][nd].  If I ever find the time, I might work on a version of the ideas of the latter made more specific to Javascript.

  [nd]: http://www.naturaldocs.org/
  [tr]: https://github.com/plaid/transcribe

CrossEye:
:herb:

src/maxBy.js
scott-christopher: Perhaps something like this might help? https://github.com/phadej/typify-parser

CrossEye: @scott-christopher: Perhaps it would.  I'm afraid I'd have to learn a lot more than I want to about JSDoc in order to integrate it, though.  :smile:

jon49:
So, before I could just take a list of objects and get the lowest value in an object and return that object with `minBy`. With this change I can no longer do that. Is there an easy way to go about that now? I imagine I'll be using applicative and apply with a filter? Upgrading from 0.14 has been quite the feat with so many changes!

davidchambers:
You could use a fold:

```javascript
> R.reduce(R.minBy(R.prop('x')), {x: Infinity}, [{x: 2}, {x: 1}, {x: 3}])
{x: 1}
```

Is this acceptable? Having to provide an initial value is a bit awkward, but it does avoid creating partial functions (i.e. functions not defined when the input is `[]`).

jon49:
Yeah, that works, just figured it out myself too (although yours is a bit more elegant), I wasn't even thinking about reduce for some reason, slowed me down, obvious though, when I see it.

That would probably be a nice example to add to `minBy` when people are migrating or comparing to `lodash` functions.

Thanks.

davidchambers:
> That would probably be a nice example to add to `minBy` when people are migrating or comparing to `lodash` functions.

Good idea.

PR #1232, opened 2015-06-26
fix metadata for R.addIndex

davidchambers:
The missing type breaks our script for generating the HTML documentation.

### DIFF
--- a/src/addIndex.js
+++ b/src/addIndex.js
@@ -19,8 +19,8 @@ var curryN = require('./curryN');
  * @category Function
  * @category List
  * @sig ((a ... -> b) ... -> [a] -> *) -> (a ..., Int, [a] -> b) ... -> [a] -> *)
- * @param {Function} fn A list iteration function that does not pass index/list to its callback
- * @return An altered list iteration function thats passes index/list to its callback
+ * @param {Function} fn A list iteration function that does not pass index or list to its callback
+ * @return {Function} An altered list iteration function that passes (item, index, list) to its callback
  * @example
  *
  *      var mapIndexed = R.addIndex(R.map);


davidchambers:
More accurately, it broke old versions of the script. @CrossEye added a workaround in https://github.com/ramda/ramda.github.io/commit/795cbe836e295f6fe627f0c35f2269d5cc0660dd, but we should fix the root cause regardless.

src/addIndex.js
```diff
@@ -20,7 +20,7 @@ var curryN = require('./curryN');
  * @category List
  * @sig ((a ... -> b) ... -> [a] -> *) -> (a ..., Int, [a] -> b) ... -> [a] -> *)
  * @param {Function} fn A list iteration function that does not pass index/list to its callback
- * @return An altered list iteration function thats passes index/list to its callback
+ * @return {Function} An altered list iteration function thats passes index/list to its callback
```
buzzdecafe: > function **thats** passes index/list 

typo

and "item, index, and list" would be clearer

src/addIndex.js
```diff
@@ -20,7 +20,7 @@ var curryN = require('./curryN');
  * @category List
  * @sig ((a ... -> b) ... -> [a] -> *) -> (a ..., Int, [a] -> b) ... -> [a] -> *)
```
buzzdecafe: > (a ..., Int, [a] -> b)

i think we should start wrapping these for clarity. For instance, this is not some `a`'s, an `Int`, and a function from `[a] -> b`. clearer IMO would be:

`((a ..., Int, [a]) -> b)`

davidchambers: I don't understand the `...` in `(a ..., Int, [a])`. Shouldn't it just be `(a, Int, [a])`?

Can you share your suggestion for the entire signature?

buzzdecafe: i just copied what you had there

davidchambers: I don't want to change a signature I don't understand. Let's leave this for a separate pull request. :)

buzzdecafe: sounds good. i also don't understand the intended meaning of the ellipsis after `a` in the function sig.

CrossEye: I would welcome a clearer version of this signature.  But the function has some inherent complexity in that it's supposed to handle various kinds of cases, including:

**Map:**

    (inVal -> outVal) -> list -> result
    ((inVal, idx, list) -> outVal) -> list -> result

**Reduce**

    ((accumVal, currVal) -> newVal) -> initVal -> list -> result
    ((accumVal, currVal, idx, list) -> newVal) -> initVal -> list -> result

The nested ellipses are meant to represent the difference between `inVal -> outVal` and `(accumVal, currVal) -> newVal`, allowing for the possibility that a function might have still more parameters than this.  The top-level ellipses are to represent the difference between `fn -> list -> result` and `fn -> initVal -> list -> result`, again with the possibility that there might be even more parameters.

Any suggestions would be welcome.

PR #1233, opened 2015-06-27
Make `commute` more efficient.

joneshf:
Since `f` is a `Functor`, the identity law provides that `map(identity) = identity`.
So we needn't actually map over the value.

See https://github.com/fantasyland/fantasy-land#functor for more information.

### DIFF
--- a/src/commute.js
+++ b/src/commute.js
@@ -1,6 +1,5 @@
 var commuteMap = require('./commuteMap');
 var identity = require('./identity');
-var map = require('./map');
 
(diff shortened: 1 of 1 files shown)


joneshf:
Sorry for the multiple commits, I'll squash these later.

davidchambers:
Very nice!

joneshf:
Squashed.

CrossEye:
D'oh!  :smile:

Issue #1067, opened 2015-05-06
reduceObj ?


Cottin:
How would you guys feel about `reduceObj` ?
In Clojure there is `reduce-kv` https://clojuredocs.org/clojure.core/reduce-kv.

It feels usable but maybe I'm overlooking some smart way of using existing function(s) in ramda to achieve the same thing?

buzzdecafe:
uh-oh

CrossEye:
Interpreting pithy warnings from @buzzdecafe sometimes takes a little guide:

#257, #364, #546, #625, #656, #920

... and probably others.  

Of course you can always just do this:

```js
var foldObj = R.useWith(R.reduce, R.identity, R.identity, R.values);
foldObj(R.add, 0, {a: 1, b: 2, c: 3}); //=> 6
```

Cottin:
Ahh, sorry, I should have searched a bit more before posting.
Thanks for links! ...and implementation suggestion!

CrossEye:
No worries.

My [chief concern][co] is with code like this:

```js
var diff = function(a, b) {return b - a;}, 
    obj1 = {a: 1, b: 2, c: 3}, 
    obj2 = {b: 2, a: 1, c: 3};
_.reduce(obj1, diff); //=> 2
_.reduce(obj2, diff); //=> 4
```

ES6 will at least standardize this, so that we don't have to worry about this oddity being implementation-dependent.  But it's still fairly nasty.

  [co]: https://github.com/ramda/ramda/pull/257#issuecomment-52532396

jdalton:
> ES6 will at least standardize this, so that we don't have to worry about this oddity being implementation-dependent. But it's still fairly nasty.

ES6 standardizes the existing de facto behavior, the one you disagree with. By ES6, `obj1` and `obj2` should be iterated differently.

CrossEye:
>> ES6 will at least standardize this, so that we don't have to worry about this oddity being implementation-dependent. But it's still fairly nasty.

> Yap, ES6 standardizes the existing de facto intuitive behavior, the one you disagree with. By ES6, `obj1` and `obj2` should be iterated differently.

I don't love this iteration order, it's true.  It's made worse by exceptions for small integers:

    Object.keys({a: 1, b: 2, 10: 3}); //=> ["10", "a", "b"]

[...]

PR #1235, opened 2015-06-28
support Map, Set, WeakMap, and WeakSet in R.equals

davidchambers:
Set the query string to `?w=` when viewing this pull request to hide the indentation changes.

### DIFF
--- a/src/internal/_equals.js
+++ b/src/internal/_equals.js
@@ -1,67 +1,95 @@
+var _arrayFromIterator = require('./_arrayFromIterator');
 var _has = require('./_has');
 var identical = require('../identical');
 var keys = require('../keys');
 var type = require('../type');
 
-// The algorithm used to handle cyclic structures is
-// inspired by underscore's isEqual
+
 module.exports = function _equals(a, b, stackA, stackB) {
-  var typeA = type(a);
-  if (typeA !== type(b)) {
-    return false;
-  }
-
-  if (typeA === 'Boolean' || typeA === 'Number' || typeA === 'String') {
-    return typeof a === 'object' ?
-      typeof b === 'object' && identical(a.valueOf(), b.valueOf()) :
-      identical(a, b);
-  }
-
   if (identical(a, b)) {
     return true;
   }
 
-  if (typeA === 'RegExp') {
-    // RegExp equality algorithm: http://stackoverflow.com/a/10776635
-    return (a.source === b.source) &&
-           (a.global === b.global) &&
-           (a.ignoreCase === b.ignoreCase) &&
-           (a.multiline === b.multiline) &&
-           (a.sticky === b.sticky) &&
-           (a.unicode === b.unicode);
+  if (type(a) !== type(b)) {
+    return false;
   }
 
-  if (Object(a) === a) {
-    if (typeA === 'Date' && a.getTime() !== b.getTime()) {
+  switch (type(a)) {
+    case 'Null':
+    case 'Undefined':
       return false;
-    }
-
-    var keysA = keys(a);
-    if (keysA.length !== keys(b).length) {
-      return false;
-    }
-
-    var idx = stackA.length - 1;
-    while (idx >= 0) {
-      if (stackA[idx] === a) {
-        return stackB[idx] === b;
-      }
-      idx -= 1;
-    }
-
-    stackA[stackA.length] = a;
-    stackB[stackB.length] = b;
-    idx = keysA.length - 1;
-    while (idx >= 0) {
-      var key = keysA[idx];
-      if (!_has(key, b) || !_equals(b[key], a[key], stackA, stackB)) {
+    case 'Array':
+    case 'Object':
+      break;
+    case 'Boolean':
+    case 'Number':
+    case 'String':
+      if (!(typeof a === typeof b && identical(a.valueOf(), b.valueOf()))) {
         return false;
       }
-      idx -= 1;
-    }
-    stackA.pop();
-    stackB.pop();
-    return true;
+      break;
+    case 'Date':
+      if (!identical(a.valueOf(), b.valueOf())) {
+        return false;
+      }
+      break;
+    case 'RegExp':
+      if (!(a.source === b.source &&
+            a.global === b.global &&
+            a.ignoreCase === b.ignoreCase &&
+            a.multiline === b.multiline &&
+            a.sticky === b.sticky &&
+            a.unicode === b.unicode)) {
+        return false;
+      }
+      break;
+    case 'Map':
+    case 'Set':
+      if (!_equals(_arrayFromIterator(a.entries()), _arrayFromIterator(b.entries()), stackA, stackB)) {
+        return false;
+      }
+      break;
+    case 'Int8Array':
+    case 'Uint8Array':
+    case 'Uint8ClampedArray':
+    case 'Int16Array':
+    case 'Uint16Array':
+    case 'Int32Array':
+    case 'Uint32Array':
+    case 'Float32Array':
+    case 'Float64Array':
+      break;
+    case 'ArrayBuffer':
+      break;
+    default:
+      // Values of other types are only equal if identical.
+      return false;
   }
-  return false;
+
+  var keysA = keys(a);
+  if (keysA.length !== keys(b).length) {
+    return false;
+  }
+
+  var idx = stackA.length - 1;
+  while (idx >= 0) {
+    if (stackA[idx] === a) {
+      return stackB[idx] === b;
+    }
+    idx -= 1;
+  }
+
+  stackA.push(a);
+  stackB.push(b);
+  idx = keysA.length - 1;
+  while (idx >= 0) {
+    var key = keysA[idx];
+    if (!(_has(key, b) && _equals(b[key], a[key], stackA, stackB))) {
+      return false;
+    }
+    idx -= 1;
+  }
+  stackA.pop();
+  stackB.pop();
+  return true;
 };
--- a/src/internal/_arrayFromIterator.js
+++ b/src/internal/_arrayFromIterator.js
@@ -1,0 +1,8 @@
+module.exports = function _arrayFromIterator(iter) {
+  var list = [];
+  var next;
+  while (!(next = iter.next()).done) {
+    list.push(next.value);
+  }
+  return list;
+};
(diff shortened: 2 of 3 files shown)


src/internal/_equals.js
```diff
@@ -1,67 +1,71 @@
...
-    if (keysA.length !== keys(b).length) {
+    case 'Boolean':
+    case 'Number':
+    case 'String':
+      return typeof a === typeof b && identical(a.valueOf(), b.valueOf());
+    case 'Date':
+      if (!identical(a.valueOf(), b.valueOf())) {
+        return false;
+      }
+      break;
```
davidchambers: Why is this branch not simply `return identical(a.valueOf(), b.valueOf());`, you ask? Good question. The test suite currently contains this assertion:

```javascript
var date1 = new Date(0);
var date4 = new Date(0);
date4.v = 'value';
assert.strictEqual(R.equals(date1, date4), false);
```

While this behaviour is reasonable, I can think of no reason it should apply to Date objects but not certain other object types. For example:

```javascript
> var a = /(?:)/
undefined
> var b = /(?:)/
undefined
> b.foo = 'bar'
'bar'
> R.equals(a, b)
true
```

We should either stop checking properties of Date objects or apply these same checks to other object types. This should happen in a separate pull request, but we can discuss the options here.

I lean towards only considering objects equal if they have the same set of properties and the values of each property are equal. It's a good thing, in my view, that `R.equals` does not consider the following values equivalent:

```javascript
> R.match(/foo(bar)?/, 'foobarbaz')
[ 'foobar', 'bar', index: 0, input: 'foobarbaz' ]
> ['foobar', 'bar']
[ 'foobar', 'bar' ]
```

asaf-romano:
How did you compare WeakMap/Set? Their keys are not enumerable.

davidchambers:
> How did you compare WeakMap/Set? Their keys are not enumerable.

All we can do is compare by identity, but something is better than nothing. :)

CrossEye:
I'm still doing triage on all the issues/prs since I've been gone.  This is one of those that's going to take me more than a few minutes to absorb.  It looks good at first glance, though.

src/internal/_equals.js
davidchambers: > We should either stop checking properties of Date objects or apply these same checks to other object types. This should happen in a separate pull request, but we can discuss the options here.

What are your thoughts on this, @CrossEye?

CrossEye: I'm sorry.  I meant to look at this one this evening, but I forgot and am falling asleep now.  I'll look at it more carefully tomorrow.

davidchambers: ((:bell:))

CrossEye: Even after I assigned it to myself.  :smile:

But... LGTM

davidchambers: I've extended the _same enumerable properties_ requirement to all object types.

src/internal/_equals.js
```diff
@@ -1,67 +1,95 @@
...
+    case 'Uint8ClampedArray':
+    case 'Int16Array':
+    case 'Uint16Array':
+    case 'Int32Array':
+    case 'Uint32Array':
+    case 'Float32Array':
+    case 'Float64Array':
+      break;
+    case 'ArrayBuffer':
+      break;
```
davidchambers: Does it even make sense to apply the notion of value-based equality to objects of these types?

CrossEye: Why wouldn't it?

Not arguing, just think I'm missing the point.

davidchambers: It just feels odd to apply the notion of value-based equality to buffers which (as I understand them) are only useful in concert with mutation. Though arrays are mutable we can choose to treat arrays as immutable values and still have something useful. I'm not sure that immutable buffers would be any use at all.

If I'm suggesting that only immutable objects (or mutable objects we pretend are immutable in languages which lack immutable data structures) should be compared by value, Python disagrees:

```python
>>> (1, 2, 3) == (1, 2, 3)
True
>>> [1, 2, 3] == [1, 2, 3]
True
```

Python has an immutable ordered collection (`tuple`) yet it still performs value-based equality on its mutable ordered collection (`list`).

davidchambers:
Could I get some sort of emoji on this pull request? :)

CrossEye:
Have an :herb: from me.

src/internal/_equals.js
```diff
@@ -1,67 +1,95 @@
...
+      if (!(a.source === b.source &&
+            a.global === b.global &&
+            a.ignoreCase === b.ignoreCase &&
+            a.multiline === b.multiline &&
+            a.sticky === b.sticky &&
+            a.unicode === b.unicode)) {
+        return false;
+      }
+      break;
+    case 'Map':
```
jdalton: with this patch maps of same values assigned in different order are seen as unequal.

davidchambers: #1364

CrossEye: Thanks @jdalton.  Very much related to #1067 !

paldepind:
Comparing the equality of sets by comparing their elements with `R.equals` is silly.

We have to ask ourselves: Why are Sets useful? Because checking for elements is fast with `Set.prototype.has`! That is the primary reason for using them.

`Set.prototype.has` does not use `R.equals`. It uses an equality test quite similar to `===`. I.e. after this this `R.equals(set1, set2) === true` does not imply `set1.has(x) === set2.has(x)`. We are breaking a fundamental law with regards to sets.

And we have no way to redefine `Set.prototype.has` because then we lose the performance benefit and the entire reason for using Sets in the first place.

I think the obsession that this library has with `R.equals` is doing more harm than good.

CrossEye:
I think you're missing the point, though.

If you have another implementation for Set, that would be welcome.  This is a fix for a very serious issue, to wit, in the current implementation, 

```js
R.equals(new Set([1, 2]), new Set([2, 1])); //=> false
```

Although the current implementation does use an `R.equals` check on the elements, it would also fail if we switched that to a `===` or `==` check, as the problem is iteration order.

That behavior needs to be fixed.  If you want to code an alternative that depends only on `Set.prototype.has`, I have no immediate objection as long as it fixes such problems... but I will point out that you were the one [who noted][sv] that "Defaulting to value equality seems very sensible for a functional programming library."



  [sv]: https://github.com/ramda/ramda/pull/1096#issuecomment-104159074

CrossEye:
> We have to ask ourselves: Why are Sets useful? Because checking for elements is fast with `Set.prototype.has`! That is the primary reason for using them.

Really?  To me the main point is that they are containers that allow only a single copy of an individual value.  That is my primary reason for using them.  Bags/Multiset have other uses of their own, as do Lists, Trees, and other data structures.  Performance is only a minor concern for me in choosing such a fundamental structure.

paldepind:
> I think you're missing the point, though.

No. You are mistaking the PR.

> If you have another implementation for Set, that would be welcome. This is a fix for a very serious issue, to wit, in the current implementation, 

That is another PR. Not this one.

> Although the current implementation does use an R.equals check on the elements, it would also fail if we switched that to a === or == check, as the problem is iteration order.

Again. It has nothing to do with this.

> That behavior needs to be fixed. If you want to code an alternative that depends only on Set.prototype.has, I have no immediate objection as long as it fixes such problems.

I'm glad you think it needs to be fixed. Then we will have to stop using `R.equals` between the elements. The algorithm is simple and will perform much better than what this PR does.

1. Compare size of set. If not equal, the sets are not equal.
2. Iterate over each element in first set while checking for presence in other set with `Set.prototype.has`. If and item is missing the sets are not equal. Otherwise they are.

> but I will point out that you were the one who noted that "Defaulting to value equality seems very sensible for a functional programming library."

I still think so. But suddenly `R.equals` is a deciding factor when it should not be.

> Really? To me the main point is that they are containers that allow only a single copy of an individual value. That is my primary reason for using them.

Yes. Of course. That is also fundamental. And that stems directly from having an efficient way to check for the presence of an element. My thinking was more along the lines of: Why do we care about Set when we can fake them with arrays. The answer is performance.

jdalton:
> Iterate over each element in first set while checking for presence in other set with `Set.prototype.has`

Not sure if `has` is total approp for `equals`:
```js
equals({a:1}, {a:1}); // true
equals(new Set([{a:1}]), new Set([{a:1}]); // false
```

paldepind:
@jdalton It certainly is not (it's close to `===`). That is the point. We can not get the semantics of `R.equals` here.

jdalton:
There is equality for values being included in a set and then there's the `equals` equivalence method. I see them as different. I think `equals` should be consistent with how it crawls values. So if that means being looser than a sets SameValueZero comparison then that's ok for `equals`.

CrossEye:
>> I think you're missing the point, though.

> No. You are mistaking the PR.

I'm pretty sure I'm not.  In 76553fb we added this implementation to `_equals` for `Set` and `Map`:

```patch
+    case 'Map':
+    case 'Set':
+      if (!_equals(_arrayFromIterator(a.entries()), \
                    _arrayFromIterator(b.entries()), stackA, stackB)) {
+        return false;
+      }
+      break;
```

The immediate trouble is that this ties us to the iteration order of Sets, which, unsurprisingly, I guess, is insertion order, so regardless of how we compare individual elements, `R.equals`, `R.identical`, `(a, b) => a === b`, whatever, we will never get `set1` to be equal to `set2` if they're defined like this:

```js
var s1 = new Set
s1.add(1);
s1.add(2);
var s2 = new Set();
s2.add(2);
s2.add(1);
```
 
And regardless of our debate in #1067, I imagine we can agree that this is a problem.

I think we do need to do something more than the fix David has proposed, but mostly to avoid using `toString` to compare them.  I'll comment on that over in #1364, though.

paldepind:
> I'm pretty sure I'm not.

Well. You wrote:

> This is a fix for a very serious issue, to wit, in the current implementation, 

And no it is not. That is #1364. This is the PR that added the current implementation.

> will never get set1 to be equal to set2 if they're defined like this:

The algorithm I proposed does not suffer from that issue. With that algorithm they would be equal.

CrossEye:
I seem to have misunderstood you entirely,  and I apologize.   I thought you were responding to the comments from @jdalton and saying that there was no issue here. 

I agree at least in part with your suggested algorithm.   I still think we need to hash out whether we would want to consider value equality of elements.  And it might be a tricky question.

paldepind:
> I seem to have misunderstood you entirely, and I apologize. I thought you were responding to the comments from @jdalton and saying that there was no issue here. 

No problem :)

Issue #1228, opened 2015-06-25
remove R.isEmpty


davidchambers:
`R.isEmpty` is sugar for `R.propEq('length', 0)`. The problem, though, is that people don't necessary *realize* that this is the case. Some might assume, for example, that `R.isEmpty({})` evaluates true (#538, #1194). A colleague just ran into problems by assuming `R.isEmpty(null)` and `R.isEmpty(undefined)` evaluate true.

Sugar is nice, but in this case the desugaring is not obvious.

buzzdecafe:
:cow2:

CrossEye:
I know this one has had it's share of controversy.  But I'm in favor of keeping it.

Saying "abc is sugar for xyz" is different from saying "abc can be implemented as xyz".  I think this is the latter case.  "`pluck` is sugar for `map(prop)`".  I have no problem with that.  It's simply noting that `pluck` is a syntactic simplification of the ideas expressed by `map(prop)`.

But this case is different.  The idea is expressed by `isEmpty`.  One implementation of this is `propEq('length', 0)`.  But we have variously considered and/or adopted other variations of the same idea of `isEmpty` that could not have been expressed by `propEq('length', 0)`.  If we hadn't gotten bogged down by all the crazy details of what it might mean, this might have been our first function that operated on lists, strings, and plain objects.  I could easily see taking that up again as in this case, it seems fairly reasonable (except for the pesky details!)

So I would rather keep the name.  It is reasonably documented.

paldepind:
:+1: All the cases shows that people confuse the name of this function with what it does. It implies something else. `R.propEq('length', 0)` on the other hand is impossible to misunderstand.

We should not keep a confusing function just because we might one day implement something that does what it says. If that is the goal we should make that change _now_ or remove it in the mean time to avoid harm.

davidchambers:
The fact that the meaning of `R.isEmpty` has changed in the past and may change again in the future adds to my concern. There are clearly many *semantically* different ways to define a function named `isEmpty` in JavaScript. The only way to determine the behaviour with certainly is to read the source code (documentation is useful, of course, but is not guaranteed to be up to date and to cover every edge case).

TheLudd:
With @CrossEye here. 

I think the function is clearly documented. If it is considered a problem that people make incorrect assumptions about functions by just looking at the name then a whole lot of ramda functions are in trouble.

[...]

PR #1236, opened 2015-06-30
define R.empty for Object and String

davidchambers:
As suggested by @scott-christopher and @paldepind in #1228

### DIFF
--- a/src/empty.js
+++ b/src/empty.js
@@ -1,21 +1,34 @@
 var _curry1 = require('./internal/_curry1');
-var _hasMethod = require('./internal/_hasMethod');
 
 
 /**
- * `empty` returns an empty list for any argument, except when the argument satisfies the
- * Fantasy-land Monoid spec. In that case, this function will return the result of invoking
- * `empty` on that Monoid.
+ * Returns the empty value of its argument's type. Ramda defines the empty
+ * value of Array (`[]`), Object (`{}`), and String (`''`). Other types are
+ * supported if they define `<Type>.empty` and/or `<Type>.prototype.empty`.
  *
  * @func
  * @memberOf R
  * @category Function
- * @sig * -> []
- * @return {Array} An empty array.
+ * @sig a -> a
+ * @param {*} x
+ * @return {*}
  * @example
  *
- *      R.empty([1,2,3,4,5]); //=> []
+ *      R.empty(Just(42));      //=> Nothing()
+ *      R.empty([1, 2, 3]);     //=> []
+ *      R.empty('unicorns');    //=> ''
+ *      R.empty({x: 1, y: 2});  //=> {}
  */
 module.exports = _curry1(function empty(x) {
-  return _hasMethod('empty', x) ? x.empty() : [];
+  if (x != null && typeof x.empty === 'function') {
+    return x.empty();
+  } else if (x != null && typeof x.constructor != null && typeof x.constructor.empty === 'function') {
+    return x.constructor.empty();
+  } else {
+    switch (Object.prototype.toString.call(x)) {
+      case '[object Array]':  return [];
+      case '[object Object]': return {};
+      case '[object String]': return '';
+    }
+  }
 });
--- a/test/empty.js
+++ b/test/empty.js
@@ -4,8 +4,42 @@ var R = require('..');
 
 
 describe('empty', function() {
-  it('returns an empty list', function() {
-    assert.deepEqual(R.empty([1, 2, 3]), []);
+
+  it('dispatches to `empty` method', function() {
+    function Nothing() {}
+    Nothing.prototype.empty = function() { return new Nothing(); };
+
+    function Just(x) { this.value = x; }
+    Just.prototype.empty = function() { return new Nothing(); };
+
+    assert.strictEqual(R.empty(new Nothing()).constructor, Nothing);
+    assert.strictEqual(R.empty(new Just(123)).constructor, Nothing);
+  });
+
+  it('dispatches to `empty` function on constructor', function() {
+    function Nothing() {}
+    Nothing.empty = function() { return new Nothing(); };
+
+    function Just(x) { this.value = x; }
+    Just.empty = function() { return new Nothing(); };
+
+    assert.strictEqual(R.empty(new Nothing()).constructor, Nothing);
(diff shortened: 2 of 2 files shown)


src/empty.js
```diff
@@ -1,21 +1,36 @@
...
-var _hasMethod = require('./internal/_hasMethod');
 
 
 /**
- * `empty` returns an empty list for any argument, except when the argument satisfies the
- * Fantasy-land Monoid spec. In that case, this function will return the result of invoking
- * `empty` on that Monoid.
+ * Returns the empty value of its argument's type. The argument must be a
+ * [Monoid](https://github.com/fantasyland/fantasy-land#monoid), Array,
+ * Object, or String.
```
scott-christopher: Should we perhaps mention that calling with an unsupported argument will now result in `undefined` being returned?

davidchambers: At the moment the behaviour is left unspecified (returning `undefined` happens to fall out of the implementation). Throwing is another option. My preference is to throw given `null` or `/xxx/` or a value of any other unsupported type, but Ramda tends to follow JavaScript's lead by failing silently (`R.head([])` does not throw, for example).

If we agree that we should return `undefined` given a value of an unsupported type, I'll update the docstring accordingly.

scott-christopher: Fair point. It was mainly suggested because it is modifying the existing behaviour of `empty` which currently defaults to an empty array and would be good to have some docs to refer people back to if questions are raised.

src/empty.js
```diff
@@ -1,21 +1,36 @@
...
+      return x.empty();
+    case x != null && typeof x.constructor != null && typeof x.constructor.empty === 'function':
+      return x.constructor.empty();
+    case repr === '[object Array]':
+      return [];
+    case repr === '[object Object]':
+      return {};
+    case repr === '[object String]':
+      return '';
+  }
```
paldepind: I would personally prefer a nested ternary operator here. `switch (true)` is quite ugly in my opinion. Something along these lines:

```javascript
return x != null && typeof x.empty === 'function' ? x.empty()
     : repr === '[object Array]'  ? []
     : repr === '[object Object]' ? {}
                                  : undefined;
```

src/empty.js
```diff
@@ -1,21 +1,36 @@
...
+ *      R.empty(Just(42));      //=> Nothing()
+ *      R.empty([1, 2, 3]);     //=> []
+ *      R.empty('unicorns');    //=> ''
+ *      R.empty({x: 1, y: 2});  //=> {}
  */
 module.exports = _curry1(function empty(x) {
-  return _hasMethod('empty', x) ? x.empty() : [];
+  var repr = Object.prototype.toString.call(x);
+  switch (true) {
+    case x != null && typeof x.empty === 'function':
```
paldepind: What about using `_hasMethod`?

davidchambers: `_hasMethod` doesn't look for a function on the constructor, which we need to do to comply with the FL spec.

src/empty.js
davidchambers: :zap:

paldepind:
I like it :smile_cat:

svozza:
Yeah, this is very nice.

buzzdecafe:
looks ok, but it does make we wonder what this means for `isEmpty`

paldepind:
@buzzdecafe https://github.com/ramda/ramda/issues/1228#issuecomment-115448064

buzzdecafe:
yeah that's what i was thinking, `isEmpty(x) <=> equals(x, empty(x))`

src/empty.js
```diff
@@ -1,21 +1,38 @@
 var _curry1 = require('./internal/_curry1');
-var _hasMethod = require('./internal/_hasMethod');
 
 
 /**
- * `empty` returns an empty list for any argument, except when the argument satisfies the
- * Fantasy-land Monoid spec. In that case, this function will return the result of invoking
- * `empty` on that Monoid.
+ * Returns the empty value of its argument's type. The argument must be a
+ * [Monoid](https://github.com/fantasyland/fantasy-land#monoid), Array,
```
CrossEye: Although this is not newly introduced, perhaps we should clean it up while we're here.  There is nothing here that I see that requires this to be a Monoid.  Monoids are a subset of Semigroups.  But there is nothing here that requires a closed associative binary operation on the set, the definition of Semigroup, `concat` in FantasyLand terms.

All the implementation needs is to have an `empty` function returning an element of its type.  Is there any reason to document it as actually then needing a full Monoid?

src/empty.js
```diff
@@ -1,21 +1,38 @@
...
+    repr === '[object Array]' ?
+      [] :
+    repr === '[object Object]' ?
+      {} :
+    repr === '[object String]' ?
+      '' :
+    // else
+      void 0
+  );
 });
```
CrossEye: That's a lot of ternary operators!

Do we really want to code in that style?

I'm especially bothered because it forces us to pre-calculate an unnecessary value, `repr`, if either of the first two cases passes.  It's not really a question of performance, but of cleanliness.

CrossEye:
I'm not sure I like the details of the implementation, but I am definitely :+1: on this overall.

And I think it could serve well as the basis for `isEmpty` too.

src/empty.js
davidchambers: I don't mind how we structure this code. Let me know your preferred approach and I'll updated this pull request.

scott-christopher:
```js
module.exports = _curry1(function empty(x) {
  if (x != null && typeof x.empty === 'function') {
    return x.empty();
  } else if (x != null && typeof x.constructor != null && typeof x.constructor.empty === 'function') {
    return x.constructor.empty();
  } else {
    switch (Object.prototype.toString.call(x)) {
      case '[object Array]':  return [];
      case '[object Object]': return {};
      case '[object String]': return '';
    }
  }
});
```

CrossEye: :+1:

davidchambers: :zap:

src/empty.js
davidchambers: Updated to read:

> Returns the empty value of its argument's type. Ramda defines the empty value of Array (`[]`), Object (`{}`), and String (`''`). Other types are supported if they define `<Type>.empty` and/or `<Type>.prototype.empty`.

CrossEye:
:herb:

PR #1237, opened 2015-06-30
diverge via exactly two unary functions in R.converge

davidchambers:
This pull request is based on suggestions by @scott-christopher and @buzzdecafe in #1229. It gives `R.converge` a dramatically simpler signature:

    R.converge :: ((b, c) -> d) -> (a -> b) -> (a -> c) -> a -> d

The implementation is now trivial:

```javascript
module.exports = curry(function converge(combine, f, g, x) {
  return combine(f(x), g(x));
});
```

I looked at a project with 34 occurrences of `R.converge`. The vast majority of these are compatible with the simpler signature. The few which are not are likely better expressed via lambdas. ;)

### DIFF
--- a/test/converge.js
+++ b/test/converge.js
@@ -3,42 +3,38 @@ var assert = require('assert');
 var R = require('..');
 
 
+//  square :: Number -> Number
+var square = function(n) { return n * n; };
+
+
 describe('converge', function() {
-  var mult = function(a, b) {return a * b;};
 
-  var f1 = R.converge(mult,
-                      function(a) { return a; },
-                      function(a) { return a; });
-  var f2 = R.converge(mult,
-                      function(a) { return a; },
-                      function(a, b) { return b; });
-  var f3 = R.converge(mult,
-                      function(a) { return a; },
-                      function(a, b, c) { return c; });
-
-  it('passes the results of applying the arguments individually to two separate functions into a single one', function() {
-    assert.strictEqual(R.converge(mult, R.add(1), R.add(3))(2), 15); // mult(add1(2), add3(2)) = mult(3, 5) = 3 * 15;
+  it('is a ternary function', function() {
+    assert.strictEqual(typeof R.converge, 'function');
+    assert.strictEqual(R.converge.length, 3);
   });
 
-  it('returns a function with the length of the "longest" argument', function() {
-    assert.strictEqual(f1.length, 1);
-    assert.strictEqual(f2.length, 2);
-    assert.strictEqual(f3.length, 3);
+  it('diverges and converges', function() {
+    assert.strictEqual(R.converge(R.subtract, square, R.inc)(10), (10 * 10) - (10 + 1));
   });
 
   it('passes context to its functions', function() {
-    var a = function(x) { return this.f1(x); };
-    var b = function(x) { return this.f2(x); };
-    var c = function(x, y) { return this.f3(x, y); };
-    var d = R.converge(c, a, b);
-    var context = {f1: R.add(1), f2: R.add(2), f3: R.add};
-    assert.equal(a.call(context, 1), 2);
-    assert.equal(b.call(context, 1), 3);
-    assert.equal(d.call(context, 1), 5);
+    var c = function(a, b) { return this.f(a, b); };
+    var f = function(a) { return this.x + a; };
+    var g = function(a) { return this.x - a; };
+    var ctx = {
+      f: function(a, b) { return Math.pow(a - b, this.x); },
+      x: 3
+    };
+
+    assert.strictEqual(R.converge(c, f, g).call(ctx, 0), 0);  // Math.pow((3 + 0) - (3 - 0), 3)
+    assert.strictEqual(R.converge(c, f, g).call(ctx, 1), 8);  // Math.pow((3 + 1) - (3 - 1), 3)
+    assert.strictEqual(R.converge(c, f, g).call(ctx, 2), 64); // Math.pow((3 + 2) - (3 - 2), 3)
   });
 
-  it('returns a curried function', function() {
-    assert.strictEqual(f2(6)(7), 42);
-    assert.strictEqual(f3(R.__).length, 3);
+  it('is curried', function() {
+    assert.strictEqual(R.converge(R.subtract).length, 2);
+    assert.strictEqual(R.converge(R.subtract)(square).length, 1);
   });
+
 });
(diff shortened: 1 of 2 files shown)


buzzdecafe:
:-1: for passing in `x`. IMO this function should take three functions and give you back one. Then you can apply that one as you prefer. So I'm proposing a signature of:

`((b, c) -> d) -> (a... -> b) -> (a... -> c) -> (a... -> d)`

davidchambers:
> IMO this function should take three functions and give you back one.

This function does do this. ;)

> `((b, c) -> d) -> (a... -> b) -> (a... -> c) -> (a... -> d)`

What do the ellipses indicate?

buzzdecafe:
> What do the ellipses indicate?

variadic.

davidchambers:
Can anyone provide a real-world example of `R.converge` being used to create a function with arity greater than one? Each of the 34 occurrences of `R.converge` in one Plaid project creates a unary function.

buzzdecafe:
here's a toy:

```js
var toy = R.converge(R.subtract, R.multiply, R.add);
toy(6, 10); //=> (- (* 6 10) (+ 6 10)) => 44
```

paldepind:
Have you noticed that `R.converge` achieves something very similar to the applicative instance of functions.

This for instance

```javascript
R.converge(add, mul(2), sub(3))(5); // 12
```

Is the same as this:

```javascript
of(add).ap(mul(2)).ap(sub(3))(5); // 12
```

Where `of` and `ap` are the fantasy land method names for applicatives and in this case as they are defined for functions.

```javascript
function of(v) {
  return function(_) { return v; };
};
function ap(g) {
  var f = this;
  return function(x) {
    return f(x)(g(x));
  }
};
```

Untested code btw.

test/converge.js
```diff
@@ -3,42 +3,25 @@ var assert = require('assert');
...
 
-  it('passes context to its functions', function() {
-    var a = function(x) { return this.f1(x); };
-    var b = function(x) { return this.f2(x); };
-    var c = function(x, y) { return this.f3(x, y); };
-    var d = R.converge(c, a, b);
-    var context = {f1: R.add(1), f2: R.add(2), f3: R.add};
-    assert.equal(a.call(context, 1), 2);
-    assert.equal(b.call(context, 1), 3);
-    assert.equal(d.call(context, 1), 5);
```
CrossEye: Why lose this?  Even if you do curry in the actual value, why should the running functions drop their context?

CrossEye:
I will have to think further on this, but my first thoughts are positive.  The two distinct variadic expansions that the current function has against this one were written on spec.  I don't recall any real-world uses for them in out iniitial design.  `useWith` had this sort of API, which was probably too complicated, but perhaps we can come back at that one too.

I'm not sure how I feel about currying in the argument to the function, but I'm going to disagree with @buzzdecafe on a different point.  There are two different ways in which the current API expands on this:  First, the helper functions -- and therefore the resulting function as well -- are variadic.  Second, the combining function does not have to be merely binary; it can be variadic as well.

If we were to keep one of these, I think it's the second one that's far more important.  Although the toy example from @buzzdecafe shows that the first could indeed happen, it seems fairly obscure, and, as @davidchambers said, perhaps should best be handled with a lambda.  The second seems more likely: perform these (say, four) transformations on the input, and combine them with this.

But I think in the end, neither is particularly important.  My own usage would almost always be covered by this simplification, and I agree that those cases where it wouldn't might best not be handled by such a function in any case.

So, for a first thought, a tentative :herb:.

test/converge.js
davidchambers: I couldn't think of an example that isn't contrived. I've reinstated the calls to `Function.prototype.call`. The code is much less elegant now. :cry:

CrossEye: Real examples would come when a user used Ramda to generate OO methods.  It's not a use-case that I want to spend a great deal of effort on, but so far it's been pretty simple to maintain this capability for our function-returning functions.

test/converge.js
```diff
@@ -3,42 +3,38 @@ var assert = require('assert');
...
+    assert.strictEqual(R.converge(c, f, g).call(ctx, 2), 64); // Math.pow((3 + 2) - (3 - 2), 3)
   });
 
-  it('returns a curried function', function() {
-    assert.strictEqual(f2(6)(7), 42);
-    assert.strictEqual(f3(R.__).length, 3);
+  it('is curried', function() {
+    assert.strictEqual(R.converge(R.subtract).length, 2);
+    assert.strictEqual(R.converge(R.subtract)(square).length, 1);
   });
```
CrossEye: This is also a substantive change in the tests.  The old tests demonstrated that `converge(f, g, h)` is curried.  Yours demonstrates that `converge` is curried.  Should we have both?

Certainly the whole show-every-function-is-curried routine has long since gotten old.  But the old unit test was something different.

davidchambers: As it stands this pull request gives `R.converge` type `((b, c) -> d) -> (a -> b) -> (a -> c) -> (a -> d)`. Since the return value is always a unary function, we needn't assert that the return value is a curried function.

If you'd like to make the type more complex we can do so, but that should happen in a separate pull request (which would supersede this one).

CrossEye: Oh, of course, that's right.  I'm still only about three-quarters of the way on-board with the overall change.  I _think_ I like it.  Giving up flexibility for simplicity is always an interesting call.

test/converge.js
davidchambers: A week has passed and I've lost context (pun not actually intended). I believe I reinstated the context-preserving behaviour at some point after you left a comment on the line above. Are the assertions below sufficient?

davidchambers:
I just found a compelling use case for converging binary functions:

```javascript
const xor = R.converge(R.difference, R.union, R.intersection);

xor([1, 2, 3, 4], [7, 6, 5, 4, 3]);
// => [1, 2, 7, 6, 5]
```

Here are the equivalent lambdas:

```javascript
const xor = (a, b) => R.difference(R.union(a, b), R.intersection(a, b));
```

```javascript
var xor = function(a, b) { return R.difference(R.union(a, b), R.intersection(a, b)); };
```

While I'm still in favour of fixing the arity of `R.converge`, in light of this example I'm no longer in favour of requiring *unary* functions as the second and third arguments.

buzzdecafe:
> While I'm still in favour of fixing the arity of R.converge, in light of this example I'm no longer in favour of requiring unary functions as the second and third arguments.

in that case we agree 100%

PR #1239, opened 2015-07-03
Make match return an empty array when there are no matches

TheLudd:
This is a breaking change which will return an empty array for the
`match` function when there are no matches. The old behavior had the
function return `null` in that case.

This deviates from the standard API but is IMHO an improvement. The deviation is also documented.

### DIFF
--- a/src/match.js
+++ b/src/match.js
@@ -1,19 +1,24 @@
+var _curry2 = require('./internal/_curry2');
+var compose = require('./internal/compose');
+var defaultTo = require('./defaultTo');
 var invoker = require('./invoker');
 
 
 /**
- * Tests a regular expression against a String
+ * Tests a regular expression against a String. Note that this function
+ * will return an empty array when there are no matches. This differs
+ * from [`String.prototype.match`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match)
+ * which returns `null` when there are no matches.
  *
  * @func
  * @memberOf R
  * @category String
- * @sig RegExp -> String -> [String] | null
+ * @sig RegExp -> String -> [String | Undefined]
  * @param {RegExp} rx A regular expression.
  * @param {String} str The string to match against
- * @return {Array} The list of matches, or null if no matches found.
- * @see R.invoker
+ * @return {Array} The list of matches.
  * @example
  *
  *      R.match(/([a-z]a)/g, 'bananas'); //=> ['ba', 'na', 'na']
  */
-module.exports = invoker(1, 'match');
+module.exports = _curry2(compose(defaultTo([]), invoker(1, 'match')));
--- a/test/match.js
+++ b/test/match.js
@@ -8,12 +8,12 @@ describe('match', function() {
 
   it('determines whether a string matches a regex', function() {
     assert.strictEqual(R.match(re, 'B17-afn').length, 1);
-    assert.strictEqual(R.match(re, 'B1-afn'), null);
+    assert.deepEqual(R.match(re, 'B1-afn'), []);
   });
 
   it('is curried', function() {
     var format = R.match(re);
     assert.strictEqual(format('B17-afn').length, 1);
-    assert.strictEqual(format('B1-afn'), null);
+    assert.deepEqual(format('B1-afn'), []);
   });
 });


buzzdecafe:
i agree this is a better API

paldepind:
I agree :+1:

svozza:
Yep, +1.

src/match.js
```diff
@@ -1,19 +1,25 @@
...
+ * Tests a regular expression against a String. Note that this function
+ * will return an empty array when there are no matches. This differs
+ * from the standard JavaScript string function which returns null when
+ * there are no matches.
  *
  * @func
  * @memberOf R
  * @category String
- * @sig RegExp -> String -> [String] | null
+ * @sig RegExp -> String -> [String]
```
davidchambers: This should really be `RegExp -> String -> [String | Undefined]` if we're to acknowledge the existence of optional capturing groups:

```javascript
> 'bye'.match(/(good)?bye/)
["bye", undefined]
```

src/match.js
```diff
@@ -1,19 +1,25 @@
...
  * @param {RegExp} rx A regular expression.
  * @param {String} str The string to match against
```
davidchambers: These should match the actual parameter names.

src/match.js
```diff
@@ -1,19 +1,25 @@
...
- * @see R.invoker
+ * @return {Array} The list of matches.
  * @example
  *
  *      R.match(/([a-z]a)/g, 'bananas'); //=> ['ba', 'na', 'na']
  */
-module.exports = invoker(1, 'match');
+module.exports = _curry2(function match(regex, string) {
+  var matches = string.match(regex);
+  return defaultTo([], matches);
```
davidchambers: `defaultTo([], string.match(regex))` seems fine to me. We needn't introduce a variable.

test/match.js
```diff
@@ -8,12 +8,12 @@ describe('match', function() {
 
   it('determines whether a string matches a regex', function() {
     assert.strictEqual(R.match(re, 'B17-afn').length, 1);
-    assert.strictEqual(R.match(re, 'B1-afn'), null);
+    assert.deepEqual(R.match(re, 'B1-afn'), []);
```
davidchambers: It would be good to assert that the return value is `[]` rather than `{}`, since `assert.deepEqual` does not distinguish the two. One option is to use `R.toString`:

```javascript
assert.strictEqual(R.toString(R.match(re, 'B1-afn')), '[]');
```

src/match.js
buzzdecafe: that's unfortunate. i'd prefer not to generate "sparse arrays" since we don't support 'em anywhere.

svozza: Yeah, it's not ideal. `R.pluck` generates spare arrays too.

megawac: How do either this or pluck generate sparse arrays

svozza:
```javascript
R.pluck('b', [{a:1, b:1}, {a:2, c:3}, {a: 3, b: 5}])
// => [1, undefined, 5]
```

buzzdecafe: i think the best course is to document/warn about it, and recommend composing with `filter(complement(isNil))` (or wrapping in `Maybe`s)

megawac: > R.pluck('b', [{a:1, b:1}, {a:2, c:3}, {a: 3, b: 5}])
// => [1, undefined, 5]

That's not a sparse array

```js
var res = R.pluck('b', [{a:1, b:1}, {a:2, c:3}, {a: 3, b: 5}])
console.log(2 in res);
```

paldepind:
An array with some items set to undefined I not sparse. And array with some keys not set at all is aparse.

src/match.js
TheLudd: >  'bye'.match(/(good)?bye/)
["bye", undefined]

What does that result really mean? Is it an option to filter out undefined values?

test/match.js
TheLudd: That sounds really weird to me. How can an assertion framework let that pass? I have checked around in the ramda test code and the `toString` is not used in any other case. To fix this I would suggest switch assertion framework. I use chai in my projects and it throws an assertion error in these cases.

src/match.js
kedashoe: > What does that result really mean? Is it an option to filter out undefined values?

It means that your subgroup did not match, eg

    'bye alice'.match(/(good)?bye (alice|bob)/)
    [ 'bye alice', undefined, 'alice' ]

test/match.js
davidchambers: We could switch assertion libraries. Using `assert.deepEqual` is almost always a bad idea. ;)

src/match.js
TheLudd: Would we loose information if `undefined` values were filtered out? Would that be preferable?

test/match.js
TheLudd: So stay with this for now and perhaps take a big look at the entire ramda project and change assertion framework if desired?

buzzdecafe: sounds good, but *way* beyond the scope of this PR

TheLudd: Yes that is what I meant, merge as is and then look at assertions in another PR

src/match.js
buzzdecafe: i don't think we should. let's capture the existing behavior. you can always `filter` the result.

TheLudd: @buzzdecafe So change sig to `RegExp -> String -> [String | Undefined]`?

buzzdecafe: *sigh* 
yes, i think so.

TheLudd: I'll sigh together with you and change it =)

TheLudd:
I think I have addressed all comments now.

CrossEye:
LGTM

src/match.js
```diff
@@ -1,19 +1,24 @@
...
+var compose = require('./internal/compose');
+var defaultTo = require('./defaultTo');
 var invoker = require('./invoker');
 
 
 /**
- * Tests a regular expression against a String
+ * Tests a regular expression against a String. Note that this function
+ * will return an empty array when there are no matches. This differs
+ * from the standard JavaScript string function which returns null when
```
davidchambers: Let's replace

    the standard JavaScript string function

with

    [`String.prototype.match`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match)

TheLudd: done

src/match.js
```diff
@@ -1,19 +1,24 @@
...
+var defaultTo = require('./defaultTo');
 var invoker = require('./invoker');
 
 
 /**
- * Tests a regular expression against a String
+ * Tests a regular expression against a String. Note that this function
+ * will return an empty array when there are no matches. This differs
+ * from [`String.prototype.match`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match)
+ * returns null when there are no matches.
```
davidchambers: I think this line needs to begin with `which`. Also, should we replace `null` with <code>&#x60;null&#x60;</code>? I'm not sure what we do elsewhere.

TheLudd: Done

2015-07-08 21:53 GMT+02:00 David Chambers <notifications@github.com>:

davidchambers:
:deciduous_tree:

src/match.js
```diff
@@ -1,19 +1,24 @@
...
  * @param {String} str The string to match against
- * @return {Array} The list of matches, or null if no matches found.
- * @see R.invoker
+ * @return {Array} The list of matches.
  * @example
  *
  *      R.match(/([a-z]a)/g, 'bananas'); //=> ['ba', 'na', 'na']
  */
-module.exports = invoker(1, 'match');
+module.exports = _curry2(compose(defaultTo([]), invoker(1, 'match')));
```
anton-rudeshko: Am I right, that if someone evil will change resulting array then everybody will suffer because of shared default array pointer?

Example:
```js
R.match(/a/, '')[1] = 'fail';
R.match(/a/, ''); // [, 'fail']
```
Reproducing on Ramda site.

Sure, there is purity around all Ramda, but it is not guaranteed in the wild and someone could get a hard pain in the ass debugging.

davidchambers: Good point. We should return a new array each time. Fancy opening a pull request?

anton-rudeshko: Okay, if you release minor version after merge =)

davidchambers: We could publish the change as v0.17.1.

anton-rudeshko: Sure, that's what I meant. See https://github.com/ramda/ramda/pull/1293

PR #1242, opened 2015-07-05
define R.isEmpty in terms of R.empty

davidchambers:
This pull request changes the meaning of `R.isEmpty(x)` from

> `x.length` is zero

to

> `x` is the empty value of its type

as discussed in #1236.

This pull request also updates `R.empty` to support Arguments.

### DIFF
--- a/src/empty.js
+++ b/src/empty.js
@@ -1,10 +1,15 @@
 var _curry1 = require('./internal/_curry1');
+var _isArguments = require('./internal/_isArguments');
+var _isArray = require('./internal/_isArray');
+var _isObject = require('./internal/_isObject');
+var _isString = require('./internal/_isString');
 
 
 /**
  * Returns the empty value of its argument's type. Ramda defines the empty
- * value of Array (`[]`), Object (`{}`), and String (`''`). Other types are
- * supported if they define `<Type>.empty` and/or `<Type>.prototype.empty`.
+ * value of Array (`[]`), Object (`{}`), String (`''`), and Arguments.
+ * Other types are supported if they define `<Type>.empty` and/or
+ * `<Type>.prototype.empty`.
  *
  * @func
  * @memberOf R
@@ -20,15 +25,20 @@ var _curry1 = require('./internal/_curry1');
  *      R.empty({x: 1, y: 2});  //=> {}
  */
 module.exports = _curry1(function empty(x) {
-  if (x != null && typeof x.empty === 'function') {
-    return x.empty();
-  } else if (x != null && typeof x.constructor != null && typeof x.constructor.empty === 'function') {
-    return x.constructor.empty();
-  } else {
-    switch (Object.prototype.toString.call(x)) {
-      case '[object Array]':  return [];
-      case '[object Object]': return {};
-      case '[object String]': return '';
-    }
-  }
+  return (
+    (x != null && typeof x.empty === 'function') ?
+      x.empty() :
+    (x != null && x.constructor != null && typeof x.constructor.empty === 'function') ?
+      x.constructor.empty() :
+    _isArray(x) ?
+      [] :
+    _isString(x) ?
+      '' :
+    _isObject(x) ?
+      {} :
+    _isArguments(x) ?
+      (function() { return arguments; }()) :
+    // else
+      void 0
+  );
 });
--- a/src/internal/_isArguments.js
+++ b/src/internal/_isArguments.js
@@ -1,0 +1,9 @@
+var _has = require('./_has');
+
+
+module.exports = (function() {
+  var toString = Object.prototype.toString;
+  return toString.call(arguments) === '[object Arguments]' ?
+    function _isArguments(x) { return toString.call(x) === '[object Arguments]'; } :
+    function _isArguments(x) { return _has('callee', x); };
+}());
--- a/src/internal/_isObject.js
+++ b/src/internal/_isObject.js
@@ -1,0 +1,3 @@
+module.exports = function _isObject(x) {
+  return Object.prototype.toString.call(x) === '[object Object]';
+};
--- a/src/isEmpty.js
+++ b/src/isEmpty.js
@@ -1,25 +1,28 @@
 var _curry1 = require('./internal/_curry1');
+var empty = require('./empty');
+var equals = require('./equals');
 
 
 /**
- * Reports whether the list has zero elements.
+ * Returns `true` if the given value is its type's empty value; `false`
+ * otherwise.
  *
  * @func
  * @memberOf R
  * @category Logic
- * @sig [a] -> Boolean
- * @param {Array} list
+ * @sig a -> Boolean
+ * @param {*} x
  * @return {Boolean}
+ * @see R.empty
  * @example
  *
  *      R.isEmpty([1, 2, 3]);   //=> false
  *      R.isEmpty([]);          //=> true
  *      R.isEmpty('');          //=> true
  *      R.isEmpty(null);        //=> false
- *      R.isEmpty(R.keys({}));  //=> true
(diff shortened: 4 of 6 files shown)


src/empty.js
```diff
@@ -26,9 +27,10 @@ module.exports = _curry1(function empty(x) {
     return x.constructor.empty();
   } else {
     switch (Object.prototype.toString.call(x)) {
-      case '[object Array]':  return [];
-      case '[object Object]': return {};
-      case '[object String]': return '';
+      case '[object Arguments]':  return (function() { return arguments; }());
```
megawac: Heads up, this isn't backwards compatible to IE7 where `toString(arguments) == '[object String]'` or something

megawac: PS, I would be in favour of dropping IE6&7 support

davidchambers: > I would be in favour of dropping IE6&7 support

Sounds good to me!

CrossEye: Anyone have a good sense of how much we have in the way of either ES3 or IE6/7 workarounds in the codebase now?

CrossEye:
:herb:, modulo discussions about IE6/7.

src/empty.js
davidchambers: Here's Underscore's [`_.isArguments` fallback][1]:

> ```javascript
> // Define a fallback version of the method in browsers (ahem, IE < 9), where
> // there isn't any inspectable "Arguments" type.
> if (!_.isArguments(arguments)) {
>   _.isArguments = function(obj) {
>     return _.has(obj, 'callee');
>   };
> }
> ```

If it's IE8 as well, we should probably provide the same fallback.


[1]: https://github.com/jashkenas/underscore/blob/1.8.3/underscore.js#L1224-L1230

davidchambers: I've added a fallback for old versions of IE.

src/empty.js
```diff
@@ -20,15 +25,20 @@ var _curry1 = require('./internal/_curry1');
...
+    _isString(x) ?
+      '' :
+    _isObject(x) ?
+      {} :
+    _isArguments(x) ?
+      (function() { return arguments; }()) :
+    // else
+      void 0
+  );
 });
```
CrossEye: I'm still mixed about these deeply nested conditional operators.  That's a lot of `? - :` pairs!

I'm not saying I dislike it.  In fact, I think I could grow to like it.  But I'm not used to doing more than two levels, and that's generally in a one-liner for a comparator:

    return a < b ?  -1 : a > b : 1 : 0;

It's certainly nice to deal with simple expressions here and not control-flow statements like `if-else` or `switch`.  Hmm.

CrossEye:
LGTM

buzzdecafe:
i understand this is required by the spec:

```js
(x != null && typeof x.empty === 'function') ?
      x.empty() :
    (x != null && typeof x.constructor != null && typeof x.constructor.empty === 'function') ?
      x.constructor.empty() :
```

but eeeww.

Nonetheless :cow2:

TheLudd:
You seem to be swinging back and forth on this =)

@CrossEye  
https://github.com/ramda/ramda/issues/1228#issuecomment-119372995
https://github.com/ramda/ramda/pull/1252#issuecomment-119709047

@buzzdecafe
https://github.com/ramda/ramda/issues/1228#issuecomment-115303714
https://github.com/ramda/ramda/issues/1228#issuecomment-119377780

I am still for the simpler definition and adding other methods for checking that non array values are empty or empty like.

CrossEye:
@TheLudd:

I have found this issue really thorny.  And I keep seeing the great points on every side of the debate.  My concern with an array-only version is the issue that started one of the many threads involved.

    isEmpty({})

really in any logical world would return true.  Having it throw, return `undefined`, or simply have undefined behavior is not a particularly useful experience.

So since I was rejecting the notion of a list-only API, I liked the idea that `isEmtpy()` is closely aligned with `empty()`.

I'm not thrilled with this choice, as it's further from the center of Ramda's ideas.  But I haven't seen anything better.

TheLudd:
My suggestion for `{}` which I don't think anyone has said either yes or no to is to define a method `hasProps` which I think is a better name for objects. `isEmpty` could be renamed to `arrayIsEmpty` or something similar if that will avoid confusion (I see no such need however). 

> I'm not thrilled with this choice, as it's further from the center of Ramda's ideas

I agree, therefore I say no. 

I actually think `isEmpty(null)` should evaluate to `false`. Why? Because `null` is not an empty array. And the documentation says that the function returns true if the array is empty. If someone reaches the conclusion that "isEmpty" is a bad name for this, then we could look at changing the name. 

If I want to check for `null/undefined` then I use `isNil`, and again if I want to do something if a value is either `null`, `undefined`, `[]` or `{}` I'd use `R.anyPass([ R.isNil, R.isEmpty, R.complement(hasProps) ])`

davidchambers:
I'd like `R.isEmpty(R.empty(x))` to evaluate true for every `x` for which `R.empty(x)` is not a type error. The names of the functions suggest this expression should never evaluate false.

> My suggestion for `{}` […] is to define a method `hasProps` which I think is a better name for objects. `isEmpty` could be renamed to `arrayIsEmpty` or something similar if that will avoid confusion […].

This is one option. I would not be surprised if `R.isEmptyArray(R.empty(x))` evaluated false for certain `x` values (such as `'foo'`).

But we don't need to define `R.isEmptyArray`. One can simply use `R.equals([])`. If `R.equals([])` is too specific, because it doesn't recognize Arguments objects, then `R.isEmptyArray` is also too specific. So we'd end up with `R.isEmptyArrayLike`. At that point using `R.propEq('length', 0)` instead seems an attractive option.

To determine whether a value is a particular empty value we can use `R.equals([])`, `R.equals({})`, `R.equals('')`, etc. To determine whether a value has length zero we can use `R.propEq('length', 0)`. We don't currently have a straightforward way to determine whether a value is the empty value of its type. If this would be useful, let's merge this pull request. If not, we should either rename the function `R.isEmptyArrayLike` or remove the function from the library.

davidchambers:
@paldepind? @raine?

pmcavoy89:
I've read through some of the issues around the `isEmpty()` controversy regarding the `null` and `undefined`. I see that @TheLudd is in favor for both to return as true as stated [here](https://github.com/ramda/ramda/pull/1242#issuecomment-123304930). 

Reading the documentation `Returns true if the given value is its type's empty value; false otherwise.` kinda is hard to read. Yes, I understand what it is saying but it is still strange and I did not get that the first time I read it. 

Reading the documentation about [null](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/null#Difference_between_null_and_undefined) kinda gives that information of why the choice was made. This just wasn't as clear to me.

CrossEye:
I feel that if the documentation is unclear, then that is what should be fixed, and we shouldn't reduce the functionality.  `isNil` is quite useful.  I wouldn't want to lose that.

pmcavoy89:
That's fair, it's just a little confusing with Javascript because of the way the language is.

Would it make more sense to update the documentation to; `Returns true if the given value is its type's empty value (uses typeof on primitive types); false otherwise.`

I feel something along those lines may help clear up the confusion?

davidchambers:
One should never apply `R.isEmpty` to `null` (or `undefined`). The function's strict type signature is:

```purescript
isEmpty :: Monoid m => m -> Boolean
```

`null` does not satisfy the constraint, so it is not a valid argument. We have several options:

  - throw;
  - arbitrarily return `true`;
  - arbitrarily return `false`; or
  - return some other arbitrary value such as `undefined` or `'nope'`.

I imagine throwing is out of the question. We needn't worry much about the handling of invalid arguments; `true`, `false`, `undefined`, and `'nope'` seem equally reasonable (or unreasonable) arbitrary values.

Issue #1241, opened 2015-07-05
Object tree filtering. How?


bySabi:
A newbie here. Thanks for "easy(compose(good,hard), R)" just a bad joke :-)

I have this object:

```
obj1 = {
    one: {
        type: Number
    },
    two: {
        type: Object
    },
    three: {
        type: "String"
    }
}
```

and need this: 
```
new_obj1 = {
    three: {
        type: "String"
    }
}
```

Say filter all object props who "type" prop value is a String

Right now I have this but not work, obviusly, that´s cause I write here :-).

```
sayX = function(x) { console.log('x: ' + x); };

new_obj1 = R.mapObj(
    R.compose(
        R.tap(sayX),
        R.filter(R.is(String)),
        R.prop("test")
    ), obj1);
```

Thanks

bySabi:
Answer my self:
```
new_obj1 = R.pickBy(
    R.compose(
        R.is(String),
        R.prop('type')
    ),
    obj1
)
```

now I´m 0.02€ less newbie :-)

CrossEye:
I'm glad you were able to solve it.  Two points:

  * There's a strange ill-parallel between `type: Number`, `type: Object`, and `type: "String"`.  If you were to add
  ```js
  four: {
      type: "Foo"
  }
  ```
  this would still be accepted by your filter.  Is that what's expected?  Or do you really want `type: String` (without the quotes?)

  * Your solution is fine.  There's nothing wrong with it.  If you're striving for a points-free solution for whatever reason, this will give it to you, just by creating a named function without including the `obj1` parameter.  But points-free can sometimes also obscure things.  This might be more easily stated as:
  ```js
  R.pickBy(
    x => x.type == "String",
    obj1
  )
  ```
  Even if you had to do this without the ES6 sugar (`function(x) {return x.type == "String";}`), this would still to my eyes be a cleaner solution than the original.  And with that sugar it makes a nice one-line function definition:
  ```js
  var myFilter = R.pickBy(x => x.type == "String");
  ```

bySabi:
Is correct. I store object name like strings. Then I parse arbitrary object who contain "type" has string and set "type" to correct reference with `window["Foo"]`, strange?, sure! newbie stuff here ...

PR #1243, opened 2015-07-06
add R.propIs and R.propSatisfies

davidchambers:
In #1241, @bySabi defined this function:

```javascript
R.compose(R.is(String), R.prop('type'))
```

In JavaScript it's (sadly) quite common to need to check a property's type. I suggest it's at least as common as checking a property's value. `R.propEq` gives us a convenient way to check a property's value. `R.propIs` would give us a convenient way to check a property's type. The code above could then be rewritten as:

```javascript
R.propIs('type', String)
```

`R.propIs(a, b, c)` will throw a TypeError if `c` is null/undefined, as we'll attempt `c[a]`. This may or may not be optimal, but it matches the current behaviour of `R.propEq`.

---

__2015-07-08 update__

I've added `R.propSatisfies` and changed the signatures of `R.propEq` and `R.propIs`.

```haskell
R.propEq        :: a -> String -> Object -> Boolean
R.propIs        :: Type -> String -> Object -> Boolean
R.propSatisfies :: (a -> Boolean) -> String -> {String: a} -> Boolean
```

The change to `R.propEq` is breaking so will warrant a `:warning:` in the 0.16.0 upgrade guide.

### DIFF
--- a/src/propIs.js
+++ b/src/propIs.js
@@ -1,0 +1,28 @@
+var _curry3 = require('./internal/_curry3');
+var is = require('./is');
+var propSatisfies = require('./propSatisfies');
+
+
+/**
+ * Returns `true` if the specified object property is of the given type;
+ * `false` otherwise.
+ *
+ * @func
+ * @memberOf R
+ * @category Type
+ * @sig Type -> String -> Object -> Boolean
+ * @param {Function} type
+ * @param {String} name
+ * @param {*} obj
+ * @return {Boolean}
+ * @see R.is
+ * @see R.propSatisfies
+ * @example
+ *
+ *      R.propIs(Number, 'x', {x: 1, y: 2});  //=> true
+ *      R.propIs(Number, 'x', {x: 'foo'});    //=> false
+ *      R.propIs(Number, 'x', {});            //=> false
+ */
+module.exports = _curry3(function propIs(type, name, obj) {
+  return propSatisfies(is(type), name, obj);
+});
--- a/src/find.js
+++ b/src/find.js
@@ -21,8 +21,8 @@ var _xfind = require('./internal/_xfind');
  * @example
  *
  *      var xs = [{a: 1}, {a: 2}, {a: 3}];
- *      R.find(R.propEq('a', 2))(xs); //=> {a: 2}
- *      R.find(R.propEq('a', 4))(xs); //=> undefined
+ *      R.find(R.propEq(2, 'a'))(xs); //=> {a: 2}
+ *      R.find(R.propEq(4, 'a'))(xs); //=> undefined
  */
 module.exports = _curry2(_dispatchable('find', _xfind, function find(fn, list) {
   var idx = 0;
--- a/src/findIndex.js
+++ b/src/findIndex.js
@@ -21,8 +21,8 @@ var _xfindIndex = require('./internal/_xfindIndex');
  * @example
  *
  *      var xs = [{a: 1}, {a: 2}, {a: 3}];
- *      R.findIndex(R.propEq('a', 2))(xs); //=> 1
- *      R.findIndex(R.propEq('a', 4))(xs); //=> -1
+ *      R.findIndex(R.propEq(2, 'a'))(xs); //=> 1
+ *      R.findIndex(R.propEq(4, 'a'))(xs); //=> -1
  */
 module.exports = _curry2(_dispatchable('findIndex', _xfindIndex, function findIndex(fn, list) {
   var idx = 0;
--- a/src/findLast.js
+++ b/src/findLast.js
@@ -21,8 +21,8 @@ var _xfindLast = require('./internal/_xfindLast');
  * @example
  *
  *      var xs = [{a: 1, b: 0}, {a:1, b: 1}];
- *      R.findLast(R.propEq('a', 1))(xs); //=> {a: 1, b: 1}
- *      R.findLast(R.propEq('a', 4))(xs); //=> undefined
+ *      R.findLast(R.propEq(1, 'a'))(xs); //=> {a: 1, b: 1}
+ *      R.findLast(R.propEq(4, 'a'))(xs); //=> undefined
  */
 module.exports = _curry2(_dispatchable('findLast', _xfindLast, function findLast(fn, list) {
   var idx = list.length - 1;
--- a/src/findLastIndex.js
+++ b/src/findLastIndex.js
@@ -21,8 +21,8 @@ var _xfindLastIndex = require('./internal/_xfindLastIndex');
  * @example
  *
  *      var xs = [{a: 1, b: 0}, {a:1, b: 1}];
- *      R.findLastIndex(R.propEq('a', 1))(xs); //=> 1
- *      R.findLastIndex(R.propEq('a', 4))(xs); //=> -1
+ *      R.findLastIndex(R.propEq(1, 'a'))(xs); //=> 1
+ *      R.findLastIndex(R.propEq(4, 'a'))(xs); //=> -1
  */
 module.exports = _curry2(_dispatchable('findLastIndex', _xfindLastIndex, function findLastIndex(fn, list) {
   var idx = list.length - 1;
--- a/src/match.js
+++ b/src/match.js
@@ -1,5 +1,5 @@
 var _curry2 = require('./internal/_curry2');
-var compose = require('./internal/compose');
+var compose = require('./compose');
 var defaultTo = require('./defaultTo');
 var invoker = require('./invoker');
 
--- a/src/propEq.js
+++ b/src/propEq.js
@@ -1,18 +1,22 @@
 var _curry3 = require('./internal/_curry3');
 var equals = require('./equals');
+var propSatisfies = require('./propSatisfies');
 
 
 /**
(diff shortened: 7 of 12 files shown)


CrossEye:
I'll need some time to mull over whether I think this is important enough to include.  But in the meantime, I think the example above is a bit misleading, simply because the original example used the name '`type`' for the field it's checking.  I think these might be clearer:

```js
R.propIs('name', String)
R.propIs('age', Number)
```

Then of course we get to the question of how this might work with Lenses or Paths.  But I'll leave those for after I've thought through the main question itself.

kedashoe:
Funny I always define `R.propIs`, however as a more general function

`String -> (v -> Boolean) -> {k: v} -> Boolean`

so you can write things like

`R.propIs('age', R.gt(R.__, 10))`

Reads kinda weird in this case though?

`R.propIs('age', R.is(Number))`

CrossEye:
@kedashoe: That strikes me as a more useful function, as awkward as that last example may read.

davidchambers:
Perhaps @kedashoe's `propIs` could be named `propSatisfies`. `propEq` and `propIs` could then be defined in terms of `propSatisfies`.

bySabi:
@davidchambers  `propIs` :+1: 
and more general `propSatisfies` or maybe `propTest`?

TheLudd:
Is this for objects or for objects/arrays? The documentation says:

> Determines whether the given property of an object has the specified type.

But the documentation for the key says `Number | String`. AFAIK objects only have string keys.

CrossEye:
@TheLudd: It would test the _value_, not the _key_.

```js
    var person = {
        name: 'Scott',
        age: 29  // again???
    }
    propIs('name', String, person); //=> true
    propIs('name', Number, person); //=> false
    propIs('age', Number, person); //=> true
```

TheLudd:
@CrossEye I understand that, I was just wondering what kind of input the function would support. I have been under the impression that ramda functions should only support one type of input but a few functions, like this one seems to support different types.

davidchambers:
@TheLudd, I copied `@param {Number|String}` from [__src/propEq.js__][1]. Perhaps that should be changed.


[1]: https://github.com/ramda/ramda/blob/v0.15.1/src/propEq.js#L13

CrossEye:
Ahh, the sig.  I hadn't even noticed it.  I don't remember the history of this, but I'm sure the original was meant to account for the fact that this will work:

    R.propEq(1, 'b', ['a', 'b', 'c']); //=> true

as well as this:

    R.propEq('1', 'b', ['a', 'b', 'c']); //=> true

because of coercion, even though the latter is logically the only truly accurate version in JS.

davidchambers:
I've updated the signature.

TheLudd:
@davidchambers Nice. And yes I think the signature for `propEq` should be changed as well.

CrossEye:
:herb:

TheLudd:
I just want to raise the issue here of argument order. Are we sure it is correct? I am not sure either way but it could also be:

```
var user = { ... };
R.propIs(Number, 'age', user)
```

The current order resembles `propEq` where prop name is first but not `propOr` where default value is first. If we were to implement `propSatisfies`, would we then put the callback function as first argument or second?

scott-christopher:
I think I agree with @TheLudd, that when partially applied `R.propIs(Number)` would probably be more reusable than `R.propIs('age')`.

CrossEye:
Agreed on the param order.  For `propEq` I remember staring blankly for a while as neither order seemed to have any great advantage.  Here it seems clearly better to put the type first, or the predicate if we go with `propSatisfies`.

davidchambers:
I've updated this pull request (see the updated description).

src/propIs.js
```diff
@@ -0,0 +1,28 @@
...
+ * @see R.propSatisfies
+ * @example
+ *
+ *      R.propIs('x', Number, {x: 1, y: 2});  //=> true
+ *      R.propIs('x', Number, {x: 'foo'});    //=> false
+ *      R.propIs('x', Number, {});            //=> false
+ */
+module.exports = _curry3(function propIs(type, name, obj) {
+  return propSatisfies(is(type), name, obj);
+});
```
davidchambers: It's a shame our composition functions don't support this:

                 +------+                 +-----------------+
    --- type --->|  is  |--- is(type) --->|                 |
                 +------+                 |                 |
                                          |                 |
    --- name ---------------------------->|  propSatisfies  |--- propSatisfies(is(type), name, obj) --->
                                          |                 |
                                          |                 |
    --- obj ----------------------------->|                 |
                                          +-----------------+

CrossEye: Well they do, unless we accept #1254.  :smile:

```js
useWith(propSatisfies, is, identity, identity);
```

davidchambers: `useWith` is cheating! :stuck_out_tongue_winking_eye:

I would like this to behave exactly as the diagram suggests it should:

```javascript
pipe(is, propSatisfies)
```

CrossEye: Actually, there's a lot to be said for that.  If we could just convince that recalcitrant so-and-so to give up his concerns about functions that don't report their arity correctly, then we could accomplish this in a pretty straightforward manner.  But there's a huge wrinkle.  How do we know how many parameters each intermediate function should consume?  If we can insist that it's just one, then this could be straightforward.

It dovetails a bit with that [gist][bu] created, although of course it's still for multiple functions, not just two.

And we could do away with that pesky `useWith` altogether!

  [bu]: https://gist.github.com/buzzdecafe/b00905f6487725173ff7

davidchambers: I was looking for that gist! I'd forgotten that it's a gist rather than a pull request, so couldn't find it.

> How do we know how many parameters each intermediate function should consume?

Each function would consume as many of the original arguments as are necessary to satisfy it. I see no ambiguity.

Composing unary functions:

              +-----+             +-----+                +-----+
    --- a --->|  f  |--- f(a) --->|  g  |--- g(f(a)) --->|  h  |--- h(g(f(a))) --->
              +-----+             +-----+                +-----+

Composing binary functions:

              +-----+
    --- a --->|     |
              |  f  |                +-----+
    --- b --->|     |--- f(a, b) --->|     |
              +-----+                |  g  |                      +-----+
    --- c -------------------------->|     |--- g(f(a, b), c) --->|     |
                                     +-----+                      |  h  |
    --- d ------------------------------------------------------->|     |--- h(g(f(a, b), c), d) --->
                                                                  +-----+

Composing a unary function, a binary function, and a ternary function:

              +-----+             +-----+
    --- a --->|  f  |--- f(a) --->|     |
              +-----+             |  g  |                   +-----+
    --- b ----------------------->|     |--- g(f(a), b) --->|     |
                                  +-----+                   |     |
    --- c ------------------------------------------------->|  h  |
                                                            |     |
    --- d ------------------------------------------------->|     |--- h(g(f(a), b), c, d) --->
                                                            +-----+

CrossEye: But your initial example takes `is`, a binary function, but wants to treat it like a unary function so that results in a unary predicate to pass into `propSatisfies`.  How do you reconcile that with your "Combining binary functions" example?

davidchambers: Ah, I see your point! One would need to write:

```javascript
pipe(type => is(type), propSatisfies)
```

CrossEye: Yes, and that becomes much less elegant, to my mind.

davidchambers: It is less elegant, yes, but as you rightly pointed out we can't know what the user intended. Also,

```javascript
pipe(type => is(type), propSatisfies)
```

is closer to the ideal than

```javascript
(type, name, obj) => propSatisfies(is(type), name, obj)
```

which is our only option as things stand.

CrossEye: > which is our only option as things stand.

Unless we count this:

    useWith(propSatisfies, is);

(and why is that cheating?)

But my question is, as we're talking about simplifying the APIs of `converge` and `useWith`, would we really need all the flexibility implied here?  Is there any simpler convention that we could assume?  Perhaps not, as I'm not coming up with anything right away, but it would be really nice if this was not too complex.

davidchambers: I don't consider the behaviour depicted in the diagrams above complex. In fact, this is how I assumed `pipe` to work when I first considered applying it to functions of more than one argument.

`useWith` is a solution in this case but it sidesteps the question of how `compose` and `pipe` should handle functions of more than one argument.

CrossEye: I just find this somewhat amusing, because the behaviour of `compose`/`pipe` seemed obvious when we wrote them, and it was fine until I tried to write `project`.  Then the missing features made me think a bit about how `compose`/`pipe` should work, but I couldn't think of any obvious way to fix them, and that was the introduction of `useWith`.  

This suggestion would mostly remove the need for `useWith`, which I think is great.

So the main thing I think about it when I consider this is how it takes the place of `useWith`.  But as flexible as `useWith` is, this suggestion is more flexible still.  I'm fascinated, but also noting how our recent trend has seemed to be sacrificing some flexibility in favor of simplicity, I'm simply a bit wary of extreme flexibility.

davidchambers: I believe the approach is simple as well as flexible. I'll work on a pull request this evening. :)

davidchambers: #1260

PR #1244, opened 2015-07-06
acknowledge non-numeric array properties in R.toString

### LINKED ISSUE #1091: R.toString should show non-numeric array properties
```javascript
> /x/.exec('xyz')
[ 'x', index: 0, input: 'xyz' ]
> R.toString(/x/.exec('xyz'))
'["x"]'
```

`eval(R.toString(/x/.exec('xyz')))` is currently not equivalent to `/x/.exec('xyz')`, which breaks the contract.

I believe `R.toString(/x/.exec('xyz'))` should evaluate to `'["x", "index": 0, "input": "xyz"]'`. The non-numeric properties would appear, alphabetically ordered, after the values of the numeric properties.

davidchambers:
Closes #1091

```javascript
var a = ['x'];
var b = /x/.exec('xyz');
```

`a` and `b` really are different values, so should have different string representations (`["x"]` and `'["x", "index": 0, "input": "xyz"]'` respectively).

`eval(R.toString(b))` throws a SyntaxError, but the executable representation is awful:

```javascript
'(function() { var list = ["x"]; list.index = 0; list.input = "xyz"; return list; }())'
```

### DIFF
--- a/src/internal/_toString.js
+++ b/src/internal/_toString.js
@@ -3,6 +3,8 @@ var _map = require('./_map');
 var _quote = require('./_quote');
 var _toISOString = require('./_toISOString');
 var keys = require('../keys');
+var reject = require('../reject');
+var test = require('../test');
 
 
 module.exports = function _toString(x, seen) {
@@ -11,11 +13,16 @@ module.exports = function _toString(x, seen) {
     return _contains(y, xs) ? '<Circular>' : _toString(y, xs);
   };
 
+  //  mapPairs :: (Object, [String]) -> [String]
+  var mapPairs = function(obj, keys) {
+    return _map(function(k) { return _quote(k) + ': ' + recur(obj[k]); }, keys.slice().sort());
+  };
+
   switch (Object.prototype.toString.call(x)) {
     case '[object Arguments]':
       return '(function() { return arguments; }(' + _map(recur, x).join(', ') + '))';
     case '[object Array]':
-      return '[' + _map(recur, x).join(', ') + ']';
+      return '[' + _map(recur, x).concat(mapPairs(x, reject(test(/^\d+$/), keys(x)))).join(', ') + ']';
     case '[object Boolean]':
       return typeof x === 'object' ? 'new Boolean(' + recur(x.valueOf()) + ')' : x.toString();
     case '[object Date]':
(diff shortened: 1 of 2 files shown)


paldepind:
This seems right. Readability should be a higher concern than `eval`ability.

CrossEye:
While it's a shame to lose the executable representation, I agree that readability is more important.  Perhaps there's a `fromString` function needed that would be able to recognize any such odd cases and handle them, defaulting to `eval` for anything else.

I would prefer a stronger visual separation of the non-numeric properties, though:

    ["x"; "index": 0, "input": "xyz"]
    ["x" | "index": 0, "input": "xyz"]
    ["x" ~ "index": 0, "input": "xyz"]

davidchambers:
> Perhaps there's a `fromString` function needed that would be able to recognize any such odd cases and handle them, defaulting to `eval` for anything else.

I like this idea.

> I would prefer a stronger visual separation of the non-numeric properties, though:
>
>     ["x"; "index": 0, "input": "xyz"]
>     ["x" | "index": 0, "input": "xyz"]
>     ["x" ~ "index": 0, "input": "xyz"]

My thinking is as follows:

1.  The string representation of `{'0': 'foo', '1': 'bar', '2': 'baz'}` is `'{"0": "foo", "1": "bar", "2": "baz"}'`.

2.  Arrays are very similar to objects in JavaScript. In fact, we could have chosen to represent `['foo', 'bar', 'baz']` as `["0": "foo", "1": "bar", "2": "baz"]`. This would naturally have resulted in `'["0": "x", "index": 0, "input": "xyz"]'` for the original example.

3.  The `[`…`]` implies an array, which in turn applies that any unlabelled values have incrementing property names starting with `'0'`: `'0'`, `'1'`, `'2'`, ….

I view `'["x", "index": 0, "input": "xyz"]'` as shorthand for `'["0": "x", "index": 0, "input": "xyz"]'`. Using comma (`,`) as the separator preserves the symmetry between array representations and object representations.

CrossEye:
>> I would prefer a stronger visual separation of the non-numeric properties, though

> My thinking is as follows: [ ... ]

That's fine.  My preference was not a strong one.  There's a part of me that would still prefer `'["0": "x", "index": 0, "input": "xyz"]'` for the original example, but that's not particularly loveable.

PR #1245, opened 2015-07-06
fix behaviour of R.take(0)

### LINKED ISSUE #1224: R.take(0) in place of a transducer is an infinite loop if it runs on an infinite iterator.
```js
function* gen() {
	yield 1;
  	yield* gen()
}

R.into([], R.take(0), gen());
```

davidchambers:
Fixes #1224

I was able to reduce the failure case: `R.into([], R.take(0), [1, 2, 3])` currently evaluates to `[1, 2, 3]` rather than `[]`. The fix is straightforward.


CrossEye:
Looks good except for the JSCS error.

davidchambers:
> except for the JSCS error

Oops. Fixed.

PR #1248, opened 2015-07-07
Making concat throw on all non-arrays

### LINKED ISSUE #1246: concat with null second argument has unexpected error
```javascript
> R.concat([1,2], null)
TypeError: can't concat object
```

For the record, this works fine:
```javascript
> R.concat([1,2], null)
[1,2]
```

CrossEye:
Fixes #1246

### DIFF
--- a/src/concat.js
+++ b/src/concat.js
@@ -2,7 +2,7 @@ var _concat = require('./internal/_concat');
 var _curry2 = require('./internal/_curry2');
 var _hasMethod = require('./internal/_hasMethod');
 var _isArray = require('./internal/_isArray');
-
+var toString = require('./toString');
 
 /**
  * Returns a new list consisting of the elements of the first list followed by the elements
@@ -25,12 +25,12 @@ var _isArray = require('./internal/_isArray');
  *      R.concat([4, 5, 6], [1, 2, 3]); //=> [4, 5, 6, 1, 2, 3]
  *      R.concat('ABC', 'DEF'); // 'ABCDEF'
  */
-module.exports = _curry2(function(set1, set2) {
-  if (_isArray(set2)) {
-    return _concat(set1, set2);
-  } else if (_hasMethod('concat', set1)) {
-    return set1.concat(set2);
+module.exports = _curry2(function(a, b) {
+  if (_isArray(a) && _isArray(b)) {
+    return _concat(a, b);
+  } else if (_hasMethod('concat', a)) {
+    return a.concat(b);
   } else {
-    throw new TypeError("can't concat " + typeof set1);
+    throw new TypeError('Cannot concat ' + toString(a) + ' and ' + toString(b));
   }
 });
--- a/test/concat.js
+++ b/test/concat.js
@@ -42,5 +42,7 @@ describe('concat', function() {
   });
   it('throws if not an array, String, or object with a concat method', function() {
     assert.throws(function() { return R.concat({}, {}); }, TypeError);
+    assert.throws(function() { return R.concat([1, 2], null); }, TypeError);
+    assert.throws(function() { return R.concat(null, [1, 2]); }, TypeError);
   });
 });
--- a/src/adjust.js
+++ b/src/adjust.js
@@ -1,5 +1,5 @@
-var _concat = require('./internal/_concat');
 var _curry3 = require('./internal/_curry3');
+var _slice = require('./internal/_slice');
 
 /**
  * Applies a function to the value at the given index of an array,
@@ -29,7 +29,7 @@ module.exports = _curry3(function(fn, idx, list) {
   }
   var start = idx < 0 ? list.length : 0;
   var _idx = start + idx;
-  var _list = _concat(list);
+  var _list = _slice(list);
   _list[_idx] = fn(list[_idx]);
   return _list;
 });
--- a/src/internal/_concat.js
+++ b/src/internal/_concat.js
@@ -10,8 +10,6 @@
  *      _concat([4, 5, 6], [1, 2, 3]); //=> [4, 5, 6, 1, 2, 3]
  */
 module.exports = function _concat(set1, set2) {
-  set1 = set1 || [];
-  set2 = set2 || [];
   var idx;
   var len1 = set1.length;
   var len2 = set2.length;


davidchambers:
While we're at it let's change the message from

```javascript
"can't concat " + typeof set1
```

to

```javascript
'Cannot concat ' + R.toString(set1) + ' and ' + R.toString(set2)
```

Also, why on earth are the parameters named `set1` and `set2`?

CrossEye:
Good idea.  I'll do it tomorrow, when I'm not falling asleep.  :smile:

CrossEye:
> Good idea. I'll do it tomorrow, when I'm not falling asleep. :smile:

Or the next day.

src/concat.js
```diff
@@ -25,12 +25,12 @@ var _isArray = require('./internal/_isArray');
  *      R.concat([4, 5, 6], [1, 2, 3]); //=> [4, 5, 6, 1, 2, 3]
  *      R.concat('ABC', 'DEF'); // 'ABCDEF'
  */
-module.exports = _curry2(function(set1, set2) {
-  if (_isArray(set2)) {
-    return _concat(set1, set2);
-  } else if (_hasMethod('concat', set1)) {
-    return set1.concat(set2);
+module.exports = _curry2(function(ls1, ls2) {
```
davidchambers: `ls1` and `ls2` look ugly to me. I'd prefer `a` and `b` (or `list1` and `list2`, to match the JSDoc metadata).

test/concat.js
```diff
@@ -42,5 +42,7 @@ describe('concat', function() {
   });
   it('throws if not an array, String, or object with a concat method', function() {
     assert.throws(function() { return R.concat({}, {}); }, TypeError);
+    assert.throws(function() { return R.concat([1, 2], null); }, TypeError);
+    assert.throws(function() { return R.concat(null, [1, 2]); }, TypeError);
```
davidchambers: Should we assert that the error message is as we expect?

src/concat.js
CrossEye: You're probably right.  I'll fix to `a` and `b` later today.

test/concat.js
CrossEye: I don't think so.  I think that's more of an implementation detail we should feel free to change without changing any public contract.

CrossEye:
:zap:

src/concat.js
```diff
@@ -25,12 +25,12 @@ var _isArray = require('./internal/_isArray');
...
  */
-module.exports = _curry2(function(set1, set2) {
-  if (_isArray(set2)) {
-    return _concat(set1, set2);
-  } else if (_hasMethod('concat', set1)) {
-    return set1.concat(set2);
+module.exports = _curry2(function(a, b) {
+  if (_isArray(a) && _isArray(b)) {
+    return _concat(a, b);
+  } else if (_hasMethod('concat', a)) {
```
davidchambers: I think we should check that both `a` and `b` have `concat` methods. If `a` has a `concat` method but `b` does not, they are not of the same type so we should throw a TypeError.

CrossEye: I think that would just be a false sense of security:

    concat("a", [3, 4, 5])
    concat(Add(7), Multiply(5))

davidchambers: If the argument is that we can't detect all invalid applications, so we shouldn't throw a descriptive exception even in cases where we *could*, we should remove the `_hasMethod('concat', a)` check. It's inconsistent with the argument.

I'm very much in favour of descriptive error messages, so my stance is that we should take reasonable steps to ensure that `a` and `b` support `concat` and are of the same type. Throwing sooner rather than later leads to shallower stack traces and better debugging experiences.

CrossEye: What mechanism would you use to check too that 

    concat(AddInt(4), AddNum(Math.PI))

should throw an error, but

    concat(AddNum(Math.PI), AddInt(4))

is perfectly legitimate because `AddInt` is a subclass of `AddNum` but not the reverse?

davidchambers: This implementation passes all the tests on master:

```javascript
module.exports = flip(invoker(1, 'concat'));
```

So if all we want to do is handle valid input we needn't do anything more.

The fact that we instead choose to include a number of checks suggests that we're interested in validating the input. I just don't understand your algorithm for determining which *detectable* contract breaches to detect and which to ignore.

```javascript
R.concat(null, []);  // currently detected
R.concat([], null);  // not currently detected; this pull request adds detection
R.concat(null, '');  // currently detected
R.concat('', null);  // not currently detected; this pull request does not add detection
R.concat([], '');    // currently detected
R.concat('', []);    // not currently detected; this pull request does not add detection
```

I argue that we should either detect and report contract breaches consistently or remove the checks entirely. Our current checks—and the check added in this pull request—strike me as half measures.

CrossEye:
> I just don't understand your algorithm for determining which _detectable_ contract breaches to detect and which to ignore.

You may be right.  This was simply a fix for #1246, without a thought of rewriting the API besides that.

> I argue that we should either detect and report contract breaches consistently or remove the checks entirely. Our current checks—and the check added in this pull request—strike me as half measures.

Again, that may be right.  I hadn't really given it deep thought.  But this is not appropriate: 

> ensure that `a` and `b` support `concat` and are of the same type

as it reject some perfectly valid `concat` scenarios.

How do you suggest we proceed?  One possibility is to merge this as a minor improvement with a plan on figuring out a better overall implementation when we get a chance.

Another is to live with the bug from #1246 until we get around to an overhaul.

Is there a third?

Or do you have an implementation in mind?

davidchambers:
I'm in favour of this definition:

```javascript
module.exports = _curry2(function concat(a, b) {
  if (a != null && typeof a.concat === 'function' &&
      b != null && typeof b.concat === 'function' &&
      type(a) === type(b)) {
    return a.concat(b);
  } else {
    throw new TypeError('Cannot concat ' + toString(a) + ' and ' + toString(b));
  }
});
```

I could live with this one:

```javascript
module.exports = flip(invoker(1, 'concat'));
```

CrossEye:
I think `type(a) === type(b)` is too restrictive, for reasons [I describe above][co].

Moreover why should `typeof b.concat === 'function'` be relevant?

I might be able to forgo any error checking and go with your second definition, leaving it as _caveat emptor_.

  [co]: https://github.com/ramda/ramda/pull/1248#discussion_r34393993

davidchambers:
> I think `type(a) === type(b)` is too restrictive, for reasons [I describe above](https://github.com/ramda/ramda/pull/1248#discussion_r34393993).

`type(AddInt(4))` and `type(AddNum(Math.PI))` would both evaluate to `'Object'`, surely? This check is to disallow `R.concat('', [])`, which is currently permitted.

> Moreover why should `typeof b.concat === 'function'` be relevant?

If `a` has `concat` and `b` does not, they cannot be of the same type.

> I might be able to forgo any error checking and go with your second definition, leaving it as _caveat emptor_.

This seems the best option, since we can't agree on which measure to take to validate the contract.

asaf-romano:
Appending the contents of a ```Set`` to an Array seems like a valid use case to me.

asaf-romano:
By the way, using the spread argument of ES6:
js```
[1, 2, 3, ..."abc"]; // 1,2,3,a,b,c
```

CrossEye:
> This check is to disallow `R.concat('', [])`, which is currently permitted.

Right, but also decides for me in building `MyListType` that my `concat` should not accept lists, and that seems unacceptable.

davidchambers:
Perhaps you're viewing `R.concat` as sugar for a method invocation (where the method could have any definition) whereas I'm seeing it as a function with the stricter requirements of FL's [`concat`][1].


[1]: https://github.com/fantasyland/fantasy-land#concat-method

CrossEye:
> Perhaps you're viewing `R.concat` as sugar for a method invocation (where the method could have any definition) 

I do want to include arbitrary dispatch here, but these cases are covered under the following as well.


> whereas I'm seeing it as a function with the stricter requirements of FL's [`concat`][fl].

If I have a semigroup, `S`, consisting of lists of Strings, implemented through JS arrays, this is a strict subgroup of the semigroup, `A`,  consisting of all arrays.  These are both monoids as well with `[]`, but that's not important to this.  So letting `s` be a member of `S` and `a` be a member of `A`.  Since s ∈ S ⊂  A, logically this holds:

    a.concat(s) ∈ A

If we were able to call the `concat` of `A` on `s`, call it `concat'`, then this would also be true 

    s.concat'(a) ∈ A

But there is no reason to assume that

   s.concat(a) ∈ S

And I think it would be reasonable to build a type that captured this behavior, accepting a list in its `concat` method and checking whether all elements are Strings before proceeding, throwing if they're not.  We would report its type as `Object`, but it should allow us to pass parameters that were `Array`s.


[fl]: https://github.com/fantasyland/fantasy-land#concat-method

CrossEye:
Closed by #1272

Issue #509, opened 2014-11-10
Rename some *With functions to *By


CrossEye:
A few months ago (#65) we decided on a naming convention: `groupBy`, `sortBy`, and in general, `*By` each use a unary function that generates a key to stand for the objects.  The key can be as simple as `get('foo')`, or as complex as desired.  But `unionWith`, `uniqWith`, and in general, `*With`, each use a binary function used to compare two elements in some way, usually as a predicate or in order to choose one of them.

I just noticed that we're not as consistent with this as I would like.  First of all is the unfortunate `useWith`, for which I have really tried to find a better name; suggestions are still welcome.

But I also noticed three others that are misnamed: `maxWith`, `minWith`, and `pickWith` should be `maxBy`, `minBy`, and `pickBy`.  There may be others.

Any objections to my running through the code with this filter in mind and fixing these for this consistency?  It would simply be a renaming of existing functions.  (Although I do see a use-case for adding a real `maxWith`, `minWith`.)

Issue #65, opened 2014-06-21
Consider changing `minWith` and `maxWith` APIs to be more like `sortBy`


CrossEye:
`sortBy` has a nicer API than do `minWith` and `maxWith`.

In `sortBy` we generate a key for the objects to sort, and the comparator to use is automatically created based on this key.  So we can do something as simple as

```
sortBy(prop('title'), list);
```

`minWith` and `maxWith`, by contrast, which don't even need to be sorted, require you to supply a comparator.

```
minWith(function(a, b) {return a.x - b.x;}, points)
```

This really should be

```
minWith(prop('x'), points);
```

Granted the current API offers some theoretic additional flexibility.  But I can't see any realistic cases that would take advantage of it.

If there are no objections, I'll do this over the weekend.

PR #1251, opened 2015-07-08
rename times to repeatWith and deprecate times

### LINKED ISSUE #1250: Rename R.times to R.repeatWith
I don't always use `times` or `repeat`, but when I do, I can never remember which is which (`repeat('x', 2)`)

Alternatively, deprecate `times` and have `repeat` take a function and just do `R.repeat(R.always(10))` if you need to

Alternatively, don't change anything :)

kedashoe:
Closes #1250

Another note: I had no idea the function passed to `times` (possibly soon to be `repeatWith`) is [passed the index](https://github.com/ramda/ramda/blob/master/src/times.js#L27) of the loop used to generate the resulting array. Do we actually want that? Basically mapping over a range from 0 to n, but quite unexpected to me, and easily enough done with `map(fn, range(0, n))`. `repeat` for example is built on `times`/`repeatWith` and uses `always` which ignores this argument.


paldepind:
> Do we actually want that? Basically mapping over a range from 0 to n, but quite unexpected to me, and easily enough done with map(fn, range(0, n)).

The documentation is very clear about it. If this is so easily done and the name so hard to remember then maybe we should just deprecate `times`? I would prefer that. A `repeatWith` that doesn't pass an index would only be useful with impure functions. Then I think it would be better to not have it at all so that one could use `R.forEach(fn, R.repeat(_, n)` instead to make the impurity apparent.

TheLudd:
I am against this. I often use `times` and I have no problem remembering the name and that it passes the index (which I basically always use). So perhaps @kedashoe is looking for an addition to the API and not a change?

On another note:
I thought it was a convention that functions ending in "with" received as its first argument a function with the signature `(a, a) -> Boolean`. That is not the case here.

scott-christopher:
> I thought it was a convention that functions ending in "with" received as its first argument a function with the signature (a, a) -> Boolean. That is not the case here.

The convention of marking a function with a `*With` suffix is just to label it as a more generic form of the common, specialised function, accepting a custom function to help produce the result. e.g. `zip` and `zipWith`.

TheLudd:
@scott-christopher Are you sure? From #509:

> A few months ago (#65) we decided on a naming convention: `groupBy`, `sortBy`, and in general, *By each use a unary function that generates a key to stand for the objects. The key can be as simple as `get('foo')`, or as complex as desired. But `unionWith`, `uniqWith`, and in general, *With, each use a binary function used to compare two elements in some way, usually as a predicate or in order to choose one of them.

scott-christopher:
@TheLudd: If you were previously referring to Ramda's internal conventions, then no, I was not across them.

If you're talking about general conventions in this space, then yes, `*With` is often used to label a generic function that accepts a function to help generate a result. There is no convention that I'm aware of that restricts `*With` suffixed functions to binary predicates.

TheLudd:
@scott-christopher I was referring to the ramda-specific convention. I am sure no such convention exists outside ramda. But if functions are to be added to ramda I think they should follow the ramda convention. And yes, `useWith` and `zipWith` currently breaks this convention.

CrossEye:
I think with this change, we can include `zipWith` as well:

> "and in general, `*With`, each use a binary function used to compare **or combine** two elements in some way, usually as a predicate or in order to choose one of them."

And I still don't like the name `useWith`!  Someday we'll find a good replacement.

davidchambers:
@kedashoe:

> [M]apping over a range from 0 to n [can be] easily enough done with `map(fn, range(0, n))`.

Indeed.

@paldepind:

> [M]aybe we should just deprecate `times`? I would prefer that. A `repeatWith` that doesn't pass an index would only be useful with impure functions.

I agree.

@TheLudd, I'm interested in your use cases. I'm surprised `map(fn, range(0, n))` is a common enough pattern to warrant a specialized function.

kedashoe:
Good points, I think `repeatWith` is not needed.

As to deprecating `times`, looking through some projects, I see a couple instances where I do exactly `map(fn, range(0, n))`, but there are other instances of mapping over other ranges. I wasn't aware of the index feature of `times`, it does seem a bit specialized and I think personally I like seeing the `map` and `range` in all instances vs the one case using `times` and the non 0 to n cases using map and range.

TheLudd:
@davidchambers I mostly use it in tests to generate a certain number of something.

`R.times` is the functional equivalence of a for loop running a specified number of times. That is perhaps not done very often but to me it is a programming "primitive" and I think it belongs in a functional "toolbelt" such as ramda.

kedashoe:
closing this, I'll leave the #1251 open for a bit, we can discuss deprecating `times` or other options there

PR #1252, opened 2015-07-08
Simplify the defenition of isEmpty

TheLudd:
This is my suggestion on how to improve `isEmpty`. 

The code examples now only show cases that match up with the description
of the function. This has been reflected in the tests as well.
Implementation is left unchanged.

Note: I left examples with strings because it has been argued that *a string is just a list of characters*. I am willing to accept this but then I also think that we should make sure that all list functions support strings.

### DIFF
--- a/test/isEmpty.js
+++ b/test/isEmpty.js
@@ -4,44 +4,25 @@ var R = require('..');
 
 
 describe('isEmpty', function() {
-  it('returns false for null', function() {
-    assert.strictEqual(R.isEmpty(null), false);
-  });
-
-  it('returns false for undefined', function() {
-    assert.strictEqual(R.isEmpty(undefined), false);
-  });
 
   it('returns true for empty string', function() {
     assert.strictEqual(R.isEmpty(''), true);
+  });
+
+  it('returns false for a non empty string', function() {
+    assert.strictEqual(R.isEmpty('f'), false);
   });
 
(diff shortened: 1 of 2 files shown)


test/isEmpty.js
```diff
@@ -1,47 +1,28 @@
 var assert = require('assert');
 
-var R = require('..');
+var isEmpty = require('..').isEmpty;
```
davidchambers: Why this change? In other test files we import `R`.

TheLudd: Just removing duplicates. I often hear people speak about how ugly it is to prefix all functions with `R.`. Since the `R` namespace was only used to access `isEmpty` I thought I could do that only once.

CrossEye: We used to do that for all our tests, and something drove us away from that.  I would rather keep the consistency, even if I can't recall the reason for it now.  I prefer it too, BTW, but find the consistency of the test suite more compelling.

CrossEye:
I'm still trying to think through all the implication of `isEmpty`.  If we decide to keep much the same code, I'm definitely in favor of these changes (except the dropped prefix.)  But there's a lot going on around this right now.

TheLudd:
Prefix restored.

davidchambers:
Closing since we decided to merge #1242 instead.

PR #1255, opened 2015-07-08
Performance boost (up to 10x)

jon49:
per https://gamealchemist.wordpress.com/2013/05/01/lets-get-those-javascript-arrays-to-work-fast/ `6. pre-allocate your arrays whenever you can.`. This is also done in lodash's code, for performance reasons. https://github.com/lodash/lodash/blob/master/lodash.src.js#L1553

### DIFF
--- a/src/internal/_map.js
+++ b/src/internal/_map.js
@@ -1,5 +1,5 @@
 module.exports = function _map(fn, list) {
-  var idx = 0, len = list.length, result = [];
+  var idx = 0, len = list.length, result = Array(len);
   while (idx < len) {
     result[idx] = fn(list[idx]);
     idx += 1;


CrossEye:
I have a dificult time thinking of performance tweaks without a demonstrated performance issue.

Have you been running into speed problems with Ramda's `map`?

There's clearly nothing wrong with this code, but is there something especially right with it?

jon49:
Just seemed like an easy and simple performance boost. Being a utility library performance should be one goal, not necessarily the main goal, of the library. That way it is abstracted away from the users of the library and then I don't have to think, is this code I'm using going to significantly reduce the code on intensive loops? I can just use it instead.

The main reason I even looked at it was because I was getting the wrong results with lodash since it was passing in the extra parameters.

I don't see why you would want the code to be intentionally slower when a simple fix can make it faster.

CrossEye:
I don't have a major objection.  But not having seen any problems, nor any demonstration besides a blog post (that warns against perf tests!), I don't have a great deal of faith that this won't in fact _hurt_ performance.  That's why the scepticism.  I'd like to see some sort of demonstration that this would have a positive impact.

megawac:
Last I checked (some time last year), there is a mixed bag to the perf gains from the `Array(len)` between JS vm implementations. Overall, I'd say it usually gives some win (its done in lodash and underscore) however, definitely not 10x. This used to be implemented via `Array(len)` (last time I touched this function at least).

jon49:
OK, got the code to test. I'm showing 5x performance on node 0.12.x. When I'm doing tons of loops all over the place I would say that is a significant improvement.

```js
function _mapNew(fn, list) {
    var idx = 0, len = list.length, result = Array(len);
    while (idx < len) {
        result[idx] = fn(list[idx]);
        idx += 1;
    }
    return result;
}
;
function _mapOriginal(fn, list) {
    var idx = 0, len = list.length, result = [];
    while (idx < len) {
        result[idx] = fn(list[idx]);
        idx += 1;
    }
    return result;
}
;
var range = function (count) {
    var idx = -1, arr = Array(count);
    while (++idx < count) {
        arr[idx] = idx;
    }
    return arr;
};
var arr = range(1e7);
var double = function (x) { return 2 * x; };
var ORIGINAL = 'mapOriginal';
var NEW = 'mapNew';
console.time(ORIGINAL);
_mapOriginal(double, arr);
console.timeEnd(ORIGINAL);
console.time(NEW);
_mapNew(double, arr);
console.timeEnd(NEW);
//# sourceMappingURL=index.js.map
```

jon49:
Here's the output:

```
mapOriginal: 227ms
mapNew: 46ms
```

jon49:
Depending on the order of the tests it changes between 2 to 5 times the original code. Not sure what I'm doing wrong in the tests to make it different.

jon49:
When I do the tests in isolation I get about a 2.77x improvement over the original code.

TheLudd:
A quick search shows that these files contain the code snippet `result = []`
 * src/mapAccumRight.js
 * src/dropRepeatsWith.js
 * src/mapAccum.js
 * src/xprod.js
 * src/splitEvery.js
 * src/uniqWith.js
 * src/unfold.js
 * src/internal/_map.js
 * src/internal/_makeFlat.js
 * src/internal/_concat.js
 * src/internal/_filter.js
 * src/internal/_filterIndexed.js
 * src/range.js

If this change goes through, perhaps all these should be changed?

jon49:
If you are OK with the change I have no problem going through each file and changing it. I just don't want to do the work if it isn't accepted.

TheLudd:
I personally will probably not have a say in whether this stays or goes =)

megawac:
>

   - src/mapAccumRight.js
   - src/dropRepeatsWith.js
   - src/mapAccum.js
   - src/xprod.js
   - src/splitEvery.js
   - src/uniqWith.js
   - src/unfold.js
   - src/internal/_map.js
   - src/internal/_makeFlat.js
   - src/internal/_concat.js
   - src/internal/_filter.js
   - src/internal/_filterIndexed.js
   - src/range.js

This change can't be applied to most of those methods; only range and map
most likely

megawac:
+1, however note this may be slower on small arrays

jon49:
I tested smaller arrays and it still seems to be faster, although not as much as before; assuming I tested it right.

```js
function _mapNew(fn, list) {
  var idx = 0, len = list.length, result = Array(len);
  while (idx < len) {
    result[idx] = fn(list[idx]);
    idx += 1
  }
  return result;
};

function _mapOriginal(fn, list) {
  var idx = 0, len = list.length, result = [];
  while (idx < len) {
    result[idx] = fn(list[idx]);
    idx += 1;
  }
  return result;
};

var range = function(count, cb) {
    var idx = -1, arr = Array(count)
    while (++idx < count) {
        arr[idx] = idx
    }
    cb(arr)
}

var double = x => 2*x

const ORIGINAL = 'mapOriginal'
const NEW = 'mapNew'

range(1, arr => {

    const LOOP = 1e7

    var o = 0
    console.time(ORIGINAL)
    while (o < LOOP) {
        _mapOriginal(double, arr)
        o += 1
    }
    console.timeEnd(ORIGINAL)

    var n = 0
    console.time(NEW)
    while (n < LOOP) {
        _mapNew(double, arr)
        n += 1
    }
    console.timeEnd(NEW)


})
```

CrossEye:
I tested the first block and found a big improvement in V8: Chrome, Chromium and Node.  It made no difference at all in Firefox.  I didn't get a chance to check in IE, but assuming that it doesn't reduce performance there, I'd be in favor of this.

> If this change goes through, perhaps all these should be changed?

Agreed.

svozza:
Every time I see things like this I instantly thinking of Vyacheslav Egorov.

https://www.youtube.com/watch?v=65-RbBwZQdU

It's very likely you're not measuring what you think you are.

jon49:
OK, I watched the video. Yeah, who knows. From all the testing I've done it seems to be better. But I could be wrong. Assuming that it is correct when using node.js it could impact performance with high volume, but changing actual algorithms might make more of a difference rather than making a micro optimization.

I did change the test to reflect the new found knowledge from the video and it appears to still be faster. But, whatever you guys want to do, maybe I'm just wasting time, but it's nice to learn :-) .

```js
function _mapNew(fn, list) {
  var idx = 0, len = list.length, result = Array(len);
  while (idx < len) {
    result[idx] = fn(list[idx]);
    idx += 1
  }
  return result;
};

function _mapOriginal(fn, list) {
  var idx = 0, len = list.length, result = [];
  while (idx < len) {
    result[idx] = fn(list[idx]);
    idx += 1;
  }
  return result;
};

var range = function(count, cb) {
    var idx = -1, arr = Array(count)
    while (++idx < count) {
        arr[idx] = idx
    }
    cb(arr)
}

var double = x => 2*x

const ORIGINAL = 'mapOriginal'
const NEW = 'mapNew'

range(1e5, arr => {

    const LOOP = 200

    var orig = []
    var o = 0
    console.time(ORIGINAL)
    while (o < LOOP) {
        orig[o] = _mapOriginal(double, arr)
        o += 1
    }
    console.timeEnd(ORIGINAL)
    console.log(orig[10][1e5-1])

    var new_ = []
    var n = 0
    console.time(NEW)
    while (n < LOOP) {
        new_[n] = _mapNew(double, arr)
        n += 1
    }
    console.timeEnd(NEW)

    console.log(new_[10][1e5-1])

})
```

megawac:
This is better because the engine doesn't have to keep increasing the size of the array. I wouldn't consider this a micro opt, just some engines use different strategies for this case, causing different performance characteristics.

:+1: none the less

CrossEye:
My main reaction is that I have no objection to making this change, so long as we check that it doesn't actually cause harm in any important environments.  (IE anyone?)

It's not one that I care about, and I don't want to make a habit of making performance changes on spec.  I'd rather only do that once we recognize a real problem.  But now that this is out there, we might as well go ahead.

jon49:
Even though it is hard to figure out performance boosts this video talks about its importance. As a library it is probably more important to be performant but 99% of the applications it doesn't really matter that much. In other words, I agree, performance isn't as big of a concern as we (I) make it out to be.

https://www.youtube.com/watch?v=UJPdhx5zTaw

buzzdecafe:
LGTM :cow2:

PR #1256, opened 2015-07-08
@see annotations for head, tail, init and last family

kedashoe:
I knew ramda had `last` but wasn't sure what the complement was called. Figured nice to have links to the whole family

### DIFF
--- a/src/drop.js
+++ b/src/drop.js
@@ -8,11 +8,11 @@ var slice = require('./slice');
  * Returns a list containing all but the first `n` elements of the given `list`.
  *
  * Acts as a transducer if a transformer is given in list position.
- * @see R.transduce
  *
  * @func
  * @memberOf R
  * @category List
+ * @see R.transduce
  * @sig Number -> [a] -> [a]
  * @sig Number -> String -> String
  * @param {Number} n The number of elements of `xs` to skip.
--- a/src/head.js
+++ b/src/head.js
@@ -8,6 +8,7 @@ var nth = require('./nth');
  * @func
  * @memberOf R
  * @category List
+ * @see R.tail, R.init, R.last
  * @sig [a] -> a | Undefined
  * @param {Array} list The array to consider.
  * @return {*} The first element of the list, or `undefined` if the list is empty.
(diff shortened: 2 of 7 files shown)


davidchambers:
:deciduous_tree: Good idea!

CrossEye:
Will JSDoc let us combine these into a single line?

```
/*
 * @see tail, init, last
 */
```

kedashoe:
> Will JSDoc let us combine these into a single line?

I was going to write it like that but only saw the precedent in ramda for multiple lines

https://github.com/ramda/ramda/blob/master/src/transduce.js#L27-L40
https://github.com/ramda/ramda/blob/master/src/internal/_predicateWrap.js#L10-L11
https://github.com/ramda/ramda/blob/master/src/reduced.js#L13-L14

However, even if it did parse them as a single value, we would probably just need to do some pre-processing before outputting multiple links ala

https://github.com/ramda/ramda.github.io/pull/15

CrossEye:
I haven't tried, but I would find it cleaner to have a single `@see` line in the source code.  If we don't have to do any mucking about with JSDoc to get it to work, I'd rather switch all occurrences of multiple `@see`s into combined lines.

It's not a big deal of course, either way.

kedashoe:
> Will JSDoc let us combine these into a single line?

As I also think that is a nicer format, I've submitted a PR to ramdocs to handle this. I'll update this PR if you guys like/merge that one.

CrossEye:
That one's been merged.

kedashoe:
> That one's been merged.

Cool, shall I update all instances of multi-line `@see`s (there aren't that many), or just the ones currently in this PR?

davidchambers:
My preference is to update the existing multi-line `@see`s in a separate pull request, merge that, then rebase this pull request. This will leave the clearest history. Does this work for you, @kedashoe?

kedashoe:
ah, I was just working on this. Decided to update the others as there only two. I also fixed the [drop](http://ramdajs.com/docs/#drop) @see. I updated the commit message as well to note the additional multi-line @see fixes. Is this okay?

davidchambers:
Looks good!

CrossEye:
:herb:

PR #1167, opened 2015-06-08
remove array length caching

davidchambers:
Closes #1165

PR #1257, opened 2015-07-08
Add `R.uniqBy`

bezhermoso:
I was converting some code using `lodash` into using `ramda` instead and I wasn't able to find a straight alternative to `_.uniqBy`. 

I was able to compose a comparator in a point-free way, but my best attempt was just an unwieldy composition using `uniqWith`, `converge`, `equals`, `nthArg`, and `compose`:

```javascript
_.uniqBy(_.property("id")); // lodash

R.uniqWith(R.converge(R.equals, R.compose(R.prop("id"), R.nthArg(0))), R.compose(R.prop("id"), R.nthArg(1)))
```

I'm totally not discounting the possibility that there is a way to do this more concisely in a point-free manner, as I'm rather new to functional paradigms. If it turns out that there *is* a more concise way, I would love to take a look at it as it will be a good study for me.

However I figured I just implement `R.uniqBy` in case its something that actually adds value to `ramda`:

A simple example:

```javascript
R.uniqBy(Math.abs)([1, -2, 3, 5, 2, -3, 5]); // [1, -2, 3, 5] 
```

### DIFF
--- a/src/uniqBy.js
+++ b/src/uniqBy.js
@@ -1,0 +1,35 @@
+var _contains = require('./internal/_contains');
+var _curry2 = require('./internal/_curry2');
+
+
+/**
+ * Returns a new list containing only one copy of each element in the original
+ * list, based upon the value returned by applying the supplied function to each
+ * list element. Prefers the first item if the supplied function produces the
+ * same value on two items. `R.equals` is used for comparison.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig (a, b) -> [a] -> [a]
+ * @param {Function} fn A function used to produce a value to use during comparisons.
+ * @param {Array} list The array to consider.
+ * @return {Array} The list of unique items.
+ * @example
+ *
+ *      R.uniqBy(Math.abs, [-1, -5, 2, 10, 1, 2]); //=> [-1, -5, 2, 10]
+ *      R.uniqBy(R.prop("id"), [{id: 1, value: 10 }, {id: 2, value: true}, {id: 1, value: "foo"}]); //=> [{id: 1, value: 10}, {id: 2, value: true}]
+ */
+module.exports = _curry2(function uniqBy(fn, list) {
+  var idx = 0, applied = [], result = [], appliedItem, item;
+  while (idx < list.length) {
+    item = list[idx];
+    appliedItem = fn(item);
+    if (!_contains(appliedItem, applied)) {
+      result.push(item);
+      applied.push(appliedItem);
+    }
+    idx += 1;
+  }
+  return result;
+});
--- a/test/uniqBy.js
+++ b/test/uniqBy.js
@@ -1,0 +1,43 @@
+var assert = require('assert');
+
+var R = require('..');
+
+
+describe('uniqBy', function() {
+
+  var objs = [
+    {i: 1, v: 1}, {i: 2, v: 2}, {i: 3, v: 3}, {i: 4, v: 4}, {i: 5, v: 5},
+    {i: 6, v: 6}, {i: 7, v: 7}, {i: 8, v: 8}, {i: 9, v: 9}, {i: 10, v: 10}
+  ];
+
+  var objs2 = [
+    {i: 1, v: 1}, {i: 2, v: 2}, {i: 2, v: 3}, {i: 4, v: 4}, {i: 5, v: 5},
+    {i: 6, v: 6}, {i: 6, v: 7}, {i: 6, v: 8}, {i: 9, v: 9}, {i: 5, v: 10}
+  ];
+
+  it('returns a set from any array (i.e. purges duplicate elements) based on predicate', function() {
+    assert.deepEqual(R.uniqBy(R.prop('i'), objs), objs);
+    assert.deepEqual(R.uniqBy(R.prop('i'), objs2), [{i: 1, v: 1}, {i: 2, v: 2}, {i: 4, v: 4}, {i: 5, v: 5}, {i: 6, v: 6}, {i: 9, v: 9}]);
+  });
+
+  it('keeps elements from the left', function() {
+    assert.deepEqual(R.uniqBy(Math.abs, [-1, 2, 4, 3, 1, 3]), [-1, 2, 4, 3]);
+  });
+
+  it('returns an empty array for an empty array', function() {
+    assert.deepEqual(R.uniqBy(R.prop('i'), []), []);
+  });
+
+  it('has R.equals semantics', function() {
+    function Just(x) {
+      this.value = x;
+    }
+    Just.prototype.equals = function(x) {
+      return x instanceof Just && R.equals(x.value, this.value);
+    };
+    assert.strictEqual(R.uniqBy(Math.abs, [-1, 1]).length, 1);
+    assert.strictEqual(R.uniqBy(R.identity, [new Just([-1]), new Just([-1])]).length, 1);
+    assert.strictEqual(R.uniqBy(R.identity, [-0, 0]).length, 2);
+    assert.strictEqual(R.uniqBy(R.identity, [NaN, NaN]).length, 1);
+  });
+});


megawac:
:+1:

src/uniqBy.js
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Returns a new list containing only one copy of each element in the original list, based
+ * upon the value returned by applying the supplied predicate to each list element. Prefers
+ * the first item if the predicate produces the same value on two items. R.equals is used for comparison.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig (a, a -> Boolean) -> [a] -> [a]
```
davidchambers: I believe the signature should be `(a -> b) -> [a] -> [a]`.

src/uniqBy.js
```diff
@@ -0,0 +1,38 @@
+var _contains = require('./internal/_contains');
+var _curry2 = require('./internal/_curry2');
+
+
+/**
+ * Returns a new list containing only one copy of each element in the original list, based
+ * upon the value returned by applying the supplied predicate to each list element. Prefers
```
davidchambers: _Predicate_ implies `a -> Boolean`. I would use _function_ here instead.

test/uniqBy.js
```diff
@@ -0,0 +1,39 @@
...
+
+  it('returns an empty array for an empty array', function() {
+    assert.deepEqual(R.uniqBy(getI, []), []);
+  });
+
+  it('is curried', function() {
+    assert.strictEqual(typeof R.uniqBy(getI), 'function');
+    assert.deepEqual(R.uniqBy(getI)(objs), objs);
+    assert.deepEqual(R.uniqBy(getI)(objs2), [{i: 1, v: 1}, {i: 2, v: 2}, {i: 4, v: 4}, {i: 5, v: 5}, {i: 6, v: 6}, {i: 9, v: 9}]);
+  });
```
davidchambers: It's no longer necessary to assert that each function we add to the library is curried, as we assert this elsewhere for *all* Ramda functions. You could remove this block.

test/uniqBy.js
```diff
@@ -0,0 +1,39 @@
...
+  var objs2 = [
+    {i: 1, v: 1}, {i: 2, v: 2}, {i: 2, v: 3}, {i: 4, v: 4}, {i: 5, v: 5},
+    {i: 6, v: 6}, {i: 6, v: 7}, {i: 6, v: 8}, {i: 9, v: 9}, {i: 5, v: 10}
+  ];
+
+  var getI = R.prop('i');
+
+  it('returns a set from any array (i.e. purges duplicate elements) based on predicate', function() {
+    assert.deepEqual(R.uniqBy(getI, objs), objs);
+    console.log(R.uniqBy(getI, objs2));
```
davidchambers: Remove this?

test/uniqBy.js
```diff
@@ -0,0 +1,39 @@
...
+    {i: 1, v: 1}, {i: 2, v: 2}, {i: 3, v: 3}, {i: 4, v: 4}, {i: 5, v: 5},
+    {i: 6, v: 6}, {i: 7, v: 7}, {i: 8, v: 8}, {i: 9, v: 9}, {i: 10, v: 10}
+  ];
+
+  var objs2 = [
+    {i: 1, v: 1}, {i: 2, v: 2}, {i: 2, v: 3}, {i: 4, v: 4}, {i: 5, v: 5},
+    {i: 6, v: 6}, {i: 6, v: 7}, {i: 6, v: 8}, {i: 9, v: 9}, {i: 5, v: 10}
+  ];
+
+  var getI = R.prop('i');
```
davidchambers: I'd prefer to see `R.prop('i')` in place of `getI` wherever the latter is referenced. I believe tests should be as self-contained as possible, so a test can be read without one having to look for variable bindings in outer scopes.

bezhermoso:
@davidchambers I have applied the changes you asked.

bezhermoso:
@megawac Well, technically `uniqWith` can be used, but I imagine that using this is a much more convenient alternative. 

See my comment above regarding the use of 
`converge + equals + nArgs + <fn>` with `useWith` to write something point-free using existing functions. Other than that, you could write a comparator function to make it read better, but I posit that  `R.uniqBy(R.prop("id"))` is more convenient than `R.uniqWith(function(a, b) { return a.id === b.id; }` or `R.uniqWith(compareIds)`.

CrossEye:
:herb:

bezhermoso:
Another argument for `uniqBy` is that this may be more optimal if you are comparing against a computed value that could potentially be expensive:

```
R.uniqWith(function(a, b) {
     return a.computation () === b.computation();
});

vs

R.uniqBy(R.call('computation'));
``` 

The complexity for the using `useWith` increases as the list grows. In `uniqBy`, the computation will only be called once per item in list.

src/uniqBy.js
```diff
@@ -0,0 +1,38 @@
...
+ *      R.uniqBy(Math.abs)([1, 2, -5, 10, -1, 2]); //=> [1, 2, -5, 10]
+ *      R.uniqBy(R.prop("id"))([{id: 1, value: 10 }, {id: 2, value: true}, {id: 1, value: "foo"}]); //=> [{id: 1, value: 10}, {id: 2, value: true}]
+ */
+module.exports = _curry2(function uniqBy(fn, list) {
+  var idx = 0, len = list.length;
+  var applied = [], result = [], appliedItem, item;
+  while (idx < len) {
+
+    item = list[idx];
+    appliedItem = fn.call(null, item);
```
megawac: Avoid call. Do `fn(item)`

asaf-romano:
For your first example, couldn't you do ```R.uniqWith(R.propEq("id"))```

src/uniqBy.js
bezhermoso: Done.

src/uniqBy.js
```diff
@@ -0,0 +1,38 @@
+var _contains = require('./internal/_contains');
+var _curry2 = require('./internal/_curry2');
+
+
+/**
+ * Returns a new list containing only one copy of each element in the original list, based
+ * upon the value returned by applying the supplied function to each list element. Prefers
+ * the first item if the supplied function produces the same value on two items. `R.equals` is used for comparison.
```
davidchambers: Let's wrap the description at 79 characters, for those of us who like to `:set columns=80`. :)

src/uniqBy.js
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig (a, b) -> [a] -> [a]
```
davidchambers: Rather than `(a, b)` we should have `(a -> b)`, a function from type `a` to type `b`.

src/uniqBy.js
```diff
@@ -0,0 +1,38 @@
...
+ * @param {Function} fn A function used to produce a value to use during comparisons.
+ * @param {Array} list The array to consider.
+ * @return {Array} The list of unique items.
+ * @example
+ *
+ *      R.uniqBy(Math.abs)([1, 2, -5, 10, -1, 2]); //=> [1, 2, -5, 10]
+ *      R.uniqBy(R.prop("id"))([{id: 1, value: 10 }, {id: 2, value: true}, {id: 1, value: "foo"}]); //=> [{id: 1, value: 10}, {id: 2, value: true}]
+ */
+module.exports = _curry2(function uniqBy(fn, list) {
+  var idx = 0, len = list.length;
```
davidchambers: We prefer not to cache `list.length`.

src/uniqBy.js
```diff
@@ -0,0 +1,38 @@
...
+  var applied = [], result = [], appliedItem, item;
+  while (idx < len) {
+
+    item = list[idx];
+    appliedItem = fn(item);
+
+    if (!_contains(appliedItem, applied)) {
+      result[result.length] = item;
+      applied[applied.length] = appliedItem;
```
davidchambers: Let's use `Array.prototype.push` here:

```javascript
result.push(item);
applied.push(appliedItem);
```

src/uniqBy.js
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    idx += 1;
```
davidchambers: The blank line above and the other two blank lines (26 and 29) could be removed.

test/uniqBy.js
```diff
@@ -0,0 +1,30 @@
...
+  });
+
+  it('keeps elements from the left', function() {
+    assert.deepEqual(R.uniqBy(R.prop('i'), [{i: 1}, {i: 2}, {i: 3}, {i: 4}, {i: 1}]), [{i: 1}, {i: 2}, {i: 3}, {i: 4}]);
+  });
+
+  it('returns an empty array for an empty array', function() {
+    assert.deepEqual(R.uniqBy(R.prop('i'), []), []);
+  });
+});
```
davidchambers: Please add assertions for `R.equals` equality similar to [these][1].


[1]: https://github.com/ramda/ramda/blob/v0.15.1/test/contains.js#L19-L29

test/uniqBy.js
```diff
@@ -0,0 +1,30 @@
...
+    {i: 6, v: 6}, {i: 6, v: 7}, {i: 6, v: 8}, {i: 9, v: 9}, {i: 5, v: 10}
+  ];
+
+  it('returns a set from any array (i.e. purges duplicate elements) based on predicate', function() {
+    assert.deepEqual(R.uniqBy(R.prop('i'), objs), objs);
+    assert.deepEqual(R.uniqBy(R.prop('i'), objs2), [{i: 1, v: 1}, {i: 2, v: 2}, {i: 4, v: 4}, {i: 5, v: 5}, {i: 6, v: 6}, {i: 9, v: 9}]);
+  });
+
+  it('keeps elements from the left', function() {
+    assert.deepEqual(R.uniqBy(R.prop('i'), [{i: 1}, {i: 2}, {i: 3}, {i: 4}, {i: 1}]), [{i: 1}, {i: 2}, {i: 3}, {i: 4}]);
```
davidchambers: `Math.abs` would make a clearer example:

```javascript
assert.deepEqual(R.uniqBy(Math.abs, [1, 2, 3, 4, -1]), [1, 2, 3, 4]);
```

src/uniqBy.js
```diff
@@ -0,0 +1,38 @@
...
+ * @memberOf R
+ * @category List
+ * @sig (a, b) -> [a] -> [a]
+ * @param {Function} fn A function used to produce a value to use during comparisons.
+ * @param {Array} list The array to consider.
+ * @return {Array} The list of unique items.
+ * @example
+ *
+ *      R.uniqBy(Math.abs)([1, 2, -5, 10, -1, 2]); //=> [1, 2, -5, 10]
+ *      R.uniqBy(R.prop("id"))([{id: 1, value: 10 }, {id: 2, value: true}, {id: 1, value: "foo"}]); //=> [{id: 1, value: 10}, {id: 2, value: true}]
```
davidchambers: Let's use `R.uniqBy(a, b)` in these examples rather than `R.uniqBy(a)(b)`.

test/uniqBy.js
CrossEye: I agree with all the notes from @davidchambers except this one.  This assertion doesn't really match the code, but the assertion is a good one to demonstrate, and something like this might be better:

```js
R.uniqBy(R.prop('i'), [{i: 1, j: 10}, {i: 2, j: 20}, {i: 3, j: 30}, {i: 1, j: 50}, {i: 2, j: 2}]);
//=> [{i: 1, j: 10}, {i: 2, j: 20}, {i: 3, j: 30}]
```

davidchambers: What's your gripe with `Math.abs`, @CrossEye? If it's the fact that the output list appears to contain the results of the transformation, this could be addressed by changing the input list to something like `[-1, 3, 4, 1, 2]`.

CrossEye: No gripe at all, and I would prefer it.  You original example, though, didn't to my mind at all demonstrate that it kept the element from the left.  The updated one does, and with that, I agree that `Math.abs` is fine, and simpler than what's there.

src/uniqBy.js
bezhermoso: Any reason why `Array.prototype.push` is prefered? Honestly I had to check against `uniqWith` and its using this method of adding elements. I will do it, but just curious as to the reasoning and why its inconsistent across `ramda`.

bezhermoso: Is it because `arr.length` is expensive on bigger arrays?

src/uniqBy.js
megawac: I thought we changed our minds? we reverted to caching

src/uniqBy.js
megawac: any reason jscs isn't enforcing

src/uniqBy.js
davidchambers: My understanding was that we reverted out of fear that the length of an array could change during execution. Surely `fn` will not mutate `list`, in this case‽

megawac: I mean, I wouldn't but I could

src/uniqBy.js
davidchambers: It's an argument from readability and correctness. If one intends to append an element to a list, use the method which does exactly that. In the past we used the `list[idx] =` approach but this allowed at least one bug to sneak through which would not have been possible with `Array.prototype.push`.

src/uniqBy.js
davidchambers: I'm not opposed to empty lines. If there really are two sections, each should have its own paragraph, and paragraphs should be visually separated by one or more empty lines. I consider this function a single section (`idx += 1`, for example, does not warrant its own paragraph).

src/uniqBy.js
davidchambers: _Sigh_

You're right. We should include test cases for such atrocities if we consider relying on the initial length part of the contract, but I won't ask @bezhermoso to do so as part of this pull request.

test/uniqBy.js
```diff
@@ -0,0 +1,41 @@
...
+
+  it('has R.equals semantics', function() {
+    function Just(x) {
+      this.value = x;
+    }
+    Just.prototype.equals = function(x) {
+      return x instanceof Just && R.equals(x.value, this.value);
+    };
+    assert.strictEqual(R.uniqBy(Math.abs, [-1, 1]).length, 1);
+    assert.strictEqual(R.uniqBy(R.identity, [new Just([-1]), new Just([-1])]).length, 1);
```
bezhermoso: @davidchambers Would like to know if this test is sufficient.

src/uniqBy.js
bezhermoso: The only time the list length would change is if the supplied function is impure which somehow mutate the list. I think that would fall on the dev to *not* do that. (Changed it to not caching the length, anyway)

test/uniqBy.js
davidchambers: These look good. There are two we could add:

```javascript
assert.strictEqual(R.uniqBy(R.identity, [-0, 0]).length, 2);
assert.strictEqual(R.uniqBy(R.identity, [NaN, NaN]).length, 1);
```

bezhermoso: Done.

src/uniqBy.js
megawac: The argument for caching in #1167 was 3rd party code could be doing funky stuff and the cost for caching is very slight.

Further, its faster when supporting arbitrary data types which set length via `defineProperty`

```js
class Klass {
  get length() {
    return this.size;
  }
}
```

src/uniqBy.js
davidchambers: Could you address this issue, @bezhermoso?

test/uniqBy.js
davidchambers: Thanks. :)

I just noticed that `R.uniqBy(Math.abs, [-1, 1])` doesn't test differences between `R.equals(a, b)` and `a === b`, so we could remove that assertion.

davidchambers:
Superseded by #1271 (so we can include `R.uniqBy` in today's release)

PR #1260, opened 2015-07-09
simplify and enhance composition functions

davidchambers:
Changes:

  - :warning: `R.composeP` and `R.pipeP` no longer perform magic to determine whether a function is a Promise-returning function. Such magic is antithetical to Ramda's philosophy, and prevented `R.composeP` and `R.pipeP` from having consistent return types.

  - :new: `R.compose`, `R.pipe`, `R.composeP`, and `R.pipeP` now return curried functions (`R.composeK` and `R.pipeK` are for unary functions only, so the resulting function needn't be curried).

  - <del>:new: <code>R.compose</code>, <code>R.pipe</code>, <code>R.composeP</code>, and <code>R.pipeP</code> now accept functions of arbitrary arities.</del>

  - `R.composeK`, and `R.pipeK` are now defined as folds with `R.identity` as the identity element. As a result, `R.pipeK()` is no longer an error.

  - Internal functions `_createComposer` and `_isThenable` are no longer required.


src/pipe.js
```diff
@@ -30,5 +34,11 @@ var reverse = require('./reverse');
  *      squareThenDoubleThenTriple(5); //=> 150
  */
 module.exports = function pipe() {
-  return compose.apply(this, reverse(arguments));
+  return reduce(function(f, g) {
+    return curryN(f.length + g.length - 1, function() {
+      return g.apply(this,
+                     prepend(f.apply(this, take(f.length, arguments)),
+                             drop(f.length, arguments)));
```
scott-christopher: A couple of variables might aid readability here, e.g.
```js
var fArgs = take(f.length, arguments),
    gArgs = drop(f.length, arguments);
return g.apply(this, prepend(f.apply(this, fArgs), gArgs));
```

scott-christopher:
I do prefer the new `pipeP` and `composeP`, though this is a potentially breaking change for them, given that they currently allow the first function to return a non-promise value, e.g. `pipeP(identity, promiseFn)(x) == promiseFn(x)`.

src/pipeP.js
```diff
@@ -1,40 +1,35 @@
...
+ *      //  followersForUser :: String -> Promise [User]
+ *      var followersForUser = R.pipeP(db.getUserById, db.getFollowers);
  */
 module.exports = function pipeP() {
-  return composeP.apply(this, reverse(arguments));
+  return reduce(function(f, g) {
+    return curryN(f.length + g.length - 1, function() {
+      return f.apply(this, take(f.length, arguments))
+             .then(pipe(prepend(__, drop(f.length, arguments)),
+                        apply(bind(g, this))));
```
scott-christopher: This might be a bit clearer with an explicit function, rather than point-free.
```js
return f.apply(this, take(f.length, arguments)).then(function (x) {
  return g.apply(this, prepend(x, drop(f.length, arguments)));
});
```

asaf-romano:
I understand the comment about composeP/pipeP's magic not following ramda's philosophy - which I'm a big fan of, generally speaking - but I have to admit that in the compose/pipe case, I'm always hoping for them to actually be merged one day. This P there always seems out of place.

Note that there are other options to make the *P versions follow ramda's philosophy better. The first I could think of is that composeP/pipeP could always return a Promise, even if no promise-returning function was passed in.

One concern I'm having with the separation between the promise and sync versions is that they are not mirrored for all the other composition functions. For instance, very often I find myself looking for convergeP.

CrossEye:
I like all of @scott-christopher's code suggestions, but I really need to think through the implications of this change at an API level before I worry to much about the implementation.

I think there might be thorns among those flowers.

src/pipeP.js
davidchambers: In the lambda both `this` and `arguments` will reference the wrong thing, so we'd need to add:

```javascript
var ctx = this;
var args = arguments;
```

We're really looking at:

```javascript
var ctx = this;
var args = arguments;
return f.apply(ctx, take(f.length, args)).then(function(x) {
  return g.apply(ctx, prepend(x, drop(f.length, args)));
});
```

This removes the need to import `bind` and `apply`, though, so I like the change. :)

src/pipe.js
davidchambers: Nice. I didn't like this idea at the time I wrote the code because `fArgs` is all the arguments for `f` whereas `gArgs` is all *but one* of the arguments for `g`. There's a lack of symmetry.

I can look at it another way, though. We're partitioning `arguments` into a list of arguments for `f` and a list of arguments for `g`. We don't mean to imply that these are *all* the arguments in either case (and the code makes it clear that this is not so).

CrossEye:
I don't know if there's something unintuitive about the proposed API, or if I'm being dense, but I'm not figuring out how to convert certain functions to it.  One example (somewhat altered to protect the embarrassed) is this:

```js
//sorter :: String -> Object -> ([Object] -> [Object])
//           env   ,  config -> object-sorter
var sorter = R.pipe(R.prop, R.prop('sortKey'), R.prop, R.sortBy);

var configs = {
  dev: {sortKey: 'age', otherProp: 'otherVal'},
  prod: {sortKey: 'name', otherProp: 'val'},
};

var objs = [
  {name: 'Fred', age: 28},
  {name: 'Wilma', age: 25},
  {name: 'Barney', age: 27},
  {name: 'Betty', age: 30}
];

sorter('dev', configs)(objs); //=> wilma, barney, fred, betty  (sorted by age (=configs.dev.sortKey))
sorter('prod', configs)(objs); //=> barney, betty, fred, wilma (sorted by name (=configs.prod.sortKey))
```

Is this straightforward to write in this new `pipe` and I'm just missing it?

src/compose.js
```diff
@@ -30,4 +30,6 @@ var _createComposer = require('./internal/_createComposer');
  *      //≅ triple(double(square(5)))
  *      squareThenDoubleThenTriple(5); //=> 150
  */
-module.exports = _createComposer(_compose);
+module.exports = function compose() {
+  return pipe.apply(this, reverse(arguments));
```
buzzdecafe: why favor `pipe` over `compose`? this seems backwards

src/pipe.js
```diff
@@ -30,5 +34,11 @@ var reverse = require('./reverse');
...
  */
 module.exports = function pipe() {
-  return compose.apply(this, reverse(arguments));
+  return reduce(function(f, g) {
+    return curryN(f.length + g.length - 1, function() {
+      var fArgs = take(f.length, arguments);
+      var gArgs = drop(f.length, arguments);
+      return g.apply(this, prepend(f.apply(this, fArgs), gArgs));
+    });
+  }, identity, arguments);
```
buzzdecafe: currying at every step of the iteration seems a bit excessive.

davidchambers:
You just need to replace the third occurrence of `R.prop` with `k => R.prop(k)`, because you don't want this function to consume any of the original arguments.

```javascript
// sorter :: String -> Map String { sortKey :: String } -> Object -> Object
const sorter =
R.pipe(R.prop,
       R.prop('sortKey'),
       k => R.prop(k),
       R.sortBy);
```

              +--------+
    --- a --->|        |
              |  prop  |                   +-----------------+                                    +--------------+                                          +----------+
    --- b --->|        |--- prop(a, b) --->| prop('sortKey') |--- prop('sortKey', prop(a, b)) --->| k => prop(k) |--- prop(prop('sortKey', prop(a, b))) --->|          |
              +--------+                   +-----------------+                                    +--------------+                                          |  sortBy  |
    --- c ------------------------------------------------------------------------------------------------------------------------------------------------->|          |--- sortBy(prop(prop('sortKey', prop(a, b))), c) --->
                                                                                                                                                            +----------+

src/pipe.js
megawac: Yeah, I'm concerned over the performance implications of this change (compose has always been pretty slow but this may be a dramatic hit)

src/compose.js
davidchambers: Decades of exposure to written English (with no countervailing exposure to a right-to-left language) has had an effect on my neural pathways: I now find it more natural to follow things that read left to right rather than right to left.

Natural (for me):

              +-----+
    --- a --->|     |
              |  f  |                +-----+
    --- b --->|     |--- f(a, b) --->|     |
              +-----+                |  g  |                      +-----+
    --- c -------------------------->|     |--- g(f(a, b), c) --->|     |
                                     +-----+                      |  h  |
    --- d ------------------------------------------------------->|     |--- h(g(f(a, b), c), d) --->
                                                                  +-----+

Unnatural (for me):

                                                                                    +-----+
                                                                                    |     |<--- a ---
                                                             +-----+                |  f  |
                                                             |     |<--- f(a, b) ---|     |<--- b ---
                                +-----+                      |  g  |                +-----+
                                |     |<--- g(f(a, b), c) ---|     |<-------------------------- c ---
                                |  h  |                      +-----+
    <--- h(g(f(a, b), c), d) ---|     |<------------------------------------------------------- d ---
                                +-----+

src/pipe.js
davidchambers: It was only by expressing composition as a fold that I was able to reveal its inherent elegance (and realize that we needn't disallow composition of zero functions). If we agree that the behaviour is correct I've no problem with someone providing an equivalent imperative implementation.

buzzdecafe: i don't object to the fold. i don't see why to curry inside every invocation of the accummulator. why not curry once after folding? that means you have to scan the list of functions to get the cumulative length, but that seems less onerous than re-currying every step. debugging through that stack trace will be insane

CrossEye:
Funny, that's what I tried first, and when it didn't work, I went down all sorts of esoteric paths, until I realized that my version of Node wasn't recognizing arrow functions.  Somehow I never tried it again when I switched to the more verbose format.

That is one strike against it to my mind, though.  While it doesn't look so bad as an arrow function, in the more verbose syntax, which some of us are stuck with in their day jobs (no Babel, no Traceur, etc. allowed!), this is much less appealing:

```js
var sorter = R.pipe(R.prop, 
                  R.prop('sortKey'),
                  function(name) {return R.prop(name);}, 
                  R.sortBy); 
```

That's certainly not enough to make me say no, but it is a blow.

CrossEye:
I really like this function, but I'm not sure it's a good replacement for `compose`/`pipe`, which have long-standing traditions in the community.  Our version may be slightly different than some in that they accept multiple parameters at the first step, but after that it does what all `compose` functions I know of do: pass the output of one function as the single input to the next one.  That seems a clean and elegant thing to do.

This proposal is powerful, interesting, and (perhaps eventually) easy to describe.  It would subsume `useWith` and offer additional abilities, but it is different from what people expect when they think of `compose`, and that bothers me.

davidchambers:
@CrossEye, you're viewing this change from a different vantage point than I am. I'm not suggesting this is a replacement for `R.useWith` or any other function. I'm looking at a use case which `R.pipe` does not support—non unary functions in positions beyond the first—and asking myself whether there's something reasonable we could do instead. The answer is yes. :)

> [Having to use `function(name) { return R.prop(name); }` in this case] is one strike against it to my mind, though.

You're expecting too much of this function. It's a simple function. It does no guesswork. It can't know what you intend.

I agree that `function(name) { return R.prop(name); }` reduces the appeal of `R.pipe` in this case. I don't consider this a strike against the new behaviour. I would write the function this way instead:

```javascript
var sorter = R.curry(function(key, configs, objs) {
  return R.sortBy(R.prop(R.prop('sortKey', R.prop(key, configs))), objs);
});
```

Here's a contrived example of a problem for which the new `R.pipe` is well suited:

```javascript
//  f :: Number -> Number -> Number -> Number
var f = R.pipe(Math.pow, R.subtract);

f(3, 4, 5);  // 3^4 - 5
```

CrossEye:
@davidchambers:
>  you're viewing this change from a different vantage point than I am. I'm not suggesting this is a replacement for `R.useWith` or any other function.

No, I'm simply noticing that this would happen to subsume `useWith` quite comfortably, which to me is a great bonus.

But you're proposing to change `compose`/`pipe`, perhaps the most central function(s) of the library, so I want to tread very carefully.  This is not just a tweak, either.  It is a radical API shift.  The new functionality overlaps a bit with the old, but there is a lot that doesn't overlap.

My biggest concern is that for the most common use of a function as a step in todays's `compose` -- as a unary function consuming the output of its predecessor -- the proposed version would require it to be wrapped as a lamba.

Imagine if someone proposed a newer, more flexible version of Haskell's `.` operator, but that it required changing this:

```haskell
process: Num -> Num
process = half .inc . square
```

into this

```haskell
process: Num -> Num
process = (\x -> half x) . (\x -> inc x) . square
```

Do you think there might be some negative reaction?  :smile:  This feels similar to me.


> I would write the function this way instead:
> 
> ```javascript
> var sorter = R.curry(function(key, configs, objs) {
>   return R.sortBy(R.prop(R.prop('sortKey', R.prop(key, configs))), objs);
> });
> ```

I think that would depend for me on how I intended to use it.  If I was only going to apply `key` and `config` in order to get a named function that I would run several times, then I would certainly prefer this:

```js
var sorter = R.pipe(R.prop, R.prop('sortKey'), R.prop, R.sortBy);
```


> Here's a contrived example of a problem for which the new R.pipe is well suited:
> ```js
> //  f :: Number -> Number -> Number -> Number
> var f = R.pipe(Math.pow, R.subtract);
> 
> f(3, 4, 5);  // 3^4 - 5
> ```

Yes, and this is why I like this function.  The only way to do something like that points-free with our existing code would involve a lot of mucking about with `nthArg` -- not pretty.  I'm just not so sure that it can reasonably become our `compose` function.

buzzdecafe:
> ```haskell
     process: Num -> Num
     process = (\x -> half x) . (\x -> inc x) . square

for my money, that holes this PR below the waterline -- at least as-is as a replacement for `compose`.   there's enough of interest here that we should continue this line of inquiry. But we gotta satisfy the definition of `compose`.

src/pipe.js
davidchambers: > why not curry once after folding?

Good idea! I've pushed an update. I think you'll like it!

src/internal/_pipe.js
```diff
@@ -0,0 +1,14 @@
+var _slice = require('./_slice');
+
+
+module.exports = function _pipe(_f, g) {
+  return {
+    length: _f.length + g.length - 1,
+    value: function() {
+      var args = _slice(arguments);
```
megawac: Any reason for this slice, seems unnecessary will end up just increasing the perf hit.

Honestly I think these slices should be implemented as `for`/`while`s because of hot they will be

src/internal/_pipe.js
```diff
@@ -0,0 +1,14 @@
+var _slice = require('./_slice');
+
+
+module.exports = function _pipe(_f, g) {
+  return {
+    length: _f.length + g.length - 1,
```
megawac: :+1:

src/internal/_pipe.js
davidchambers: > Any reason for this slice, seems unnecessary will end up just increasing the perf hit.

Are you suggesting `Array.prototype.slice.call(arguments, _f.length)` rather than `args.slice(_f.length)`?

> Honestly I think these slices should be implemented as `for`/`while`s because of hot they will be

Okay. We can do that.

davidchambers:
> > ```haskell
> > process: Num -> Num
> > process = (\x -> half x) . (\x -> inc x) . square
> > ```
>
> for my money, that holes this PR below the waterline -- at least as-is as a replacement for `compose`.   there's enough of interest here that we should continue this line of inquiry. But we gotta satisfy the definition of `compose`.

Okay. I'll update the patch and strike this line from the description:

> `R.compose`, `R.pipe`, `R.composeP`, and `R.pipeP` now accept functions of arbitrary arities.

I don't want the other improvements to be abandoned on account of an orthogonal issue.

src/internal/_pipe.js
megawac: > Any reason for this slice, seems unnecessary will end up just increasing the perf hit.

I was suggesting `_slice(arguments, 0, _f.length)` unless I missed something. Regardless, looping makes a lot of sense here

src/internal/_pipe.js
```diff
@@ -0,0 +1,15 @@
...
+      var gArgs = [];
+      var idx = 0;
+      while (idx < arguments.length) {
+        (idx < _f.length ? fArgs : gArgs).push(arguments[idx]);
+        idx += 1;
+      }
+      return g.apply(this, [_f.value.apply(this, fArgs)].concat(gArgs));
+    }
+  };
+};
```
davidchambers: How does this look, @megawac?

megawac: Hmm, I'd probably try to avoid the concat as well

megawac: easy way to do that would be to use `arr[arr.length]` style and leave `[0]` undefined for `gArgs` until you apply `_f`

CrossEye:
> Okay. I'll update the patch and strike this line from the description:

>> `R.compose`, `R.pipe`, `R.composeP`, and `R.pipeP` now accept functions of arbitrary arities.

> I don't want the other improvements to be abandoned on account of an orthogonal issue.

My problem is that to me this is still the same problem because `pipe` really is just `compose` with the order reversed.

I would be much happier having this function under a different name.

src/internal/_pipe.js
davidchambers: Instead I decided to `var gArgs = [null];` and later `gArgs[0] = _f.value.apply(this, fArgs);`.

CrossEye:
> I would be much happier having this function under a different name.

`combine`? `sequence`? `seq`? `flow`?

test/pipe.js
```diff
@@ -4,20 +4,41 @@ var R = require('..');
...
+    assert.deepEqual(R.pipe(g, f, f)(1, 2), [[[1, 2]]]);
+    assert.deepEqual(R.pipe(g, f, f)(1)(2), [[[1, 2]]]);
+    assert.deepEqual(R.pipe(h, f, f)(1, 2, 3), [[[1, 2, 3]]]);
+    assert.deepEqual(R.pipe(h, f, f)(1, 2)(3), [[[1, 2, 3]]]);
+    assert.deepEqual(R.pipe(h, f, f)(1)(2, 3), [[[1, 2, 3]]]);
+    assert.deepEqual(R.pipe(h, f, f)(1)(2)(3), [[[1, 2, 3]]]);
+
+    var $g = R.curry(g);
+
+    assert.strictEqual(R.pipe($g, $g).length, 2);
```
scott-christopher: Should the length of `R.pipe($g, $g)` be 3 rather than two?

src/pipe.js
```diff
@@ -1,34 +1,28 @@
...
+ *      var f = R.pipe(Math.pow, R.dec);
  *
- *      //≅ triple(double(square(5)))
- *      squareThenDoubleThenTriple(5); //=> 150
+ *      f(3, 4); // 3^4 - 1
  */
 module.exports = function pipe() {
-  return compose.apply(this, reverse(arguments));
+  return curryN(arguments.length === 0 ? 1 : arguments[0].length,
+                reduce(_pipe, {length: 1, value: identity}, arguments).value);
```
scott-christopher: Should the length passed to `curryN` be the length of the first function, or should we use the length returned by `reduce`?

```js
var composed = reduce(_pipe, {length: 1, value: identity}, arguments);
return curryN(composed.length, composed.value);
```

davidchambers:
If we're going to add the versions of `compose` and `pipe` initially proposed in this pull request it would be more consistent to name them `composeX` and `pipeX` (for some `X`).

Let's discuss the possibility of adding those functions in a separate thread, if someone is sufficiently excited by the prospect to open an issue. I'd like this thread to focus on this pull request's current form.

test/pipe.js
```diff
@@ -4,20 +4,41 @@ var R = require('..');
...
+    assert.deepEqual(R.pipe($g, $g)(1, 2)(3), [[1, 2], 3]);
+    assert.deepEqual(R.pipe($g, $g)(1)(2)(3), [[1, 2], 3]);
+  });
+
+  it('returns the identity function (the identity element of function composition) given no arguments', function() {
```
scott-christopher: Is there a use case for calling `pipe` with no arguments? It seems like this outcome might just be due to the implementation rather than something we want to explicitly test.

src/pipe.js
scott-christopher: We could also remove `identity` if there is a requirement for at least one function to be passed in.
```js
var composed = reduce(_pipe, {
    length: arguments[0].length,
    value: arguments[0]
  }, _slice(arguments, 1));
  return curryN(composed.length, composed.value);
```

test/pipe.js
davidchambers: Previously, yes, but I've removed support for functions of arbitrary arity, since it prevented composition of functions which return functions (such as `$g`).

test/pipe.js
davidchambers: There's no use case, but it is the correct behaviour, wouldn't you agree? Just as `R.sum([])` gives zero, the identity element of addition, `R.pipe()` should give the identity function, the identity element of function composition.

scott-christopher: If `pipe` were to accept an array of functions then I'd agree, as the array could be obtained from elsewhere though in this case the caller would have to explicitly call it with no functions. Meh, it's not a biggy for me either way.

src/pipe.js
davidchambers: I realized that we can't actually use `R.identity` as the initial value to `R.reduce`, because we allow the leftmost function to have arity greater than one. This suggests that Ramda's function composition has no identity element!

It's therefore necessary to add a special case for zero arguments, which is a bit ugly:

```javascript
return arguments.length === 0 ?
  identity :
  curryN(arguments[0].length, reduce(_pipe, arguments[0], tail(arguments)));
```

test/pipe.js
scott-christopher: Ok. I think I need to re-digest the conversation that lead to the removal.

src/pipe.js
scott-christopher: I'd personally be happy with requiring at least one function be passed to `pipe`.

test/pipe.js
davidchambers: The problem, as @CrossEye and @buzzdecafe rightly pointed out, is that it's perfectly valid to do something like this:

```javascript
> R.pipe(R.add, R.map)(10)([1, 2, 3])
[ 11, 12, 13 ]
```

The version of `pipe` I initially proposed did not allow this, which is a showstopper.

src/pipe.js
davidchambers: I think you're right. Treating the leftmost function specially means that `R.pipe(R.identity, f)` is not equivalent to `R.pipe(f, R.identity)` (which surely breaks some law). I'll have `R.pipe` and `R.compose` throw if given no arguments.

test/pipe.js
scott-christopher: Yep, that unfortunately makes sense.

src/pipe.js
CrossEye: > `R.pipe(R.identity, f)` is not equivalent to `R.pipe(f, R.identity)` (which surely breaks some law).

I think it would if we wanted to treat composition as a Monoid, which could be done if we insisted that all functions in a composition were unary, including the initial one.  This makes sense in Haskell, since functions are always unary.

This was an early decision that although we respected Haskell's unary-functions-only design, there was no reason to try to do anything like it with Ramda:  the language doesn't have the necessary auto-currying capability to support it, and it would simply be an artificial barrier.  I still think this is right; there's little practical benefit to composition-as-a-Monoid, but a lot of practical benefit to being able to construct various polyadic functions using Ramda's `compose`.

test/pipe.js
CrossEye: > ```js
> R.pipe(R.add, R.map)(10)([1, 2, 3])
> ```

Jeez, why is that I can't come up with these simple examples, and end up digging into my own poorly written code to demonstrate a point?!  :smile:

src/compose.js
```diff
@@ -1,33 +1,26 @@
...
  *
- *      var triple = function(x) { return x * 3; };
- *      var double = function(x) { return x * 2; };
- *      var square = function(x) { return x * x; };
- *      var squareThenDoubleThenTriple = R.compose(triple, double, square);
+ *      var f = R.compose(R.dec, Math.pow);
  *
- *      //≅ triple(double(square(5)))
- *      squareThenDoubleThenTriple(5); //=> 150
+ *      f(3, 4); // 3^4 - 1
```
CrossEye: This example is an improvement, but I'd like to capture that it can take more than two functions in the example.  Perhaps we could add back `double` and incorporate that?  `R.compose(R.dec, double, Math.pow)` ?

src/pipe.js
```diff
@@ -1,34 +1,31 @@
...
+ *      f(3, 4); // 3^4 - 1
  */
 module.exports = function pipe() {
-  return compose.apply(this, reverse(arguments));
+  if (arguments.length === 0) {
+    throw new Error('pipe requires at least one argument');
+  }
+  return curryN(arguments[0].length,
+                reduce(_pipe, arguments[0], tail(arguments)));
 };
```
CrossEye: I'm really glad to have a `reduce`-based compose again.  The insight that the compose I wanted _was_ a `reduce` felt very profound at the time.  How things change!

src/pipe.js
```diff
@@ -1,34 +1,31 @@
...
+ *
  * @func
  * @memberOf R
  * @category Function
- * @sig ((a... -> b), (b -> c), ..., (x -> y), (y -> z)) -> (a... -> z)
- * @param {...Function} functions A variable number of functions.
- * @return {Function} A new function which represents the result of calling each of the
- *         input `functions`, passing the result of each function call to the next, from
- *         left to right.
+ * @sig ((a -> b), (b -> c), ..., (y -> z)) -> (a -> b -> c -> ... -> y -> z)
```
CrossEye: None of the tests exercise this change in sig.  Is it still valid after your refactoring?

test/pipe.js
davidchambers: I find one-line examples instructive. You may have noticed that I've replaced long examples with short ones in a few places. As a reader of documentation this is what I prefer, but please speak up if you'd like to keep the long examples.

CrossEye: No, no, so long as it's clear, I usually prefer the short examples too.  I'm just lamenting that I wasn't easily able to find what, when I see yours, seems as though it should have been obvious.  Although I cleaned up the actual example that brought me up short, I didn't boil it down to its essence as you did here.

src/pipe.js
davidchambers: As I rode my bicycle home last night I tried to reconcile all these things which *almost* fit together:

  - the composition functions originally proposed compose functions of arbitrary arities…
  - these composition functions *do* have an identity element…
  - which means they can be defined as folds with no special case for zero arguments…
  - but these functions don't allow partial application of curried functions…
  - such as `R.pipe(add, R.map)(10, [1, 2, 3])`…
  - because curried functions report their length according to `f(a, b)` rather than `f(a)(b)`…
  - so could we make `R.add` et al. unary functions and have everything?

I was very excited until I realized that `R.pipe(Math.pow, R.subtract)` would then return a binary function rather than a ternary function. I went to bed thinking I had explored another avenue which doesn't offer a solution.

In writing this now, though, I see that this may not be a problem. We can't write `R.pipe(Math.pow, R.subtract)(3, 4, 5)` but we can write `R.pipe(Math.pow, R.subtract)(3, 4)(5)` or `R.pipe(Math.pow, R.subtract)(3)(4)(5)`. So it seems this really *does* work, unless I'm missing something!

Are we happy to report the length of `R.add` et al. as one? We'd be reporting the minimum number of arguments to which a function should be applied rather than the maximum. This is consistent with variadic functions which report their lengths as zero (rather than infinity).

src/pipe.js
davidchambers: This signature is intended to indicate that the return value is now a curried function.

src/compose.js
davidchambers: Good idea. I've updated the example to:

```javascript
var f = R.compose(R.inc, R.negate, Math.pow);

f(3, 4); // -(3^4) + 1
```

src/pipe.js
CrossEye: I'm nor sure what would be a reasonable sig for your original proposal.  The change I was noticing was that the resulting function now can have an arity greater than the original (unary-only?) first one, according to this sig.  This is a substantial change from current behavior, and there are no tests for it.

I haven't downloaded this version to play with it as I did the earlier one.  It didn't _look_ like this version actually did that, so I guess my question was a way of asking if I was reading the code correctly.

In any case, I think this sig is wrong.  In any version of `compose` we're likely to have the intermediate types in a string of binary functions (`b`, `c`, and `y` above) shouldn't show up in the resulting function.  

The older sig was somewhat misleading too.  That `a...` wasn't descriptive enough.  This might have been a better version:

    pipe :: ((a -> b -> ... -> n), (n -> o), (o -> p), ..., (x ->y), (y -> z)) -> (a -> b -> ... -> n -> z)

src/lens.js
```diff
@@ -25,7 +25,9 @@ var map = require('./map');
...
 module.exports = _curry2(function lens(getter, setter) {
-  return _curry2(function(f, s) {
-    return map(function(v) { return setter(v, s); }, f(getter(s)));
-  });
+  return function(f) {
+    return function(s) {
+      return map(function(v) { return setter(v, s); }, f(getter(s)));
+    };
+  };
 });
```
davidchambers: For reasons I don't yet understand it's necessary for a lens to be a unary function in order for lenses to compose.

davidchambers:
I've rebased this pull request. Could I get a "LGTM" from @buzzdecafe or @CrossEye?

src/pipe.js
davidchambers: Oops. The `a -> b -> c -> ... -> y -> z` was a brain fart. I'll fix it now. :)

davidchambers: Here's a slight modification which indicates that the leftmost function needn't be a curried function:

    (((a, b, ..., n) -> o), (o -> p), ..., (x -> y), (y -> z)) -> (a -> b -> ... -> n -> z)

davidchambers: :zap:

CrossEye: Yes, that looks right.  My question, I guess is whether after your changes, if that is _still_ the correct sig?  This PR has gone through some radical changes.  The original one radically altered the API.  Are we now entirely back to this API; is it now, at least as far as `compose` / `pipe` a matter of implementation?  Or are there substantial API differences too?

davidchambers: I've updated the pull request description to make it clear what the changes are now. The only API change for `compose` and `pipe` is that we now return a curried function.

CrossEye: Ok, thanks.  I'm getting lost in the activity.  `composes` and `lenses` and `curries`, oh my!

CrossEye:
:herb:

buzzdecafe:
i must say I am surprised, @CrossEye ! this PR will curry the output of `compose` -- something i've been trying to get in the lib for quite a while. better merge this one quick @davidchambers

davidchambers:
Perhaps I bamboozled him by making several composition-related changes at once, and changing the changes several times. ;)

paldepind:
The original proposal looks a lot like stack-based/concatenative programming – except that each function can only return a single value.

CrossEye:
Hey, I've been swindled!  :smile:


Actually, I knew that this was inevitable.  At some point, though, I will probably open a PR for still another version that will work with functions that don't report their arity consistently.

buzzdecafe:
can't have enough composition functions! that's what i always say

anton-rudeshko:
I'm a little bit lost in discussion. Is this fixes #736?

CrossEye:
Now, after #1308 is raised, I realize that this still has the same issues I was raising before:  http://bit.ly/1TX3xtP.  Uggh.  I had thought the only issues were the obscure ones of functions that didn't report their arity properly.  This makes me want to reconsider this.

kedashoe:
This actually got me last night as well. Any time the function that executes first in a `pipe`/`compose` takes optional arguments (like many standard library functions) you have to be careful.

davidchambers:
I'm confused by your comment, @CrossEye. `R.compose(someUnaryFunction, R.equals)` *should* return a curried binary function, should it not?

anton-rudeshko:
Just be careful with functions with optional args. You can just use `R.nAry` on first function. I think that this make sense.

CrossEye:
@davidchambers:

That's what, in the end, this change effectively ended up doing.  The original was a much bigger redesign, and I'm afraid I might have let down my guard.  I originally didn't want to make this change without adding a new alternative function that didn't automatically curry the result, for the reasons @kedashoe, @anton-rudeshko, and @pzavolinsky (in #1308) noted.  I do want the new behavior to be the default, but I think there is reasonable call for not currying here too.  I should have either immediately added (or asked you to add as part of this PR) a function that will allow the style of code that's now been broken by these changes.  I believe it's an important enough use case.

PR #1262, opened 2015-07-09
support strings in index and slice functions

### LINKED ISSUE #1199: R.init works on strings, R.tail doesn't

davidchambers:
Closes #1199

Changes:

  - :scroll: `R.nth`: reword description; add doctests; update behaviour
  - :scroll: `R.head`: reword description; add doctests
  - :scroll: `R.last`: reword description; add doctests
  - :scroll: `R.slice`: reword description
  - :scroll: `R.tail`: reword description; add doctests; update behaviour
  - :scroll: `R.init`: reword description; add doctests
  - :scroll: `R.take`: reword description
  - :scroll: `R.drop`: reword description
  - :floppy_disk: `R.nthChar`: deprecate, since `R.nth` is now polymorphic
  - :floppy_disk: `R.nthCharCode`: deprecate, since it has a niche role already served by `R.invoker`

### DIFF
--- a/src/drop.js
+++ b/src/drop.js
@@ -5,9 +5,8 @@ var slice = require('./slice');
 
 
 /**
- * Returns a list containing all but the first `n` elements of the given `list`.
- *
- * Acts as a transducer if a transformer is given in list position.
+ * Returns all but the first `n` elements of the given list, string, or
+ * transducer/transformer (or object with a `drop` method).
  *
  * @func
  * @memberOf R
@@ -15,9 +14,9 @@ var slice = require('./slice');
  * @see R.transduce
  * @sig Number -> [a] -> [a]
  * @sig Number -> String -> String
- * @param {Number} n The number of elements of `xs` to skip.
- * @param {Array} xs The collection to consider.
- * @return {Array}
+ * @param {Number} n
+ * @param {*} list
+ * @return {*}
  * @example
  *
  *      R.drop(1, ['foo', 'bar', 'baz']); //=> ['bar', 'baz']
--- a/src/head.js
+++ b/src/head.js
@@ -2,19 +2,23 @@ var nth = require('./nth');
 
 
 /**
- * Returns the first element in a list.
- * In some libraries this function is named `first`.
+ * Returns the first element of the given list or string. In some libraries
+ * this function is named `first`.
  *
  * @func
  * @memberOf R
  * @category List
  * @see R.tail, R.init, R.last
  * @sig [a] -> a | Undefined
- * @param {Array} list The array to consider.
- * @return {*} The first element of the list, or `undefined` if the list is empty.
+ * @sig String -> String
+ * @param {*} list
+ * @return {*}
  * @example
  *
  *      R.head(['fi', 'fo', 'fum']); //=> 'fi'
  *      R.head([]); //=> undefined
+ *
+ *      R.head('abc'); //=> 'a'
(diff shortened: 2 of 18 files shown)


src/drop.js
```diff
@@ -14,10 +14,9 @@ var slice = require('./slice');
  * @category List
  * @see R.transduce
  * @sig Number -> [a] -> [a]
- * @sig Number -> String -> String
- * @param {Number} n The number of elements of `xs` to skip.
- * @param {Array} xs The collection to consider.
- * @return {Array}
+ * @param {Number} n
+ * @param {*} coll
```
kedashoe: Maybe `list` instead of `coll` for these? Collection seems like a more general term to me that could include objects? This could be a useful distinction for ramda as there are operations that could make sense for one, the other, or both.

davidchambers: Works for me. :)

davidchambers: :zap:

paldepind:
I like this. Except that readers of the documentation will have to guess what an "ordered collection" is.

davidchambers:
> [R]eaders of the documentation will have to guess what an "ordered collection" is.

What do you suggest we state instead? I was reluctant to mention Array and String explicitly since several of these functions dispatch. Perhaps this is not a problem. Something similar to the following seems reasonable (for `R.take`).

_Returns the first `n` element of the given ordered collection, which must be an array, a string, or an object with a `take` method._

buzzdecafe:
> Returns the first n element of the given ordered collection, which must be an array, a string, **transducer/transformer**, or an object with a take method.

davidchambers:
Good catch, @buzzdecafe. I'll update these docstrings.

paldepind:
Yes. That would be better in my opinion.

davidchambers:
I've updated the descriptions.

src/tail.js
```diff
@@ -1,23 +1,28 @@
 var _checkForMethod = require('./internal/_checkForMethod');
-var _slice = require('./internal/_slice');
+var slice = require('./slice');
 
 
 /**
- * Returns all but the first element of a list. If the list provided has the `tail` method,
- * it will instead return `list.tail()`.
+ * Returns all but the first element of the given list or string (or object
+ * with a `tail` method).
```
davidchambers: I parenthesized the last subexpression to avoid ambiguity.

_Returns all but the first element of the given list, string, or object with a `tail` method._

In the sentence above it's unclear whether _with a `tail` method_ relates to _list, string, or object_ or just _object_.

src/head.js
```diff
@@ -2,19 +2,23 @@ var nth = require('./nth');
...
+ * this function is named `first`.
  *
  * @func
  * @memberOf R
  * @category List
  * @see R.tail, R.init, R.last
  * @sig [a] -> a | Undefined
- * @param {Array} list The array to consider.
- * @return {*} The first element of the list, or `undefined` if the list is empty.
+ * @sig String -> String
```
davidchambers: It's necessary to include a second signature since `R.head('')` gives `''` rather than `undefined`. Should we also provide explicit signatures for functions such as `R.tail` for which the list signature *is* appropriate?

It's worth noting that this would not be a problem if the signature were `R.head :: [a] -> Maybe a`. ;)

CrossEye:
LGTM

PR #1263, opened 2015-07-10
fix typo for evolve docs

davidchase:
Small PR adds missing `data` key and shows a key that does not get transformed.


CrossEye:
Perfect.  Thank you.

This will make it into the docs the next time we release.  If it seems urgent, we can do the same thing in the output docs.

PR #1264, opened 2015-07-10
deprecate R.times

### LINKED ISSUE #1250: Rename R.times to R.repeatWith

davidchambers:
Closes #1250

As discussed in #1251, `R.times(f, n)` is sugar for `R.map(f, R.range(0, n))`. In my view this sugar is not warranted.

As an aside, several of my colleagues have complained about the number of functions in the library. Although my position is that no one is being forced to use *all* the functions, we should remain on the lookout for functions which no longer pull their weight.

### DIFF
--- a/src/times.js
+++ b/src/times.js
@@ -15,6 +15,7 @@ var _curry2 = require('./internal/_curry2');
  * @param {Function} fn The function to invoke. Passed one argument, the current value of `n`.
  * @param {Number} n A value between `0` and `n - 1`. Increments after each function call.
  * @return {Array} An array containing the return values of all calls to `fn`.
+ * @deprecated since v0.18.0
  * @example
  *
  *      R.times(R.identity, 5); //=> [0, 1, 2, 3, 4]


CrossEye:
I have no major objection.  Anyone?  @TheLudd?

TheLudd:
Yes. I say it stays because `R.times(n)` is the functional variant of `for(var i = 0; i < n; i++)`. According to the article "The Philosopho of Ramda" by @CrossEye, ramda is a:

> simply a collection of functions designed to make it easier to work in the composable-function style

I see this as a part of that collection. 

@davidchambers 
> In my view this sugar is not warranted.

To me this is not a good argument, it more seems like an opinion that originates in the fact that you do not use this. Please correct me if I am wrong. But if I were to argue the same then ~70% of the lib should go.

> As an aside, several of my colleagues have complained about the number of functions in the library. 

Sounds a bit silly to me. I am the one pushing ramda in my team, and while I have heard people say that once you want to accomplish something it can be difficult to know if ramda already has a function for it, no one has stated that the library "has too many functions". What library or framework of ramdas size or bigger do anyone know completely? IMO it is not really possible as a mere user to know all about it. But ramda has good documentation and a nice community in gitter. I have on several occasions asked for a function there because I couldn't find it. I don't say that the lib should be reduced because of this.

Recently, two functions were added #1243 that we never use (that were admittedly "just sugar" by the author). That is not a reason for me to say no to them but I do find the arguments for the inclusion of some functions and the removal of others to sometimes contradict eachother.

davidchambers:
@TheLudd, you're absolutely right that how useful I consider a function is heavily biased by the sorts of problems I've solved with Ramda.

Yesterday I shared with @benperez an idea I've been considering lately. I believe it's instructive to view every Ramda function (and this does not just apply to Ramda functions) as falling somewhere on this two-dimensional plane:

                              useful
                                ▲
                                |
                                |
                                |
                                |
                                |
    poorly defined ◀︎------------+------------▶︎ well defined
                                |
                                |
                                |
                                |
                                |
                                ▼
                            not useful

Usefulness is a function of two factors: how often a function is used and how much it simplifies the expression of a certain idea.

A function is well defined if there is exactly one clear definition: the function must take *these* arguments of *these* types in *this* order and return a value of *this* type. To be at the very right of the square the function must even have exactly one obvious *name*.

Let's consider some functions which fall in the different quadrants, clockwise from the upper-right quadrant:

  - `R.map` is useful and well defined. I use this function a great deal, and I'm sure others do too. Without this function we'd be forced to express mapping operations as folds, which in many cases is not such a clear expression of the idea. This function is also well defined. It takes two arguments. Currying dictates that the function must come first. Most of us agree that the provided function should be applied to the value only. The function also has a well-established name.

  - `R.call` is well defined but not useful. I scarcely ever use this function. It's not all the way at the bottom of the square, though, as it does simplify certain expressions when it *is* useful. This function is well defined. It mirrors `R.apply`, another well-defined function. Its name is canonical as it's taken from `Function.prototype.call`.

  - Fortunately I don't see any functions which are poorly defined and not useful.

  - `R.isEmpty` is a terrific example of a function which is useful but poorly defined. We have made breaking changes to its behaviour on two occasions that I can recall, and we're currently discussing further changes in one thread and the possibility of removing the function in another. There's clearly not exactly one obvious definition. It's clearly a useful function, though, as the aforementioned threads have received much attention.

My goal is to get to the point where every Ramda function is in the right half of the square. I'm fine with the library containing a number of well-defined but not terribly useful functions. I'm *not* fine with the library containing a number of useful but poorly defined functions.

Where does `R.times` lie on the x-axis? I consider it somewhere slightly to the left of the origin, as the values to which the provided function is applied aren't well defined. We're not dealing with array indexes, so I see no particular reason for `0`, `1`, `2`, … rather than `1`, `2`, `3`, … or some other sequence. This suggests that perhaps the start value should be parameterized. If there are competing definitions we don't have a well-defined function.

Where does `R.times` fall on the y-axis? I've only used it once as far as I can recall (and later realized `R.map` would have been better), but this may be a reflection of the problem domains in which I've been using Ramda. Some people may use this function frequently. The other factor of usefulness is how much the function simplifies the expression of an idea. Not much, I argue. `R.times(f, n)` is not much clearer than `R.map(f, R.range(0, n))`.

CrossEye:
@TheLudd:
> Yes. I say it stays because `R.times(n)` is the functional variant of `for(var i = 0; i < n; i++)`. 

Don't say that like it's a good thing.  @buzzdecafe has made it clear that one of his goals is to never have to write a `for`-loop again, and while my aversion is maybe not quite as strong, I'm close.

To me loops should be used to accomplish something.  It's extremely rare that that "something" isn't covered by `map`, `find`, `filter`, or `reduce`.

This isn't to say that you're wrong to want to do so, but it's pretty far from my own experience with FP.

>> In my view this sugar is not warranted.

> To me this is not a good argument, it more seems like an opinion that originates in the fact that you do not use this. 

I think there's something there.  David is, however, continually on the lookout for things that can be pared down, and that's a very good thing.  The library is growing rather quickly.

There is a difficult call when it comes to syntactic sugar.  Some things just seem so clearly useful that it's a no-brainer: `pipe` as a reversed `compose`.  Others I wouldn't want to drop because the names are the clear ones: whether or not `head` can be implemented as `nth(1)`, the name `head` is clear to me and to anyone whose dealt with this style programming.  Others get more tricky:  Is there a good reason for `inc`, which doesn't, arguably, hold any sacred space in the community mind, and which is probably less clear than `add(1)`.  This one wins mostly because of its pairing with `dec`, where the equivalent is the much less pretty, `subtract(__, 1)`.  We haven't formulated a specific policy, but there is a clear sense that some instances of syntactic sugar are much more reasonable than others.

> while I have heard people say that once you want to accomplish something it can be difficult to know if ramda already has a function for it, no one has stated that the library "has too many functions".

There are different audiences with varying expectations.  Those who want Ramda to act as a drop-in replacement for Underscore/lodash are often surprised -- they seem almost hurt -- to find that we don't happen to include their favourite function and have no intention of doing so.  Such folks would probably like as large a library as possible.  Others, and perhaps these are more likely people coming to Javascript from more functional languages, really want to see the core functional primitives, written as abstractly as we can pragmatically do so, and little else.  And of course there's a huge spectrum of prior experiences and current projects that play into these expectations.

We also have to weigh some of this against the founders.  @buzzdecafe and I started this library as a personal playground, a learning experience.  It wasn't until someone else picked it up as the library for a class that we decided to blog about it.  But for us there's always a tension between making this library as useful to others as possible and making it as fun for ourselves as we can.  And now too there's a large community of contributors whose expectations need to be taken into account.

Deciding what to include and what not to include is a complex balance, not just on the two axes that David describes,  but a number of other more squishy factors.

Back to the specific point, can you describe more specifically _how_ you use `R.times`?  That might help explain to us why it's important to keep.  Or, if others could demonstrate equally good or better ways to achieve the same thing, that might help convince you that it's not so important.  I would try to do this myself, but I don't use `R.times` at all.

CrossEye:
:herb:

I do have one caveat.  I think we need to, sometime soon, figure out how to describe types.  We've had need for an Ordered type for sigs and we definitely need a Lens type that we can describe.  My guess is that our best bet will simply be to include another file parallel to `ReadMe.md` in the docs directory with type descriptions and figure out how to get JSDoc to link to it appropriately.

But however we do it, I think this is becoming critical.  The Lens type is too complex to just assume people are going to get it from some simple sigs.

davidchambers:
`times` is a good candidate for a ramda-contrib package.

TheLudd:
@davidchambers You explain your reasoning about how you view functions in a very elaborate way but I really feel that we are still at the same place that we were before. You don't think this is a useful function because you have almost never used it yourself. You also say that it is not much clearer than `R.map(f, R.range(0, n))`. As a user of `times` I disagree. I feel that the code where I use `times` would be less expressive. 

You also add `propIs` which, according to you, is just `R.compose(R.is(String), R.prop('type'))` *even* after also adding `propSatisfies` which is a more general version that could be used to create  the `propIs` function.

My conclusion of this is that when it comes to functions you need in your project then your opinion is that:
 * The library does not contain too many functions already
 * Sugar is warranted 

But I often feel that you argue the opposite when it comes to functions you do not use. My point here is that, just like you say, you are heavily biased on what should be included based on the kind of programming you do and that is not IMHO a good base on which to make these decisions.

TheLudd:
@CrossEye 

> Don't say that like it's a good thing! :smile: @buzzdecafe has made it clear that one of his goals is to never have to write a for-loop again, and while my aversion is maybe not quite as strong, I'm close.

I'm not sure if I misunderstand you or if you misunderstood me because it looks to me as if we are on the same side of the argument. I don't write for loops because I have ramda. `map`, `filter`, `find`, `reduce` and `times` takes care of the cases that would otherwise be for loops.

> Back to the specific point, can you describe more specifically how you use R.times? 

I use it in tests to generate a set of test data with a given size. We also use it in our front end to work with an interface consisting of a grid. In that scenario there are several features where we go from a number (the number of rows or columns) and do an operation which takes the row or column index as its input.

TheLudd:
@davidchambers 

> times is a good candidate for a ramda-contrib package.

Sure, I could agree to that. But I think that I would want the contrib package to be up and available before the move. If so, would you also consider moving the mentioned prop-functions there since they are very specific (no other functions depend on them) and they were added quite late (the demand does not seem to be that big).

davidchambers:
How about a fusion of `map` and `range`?

```haskell
R.mapRange :: (Number -> a) -> Number -> Number -> [a]
```

```javascript
// Calculate the sum of the squares of [1, 100].
R.sum(R.map(x => x * x, 1, 101));
// => 338350
```

CrossEye:
>> Don't say that like it's a good thing! :smile: @buzzdecafe has made it clear that one of his goals is to never have to write a for-loop again...

> I'm not sure if I misunderstand you or if you misunderstood me because it looks to me as if we are on the same side of the argument. I don't write for loops because I have ramda. `map`, `filter`, `find`, `reduce` and `times` takes care of the cases that would otherwise be for loops.

Well, as far as I can tell, `times` is not for anything on its own.  It's just a loop with a different syntax, and that's what I was objecting to.  (My objection is not actually strong, though.)


> How about a fusion of `map` and `range`?

If I understand correctly, this might not do.  It sounds as though there's a reasonable chance this is being used for side-effects.

TheLudd:
@davidchambers 

> How about a fusion of map and range?

I would decline this as well even though it looks well defined. The reason I decline is that while the use of `times` is more rare that other loop functions, the use case where the starting index is not 0 is for me non existent. Add `mapRange` if there is demand for it. I would prefer to keep things as is and do something like `R.map(R.compose(myFn, R.add(1)), 100)` if I wanted indexes from 1 to 100. Not super pretty but AFAIK such a rare case that sugar is not warranted.

TheLudd:
@CrossEye Perhaps we have different stand points on when to use for loops. I have noticed that they are not used in the ramda source code in cases where they actually could be. That is fine but when I learned programming I was taught to use loops like this:

 * Use for loops when you know how many iterations you want to do
 * Use while loops when you don't, where the breaking of the iteration depends on some condition that may or may not change with each iteration.

Meaning for loops are used for lists, you want to do one iteration for each element (`map`, `filter` et al.). Or a for loop can be used when you don't have a list but just a number reflecting the iterations you want to do, this is `times`. 

Now, lets not get into whether the claim about when to use `for` or `while` is a good one, I think we can all agree that we want to use neither. I just wanted to clarify what I meant in earlier statements about `times` being the functional equivalent of `for(var i  = 0; i < n; i++)` where `n` would be the number supplied to `times`.

davidchambers:
I remain in favour of this change, but we should close this pull request if we're not going to merge it.

davidchambers:
I'll give you this one, @TheLudd. ;)

PR #1265, opened 2015-07-10
Adding omitBy; complement to pickBy

Bpless:


src/omitBy.js
```diff
@@ -0,0 +1,25 @@
...
+var pickBy = require('./pickBy');
+
+/**
+ * Returns a partial copy of an object omitting the keys that
+ * satisfy the supplied predicate.
+ *
+ * @func
+ * @memberOf R
+ * @category Object
+ * @sig [String] -> {String: *} -> {String: *}
```
davidchambers: The first argument should be something like `((a, String) -> Boolean)` rather than `[String]`.

src/omitBy.js
```diff
@@ -0,0 +1,25 @@
...
+ * @param {Function} pred A predicate to determine whether or not a key
+ *        should be excluded from the output object.
+ * @param {Object} obj The object to copy from
+ * @return {Object} A new object with excluded properties.
+ * @example
+ *
+ *      var isUpperCase = function(val, key) { return key.toUpperCase() === key; }
+ *      R.omitBy(isUpperCase, {a: 1, b: 2, A: 3, B: 4}); //=> {a: 1, b: 2}
```
davidchambers: This is an unusual example, as it ignores the value. I'd prefer something like this:

```javascript
var isPositive = function(n) { return n > 0; };
R.omitBy(isPositive, {a: -1, b: 0, c: 1, d: 2}); //=> {a: -1, b: 0}
```

davidchambers:
With #1260 this could be defined `R.pipe(R.complement, R.pickBy)`.

src/omitBy.js
Bpless: Fixed

src/omitBy.js
Bpless: Fixed

davidchambers:
I think `R.pickBy` and `R.omitBy` should only pass the value to the predicate.

I would then go a step further and support objects in `R.filter` (and by extension `R.reject`). We could then deprecate `R.pickBy` and `R.omitBy`. If this is off the table I'll work on a pull request for Sanctuary. :)

CrossEye:
> With #1260 this could be defined `R.pipe(R.complement, R.pickBy)`.

... which is slightly cleaner than `R.useWith(R.pickBy, R.complement, R.identity)`

CrossEye:
> I think `R.pickBy` and `R.omitBy` should only pass the value to the predicate.

I would support this if we create an equivalent to `addIndex` for these object functions.  I would like to have one in fact.

> I would then go a step further and support objects in `R.filter` (and by extension `R.reject`). We could then deprecate `R.pickBy` and `R.omitBy`.

I'm pretty strongly opposed.  Clearly `filter` and `pickBy` are related functions, but I don't think they're close enough to put in a single function and still match Ramda's insistence on simplicity.

davidchambers:
Thanks to #1429, `R.reject` can now operate on objects. For example:

```javascript
> R.reject(x => x % 2 === 0, {a: 1, b: 2, c: 3, d: 4})
{ a: 1, c: 3 }
```

I don't want to introduce `R.omitBy`, as I'd like to remove `R.pickBy` one day. Sorry, @Bpless! :trollface:

PR #1270, opened 2015-07-12
reduce the number of internal functions

davidchambers:
There are currently a small number of functions defined in __src/internal__ but never used. This pull request removes the unused files.

There are currently several cases where __src/internal/_foo.js__ exists only for __src/foo.js__ to be defined as `curry(_foo)`. This pull request coalesces each such pair of files.

This pull request takes additional reasonable measures to reduce the number of files in __src/internal__.


src/add.js
```diff
@@ -1,21 +1,19 @@
...
  * @memberOf R
  * @category Math
  * @sig Number -> Number -> Number
- * @sig String -> String -> String
- * @param {Number|String} a The first value.
- * @param {Number|String} b The second value.
- * @return {Number|String} The result of `a + b`.
+ * @param {Number} a
+ * @param {Number} b
+ * @return {Number}
```
davidchambers: Is this change controversial? I don't think we should document the fact that `R.add` happens to operate on strings: `R.concat` performs that function.

src/addIndex.js
```diff
@@ -38,6 +38,6 @@ module.exports = _curry1(function addIndex(fn) {
       return result;
     };
 
-    return fn.apply(this, _prepend(indexedFn, _slice(arguments, 1)));
+    return fn.apply(this, prepend(indexedFn, _slice(arguments, 1)));
```
megawac: This may be a perf regression

src/allPass.js
```diff
@@ -22,4 +22,4 @@ var curry = require('./curry');
  *      f(11); //=> false
  *      f(12); //=> true
  */
-module.exports = curry(_predicateWrap(_all));
+module.exports = curry(_predicateWrap(all));
```
megawac: `curry2`?

src/addIndex.js
davidchambers: How about this?

```javascript
var args = _slice(arguments);
args[0] = indexedFn;
return fn.apply(this, args);
```

src/allPass.js
megawac: Does this have to be curried at all?

src/anyPass.js
```diff
@@ -1,5 +1,5 @@
-var _any = require('./internal/_any');
```
megawac: IDK, I kind of think keeping internal versions of these make sense (for performance and future extensions)

src/insert.js
```diff
@@ -23,5 +21,7 @@ var _slice = require('./internal/_slice');
  */
 module.exports = _curry3(function insert(idx, elt, list) {
   idx = idx < list.length && idx >= 0 ? idx : list.length;
-  return _concat(_append(elt, _slice(list, 0, idx)), _slice(list, idx));
+  var result = _slice(list);
+  result.splice(idx, 0, elt);
+  return result;
```
davidchambers: Does this get the @megawac seal of approval? :)

megawac:
I'm concerned about the performance hits for functions like `reject`, `addIndex` and others which will now be double or triple unwrapping curried functions on each step

src/insert.js
megawac: stamped!

src/allPass.js
davidchambers: It does to satisfy one of the invariants added in #836. I've switched to `_curry1`, though.

src/addIndex.js
megawac: Meh, I'm still not a big fan of the `addIndex` functions (too hacky and can't be extended for `addKey`). I'd still rather see the `Index`/`Key` functions be the default and the none `key`/`index` functions just apply `unary` on the predicate

megawac: Actually, this change is fine as is, I didn't realize it's only applied once

CrossEye:
:100:

I'm very much in favor of this.  If there are perf regressions, I'm bothered only if they're substantial (not that I have a precise figure in mind.)  I love the code clean-up this entails.

src/allPass.js
CrossEye: Good point.

src/insert.js
CrossEye: I thought we tested that approach and the `concat` one was faster.  But it's only vague memories now, I'm afraid.

buzzdecafe:
it might be nice to capture some data on the cost so that someone can circle back later if necessary

src/insert.js
megawac: I'd be very surprised if that was the case

davidchambers:
@megawac, I've reinstated `_complement`. Removing that may be a step too far.

PR #1272, opened 2015-07-12
remove all type checking from R.concat

davidchambers:
Supersedes #1248

### DIFF
--- a/src/concat.js
+++ b/src/concat.js
@@ -1,23 +1,18 @@
-var _concat = require('./internal/_concat');
-var _curry2 = require('./internal/_curry2');
-var _hasMethod = require('./internal/_hasMethod');
-var _isArray = require('./internal/_isArray');
+var flip = require('./flip');
+var invoker = require('./invoker');
 
 
 /**
- * Returns a new list consisting of the elements of the first list followed by the elements
- * of the second.
+ * Returns the result of concatenating the given lists or strings.
  *
  * @func
  * @memberOf R
  * @category List
  * @sig [a] -> [a] -> [a]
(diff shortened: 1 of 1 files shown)


src/concat.js
```diff
@@ -25,12 +20,4 @@ var _isArray = require('./internal/_isArray');
...
-module.exports = _curry2(function concat(set1, set2) {
-  if (_isArray(set2)) {
-    return _concat(set1, set2);
-  } else if (_hasMethod('concat', set1)) {
-    return set1.concat(set2);
-  } else {
-    throw new TypeError("can't concat " + typeof set1);
-  }
-});
+module.exports = flip(invoker(1, 'concat'));
```
davidchambers: Shall we do this instead?

```javascript
module.exports = _curry2(function concat(a, b) { return a.concat(b); });
```

davidchambers:
((:bell:)) @CrossEye

CrossEye:
Somehow I missed this one when it came through.  Let me think about it a bit.  I'm not completely convinced that we _can't_ do reasonable type checking here.  I just thought that the approach you were suggesting in #1248 was too restrictive.  I'll look at it this evening.

CrossEye:
I guess I'm willing to try this and see if it causes issues.

I'm far from certain about it, though.  I really would rather stop some of the bad cases, but perhaps as we don't do that for any other Semigroups (how would we know if someone's `Maybe List` element could concatenate a `Maybe String`, or that the reverse is much less likely?) it just shouldn't bother me.

So I guess it's :herb: from me.  I'll close #1248, and we can always come back to this if it bites us.

PR #1273, opened 2015-07-12
add @see annotations for related functions

### LINKED ISSUE #1119: Function inverses in the documentation
I've spent a lot of time searching through the documentation looking for the opposite of a function that I have found in the documentation. 

For example, I've found the function `take` which is useful. But now I'm looking for the opposite function -- rather than take the first n, I'd like to `drop` the first n. It would be nice if there were notes in the documentation of these opposites. It would just be a nice touch so I dont have to search around as long.

davidchambers:
Closes #1119

### DIFF
--- a/src/F.js
+++ b/src/F.js
@@ -8,8 +8,8 @@ var always = require('./always');
  * @memberOf R
  * @category Function
  * @sig * -> false
- * @see R.always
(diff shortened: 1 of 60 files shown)


buzzdecafe:
:cow2:

PR #1274, opened 2015-07-12
mention "constant" in the description of R.always

### LINKED ISSUE #1038: Mention "constant" in the docs

davidchambers:
Closes #1038

### DIFF
--- a/src/always.js
+++ b/src/always.js
@@ -2,8 +2,11 @@ var _curry1 = require('./internal/_curry1');
 
 
 /**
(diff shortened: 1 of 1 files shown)


buzzdecafe:
:cow2:

PR #1275, opened 2015-07-12
support strings in R.reverse

davidchambers:
I hope this is not controversial. :)

### DIFF
--- a/src/reverse.js
+++ b/src/reverse.js
@@ -1,24 +1,31 @@
+var _asChars = require('./internal/_asChars');
 var _curry1 = require('./internal/_curry1');
 var _slice = require('./internal/_slice');
 
 
 /**
- * Returns a new list with the same elements as the original list, just
- * in the reverse order.
+ * Returns a new list or string with the elements or characters in reverse
+ * order.
  *
  * @func
  * @memberOf R
  * @category List
  * @sig [a] -> [a]
- * @param {Array} list The list to reverse.
- * @return {Array} A copy of the list in reverse order.
+ * @sig String -> String
+ * @param {Array|String} list
+ * @return {Array|String}
  * @example
  *
  *      R.reverse([1, 2, 3]);  //=> [3, 2, 1]
  *      R.reverse([1, 2]);     //=> [2, 1]
  *      R.reverse([1]);        //=> [1]
  *      R.reverse([]);         //=> []
+ *
+ *      R.reverse('abc');      //=> 'cba'
+ *      R.reverse('ab');       //=> 'ba'
+ *      R.reverse('a');        //=> 'a'
+ *      R.reverse('');         //=> ''
  */
-module.exports = _curry1(function reverse(list) {
+module.exports = _curry1(_asChars(function reverse(list) {
   return _slice(list).reverse();
-});
+}));
--- a/src/internal/_asChars.js
+++ b/src/internal/_asChars.js
@@ -1,0 +1,17 @@
+var _isString = require('./_isString');
+var _slice = require('./_slice');
+
+
+module.exports = function _asChars(f) {
+  return function() {
+    var idx = arguments.length - 1;
+    var xs = arguments[idx];
+    if (_isString(xs)) {
+      var args = _slice(arguments);
+      args[idx] = xs.split('');
+      return f.apply(this, args).join('');
+    } else {
+      return f.apply(this, arguments);
+    }
+  };
+};
--- a/test/reverse.js
+++ b/test/reverse.js
@@ -4,12 +4,19 @@ var R = require('..');
 
 
 describe('reverse', function() {
+
   it('reverses arrays', function() {
-    assert.deepEqual(R.reverse([1, 2, 3, 4]), [4, 3, 2, 1]);
+    assert.deepEqual(R.reverse([]), []);
+    assert.deepEqual(R.reverse([1]), [1]);
+    assert.deepEqual(R.reverse([1, 2]), [2, 1]);
+    assert.deepEqual(R.reverse([1, 2, 3]), [3, 2, 1]);
   });
 
-  it('returns the empty list to itself', function() {
-    assert.deepEqual(R.reverse([]), []);
+  it('reverses strings', function() {
+    assert.strictEqual(R.reverse(''), '');
+    assert.strictEqual(R.reverse('a'), 'a');
+    assert.strictEqual(R.reverse('ab'), 'ba');
+    assert.strictEqual(R.reverse('abc'), 'cba');
   });
 
 });


src/reverse.js
```diff
@@ -1,24 +1,33 @@
...
+ *      R.reverse('abc');      //=> 'cba'
+ *      R.reverse('ab');       //=> 'ba'
+ *      R.reverse('a');        //=> 'a'
+ *      R.reverse('');         //=> ''
  */
 module.exports = _curry1(function reverse(list) {
-  return _slice(list).reverse();
+  return _isString(list) ?
+    list.split('').reverse().join('') :
+    _slice(list).reverse();
```
buzzdecafe: `reduceRight(concat, empty(list), list)`

davidchambers: `reduceRight` doesn't accept a string as its third argument, does it?

buzzdecafe: http://bit.ly/1L9z0pC

davidchambers: I believe that relies on the environment supporting bracket notation for character access. It doesn't work for arrays, though, as one ends up doing `R.concat([], 3)` when reversing `[1, 2, 3]`. This is currently a type error, but interestingly it will work if and when #1272 is merged. Relying on `[] ++ 3` working would feel icky, of course!

buzzdecafe: yeah that's unfortunate. `reduceRight` should really be a dual of `reduce` with its tranducer-y nature

davidchambers:
While I'd love someone to suggest an elegant implementation, do we agree on the behaviour change? If so, we shouldn't let the lack of an elegant implementation prevent this function from behaving as desired.

TheLudd:
I disagree. This proves to me that strings are not just arrays of characters. There are many other functions that will need to be implemented differently for strings and arrays. I don't think the question should be only for reverse but for all array functions.

Here is an example of another non working example: http://bit.ly/1D7QiCv

davidchambers:
Would you prefer to add `R.strReverse`, @TheLudd?

TheLudd:
> Would you prefer to add R.strReverse, @TheLudd?

Yes (if it is deemed useful). Especially when looking at the bigger picture. We can go one of these ways:

 1. Make all array functions support strings
 2. Make some of the array functions support strings and create string specific functions where needed
 3. Make no array functions support strings and create string specific functions where needed

I think 1 is unfeasible. This means that `map`, `filter` et al should support strings and we don't want to go down that path right? In many array functions there is no reason for a string equivelent (ex `all`). 

I don't like 2 as there is no good definition of when a function should support strings or not. Hence I prefer 3. To me that is more simple than easy. But also not very hard...

davidchambers:
Does anyone else object to this change?

CrossEye:
I have a slightly queasy feeling that I imagine is parallel to objection by @TheLudd.  

But I'm more hopeful about his Option 2.  I think that there is a very real sense in which we should consider a String to be a list of characters.  (Of course the only representation of a character in JS is a length-1 String, so there's an odd infinite regress if we take that too literally.  But the idea is clear.)

I would not mind revisiting this to make any function `[a] -> b` also specifically work as `String -> b`.  But this is meant for an abstract type, `a`.  I wouldn't do this for, say, `pluck:: k -> [{k : v}] -> [v]`, since the list must be of type Object.

But `map`, `filter`, and `all` were mentioned in this context, and I can easily see any of them:

```js
var isVowel = x => {
  var y = R.toLower(x);
  return y == 'a' || y == 'e' || y == 'i' || y == 'o' || y == 'u';
};
R.map(R.ifElse(isVowel, R.toUpper, R.identity), "abcdefg") ; //=> "AbcdEfg"
R.filter(isVowel, "sequoia"); //=> "euoia"
R.map(R.all(isVowel), ["abc", "eieio", "oyvay"]); //=> [false, true, false]
```

So, no, I have no objection.  And I also think I'd like to see a coordinated approach to handling Strings as lists where appropriate.  This still leaves open questions.  Like `prepend:: a -> [a] -> [a]`.  Do we expect the `a`'s in the first `[a]` to each represent single-character strings?  If so, then would we want to do something to prevent, `prepend('un', 'do')`, insisting that it would have to be written as `prepend('u', prepend('n', 'do'))`?  But that's for later, and only if others like the basic idea.

davidchambers:
Thanks, @CrossEye. I've updated the patch to include an `_asChars` decorator which could be used in the definitions of `R.map`, `R.filter`, and other functions. If this pull request is merged I'll open a follow-up pull request to update these functions.

CrossEye:
:herb:, although the number of decorators is starting to scare me.  :smile:

paldepind:
I really don't think `_asChars` should be a decorator. Our stack traces are horrible enough as is.

If we decide not to support string handling in browsers without array indexing for strings it could be reduced to a function call at the end.

CrossEye:
@paldepind: I understand your objection; I'm worried too.  But I don't follow this:

> If we decide not to support string handling in browsers without array indexing for strings it could be reduced to a function call at the end.

Could you give some more detail of what you mean?

paldepind:
The decorator does two things:

* Before decorated function is invoked: If input is a string it convert it to an array.
* After decorated function is invoked: Computed array is converted to a string if input was string.

If we can assume that strings can be indexed like arrays (IE>7) the first step is unnecessary. Then a decorator is definitely unnecessary.

The decorator also leads to unnecessary extra work being done. If I invoke current `R.reverse` with a string then an array will be created out of the string and that created array will be copied for no good reason. There is no reason to call `_slice` on an array you've just created with `split`. Wasted CPU work, wasted memory.

The decorator is also too naive. It will not work as expected when used with `R.map`. Consider this:

```js
R.map((char) => char.charCodeAt(0), 'hello world')
```

The end result _should_ be a list of numbers but the `_asChars` decorator would convert it to a string with `join('')`. Definitely not what one wants.

All in all I think this decorator is horrible.

I think making all list functions accept strings is a road to a lot of complexity with very little benefit. Then I'd much rather have users `split` and `join` as needed. It is not much to demand for a rare use case.

Of the ways @TheLudd mentioned [before](https://github.com/ramda/ramda/pull/1275#issuecomment-128929705) I prefer number 2. Sure, it is a bit ad-hoc but I also think it is the practical standpoint between the ideal: strings are just arrays of characters, and the reality in JS: strings are primitives, characters don't exist and arrays of single length strings are nothing like actual strings. After number 2 I prefer number 3.

TheLudd:
> strings are just arrays of characters

> strings are primitives, characters don't exist and arrays of single length strings are nothing like actual strings

You managed to put both those two in the same sentence @paldepind :smile: 

In many languages strings are implemented as "an array of characters", but what is an array? Wikipedia gave these two short definitions on [this page](https://en.wikipedia.org/wiki/Array#In_computer_science)

> * Array data structure, an arrangement of items at equally spaced addresses in computer memory
* Array data type, used in a programming language to specify a variable that can be indexed

When you say that a string is just an array of characters are you referring to the data structure or the data type? If you refer to the data structure then you may be correct. But all of that is beneath the surface of the code and not something we deal with in every day programming. 

When we talk about arrays in JavaScript (what in ramda is referred to as lists) we talk about the array data type, specifically the JavaScript array data type. And from what I can tell, a string is not a  *JavaScript array data type* of characters.

davidchambers:
> ```javascript
> R.map((char) => char.charCodeAt(0), 'hello world')
> ```

This doesn't seem valid to me. Shouldn't the result of mapping over a string of length N be a string of length N? The function would therefore need to have type `Char -> Char` when the second argument to `R.map` has type `String`. If this is the case, `R.map(R.__, 'abc')` and `R.map(R.__, ['a', 'b', 'c'])` are not interchangeable expressions.

Perhaps having `R.map` support strings is a step too far. I'm very pleased that `R.indexOf`, `R.head`, `R.take`, `R.reverse`, and several other functions are polymorphic, but perhaps we should take this train no further.

TheLudd:
> Perhaps having R.map support strings is a step too far

Definitely, please never go there :smile:

paldepind:
@davidchambers
> This doesn't seem valid to me. Shouldn't the result of mapping over a string of length N be a string of length N?

Why? Your `R.map` handles strings as if they where lists of characters. Why should my map function not be allowed to turn a character into something else?

> The function would therefore need to have type `Char -> Char` when the second argument to R.map has type String.

To me that is a limitation of the implementation but conceptually it should be allowed. When mapping over a string the type should be: `(Char -> a) -> [Char] -> [a]`.

@TheLudd
> You managed to put both those two in the same sentence @paldepind :smile:

I was comparing ideal world vs JavaScript reality.

> When we talk about arrays in JavaScript (what in ramda is referred to as lists) we talk about the array data type, specifically the JavaScript array data type. And from what I can tell, a string is not a JavaScript array data type of characters.

Yes. That is also what I wrote.

CrossEye:
>> This doesn't seem valid to me. Shouldn't the result of mapping over a string of length N be a string of length N?

> Why? Your `R.map` handles strings as if they where lists of characters. Why should my map function not be allowed to turn a character into something else?

Agreed.  It did sound good for a short while...

This is clearly why Clojure has some common behavior on many iterable types, but often ends up turning the results into a single common one.  You have to then create your specific type out of the result if you don't like that default behavior.

TheLudd:
> Yes. That is also what I wrote.

@paldepind Now I am a little bit confused. Are we agreeing that strings are not just arrays of characters given that by array we are referring to the JavaScript array type?

paldepind:
@TheLudd Yes. Of course. I'm not sure what I wrote that made you think otherwise.

TheLudd:
@paldepind [this](https://github.com/ramda/ramda/pull/1184#issuecomment-115599652) and [this](https://github.com/ramda/ramda/pull/1214#issuecomment-114018769)

paldepind:
@TheLudd I'm sorry if I've been unclear. In both of these cases I'm taking about lists of characters as an abstract thing.

> To me a string is just a list of characters.

That's quite different that saying: Strings in JavaScript are arrays of characters. Also earlier in this thread I clearly state that JavaScript does not implement strings as arrays of characters.

Anyway. I hope the confusion is sorted out and that we now know that we both agree on that matter! :smile:

scott-christopher:
In languages like Haskell, a `String` is just a type alias for a `[Char]`, but we can only mimic that in JS and can never be sure whether a function resulting in an array of single letter strings should automatically be converted back to a string or not.

Perhaps a middle ground here is to expose `toArray`/`fromArray` functions, allowing the user to be in control of the isomorphism like we already do for objects with `toPairs`/`fromPairs`. In fact, I would almost go as far to suggest deprecating `toPairs`/`fromPairs` in favour of `toArray`/`fromArray` that could dispatch on its argument's type.

We could also consider exposing something like an `asArray` function, similar to `addIndex` that wraps functions like `map`, etc, handling the `toArray` and `fromArray` before and after to provide the same functionality that `_asChars` does here.

CrossEye:
@scott-christopher: If I follow correctly, then your suggestion would have

```js
toArray('abc'); //=> ['a', 'b', 'c'] // (_asChars initial behavior)
toArray({a: 1, b: 2}); //=> [['a', 1], ['b', 2]] // (toPairs behavior)
```

If so, then my first impression is negative.  I don't have a major objection to polymorphic behavior regarding input parameters, but I really like consistent return types (even if they're parameterized.)  This just looks like a bunch of unrelated functions stuck together with a common name.

(I suppose others may see my support for delegation in the same vein.  I don't, but I recognize that possible connection.)

But I like the idea of an `addIndex`-style API doing much like what @davidchambers proposed, only under the control of the user.  It won't quite work, I believe, but something close to this?:

```js
R.asArray(R.split(''), R.into(''), R.filter)(isVowel, "sequoia"); //=> "euoia"
```

`R.into` would probably not quite work, but it's got a similar feel.

scott-christopher:
It also occurred to me that a generic `fromArray` would suffer from pretty fuzzy duck typing too (e.g. Is the head element a string of length 1, or an array of length 2, etc). So I think if we were to introduce support for `String -> [Char]` and `[Char] -> String`, they should be specialised functions.

I'll raise a separate PR to discuss the idea of an `iso` function for handling the wrapping I mentioned above.

PR #1276, opened 2015-07-13
avoid looping infinitely on bad input in R.range

### LINKED ISSUE #1222: Calling range with letters causes Chrome to crash
Calling `R.range('a', 'z')` breaks Chrome so badly it needs to be restarted.

Although `range` is only supposed to work on numbers, it should fail gracefully rather than taking out the browser (especially since range works on characters in other languages so its not an unreasonable error to make when developing).

davidchambers:
Closes #1222

### DIFF
--- a/test/range.js
+++ b/test/range.js
@@ -4,6 +4,7 @@ var R = require('..');
 
 
 describe('range', function() {
+
   it('returns list of numbers', function() {
     assert.deepEqual(R.range(0, 5), [0, 1, 2, 3, 4]);
     assert.deepEqual(R.range(4, 7), [4, 5, 6]);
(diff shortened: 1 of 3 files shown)


test/range.js
```diff
@@ -25,4 +26,15 @@ describe('range', function() {
     result.push(5);
     assert.deepEqual(R.range(10, 5), []);
   });
+
+  it('terminates given bad input', function() {
+    assert.throws(
+      function() { R.range('a', 'z'); },
+      function(err) {
+        return err.constructor === TypeError &&
+               err.message === 'Both arguments to range must be numbers';
```
CrossEye: I really believe this style of testing the specific message in a unit-test is over-specification.

Checking that we throw a TypeError should be plenty.

PR #1280, opened 2015-07-15
Revert "add R.propIs and R.propSatisfies"

### LINKED ISSUE #1279: Revert change of propEq
I consider the changing of argument order in `propEq` to be a mistake. And it feels like an outcome of a misunderstanding of a [comment](https://github.com/ramda/ramda/pull/1243#issuecomment-119541173) I made. 

I noticed that there were some functions beginning with "prop" that took the prop name first and others that didn't. I didn't say this was a bad thing. I just wanted to highlight that it could be both and in the case of #1243 I thought it should be changed.

I do *not* feel that `propEq` should have the argument order changed. We use that function heavily and we sometimes only apply the first or just the first two arguments but we have *never* done `propEq(R.__, someValue)`. Now we would have to `propEq(R.__, 'someProp')` all over our code.

I vote to revert that change.

I also have some issues with how this was done, regardless of whether the change itself was good or not:
  * This change was merged in a PR unrelated to `propEq`
  * The change was not discussed
  * The PR was merged by the author

IMHO this is bad practice and I don't know if ramda has any guidelines for how to handle PRs but each of the items goes against what I find to be the purpose of PRs.

davidchambers:
Reverts #1243; closes #1279

If we merge this we should publish v0.17.0.


TheLudd:
LGTM, no need for me to remove the new functions though.

CrossEye:
:-1:

I think both `propIs` and `propSatisfies` are useful functions.  The fact that their natural parameter order differs from that of `propEq` is slightly unfortunate, but far from tragic.  Yes, we want the library to be consistent, but the consistency that we most want to support is _usefulness_.

It was Gleb Bahmutov who first [pointed out][gb] that what really mattered in our parameter order was that we put those parameters less likely to change ahead of those more likely to.  I've pretty well adopted it as a mantra on my own.  I think it usually makes it clear just how a function should be designed.  If we continue to stress that, then we will maintain a strong consistency, even if on occasion two related functions are less similar than we might prefer.

So I'd like to revert only the parameter-order changes to `propEq`.



  [gb]: http://glebbahmutov.com/blog/put-callback-first-for-elegance/

svozza:
It would be nice to get a decision on this quickly, I'm holding off on upgrading to v0.16 because I don't want to spend the time changing the huge amount of `propEq` invocations in our codebase only to have to change them all back again.

buzzdecafe:
My $0.02:

Revert `propEq`: Yes
Revert entire commit: No

CrossEye:
Anyone else?  @paldepind?  @raine?

raine:
`propEq` that takes the property name first makes more sense.

davidchambers:
I'll open a new pull request to revert just the `R.propEq` change.

PR #1286, opened 2015-07-17
revert change to R.propEq argument order

### LINKED ISSUE #1279: Revert change of propEq

davidchambers:
Closes #1279

### DIFF
--- a/src/find.js
+++ b/src/find.js
@@ -21,8 +21,8 @@ var _xfind = require('./internal/_xfind');
  * @example
  *
  *      var xs = [{a: 1}, {a: 2}, {a: 3}];
- *      R.find(R.propEq(2, 'a'))(xs); //=> {a: 2}
- *      R.find(R.propEq(4, 'a'))(xs); //=> undefined
+ *      R.find(R.propEq('a', 2))(xs); //=> {a: 2}
+ *      R.find(R.propEq('a', 4))(xs); //=> undefined
  */
 module.exports = _curry2(_dispatchable('find', _xfind, function find(fn, list) {
   var idx = 0;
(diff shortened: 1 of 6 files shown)


CrossEye:
:herb:

buzzdecafe:
:cow2:

look good to you, @TheLudd?

TheLudd:
:horse:

svozza:
Is there going to be a 0.16.1 release for this or will it be rolled up into 0.17?

TheLudd:
I'd say `0.16.1` but it majes luttle dufference qhatever the choice is.

davidchambers:
> Is there going to be a 0.16.1 release for this or will it be rolled up into 0.17?

I'll publish v0.17.0 today. This will be the only change in the release. I don't consider v0.16.1 appropriate for an API change: it would cause problems for those specifying `"0.16.x"` as the Ramda version in their __package.json__.

asaf-romano:
Has the plan changed?

davidchambers:
> Has the plan changed?

No. I didn't publish a new version on Friday as I said I would. Sorry about that. I've just done so. :)

PR #1287, opened 2015-07-18
rename isSet to allUniq

### LINKED ISSUE #1278: Rename isSet
isSet naming is too confusing, imo. My first guess would be that it is either ```has``` or ```is(Set)```.

asaf-romano:
Closes #1278

### DIFF
--- a/src/allUniq.js
+++ b/src/allUniq.js
@@ -14,11 +14,11 @@ var _indexOf = require('./internal/_indexOf');
  * @return {Boolean} `true` if all elements are unique, else `false`.
  * @example
  *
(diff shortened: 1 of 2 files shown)


davidchambers:
Perhaps this pull request should add `R.allUniq` and deprecate `R.isSet` (rather than remove it).

TheLudd:
@davidchambers good catch =)

davidchambers:
@asaf-romano, would you like to update this pull request?

asaf-romano:
will do.

davidchambers:
It would have been nice to update this pull request rather than open a new one.

PR #1293, opened 2015-07-20
Fix `R.match` empty array default

anton-rudeshko:
Problem: if someone evil will change resulting default array then everybody will suffer because of shared default array pointer. Example:
```js
R.match(/a/, '')[1] = 'fail';
R.match(/a/, ''); // [, 'fail']
```
-----
Imperative style looks better and cleaner here IMHO.
Functional approach may look like this:
```js
var defaultToEmptyArray = ifElse(isNil, constructN(0, Array), identity);
R.match = compose(defaultToEmptyArray, invoker(1, 'match'));
```

### DIFF
--- a/test/match.js
+++ b/test/match.js
@@ -5,15 +5,27 @@ var R = require('..');
 
 describe('match', function() {
   var re = /[A-Z]\d\d\-[a-zA-Z]+/;
+  var matching = 'B17-afn';
+  var notMatching = 'B1-afn';
 
   it('determines whether a string matches a regex', function() {
-    assert.strictEqual(R.match(re, 'B17-afn').length, 1);
-    assert.deepEqual(R.match(re, 'B1-afn'), []);
+    assert.strictEqual(R.match(re, matching).length, 1);
+    assert.deepEqual(R.match(re, notMatching), []);
   });
 
   it('is curried', function() {
     var format = R.match(re);
-    assert.strictEqual(format('B17-afn').length, 1);
-    assert.deepEqual(format('B1-afn'), []);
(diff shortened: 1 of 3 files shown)


buzzdecafe:
LGTM
:cow2:

test/match.js
```diff
@@ -5,15 +5,27 @@ var R = require('..');
...
+  });
+
+  it('defaults to a different empty array each time', function() {
+    var first = R.match(re, notMatching);
+    var second = R.match(re, notMatching);
+    assert.notStrictEqual(first, second);
+  });
+
+  it('throws on null input', function() {
+    assert.throws(function() { R.match(re, null); }, TypeError, /null does not have a method named "match"/);
```
davidchambers: I imagine @CrossEye will object to the specificness of this assertion.

CrossEye: Ok, since I've been summoned, I object to the specificness of this assertion!  :smile:

I see no reason for the test to assert the specific message supplied.  A test is part of our public interface to my mind, and I would prefer to feel free to change this without feeling as though I've changed the API.  Specifying an error, or at most specifying the type of the error, should be plenty.

anton-rudeshko:
Okay, I removed assert for exact error message.

anton-rudeshko:
So?

CrossEye:
Thanks, @anton-rudeshko!  Looks great.

anton-rudeshko:
Thank you, will be waiting for v0.17.1 with this fix.

davidchambers:
> Thank you, will be waiting for v0.17.1 with this fix.

Wait no longer. :)

Issue #1292, opened 2015-07-20
Unexpected behaviour of R.clone


nomosyn:
We have:
```
(new Mongo.ObjectID()) instanceof Mongo.ObjectID // => true ✓
(EJSON.clone(new Mongo.ObjectID())) instanceof Mongo.ObjectID // => true ✓
```

But:
```
(R.clone(new Mongo.ObjectID())) instanceof Mongo.ObjectID // false (EXPECTED TO BE TRUE)
```
Maybe some inpirations could be taken from [EJSON.clone](https://github.com/meteor/meteor/blob/devel/packages/ejson/ejson.js) ?

CrossEye:
I could definitely see borrowing this:

```js
  if (typeof v.clone === 'function') {
    return v.clone();
  }
```

I would certainly not want us to try to get into the business of figuring out someone else's inheritance models and trying to capture them in our clones, but if they have a clone method, perhaps we could use it?

nomosyn:
Yes, just like [R.equals](https://github.com/ramda/ramda/blob/master/src/equals.js) does it...
Would be useful to me, probably to others (anyone?), I cannot see why it would be a bad thing...

davidchambers:
Dispatching to `clone` seems reasonable. I agree with Scott that we shouldn't go any further than that.

CrossEye:
@nomosyn: Do you feel like opening a PR for this???

nomosyn:
haha yes let's do it! :-)

PR #1295, opened 2015-07-21
Make clone dispatch to `clone` method if present.

nomosyn:
Fixes: #1292

### DIFF
--- a/test/clone.js
+++ b/test/clone.js
@@ -189,3 +189,17 @@ describe('deep clone edge cases', function() {
     assert.notStrictEqual(R.clone(list), list);
   });
 });
+
+
+describe('Let `R.clone` use an arbitrary user defined `clone` method', function() {
+
+  it('dispatches to `clone` method if present', function() {
+    function ArbitraryClone(x) { this.value = x; }
+    ArbitraryClone.prototype.clone = function() { return new ArbitraryClone(this.value); };
+
+    var obj = new ArbitraryClone(42);
+    var arbitraryClonedObj = R.clone(obj);
+    assert.deepEqual(arbitraryClonedObj, new ArbitraryClone(42), true);
+    assert.strictEqual(arbitraryClonedObj instanceof ArbitraryClone, true);
+  });
+});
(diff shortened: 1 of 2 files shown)


test/clone.js
```diff
@@ -189,3 +189,20 @@ describe('deep clone edge cases', function() {
...
 });
+
+
+describe('Let `R.clone` use an arbitrary user defined `clone` method', function() {
+
+  it('dispatches to `clone` method if present', function() {
+    function ArbitraryClone(x) { this.value = x; }
+
+    // user defined `clone` method can or cannot make sense: it's arbitrary.
+    ArbitraryClone.prototype.clone = function() { return new ArbitraryClone(2 * this.value); };
```
CrossEye: Although clearly an `ArbitraryClone` would be allowed to do whatever it wanted to, I'd rather _not_ call attention to it in the test suite, just doing, `return new ArbitraryClone(this.value)`.

If you find this that important, I think it should just be a comment.  ("Note that if Ramda delegates to the object's 'clone' method, it does no checking that data is similar," or some such.)  But personally, I don't find it really worth spending any effort on.

CrossEye:
Thank you.  Except for the minor issue mentioned about the new test, this looks good to me.

nomosyn:
Ok! Updated.

buzzdecafe:
:cow2:

nomosyn:
mmh... That's a cow.

davidchambers:
Does the special case for arrays in `_hasMethod` both anyone else in this context? I expect `let x = []; x.clone = () => 42; let y = R.clone(x);` to bind `42` to `y`. The fact that it doesn't suggests an implementation detail is affecting the API.

CrossEye:
> Does the special case for arrays in `_hasMethod` both anyone else in this context? 

No more than it does for `chain` or `of` or `tail`.  It seems slightly unfortunate, but not tragic.  If you extend an array, or `Array.prototype` with a new method, then our dispatch mechanism might well skip it when trying to convert your API from `obj.method(...params)` to `function(...params, obj)`.  But your original sytax will still work.  That seems as far as it goes.

PR #1299, opened 2015-07-22
Handle empty list of predicates in _predicateWrap

### LINKED ISSUE #1298: allPass with empty list of predicates
R.allPass seems to fail when partially applied with an empty list of predicates. To reproduce:

`R.allPass([], 1);`

> true

`R.allPass([]);`

> ramda.js:85 Uncaught Error: First argument to _arity must be a non-negative integer no greater than ten

MartijnDwars:
Math.max on an empty array returns -Infinity, causing things to go wrong. Instead, I reduce the list of lengths with `max` as iterator function and 0 as default value.

I'd love to hear back from you. If there is anything I can improve, let me know.

Fixes #1298

### DIFF
--- a/src/internal/_predicateWrap.js
+++ b/src/internal/_predicateWrap.js
@@ -1,7 +1,8 @@
 var _arity = require('./_arity');
 var _slice = require('./_slice');
+var max = require('../max');
 var pluck = require('../pluck');
-
+var reduce = require('../reduce');
 
 /**
  * Create a predicate wrapper which will call a pick function (all/any) for each predicate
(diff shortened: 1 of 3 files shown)


scott-christopher:
It would probably be worthwhile adding similar tests to `anyPass` along with this change as well:

```js
it('returns false for an empty predicate list', function() {
  assert.strictEqual(R.anyPass([])(3), false);
});
```

CrossEye:
:herb:

buzzdecafe:
:cow2:

PR #1303, opened 2015-07-22
support any Chain in R.unnest

davidchambers:
If we're to provide sugar for `R.chain(R.identity)` it should work for any [Chain][1], not just Array-like ones.

This change enables nested maybes to be unnested:

```javascript
> R.unnest(Just(Just(42)))
Just(42)
```

[1]: https://github.com/fantasyland/fantasy-land#chain

### DIFF
--- a/test/unnest.js
+++ b/test/unnest.js
@@ -1,9 +1,11 @@
 var assert = require('assert');
 
 var R = require('..');
+var Maybe = require('./shared/Maybe');
 
 
 describe('unnest', function() {
+
   it('only flattens one layer deep of a nested list', function() {
     var nest = [1, [2], [3, [4, 5], 6, [[[7], 8]]], 9, 10];
     assert.deepEqual(R.unnest(nest), [1, 2, 3, [4, 5], 6, [[[7], 8]], 9, 10]);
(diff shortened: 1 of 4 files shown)


test/unnest.js
```diff
@@ -26,4 +27,38 @@ describe('unnest', function() {
...
+      }
+      this.value = x;
+    }
+    Just.prototype.chain = function(f) {
+      return f(this.value);
+    };
+    Just.prototype.toString = function() {
+      return 'Just(' + R.toString(this.value) + ')';
+    };
+
```
CrossEye: Or just replace lines 32 - 55 with 

```js
var Maybe = require('./shared/Maybe');
```

:smile: 

See [`test/commute.js`][co] or [`test/lift.js`][li] for examples.

  [co]: https://github.com/ramda/ramda/blob/292e081fff69ad9f6ba39e7295de032c6ff4ee8a/test/commute.js#L2
  [li]: https://github.com/ramda/ramda/blob/292e081fff69ad9f6ba39e7295de032c6ff4ee8a/test/lift.js#L6

CrossEye:
I definitely like this abstraction.  Very nice!

:herb:

buzzdecafe:
:cow2:

PR #1313, opened 2015-07-26
add R.pair

### LINKED ISSUE #1310: add R.pair
```haskell
pair :: a -> b -> (a,b)
```

Didn't someone once say that all numbers except 0 and 1 are arbitrary? In the context of Ramda, I disagree: 2 is also a number with special significance. `R.fromPairs` and `R.toPairs` use a pair to represent a map entry.

This morning I wanted to transform `[{ name :: String, sig :: String, description :: String }]` into `Map String { name :: String, sig :: String, description :: String }`. Here's my solution:

```javascript
R.fromPairs(R.map(R.converge(R.concat, R.compose(R.of, R.prop('name')), R.of), list))
```

With `R.pair`, I could have written:

```javascript
R.fromPairs(R.map(R.converge(R.pair, R.prop('name'), R.identity), list))
```

I believe this function would be a good companion of `R.createMapEntry`. The implementation is trivial.

davidchambers:
Closes #1310

### DIFF
--- a/src/createMapEntry.js
+++ b/src/createMapEntry.js
@@ -11,6 +11,7 @@ var _curry2 = require('./internal/_curry2');
  * @param {String} key
  * @param {*} val
  * @return {Object}
+ * @see R.pair
  * @example
  *
  *      var matchPhrases = R.compose(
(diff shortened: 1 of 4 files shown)


paldepind:
:+1: What about adding a `@see` to `fromPairs` as well?

CrossEye:
:herb:

davidchambers:
> What about adding a `@see` to `fromPairs` as well?

:zap:

PR #1315, opened 2015-07-28
Adds R.when and R.unless

scott-christopher:
Not sure how welcome this will be, but in the trend of suggesting sugar for common patterns, I am continually finding myself writing `R.ifElse(pred, doSomething, R.identity)`, or `R.ifElse(pred, R.identity, doSomethingElse)`.

This change introduces `R.when` and `R.unless` as sugar for the above.

### DIFF
--- a/src/ifElse.js
+++ b/src/ifElse.js
@@ -9,6 +9,7 @@ var curryN = require('./curryN');
  * @func
  * @memberOf R
  * @category Logic
+ * @see R.unless, R.when
  * @sig (*... -> Boolean) -> (*... -> *) -> (*... -> *) -> (*... -> *)
  * @param {Function} condition A predicate function
  * @param {Function} onTrue A function to invoke when the `condition` evaluates to a truthy value.
@@ -17,11 +18,13 @@ var curryN = require('./curryN');
  *                    function depending upon the result of the `condition` predicate.
  * @example
  *
- *      // Flatten all arrays in the list but leave other values alone.
- *      var flattenArrays = R.map(R.ifElse(Array.isArray, R.flatten, R.identity));
- *
- *      flattenArrays([[0], [[10], [8]], 1234, {}]); //=> [[0], [10, 8], 1234, {}]
- *      flattenArrays([[[10], 123], [8, [10]], "hello"]); //=> [[10, 123], [8, 10], "hello"]
+ *      var incCount = R.ifElse(
+ *        R.has('count'),
+ *        R.over(R.lensProp('count'), R.inc),
+ *        R.assoc('count', 1)
+ *      );
+ *      incCount({});           //=> { count: 1 }
+ *      incCount({ count: 1 }); //=> { count: 2 }
  */
 module.exports = _curry3(function ifElse(condition, onTrue, onFalse) {
   return curryN(Math.max(condition.length, onTrue.length, onFalse.length),
--- a/src/unless.js
+++ b/src/unless.js
@@ -1,0 +1,30 @@
+var _curry3 = require('./internal/_curry3');
+
+
+/**
+ * Tests the final argument by passing it to the given predicate function.
+ * If the predicate is not satisfied, the function will return the
+ * result of calling the `whenFalseFn` function with the same argument. If the
+ * predicate is satisfied, the argument is returned as is.
+ *
+ * @func
+ * @memberOf R
+ * @category Logic
+ * @see R.ifElse, R.when
+ * @sig (a -> Boolean) -> (a -> a) -> a -> a
+ * @param {Function} pred        A predicate function
+ * @param {Function} whenFalseFn A function to invoke when the `pred` evaluates
+ *                               to a falsy value.
+ * @param {*}        x           An object to test with the `pred` function and
+ *                               pass to `whenFalseFn` if necessary.
+ * @return {*} Either `x` or the result of applying `x` to `whenFalseFn`.
+ * @example
+ *
+ *      // coerceArray :: (a|[a]) -> [a]
+ *      var coerceArray = R.unless(R.isArrayLike, R.of);
+ *      coerceArray([1, 2, 3]); //=> [1, 2, 3]
+ *      coerceArray(1);         //=> [1]
+ */
+module.exports = _curry3(function unless(pred, whenFalseFn, x) {
+  return pred(x) ? x : whenFalseFn(x);
+});
--- a/src/when.js
+++ b/src/when.js
@@ -1,0 +1,33 @@
+var _curry3 = require('./internal/_curry3');
+
+
+/**
+ * Tests the final argument by passing it to the given predicate function.
+ * If the predicate is satisfied, the function will return the result
+ * of calling the `whenTrueFn` function with the same argument. If the predicate
+ * is not satisfied, the argument is returned as is.
+ *
+ * @func
+ * @memberOf R
+ * @category Logic
+ * @see R.ifElse, R.unless
+ * @sig (a -> Boolean) -> (a -> a) -> a -> a
+ * @param {Function} pred       A predicate function
+ * @param {Function} whenTrueFn A function to invoke when the `condition`
+ *                              evaluates to a truthy value.
+ * @param {*}        x          An object to test with the `pred` function and
+ *                              pass to `whenTrueFn` if necessary.
+ * @return {*} Either `x` or the result of applying `x` to `whenTrueFn`.
+ * @example
+ *
+ *      // truncate :: String -> String
+ *      var truncate = R.when(
+ *        R.propSatisfies(R.gt(R.__, 10), 'length'),
+ *        R.pipe(R.take(10), R.append('…'), R.join(''))
+ *      );
+ *      truncate('12345');         //=> '12345'
+ *      truncate('0123456789ABC'); //=> '0123456789…'
+ */
+module.exports = _curry3(function when(pred, whenTrueFn, x) {
+  return pred(x) ? whenTrueFn(x) : x;
+});
--- a/src/reduced.js
+++ b/src/reduced.js
@@ -21,7 +21,7 @@ var _reduced = require('./internal/_reduced');
  * @example
  *
  *      R.reduce(
- *        R.pipe(R.add, R.ifElse(R.lte(10), R.reduced, R.identity)),
+ *        R.pipe(R.add, R.when(R.gte(R.__, 10), R.reduced)),
  *        0,
  *        [1, 2, 3, 4, 5]) // 10
  */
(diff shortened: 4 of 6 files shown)


src/unless.js
```diff
@@ -0,0 +1,38 @@
...
+ *                    depending upon the result of the `condition` predicate.
+ * @example
+ *
+ *      var coerceArray = R.unless(
+ *        R.isArrayLike,
+ *        R.ifElse(
+ *          R.has('toArray'),
+ *          R.invoker(0, 'toArray')
+ *          R.of
+ *        )
```
CrossEye: I think mixing `ifElse` in the example will just serve to confuse things.  Perhaps a simpler example here?

src/when.js
```diff
@@ -0,0 +1,33 @@
...
+ * @see R.ifElse, R.unless
+ * @sig (a -> Boolean) -> (a -> a) -> (a -> a)
+ * @param {Function} condition A predicate function
+ * @param {Function} whenTrue  A function to invoke when the `condition`
+ *                             evaluates to a truthy value.
+ * @return {Function} A new function that will process the `whenTrue` function
+ *                    depending upon the result of the `condition` predicate.
+ * @example
+ *
+ *      var truncate = R.when(
```
CrossEye: Although we don't usually put them in examples, I think here it might be worth including the signature of `truncate`.

CrossEye:
I wouldn't stand in the way of this, but I'd want to know first if others are feeling the same need.  This is clear and obvious enough, but fills no special void in my life.  :smile:

scott-christopher:
Yeah, I'm not too bothered whether this lands or not. I've just found it tends to read a bit better as `when` or `unless`.

src/unless.js
scott-christopher: Agreed. I've simplified it to just defer to `R.of` instead.

src/when.js
scott-christopher: I've pushed up a change with the signature of truncate included.

CrossEye:
Well, as I said, I don't have any real objection, but think it should probably be more than one person's itch that needed to be scratched here.  If others have found a similar need, I'm all for it.  I simply haven't myself.

src/unless.js
```diff
@@ -0,0 +1,31 @@
...
+ * @example
+ *
+ *      // coerceArray :: (a|[a]) -> [a]
+ *      var coerceArray = R.unless(R.isArrayLike, R.of);
+ *      coerceArray([1, 2, 3]); //=> [1, 2, 3]
+ *      coerceArray(1);         //=> [1]
+ */
+module.exports = _curry2(function(pred, whenFalseFn, x) {
+  return ifElse(pred, identity, whenFalseFn, x);
+});
```
davidchambers: We like to use named function expressions rather than anonymous function expressions at the top level.

Also, what is the `x` doing?

CrossEye:
With these changes, the code looks good.  So a provisional :herb: from me, if there are others who find it compelling.

src/unless.js
```diff
@@ -0,0 +1,31 @@
+var _curry2 = require('./internal/_curry2');
+var identity = require('./identity');
```
davidchambers: We could use the internal `_identity` instead.

davidchambers:
I like these functions. While we can certainly live without them, they are nice sugar in the cases where they're applicable. I looked at a Plaid project which makes heavy use of Ramda: the "when" pattern is used in a handful of places.

src/unless.js
scott-christopher: Updated.

src/unless.js
scott-christopher: Remnants of a refactor :) I've pushed an update with `x` removed.

scott-christopher:
Is it worth adding `@see R.unless, R.when` to `R.ifElse`?

davidchambers:
> Is it worth adding `@see R.unless, R.when` to `R.ifElse`?

Yes. I favour lots of linking. :)

src/ifElse.js
```diff
@@ -9,6 +9,7 @@ var curryN = require('./curryN');
  * @func
  * @memberOf R
  * @category Logic
+ * @see R.unless, R.when
```
scott-christopher: I've added the link here.

src/when.js
```diff
@@ -0,0 +1,34 @@
+var _curry2 = require('./internal/_curry2');
+var _identity = require('./internal/_identity');
+var ifElse = require('./ifElse');
+
+
+/**
+ * Creates a function that will first pass its argument to the given predicate
+ * function. If the predicate returns true, the function will return the result
```
davidchambers: It needn't actually return the `true` value. Perhaps we could write _if the predicate is satisfied_ to sidestep the matter of implicit type coercion.

davidchambers:
:deciduous_tree:

src/when.js
scott-christopher: I originally had `If the predicate is satisfied` but was worried that may not be clear to some people what that actually means.

Happy to change it, as I'd prefer it myself.

scott-christopher: I've pushed up a change to amend this.

src/when.js
```diff
@@ -0,0 +1,34 @@
...
+ *      // truncate :: String -> String
+ *      var truncate = R.when(
+ *        R.propSatisfies(R.gt(R.__, 10), 'length'),
+ *        R.pipe(R.take(10), R.append('…'), R.join(''))
+ *      );
+ *      truncate('12345');         //=> '12345'
+ *      truncate('0123456789ABC'); //=> '0123456789…'
+ */
+module.exports = _curry2(function when(pred, whenTrueFn) {
+  return ifElse(pred, whenTrueFn, _identity);
```
scott-christopher: I'm wondering whether it's a bit unnecessary to use `ifElse` for the implementation here. These are unary functions by design, so could be implemented as simply as:
```js
_curry3(function when(pred, whenTrueFn, x) {
  return pred(x) ? whenTrueFn(x) : x;
});
```

paldepind: @scott-christopher Agreed. And likewise for unless

```js
_curry3(function unless(pred, fn, x) {
  return pred(x) ? x : fn(x);
});
```

scott-christopher:
I've updated this PR to use a simpler ternary implementation.

src/when.js
scott-christopher: One step ahead of you :D

paldepind:
I really like this addition :+1:  I've been using the `R.when` pattern quite a bit and this reads a lot better.

The current example of `R.ifElse` needs to be updated since it now exemplifies a case that would be better handled by `R.when`. Also, the example for `R.reduced` should be updated to use `R.when`.

scott-christopher:
I've updated the examples as suggested by @paldepind

paldepind:
Looks good to me :)

src/when.js
```diff
@@ -0,0 +1,32 @@
...
+ * Creates a function that will first pass its argument to the given predicate
+ * function. If the predicate is satisfied, the function will return the result
+ * of calling the `whenTrue` function with the same argument. If the predicate
+ * is not satisfied, the argument is returned as is.
+ *
+ * @func
+ * @memberOf R
+ * @category Logic
+ * @see R.ifElse, R.unless
+ * @sig (a -> Boolean) -> (a -> a) -> (a -> a)
```
davidchambers: Since all three arguments may now be provided up front, the signature should really be

    (a -> Boolean) -> (a -> a) -> a -> a

rather than

    (a -> Boolean) -> (a -> a) -> (a -> a)

scott-christopher: Good catch. I've pushed up the modified sig.

src/when.js
```diff
@@ -0,0 +1,32 @@
...
+ * of calling the `whenTrue` function with the same argument. If the predicate
+ * is not satisfied, the argument is returned as is.
+ *
+ * @func
+ * @memberOf R
+ * @category Logic
+ * @see R.ifElse, R.unless
+ * @sig (a -> Boolean) -> (a -> a) -> a -> a
+ * @param {Function} condition A predicate function
+ * @param {Function} whenTrue  A function to invoke when the `condition`
```
davidchambers: Should these names match the actual parameter names?

scott-christopher: Yes they should, thanks. Updated.

scott-christopher: Actually, I'll need to update the docs for these now that they accept all 3 args. Stand by.

src/when.js
```diff
@@ -0,0 +1,32 @@
...
+ * is not satisfied, the argument is returned as is.
+ *
+ * @func
+ * @memberOf R
+ * @category Logic
+ * @see R.ifElse, R.unless
+ * @sig (a -> Boolean) -> (a -> a) -> a -> a
+ * @param {Function} pred       A predicate function
+ * @param {Function} whenTrueFn A function to invoke when the `condition`
+ *                             evaluates to a truthy value.
```
davidchambers: This is nitpicky, but "evaluates" should probably be indented to align with the "A".

src/when.js
scott-christopher: I've updated the docs to reflect the new signature.

src/when.js
```diff
@@ -0,0 +1,34 @@
...
+ * @category Logic
+ * @see R.ifElse, R.unless
+ * @sig (a -> Boolean) -> (a -> a) -> a -> a
+ * @param {Function} pred       A predicate function
+ * @param {Function} whenTrueFn A function to invoke when the `condition`
+ *                              evaluates to a truthy value.
+ * @param {*}        x          An object to test with the `pred` function and
+ *                              pass to `whenTrueFn` if necessary.
+ * @return {Function} A new function that will process the `whenTrue` function
+ *                    depending upon the result of the `condition` predicate.
```
davidchambers: The return value is no longer a function, if I've understood correctly.

scott-christopher: You're correct. (I should have just scrapped the entire docs and started again when I changed the signature...)

I've pushed up the changed docs.

src/when.js
```diff
@@ -0,0 +1,33 @@
...
+ * @memberOf R
+ * @category Logic
+ * @see R.ifElse, R.unless
+ * @sig (a -> Boolean) -> (a -> a) -> a -> a
+ * @param {Function} pred       A predicate function
+ * @param {Function} whenTrueFn A function to invoke when the `condition`
+ *                              evaluates to a truthy value.
+ * @param {*}        x          An object to test with the `pred` function and
+ *                              pass to `whenTrueFn` if necessary.
+ * @return {*} Either `x` or the result of calling `whenTrueFn` with `x`.
```
davidchambers: I'd prefer to write _applying … to_ rather than _calling … with_. Thoughts?

scott-christopher: Not bothered either way. I've updated it with your suggestion.

davidchambers:
Thanks for all the speedy tweaks, @scott-christopher!

PR #1316, opened 2015-07-29
add examples of R.maxBy and R.minBy in concert with R.reduce

davidchambers:
As suggested by @jon49 in https://github.com/ramda/ramda/pull/1231#issuecomment-121691615


CrossEye:
:herb:

PR #1323, opened 2015-07-30
rename R.createMapEntry

davidchambers:
Just an idea. `singleton` is quite a bit shorter than `createMapEntry`, though slightly less descriptive.

### DIFF
--- a/src/objOf.js
+++ b/src/objOf.js
@@ -1,4 +1,4 @@
-var objOf = require('./objOf');
+var _curry2 = require('./internal/_curry2');
 
 
 /**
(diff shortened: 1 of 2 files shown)


scott-christopher:
I tend to associate `singleton` with lists more so than maps. Given the similarity to `R.of`, perhaps this could be called `R.mapOf`?

davidchambers:
> I tend to associate `singleton` with lists more so than maps.

So do I.

> Given the similarity to `R.of`, perhaps this could be called `R.mapOf`?

I like it! :thumbsup:

kedashoe:
I like the sound of `mapOf` better than `objOf`, but maybe `objOf` is more appropriate? Right now if you filter the ramda docs for `map`, `createMapEntry`/`mapOf` is the odd fn out, and if you filter for `obj` you get your object functions except you don't see `createMapEntry`/`mapOf`. Also `Map`s are here/on the way and `mapOf` would not give you one. Maybe even just `obj`? `var x = R.obj('a', 100);`

scott-christopher:
Yeah, I agree that `objOf` would be more appropriate and consistent with Ramda naming than `mapOf`.

I actually have a helper function I use called `obj`, which is similar except it creates an object from multiple keys:
```js
var obj = R.unapply((keys) => R.compose(R.zipObj(keys), R.unapply(R.identity)));

var abc = obj('a', 'b', 'c');
abc(1, 2, 3); //=> {"a": 1, "b": 2, "c": 3}
```

CrossEye:
I've never liked '`createMapEntry`'; it's never seemed to match.  '`objOf`' (or '`obj`') sounds _much_ better.

buzzdecafe:
:+1: `objOf`

CrossEye:
I think we have a winner.

raine:
Woohoo!

paldepind:
`objOf` is really nice.

Regarding the PR I think the old deprecated function should simply be an alias to the new name. In the current PR the source contains the same function twice.

davidchambers:
I've renamed the function `objOf`, and made `createMapEntry` an :alien: as @paldepind suggested.

buzzdecafe:
μ :cow2:

Issue #1281, opened 2015-07-15
Clojure like update-in?


epeli:
In Clojure one can use [update-in](https://clojuredocs.org/clojure.core/update-in) to update value inside a deeply nested data structure.

So I was wondering whether Rambda could include something like it? (Or does it already?)

For example using it it would be easy to merge some new properties to a deeply nested object in immutable fashion:

```js
var coll = {foo: {bar: {a: 1}};
R.updatePath(["foo", "bar"], R.merge({b: 2}), coll);
// -> {foo: {bar: {a: 1, b: 2}}
```

TheLudd:
Will `assocPath` do?

epeli:
Nope. Afaik it can be only used to set the value. You have to type the path twice:

```js
R.assocPath(["foo", "bar"], R.merge(coll.foo.bar, {b: 2}), coll)
```

TheLudd:
Or just `R.assocPath(["foo", "bar"], 2, coll)`

epeli:
Well, that's the solution for the specific example case. There are quite a lot more use cases for this.

TheLudd:
Ah you mean like [update](http://ramdajs.com/docs/#adjust) but for nested property paths?

buzzdecafe:
this should work:

`R.assocPath(['foo', 'bar', 'b'], 2, coll); //=> {foo: {bar: {a: 1, b: 2}}}`

epeli:
Please read the Clojure documentation I've linked. Also the icepick module implements this:

https://github.com/aearly/icepick#updateincollection-path-callback

So in other words it's just like `R.assocPath` but instead of the value you pass in a function and the existing value (if any) is passed to the function and the return value of the function is set as the new value.

epeli:
@TheLudd Oh, If you mean `adjust` then yes!

TheLudd:
I did, correction made.

CrossEye:
[`evolve`][ev] is close, although it uses similar structures rather than separated paths

    R.evolve({foo: {bar: R.merge({b: 2})}}, coll);
    //=> {foo: {bar: {a: 1, b: 2}}

  [ev]: http://ramdajs.com/docs/#evolve

TheLudd:
This makes me think... for each `prop` function there could be an analogous `path` function and an `index` function. We have `adjust` but not `adjustProp` or `adjustPath` right? Is there some way to solve this so that there would be no gaps between indexes, paths and props?

[...]

PR #1328, opened 2015-08-06
Adds adjustPath

epeli:
First stab at #1281

Not quite sure this is right name for the function because plain `adjust` deals with arrays – not objects – but neither did the original proposal of `updatePath` feel any better since `update` is also for arrays and is used for setting values...

### DIFF
--- a/src/adjustPath.js
+++ b/src/adjustPath.js
@@ -1,0 +1,38 @@
+
+var _curry3 = require('./internal/_curry3');
+var _slice = require('./internal/_slice');
+var assoc = require('./assoc');
+
+
+/**
+ * Makes a shallow clone of an object, setting or overriding the nodes required
+ * to create the given path, applies the function with the value at the tail
+ * end of that path and places the result at the end.  Note that this copies
+ * and flattens prototype properties onto the new object as well.  All
+ * non-primitive properties are copied by reference.
+ * @see R.assocPath
+ * @see R.adjust
+ *
+ * @func
+ * @memberOf R
+ * @category Object
+ * @sig (a -> a) -> [a] -> {k: v} -> {k: v}
+ * @param {Function} fn The function to apply.
+ * @param {Array} path the path to set
+ * @param {Object} obj the object to clone
+ * @return {Object} a new object similar to the original except along the specified path.
+ * @example
+ *
+ *      R.adjustPath(R.add(2), ['a', 'b'], {a: {b: 1}}); //=> {a: {b: 3}}
+ *
+ */
+module.exports = _curry3(function assocPath(fn, path, obj) {
+  switch (path.length) {
+    case 0:
+      return obj;
+    case 1:
+      return assoc(path[0], fn(obj[path[0]]), obj);
+    default:
+      return assoc(path[0], assocPath(fn, _slice(path, 1), Object(obj[path[0]])), obj);
+  }
+});
--- a/test/adjustPath.js
+++ b/test/adjustPath.js
@@ -1,0 +1,46 @@
+var assert = require('assert');
+
+var R = require('..');
+
+
+describe('adjustPath', function() {
+  it('makes a shallow clone of an object, overriding only what is necessary for the path', function() {
+    var obj1 = {a: {b: 1, c: 2, d: {e: 3}}, f: {g: {h: 4, i: 5, j: {k: 6, l: 7}}}, m: 8};
+    var obj2 = R.adjustPath(R.add(1), ['f', 'g', 'i'],  obj1);
+    assert.deepEqual(obj2,
+        {a: {b: 1, c: 2, d: {e: 3}}, f: {g: {h: 4, i: 6, j: {k: 6, l: 7}}}, m: 8}
+    );
+    // Note: reference equality below!
+    assert.strictEqual(obj2.a, obj1.a);
+    assert.strictEqual(obj2.m, obj1.m);
+    assert.strictEqual(obj2.f.g.h, obj1.f.g.h);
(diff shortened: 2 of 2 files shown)


davidchambers:
I'm interested in exploring the possibility of *removing* some or all of the `*Path` functions now that we have composable lenses.

The example from the docstring,

> ```javascript
> R.adjustPath(R.add(2), ['a', 'b'], {a: {b: 1}}); //=> {a: {b: 3}}
> ```

could be written

```javascript
var aLens = R.lensProp('a');
var bLens = R.lensProp('b');
var a = {b: R.compose(aLens, bLens)};

R.over(a.b, R.add(2), {a: {b: 1}}); //=> {a: {b: 3}}
```

As discussed in #1281, lenses are great when a path is used more than once, but are less convenient for one-offs. To address this, we could add `R.toLens :: String -> Lens s a`. The example could then be written

```javascript
R.over(R.toLens('.a.b'), R.add(2), {a: {b: 1}}); //=> {a: {b: 3}}
```

epeli:
Hmm cool. Now that you mention `R.toLens` I'm starting to warm up to lenses bit more. Although I'd like to see it with an array param instead of a string (less edge cases with dot and easier to build dynamically etc.).

```js
R.over(R.lensPath(['a', 'b']), R.add(2), {a: {b: 1}});
```
I think it would go nicely along with `lensProp` and `lensIndex`. 

And it is indeed pretty clear that `adjustPath` is just sugar for this. But I would still argue it's worth it. Again take a look at Clojure. It has bunch of libraries implementing lenses but `update-in` and friends are in the core just because it so useful and terse in so many situations.

---


Here my use case with React.js and Flux: I want to thread my data as immutable because then I can make my components "pure" which yields to insane performance benefits on avoided DOM mutations.

I have bunch of stores implementing an immutable state object

```js
this.state = { comments: {} };
```

and methods which update it

```js
addComment(comment) {
  this.state = R.assocPath(["comments", comment.id], comment, this.state);
}
```

```js
incLikes(commentId) {
  this.state = R.adjustPath(R.add(1), ["comments", commentId, "likeCount"], this.state);
}
```

etc.

And I do find the `*path` functions super convenient __and readable__ for operations like these.


Thanks for hearing me out.

scott-christopher:
A `lensPath` function is pretty straightforward to implement using the existing `lensIndex` and `lensProp` functions:
```js
var lensPath = R.reduce((lens, prop) => {
  var propToLens = R.ifElse(R.is(Number), R.lensIndex, R.lensProp);
  return R.compose(lens, propToLens(prop));
}, R.identity);

// Or the more convoluted, point-free edition
var lensPath = R.reduce(R.useWith(R.compose, R.identity, R.ifElse(R.is(Number), R.lensIndex, R.lensProp)), R.identity);

var obj = { x: [0, { y: "moo" }] }

var x1yLens = lensPath(['x', 1, 'y']);

R.view(x1yLens, obj);       // "moo"
R.set(x1yLens, "baa", obj); // {"x": [0, {"y": "baa"}]}
```

davidchambers:
> ```javascript
> incLikes(commentId) {
>   this.state = R.adjustPath(R.add(1), ["comments", commentId, "likeCount"], this.state);
> }
> ```

This is a compelling example. `R.adjustPath` does seem useful.

For comparison, here's how this could be written using lenses:

```javascript
incLikes(commentId) {
  const lens = R.compose(L.comments, R.lensProp(commentId), L.likeCount);
  this.state = R.over(lens, R.add(1), this.state);
}
```

paldepind:
> ```js
incLikes(commentId) {
  this.state = R.adjustPath(R.add(1), ["comments", commentId, "likeCount"], this.state);
}
```

If using ECMAScript 6 one can also do this using `R.evolve`.
```js
incLikes(commentId) {
  this.state = R.evolve({comments: {[commentId]: {likeCount: R.add(1)}}}, this.state);
}
```

But this does seem like a natural extension to the current object functions.

whodidthis:
Nice, I was just looking for `clojure.core/update-in`

epeli:
Played today using lenses where ever I could and I noticed that when using `over` or `set` the intermediate are not created automatically if missing like in `assocPath` or the proposed `adjustPath`.

Ex.

```js
var lens = R.compose(R.lensProp("a"), R.lensProp("b"));
R.over(lens, R.assoc("foo", "val"), {});
```

throws `TypeError: Cannot read property 'b' of undefined`.

Is this by design? If so then this is another argument for this.

davidchambers:
It's interesting to consider a lens whose focus may not exist. I need to give this more thought, both from Ramda's perspective and from Sanctuary's.

scott-christopher:
In the absence of prisms and isos in Ramda for gracefully handling partial objects, you should still be able to achieve what you want with a lens such as:
```js
var partialLensProp = p => R.lens(
  R.ifElse(R.isNil, R.identity, R.prop(p)),
  (v, obj) => R.isNil(obj) ? R.createMapEntry(p, v) : R.assoc(p, v, obj)
);

var abLens = R.compose(partialLensProp("a"), partialLensProp("b"));

var emptyObj = {};
var abcBar = { a: { b: { c: "bar" }}};

console.log(R.view(abLens, emptyObj));
// undefined

console.log(R.view(abLens, abcBar));
// { c: 'bar' }

console.log(R.over(abLens, R.assoc("foo", "val"), emptyObj));
// { a: { b: { foo: 'val' } } }

console.log(R.over(abLens, R.assoc("foo", "val"), abcBar));
// { a: { b: { c: 'bar', foo: 'val' } } }
```

hzamani:
+1 for adjustPath

CrossEye:
Closed in favor of #2317

PR #1331, opened 2015-08-11
escape escape sequences in R.toString

davidchambers:
`R.toString('\n')` should evaluate to `'"\\n"'` rather than `'"\n"'` in order to uphold the invariant:

```javascript
R.equals(x, eval(R.toString(x))) === true  // for *most* values of x
```

Otherwise, we're essentially doing

```javascript
eval('"
"')
```

which is syntactically invalid.

Before:

```javascript
> R.map(R.converge(R.equals, R.identity, R.compose(eval, R.toString)),
...     ['\b', '\f', '\n', '\r', '\t', '\v', '\0', '\\', '"', "'"])
SyntaxError: Unexpected token ILLEGAL
```

After:

```javascript
> R.map(R.converge(R.equals, R.identity, R.compose(eval, R.toString)),
...     ['\b', '\f', '\n', '\r', '\t', '\v', '\0', '\\', '"', "'"])
[ true, true, true, true, true, true, true, true, true, true ]
```

One practical benefit of this change is better REPL output. Rather than the [current output][1]

```javascript
"foo
bar
"
```

we'll get

```javascript
"foo\nbar\n"
```

which matches the behaviour of the Node REPL.

@mathiasbynens's [JavaScript character escape sequences][2] proved to be an invaluable reference while working on this patch.

[1]: http://ramdajs.com/repl/?v=0.17.1#'foo%5Cnbar%5Cn'
[2]: https://mathiasbynens.be/notes/javascript-escapes

### DIFF
--- a/src/internal/_quote.js
+++ b/src/internal/_quote.js
@@ -1,3 +1,13 @@
 module.exports = function _quote(s) {
-  return '"' + s.replace(/"/g, '\\"') + '"';
+  var escaped = s
+    .replace(/\\/g, '\\\\')
+    .replace(/[\b]/g, '\\b')  // \b matches word boundary; [\b] matches backspace
+    .replace(/\f/g, '\\f')
+    .replace(/\n/g, '\\n')
+    .replace(/\r/g, '\\r')
+    .replace(/\t/g, '\\t')
+    .replace(/\v/g, '\\v')
+    .replace(/\0/g, '\\0');
+
+  return '"' + escaped.replace(/"/g, '\\"') + '"';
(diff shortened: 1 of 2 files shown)


kedashoe:
+1. Not that I care too much about performance, but I wonder if maybe it is worth checking if the string actually contains at least one backslash before running through that gauntlet of `replace`s?

davidchambers:
@kedashoe, keep in mind that `'foo\r\nbar\r\n'` doesn't actually contain any backslash characters. ;)

PR #1333, opened 2015-08-12
isTruthy - isFalsy

thefourtheye:
Introducing `isTruthy` and `isFalsy`

This are simple helper functions which will come in handy to test if
the given value is Truthy or Falsy.

### DIFF
--- a/src/isFalsy.js
+++ b/src/isFalsy.js
@@ -1,0 +1,22 @@
+var _curry1 = require('./internal/_curry1');
+
+/**
+ * Returns a boolean indicating whether the input is a Falsy value or not
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig Ord a => a -> Boolean
+ * @param {*} The JavaScript object whose falsiness has to be determined
+ * @return {*} true, if the object is falsy, false otherwise
+ * @example
+ *
+ *      R.isFalsy('')        //=> false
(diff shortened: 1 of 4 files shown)


davidchambers:
Thanks for the pull request. I don't think these functions are necessary, though.

```javascript
R.isTruthy ≍ Boolean
R.isFalsy  ≍ R.complement(Boolean)
```

thefourtheye:
Oh. Is there any function which is equivalent to `isTruthy`? I wanted to filter out a list of Truthy items from a list. But I couldn't find the proper filter function.

davidchambers:
```javascript
> R.filter(Boolean, ['', 'x', 0, 1])
[ 'x', 1 ]
> R.reject(Boolean, ['', 'x', 0, 1])
[ '', 0 ]
```

thefourtheye:
@davidchambers Ah, `Boolean`. Thanks :-)

davidchambers:
No problem. :)

thefourtheye:
@davidchambers I was using `R.identity` along with `Array.prototype.filter`, and it was working fine. I think for the `filter` case that also should be enough :-)

davidchambers:
I like to use `Boolean` rather than `R.identity` because it matches the type "required" by `R.filter`:

    R.filter :: (a -> Boolean) -> [a] -> [a]

`R.identity` works equally well, but I like to avoid implicit type coercion.

PR #1335, opened 2015-08-13
Simplify logical functions

paldepind:
This is a second attempt at simplifying Ramdas logical functions. The first attempt is #1218.

This PR makes the following changes:

* `and` and `or` does not dispatch. This didn't receive any resistance in the first PR so I want repeat the arguments.
* `and` and `or` are now documented as if they only worked on booleans. We no longer advertise that you can pass non-logical values to these logical functions. They are both documented with a truth table which is how these functions are [often defined in math](https://en.wikipedia.org/wiki/Logical_conjunction#Truth_table).
* `defaultTo` now returns it first argument if it's last argument is `NaN`.

These changes are made to achieve the following advantages:

* `and` and `or` are now semantically significantly simpler since they don't reference the quite complex and arbitrary concepts "truthy" and "falsy". The denotation (the meaning) of these functions are now equal to their counterparts in math. I believe simple denotation is a fundamental to functional programming (some people even prefer the term denotational programming to functional programming).
* `and` and `or` are now _single purpose_ functions. This makes code clearer and more explicit. This is not enforced in code, only in documentation.
* `defaultTo` now handles `NaN` this makes it more useful in many cases (for instance with `parseInt`) and removes the need to misuse `or` to achieve the same. Using `defaultTo` is also safer. Consider code like this `R.or(parseInt(student.score), 40)`. This would result in an unexpected behavior if the result of `parseInt(student.score)` returned `0`. `defaultTo` does not suffer from this problem.

I would also like to point out that the current behavior of `defaultTo` appears to [surprise people](https://gitter.im/ramda/ramda?at=55b8f72e4c04f0cc22e73104).

This PR would make it possible to replace code like this (an example by @CrossEye in the previous PR)

```js
R.map(student => R.assoc('grade', R.or(parseInt(student.score), 0), student), students);
```

With this code that better expresses the intent and doesn't rely on a multi-purpose `or`.

```js
R.map(student => R.assoc('grade', R.defaultTo(0, parseInt(student.score)), student), students);
```

### DIFF
--- a/src/and.js
+++ b/src/and.js
@@ -1,28 +1,24 @@
 var _curry2 = require('./internal/_curry2');
-var _hasMethod = require('./internal/_hasMethod');
 
 
 /**
- * A function that returns the first argument if it's falsy otherwise the second
- * argument. Note that this is NOT short-circuited, meaning that if expressions
- * are passed they are both evaluated.
- *
- * Dispatches to the `and` method of the first argument if applicable.
+ * Returns `true` if both arguments are `true`; `false` otherwise.
  *
  * @func
  * @memberOf R
  * @category Logic
  * @sig * -> * -> *
- * @param {*} a any value
- * @param {*} b any other value
- * @return {*} the first argument if falsy otherwise the second argument.
+ * @param {Boolean} a A boolean value
+ * @param {Boolean} b A boolean value
+ * @return {Boolean} `true` if both arguments are `true`, `false` otherwise
  * @see R.both
  * @example
  *
+ *      R.and(true, true); //=> true
+ *      R.and(true, false); //=> false
  *      R.and(false, true); //=> false
- *      R.and(0, []); //=> 0
- *      R.and(null, ''); => null
+ *      R.and(false, false); //=> false
  */
 module.exports = _curry2(function and(a, b) {
-  return _hasMethod('and', a) ? a.and(b) : a && b;
+  return a && b;
 });
--- a/src/defaultTo.js
+++ b/src/defaultTo.js
@@ -2,8 +2,8 @@ var _curry2 = require('./internal/_curry2');
 
 
 /**
- * Returns the second argument if it is not null or undefined. If it is null
- * or undefined, the first (default) argument is returned.
+ * Returns the second argument if it is not `null`, `undefined` or `NaN`
+ * otherwise the first argument is returned.
  *
  * @func
  * @memberOf R
@@ -19,7 +19,8 @@ var _curry2 = require('./internal/_curry2');
  *      defaultTo42(null);  //=> 42
  *      defaultTo42(undefined);  //=> 42
  *      defaultTo42('Ramda');  //=> 'Ramda'
+ *      defaultTo42(parseInt('string')); //=> 42
  */
 module.exports = _curry2(function defaultTo(d, v) {
-  return v == null ? d : v;
+  return v == null || v !== v ? d : v;
 });
--- a/test/defaultTo.js
+++ b/test/defaultTo.js
@@ -6,9 +6,10 @@ describe('defaultTo', function() {
 
   var defaultTo42 = R.defaultTo(42);
 
-  it('returns the default value if input is null/undefined', function() {
+  it('returns the default value if input is null, undefined or NaN', function() {
     assert.strictEqual(42, defaultTo42(null));
     assert.strictEqual(42, defaultTo42(undefined));
+    assert.strictEqual(42, defaultTo42(NaN));
   });
 
   it('returns the input value if it is not null/undefined', function() {
--- a/src/or.js
+++ b/src/or.js
@@ -1,28 +1,25 @@
 var _curry2 = require('./internal/_curry2');
-var _hasMethod = require('./internal/_hasMethod');
 
 
 /**
- * A function that returns the first truthy of two arguments otherwise the
- * last argument. Note that this is NOT short-circuited, meaning that if
- * expressions are passed they are both evaluated.
- *
- * Dispatches to the `or` method of the first argument if applicable.
+ * Returns `true` if one or both of its arguments are `true`. Returns `false`
+ * if both arguments are `false`.
  *
  * @func
  * @memberOf R
  * @category Logic
  * @sig * -> * -> *
- * @param {*} a any value
- * @param {*} b any other value
- * @return {*} the first truthy argument, otherwise the last argument.
+ * @param {Boolean} a A boolean value
+ * @param {Boolean} b A boolean value
+ * @return {Boolean} `true` if one or both arguments are `true`, `false` otherwise
  * @see R.either
  * @example
  *
+ *      R.or(true, true); //=> true
+ *      R.or(true, false); //=> true
  *      R.or(false, true); //=> true
- *      R.or(0, []); //=> []
- *      R.or(null, ''); => ''
+ *      R.or(false, false); //=> false
  */
 module.exports = _curry2(function or(a, b) {
-  return _hasMethod('or', a) ? a.or(b) : a || b;
+  return a || b;
 });
--- a/test/and.js
+++ b/test/and.js
@@ -5,35 +5,15 @@ var R = require('..');
 
 describe('and', function() {
   it('compares two values with js &&', function() {
-    var someAr = [];
-    assert.strictEqual(R.and(1, 1), 1);
-    assert.strictEqual(R.and(1, 0), 0);
-    assert.strictEqual(R.and(true, someAr), someAr);
-  });
-
-  it('dispatches to `and` method', function() {
-    function Nothing() {}
-    Nothing.prototype.and = function() { return this; };
-
-    function Just(x) { this.value = x; }
-    Just.prototype.and = R.identity;
-
-    var n1 = new Nothing();
-    var n2 = new Nothing();
-    var j1 = new Just(1);
-    var j2 = new Just(2);
-
-    assert.strictEqual(R.and(n1, n2), n1);
-    assert.strictEqual(R.and(n2, n1), n2);
-    assert.strictEqual(R.and(n1, j2), n1);
-    assert.strictEqual(R.and(j2, n1), n1);
-    assert.strictEqual(R.and(j1, j2), j2);
-    assert.strictEqual(R.and(j2, j1), j1);
+    assert.strictEqual(R.and(true, true), true);
+    assert.strictEqual(R.and(true, false), false);
+    assert.strictEqual(R.and(false, true), false);
+    assert.strictEqual(R.and(false, false), false);
   });
 
   it('is curried', function() {
     var halfTruth = R.and(true);
     assert.strictEqual(halfTruth(false), false);
-    assert.strictEqual(halfTruth('lie'), 'lie');
+    assert.strictEqual(halfTruth(true), true);
   });
 });
--- a/test/or.js
+++ b/test/or.js
@@ -5,36 +5,14 @@ var R = require('..');
 
 describe('or', function() {
   it('compares two values with js &&', function() {
-    var someAr = [];
-    assert.strictEqual(R.or(1, 0), 1);
-    assert.strictEqual(R.or(0, 1), 1);
-    assert.strictEqual(R.or(someAr, false), someAr);
-    assert.strictEqual(R.or('', 0), 0);
-  });
-
-  it('dispatches to `or` method', function() {
-    function Nothing() {}
-    Nothing.prototype.or = R.identity;
-
-    function Just(x) { this.value = x; }
-    Just.prototype.or = function() { return this; };
-
-    var n1 = new Nothing();
-    var n2 = new Nothing();
-    var j1 = new Just(1);
-    var j2 = new Just(2);
-
-    assert.strictEqual(R.or(n1, n2), n2);
-    assert.strictEqual(R.or(n2, n1), n1);
-    assert.strictEqual(R.or(n1, j2), j2);
-    assert.strictEqual(R.or(j2, n1), j2);
-    assert.strictEqual(R.or(j1, j2), j1);
-    assert.strictEqual(R.or(j2, j1), j2);
+    assert.strictEqual(R.or(true, true), true);
+    assert.strictEqual(R.or(true, false), true);
+    assert.strictEqual(R.or(false, true), true);
+    assert.strictEqual(R.or(false, false), false);
   });
 
   it('is curried', function() {
-    assert.strictEqual(R.or('lie')(false), 'lie');
+    assert.strictEqual(R.or(false)(false), false);
     assert.strictEqual(R.or(false)(true), true);
-    assert.strictEqual(R.or('')(0), 0);
   });
 });


src/defaultTo.js
```diff
@@ -19,7 +19,8 @@ var _curry2 = require('./internal/_curry2');
  *      defaultTo42(null);  //=> 42
  *      defaultTo42(undefined);  //=> 42
  *      defaultTo42('Ramda');  //=> 'Ramda'
+ *      defaultTo42(parseInt('string')); //=> 42
  */
 module.exports = _curry2(function defaultTo(d, v) {
-  return v == null ? d : v;
+  return v == null || v !== v ? d : v;
```
TheLudd: I'd prefer `isNaN` here instead of `v !== v`. Is there a reason why this is favorable?

paldepind: `isNaN` is broken (`isNaN({})` for instance is `true`) and `Number.isNaN` lacks browser support. `NaN` is the only value that isn't equal to itself so this is AFAIK the simplest way to check for it.

davidchambers:
I like the changes to `R.and` and `R.or`. I don't have a problem with the change to `R.defaultTo`, though `R.defaultTo` and `R.propOr` have always seemed to me like (useful) half measures.

Here's Scott's example written *without* `R.defaultTo`:

```javascript
R.map(R.converge(R.assoc('grade'),
                 R.pipe(S.get('score'),
                        R.chain(S.parseInt(10)),
                        S.fromMaybe(0)),
                 R.identity),
      students);
```

src/and.js
```diff
@@ -1,28 +1,24 @@
...
-var _hasMethod = require('./internal/_hasMethod');
 
 
 /**
- * A function that returns the first argument if it's falsy otherwise the second
- * argument. Note that this is NOT short-circuited, meaning that if expressions
- * are passed they are both evaluated.
- *
- * Dispatches to the `and` method of the first argument if applicable.
+ * Returns `true` if both arguments are `true` and otherwise `false`.
```
davidchambers: I suggest this tweak:

```diff
- * Returns `true` if both arguments are `true` and otherwise `false`.
+ * Returns `true` if both arguments are `true`; `false` otherwise.
```

src/and.js
```diff
@@ -1,28 +1,24 @@
...
  *
  * @func
  * @memberOf R
  * @category Logic
  * @sig * -> * -> *
- * @param {*} a any value
- * @param {*} b any other value
- * @return {*} the first argument if falsy otherwise the second argument.
+ * @param {Boolean} a A boolean value
+ * @param {Boolean} b A boolean value
```
davidchambers: These descriptions are redundant. Let's nix 'em. :)

src/defaultTo.js
davidchambers: I'm fine with this. The other option is to encapsulate the `v !== v` logic in an internal `_isNaN` function.

src/defaultTo.js
```diff
@@ -2,8 +2,8 @@ var _curry2 = require('./internal/_curry2');
 
 
 /**
- * Returns the second argument if it is not null or undefined. If it is null
- * or undefined, the first (default) argument is returned.
+ * Returns the second argument if it is not `null`, `NaN` or `undefined`
+ * otherwise the first arguments is returned.
```
davidchambers: I would list the "invalid" values in one of the following orders:

  - `[null, undefined, NaN]`
  - `[NaN, null, undefined]`

It seems nice to keep `null` and `undefined` together, since we often say "null or undefined" (and we have a function which considers both values "nil").

Also, there's a typo on the line above:

```diff
- * otherwise the first arguments is returned.
+ * otherwise the first argument is returned.
```

test/defaultTo.js
```diff
@@ -6,9 +6,10 @@ describe('defaultTo', function() {
 
   var defaultTo42 = R.defaultTo(42);
 
-  it('returns the default value if input is null/undefined', function() {
+  it('returns the default value if input is null, NaN undefined', function() {
```
davidchambers:
```diff
-  it('returns the default value if input is null, NaN undefined', function() {
+  it('returns the default value if input is null, undefined, or NaN', function() {
```

src/and.js
paldepind: Nix them how? Aren't they required to be there?

paldepind:
Thank you for your comments @davidchambers. I have updated the PR.

src/defaultTo.js
TheLudd: There is also `R.identical(NaN, v)`

paldepind: Yes. I did consider that. But it seems like a bigger overhead to me. Then I'd rather have an internal `_isNaN` and make `R.identical` use it as well. I guess it depends on how familiar the `v !== v` is to people.

buzzdecafe:
i recommend waiting for @CrossEye to return before merging this one, he had some reservations last time.

paldepind:
@buzzdecafe Yes. I agree, of course.

src/and.js
buzzdecafe: i think you can just have:

```
* @param {Boolean} a
* @param {Boolean} b
```

davidchambers: Yes. That's what I had in mind. :)

CrossEye:
Thanks for waiting.  I'm afraid I may once again be the fly in the ointment here.

I have much the same objection I had in #1218.  Since this does not change the behavior of `and` and `or`, but only their documentation -- well, except for removing the dispatching; I'm not particularly attached to dispatching these functions, but don't really see a reason to remove it now that it's here, either -- since this is a smaller change, my objection is perhaps less strenuous.  But it's the same objection.

The goal of these functions, as I see it, is to provide pure function versions of the built-in operators of the language.  Documenting them as something else because that something else is logically simpler misses the point.

The functions already serve fine as `Bool -> Bool -> Bool`, following the relevant truth-tables.  Anyone who wants a function to combine two boolean values with a logical `or` or `and` can use these functions without concern.  But the functions generalize on this notion in a useful manner.  Note that my descriptions of `and`, "returns the first falsey value or the last value," and of `or`, "returns the first truthy value or the last value" perfectly well describe boolean truth tables:

    and(T, T) => T, and(T, F) => F
    and(F, T) => F  and(F, F) => F

    or(T, F) => T,  or(T, F) => T
    or(F, T) => T,  or(F, F) => F

But they also serve to describe the additional features in Javascript's `&&` and `||` operators.

I simply don't see any reason to try to restrict these function, either in practice in #1218 or in documentation here, to a smaller domain.

I haven't really thought hard about `defaultTo` yet, but it seemed as it that was done mostly to satisfy my desires for `||`-like functionality.  While I don't object, I wonder where the paired `&&`-like functionality is.  Moreover, unless we change `and` and `or`, this doesn't really seem all that necessary, but I may be misunderstanding this part.

paldepind:
> The goal of these functions, as I see it, is to provide pure function versions of the built-in operators of the language.

Why is that the goal? What about `R.equals` and `R.identical`? They are very much like built-in operators but yet our implementation is different.

> I simply don't see any reason to try to restrict these function, either in practice in #1218 or in documentation here, to a smaller domain.

Because using `R.or` and `R.and` with anything but booleans results in code that is

* Harder to understand – because they are now double purpose. They can be used _either_ as the logical operators we know from math _or_ as a form of default to fallback value mechanism. Whenever one sees an `R.or` one has to spend time and energy on determining which case one is looking at. What comes in and out? It could be _anything_ (there is a * in the type signature).
* Error prone – Using `R.or` with anything but booleans is notoriously error prone. David had a splendid rundown [here](https://github.com/ramda/ramda/issues/843#issuecomment-77297120) of bugs in Ramda cause by them.

It is that simple. Ramda should _not_ promote writing hard to understand error prone code. We should promote precise simple functions where each _do one thing_. And if one looks at Ramda it is very much what it does in most of its functions.

CrossEye:
>> The goal of these functions, as I see it, is to provide pure function versions of the built-in operators of the language.

> Why is that the goal? What about `R.equals` and `R.identical`? They are very much like built-in operators but yet our implementation is different.

Why should the goal be the much more restricted one of providing boolean-only `and` / `or`?

What we have is a very simple definition, and a very powerful one.  The language's `&&` and `||` operators are among the good parts of the language.  Almost every language out there returns the first falsey or the last value for `and`, for that language's equivalent of falsey.  It is also the definition familiar to JS developers.

As to `equals` and `identical`, note that we renamed `eq` to `equals`; it should be less likely to be confused with the operator, and I, for one, would certainly never confuse `identical` with an operator.   

>> I simply don't see any reason to try to restrict these function, either in practice in #1218 or in documentation here, to a smaller domain.

> Because using `R.or` and `R.and` with anything but booleans results in code that is
> 
> * Harder to understand – because they are now double purpose. They can be used _either_ as the logical operators we know from math _or_ as a form of default to fallback value mechanism. 

This is not a different, separate behavior; it's simply a higher abstraction.

> Whenever one sees an `R.or` one has to spend time and energy on determining which case one is looking at. What comes in and out? It could be _anything_ (there is a * in the type signature).

That type signature should be changed, though.  `a -> a -> a` is more sensible.  `a -> b -> a | b` is precise.

> * Error prone – Using `R.or` with anything but booleans is notoriously error prone. David had a splendid rundown [here](https://github.com/ramda/ramda/issues/843#issuecomment-77297120) of bugs in Ramda cause by them.

Most of those, though were about changes to behavior, not actually bugs.  (#505 and #507 are clearly bugs; the others are more questionable.)  That a powerful tool can be misused is an argument for not including it; we've all seen foot-guns.  But the evidence for that has to be overwhelming; else most powerful tools would probably never pass muster.  Should Ramda, for instance, clone the accumulation object passed through `reduce` because there's a chance that a user would break something by mutating it along the way?  Our answer to these has generally been no.  This is a library for grown-ups, and we answer "Doctor, it hurts when I do this," with, "Then, don't do that!"


> It is that simple. Ramda should _not_ promote writing hard to understand error prone code. We should promote precise simple functions where each _do one thing_. And if one looks at Ramda it is very much what it does in most of its functions.

Like `map` for any functor, where it performs that functor's `map` operation on the supplied function and the functor instance, or where `map` calls the supplied function on each element in a list returning a list made up of the respective results?  From one perspective, this is two very different behaviors.  But at a higher level, it is _exactly the same thing_.  That's what I am suggesting here.

Even it it wasn't, though, it's the JS behavior that I would want to reify.

davidchambers:
> What we have is a very simple definition, and a very powerful one. The language's `&&` and `||` operators are among the good parts of the language.

Powerful? Yes. Simple? No.

```javascript
x || y
```

expands to

```javascript
(x === false || x === null || x === undefined || x === 0 || x !== x || x === '') ? y : x
```

This is almost certainly not what the author of `x || y` *meant*. Perhaps she meant

```javascript
Number.isNaN(x) ? y : x
```

or maybe

```javascript
x === '' ? y : x
```

Much mental effort is expended by readers determining, from surrounding context, which of the six values are *intentionally* matched and which are *inadvertently* matched. This could have been avoided by avoiding the `||` "macro".

In my view `&&` and `||` are far from simple. Furthermore, it's impossible to capture their evaluation semantics in a function, so `R.and` and `R.or` will always be significantly different from their operator counterparts.

One thing we haven't discussed is the possibility of aligning these functions with `R.empty` and `R.isEmpty`. That is something I could support, since at least then both `''` and `[]` would be treated as empty/false values.

Given a choice between the current semantics and the semantics proposed in this pull request, I remain firmly in favour of the latter.

CrossEye:
>> What we have is a very simple definition, and a very powerful one. The language's `&&` and `||` operators are among the good parts of the language.

> Powerful? Yes. Simple? No.

> ```javascript
> x || y
> ```
> 
> expands to
> 
> ```javascript
> (x === false || x === null || x === undefined || x === 0 || x !== x || x === '') ? y : x
> ```

I think this misses the point.  This is a more appropriate expansion.

> ```javascript
> x || y ; //~~> isFalsey(x) ? y : x
> ```

Granted, one would have to expand `isFalsey` if desired.  But that is the logical expansion of `||`.  And I think that the above is probably not the best expansion for `isFalsey`, recursively depending on `||`.

By the same token, one appropriate expansion of a plain logical `and` and `or` are:

```js
and(x, y) //~~> (x === false)  ? false : (y === true)
or(x, y) //~~> (x === false)  ? (y === true) : false
```

This makes the parallel clear:

```js
and(x, y) //~~> isFalsey(x)  ? x : y
or(x, y) //~~> isFalsey(x) ? y : x
```

> Furthermore, it's impossible to capture their evaluation semantics in a function, so `R.and` and `R.or` will always be significantly different from their operator counterparts.

This is the most important point here.  In some sense, this would be a good reason to revert #843, although I really don't want to go there.

#1218 attempted to reduce the capabilities of two of our functions.  This one mostly does the same via documentation, without actually changing the function, and also expands the capabilities of a third function to cover the missing cases of `or`, but simply ignores the existing abilities of `and`.  It claims to make code easier to understand, but for code that supplies the only types that are documented, makes _no difference in behavior whatsoever_.  How is this progress?

> Much mental effort is expended by readers determining, from surrounding context, which of the six values are *intentionally* matched and which are *inadvertently* matched. This could have been avoided by avoiding the `||` "macro".

How does this PR address that concern at all?  It does not stop people from using `||`, nor does it change the semantics of `or`, only replacing the documentation of how it actually works with a gloss on how we would prefer people use it.

paldepind:
> I think this misses the point.

No. It is right on point. The ad-hoc arbitrary concepts of truthy and falsey is exactly the root cause of the problem.

IMO no function at all in Ramda should depend on them. But I have given up hope.

kedashoe:
> But I have given up hope.

What about

> One thing we haven't discussed is the possibility of aligning these functions with R.empty and R.isEmpty. That is something I could support, since at least then both '' and [] would be treated as empty/false values.

? Seems like a nice suggestion to me that fits with Ramda. Strictly boolean `and` and `or` functions as in your previous pull request, and `firstEmpty` and `unEmpty` (except with better names) for the truthy/falsey behaviour of the `&&` and `||` operators.

CrossEye:
@paldepind:

> But I have given up hope.

That's funny.  I thought I was the one fighting a rear-guard action here.

@kedashoe: I think this might be the best bet, but I still expect a big discussion over who gets to keep the good names.   :smile:

paldepind:
> By the same token, one appropriate expansion of a plain logical and and or are:

> and(x, y) //~~> (x === false)  ? false : (y === true)
or(x, y) //~~> (x === false)  ? (y === true) : false

Proper logical functions just look like this:

```js
and(x, y) // x === false ? x : y
or(x, y) // x === true ? x : y
```

If you can only pass booleans to the functions then that is what they does. Notice that they still return the first or second argument but without the concept of truthy and falsey that is not an issue.

> How does this PR address that concern at all? It does not stop people from using ||, nor does it change the semantics of or, only replacing the documentation of how it actually works with a gloss on how we would prefer people use it.

It's a compromise. IMO we could throw on non-booleans in `or` and `and` but I know that is unlikely to happen. If I was using Ramda in a corporate setting I'd want that.

> Seems like a nice suggestion to me that fits with Ramda. Strictly boolean and and or functions as in your previous pull request, and firstEmpty and unEmpty (except with better names) for the truthy/falsey behaviour of the && and || operators.

That would be fine with me. But I actually don't think there is a use case for an `unEmpty`. That is also why this PR only tries to provide an alternative to `or`. I often see people using `||` to provide fallback values but I don't see people use `&&` to get anything but a truth value out. I could be wrong though.

> I think this might be the best bet, but I still expect a big discussion over who gets to keep the good names. :smile:

What do you mean? `or` and `and` of course should be the logical operators. They match every day usage of the words "John took out the trash _or_ it is still in the garage". No one says "I'd like the blue bucket or the red bucket" to mean "I'd like the blue bucket unless it is empty, otherwise the red".

CrossEye:
> Proper logical functions just look like this:
> 
> ```js
> and(x, y) // x === false ? x : y
> or(x, y) // x === true ? x : y
> ```

That's funny, as I feel it entirely demonstrates my point, that Javascript's `&&` and `||` are clear-cut generalizations of this notion.  If you said one of these instead, it's a little harder to demonstrate, although they are all equivalent:

```js
and(x, y) // x === false ? false : y === true ? true : false
or(x, y) // x === true ? true : y === true ? true : false
```
or 

```js
and(x, y) // x === false ? false : y === true
or(x, y) // x === true ? true : y === true
```

> It's a compromise. IMO we could throw on non-booleans in `or` and `and` but I know that is unlikely to happen. If I was using Ramda in a corporate setting I'd want that.

To me exceptions are a very last resort.  Only if there's no reasonable way to handle user input would I choose an exception.  Because they don't compose, and force all sorts of craziness into a codebase, I choose to avoid them as much as possible, especially in my corporate setting!  :smile:

>  I often see people using `||` to provide fallback values but I don't see people use `&&` to get anything but a truth value out. I could be wrong though.

I see plenty of things like this:

```js
var city = account && account.address && account.address.city; //=> undefined | City
```

Granted Ramda has some nicer tools to deal with this, but that's a very common non-boolean usage of `&&`.

>> I think this might be the best bet, but I still expect a big discussion over who gets to keep the good names. :smile:

> What do you mean? `or` and `and` of course should be the logical operators. They match every day usage of the words "John took out the trash _or_ it is still in the garage". 

What do you mean?  `or` and `and` of course should be the language operators.  They match every day pronunciation of the code `trash.isOut() || garage.contains(trash)`.  

:smile:

> But I'm happy to hear that there is hope :)

There is, but I'm still somewhat reluctant in this whole endeavor.  I'm very much reminded of [a comment][co] I made a few months ago:

>>> I'm starting to feel fairly lonely in the stand I'm taking, but I do still feel that Ramda should not be particularly prescriptive about how users write their Javsacript. To my mind, it's supposed to be a library that works with the grain of the language. It's [ ... ] not designed to dictate too strongly how users could and should write their code.

It's the same attitude that makes me see no point in changing a function, which already does what you want when used the way you suggest, to restrict it so that others cannot use it in other ways, simply because you find that style of coding problematic.  Even doing this simply through documentation bothers me.  I really don't think Ramda should be that prescriptive.

  [co]: https://github.com/ramda/ramda/issues/1107#issuecomment-106009092

paldepind:
> That's funny, as I feel it entirely demonstrates my point, that Javascript's && and || are clear-cut generalizations of this notion.

paldepind:
> That's funny, as I feel it entirely demonstrates my point, that Javascript's && and || are clear-cut generalizations of this notion.

They are "generalizations" that only makes sense if you have a concept of truthy and falsey. Note my next sentence (emphasis mine):

> Notice that they still return the first or second argument but _without the concept of truthy and falsey that is not an issue_.

Without falsey and truthy returning the argument is an implementation detail and as you showed they can be implemented without returning any of the arguments. I.e. it is not a part of the semantics of the function only a part of the operation of the function. That is a big difference.

> I see plenty of things like this:

I've honestly never seen code like that. But I'd not be surprised if it's common.

> Granted Ramda has some _nicer_ tools to deal with this, but that's a very common non-boolean usage of &&.

(emphasis mine). I'm glad you acknowledge that there is nicer ways to achieve that without misusing logical operators.

> What do you mean? or and and of course should be the language operators. They match every day pronunciation of the code `trash.isOut() || garage.contains(trash)`. 

Are you saying that `isOut` and `contains` does not return booleans? Because if they do you're saying the exact same thing as me (which doesn't seem to be the intention) and if they don't then you have some horribly named method names.

> It's the same attitude that makes me see no point in changing a function, which already does what you want when used the way you suggest, to restrict it so that others cannot use it in other ways, simply because you find that style of coding problematic. Even doing this simply through documentation bothers me. I really don't think Ramda should be that prescriptive.

If you think about it that is what Ramda does _all the time_. Why do we say no to people who wants to be able to do `R.path('foo.bar')`? You can implement that along with the current behavior. Who do we say no to `R.foldObj`? Having it wouldn't hurt those who don't like it? Why can't I pass constants to `R.cond` to avoid using `R.always`? Doesn't that sound exactly like this

> to restrict it so that others cannot use it in other ways, simply because you find that style of coding problematic

I simply don't understand why being prescriptive in this case bothers you when you are a huge proponent of it in so many other cases.

CrossEye:
>> I really don't think Ramda should be that prescriptive.

> If you think about it that is what Ramda does _all the time_. 

Ramda provides tools to make it easier to program in a functional manner in JS, especially composable utility functions.  It has a fairly strong set of rules for its own functions, stressing things like API simplicity and immutability of input data.  But it doesn't try to impose particular discipline on its users.  Most of its function decorators maintain the execution context, allowing users to use Ramda-created functions as OO-methods if they choose; we don't enforce FP even here.  I love Sanctuary, but the idea of `head :: [a] -> Maybe(a)` is the type of thing that I believe Ramda should not be insisting upon.  If users want to risk bad data, and accept the possibility of `null` / `undefined` values, it is the users' prerogative to do so.  There is a really good argument to be made for the mathematical purity of only working with unary functions.  But that is too prescriptive for Ramda.  We have no place telling our users that they should only be using unary functions; our function-generating functions should support polyadic outputs.  We have perfectly serviceable `and` / `or` functions, ones that work properly as binary boolean operators, but ones that also generalize in a certain direction to act similarly to the closest equivalent operators in the language.  To say that users are not smart enough to know when to use them with other than boolean values is too prescriptive.

> Why do we say no to people who wants to be able to do `R.path('foo.bar')`? You can implement that along with the current behavior. 

It breaks our rule on simplicity.  I actually didn't want to change from this style API to the current one because this one is easier; it was hard to give up; it is how I've done similar features in my own projects.  But in a library, the robustness of the current solution, and the true simplicity it entails outweighs the easiness and apparent simplicity of the string-base solution.

Note that this could have gone the other way.  If we were to insist that the key functionality that we desired was to access a node of a nested object graph using a dot-separated string, then `R.path['foo', 'bar']` doesn't meet the spec.  But we weren't that specific here.  I think this is the main point of contention between you and me here.  I feel the fundamental goal here is to reify the language operators, and I believe you see it as to provide logical boolean operators.  That makes it hard to come together.

> Who do we say no to `R.foldObj`? Having it wouldn't hurt those who don't like it? 

Because it doesn't fit well with the philosophy of Ramda.  We can't have useful properties such as

```js
equals(obj1, obj2); 
//~~> 
equals(foldObj(fn, init, obj1), foldObj(fn, init, obj2));
```

Things that can't be made lawful, things that break standard FP concerns such as the irrelevance of order of operations, these should not become Ramda functions.


> Why can't I pass constants to `R.cond` to avoid using `R.always`? Doesn't that sound exactly like this

Again, API simplicity.  Our functions have a single, simple signature.  Sometimes it's harder to see this, when we manually handle the list case and delegate others, but that's the basic idea.  `map` works on any Functor instance, including, through our bespoke code, a list.


>> to restrict it so that others cannot use it in other ways, simply because you find that style of coding problematic

> I simply don't understand why being prescriptive in this case bothers you when you are a huge proponent of it in so many other cases.

I'm very concerned about what Ramda does, especially in its API, but also in its internal implementation.  I'm not concerned about how Ramda users choose to write their code.

I think that distinction is reasonable.  Do you agree?

paldepind:
> To say that users are not smart enough to know when to use them with other than boolean values is too prescriptive.

But it is fine to say that they are not smart enough to use `reduceObj`?

>  I feel the fundamental goal here is to reify the language operators, and I believe you see it as to provide logical boolean operators. That makes it hard to come together.

Well, `||` and `&&` are logical boolean operators. But sadly we inherited them from a language that didn't have booleans and thus some harmful historical cruft came with them. C didn't have booleans so they had to use 0 for false. There are no _good_ reasons for such semantics.

> We can't have useful properties such as

What about this properties?

```js
equals(R.and(p, q), true); 
//~~> 
equals(p, true)
equals(q, true)
```

It seems nice to immediately be able to infer the types of `p` and `q`.

> things that break standard FP concerns

A standard FP concern is simple denotation as I mentioned in my first post. Truthy and falsey breaks that.

> It breaks our rule on simplicity.

So does truthy and falsey. You wrote "simple" six times in your last post. So you care a lot about it. I don't understand why the complexity of truthy and falsey somehow doesn't matter.

> I'm not concerned about how Ramda users choose to write their code.

But you are concerned about what tools you provide them. And since `reduceObj` is problematic you want let them have it. Neither should we give them truthy and falsey dependent logical operators.

> I think that distinction is reasonable. Do you agree?

No. I don't see it. It's the same to me. It's all about simple semantics that makes it hard to shoot oneself in the foot.

@davidchambers 

> One thing we haven't discussed is the possibility of aligning these functions with R.empty and R.isEmpty. That is something I could support, since at least then both '' and [] would be treated as empty/false values.

I'm not sure I see how that would make for an alternative to `R.defaultTo(0, parseInt(student.score))`?

CrossEye:
> But it is fine to say that they are not smart enough to use `reduceObj`?

No, the problem is that there is no way to make a reasonable `reduceObj` that follows reasonable laws.  If supplying equivalent values to a function yields differing results, then the function is not pure, and to my mind has no place in Ramda.

In Underscore, you have behavior like this:

```js
var obj1 = {};
obj1.a = 1;
obj1.b = 2;

var obj2 = {};
obj2.b = 2;
obj2.a = 1;

var diff = (a, b) => b - a;

_.isEqual(obj1, obj2); //=> true
_.reduce(obj1, diff, 0); //=> 1
_.reduce(obj2, diff, 0); //=> -1
```

That won't fly for Ramda.  There are, of course, options.  We could define our own property iteration order.  But that would likely surprise users, and it's certainly not somewhere I really want to go, not for the benefit of being able to supply what strikes me as a fairly meaningless function.  (While it makes sense to map arbitrary objects, I can't see a reason to try to fold them; folding has to do with ordering and objects don't have intrinsic order to them, even if an iteration order is [specified][sp].)

So this is not about users not being smart enough to use such a function, but about it being impossible to create a function that meets Ramda's standards.

Further discussion about this should probably take place in #1067.


>>  I feel the fundamental goal here is to reify the language operators, and I believe you see it as to provide logical boolean operators. That makes it hard to come together.

> Well, `||` and `&&` are logical boolean operators. But sadly we inherited them from a language that didn't have booleans and thus some harmful historical cruft came with them. C didn't have booleans so they had to use 0 for false. There are no _good_ reasons for such semantics.

In the dynamically-typed JS, many find being able write this:

```js
if (test) { /* ... */ }
```

instead of 

```js
if (test !== null && test !== undefined && test !== '') { /* ... */ }
```

a significant advantage.  I have to admit to being one of them.  There are certainly mistakes in what was chosen as falsey values -- I would certainly have included empty arrays / empty objects -- but I'm not at all unhappy with the basic ideas.  That this carries over to `&&` and `||` is then unsurprising.


>> It breaks our rule on simplicity.

> So does truthy and falsey. You wrote "simple" six times in your last post. So you care a lot about it. I don't understand why the complexity of truthy and falsey somehow doesn't matter.

We discussed this [in #1218][si].  Simplicity comes into play more with _how_ we accomplish something than with _what_ we choose to do.  If the goal is my notion of reifying the language operators, then there is no _additional_ complexity in truthy/falsey, only the _inherent_ complexity of the goal.  Obviously if the goal is different, we have other results.  But choosing goals chiefly on the basis of simplicity leads to doing less and less.


>> I'm very concerned about what Ramda does, especially in its API, but also in its internal implementation. I'm not concerned about how Ramda users choose to write their code.
>> 
>> I think that distinction is reasonable. Do you agree?

> No. I don't see it. It's the same to me. It's all about simple semantics that makes it hard to shoot oneself in the foot.

This is quite fundamental.  And it scares me a bit.  A library that dictates how users should write their code is not much different from a framework.  That is at the far end of the spectrum from where I've always envisioned Ramda operating.



  [si]: https://github.com/ramda/ramda/pull/1218#issuecomment-114631316
  [sp]: http://www.ecma-international.org/ecma-262/6.0/#sec-ordinary-object-internal-methods-and-internal-slots-ownpropertykeys

paldepind:
> So this is not about users not being smart enough to use such a function, but about it being impossible to create a function that meets Ramda's standards.

Any function using falsey and truthy doesn't meet those standards either IMO.

> Further discussion about this should probably take place in #1067.

I am not arguing for `reduceObj`. I am using it as an example.

> In the dynamically-typed JS, many find being able write this:

I find such code horrible. I look at it and I have no idea about what `test` is. Ramda and other utility libraries has plenty functions for writing precise code instead of that.

> Simplicity comes into play more with how we accomplish something than with what we choose to do.

Yes. And the proposed `and`, `or` and `defaultTo` accomplishes what the current `and` and `or` does but in a simpler and more explicit way.

Any way. This is going nowhere. How do we arrive at a decision with regards to this matter?

jdalton:
>> To say that users are not smart enough to know when to use them with other than boolean values is too prescriptive.

> But it is fine to say that they are not smart enough to use `reduceObj`?

Or
```js
// smart enough to handle
var o1 = { 'a': 1, 'b': 2 };
var o2 = { 'b': 2, 'a': 1 };

R.toString(o1); // '{"a": 1, "b": 2}'
R.toString(o2); // '{"a": 1, "b": 2}'

// but then...

R.keys(o1); // ["a", "b"]
R.keys(o2); // ["b", "a"]

R.toPairs(o1); // [["a",1],["b",2]]
R.toPairs(o2); // [["b",2],["a",1]]

R.values(o1); // [1, 2]
R.values(o2); // [2, 1]

R.mapObj(console.log.bind(console), o1); // logs 1 then 2
R.mapObj(console.log.bind(console), o2); // logs 2 then 1

var o1 = { get a() { console.log('a'); return 1; }, get b() { console.log('b'); return 2; } };
var o2 = { get b() { console.log('b'); return 2; }, get a() { console.log('a'); return 1; } };

R.merge(o1, { 'c': 3 }); // logs "a" then "b" and produces {'a': 1, 'b': 2, 'c': 3}
R.merge(o2, { 'c': 3 }); // logs "b" then "a" and produces {'b': 2, 'a': 1, 'c': 3}
```

davidchambers:
> > One thing we haven't discussed is the possibility of aligning these functions with `R.empty` and `R.isEmpty`. That is something I could support, since at least then both `''` and `[]` would be treated as empty/false values.
>
> I'm not sure I see how that would make for an alternative to `R.defaultTo(0, parseInt(student.score))`?

It wouldn't. I'd write the above as

```javascript
// score :: { score :: String } -> Number
const score =
R.pipe(R.prop('score'),
       R.flip(parseInt)(10),
       R.when(Number.isNaN, R.always(0)));
```

if using Ramda alone, or as

```javascript
// score :: { score :: String } -> Number
const score =
R.pipe(S.get('score'),
       R.filter(R.is(String)),  // remove if we'd prefer to throw if value is not of the correct type
       R.chain(S.parseInt(10)),
       S.fromMaybe(0));
```

if using Sanctuary as well.

I strongly dislike the idea of these sets:

```javascript
{false, null, undefined, 0, NaN, ''}  {<all other values>}
```

The set on the left is a grab bag of values of different types. So, by extension, is the set on the right.

On the other hand I'm comfortable with these sets:

```javascript
{''}  {<all other string values>}
{[]}  {<all other array values>}
```

That's the point I was trying to make with the comment quoted above. It's not a proposal: it's just an observation. Monoid? Good. Grab bag? Bad.

paldepind:
> I strongly dislike the idea of these sets:

I completely agree.

I think a point can be made for `{null, undefined, NaN}` as they are all a form of useless values. JavaScript has too many of those but there's nothing we can do about that.

Not that this PR does not change the behavior of `R.defaultTo` because I've had a need for it myself. It's changed because some users seemed to expect that behavior. And because I hoped that seeing how it is a much better solution than misusing `or` would help getting the change to `or` through.

CrossEye:
I want to apologize. I thought I'd have plenty of time to write this weekend, but I ended up spending most of my time on home improvement projects and have not have had the time to write. Most of my thoughts are in order, but it will still take me a bit of time to write them up. I will get to it soon.

CrossEye:
I've spent a lot of time thinking through this issue and the issues surrounding a potential `reduceObj` function.  My position has not changed drastically on either, but I will try -- here and there -- to summarize what I think, and why.  I'm not attempting to shut off debate on either topic but to stop something I do too often: over-engaging in back-and-forth reflexive argument on these issues.  If I don't respond to someone's points, it does not mean either that I agree or that I disagree, only  that I have already had my say.  I will try to respond if someone is succeeding in changing my mind, though.

This issue is the simpler one for me. First of all, I'm concerned less about it.  If everyone wants to switch to a boolean-only version of `and` and `or`, I won't try to stand in the way.  I would rather not be half-hearted about it, preferring #1218 to this PR.  But really, I would rather simply revert #843.  I would prefer to use `and` and `or` -- if they're not going to represent the language's operators -- as a way to combine _functions_ that return booleans than as a way to combine booleans.  That is the main use I have for any of these functions in the first place:  `var healthyBMI = and(gte(__, 18.5), lt(__ 25))`.  I think that reads better than '`both`'.

However, most of all, I would prefer to keep the current behavior.

@paldepind has been arguing that 'truthy' and 'falsey' are incoherent, and that therefore supporting their usage in these functions is harmful.

I feel precisely the opposite.  I find the notion of 'falsey' in Javascript to be one of the very good parts of the language.  It's not perfect; in fact, it's annoyingly wrong, as falsey should also include empty objects and empty arrays.  But it's extremely useful, and I don't find that I ever get tripped up by the sorts of issues that seem to worry @paldepind.  I also work with and help mentor a very large team, often including as many as 30 developers with hands in the Javascript codebase, from rank beginners to very experienced.  There are all sorts of problems that come up on code review, but this is not a significant one.

I don't use `and` and `or` much.  I'm substantially more likely to use `both` and `either` in the current incarnation.  But I was willing to change the names in #843 as this would allow us to offer approximations to more of the language's operators as Ramda functions.  Early on, I was sure that we eventually wanted to reify not only all   the operators but all the methods on all the built-in prototypes.  Eventually, I came to realize that there was simply no reason for that, but I still have a desire to be able to create expressions with pure functions rather than one-off lambdas.  So these functions as operators make sense to me.

Having these functions as boolean operators also makes sense, of course.  There's no denying that there are times when one wants to use versions of `and` and `or` that match the logical truth tables.

_But these functions already do that!_  This is why I have had such a strong objection to this PR and to #1218 before it..  They do not increase the simplicity of ones own code: if you pass only booleans to `and`/`or`, you will get the expected booleans back.  They do not increase the simplicity of the implementation: that's a wash.  They do not increase the simplicity of the actual API: I would argue that they actually increase it, more on this just below.  The only way they increase the simplicity is in the documented API.  And that's a horrible trade-off.

As to the actual API: since we are in an untyped language, there is no way to enforce that what the user passes in is really a boolean, without type-checking and choosing to throw an exception on mismatches -- definitely not the Ramda way.  But if the user can still pass other types, and we do our own coercion with code like `if x === true`, then our code, assuming `not` is changed to match:  `not == (x => x !== true)`, fails to support even the very fundamental rules of logic, such as [De Morgan's Laws][dm]: 

    ((x, y) => and (x, not(y))(42) ==> false
    ((x, y) => not(or(not(x), y)))(42) ==> true

This sort of failure tells me that functions aren't even doing what's proposed!  And what's proposed seems designed only to appeal to someone's notions of purity, not for any real practical benefit.  So, while I won't really fight hard, it still feels absurd to me.

But the worse thing to me about either of these proposals is the paternalistic feel to them.  No one is pointing to problems wrought by the current incarnations of `and` or `or`.  This change will not allow anyone to write anything they couldn't before.  And in championing it, the main point is that these functions should only be passed booleans, and should operate as boolean operators.  But presumably those who feel that way, already use the functions that way, and will gain nothing themselves.  All they can do is to prevent others who might prefer the current behavior from being allowed to have it -- because of the proponents' own sensibilities.

So, no, I'm still definitely not in favor of this.  Quite the opposite.  But if that's what the group really wants, I will not stand in the way.   


-----

Well that took forever.  I guess `reduceObj` will have to wait another day...


  [dm]: https://en.wikipedia.org/wiki/De_Morgan%27s_laws

paldepind:
>  I find the notion of 'falsey' in Javascript to be one of the very good parts of the language. It's not perfect; in fact, it's annoyingly wrong, as falsey should also include empty objects and empty arrays.

Basically you want empty things to be false. The notion of empty things makes a lot of sense – it's captured by monoids. The notion of true and false also makes sense. But these are _separate_ ideas, separate things and they should be kept separate. Mixing them together is not doing any good.

> They do not increase the simplicity of the actual API:

Yes they do. By not relying on the notion of falsey and truthy.

CrossEye:
>>  I find the notion of , 'falsey' in Javascript to be one of the very good parts of the language. It's not perfect; in fact, it's annoyingly wrong, as falsey should also include empty objects and empty arrays.

> Basically you want empty things to be false. The notion of empty things makes a lot of sense – it's captured by monoids.

Well, no.  I don't want `1`, the empty value for the mulitplicative monoid over numbers to be falsey, nor do I want `Nothing`, the empty value for the most common monoid over Maybe(a) to be falsey.  Nor does it seem to me that `NaN`, which I do want to be falsey, has much to do with empty.  (I also don't know of a monoid for which `{}` is the empty value, although I can imagine such is certainly possible.)  I don't think I'm combining categories here.

>> fails to support even the very fundamental rules of logic, such as De Morgan's Laws: 

> They satisfy all the rules of logic as long as you pass booleans to them.

So does the current implementation.  What we have now seems to do everything you want it to do if you use it the way you say you want to use it.  That's part of why I have a hard time seeing the drive to change it.

paldepind:
> If everyone wants to switch to a boolean-only version of and and or, I won't try to stand in the way.

Ok. Can this be merged?

davidchambers:
Since [`S.and`][1] and [`S.or`][2] behave just as I would like (which is not surprising!), I don't use `R.and` or `R.or`. As a Ramda user, the behaviour of these functions doesn't bother me. As a Ramda contributor, though, I'd love to see all references to "truthy" and "falsy" removed. Using these concepts leads to code which is less precise than it should be (e.g. `amount || 100` rather than `amount == null ? 100 : amount`), which in turn leads to bugs (e.g. `100` rather than `0` when `amount` is `0`).


[1]: https://github.com/plaid/sanctuary#and
[2]: https://github.com/plaid/sanctuary#or

PR #1337, opened 2015-08-13
update description of R.over

davidchambers:
The current description was copied from __set.js__ but not altered, meaning that `R.over` and `R.set` have the same description.

/cc @soroushjp

### DIFF
--- a/src/over.js
+++ b/src/over.js
@@ -3,7 +3,8 @@ var _curry3 = require('./internal/_curry3');
 
 /**
  * Returns the result of "setting" the portion of the given data structure
- * focused by the given lens to the given value.
+ * focused by the given lens to the result of applying the given function to
+ * the focused value.
  *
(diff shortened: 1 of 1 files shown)


src/over.js
```diff
@@ -3,7 +3,8 @@ var _curry3 = require('./internal/_curry3');
 
 /**
  * Returns the result of "setting" the portion of the given data structure
- * focused by the given lens to the given value.
+ * focused by the given lens to the result of applying the given function to
+ * the focused value.
```
buzzdecafe: is there a way to include more prepositions in that sentence? :stuck_out_tongue:

davidchambers: It's wordy, I agree!

CrossEye: Well, this is the definition of '`over`'! 

:smile:

PR #1339, opened 2015-08-15
Add K-combinator

TheLudd:
This adds the K-combinator which is similar to the existing `always` but
has a few different properties:

 * If provided with two arguments it returns the constant value
   immediately
 * It can be flipped and will then return the second value
 * It is a known concept from computer sience and maths

The somewhat hacky implementation was made to pass all standard tests.

### DIFF
--- a/src/K.js
+++ b/src/K.js
@@ -1,0 +1,22 @@
+var _curry2 = require('./internal/_curry2');
+
+/**
+ * K-combinator. Takes two arguments and returns the first.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig a -> b -> a
+ *
+ * @param {*} a The value to return
+ * @param {*} b A value to discard
+ * @return {*} The first parameter
+ * @example
+ *    R.K('foo', 'bar') //=> 'foo'
+ *    R.K('foo')('bar') //=> 'foo'
+ *
+ */
+/*jshint unused:false */
+module.exports = _curry2(function(a, b) {
+  return a;
+});
--- a/test/K.js
+++ b/test/K.js
@@ -1,0 +1,29 @@
+var assert = require('assert');
+var R = require('..');
+
+describe('K', function() {
+
+  var gimmie5 = R.K(5);
+  var flippedK = R.flip(R.K);
+
+  it('takes two arguments and returns the first', function() {
+    var result = R.K('a', 'b');
+    assert.equal('a', result);
+  });
+
+  it('can take the arguments seperately', function() {
+    var result = gimmie5(10);
+    assert.equal(5, result);
+  });
+
+  it('can be flipped', function() {
+    var result = flippedK('a', 5);
+    assert.equal(5, result);
+  });
+
+  it('can be flipped and used partially', function() {
+    var result = flippedK('a')(5);
+    assert.equal(5, result);
+  });
+
+});


TheLudd:
Will add documentation if it is accepted.

davidchambers:
I use [`S.K`][1] frequently. I've yet to require the looseness afforded by [`R.always`][2].

> ```javascript
> it('can be called with no arguments second', function() {
>   var result = gimmie5();
>   assert.equal(5, result);
> });
> ```

I don't agree with this. `R.K` should be a curried binary function like Haskell's [`const`][3]. If we're to let programmer "convenience" affect the function's type, we should give the function a different name. A binary function masquerading as a unary function which, when applied to one argument, returns a function which may be applied to zero arguments, is different in significant ways from the K combinator.

What do you have in mind for `R.always`? I would be confused to see both functions in the library.


[1]: https://github.com/plaid/sanctuary#K
[2]: http://ramdajs.com/docs/#always
[3]: http://hackage.haskell.org/package/base-4.8.1.0/docs/Prelude.html#v:const

TheLudd:
> What do you have in mind for R.always?

My thoughts were that it would be removed but considering the things you pointed out it could stay, and this could be added as a strictly binary function.

paldepind:
I don't understand the need for this when we have `R.always`?

TheLudd:
I actually wanted the reverse, a binary function the returns the second argument. Which you can get if you flip this one.

paldepind:
I don't understand the need for that either ;)

But I don't see a problem with making `R.always` behave like `K`.

```js
var always = _curry2(function(a, b) { return a; });
```

It should be backwards compatible with the current `always`.

TheLudd:
@paldepind What you wrote is not compatible, this PR is but see the objections by @davidchambers

paldepind:
@TheLudd Why is it not compatible?

TheLudd:
You must give it two arguments. You cannot give it one aregument and treat the resulting function as a nullary one.

paldepind:
Oh. I see.

I still don't see much use case for this though. Why should this be included in Ramda alongside `always`?

TheLudd:
This received some buzz on gitter so I decided to beef it up. Documentation is added

@davidchambers I changed it to a binary function. The previous implementation was due to a misconception of mine. I have since reached enlightenment.

@paldepind While this is similar to `always` many people seem to agree that this is the better implementation. The use of flipping it came from a `reduce` function where I wanted to discard the accumulator. But there are surely more uses. The fact that this supports flipping simply makes it more powerful than `always` and that seems like a good thing regardless of use cases imaginable. 

Like I said before this is not fully compatible with `always` so I'll leave it's fate up to others to decide.

src/K.js
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * @param {*} a The value to return
+ * @param {*} b A value to discard
+ * @return {*} The first parameter
+ * @example
+ *    R.K('foo', 'bar') //=> 'foo'
+ *    R.K('foo')('bar') //=> 'foo'
+ *
+ */
+module.exports = _curry2(function(a) {
```
paldepind: This should be `_curry2(function K(a, b) {` IMO.

paldepind:
I don't have an opinion as long as we don't have _both_ `always` and `K`.

src/K.js
TheLudd: IMO too. If you can convince the linter as well then we all can agree :)

TheLudd:
> I don't have an opinion as long as we don't have both always and K.

I see no problem in removing `always` if this is merged.

CrossEye:
If we decide to do this, I would rather stick with the name `always`.  Although I admit to missing `R.I`, I think combinator names are generally too obscure, and a wee bit terse.

I don't know that anything has changed since #733, but perhaps people's thinking has changed.  @buzzdecafe, do you have the same objections?  The distinction is, as @davidchambers describes it, a "looseness" that allows one to use `var fn = R.always(val); /* ... */ fn();` without passing any value to `fn`.  Although I argued for it at the time, I've looked back and found that I haven't actually needed that looseness either, and could live with `K`-style behavior.  The question for me is whether there's a strong reason to change it now.

As to the original impetus, is there a great advantage to this

```js
R.flip(R.K)
```

over this

```js
(a, b) => b
```

or this

```js
a => R.identity
```

?

src/K.js
CrossEye:
```js
/*jshint unused:false */
```

jethrolarson:
If we add K we have to add S too :)

davidchambers:
It'd be odd to have both [`R.always`][1] and `R.K`. As a Sanctuary user I'd prefer to keep `R.always` as it is now. I can then use [`S.K`][2] most of the time, and fall back to `R.always` in the rare cases where I need a nullary function which always returns a specific value. I do prefer the stricter `a -> b -> a` type, though, so I'd like Ramda's "const" function to have this type. I don't have a strong opinion as to which name the function should have.


[1]: http://ramdajs.com/docs/#always
[2]: https://github.com/plaid/sanctuary#K

CrossEye:
> I do prefer the stricter `a -> b -> a` type, though, so I'd like Ramda's "const" function to have this type.

Is this just for your reasoning about the function, or do you see any practical benefits to it?  I've never seen any case, until @TheLudd  presented his `flip` requirement, where the current `always` couldn't be used fine but where a stricter `K :: a -> b -> a` would work.  Do you have any examples?

davidchambers:
> Is this just for your reasoning about the function, or do you see any practical benefits to it?

I'm not suggesting that there are practical benefits (though @TheLudd has presented one). I prefer `a -> b -> a` purely for its elegance.

davidchambers:
Having `R.always` return a function which accepts any number of arguments is arguably more "practical", and thus more in line with Ramda's philosophy. Those who appreciate the elegance and strictness of `a -> b -> a` can use [`S.K`][1] instead.


[1]: https://github.com/plaid/sanctuary#K

PR #1344, opened 2015-08-21
handle non-existent paths in R.lensIndex and R.lensProp

davidchambers:
These changes are motivated by https://github.com/ramda/ramda/pull/1328#issuecomment-133120377. I'm keen to hear from @epeli, @scott-christopher, and others.

### DIFF
--- a/src/lensIndex.js
+++ b/src/lensIndex.js
@@ -1,7 +1,5 @@
 var _curry1 = require('./internal/_curry1');
 var lens = require('./lens');
-var nth = require('./nth');
-var update = require('./update');
 
 
 /**
@@ -24,5 +22,22 @@ var update = require('./update');
  *      R.over(headLens, R.toUpper, ['a', 'b', 'c']); //=> ['A', 'b', 'c']
  */
 module.exports = _curry1(function lensIndex(n) {
-  return lens(nth(n), update(n));
+  return lens(
+    function(_xs) {
+      var xs = _xs == null ? [] : _xs;
+      return xs[n];
+    },
+    function(x, _xs) {
+      var xs = _xs == null ? [] : _xs;
+      var result = [];
+      var max = Math.max(n, xs.length - 1);
+      var idx = 0;
+      while (idx <= max) {
+        result.push(xs[idx]);
+        idx += 1;
+      }
+      result[n] = x;
+      return result;
+    }
+  );
 });
--- a/src/lensProp.js
+++ b/src/lensProp.js
@@ -1,7 +1,5 @@
 var _curry1 = require('./internal/_curry1');
-var assoc = require('./assoc');
 var lens = require('./lens');
-var prop = require('./prop');
 
 
 /**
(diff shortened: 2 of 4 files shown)


scott-christopher:
I think we should keep these bundled lenses as they are, as supporting partial objects violates one of the lens laws:
```js
var lens = R.lensProp('a');
R.equals(
  R.set(lens, R.view(lens, {}), {}),
  {}
); // false
```

davidchambers:
`R.set(lens, R.view(lens, {}), {})` gives `{a: undefined}` on master. Are you suggesting it should give `{}` instead, @scott-christopher? I'm not sure we could do this without having `R.view` return a maybe.

scott-christopher:
Both implementations will give `{a: undefined}` because they shouldn't be operating on partial objects. My comment was more so suggesting that `lensProp`/`lensIndex` should only be used with non-partial objects. Having said that, I'm not aware of the consequences if a lens does not adhere to that particular law.

I'm pretty sure Prisms and Isos are intended for this sort of thing (though they're probably a little too far down the path to introduce in Ramda).

Perhaps @tel could provide some insight?

tel:
This is definitely a prism thing. Lenses cannot conceive of "choice" within the structure yet that's exactly what you're hitting when you consider structures which could or could not be there.

Another way of saying it is that lenses operate only and exactly when there is exactly one subpart being focused on. You need prisms or traversals to handle the situation where you have 0-or-1 target or 0-to-many targets respectively.

As a tangible example of how this plays out, the `view` analogue for `Prism s a` looks like

    preview :: Prism s a -> s -> Maybe a

And for Traversal it's (essentially)

    toList :: Traversal s a -> s -> [a]

And as to what goes wrong when you violate the lens laws? They're just going to accumulate inconsistent behavior, especially as you combine them.

The right way to think of it is that `propLens('key')` is a lens like `Lens { key: a | r } a` where `r` is a row type indicating some set of unknown remaining keys. That type cannot unify with a type where values `{}` reside unless we think of `{}` as having a type like `{ val: Maybe a | r }` for some `r` and take missing values to be implicit Nothings. At this point our lens ought to unify and end up at the type `Lens { val: Maybe a | r } (Maybe a)` so the logic forces us to a Maybe type again.

epeli:
Seems to solve my issue. Don't know about lens laws but this feels practical at least.

Just wondering whether it creates the intermediate values even when the lens is used just for viewing?

davidchambers:
> Just wondering whether it creates the intermediate values even when the lens is used just for viewing?

No, it doesn't. I added some logging to be sure:

```javascript
> R.view(R.compose(a, b, c), {a: {b: {c: 42}}})
accessing...
accessing...
accessing...
42
> R.set(R.compose(a, b, c), 99, {a: {b: {c: 42}}})
accessing...
accessing...
accessing...
reconstructing...
reconstructing...
reconstructing...
{ a: { b: { c: 99 } } }
```

davidchambers:
Let's summarize the situation:

  - As @scott-christopher rightly pointed out, the behaviour of the lens functions in this branch violates the following lens law:

    _Setting the value of the focus to the result of getting the value of the focus should have no effect._

    Since `R.view(R.lensProp('x'), {})` gives `undefined`, the result of a round trip is `{x: undefined}` rather than `{}`.

  - The behaviour of the lens functions in the latest release (and on __master__) violates the same law in the same manner.

  - As @tel pointed out, preserving the law requires the use of Maybe so that `R.view` can indicate the absence of a property.

  - Ramda does not currently use Maybe anywhere.

Here are our options as I see them:

1.  Make `R.view` a Maybe-returning function, and update `R.over` and `R.set` accordingly (or leave all three functions as they are and introduce corresponding prism functions).

2.  Make `R.lensProp` and `R.lensIndex` throw if the specified property/index does not exist.

3.  Leave things as they are.

4.  Merge this pull request.

I'm no longer in favour of the last option. Now that I have a better understanding of the issues, my preference is not to support partial objects. As a result, I prefer the second option to the third, since we should be consistent in disallowing partial objects (we currently allow digging one level too far into a partial object, but no further).

PR #1345, opened 2015-08-21
Added feature pickValues

cullophid:
works like pick, but returns the values as an Array

pickValues is similar to compose(values, pick([...])
but ensures that values are returned in the correct order on all platforms.

I have used the function quite alot.
E.g Google charts expect data as a set of touples: [x, y].
pickValues makes this very easy:

```
// data [ {time: '2015-01-01', revenue: 5, ...}, ... ]
var prepareDataForChart = R.map(R.pickValues(['time', 'revenue']))

prepareDataForChart(data);
// => [ ['2015-01-10', 5], ... ]
```

### DIFF
--- a/src/pickValues.js
+++ b/src/pickValues.js
@@ -1,0 +1,31 @@
+var _curry2 = require('./internal/_curry2');
+
+
+/**
+ * Returns a list of values matching the keys specified.
+ * property is ignored.
+ *
+ * @func
+ * @memberOf R
+ * @category Object
+ * @sig [k] -> {k: v} -> [v]
+ * @param {Array} names an array of String property names get values for
+ * @param {Object} obj The object to get values from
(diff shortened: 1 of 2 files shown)


TheLudd:
Is this not [pluck](http://ramdajs.com/0.17/docs/#pluck)?

cullophid:
pluck return an array of 1 specific property from a list of objects.
pickValues returns a array of specified values from a single object

davidchambers:
This is quite similar to [`R.props`][1].


[1]: http://ramdajs.com/docs/#props

cullophid:
This is totally a duplicate of R.props, I Missed that :(

kedashoe:
I had a similar experience recently and fumbled around for a bit before eventually finding `props`. What do you think about some `@see` annotations, @davidchambers? on `pluck` at least and maybe some of the other similar functions?

PR #1347, opened 2015-08-22
Rename isSet to allUniq

### LINKED ISSUE #1278: Rename isSet

asaf-romano:
Closes #1278

### DIFF
--- a/src/allUniq.js
+++ b/src/allUniq.js
@@ -1,0 +1,31 @@
+var _curry1 = require('./internal/_curry1');
+var _indexOf = require('./internal/_indexOf');
+
+
+/**
+ * Returns `true` if all elements are unique, in `R.equals` terms,
+ * otherwise `false`.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig [a] -> Boolean
+ * @param {Array} list The array to consider.
+ * @return {Boolean} `true` if all elements are unique, else `false`.
+ * @example
+ *
+ *      R.allUniq(['1', 1]); //=> true
+ *      R.allUniq([1, 1]);   //=> false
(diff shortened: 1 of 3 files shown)


src/allUniq.js
```diff
@@ -0,0 +1,31 @@
...
+ * @example
+ *
+ *      R.allUniq(['1', 1]); //=> true
+ *      R.allUniq([1, 1]);   //=> false
+ *      R.allUniq([[42], [42]]); //=> false
+ */
+module.exports = _curry1(function allUniq(list) {
+  var len = list.length;
+  var idx = 0;
+  while (idx < len) {
```
davidchambers: We could write `idx < list.length` here, since `list.length` cannot change.

davidchambers:
LGTM

CrossEye:
Absolutely.

src/allUniq.js
davidchambers: Thoughts on this suggestion, @CrossEye?

CrossEye: All I remember is that when we went down that path before, @megawac returned to show us the error of our ways.  I don't remember details now, and I'm in the middle of something tricky and only coming up briefly for air, so don't want to go searching for it.  But we tried to stop caching array lengths, and it made sense to everyone until Graeme showed us that we missed something important.  I wish I remembered exactly what it was.  If it doesn't affect this case, I don't mind switching.  But I also don't find it very important, either way.

paldepind: It doesn't affect this case since we're not invoking any user supplied function in the loop. The problems happens when you invoke a function that might modify the array you're looping over.

I agree with @davidchambers that we should not cache the length here.

PR #1349, opened 2015-08-25
replace R.containsWith examples

davidchambers:
The current examples were fine at the time they were written, but now that we have [value-based equality][1] they could be more simply expressed in terms of [`R.contains`][2].

[1]: https://github.com/ramda/ramda/pull/1096
[2]: http://ramdajs.com/docs/#contains

### DIFF
--- a/src/containsWith.js
+++ b/src/containsWith.js
@@ -16,8 +16,10 @@ var _curry3 = require('./internal/_curry3');
  * @return {Boolean} `true` if `x` is in `list`, else `false`.
  * @example
  *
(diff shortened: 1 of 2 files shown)


CrossEye:
LGTM.

As to the discussion in the [Gitter Room][gr], there is an interesting case to be made for replacing this with a `containsBy` implementation.


  [gr]: https://gitter.im/ramda/ramda

davidchambers:
> [T]here is an interesting case to be made for replacing this with a `containsBy` implementation.

Absolutely. I first wanted to update the example to better frame the discussion. :)

PR #1350, opened 2015-08-25
deprecate R.functions and R.functionsIn

davidchambers:
These functions are oddly specific: _given an object, give me the names of each of its properties whose value is a function_. Why do we provide this for Function but not for RegExp or Date?

This could instead live in the [cookbook][1]:

```javascript
const functions = R.compose(R.keys, R.pickBy(R.is(Function)));

functions({abs: Math.abs, negate: R.negate, x: 0, y: 0}); //=> ['abs', 'negate']
```

### DIFF
--- a/src/functions.js
+++ b/src/functions.js
@@ -12,6 +12,7 @@ var keys = require('./keys');
  * @sig {*} -> [String]
  * @param {Object} obj The objects with functions in it
  * @return {Array} A list of the object's own properties that map to functions.
+ * @deprecated since v0.18.0
  * @example
  *
  *      R.functions(R); // returns list of ramda's own function names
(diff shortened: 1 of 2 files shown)


CrossEye:
I've never used them.

@prathamesh7pute, these were [your suggestion][ps].  Any objection to pulling them out?


  [ps]: https://github.com/ramda/ramda/pull/290

prathamesh7pute:
Agree with the @davidchambers  this can live in cookbook. 

@CrossEye No objection in pulling them out. 

It was added to provide utility functions just like in underscore or lodash (_.methods / _.functions)  which is required in [ramdangular](https://github.com/Ramda/ramdangular) for exposing functions to Angular templates.

If these functions are pulled out we need to update [ramdangular](https://github.com/Ramda/ramdangular) with the above code. (will add a fix in [ramdangular](https://github.com/Ramda/ramdangular) depending on decision)

paldepind:
I agree that these functions are too specific.

davidchambers:
I've added the [recipe][1] to the cookbook.


[1]: https://github.com/ramda/ramda/wiki/Cookbook#get-an-objects-method-names

PR #1351, opened 2015-08-25
add R.eqBy

davidchambers:
I've written `a.x === b.x` and `f(a) === f(b)` many times. Ramda should provide a function for more succinctly expressing such relationships.

### DIFF
--- a/src/eqBy.js
+++ b/src/eqBy.js
@@ -1,0 +1,23 @@
+var _curry3 = require('./internal/_curry3');
+var equals = require('./equals');
+
+
+/**
+ * Takes a function and two values in its domain and returns `true` if
+ * the values map to the same value in the codomain; `false` otherwise.
+ *
+ * @func
+ * @memberOf R
+ * @category Relation
+ * @sig (a -> b) -> a -> a -> Boolean
+ * @param {Function} f
+ * @param {*} x
+ * @param {*} y
+ * @return {Boolean}
+ * @example
+ *
+ *      R.eqBy(Math.abs, 5, -5); //=> true
+ */
+module.exports = _curry3(function eqBy(f, x, y) {
+  return equals(f(x), f(y));
+});
--- a/test/eqBy.js
+++ b/test/eqBy.js
@@ -1,0 +1,28 @@
+var assert = require('assert');
+
+var R = require('..');
+
+
+describe('eqBy', function() {
+
+  it('determines whether two values map to the same value in the codomain', function() {
+    assert.strictEqual(R.eqBy(Math.abs, 5, 5), true);
+    assert.strictEqual(R.eqBy(Math.abs, 5, -5), true);
+    assert.strictEqual(R.eqBy(Math.abs, -5, 5), true);
+    assert.strictEqual(R.eqBy(Math.abs, -5, -5), true);
+    assert.strictEqual(R.eqBy(Math.abs, 42, 99), false);
+  });
+
+  it('has R.equals semantics', function() {
+    function Just(x) { this.value = x; }
+    Just.prototype.equals = function(x) {
+      return x instanceof Just && R.equals(x.value, this.value);
+    };
+
+    assert.strictEqual(R.eqBy(R.identity, 0, -0), false);
+    assert.strictEqual(R.eqBy(R.identity, -0, 0), false);
+    assert.strictEqual(R.eqBy(R.identity, NaN, NaN), true);
+    assert.strictEqual(R.eqBy(R.identity, new Just([42]), new Just([42])), true);
+  });
+
+});


src/eqProps.js
```diff
@@ -14,7 +14,7 @@ var equals = require('./equals');
  * @param {Object} obj1
  * @param {Object} obj2
  * @return {Boolean}
- *
+ * @deprecated since v0.18.0
```
davidchambers: I suggest deprecating [`R.eqProps`][1] since `R.eqProps('x')` is sugar for `R.eqBy(R.prop('x'))`. I'd prefer to promote the use of the more general function in this case.


[1]: http://ramdajs.com/docs/#eqProps

CrossEye: I have no issue at all with including syntactic sugar.

But I agree here.  This is not a common enough case to warrant dedicated sugar.

CrossEye:
:herb:

src/eqProps.js
davidchambers: I'm pleased we agree! The problem with providing lots of sugar is that users have less exposure to the underlying general-purpose functions, so knowledge is less transferable. Sugar is best when those reading the code understand the "macro expansion" that will take place.

TheLudd: > The problem with providing lots of sugar is that users have less exposure to the underlying general-purpose functions, so knowledge is less transferable. Sugar is best when those reading the code understand the "macro expansion" that will take place.

Here I come and disagree again. I think it is up to each and everyone to gain an understanding of how the code works. 

The problem with *not* adding sugar to common functions is that I now have to write `R.eqBy(R.propEq('something'))` in every place where I used to have `R.propEq`

I could of course define `propEq` myself but this requires me to put it in some other module or a module wrapping ramda that I then have to import in all my projects.

What bothers me the most is that the decisions of what sugar should be included and what should not feels very arbitrary.

TheLudd:
Yes to `eqBy`, no to deprecation of `propEq`

davidchambers:
> Yes to `eqBy`, no to deprecation of `propEq`

I'm proposing deprecation of [`R.eqProps`][1], not [`R.propEq`][2]. The fact that this patch removes one of these similarly named functions is another reason to like it. ;)


[1]: http://ramdajs.com/docs/#eqProps
[2]: http://ramdajs.com/docs/#propEq

TheLudd:
@davidchambers I simply wrote the wrong function, updated

src/eqProps.js
davidchambers: > The problem with *not* adding sugar to common functions is that I now have to write `R.eqBy(R.propEq('something'))` in every place where I used to have `R.propEq`

Let's be fair with our comparison:

```javascript
// before
R.eqProps('something')

// after
R.eqBy(R.prop('something'))
```

The latter doesn't seem too bad to me. In fact, I find it clearer.

TheLudd: Sure, it is not many extra characters that are required but again 

> What bothers me the most is that the decisions of what sugar should be included and what should not feels very arbitrary.

to me `R.always(true)` is clearer than `R.T` and [propIs](https://github.com/ramda/ramda/blob/master/src/propIs.js#L27) also feels quite unneccesary.

davidchambers: I agree that `R.always(true)` is clearer than `R.T`. I wouldn't object to removing `R.T` and `R.F`.

`R.propEq` and `R.propIs` are both unnecessary. Both functions are useful, though. It depends on the domain whether one is more likely to ask _are you 42?_ (a value-level enquiry) or _are you a Number?_ (a type-level enquiry).

TheLudd: I find `propEq` very useful. I also find `eqProps` useful. I never use `propIs` but that does not give me an urge to remove it from the code base. If others use it, that is fine by me. Same goes for `R.T` and `R.F`.

paldepind:
I still think `R.equals` should be called `R.eq`. But given that it is called `R.equals` this function should be called `R.equalsBy`.

davidchambers:
> I still think `R.equals` should be called `R.eq`.

I agree.

> But given that it is called `R.equals` this function should be called `R.equalsBy`.

I did consider it. `R.equalBy` sounds good, but `R.equalsBy` is awkward. `R.eqBy` is arguably less inconsistent than `R.equalsBy` would be, since `R.eqProps`, `R.propEq`, and `R.pathEq` all use the abbreviation.

CrossEye:
Part of the reason I like it, is that I know for certain that if I'd considered the more general `eqBy` originally, I would have absolutely preferred it to `eqProps`.  This feels like a matter of fixing an initial oversight.

But @TheLudd is right that choosing what sugar is worth keeping is more than almost anything in the library a matter of individual taste.  Anyone want to take a swing at defining useful objective criteria?  We need to keep in mind that one person's central function is someone else's sugar.  We obviously do not want a library that exposes only the `S`, `K`, and `I` combinators.  Nor do we want one that includes `filterOddsAndSquare`.  We have to find a reasonable middle ground.  Can we do that objectively, or does it simply come down to matters of taste?

davidchambers:
I've reverted the deprecation of `R.eqProps`. I'll merge this pull request once the tests pass since `R.eqBy` itself is not contentious.

PR #1352, opened 2015-08-25
run test suite in browsers via Testem

### LINKED ISSUE #1348: Test cases fail in `testem`
It seems that Ramda use Test'em to test the compatibility of browsers, see #244. So I run Test'em in Ramda, but most of the test cases are failed. Is that an issue?

davidchambers:
Closes #1348

This pull request also removes several defunct files and generally gives the various test directories a spring clean.

I removed __test/shared/Maybe.js__ because internal dependencies among the test files add complexity to what is already a complex combination of assertion libraries, testing frameworks, and shims. (I actually *prefer* to define algebraic data types in the test suite wherever necessary, but this was not my reason for removing the shared Maybe definition.)

I updated __.travis.yml__ as suggested by @scott-christopher, so Travis should run the test suite in (various?) browsers for future pull requests.


lib/test/index.html
```diff
@@ -1,21 +0,0 @@
...
-  <script src="mocha.js"></script>
-  <script src="assert-shim.js"></script>
-  <!--[if lt IE 9]>
-    <script src="../lib/sauce/ie8-shim.js"></script>
-  <![endif]-->
-  <script>mocha.setup('bdd')</script>
-  <script src="bundle.js"></script>
-  <script src="../sauce/mocharun.js"></script>
-</body>
-</html>
```
CrossEye: Is this removed because it's no longer supportable or simply because you don't see much value in it?

I find it sometimes quite useful.

test/chain.js
```diff
@@ -32,7 +30,12 @@ describe('chain', function() {
...
 
   it('dispatches to transformer objects', function() {
-    assert.strictEqual(_isTransformer(R.chain(add1, listXf)), true);
+    var listXf = {
+      '@@transducer/init': function() { return []; },
+      '@@transducer/step': function(acc, x) { return acc.concat([x]); },
+      '@@transducer/result': function(x) { return x; }
+    };
+    assert.strictEqual(typeof R.chain(add1, listXf)['@@transducer/step'], 'function');
   });
```
CrossEye: :+1:

lib/test/index.html
davidchambers: The unused files in __lib/test__ threw me off as I tried to make sense of all the various test scripts. I'd prefer to have just one way to run browser tests: `npm run testem`. In my case this runs the test suite in Chrome, Firefox, Opera, and Safari in less than a minute. It's neat!

test/commute.js
```diff
@@ -17,7 +16,26 @@ describe('commute', function() {
   });
 
   it('works on Algebraic Data Types such as "Maybe"', function() {
```
CrossEye: Note that you didn't change the name, `Maybe` -> `Identity`.

But are there compelling reasons to not allow a test to require an implementation of `Maybe` -- I don't care if its that one, the live one from ramda-fantasy, or from Folktale?

I know that you prefer unit tests to have no external references, but this seems to be pushing it too far. Should we also implement our own mini Promise library just to avoid an include in a test file?

I also personally prefer the semantics of `Maybe`, as people actually understand that one for the most part, and recognize that it, in fact, does something useful.  Using `Identity` feels as though it doesn't really demonstrate anything.

lib/test/index.html
CrossEye: I'm going to bed, but I'll have to try that out.  I'd like to know that I can easily debug in it.  That's what I liked about `index.html`.

It bothers me a bit, for the same reason that dropping `ramda.js` from the download did.  I would really like the pull from git (or even the download/unzip) to be functioning, and demonstrable.  I'd rather not assume the user has any tools, even Node.

CrossEye:
A few inline objections.  But overall, :100:

Very nice!

test/commute.js
davidchambers: > Note that you didn't change the name, `Maybe` -> `Identity`.

I considered doing so, but it didn't seem appropriate. I was reading _such as "Maybe"_ as a hint to readers as to what an algebraic data type is. Given this interpretation, _such as "Identity"_ would be less useful as I imagine there are fewer people familiar with Identity than with Maybe.

> But are there compelling reasons to not allow a test to require an implementation of `Maybe` […]?

It makes it clear which methods the ADT must provide. `R.commute`, for example, requires `ap` and `map`.

I'm not *strongly* opposed to adding a Maybe implementation as a dependency.

> I also personally prefer the semantics of `Maybe`, as people actually understand that one for the most part, and recognize that it, in fact, does something useful. Using `Identity` feels as though it doesn't really demonstrate anything.

The example previously on line 20 just involved Just. I thought since we were only showing the behaviour of Just we might as well give the container a more straightforward name. What we should do instead, though, is test the behaviour of Nothing as well as Just. I'll update this.

src/internal/_toString.js
```diff
@@ -18,6 +18,14 @@ module.exports = function _toString(x, seen) {
...
   };
 
+  //  Object.prototype.toString.call(x) does not behave correctly in
+  //  versions of PhantomJS prior to 2.0 when x is null or undefined.
+  if (x === null) {
+    return 'null';
+  } else if (x === void 0) {
+    return 'undefined';
+  }
+
```
davidchambers: This is a workaround for ariya/phantomjs#11722. We should be able to revert this change once Travis gives us the option of running PhantomJS 2.0 rather than 1.9.

test/commute.js
CrossEye: > What we should do instead, though, is test the behaviour of Nothing as well as Just. I'll update this.

:+1:

TheLudd: Or should we perhaps move this to ramda-fantasy? Is there a good use case for `commute` with arrays?

davidchambers: > Is there a good use case for `commute` with arrays?

Ramda is designed to work with other FL-compatible libraries (though it may be used alone). If this function were moved to ramda-fantasy, a Sanctuary user might need to add ramda-fantasy as a dependency just for this function. As a result we might end up defining `commute` in Sanctuary as well. Having `commute` in Ramda saves us from implementing it in N other libraries. :)

TheLudd: Sure. Then I would suggest removing the array examples. They are just confusing to me and imply that this is the main usage.

davidchambers: Examples in the docstring, or examples in the test suite?

TheLudd: [These](http://ramdajs.com/0.17/docs/#commute) examples in the docs. Replace them with `Future` instead so we have `Maybe` and `Future`. 

Just a suggestion...

davidchambers:
Could I get a `:herb:` from you, @CrossEye?

CrossEye:
Ping @davidchambers: Several tests are failing in the Travis build, around `equals`.  Is there some rebase needed to match another commit?  Or is this really environmental?

davidchambers:
Superseded by #1517

PR #1353, opened 2015-08-25
add R.containsBy

davidchambers:
See #1349

[`R.containsWith`][1] is usually used in conjunction with a function such as `(a, b) => a.x === b.x` or `(a, b) => f(a) === f(b)` ([occurrences on GitHub][2]). `R.containsBy` would be more convenient in such cases.

In the rare cases where the symmetry of `R.containsBy` is undesirable, one can always use [`R.any`][3]. It's thus unnecessary to retain `R.containsWith`.

[1]: http://ramdajs.com/docs/#containsWith
[2]: https://github.com/search?p=8&q=ramda+containsWith+language%3Ajavascript&type=Code
[3]: http://ramdajs.com/docs/#any

### DIFF
--- a/test/containsBy.js
+++ b/test/containsBy.js
@@ -1,0 +1,15 @@
+var assert = require('assert');
+
+var R = require('..');
+
+
+describe('containsBy', function() {
+
+  it('determines whether a projected list contains a projected value', function() {
+    assert.strictEqual(R.containsBy(Math.abs, 5, [1, 2, 3]), false);
+    assert.strictEqual(R.containsBy(Math.abs, 5, [4, 5, 6]), true);
+    assert.strictEqual(R.containsBy(Math.abs, 5, [-1, -2, -3]), false);
+    assert.strictEqual(R.containsBy(Math.abs, 5, [-4, -5, -6]), true);
+  });
+
+});
--- a/src/containsBy.js
+++ b/src/containsBy.js
@@ -1,0 +1,30 @@
+var _curry3 = require('./internal/_curry3');
+var any = require('./any');
+var compose = require('./compose');
+var equals = require('./equals');
+
+
+/**
+ * Returns `true` if `f(x)` is equal, in `R.equals` terms, to one or more
+ * of the elements of `R.map(f, xs)`; `false` otherwise.
+ *
+ * Short-circuits as soon as a match is found.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig (a -> b) -> a -> [a] -> Boolean
+ * @param {Function} f
+ * @param {*} x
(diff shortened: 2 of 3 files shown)


test/containsBy.js
```diff
@@ -0,0 +1,15 @@
+var assert = require('assert');
+
+var R = require('..');
+
+
+describe('containsBy', function() {
+
+  it('determines whether a projected list contains a projected value', function() {
```
davidchambers: Is "projected" the right word? I'm trying to suggest the mapping of a domain to a codomain.

TheLudd:
I have always found `containsWith` very confusing. This is perhaps somewhat better but what I would have expected is something like this `(a -> Boolean) -> [a] -> Boolean`. I.e. the first argument is a predicate function. If it returns true for any element in the list then this function returns true, otherwise false.

davidchambers:
> This is perhaps somewhat better but what I would have expected is something like this `(a -> Boolean) -> [a] -> Boolean`. I.e. the first argument is a predicate function. If it returns true for any element in the list then this function returns true, otherwise false.

You're precisely defined [`R.any`][1]. ;)

I'm happy to deprecate `R.containsWith` without introducing `R.containsBy`. I'm eager to remove a confusing function, and I'm willing to add `R.containsBy` if necessary to achieve this.


[1]: http://ramdajs.com/docs/#any

TheLudd:
> You're precisely defined R.any. ;)

Oh, didn't realize that. Must be because I want to use `contains` with a twist and I have naturally been searching for a similarly named function. Thanks

test/containsBy.js
CrossEye: I think "projected" is perfect.

CrossEye:
:herb:

davidchambers:
It seems we're comfortable deprecating `R.containsWith`. The question is whether `R.containsBy` is actually a useful addition. Can anyone think of a situation in which it's more convenient than `R.any`?

TheLudd:
If we only had `R.any` then all usage of `containsBy` would need to be changed from:
```
R.containsBy(someFn, someValue)
```
to 
```
R.any(R.compose(R.eq(someValue), someFn))
```
Idk. I think these two cases look very weird and would certainly not miss it. I think this (and other functions) should only be added from a real need users of ramda feel that they have.

davidchambers:
> I think this […] should only be added from a real need users of ramda feel that they have.

I agree. I suggest we open a separate pull request simply to deprecate `R.containsWith`. We can then close this pull request. If in the future a need for `R.containsBy` arises, we'll know where to look for an implementation. :)

davidchambers:
We'll revisit this if there is demand for `R.containsBy`.

PR #1357, opened 2015-08-26
Change `mapAccum` signature to use arrays

paldepind:
The current signature of `mapAccum` and `mapAccumRight` are a bit confusing IMO. They use a form of tuple syntax that does not really have a meaning in JavaScript. This PR changes the tuple signature to use arrays which I think better indicates that the function actually returns an array.

There might be a bit of ambiguity between `[a]` which is an array of `a`s and `[a, b]` which is an array where the first element is an `a` and the second a `b`.

### DIFF
--- a/src/mapAccum.js
+++ b/src/mapAccum.js
@@ -12,7 +12,7 @@ var _curry3 = require('./internal/_curry3');
  * @func
  * @memberOf R
  * @category List
- * @sig (acc -> x -> (acc, y)) -> acc -> [x] -> (acc, [y])
+ * @sig (acc -> x -> [acc, y]) -> acc -> [x] -> [acc, [y]]
  * @param {Function} fn The function to be called on every element of the input `list`.
  * @param {*} acc The accumulator value.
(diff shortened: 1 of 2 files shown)


CrossEye:
No objection, although I don't see one as significantly better than the other.  The trouble is that we're making this up as we go along.  It really would be nice to find some real consistencies.

davidchambers:
I prefer `(a, b)` as the representation of a pair, but my preference is not strong.

paldepind:
This change is consistent with `R.fromPairs`, `R.toPairsIn` and `R.toPairs`. It is inconsistent with `R.pairs`.

I think `(a, b)` makes senses in languages where that syntax actually works. But in JavaScript we use arrays to represent pairs so I think the syntax should use arrays as well. I.e. my basic argument is that I expect this to be easier to understand for most JavaScript programmers.

davidchambers:
If we are to change this we should update `R.pair` as well.

davidchambers:
Let's move the discussion to #1554.

PR #1359, opened 2015-08-27
add alternative to type signature of R.split

davidchambers:
As suggested by @donnut in donnut/typescript-ramda#26


paldepind:
:+1:

CrossEye:
:herb:

PR #1364, opened 2015-09-01
ignore enumeration order when comparing maps or sets in R.equals

davidchambers:
Thank you for bringing this to my attention, @jdalton!

### DIFF
--- a/src/internal/_equals.js
+++ b/src/internal/_equals.js
@@ -2,6 +2,8 @@ var _arrayFromIterator = require('./_arrayFromIterator');
 var _has = require('./_has');
 var identical = require('../identical');
 var keys = require('../keys');
+var sortBy = require('../sortBy');
+var toString = require('../toString');
 var type = require('../type');
 
 
@@ -45,7 +47,10 @@ module.exports = function _equals(a, b, stackA, stackB) {
       break;
     case 'Map':
     case 'Set':
-      if (!_equals(_arrayFromIterator(a.entries()), _arrayFromIterator(b.entries()), stackA, stackB)) {
+      if (!_equals(sortBy(toString, _arrayFromIterator(a.entries())),
+                   sortBy(toString, _arrayFromIterator(b.entries())),
+                   stackA,
+                   stackB)) {
         return false;
       }
       break;
--- a/src/internal/_toString.js
+++ b/src/internal/_toString.js
@@ -1,4 +1,3 @@
-var _contains = require('./_contains');
 var _map = require('./_map');
 var _quote = require('./_quote');
 var _toISOString = require('./_toISOString');
@@ -10,7 +9,14 @@ var test = require('../test');
 module.exports = function _toString(x, seen) {
   var recur = function recur(y) {
     var xs = seen.concat([x]);
-    return _contains(y, xs) ? '<Circular>' : _toString(y, xs);
+    var idx = 0;
+    while (idx < xs.length) {
+      if (xs[idx] === y) {
+        return '<Circular>';
+      }
+      idx += 1;
+    }
+    return _toString(y, xs);
   };
 
   //  mapPairs :: (Object, [String]) -> [String]
--- a/test/equals.js
+++ b/test/equals.js
@@ -207,6 +207,8 @@ describe('equals', function() {
       assert.strictEqual(R.equals(new Map([[1, 'a']]), new Map([])), false);
       assert.strictEqual(R.equals(new Map([[1, 'a']]), new Map([[1, 'a']])), true);
       assert.strictEqual(R.equals(new Map([[1, 'a']]), new Map([[1, 'b']])), false);
+      assert.strictEqual(R.equals(new Map([[1, 'a'], [2, 'b']]), new Map([[2, 'b'], [1, 'a']])), true);
+      assert.strictEqual(R.equals(new Map([[1, 'a'], ['1', 'a']]), new Map([['1', 'a'], [1, 'a']])), true);
       assert.strictEqual(R.equals(new Map([[1, 'a'], [2, new Map([[3, 'c']])]]), new Map([[1, 'a'], [2, new Map([[3, 'c']])]])), true);
       assert.strictEqual(R.equals(new Map([[1, 'a'], [2, new Map([[3, 'c']])]]), new Map([[1, 'a'], [2, new Map([[3, 'd']])]])), false);
       assert.strictEqual(R.equals(new Map([[[1, 2, 3], [4, 5, 6]]]), new Map([[[1, 2, 3], [4, 5, 6]]])), true);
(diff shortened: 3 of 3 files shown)


CrossEye:
This sounds [awfully familiar][ss].  :+1:


  [ss]: https://github.com/ramda/ramda/issues/1067#issuecomment-135923353

CrossEye:
Thinking about it again, I'm really not thrilled with a `toString` implementation of this.  I wonder, given how heavyweight `toString` has to be, whether an implementation based on bi-directional tests of `otherSet.has(nextEltOfThisSet)` would be horribly slower.  Or, I suppose for value equality, perhaps using `containsWith(equals)` on the iterable output of the Set.

davidchambers:
Would you like to take a crack at an alternative implementation, @CrossEye?

paldepind:
@davidchambers 

I've sketched one out in another thread:

> The algorithm is simple and will perform much better than what this PR does.

> 1. Compare size of set. If not equal, the sets are not equal.
2. Iterate over each element in first set while checking for presence in other set with `Set.prototype.has`. If an item is missing the sets are not equal. Otherwise they are.

CrossEye:
Yes, if nobody beats me to it, I'll get to this in a few days, either as Simon says with `Set.prototype.has` or using `containsWith(equals)`.

Even though Simon had pointed out our earlier, I missed the fact that this doesn't have to be symmetric.   A length comparison captures half the test.

paldepind:
> Yes, if nobody beats me to it, I'll get to this in a few days

:+1: 

> or using `containsWith(equals)`.

Unless I'm mistaking something that would be an n^2 algorithm at it would still not satisfy `R.equals(set1, set2) === true -> set1.has(x) === set2.has(x)`.

My proposal should be n*log(n) and would satisfy the above property.

davidchambers:
The question is whether to use value-based equality, as we do for other data types, or the identity-based equality of the set itself.

Choosing identity-based equality would lead to an interesting contradiction:

```javascript
> R.equals(new Set([]), new Set([]))
true
> R.equals(new Set([new Set([])]), new Set([new Set([])]))
false
```

We'd be comparing the *container* by value, but its *contents* by identity. Two empty containers would be equal, but two sets each containing one (non-identical) empty container would not be equal.

Value-based equality is more internally consistent. It is, though, *inconsistent* with built-in methods such as [`Set.prototype.has`][1]. This isn't necessarily a problem. After all, value-based equality is also inconsistent with [`Array.prototype.indexOf`][2]. We address this issue by providing [`R.indexOf`][3] and [`R.contains`][4]. We could update the latter to operate on sets as well.

The one remaining problem is that a value such as `new Set([{}, {}])` makes no sense in the presence of value-based equality. This suggests we should provide functions for creating sets, and for deriving sets from other sets.


[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/has
[2]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf
[3]: http://ramdajs.com/docs/#indexOf
[4]: http://ramdajs.com/docs/#contains

CrossEye:
@davidchambers:

You beat me to it by minutes.  I was going to note that 

```js
var s = new Set([{a: 1}, {a: 1}]);
arrayFromIterator(s.values()).length; //=> 2
```

And this would make it difficult to do value-based equality directly.  If we're going to interact with native sets, we might have nothing better than dealing with identity equality, using `has`.

CrossEye:
@paldepind:

> Unless I'm mistaking something that would be an n^2 algorithm at it would still not satisfy `R.equals(set1, set2) === true -> set1.has(x) === set2.has(x)`

I was thinking that's what we would want, that `has` was too strong.  But I believe I'm coming around to your view on this.  I'm struggling a bit though, and would like to hear more of David's ideas of providing "functions for creating sets, and for deriving sets from other sets."

davidchambers:
Here's a rough outline of what I have in mind:

```javascript
> R.createSet([1, 2, 3, 1, 2, 3])
new Set([1, 2, 3])
> R.createSet([{}, {}])
new Set([{}])
> R.union(R.createSet([1, {}]), R.createSet([2, {}]))
new Set([1, 2, {}])
```

This is a problem, though:

```javascript
> new Set([0, -0]).size
1
```

It looks like I need to create my own Set type!

paldepind:
@davidchambers 

> We address this issue by providing R.indexOf and R.contains. We could update the latter to operate on sets as well.

[I discuss that here](https://github.com/ramda/ramda/pull/1235#issuecomment-136806231). Note this in particular:

> And we have no way to redefine `Set.prototype.has` because then we lose the performance benefit and the entire reason for using Sets in the first place.

> It looks like I need to create my own Set type!

Exactly. But then you wont get the performance benefits that native sets has.

CrossEye:
@davidchambers:

> It looks like I need to create my own Set type!

Yes, that's what worries me about any attempts at value-equality.

Think also:

```javascript
var s  = R.createSet([{}, {}]); //=> new Set([{}])
s.add({})
arrayFromIterator(s.values()).length; //=> 2
```

So we'd need to override `add` as well...

davidchambers:
Rather than `s.add({})` one could write `s = R.union(s, R.createSet([{}]))`.

CrossEye:
> Rather than `s.add({})` one could write `s = R.union(s, R.createSet([{}]))`.

Right.  But then essentially we're not using native sets at all except as backing stores.  That's not necessarily a bad thing.  It certainly fits with our ideas of immutability.  But it begs the question of how we deal with Sets not created through such calls.  It's difficult to say we simply ignore them.  But we really do want to be dealing with value equality as best we can, and these are not going to make it easy.  They may well make it impossible.  As @paldepind pointed out, at the very least they will make it less performant.

davidchambers:
Value-based equality doesn't seem viable for the built-in Set type. I look forward to seeing an implementation of the alternative approach.

CrossEye:
Well my (not-so-good-)idea was for an attempt to maintain value equality.  I think we're left with the approach Simon presented.  That should be quite simple to implement.

jdalton:
So y'all are cool with:
```
equals(new Set([{a:1}]), new Set([{a:1}])); // false
```
?

CrossEye:
@jdalton: No, and I'm still trying to figure out a good answer to that.

I'm not sure there is a good answer.  We can deal with that case in a fairly straightforward manner, but it would likely give us an odd answer to 

```js
equals(new Set([{a:1}]), new Set([{a:1}, {a:1}]));
```

The problem is to even figure out _what_ answer we would want from that second one.  As native Sets, we have a two-element Set compared to a one-element one, and hence should be unequal..  As value-based Sets, they should each only contain a single element, and hence should be equal.

This is one of the first times where I'm feeling that the language may leave us no good way to respond.

jdalton:
> This is one of the first times where I'm feeling that the language may leave us no good way to respond.

I'm currently treating maps and sets as wrappers around the values so I dig in and perform and unordered comparison on them. My `isEqual` method is a loose equivalence method and not the single source or comparison truth in the lib though.

davidchambers:
Since we're going to try a different approach, I'll close this pull request.

CrossEye:
> Yes, if nobody beats me to it, I'll get to this in a few days, 

Someone beat me to it, or at least to something closely related: #1374.

PR #1369, opened 2015-09-03
dispatch recursively in R.equals

### LINKED ISSUE #1368: R.equals does not dispatch on nested values
Example:
```js
var left = S.Left(1);
var right = S.Right(1);
console.log(right.equals(left)); //=> false
console.log(R.equals(left, right)); //=> false
console.log(R.equals([left], [right])); //=> true
```
Right now only the top level checks for an `equals` method on the objects, once you get to the internal `_equals` there is no more dispatching.

davidchambers:
Fixes #1368

It's unusual for dispatching to take place in a private function, but it's necessary in this case.

### DIFF
--- a/src/internal/_equals.js
+++ b/src/internal/_equals.js
@@ -14,10 +14,16 @@ module.exports = function _equals(a, b, stackA, stackB) {
     return false;
   }
 
+  if (a == null || b == null) {
+    return false;
+  }
+
+  if (typeof a.equals === 'function' || typeof b.equals === 'function') {
+    return typeof a.equals === 'function' && a.equals(b) &&
+           typeof b.equals === 'function' && b.equals(a);
+  }
+
   switch (type(a)) {
-    case 'Null':
-    case 'Undefined':
-      return false;
     case 'Array':
     case 'Object':
       break;
--- a/src/equals.js
+++ b/src/equals.js
@@ -1,6 +1,5 @@
 var _curry2 = require('./internal/_curry2');
 var _equals = require('./internal/_equals');
-var _hasMethod = require('./internal/_hasMethod');
 
 
 /**
@@ -26,6 +25,5 @@ var _hasMethod = require('./internal/_hasMethod');
  *      R.equals(a, b); //=> true
  */
 module.exports = _curry2(function equals(a, b) {
-  return _hasMethod('equals', a) ? a.equals(b) :
-         _hasMethod('equals', b) ? b.equals(a) : _equals(a, b, [], []);
+  return _equals(a, b, [], []);
 });
--- a/test/equals.js
+++ b/test/equals.js
@@ -242,7 +242,7 @@ describe('equals', function() {
     });
   }
 
-  it('dispatches to `equals` method', function() {
+  it('dispatches to `equals` method recursively', function() {
     function Left(x) { this.value = x; }
     Left.prototype.equals = function(x) {
       return x instanceof Left && R.equals(x.value, this.value);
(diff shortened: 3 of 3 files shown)


src/internal/_equals.js
```diff
@@ -14,10 +14,19 @@ module.exports = function _equals(a, b, stackA, stackB) {
...
+  }
+
+  if (typeof a.equals === 'function') {
+    return a.equals(b);
+  }
+
+  if (typeof b.equals === 'function') {
+    return b.equals(a);
+  }
+
```
CrossEye: Lawn-mowing must have done me in.  I swear I submitted this comment last night!  :smile:

There is probably nothing we can do about this; I'm guessing its a well-known problem with some fancy name.  But I noticed it here; this sort of implementation can lead to a non-commutativity of `equals` in the face of type inheritance:

```js
var s = new Square(5);
var r = new Rectangle(5, 5);
R.equals(s, r); //=> false (r not instance of Square)
R.equals(r, s); //=> true (s instance of Rectangle, r.height (/width) == s.height (/width) )
```

Not expecting any change.  Just sayin'.

buzzdecafe: yes, to keep commutativity you may have to:

```js
if (typeof a.equals === 'function') {
    return a.equals(b) && b.equals(a);
}
```

davidchambers: Where does responsibility lie for ensuring that both values have the same constructor? In Ramda's `_equals` function, or in the various `equals` methods to which it dispatches?

This is another example of the issue @CrossEye raised above:

```javascript
function Point(x, y) {
  this.x = x;
  this.y = y;
}
Point.prototype.equals = function(point) {
  return this instanceof Point &&
         this.x === point.x && this.y === point.y;
};

ColorPoint = function(x, y, color) {
  this.x = x;
  this.y = y;
  this.color = color;
}
ColorPoint.prototype = new Point(0, 0);
ColorPoint.prototype.equals = function(point) {
  return point instanceof ColorPoint &&
         this.x === point.x && this.y === point.y &&
         this.color === point.color;
};

var p = new Point(2, 2);
var cp = new ColorPoint(2, 2, 'red');

p.equals(cp);  // => true
cp.equals(p);  // => false
```

The question is whether in the OO world it makes sense for `p.equals(cp)` to evaluate `true`, or whether `this instanceof Point` should be replaced with `this.constructor === Point` to ensure that a ColorPoint instance is never considered equal to a Point instance.

Regardless of the answer to this question I believe we should ensure that both values have the same constructor in `_equals`. `x` ≍ `y` implies `x.constructor` ≍ `y.constructor` (for every "accessible" value of `x`). If I have a function which returns `x.constructor.name` and you tell me that `x` and `y` are equivalent, I should be able to replace `x` with `y` without any observable effect. If I cannot, `x` and `y` are clearly not equivalent.

One option is to add this after the `a == null || b == null` check:

```javascript
if (a.constructor !== b.constructor) {
  return false;
}
```

We could then be sure that `a` and `b` are of the same type, so any (non-broken) `equals` method would guarantee `a.equals(b) === b.equals(a)`.

Unfortunately, this change would prevent values created in different environments from being considered equal, which is undesirable (`[1, 2, 3]` in an iframe should be equal to `[1, 2, 3]` in the parent frame). We could instead compare `constructor.name`, but there's no guarantee that two constructors with the same name are equivalent.

I don't have a solution to the cross-frame issue, so @buzzdecafe's suggestion of checking both `a.equals(b)` and `b.equals(a)` seems the best course of action.

davidchambers:
I've made a change to guarantee commutativity.

CrossEye:
I like it, and I actually prefer this to taking on the constructor issue directly.  If my Rectangle constructor function want to claim that `(new Rectangle(4, 7)).equals({width: 4, height: 7, equals: testHeightAndWidth})` even though the constructors don't match, that's not our concern.

:herb:

PR #1374, opened 2015-09-06
equals should return true for Maps with same keys and values but different order insertion

gonzaloruizdevilla:
this two maps should be considered the same:

```

new Map([[[1, 2, 3], [4, 5, 6]],    [[7, 8, 9], [10, 11, 12]]]),
new Map([[[7, 8, 9], [10, 11, 12]], [[1, 2, 3], [4, 5, 6]]])
```

on the way of solving it, I made `map` and `find` compatible with iterables.
If this looks like the way to go, I will make `all`compatible with iterables, so there's no need to use map and simplify the implementation.

### DIFF
--- a/src/find.js
+++ b/src/find.js
@@ -1,5 +1,6 @@
 var _curry2 = require('./internal/_curry2');
 var _dispatchable = require('./internal/_dispatchable');
+var _find = require('./internal/_find');
 var _xfind = require('./internal/_xfind');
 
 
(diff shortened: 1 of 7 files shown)


CrossEye:
Thank you.  See also #1364.  

This is a lot to consider.  I like having `map` work on iterables.  It would be nice to have a consolidated approach to figure out exactly which list functions should handle iterables in a uniform manner, but perhaps this could just be considered the start of that.

gonzaloruizdevilla:
I see a few different approaches to do this:

* make two implementations for arrays and iterables as in map with `_arrayMap` and `_iteratorMap`. Maybe this can be the one with the best performance.
* always use the iterator implementation. If object received doesn't have [Symbol.iterator] and is arrayLike, we can construct the iterator to reuse the implementation.
* reimplement all this list functions using `reduce` (which already supports iterator). This one should be the one needing less LOC

tycho01:
> It would be nice to have a consolidated approach to figure out exactly which list functions should handle iterables in a uniform manner, but perhaps this could just be considered the start of that.

@CrossEye if you'd find it of any use, for the [TS typings](https://github.com/donnut/typescript-ramda/blob/master/index.d.ts) I'd tested the Ramda function to check which ones handled what -- the `List` you'll see in there are the cases where I found 'any' array-like inputs (-> array + string) to be supported (only few cases now actually required array). Not sure it'd be of significant use here though... maybe it'd help find some not listed under `List` in the Ramda docs, fwiw.

PR #1378, opened 2015-09-08
remove _asChars

davidchambers:
In #1275 we discussed the possibility of supporting strings in list functions such as `R.map`, `R.filter`, and `R.all`. It now appears that `R.reverse` may be the last of the list functions which can sensibly support strings. Since `_asChars` is currently used in just one place and is unlikely to be used elsewhere in future, we should simplify the implementation of `R.reverse`. This should please @paldepind. :)


paldepind:
I am pleased :smile: This looks splendid.

CrossEye:
:herb:

PR #1382, opened 2015-09-13
add R.words and R.unwords

davidchambers:
These are taken straight from the Prelude: [`words`][1]; [`unwords`][2].

[1]: http://hackage.haskell.org/package/base-4.8.1.0/docs/Prelude.html#v:words
[2]: http://hackage.haskell.org/package/base-4.8.1.0/docs/Prelude.html#v:unwords

### DIFF
--- a/src/unwords.js
+++ b/src/unwords.js
@@ -1,0 +1,18 @@
+var join = require('./join');
+
+
+/**
+ * Joins (with separating spaces) a list of words.
+ *
+ * @func
+ * @memberOf R
+ * @category String
+ * @sig [String] -> String
+ * @param {Array}
+ * @return {String}
+ * @see R.words
+ * @example
+ *
+ *      R.unwords(['foo', 'bar', 'baz']); //=> 'foo bar baz'
+ */
+module.exports = join(' ');
--- a/src/words.js
+++ b/src/words.js
@@ -1,0 +1,23 @@
+var compose = require('./compose');
+var isEmpty = require('./isEmpty');
+var reject = require('./reject');
+var split = require('./split');
+
+
+/**
+ * Splits a string into a list of words, which were delimited by whitespace
+ * characters.
+ *
+ * @func
+ * @memberOf R
+ * @category String
+ * @sig String -> [String]
+ * @param {String}
+ * @return {Array}
+ * @see R.unwords
+ * @example
+ *
+ *      R.words(' foo  bar  baz '); //=> ['foo', 'bar', 'baz']
+ *      R.words('foo\nbar\nbaz\n'); //=> ['foo', 'bar', 'baz']
+ */
+module.exports = compose(reject(isEmpty), split(/\s+/));
--- a/test/unwords.js
+++ b/test/unwords.js
@@ -1,0 +1,14 @@
+var assert = require('assert');
+
+var R = require('..');
+
+
+describe('unwords', function() {
+
+  it('joins (with separating spaces) a list of words', function() {
+    assert.deepEqual(R.unwords([]), '');
+    assert.deepEqual(R.unwords(['foo', 'bar', 'baz']), 'foo bar baz');
+    assert.deepEqual(R.unwords([' foo', ' bar', ' baz']), ' foo  bar  baz');
+  });
+
+});
--- a/test/words.js
+++ b/test/words.js
@@ -1,0 +1,16 @@
+var assert = require('assert');
+
+var R = require('..');
+
+
+describe('words', function() {
+
+  it('splits a string into a list of words', function() {
+    assert.deepEqual(R.words(''), []);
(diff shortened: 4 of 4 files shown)


CrossEye:
I'm looking for other support for this beyond @davidchamber's request.  `unwords` is pure sugar, and `words` is only one-step away from pure sugar, so I'd like to know if others find this particularly sweet.

I have definitely used versions of both of these functions, so I recognize their usefulness, but are they important enough to include?

raine:
Not sure if you can have `words` and `unwords` without also including `lines` and `unlines` and maybe `chars` and `unchars`. Personally I find these useful and have them in my own [ramda extension](https://github.com/raine/my-ramda-ext/blob/master/src/index.js#L3-L8).

davidchambers:
> Not sure if you can have `words` and `unwords` without also including `lines` and `unlines` and maybe `chars` and `unchars`.

I'd like to include all six functions. :)

> Personally I find these useful and have them in my own [ramda extension](https://github.com/raine/my-ramda-ext/blob/master/src/index.js#L3-L8).

Out of interest, I noticed some differences between your functions and their Haskell namesakes:

| Operation                 | Prelude           | Raine                     |
| ------------------------- | ----------------- | ------------------------- |
| `unlines ["foo", "bar"]`  | `"foo\nbar\n"`    | `"foo\nbar"`              |
| `lines "foo\nbar\n"`      | `["foo", "bar"]`  | `["foo", "bar", ""]`      |
| `words "foo\nbar\n"`      | `["foo", "bar"]`  | `["foo\nbar\n"]`          |
| `words " foo bar "`       | `["foo", "bar"]`  | `["", "foo", "bar", ""]`  |

paldepind:
I don't have an opinion on this. But I am a bit confused. I though you, @davidchambers, wanted to decrease the number of functions we have. But now you want to include a name for `R.join(' ')`. It is even less sugar than `R.times`, discussed in #1264.

davidchambers:
I'd like to see Ramda continue to grow in size, @paldepind, but I'm always eager to prune functions which are not [well defined][1]. In contrast to `R.times`, I consider these functions well defined by virtue of being part of the Prelude and having descriptive, unambiguous names.

As for the question of providing sugar for `R.join(' ')`, I would not be in favour but for the position of `unwords` as a member of a family of functions. `unlines` desugars to an expression which would be tedious to write over and over:

```javascript
// unlines :: [String] -> String
const unlines = R.compose(R.join(''), R.map(R.concat(R.__, '\n')));
```

Were we to provide `lines` and `unlines`, it would be strange to provide `words` but not `unwords`.


[1]: https://github.com/ramda/ramda/pull/1264#issuecomment-120645134

paldepind:
What isn't well defined about `R.times`? It's equal to `R.map(f, R.range(0, n))`?

> A function is well defined if there is exactly one clear definition

By that definition `unlines` is not well defined. Raine and the Haskell prelude has given two equally good implementations.

I don't think being a part of Haskell's prelude is a good reason for using a function. In my experience it's more normal to do plain text IO in Haskell than in JavaScript. I think being included in Underscore or Lodash is actually a better reason for including a function.

That being said I am fine with all of these functions being included. I do prefer Raine's `unlines` though. As far as I can tell the other behaviour is mostly relevant if you're doing a CLI program where a trailing newline is necessary. If you're, for instance, setting the value of an HTML textbox a trailing newline would be unwanted. Furthermore it's easier to add a newline than to remove one.

CrossEye:
> I don't think being a part of Haskell's prelude is a good reason for using a function.

I agree.

> I think being included in Underscore or Lodash is actually a better reason for including a function.

I disagree.

I think we should be designing our own API based on functions that the community expresses interest in having, subject of course to whatever restrictions we see fit to impose (no mutation, pure functions only, etc.)  When it comes to the details of the API of those functions, I would expect to receive more inspiration from other functional languages than from other JS libraries.  When it comes to naming, I would expect almost equal degrees of influence.

But the important point is that it's not important to include `foldObj` because Underscore does or to include `unwords` because Haskell has it.  Those should be at most minor support for some other argument about the usefulness of the functions or about the difficulties not having them supplied in Ramda actually presents to users.

davidchambers:
Perhaps these functions would be more at home in Sanctuary, which is more closely aligned with Haskell and PureScript. I'm happy to close this pull request and open a Sanctuary pull request if others agree.

CrossEye:
I don't particularly want to drop this, not if there's support for it.

I simply don't want anyone to think -- ourselves included -- that we're trying to port over all the Haskell functions.  We need to think carefully about what we want.  That's all.

davidchambers:
Superseded by plaid/sanctuary#87

We can let Sanctuary test the waters. We can revive this patch if at some point there's support for including these functions in Ramda.

PR #1388, opened 2015-09-16
make R.anyPass and R.allPass return curried functions

davidchambers:
Before:

```javascript
R.allPass([R.gte, R.lte])(5, 5);  // => true
R.allPass([R.gte, R.lte])(5)(5);  // TypeError: boolean is not a function
R.allPass([R.gte, R.lte], 5)(5);  // TypeError: boolean is not a function
R.allPass([R.gte, R.lte], 5, 5);  // => true
```

After:

```javascript
R.allPass([R.gte, R.lte])(5, 5);  // => true
R.allPass([R.gte, R.lte])(5)(5);  // => true
R.allPass([R.gte, R.lte], 5)(5);  // => unary function (extra arg ignored)
R.allPass([R.gte, R.lte], 5, 5);  // => binary function (extra args ignored)
```


CrossEye:
Glad to have the fix.  Not sure how much I like the duplication.  Those two 13-line functions are awfully similar.  But maybe that's for the best.  It certainly reduces stacktraces not to have `predicateWrap`.  Hmmm.

TheLudd:
I was under the impression that these functions were of airity 2, took one list of unary functions and one value. I find both the new and old implementations complex. Just curious if there is any demand for this flexibility.

davidchambers:
> Not sure how much I like the duplication. Those two 13-line functions are awfully similar.

The functions differ in four ways. I don't think it's a good idea to parameterize these differences. Also, they'd be ten-line functions were they to use `for` loops. ;)

> I was under the impression that these functions were of airity 2, took one list of unary functions and one value.

I'd like to discuss this possibility in a separate thread, as it also concerns `R.both`, `R.either`, and `R.ifElse` (and perhaps other functions).

Issue #1318, opened 2015-07-29
Reconsider changes to pipe and compose


CrossEye:
In #1260, after a fairly tortuous path, we ended up reworking a number of the `composeX` / `pipeX` functions.  I'm afraid at this point that I believer we broke them, at lest the default `pipe`/`compose` versions.

Today, attempting to answer [this Stack Overflow question][so], I tried to use this code as a points-free alternative:

```js
// getLocalDeps :: [{ source :: { value :: String }}] -> RegExp -> [String]
const getLocalDeps =  R.useWith(
  R.flip(R.call),
  R.map(R.path(['source', 'value'])),
  R.pipe(R.test, R.filter)
);
```

I was stuck for quite a while trying to figure out [why it didn't work][v16].  Everything looks right.  Eventually I realized that `pipe` wasn't doing what it was supposed to.  It [works in the previous release][v15].  And of course I can [simply use a lambda instead][v17]:

```js
// getLocalDeps :: [{ source :: { value :: String }}] -> RegExp -> [String]
const getLocalDeps =  R.useWith(
  R.flip(R.call),
  R.map(R.path(['source', 'value'])),
  regex => R.filter(R.test(regex))
);
```
 or this:

```js
  // ...
  R.pipe(regex => item => R.test(regex, item), R.filter)
  //...
```

But those both miss some of the elegance of the original.  I really think that we want to be able to use `pipe`/`compose` in this manner.  This strikes me as a very common requirement.

  [so]: http://stackoverflow.com/questions/31701246/
  [v15]: http://bit.ly/1MtZhjB
  [v16]: http://bit.ly/1MtZfbs
  [v17]: http://bit.ly/1MtZhQz

davidchambers:
I'd prefer to leave `R.compose` and `R.pipe` as they are now, and add `R.composeN` and `R.pipeN` to support functions with optional arguments and functions which misreport their arities. We do this for `R.curry` with `R.curryN`.

asaf-romano:
Please also see #1314. 0.16+ is so far a no-go for me because many of my async pipelines happen to start with a sync function :-/

CrossEye:
But note that this is neither of those.  I simply want to use `R.pipe(R.test, R.filter)`, pass it the regex and get back `[String] -> [String]`.  That worked before #1260.  It's not a matter of optional arguments or poorly reporting of arities.  I could also solve it for this problem with 

```js
  // ...
  R.pipe(R.unary(R.test), R.filter)
[...]

PR #1391, opened 2015-09-20
do not curry functions returned by R.{compose,pipe}{,P}

davidchambers:
This pull request partially reverts #1260. It means a pipeline may begin with a function with optional arguments, in which case the resulting function will itself have optional arguments. As a result, `R.compose`, `R.pipe`, `R.composeP`, and `R.pipeP` no longer return curried functions.

Various more significant changes have been discussed in #1318, but we've yet to reach consensus and our indecision is blocking the release of v0.18.0. My hope is that this pull request will unblock us.

As stated in https://github.com/ramda/ramda/issues/1318#issuecomment-132005376, I prefer simpler definitions of `compose` and `pipe`. I've opened plaid/sanctuary#94 to add these, along with `meld`, to Sanctuary. @CrossEye has expressed interest in `meld`; I'll open a Ramda pull request for a variadic `meld` if and when this pull request is merged.

### DIFF
--- a/src/compose.js
+++ b/src/compose.js
@@ -9,7 +9,7 @@ var reverse = require('./reverse');
  * @func
  * @memberOf R
  * @category Function
- * @sig ((y -> z), (x -> y), ..., (o -> p), ((a, b, ..., n) -> o)) -> (a -> b -> ... -> n -> z)
+ * @sig ((y -> z), (x -> y), ..., (o -> p), ((a, b, ..., n) -> o)) -> ((a, b, ..., n) -> z)
  * @param {...Function} functions
  * @return {Function}
  * @see R.pipe
--- a/src/composeK.js
+++ b/src/composeK.js
@@ -2,6 +2,7 @@ var chain = require('./chain');
 var compose = require('./compose');
 var identity = require('./identity');
 var map = require('./map');
+var prepend = require('./prepend');
 
 
 /**
@@ -37,7 +38,5 @@ var map = require('./map');
  *      //=> Nothing()
  */
 module.exports = function composeK() {
-  return arguments.length === 0 ?
-    identity :
-    compose.apply(this, map(chain, arguments));
+  return compose.apply(this, prepend(identity, map(chain, arguments)));
 };
--- a/src/pipe.js
+++ b/src/pipe.js
@@ -1,5 +1,5 @@
+var _arity = require('./internal/_arity');
 var _pipe = require('./internal/_pipe');
-var curryN = require('./curryN');
 var reduce = require('./reduce');
 var tail = require('./tail');
 
@@ -13,7 +13,7 @@ var tail = require('./tail');
  * @func
  * @memberOf R
  * @category Function
- * @sig (((a, b, ..., n) -> o), (o -> p), ..., (x -> y), (y -> z)) -> (a -> b -> ... -> n -> z)
+ * @sig (((a, b, ..., n) -> o), (o -> p), ..., (x -> y), (y -> z)) -> ((a, b, ..., n) -> z)
  * @param {...Function} functions
  * @return {Function}
  * @see R.compose
@@ -27,6 +27,6 @@ module.exports = function pipe() {
   if (arguments.length === 0) {
     throw new Error('pipe requires at least one argument');
   }
-  return curryN(arguments[0].length,
+  return _arity(arguments[0].length,
                 reduce(_pipe, arguments[0], tail(arguments)));
 };
--- a/src/pipeP.js
+++ b/src/pipeP.js
@@ -1,5 +1,5 @@
+var _arity = require('./internal/_arity');
 var _pipeP = require('./internal/_pipeP');
-var curryN = require('./curryN');
 var reduce = require('./reduce');
 var tail = require('./tail');
 
@@ -25,6 +25,6 @@ module.exports = function pipeP() {
   if (arguments.length === 0) {
     throw new Error('pipeP requires at least one argument');
   }
-  return curryN(arguments[0].length,
+  return _arity(arguments[0].length,
                 reduce(_pipeP, arguments[0], tail(arguments)));
 };
--- a/test/compose.js
+++ b/test/compose.js
@@ -11,27 +11,12 @@ describe('compose', function() {
   });
 
   it('performs right-to-left function composition', function() {
-    var f = function(a) { return [a]; };
-    var g = function(a, b) { return [a, b]; };
-    var h = function(a, b, c) { return [a, b, c]; };
+    //  f :: (String, Number?) -> ([Number] -> [Number])
+    var f = R.compose(R.map, R.multiply, parseInt);
 
-    assert.strictEqual(R.compose(f, f, f).length, 1);
-    assert.strictEqual(R.compose(f, f, g).length, 2);
-    assert.strictEqual(R.compose(f, f, h).length, 3);
-
-    assert.deepEqual(R.compose(f, f, f)(1), [[[1]]]);
-    assert.deepEqual(R.compose(f, f, g)(1, 2), [[[1, 2]]]);
-    assert.deepEqual(R.compose(f, f, g)(1)(2), [[[1, 2]]]);
-    assert.deepEqual(R.compose(f, f, h)(1, 2, 3), [[[1, 2, 3]]]);
-    assert.deepEqual(R.compose(f, f, h)(1, 2)(3), [[[1, 2, 3]]]);
-    assert.deepEqual(R.compose(f, f, h)(1)(2, 3), [[[1, 2, 3]]]);
-    assert.deepEqual(R.compose(f, f, h)(1)(2)(3), [[[1, 2, 3]]]);
-
-    var $g = R.curry(g);
-
-    assert.strictEqual(R.compose($g, $g).length, 2);
-    assert.deepEqual(R.compose($g, $g)(1, 2)(3), [[1, 2], 3]);
-    assert.deepEqual(R.compose($g, $g)(1)(2)(3), [[1, 2], 3]);
+    assert.strictEqual(f.length, 2);
+    assert.deepEqual(f('10')([1, 2, 3]), [10, 20, 30]);
+    assert.deepEqual(f('10', 2)([1, 2, 3]), [2, 4, 6]);
   });
 
   it('passes context to functions', function() {
--- a/test/composeP.js
+++ b/test/composeP.js
@@ -12,26 +12,34 @@ describe('composeP', function() {
     assert.strictEqual(R.composeP.length, 0);
   });
 
-  it('performs right-to-left composition of Promise-returning functions', function() {
+  it('performs right-to-left composition of Promise-returning functions', function(done) {
     var f = function(a) { return Q.Promise(function(res) { res([a]); }); };
     var g = function(a, b) { return Q.Promise(function(res) { res([a, b]); }); };
-    var h = function(a, b, c) { return Q.Promise(function(res) { res([a, b, c]); }); };
 
-    assert.strictEqual(R.composeP(f, f, f).length, 1);
-    assert.strictEqual(R.composeP(f, f, g).length, 2);
-    assert.strictEqual(R.composeP(f, f, h).length, 3);
+    assert.strictEqual(R.composeP(f).length, 1);
+    assert.strictEqual(R.composeP(g).length, 2);
+    assert.strictEqual(R.composeP(f, f).length, 1);
+    assert.strictEqual(R.composeP(f, g).length, 2);
+    assert.strictEqual(R.composeP(g, f).length, 1);
+    assert.strictEqual(R.composeP(g, g).length, 2);
 
-    R.composeP(f, f, f)(1).then(function(result) {
-      assert.deepEqual(result, [[[1]]]);
-    });
+    R.composeP(f, g)(1).then(function(result) {
+      assert.deepEqual(result, [[1, undefined]]);
 
-    R.composeP(f, f, g)(1)(2).then(function(result) {
-      assert.deepEqual(result, [[[1, 2]]]);
-    });
+      R.composeP(g, f)(1).then(function(result) {
+        assert.deepEqual(result, [[1], undefined]);
 
-    R.composeP(f, f, h)(1)(2)(3).then(function(result) {
-      assert.deepEqual(result, [[[1, 2, 3]]]);
-    });
+        R.composeP(f, g)(1, 2).then(function(result) {
+          assert.deepEqual(result, [[1, 2]]);
+
+          R.composeP(g, f)(1, 2).then(function(result) {
+            assert.deepEqual(result, [[1], undefined]);
+
+            done();
+          })['catch'](done);
+        })['catch'](done);
+      })['catch'](done);
+    })['catch'](done);
   });
 
   it('throws if given no arguments', function() {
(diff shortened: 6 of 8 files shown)


CrossEye:
:herb:

Well, you just saved me another hour or two on my own PR, which probably would not be as well-written!

I'm all for including `meld`.  (And the name is better than either `mix` or `combine`, which were the two I was trying to decide between.)

paldepind:
Looks good. But to me this still has nothing to do with optional arguments.

davidchambers:
> But to me this still has little to do with optional arguments.

Many of the reported issues have involved "unary" functions with optional arguments (such as `JSON.parse` and `parseInt`) appearing at the beginning of a pipeline. For example:

```javascript
const f = R.compose(R.inc, parseInt);

f('10', 2);   // => 3
f('10', 10);  // => 11
f('10');      // => 11 (previously Number -> Number)
```

You're right, though. Optional arguments are not the only issue. Any curried function at the beginning of a pipeline is currently problematic. When placing a function of type `a -> b -> c` at the beginning of a pipeline one may intend to provide one argument to the pipeline and for the second function in the pipeline to receive a `b -> c` function. This is not currently possible if the `a -> b -> c` reports its `length` as `2` rather than `1`.

So there are two issues: functions with optional arguments at the beginning of a pipeline; and curried functions at the beginning of a pipeline. At Plaid we had more occurrences of the former (lots of pipelines beginning with `JSON.parse`); others may have had more occurrences of the latter. The workaround in both cases has been to begin the pipeline with `R.identity`.

CrossEye:
I think the discussions about optional arguments are my fault.  I raised them as a potential problem with currying the result of compose, and that seems to have stuck as the main issue.  I'm guessing David is using that as short-hand that includes curried functions, which might be seen as functions with arity 1 but with optional parameters.

I agree that the original issue of named optional parameters is minor compared to the issue of people expecting this:

    compose(flip(map)([1, 2, 3]), add)(10); //=> [11, 12, 13]

and not getting it from the post #1260 code.

davidchambers:
> I think the discussions about optional arguments are my fault.

My perception of the issues has been greatly influenced by all the occurrences of `R.pipe(JSON.parse, ...)` in various Plaid projects. I agree that curried functions are the more important concern (though the workaround is the same in both cases, as is the fix).

CrossEye:
Anyone else have input into this?  @buzzdecafe?  @raine? @TheLudd?  @scott-christopher?  I'd like to get 0.18 out soon, and this is the only big thing that I know of that we're still waiting for.

scott-christopher:
This change looks fine to me. I've tested this locally and can confirm the example above works as expected:
```
> R.compose(R.flip(R.map)([1, 2, 3]), R.add)(10)
[ 11, 12, 13 ]
```

buzzdecafe:
pretty disappointing that we have run up against this limitation. but this *has to work*:

```js
R.compose(R.flip(R.map)([1, 2, 3]), R.add)(10) //=> [ 11, 12, 13 ]
```

ivan-kleshnin:
I think it's a mistake to remove currying from `pipe` and `compose`. Behavior of 0.17 Ramda is not ok but I propose to change it in different way. You guys have already made of lot of valuable arguments about the subject. I will reiterate the problem a bit, to make it understandable for occasional reader. 
It's an interesting topic and people may link here.

### My use experience

I always use `pipe`, never use `compose`. I see no sense in the latter because it does not reverse argument order. I use `pipe` even for oneliners because I find it the most readable.

### The problem

**What `pipe(add, map)(10)([1, 2, 3])` should return?**

It's not obvious at all. In Ramda 0.17 it was equal to

```js
map(() => 10 + [1, 2, 3])(); // just wrong
```

rather than 

```js
map(x => x + 10)([1, 2, 3]); // [11, 12, 13]
``` 

### Current solution

Remove autocurrying from `pipe` and `compose` in 0.18.

### Alternative proposal

Define `pipe` and `compose` only for monadic (unary) functions. It was proposed by @davidchambers eariler. Make `pipe` a nonvariadic function. Add `pipe2`, `pipe3`, `compose2`, `compose3` etc versions with appropriate signatures.

```js
pipe([
  doA,
  doB,
  doC,
])(data);

pipe2(doA, doB)(data); 
pipe2(doA)(doB)(data);
comp2(doB, doA)(data);
comp2(doB)(doA)(data);

pipe3(doA, doB, doC)(data);
comp3(doC, doB, doA)(data);
``` 

### Proposal motivation

Currying is a cornerstone of Ramda. Removing it from the most crucial functions in the library 
looks like we are not going to keep currying the first class citizen. Current `pipe` is variadic which kinda breaks "dogfood" expectations. Yes, it's 2 characters less to write but instead it requires to care about trailing comma...

The code
```js
compose(flip(map)([1, 2, 3]), add);
```
is smart but probably "too smart" to live in production. It's probably worth replacing with exact logic it's supposed to reproduce.

```js
let adder = curry((operand, data) => map(add(operand), data));
```

### References

It's worth to look how other languages approach function composition.

**Haskell**: `.` function is defined only for monadic functions.
**Elm**: `>>` is defined only for monadic functions.
**PureScript**: `>>>` is defined only for monadic functions
**Clojure**: functions are not curried by default and there seems to be no official `curry` function 
That's why:
1. `(defn pipe [a & args] ((apply comp (reverse args)) a))`
2. `((pipe + map) 10 [1 2 3])` -- Wrong number of args (1) passed to... 
**Scala**: functions are not curried by default...

### Conclusion

Ramda aims to be a practical library. There is arguably much more practical sense to keep autocurrying for `pipe` and `compose` and require from people to implement their tricky cases in more explicit manner than invent some *very special* behavior for polyadic functions and require from people to wrap every `pipe` and `compose` call in additional `curry` layer.

CrossEye:
@ivan-kleshnin:

While this is an interesting proposal, I have several serious objections. 

> Ramda is based on currying while JS is ultra variadic.

Currying is used heavily in Ramda.  But saying that Ramda is based on currying is an overstatement.   One of Ramda's main themes is functional composition, and currying often helps with that.

This:
> Some bridge from one realm to another has to exist. But I believe this bridge should stay in user-code and stay explicit, instead of lurking into the library core. 

and this:
> When we cut autocurrying from the most crucial functions in the library we're basically saying it's not a first class citizen anymore.

sounds almost in direct opposition.  The following replacement for the latter actually sounds more in keeping with the former:

>> When we add auto-currying to the most crucial functions in the library we're basically saying that the user is not in charge any more.

But I don't buy the basic premise.  If we wanted everything to remain explicit in this manner, we wouldn't ever be auto-currying returned functions.  But we do so, and for a very good reason: The ability to partially apply the more fixed parameters in a simple manner makes for much cleaner, more readable code.

But users want to do this at all sorts of levels.  When we changed this in 0.16, in all sorts of forums, l was regularly answering questions about why things had been broken, or why Ramda was not acting as users expected it should.  And it always came down to this same problem: people expect these to be equivalent:

```js
compose(f, g)(x)
f(g(x))
```

(or the equivalent with `pipe`.)

If `g` happened to be a curried polyadic function, and the first parameter of `f` was a function, this was something that had worked in Ramda since day one; we broke it in #1260.

Who do we satisfy?
------------------

There is a clear argument for currying the results.  The [Alice and Bob story][ab] from @davidchambers demonstrates that either answer is going to make someone unhappy.  I think there is a strong argument to be made, though, for satisfying Bob, who wants the pre-#1260 / post-#1318 behavior, over Alice who prefers the behavior in between.

What Alice wants is to be able to do this:

```js
var fn = R.pipe(R.add, Math.sqrt);
var fn42 = fn(42);
fn42(39); //=> 9  (== sqrt(42 + 39) 
```

What Bob wants to be able to do is this:

```js
var fn = R.pipe(R.add, R.map);
var fn42 = fn(42);
fn([1, 2, 3]) //=> [43, 44, 45]  (== [42 +1, 42 + 2, 42 + 3])
```

In the pre-#1260/post-#1391 code, Alice could fix her issue by currying the results:

```js
var fn = R.curry(R.pipe(R.add, Math.sqrt));
```

In the #1260 - #1391 code, Bob can fix his issue by adding unary no-op:

```js
var fn = R.pipe(R.identity, R.add, R.map);
```

or by wrapping the polyadic function in a lambda:

```js
var fn = R.pipe(n => R.add(n), R.map);
// or
var fn = R.pipe(R.unary(R.add), R.map);
```

To me the difference between the types of solutions is clear.  Alice can fix her issue at the border of the library invocation, by wrapping the results in another call.  The existing `pipe` won't break her application; she just has to wrap its output.

Bob has deeper problems.  If he calls `pipe` against his functions, he's stuck.  He can't convert the result into something useful.  Instead, he has to dig into the internals of how he calls that function and alter them.  This strikes me as significantly more intrusive.

(It's also fairly telling to me that we've had only a handful of complaints about restoring the older behavior; we had many times more about the changes in 0.16 / 0.17.  Popularity will never dictate our design, but it does help reinforce the idea that we're on the right track.)

Basic Proposal
--------------

The essence of the proposal was this:

> Define pipe and compose only for monadic (unary) functions. 

(I'd stay away from the word "monadic" here, and stick with "unary".  "Monadic" also means "having to do with "monads", which is too often relevant to FP code, even in JS.  It just confuses things.)

The other parts of the proposal had to do with the structural details of `pipe` /  `compose` themselves.  Those are orthogonal to the question of whether the first function to be invoked in the chain may be polyadic; we can deal with those separately if necessary.

Although a justification is made for this suggestion by looking at other languages, those are mostly irrelevant.  Clojure doens't really do significant currying.  Haskell and Purescript _only have_ unary functions.  Syntactic tricks sometimes make it look as though there can have others, but the underlying system is so different from JS that it doesn't offer any real guidance here..  Elm is less clear to me; I haven't played with it at all.  But at first glance it seems to be cut from the same cloth as Haskell.  And Scala doesn't offer any related features.  Perhaps @buzzdecafe can give us feedback from F# and OCaml.


My Take
--------

My feeling is that restricting `compose` / `pipe` to unary functions is far to great a burden on JS developers, who are used to building polyadic functions at will, and who use them for all sorts of reasonable purposes.  While doing this might help make code written in Ramda look a little closer to code written in Haskell or ML, that has never been a design goal of the library.

So, while I might still like to see something like Sanctuary's `meld` function included, I'm pretty happy with the essence of `pipe` / `compose` in Ramda.


  [ab]: https://github.com/ramda/ramda/issues/1318#issuecomment-139103766

davidchambers:
> Define `pipe` and `compose` only for [unary] functions.

[`S.compose`][1] and [`S.pipe`][2] are available for those who prefer this behaviour. :)


[1]: https://github.com/plaid/sanctuary#compose
[2]: https://github.com/plaid/sanctuary#pipe

ivan-kleshnin:
@CrossEye thank you for thorough answer with valid arguments. Can't deny them.

@davidchambers interesting proposal. 
I'm still a bit confused about relations between Ramda, Fantasy Land and Sanctuary. 
I see some maybe-related code in Ramda and both FL and Sanctuary provide their own typesystem.
Can't see a bigger picture which would include all of them.

davidchambers:
> I'm still a bit confused about relations between Ramda, Fantasy Land and Sanctuary.

Fantasy Land is a specification. Ramda and Sanctuary are two of the [many libraries][1] which respect the specification.

Ramda exports several unsafe functions such as `R.head` and `R.find`. Sanctuary exports safe versions of these functions (backed by the Either and Maybe types). Those who want more type safety than Ramda provides (and descriptive error messages!) can use Sanctuary in conjunction with Ramda. Others can use Ramda on its own.

I hope this clarifies the situation. :)


[1]: https://github.com/fantasyland/fantasy-land/blob/master/implementations.md

CrossEye:
A little more on the relation of Ramda/Sanctuary with FantasyLand:

Although what is specified by FantasyLand is what Haskell calls `typeclasses` the basic notion is simply one of _algebraic types_, types defined by a collection of functions/methods that apply to them and laws about how those functions interact.  

The specific types are ones that have shown themselves to be quite useful for many developers over a long time.  They are generally fairly abstract, but the laws specifying their interaction make them quite useful.  For instance,  `Functor` defines `map`, and the composition law of `Functor` says that (in Ramda terms)

```js
compose(map(f), map(g)) ≍ map(compose(f, g))
```

This is very much like the mathematical distributive property of multiplication over addition:

```
 (x * y) + (x * z) ≍ x * (y + z) 
```

(You can see this by considering those as functions of `x`:

```js
add(mult(y), mult(z)) ≍ mult(add(y, z))
```


This implies that by purely mechanical means, without knowing any details about our type or how its `map` method is implemented, we might be able to rewrite expressions involving it into a simpler form.  That is the power of the FantasyLand abstractions.

A number of libraries create types that implement some of these algebraic specifications, including [Folktale][fo], [RamdaFantasy][rf], and Sanctuary[sa].  Ramda itself does not try to implement any of these specifications.  But many of its functions delegate to named methods on the relevant objects, and this include almost all of the functions specified by the various algebraic types of FantasyLand.

So a user can use, say Folktale's `Maybe` implementation, which implements, among others, the `Functor` and `Apply` specifications, and know that they will interoperate with Ramda's `map` and `ap` functions properly.  For example,

```js
R.map(square, Maybe.Just(5)); //=> Maybe.Just(25)
R.map(square, Maybe.Nothing); //=> Maybe.Nothing
```

Moreover a few other Ramda functions, such as `lift` are defined in terms of these FantasyLand-specified functions, so they will also work on these types.

```js
const mAdd = R.lift(R.add);
mAdd(Maybe.Just(35), Maybe.Just(7)); //=> Maybe.Just(42)
```

Ramda has no `Maybe`-related code; there are some examples and tests that mention `Maybe`s, to help document how Ramda functions work with these algebraic types.  But Ramda itself does not use them at all.

When @davidchambers says that Sanctuary exports safe versions of unsafe Ramda functions, he's talking about Sanctuary defining its own Maybe and Either types, (which meet various FantasyLand specifications for these types) and then using them as the result of functions that might otherwise throw exceptions or return `null`/`undefined` values.

Sanctuary was originally built mostly as an adjunct to Ramda to provide safe versions of some of these functions, but it's grown significantly.  While it still is built on top of Ramda, it is a substantive piece of work on its own.


In summary, Sanctuary is a collection of specifications.  Ramda does not implement them at all, but works very well with types that do implement them.  Sanctuary implements a few and uses them to replace some Ramda functions with safer versions. 

  [fo]: https://github.com/folktale
  [sa]: https://github.com/plaid/sanctuary

scott-christopher:
Great overview, @CrossEye.

Minor correction - I believe you mean _FantasyLand_ here:
> In summary, Sanctuary is a collection of specifications.

CrossEye:
@scott-christopher: Thank you.

I've updated to correct that mistake.  Thanks.

ivan-kleshnin:
@davidchambers, @CrossEye thanks. It's clear now.
I have an additional question: why some functions in Ramda are "unsafe" then?

CrossEye:
@ivan-kleshnin:

Ramda made the decision not to require users to know anything about the algebraic types in order to use it.  This was a controversial decision, and it was part of the reason for the creation of Sanctuary.  The reason for not introducing, for instance, `Maybe` is pretty straightforward:

Replacing the current `head` function `[a] -> a | Undefined` with the simpler `[a] -> Maybe a` would require a significant change in code structure on the part of many users.  While the benefits of that new structure are great, Ramda is meant to introduce important FP concepts to the Javascript world and not to serve every FP need it might ever have.  (There's always [Bilby][bi] for that.)  It seemed more important that it was immediately useful to users than that it was as powerful as it might be.


  [bi]: http://bilby.brianmckenna.org/

ivan-kleshnin:
Haskell throws an exception on `head []`. I wonder which behavior is better... I don't solve enough algorithmic tasks to answer. But your points are clear.

PR #1397, opened 2015-09-23
make R.useWith a binary function

### LINKED ISSUE #1104: R.useWith doesn't work as expected if called with R.__ for the function argument
See http://bit.ly/1Bdkpls vs. http://bit.ly/1K3MEZ5

syves:
Closes #1104 

The examples given for this function are quite verbose. Would you be interested in having me simplify these for you? A one line example function would make grasping it's purpose much easier.

### DIFF
--- a/src/useWith.js
+++ b/src/useWith.js
@@ -1,10 +1,11 @@
 var _arity = require('./internal/_arity');
+var _curry2 = require('./internal/_curry2');
 var _slice = require('./internal/_slice');
 var curry = require('./curry');
 
 
-/**
- * Accepts a function `fn` and any number of transformer functions and returns a new
+/*
+ * Accepts a function `fn` and a list of transformer functions and returns a new curried
  * function. When the new function is invoked, it calls the function `fn` with parameters
  * consisting of the result of calling each supplied handler on successive arguments to the
  * new function.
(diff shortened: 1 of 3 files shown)


davidchambers:
Thanks, @syves! I far prefer the binary version of `useWith`. :)

> The examples given for this function are quite verbose. Would you be interested in having me simplify these for you?

Absolutely! The best idea may be to propose new examples in a comment in this thread, then update the patch if people agree they're preferable. I'm fond of trivial examples rather than "realistic" examples, so you're likely to get a `:+1:` from me.

src/useWith.js
```diff
@@ -44,20 +44,18 @@ var curry = require('./curry');
...
+ *      var addDoubleAndSquareWithExtraParams = R.useWith(addAll, [double, square, R.identity]);
  *      // addDoubleAndSquareWithExtraParams.length //=> 3
  *      //≅ addAll(double(10), square(5), R.identity(100));
  *      addDoubleAndSquare(10, 5, 100); //=> 145
  */
-module.exports = curry(function useWith(fn /*, transformers */) {
-  var transformers = _slice(arguments, 1);
-  var tlen = transformers.length;
-  return curry(_arity(tlen, function() {
+module.exports = curry(function useWith(fn, transformers) {
```
davidchambers: We have an optimized [`_curry2`][1] function for currying binary functions. Let's use it here in place of `curry`.


[1]: https://github.com/ramda/ramda/blob/v0.17.1/src/internal/_curry2.js

jdalton:
@syves  mind changing `converge` too?

syves:
@jdalton I'm happy to look into Converge, but I don't see any open issues, could you open one?

jdalton:
#1401

CrossEye:
:herb:  

And :+1: on better examples, too.

davidchambers:
I consider these examples sufficient:

```javascript
R.useWith(Math.pow, [R.identity, R.identity])(3, 4); //=> 81
R.useWith(Math.pow, [R.identity, R.identity])(3)(4); //=> 81
R.useWith(Math.pow, [R.dec, R.inc])(3, 4); //=> 32
R.useWith(Math.pow, [R.dec, R.inc])(3)(4); //=> 32
```

syves:
I updated this pr.

test/useWith.js
```diff
@@ -8,12 +8,12 @@ describe('useWith', function() {
   function add1(x) { return x + 1; }
   function mult2(x) { return x * 2; }
   function div3(x) { return x / 3; }
-  var f = R.useWith(max, add1, mult2, div3);
+  var f = R.useWith(max, [add1, mult2, div3]);
 
   it('takes a arbitrary number of function arguments and returns a function', function() {
```
davidchambers: This description is no longer accurate. Let's state that it _takes a function and a list of functions and returns a function_.

davidchambers:
Thanks, @syves! I noticed one test description which should be updated. Once you've done this you'll have a `:deciduous_tree:` from me. :)

test/useWith.js
syves: I updated the test description.

syves: @davidchambers is this description as brief as it could be?
--> it('nonetheless has the correct arity', function() {
    assert.strictEqual(f.length, 3);
  });
--> perhaps nonetheless is superfluous ;)

davidchambers: I agree that "nonetheless" is superfluous! :)

syves:
updated pr!

davidchambers:
:deciduous_tree: Thanks, @syves. Lovely first contribution to Ramda!

PR #1400, opened 2015-09-24
remove _hasMethod

### LINKED ISSUE #1399: R.ap() does not dispatch correctly when using native Array as functor
In the light of this functor:

```js
// Identity
var Identity = function(val) {
  this.val = val;
};
Identity.prototype.map = function(fn) { // Functor
  return Identity.of(fn(this.val));
};
Identity.prototype.ap = function(app) { // Apply
  return app.map(this.val);
};
Identity.of = function(val) { // Applicative
  return new Identity(val);
};
```

The native JavaScript Array is a functor by the Fantasy Land definition. In addition, I've implemented the Apply and Applicative spec directly on the Array prototype:
```js
Array.prototype.ap = function(app) { // Apply
  return this.map(fn => app.map(fn));
};
Array.of = function(val) { // Applicative
  return [val];
};
```

Then when I'd like to use my Apply, this works for my Identity functor, but not for my Array functor:
```js
R.ap(Identity.of(R.add(5)), Identity.of(50)); //=> Identity(55)
R.ap([R.add(3), R.add(5)], Identity.of(50)); //=> []
```

I would expect the last line to return the same as using `.ap()` directly on the functor:
```js
[R.add(3), R.add(5)].ap(Identity.of(50)); //=> [Identity(53), Identity(55)]
```
Instead, I get an empty array.

My guess would be that Ramda's `.ap()` does not dispatch correctly somehow when I'm extending JavaScript's native Array as a functor. From a quick scan of the ramda source, I'd say the problem is in the hardcoded check for [`isArray` within `hasMethod`](https://github.com/ramda/ramda/blob/v0.17.1/src/internal/_hasMethod.js#L22).

Is this expected behaviour? For me it's certainly not.

I realize that piggybacking the `.ap()` implementation on top of the native JavaScript Array is probably not the best solution, but adding another array-like type to JavaScript is not a great solution either imho.

davidchambers:
Closes #1399

`_hasMethod` does more than its name suggests: `_hasMethodAndIsNotAnArray` would be a more accurate name. It's currently used in places where arrays needn't be treated specially, such as in `R.clone` (see https://github.com/ramda/ramda/pull/1295#issuecomment-123403131).

This pull request replaces all occurrences of `_hasMethod` with the logic we actually mean to express. In every case this reduces the number of checks we perform: in some cases the null check is unnecessary; in some cases the `_isArray` check is inappropriate.


CrossEye:
Although `_hasMethod` is the last vestige of an older dispatch system, what does it help to clean up this one and not touch `_dispatchable`, which does the same sort of `isArray()` testing?

I suppose it's good to get rid of it, but the basic question is left unanswered:

Do we delegate first on lists for the functions which Ramda defines, allowing object methods to override Ramda's definitions even there?  And if the latter, do we throw out our internal `map`, `filter`, etc.?  And if so, would we do something to make these functions unary?

davidchambers:
Here's the general approach I suggest we take to resolve `R.asdf(x, y)` which may dispatch to `y.asdf`:

1.  If `y` is an Array and `'asdf'` is a member of the set of known Array method names, use code path for array-like objects (which might simply involve dispatching).

2.  Otherwise, if `typeof y['asdf'] === 'function'`, dispatch.

3.  Otherwise, use code path for array-like objects.

PR #1405, opened 2015-09-25
dropLast and dropLastWhile can act as transducer

gonzaloruizdevilla:
two more list functions that can act as transducers

### DIFF
--- a/src/dropLast.js
+++ b/src/dropLast.js
@@ -1,5 +1,7 @@
 var _curry2 = require('./internal/_curry2');
-var take = require('./take');
+var _dispatchable = require('./internal/_dispatchable');
+var _dropLast = require('./internal/_dropLast');
+var _xdropLast = require('./internal/_xdropLast');
 
 
 /**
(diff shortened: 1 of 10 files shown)


davidchambers:
You understand the transducer code better than I do, @gonzaloruizdevilla! Thanks for the patch. LGTM.

gonzaloruizdevilla:
you're welcome!  :)

CrossEye:
I'm looking at it now, and feeling confused, as I always wanted transducers to help with laziness.  Obviously `lastXXX` is not going to work with lazy collections, so again my expectations are a little off.  :smile:

CrossEye:
Still a bit confused by the impl, but the tests look good, and if it's passing those, I think I just need to spend more time with transducers.  :smiling_imp:

gonzaloruizdevilla:
I'll try to explain the implementations. They should work quite well with lazy sources (takeLast for example won't be so good).

dropLast `n`: It starts storing elements until it has `n`elements. Afterwards, when a new element comes the oldest one goes to the next transducer (FIFO). (I think I should remove the `retained` array in `'@@transducer/result'` to prevent retaining inputs in memory after the processing has been finished when the reference to the transducer is not removed).

dropLastWhile `fn`: if fn returns true, the item might be dropped so it gets retained. If fn returns false, it implies that what was retained until that point is not going to be dropped, so all the retained items are sent to the next transducer, and retained array reseted. (Again, here I have to do some more work to prevent maintaining references to inputs when they are not longer needed).

davidchambers:
Thanks for the explanations, @gonzaloruizdevilla.

Could you squash the commits?

gonzaloruizdevilla:
Sorry for the delay. Is it ok like this?
One commit for each new feature and one commit for each fix on a different function.

davidchambers:
Thanks! Merging.

Issue #1054, opened 2015-04-28
pathOr ?


Cottin:
Since there is `propOr` would it make sense to include a `pathOr` function?

davidchambers:
I agree that it makes sense. The problem is that since Ramda doesn't have an unambiguous way to represent a failed operation, we end up needing a `R.____Or` function for each function which may fail. We could end up with `R.propOr`, `R.pathOr`, `R.headOr`, `R.lastOr`, `R.findOr`, and several others.

[Sanctuary][1] was conceived as an alternative to a proliferation of such functions.


[1]: https://github.com/plaid/sanctuary

CrossEye:
Nonetheless, we can make a judgment call about which functions we find worth supporting.  At the moment, we have `reduce` and `reduceRight`, and have `scan` but no `scanRight`.  Perhaps we should correct that, but ["a foolish consistency is the hobgoblin of little minds..."][rwe]


  [rwe]: http://en.wikipedia.org/wiki/Self-Reliance

davidchambers:
`R.pathOr` seems useful to me. `R.headOr` and `R.lastOr` aren't as useful since the alternative, `xs.length > 0 ? ... : ...`, is manageable.

CrossEye:
And I meant to say that.  I would welcome a PR for `pathOr`.

davidchambers:
If anyone is interested in pursuing this, please submit a pull request. :)

PR #1192, opened 2015-06-15
Revert "remove array length caching"

megawac:
As discussed reverts ramda/ramda#1167

----------------------

Probably justifies a patch release (0.15.1)

PR #1406, opened 2015-09-26
Add R.pathOr

jimf:
Implement R.pathOr as per the invitation for PR in #1054.

### DIFF
--- a/src/pathOr.js
+++ b/src/pathOr.js
@@ -1,0 +1,24 @@
+var _curry3 = require('./internal/_curry3');
+var defaultTo = require('./defaultTo');
+var path = require('./path');
+
+
+/**
+ * If the given, non-null object has a value at the given path, returns
+ * the value at that path. Otherwise returns the provided default value.
+ *
+ * @func
+ * @memberOf R
+ * @category Object
+ * @sig a -> [String] -> Object -> a
+ * @param {*} d The default value.
+ * @param {Array} p The path to use.
+ * @return {*} The data at `path` of the supplied object or the default value.
+ * @example
+ *
+ *      R.pathOr('N/A', ['a', 'b'], {a: {b: 2}}); //=> 2
+ *      R.pathOr('N/A', ['a', 'b'], {c: {b: 2}}); //=> "N/A"
+ */
+module.exports = _curry3(function pathOr(d, p, obj) {
+  return defaultTo(d, path(p, obj));
+});
--- a/src/path.js
+++ b/src/path.js
@@ -20,8 +20,10 @@ module.exports = _curry2(function path(paths, obj) {
     return;
   } else {
     var val = obj;
-    for (var idx = 0, len = paths.length; idx < len && val != null; idx += 1) {
+    var idx = 0;
+    while (val != null && idx < paths.length) {
       val = val[paths[idx]];
+      idx += 1;
     }
     return val;
   }
--- a/test/pathOr.js
+++ b/test/pathOr.js
@@ -1,0 +1,55 @@
+var assert = require('assert');
+
+var R = require('..');
+
+
+describe('pathOr', function() {
+  var deepObject = {a: {b: {c: 'c'}}, falseVal: false, nullVal: null, undefinedVal: undefined, arrayVal: ['arr']};
+
+  it('takes a path and an object and returns the value at the path or the default value', function() {
+    var obj = {
+      a: {
+        b: {
+          c: 100,
+          d: 200
+        },
+        e: {
+          f: [100, 101, 102],
+          g: 'G'
+        },
+        h: 'H'
+      },
+      i: 'I',
+      j: ['J']
+    };
+    assert.strictEqual(R.pathOr('Unknown', ['a', 'b', 'c'], obj), 100);
+    assert.strictEqual(R.pathOr('Unknown', [], obj), obj);
+    assert.strictEqual(R.pathOr('Unknown', ['a', 'e', 'f', '1'], obj), 101);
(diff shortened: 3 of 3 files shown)


jimf:
NOTE: I've more or less re-implemented R.path here, returning the default value in place of `undefined`. I'm thinking it would be preferable to instead leverage R.path directly and return the default if the return value was `undefined`, but I'm open to suggestions. New to this codebase.

davidchambers:
Thanks for the patch, @jimf!

src/pathOr.js
```diff
@@ -0,0 +1,30 @@
+var _curry3 = require('./internal/_curry3');
+
+
+/**
+ * If the given, non-null object has an own property at the given path, returns
+ * the value at that path. Otherwise returns the provided default value.
```
davidchambers: We're doing `val[paths[idx]]`, so the property needn't be an *own* property. Let's update the description accordingly.

src/pathOr.js
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * @func
+ * @memberOf R
+ * @category Object
+ * @sig a -> [String] -> {k: v} -> v | a
```
davidchambers: I suggest `a -> [String] -> Object -> a` as the type signature. The default value should be of the same type as the expected value.

src/pathOr.js
```diff
@@ -0,0 +1,30 @@
...
+ * @param {*} d The default value.
+ * @param {Array} path The path to use.
+ * @return {*} The data at `path`.
+ * @example
+ *
+ *      R.pathOr('N/A', ['a', 'b'], {a: {b: 2}}); //=> 2
+ *      R.pathOr('N/A', ['a', 'b'], {c: {b: 2}}); //=> "N/A"
+ */
+module.exports = _curry3(function path(d, paths, obj) {
```
davidchambers: It seems to me the second parameter should be named `path`. There is, after all, only one path.

davidchambers: Also, the name of the function should be `pathOr`.

src/pathOr.js
```diff
@@ -0,0 +1,30 @@
...
+  if (obj == null) {
+    return d;
+  } else {
+    var val = obj;
+    for (var idx = 0, len = paths.length; idx < len && val != null; idx += 1) {
+      val = val[paths[idx]];
+    }
```
davidchambers: Ramda style favours `while` loops over `for` loops. Also, we prefer not to cache array lengths unless necessary. We might write:

```javascript
var val = obj;
var idx = 0;
while (val != null && idx < path.length) {
  val = val[path[idx]];
  idx += 1;
}
```

src/pathOr.js
jimf: +1

src/pathOr.js
jimf: +1, that makes sense.

src/pathOr.js
jimf: +1 on `path` vs `paths`. I was just staying consistent with `R.path`'s implementation. As for the function name, doh!

src/pathOr.js
jimf: +1, will change.

jimf:
@davidchambers your comments have been addressed.

A couple of your comments would apply to the `R.path` implementation as well. Thoughts on making those same mods there, either as part of this PR or another?

raine:
I'm wondering about the reasons against implementing this in terms of `path` and `defaultTo` internally.

Are we trying to keep functions independent in case they are deprecated?

davidchambers:
> I'm wondering about the reasons against implementing this in terms of `path` and `defaultTo` internally.

It seems we could use these functions to avoid duplicating code. Mind trying that, @jimf?

> A couple of your comments would apply to the `R.path` implementation as well. Thoughts on making those same mods there, either as part of this PR or another?

Let's include the `R.path` tweaks in this pull request.

jimf:
+1, fully agree.

jimf:
@davidchambers @raine comments addressed!

src/pathOr.js
```diff
@@ -0,0 +1,24 @@
...
+ * If the given, non-null object has a value at the given path, returns
+ * the value at that path. Otherwise returns the provided default value.
+ *
+ * @func
+ * @memberOf R
+ * @category Object
+ * @sig a -> [String] -> Object -> a
+ * @param {*} d The default value.
+ * @param {Array} p The path to use.
+ * @return {*} The data at `path`.
```
davidchambers: This description is not entirely accurate. Let's either reword or remove it.

davidchambers:
There's just one minor issue for you to address, @jimf, before we merge this pull request.

jimf:
@davidchambers Updated!

jimf:
Regarding my second commit, I just want to resurface #1192 and make sure we still want that.

davidchambers:
> Regarding my second commit, I just want to resurface #1192 and make sure we still want that.

We're in favour of `while` loops over `for` loops, and we like to avoid manual `.length` caching unless there's a possibility of the value changing during iteration. In this case a `while` loop without manual `.length` caching exactly matches our preferences. Thank you for fixing the style violations. :)

Please squash the commits (or append an "s" to the username mentioned in the second commit message).

jimf:
@davidchambers Thanks for the clarification. Commits have been reworded and squashed.

davidchambers:
Thanks! I'll merge as soon as Travis gives the go-ahead.

davidchambers:
:deciduous_tree:

jimf:
Thanks! :beers:

Issue #1402, opened 2015-09-24
InvertObj yields unexpected results when called with a deep nested object


dumconstantin:
```javascript
var obj = {
    a: { foo: 'bar' },
    b: { baz: 'bam' }
}
var invertedObj = R.invertObj(obj); // Object {[object Object]: "b"}
```
- Prop "a" is lost without trace (this also happens if both "a" and "b" have the same value .e.g Object {foo: "b"})
- Object.keys(invertedObj).length !== Object.keys(obj)
- The initial value is replaced by its toString() value making this operation irreversible.

This is very subtle bug to figure out, maybe an error/complete invalidation of result might save the day? Or a debate if the benefits of this function out weight the possible bugs that might occur?

The following is a naive JSON implementation but does it make sense?

```javascript
  // Extracted from https://github.com/ramda/ramda/blob/master/src/invertObj.js

  // Add N space characters together
  var spacesN = R.compose(R.join(''), R.times(R.always(' ')));
  var history = [];
  var newKey = "";

  // Added conditional
  if (typeof obj[key] === 'object') {
    newKey = JSON.stringify(obj[key]);
  } else {
    newKey = obj[key]
  }
  // Makes sure duplicates don't overwrite each other
  if (history.indexOf(newKey) !== -1) {
     newKey += spacesN(idx);
  }
  history.push(newKey);
  out[newKey] = key;
```
Of course, the above code might introduce even more subtle bugs.

CrossEye:
This function won't lose the duplicate: [`invert`][in].

I'm afraid that neither makes much sense if your values are not Strings (or easily converted to Strings.) 

Could you put your code together as a complete sample function, either here or in a gist, or in the [repl][re]?


  [in]: http://ramdajs.com/docs/#invert
  [re]: http://ramdajs.com/repl/

dumconstantin:
I've put some code together here:

da7edbdebcefe605095e8b2031c9c2316203c4fa

If this is a good idea I can add further support for non-primitive types and get some tests going for those.

CrossEye:
Ok, I see what you're doing.  Honestly, this seems like quite bizarre behavior to me.

My thought is that this is simply a case of "Doctor it hurts when I do _this_."  "Then don't _do_ that!".

I simply don't see a good reason to use `invert` on objects that aren't simple, one-level deep data objects.  If you need more sophisticated processing, then you probably need a different tool.

What do others think?

davidchambers:
[...]

PR #1407, opened 2015-09-27
Add docs on duplication and non-primitive objects for R.invertObj

dumconstantin:
R.invertObj has some cases that developers might need  to consider when choosing this function.
See issue for more details: #1402

### DIFF
--- a/src/invertObj.js
+++ b/src/invertObj.js
@@ -4,7 +4,10 @@ var keys = require('./keys');
 
 /**
  * Returns a new object with the keys of the given object
- * as values, and the values of the given object as keys.
+ * as values, and the values of the given object, which are
+ * coerced to strings, as keys.
+ * Note that the last key found is preferred when handling
+ * the same value.
  *
  * @func
  * @memberOf R


src/invertObj.js
```diff
@@ -5,6 +5,9 @@ var keys = require('./keys');
 /**
  * Returns a new object with the keys of the given object
  * as values, and the values of the given object as keys.
+ * Note that the last key found is prefered when handling
+ * the same value. Non-primitive values are converted to
+ * strings which might yield undesired results.
```
CrossEye: I think we can drop the "which might yield undesired results."

davidchambers: Why "non-primitive"? Primitives such as `true` and `42` are also coerced to strings. Perhaps this sentence should read:

_Values of the input object are coerced to strings to determine the keys of the output object._

dumconstantin: Thanks for the corrections!
- Removed the *undesired results*
- Added the *coercion to string* in the first sentence to avoid repetition

CrossEye:
Looks good.  Could you squash the commits?

dumconstantin:
Squashed :)

src/invertObj.js
```diff
@@ -4,7 +4,10 @@ var keys = require('./keys');
 
 /**
  * Returns a new object with the keys of the given object
- * as values, and the values of the given object as keys.
+ * as values, and the values of the given object, which are
+ * coerced to string, as keys.
+ * Note that the last key found is prefered when handling
```
davidchambers: `s/prefered/preferred/`

src/invertObj.js
```diff
@@ -4,7 +4,10 @@ var keys = require('./keys');
 
 /**
  * Returns a new object with the keys of the given object
- * as values, and the values of the given object as keys.
+ * as values, and the values of the given object, which are
+ * coerced to string, as keys.
```
davidchambers: `s/string/strings/`

davidchambers:
Nice changes, @dumconstantin. I left two comments on minor things. Once you've addressed these, we'll merge. Please use `git commit --amend` to avoid introducing extra commits. :)

dumconstantin:
@davidchambers done :)

davidchambers:
:deciduous_tree:

dumconstantin:
Amazing :) Thanks!

Issue #1401, opened 2015-09-24
Align converge with useWith


jdalton:
Align converge with useWith in making it a binary function. Related to #1397.

CrossEye:
If we're going to do one, we definitely should do both...

Issue #1140, opened 2015-06-01
R.is implementation has the problematic characteristics of instanceof 


asaf-romano:
Long story short: http://bit.ly/1I3wwqf

This isn't an issue in node.js environments (at least not in non-esoteric usage... I guess one could get multiple globals through the vm modules), because node uses a single global object for all modules. However, in browser environments, and probably in some other JS environments, having to deal with multiple global scopes is a quite common obstacle, and also one of the reasons why some developers avoid using instanceof.

There is no perfect solution for this, but I think that at least for native js types, Ramda should work "as expected" (i.e. in the practical manner ;-)). The obvious example is arrays: I don't think the current difference between Array.isArray to R.is(Array) is something that can ever be beneficial to anyone.

On a side note, an interesting edge case is |R.is(Object, Object.create(null))|, which returns false (as expected) even though Object.create(null)| walks and does almost all of the things an object does - it's just not an Object (Object.create(null) is often used to avoid accidentally picking Object.prototype keys when using an object in the role of a map).

CrossEye:
Somehow I missed this one when it was posted.  You're right that this is unfortunate.  Obviously we can switch to the sort of solution kangax gave some years ago for the known types, but it would be some pretty ugly code, involving something like:

    Object.prototype.toString.call(val) === '[object Array]'

    Object.prototype.toString.call(val) === '[object Date]'

etc., with a separate path for anything other than the known types.  Do you have any other suggestion?

asaf-romano:
This seems to work ok:
http://tinyurl.com/o3eab7p

("Make short URL" doesn't work there for whatever reason)

benjycui:
@asaf-romano Function.name [does not work in IE](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name).
```js
var fixedIsImpl = R.curry((ctor, obj) => {
  if (R.type(obj) == "Object" && ctor != Object) {
    return R.is(ctor, obj);
  }
                      
  return R.type(obj) == ctor.name;  // I mean ctor.name
});
```

@CrossEye I can not find any documentation about the browser compatibility of Ramda. Is that OK?

buzzdecafe:
@benjycui we have a sauce suite, we just have to get it running again. the problem was it was running *too much* (and timing out). Anyways, ramda should be ES3-compatible, for better or worse.

asaf-romano:
[...]

PR #1408, opened 2015-09-27
implement map, ap and chain for functions

benperez:
This change was motivated by some [discussion](https://github.com/plaid/sanctuary/issues/102) around porting some higher order Ramda functions into Sanctuary. The common theme among many of these functions is that they are essentially gluing together the results of different functions in different ways. If we implement Functor and Applicative for functions then we get this behavior for free.
```haskell
instance Functor ((->) r) where
    fmap = (.)

instance Applicative ((->) a) where
    pure = const
    (<*>) f g x = f x (g x)
```
The following simplifications become possible
 - `R.both` -> `R.lift(R.and)`
 - `R.either` -> `R.lift(R.or)`
 - `R.negate` -> `R.lift(R.not)`
 - `R.converge` -> `R.lift`

There are 2 issues that I ran into

What to do about converging functions that take multiple arguments, for example: 
```javascript
const f = R.converge((a, b) => a + b, a => a, (a, b) => a);
f.length; // 2
```
Right now `ap` just assumes there will only be one argument supplied to a function (but I could easily change it to apply the function to the entire arguments object). `R.converge` takes the max length of all of the functions but it can only do this because it knows the length of both input functions at the same time. If you were to perform the equivalent operation with `R.lift` you'd supply the input functions 1 at a time (although both at once is possible as well):
```javascript
const g = R.lift((a, b) => a + b); 
// g :: (Applicative f, Num c) => f c -> f c -> f c
const h = g(a => a); // should the function returned have length 1 now? How do you keep track of that? 
// h :: (Applicative f, Num c) => f c -> f c
const i = h((a, b) => a); // should the function returned have length 2 now?
```
I think the simplest solution is just to make `R.both`, `R.either`, etc. just take unary functions as arguments and call it a day.

The second problem is around preserving context but it seems like that has already been figured out for converge, I'm just not sure how to add it in to `R.map` and `R.ap`.

### DIFF
--- a/src/ap.js
+++ b/src/ap.js
@@ -1,11 +1,14 @@
 var _concat = require('./internal/_concat');
 var _curry2 = require('./internal/_curry2');
 var _reduce = require('./internal/_reduce');
+var curryN = require('./curryN');
 var map = require('./map');
 
 
 /**
  * ap applies a list of functions to a list of values.
+ *
+ * ap also treats functions as applicatives.
  *
  * @func
  * @memberOf R
@@ -18,8 +21,15 @@ var map = require('./map');
  *
  *      R.ap([R.multiply(2), R.add(3)], [1,2,3]); //=> [2, 4, 6, 4, 5, 6]
  */
-module.exports = _curry2(function ap(fns, vs) {
-  return typeof fns.ap === 'function' ?
-    fns.ap(vs) :
-    _reduce(function(acc, fn) { return _concat(acc, map(fn, vs)); }, [], fns);
+module.exports = _curry2(function ap(applicative, fn) {
+  return (
+    typeof applicative.ap === 'function' ?
+      applicative.ap(fn) :
+    typeof applicative === 'function' ?
+      curryN(Math.max(applicative.length, fn.length), function() {
+        return applicative.apply(this, arguments)(fn.apply(this, arguments));
+      }) :
+    // else
+      _reduce(function(acc, f) { return _concat(acc, map(f, fn)); }, [], applicative)
+  );
 });
--- a/src/complement.js
+++ b/src/complement.js
@@ -1,5 +1,5 @@
-var _complement = require('./internal/_complement');
-var _curry1 = require('./internal/_curry1');
+var lift = require('./lift');
+var not = require('./not');
 
 
 /**
@@ -10,6 +10,8 @@ var _curry1 = require('./internal/_curry1');
  *
  *   - applying `g` to zero or more arguments will give __false__ if applying
  *     the same arguments to `f` gives a logical __true__ value.
+ *
+ * `R.complement` will work on all other functors as well.
  *
  * @func
  * @memberOf R
@@ -25,4 +27,4 @@ var _curry1 = require('./internal/_curry1');
  *      isOdd(21); //=> true
  *      isOdd(42); //=> false
  */
-module.exports = _curry1(_complement);
+module.exports = lift(not);
--- a/src/internal/_map.js
+++ b/src/internal/_map.js
@@ -1,8 +1,16 @@
-module.exports = function _map(fn, list) {
-  var idx = 0, len = list.length, result = Array(len);
-  while (idx < len) {
-    result[idx] = fn(list[idx]);
-    idx += 1;
+var curryN = require('../curryN');
+
+module.exports = function _map(fn, functor) {
+  if (typeof functor === 'function') {
+    return curryN(functor.length, function() {
+      return fn.call(this, functor.apply(this, arguments));
+    });
+  } else {
+    var idx = 0, len = functor.length, result = Array(len);
+    while (idx < len) {
+      result[idx] = fn(functor[idx]);
+      idx += 1;
+    }
+    return result;
   }
-  return result;
 };
--- a/test/liftN.js
+++ b/test/liftN.js
@@ -28,12 +28,6 @@ describe('liftN', function() {
     assert.deepEqual(addN3([1, 10], [2], [3]), [6, 15]);
   });
 
-  it('produces a cross-product of array values', function() {
-    assert.deepEqual(addN3([1, 2, 3], [1, 2], [1, 2, 3]), [3, 4, 5, 4, 5, 6, 4, 5, 6, 5, 6, 7, 5, 6, 7, 6, 7, 8]);
-    assert.deepEqual(addN3([1], [2], [3]), [6]);
-    assert.deepEqual(addN3([1, 2], [3, 4], [5, 6]), [9, 10, 10, 11, 10, 11, 11, 12]);
-  });
-
   it('can lift functions of any arity', function() {
     assert.deepEqual(addN3([1, 10], [2], [3]), [6, 15]);
     assert.deepEqual(addN4([1, 10], [2], [3], [40]), [46, 55]);
@@ -49,5 +43,25 @@ describe('liftN', function() {
   it('works with other functors such as "Maybe"', function() {
     var addM = R.liftN(2, R.add);
     assert.deepEqual(addM(Maybe(3), Maybe(5)), Maybe(8));
+  });
+
+  it('interprets [a] as a functor', function() {
+    assert.deepEqual(addN3([1, 2, 3], [10, 20], [100, 200, 300]), [111, 211, 311, 121, 221, 321, 112, 212, 312, 122, 222, 322, 113, 213, 313, 123, 223, 323]);
+    assert.deepEqual(addN3([1], [2], [3]), [6]);
+    assert.deepEqual(addN3([1, 2], [10, 20], [100, 200]), [111, 211, 121, 221, 112, 212, 122, 222]);
+  });
+
+  it('interprets ((->) r) as a functor', function() {
+    var convergedOnInt = addN3(R.add(2), R.multiply(3), R.subtract(4));
+    var convergedOnBool = R.liftN(2, R.and)(R.gt(R.__, 0), R.lt(R.__, 3));
+    assert.strictEqual(typeof convergedOnInt, 'function');
+    assert.strictEqual(typeof convergedOnBool, 'function');
+    assert.strictEqual(convergedOnInt(10), (10 + 2) + (10 * 3) + (4 - 10));
+    // jscs:disable disallowYodaConditions
+    assert.strictEqual(convergedOnBool(0), (0 > 0) && (0 < 3));
+    assert.strictEqual(convergedOnBool(1), (1 > 0) && (1 < 3));
+    assert.strictEqual(convergedOnBool(2), (2 > 0) && (2 < 3));
+    assert.strictEqual(convergedOnBool(3), (3 > 0) && (3 < 3));
+    // jscs:enable disallowYodaConditions
   });
 });
--- a/src/both.js
+++ b/src/both.js
@@ -1,11 +1,12 @@
-var _curry2 = require('./internal/_curry2');
+var and = require('./and');
+var lift = require('./lift');
 
 
 /**
  * A function wrapping calls to the two functions in an `&&` operation, returning the result of the first
- * function if it is false-y and the result of the second function otherwise.  Note that this is
- * short-circuited, meaning that the second function will not be invoked if the first returns a false-y
- * value.
+ * function if it is false-y and the result of the second function otherwise.
+ *
+ * `R.both` will work on all other applicatives as well.
  *
  * @func
  * @memberOf R
@@ -23,8 +24,4 @@ var _curry2 = require('./internal/_curry2');
  *      f(100); //=> true
  *      f(101); //=> false
  */
-module.exports = _curry2(function both(f, g) {
-  return function _both() {
-    return f.apply(this, arguments) && g.apply(this, arguments);
-  };
-});
+module.exports = lift(and);
--- a/src/chain.js
+++ b/src/chain.js
@@ -23,6 +23,11 @@ var map = require('./map');
  *      var duplicate = n => [n, n];
  *      R.chain(duplicate, [1, 2, 3]); //=> [1, 1, 2, 2, 3, 3]
  */
-module.exports = _curry2(_dispatchable('chain', _xchain, function chain(fn, list) {
-  return _makeFlat(false)(map(fn, list));
+module.exports = _curry2(_dispatchable('chain', _xchain, function chain(fn, monad) {
+  if (typeof monad === 'function') {
+    return function() {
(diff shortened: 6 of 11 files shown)


davidchambers:
OMG, this is some incredible @joneshf​-level jutsu!

test/liftN.js
```diff
@@ -50,4 +44,25 @@ describe('liftN', function() {
...
+    assert.deepEqual(addN3([1, 2], [3, 4], [5, 6]), [9, 10, 10, 11, 10, 11, 11, 12]);
+  });
+
+  it('correctly interprets ((->) r) as a functor', function() {
+    var convergedOnInt = addN3(R.add(2), R.multiply(3), R.subtract(4));
+    var liftedAnd = R.liftN(2, R.and);
+    var convergedOnBool = liftedAnd(R.gt(R.__, 0), R.lt(R.__, 3));
+    assert.strictEqual(typeof convergedOnInt, 'function');
+    assert.strictEqual(typeof convergedOnBool, 'function');
+    assert.deepEqual(convergedOnInt(10), (10 + 2) + (10 * 3) + (4 - 10));
```
davidchambers: We can use `assert.strictEqual` here (and below).

test/liftN.js
```diff
@@ -50,4 +44,25 @@ describe('liftN', function() {
...
+
+  it('correctly interprets [] as a functor', function() {
+    assert.deepEqual(addN3([1, 2, 3], [1, 2], [1, 2, 3]), [3, 4, 5, 4, 5, 6, 4, 5, 6, 5, 6, 7, 5, 6, 7, 6, 7, 8]);
+    assert.deepEqual(addN3([1], [2], [3]), [6]);
+    assert.deepEqual(addN3([1, 2], [3, 4], [5, 6]), [9, 10, 10, 11, 10, 11, 11, 12]);
+  });
+
+  it('correctly interprets ((->) r) as a functor', function() {
+    var convergedOnInt = addN3(R.add(2), R.multiply(3), R.subtract(4));
+    var liftedAnd = R.liftN(2, R.and);
```
davidchambers: We could use `R.lift(R.and)` here, I believe.

src/internal/_map.js
```diff
@@ -1,8 +1,14 @@
-module.exports = function _map(fn, list) {
-  var idx = 0, len = list.length, result = Array(len);
-  while (idx < len) {
-    result[idx] = fn(list[idx]);
-    idx += 1;
+module.exports = function _map(fn, functor) {
+  if (functor instanceof Function) {
+    return function(x) {
+      return fn(functor(x));
+    };
```
davidchambers: This is just `compose`! :smile:

src/ap.js
```diff
@@ -19,8 +19,14 @@ var map = require('./map');
...
+  return _hasMethod('ap', functor) ?
+           functor.ap(fn) :
+         functor instanceof Function ?
+           function(x) {
+             return functor(x)(fn(x));
+           } :
+           _reduce(function(acc, f) {
+             return _concat(acc, map(f, fn));
+           }, [], functor);
 });
```
davidchambers: Perhaps this would read better formatted like so:

```javascript
module.exports = _curry2(function ap(ftor, fn) {
  return (
    _hasMethod('ap', ftor) ?
      ftor.ap(fn) :
    ftor instanceof Function ?
      function(x) { return ftor(x)(fn(x)); } :
    // else
      _reduce(function(acc, f) { return _concat(acc, map(f, fn)); }, [], ftor)
  );
});
```

Also, let's use `typeof ftor === 'function'` rather than `ftor instanceof Function` for the reason given in #1140.

test/liftN.js
```diff
@@ -50,4 +44,25 @@ describe('liftN', function() {
...
+  it('correctly interprets [] as a functor', function() {
+    assert.deepEqual(addN3([1, 2, 3], [1, 2], [1, 2, 3]), [3, 4, 5, 4, 5, 6, 4, 5, 6, 5, 6, 7, 5, 6, 7, 6, 7, 8]);
+    assert.deepEqual(addN3([1], [2], [3]), [6]);
+    assert.deepEqual(addN3([1, 2], [3, 4], [5, 6]), [9, 10, 10, 11, 10, 11, 11, 12]);
+  });
+
+  it('correctly interprets ((->) r) as a functor', function() {
+    var convergedOnInt = addN3(R.add(2), R.multiply(3), R.subtract(4));
+    var liftedAnd = R.lift(R.and);
+    var convergedOnBool = liftedAnd(R.gt(R.__, 0), R.lt(R.__, 3));
```
davidchambers: Let's remove the `liftedAnd` variable. `liftedAnd` doesn't convey any more information than `R.lift(R.and)`. ;)

CrossEye:
I love this.  I've always known that Haskell thinks of functions as functors over composition, although I didn't remember the applicative definition.

I hadn't ever _considered_ the possibility of including them here.  Nice done!  :fireworks: 

As to the two issues you raise,  I think we'll have to think pretty hard about both of them.  I'm quite loathe to go down the everything-is-unary route.  I don't know that we'd want to change the functions' APIs simply because we found an elegant new implementation that would demand it.

The `this` context has always been to me something very nice to have so long as it doesn't interfere too much with what else we want to do.  It shouldn't drive too much of our design.  But it _is_ nice to have, so as long as we can reasonably accommodate it, I feel we should.

But in any case, I would still want to extend `map` and `ap` this way; those are beautiful.

test/liftN.js
```diff
@@ -50,4 +44,25 @@ describe('liftN', function() {
     var addM = R.liftN(2, R.add);
     assert.deepEqual(addM(Maybe(3), Maybe(5)), Maybe(8));
   });
+
+  it('correctly interprets [] as a functor', function() {
+    assert.deepEqual(addN3([1, 2, 3], [1, 2], [1, 2, 3]), [3, 4, 5, 4, 5, 6, 4, 5, 6, 5, 6, 7, 5, 6, 7, 6, 7, 8]);
```
davidchambers: It'd be easier to understand how the output is derived given `[1, 2, 3]`, `[10, 20]`, and `[100, 200, 300]` as inputs.

test/liftN.js
```diff
@@ -50,4 +44,25 @@ describe('liftN', function() {
...
+    assert.deepEqual(addN3([1], [2], [3]), [6]);
+    assert.deepEqual(addN3([1, 2], [3, 4], [5, 6]), [9, 10, 10, 11, 10, 11, 11, 12]);
+  });
+
+  it('correctly interprets ((->) r) as a functor', function() {
```
davidchambers: What does `((->) r)` indicate?

benperez: It is the function type constructor with its first argument partially applied. Instances of Functor and Applicative must have type constructors that take exactly one argument, so we need to partially apply the first argument to `(->)`

test/liftN.js
```diff
@@ -50,4 +44,24 @@ describe('liftN', function() {
     var addM = R.liftN(2, R.add);
     assert.deepEqual(addM(Maybe(3), Maybe(5)), Maybe(8));
   });
+
+  it('correctly interprets [] as a functor', function() {
```
davidchambers: The word "correctly" is redundant.

test/liftN.js
davidchambers: I'd like to state this in a way that makes sense in the context of JavaScript. Would it be accurate to state that `liftN` _interprets `a -> b` as a functor_? It seems to me we're trying to say that unary functions are functors.

test/liftN.js
```diff
@@ -50,4 +44,24 @@ describe('liftN', function() {
...
+    assert.deepEqual(addN3([1, 2, 3], [10, 20], [100, 200, 300]), [111, 211, 311, 121, 221, 321, 112, 212, 312, 122, 222, 322, 113, 213, 313, 123, 223, 323]);
+    assert.deepEqual(addN3([1], [2], [3]), [6]);
+    assert.deepEqual(addN3([1, 2], [10, 20], [100, 200]), [111, 211, 121, 221, 112, 212, 122, 222]);
+  });
+
+  it('correctly interprets ((->) r) as a functor', function() {
+    var convergedOnInt = addN3(R.add(2), R.multiply(3), R.subtract(4));
+    var convergedOnBool = R.lift(R.and)(R.gt(R.__, 0), R.lt(R.__, 3));
```
davidchambers: I just realized that we're in __test/liftN.js__, so we should use `R.liftN` here after all. Sorry!

benperez:
@CrossEye I've updated this PR to do 2 things: 

1. addressed arity issues in `R.map` and `R.ap` for functions
2. switched the existing implementations of `R.complement`, `R.both`, and `R.either` to use `R.lift`.

Number 1 means that lifted functions can operate on functions of any length, which is consistent with the behavior of converged functions now and doesn't require Ramda to support "unary functions only".

Number 2 preserves the existing behavior of these functions with a cleaner, more generalized implementation __and__ has the additional benefit of generalizing them for all applicatives (in the case of `R.both` and `R.either`) and functors (in the case of `R.complement`).

I'm going to wait for something addressing https://github.com/ramda/ramda/issues/1401 to land before tackling the issue of context in lifted functions, especially with regards to `R.converge`.

davidchambers:
> switched the existing implementations of `R.complement`, `R.both`, and `R.either` to use `R.lift`

These definitions are absolutely beautiful! :heart:

test/liftN.js
benperez: My understanding is that the return type of the function is the important part.
```javascript
var both = R.liftN(2, R.and);
// both :: Applicative f => f Bool -> f Bool -> f Bool
```
When we're talking about functions as _Applicative Functors_ we interpret 
`f Bool` to mean `((->) r) Bool` or `r -> Bool` 
so one interpretation of the type of the type of `both` is
```haskell
both :: (r -> Bool) -> (r -> Bool) -> (r -> Bool)
```
> Would it be accurate to state that liftN interprets a -> b as a functor? It seems to me we're trying to say that unary functions are functors.

In haskell I'm pretty sure that only unary functions can be applicative functors. In Javascript that's not the case, since we can hack around the arguments object to accept whatever number of arguments we want. We could indicate this by changing the description to something like:
> correctly interprets (* ->) as a functor

davidchambers: I'll let you decide what makes the most sense. If that's sticking with `((->) r)`, that's fine with me.

CrossEye:
Fantastic!

Any objections, anyone?

kedashoe:
Amazing! Still trying to understand exactly what is going on, but a few comments:

- I believe the descriptions for `both` and `either` need updating as they do not short circuit anymore
- Should the descriptions for `map` and `ap` also be updated to mention this new behavior when they get functions?
- In `ap` we have the variable `ftor`, in `map` it is called `functor`
- Should we not include some tests in `test/map.js` and `test/ap.js`? Possibly (probably?) anything that breaks this behavior would be caught by the tests in `liftN`, but maybe appropriate to include tests closer to the source?

Great stuff!

scott-christopher:
This looks good to me.

Is wanting a `chain` implementation for functions going too far? :smile: 
```js
module.exports = _curry2(_dispatchable('chain', _xchain, function chain(fn, list) {
  if (typeof list == 'function') {
    return function(env) {
      return fn(list(env))(env);
    };
  }
  return _makeFlat(false)(map(fn, list));
}));
```

I would also be interested in trying to consolidate the dispatching, but don't really want to hijack this PR in the process.

davidchambers:
Great points, @kedashoe.

> Is wanting a `chain` implementation for functions going too far?

We can do that too? Cool! What do you say, @benperez?

benperez:
@kedashoe Thanks, I'll update the descriptions for `both` &  `either` &  `map` & `ap` and update the naming inconsistency. I'll also add tests in `map` and `ap`.

@scott-christopher Very nice! Will add that in tomorrow morning!

CrossEye:
> I believe the descriptions for both and either need updating as they do not short circuit anymore

That's actually a bit of a bummer.

It was one of the nice things about these functions.  But perhaps it wasn't really a particularly functional concern, just a standard JS one.

src/ap.js
```diff
@@ -19,8 +21,15 @@ var map = require('./map');
...
-  return _hasMethod('ap', fns) ? fns.ap(vs) : _reduce(function(acc, fn) {
-    return _concat(acc, map(fn, vs));
-  }, [], fns);
+module.exports = _curry2(function ap(ftor, fn) {
+  return (
+    _hasMethod('ap', ftor) ?
+      ftor.ap(fn) :
+    typeof ftor === 'function' ?
+      curryN(Math.max(ftor.length, fn.length), function() {
+        return ftor.apply(this, _slice(arguments))(fn.apply(this, _slice(arguments)));
```
buzzdecafe: no need to slice twice

paldepind:
I think this is a good idea. But if we take this step we should go further still.

Objects for instance are functors.

Many other instances exists for native JavaScript data structures. I implemented a good bunch of them a while back in [seamless-fantasy](https://github.com/paldepind/seamless-fantasy).

src/ap.js
davidchambers: No need to slice *at all*. :)

davidchambers:
> I think this is a good idea. But if we take this step we should go further still.
>
> Objects for instance are functors.

I agree. Could you open an issue for discussion?

src/ap.js
buzzdecafe: indeed, even better

benperez:
I've updated this PR to address all comments.

buzzdecafe:
:cow2:

src/complement.js
```diff
@@ -11,6 +11,8 @@ var _curry1 = require('./internal/_curry1');
  *   - applying `g` to zero or more arguments will give __false__ if applying
  *     the same arguments to `f` gives a logical __true__ value.
  *
+ * `R.not` will work on all other functors as well.
```
davidchambers: Should this be in the docstring for `R.not` instead?

benperez: Good catch, should have said `R.complement`. :zap:

CrossEye:
>> Objects for instance are functors.

> I agree. Could you open an issue for discussion?

Yes, definitely worth discussing, but also should be a separate issue.

davidchambers:
:deciduous_tree:

davidchambers:
> Objects for instance are functors.

#1426

mattS-H:
The loss of short-circuiting is a real issue. I don't know why this has been ignored. For me its a performance problem primarily.

svozza:
https://github.com/ramda/ramda/pull/1475

PR #1409, opened 2015-09-28
fix R.commute and R.commuteMap

davidchambers:
Before:

```javascript
> R.commute(S.Either.of, [S.Left('XXX'), S.Left('YYY'), S.Left('ZZZ')])
Left("ZZZ")
```

After:

```javascript
> R.commute(S.Either.of, [S.Left('XXX'), S.Left('YYY'), S.Left('ZZZ')])
Left("XXX")
```

It seems to me the leftmost failure should be returned, as is the case with Haskell's [`sequence`][1]:

```haskell
> sequence [Right 1, Right 2]
Right [1,2]
> sequence [Right 1, Left "XXX"]
Left "XXX"
> sequence [Left "XXX", Right 1]
Left "XXX"
> sequence [Left "XXX", Left "YYY"]
Left "XXX"
```

[1]: http://hackage.haskell.org/package/base-4.8.1.0/docs/Prelude.html#v:sequence

### DIFF
--- a/test/commute.js
+++ b/test/commute.js
@@ -1,5 +1,6 @@
-var Maybe = require('./shared/Maybe');
+var S = require('sanctuary');
 
+var Maybe = require('./shared/Maybe');
 var R = require('..');
 var eq = require('./shared/eq');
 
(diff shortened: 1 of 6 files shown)


CrossEye:
:herb:

test/commute.js
```diff
@@ -1,9 +1,16 @@
...
 
+var S = require('sanctuary');
+
+var Maybe = require('./shared/Maybe');
 var R = require('..');
 
 
+var eq = function(actual, expected) {
+  assert.strictEqual(R.toString(actual), R.toString(expected));
+};
```
davidchambers: @CrossEye, would you be happy to use this function in place of `assert.strictEqual` and `assert.deepEqual` throughout the test suite (except for testing `R.toString` itself)? If so, I submit a pull request to add __test/shared/eq.js__.

`assert.deepEqual` is horribly broken: `[]` and `{}` are considered equal, as are `[1, 2, 3]` and `['1', '2', '3']`. It's no good for comparing values of ADTs, either, since `Left(42)` and `Right(42)` may have the same internals. In such cases one is forced to use `R.equals` with `assert.strictEqual`, but then in the case of failure one sees

    AssertionError: false === true

rather than a descriptive message such as

    AssertionError: 'Left(42)' === 'Right(42)'

CrossEye: I certainly have no objection.

I wish I had started this project with different testing tools altogether.  But I've never had the heart to try to go back and retrofit them.  But this would certainly be a help.

davidchambers: #1415

PR #1412, opened 2015-09-28
docs/useWith: update example

raine:


src/useWith.js
```diff
@@ -25,9 +25,7 @@ var curry = require('./curry');
  *
  *      var double = y => y * 2;
  *      var square = x => x * x;
- *      var add = (a, b) => a + b;
- *      // Adds any number of arguments together
- *      var addAll = () => R.reduce(add, 0, arguments);
```
raine: `arguments` is not available with arrow syntax.

asaf-romano:
(...args) is

davidchambers:
Thanks, @raine. These examples will actually be removed as part of #1397, which is very close to being merged. I'd like to close this pull request to avoid introducing merge conflicts unnecessarily.

raine:
No problem, missed that one.

PR #1423, opened 2015-10-04
Add R.without

raine:
As I proposed on gitter,

```js
without :: a -> [a] -> [a]
```

### DIFF
--- a/src/without.js
+++ b/src/without.js
@@ -1,0 +1,27 @@
+var _contains = require('./internal/_contains');
+var _curry2 = require('./internal/_curry2');
+var flip = require('./flip');
+var reject = require('./reject');
+
+
+/**
+ * Returns a new list without values in the first argument.
+ * `R.equals` is used to determine equality.
+ *
+ * Acts as a transducer if a transformer is given in list position.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig [a] -> [a] -> [a]
+ * @param {Array} list1 The values to be removed from `list2`.
+ * @param {Array} list2 The array to remove values from.
+ * @return {Array} The new array without values in `list1`.
+ * @see R.transduce
+ * @example
+ *
+ *      R.without([1, 2], [1, 2, 1, 3, 4]); //=> [3, 4]
+ */
+module.exports = _curry2(function(xs, list) {
+  return reject(flip(_contains)(xs), list);
+});
--- a/test/without.js
+++ b/test/without.js
@@ -1,0 +1,31 @@
+var R = require('..');
+var eq = require('./shared/eq');
+
+
+describe('without', function() {
+  it('returns an array not containing values in the first argument', function() {
+    eq(R.without([1, 2], [1, 2, 1, 4, 5]), [4, 5]);
+  });
+
+  it('is curried', function() {
+    var withoutOnes = R.without([1]);
+    eq(withoutOnes([1, 2, 3, 5, 1]), [2, 3, 5]);
+  });
+
+  it('can act as a transducer', function() {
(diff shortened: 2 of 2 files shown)


asaf-romano:
I'd personally prefer a version taking a list as the first argument, similar to ``omit``.

davidchambers:
> I'd personally prefer a version taking a list as the first argument, similar to `omit`.

I had the same thought, then realized this would be [`R.difference`][1] with arguments flipped. @raine, is `without` significantly more convenient than `difference` in the situations in which you've used it?


[1]: http://ramdajs.com/docs/#difference

raine:
`R.difference` returns a set, i.e. removes duplicate values

davidchambers:
Good point!

CrossEye:
:herb:

asaf-romano:
Just this morning I used R.difference as a substitute for R.without (the list variant), and as always got confused with the arguments order. I guess that's because``R.difference`` is a binary function in which both arguments are considered "data" and have the same shape, albeit the order matters.  What I really want to do is to express this idea in a pipeline:

``R.without(["_id", "__version__"]) // filter db stuff``

davidchambers:
I don't have a view as to whether the first argument should be of type `a` or `[a]`. If you believe `a` to be the better option, @raine, I'm happy for you to merge this pull request.

raine:
I updated this to take a list as first argument. Please have another look.

src/without.js
```diff
@@ -0,0 +1,25 @@
...
+ * @category List
+ * @sig [a] -> [a] -> [a]
+ * @param {Array} list1 The values to be removed from `list2`.
+ * @param {Array} list2 The array to remove values from.
+ * @return {Array} The new array without values in `list1`.
+ * @example
+ *
+ *      R.without([1, 2], [1, 2, 1, 3, 4]); //=> [3, 4]
+ */
+module.exports = useWith(reject, [flip(_contains), identity]);
```
davidchambers: I'd prefer not to use `useWith` internally. A lambda would reduce the number of Ramda functions involved in this definition, reducing the footprint of `require('ramda/src/without')` when using Browserify or similar.

src/without.js
```diff
@@ -0,0 +1,25 @@
+var _contains = require('./internal/_contains');
+var flip = require('./flip');
+var identity = require('./identity');
+var reject = require('./reject');
+var useWith = require('./useWith');
+
+/**
```
davidchambers: Nit: Two empty lines after imports, please. :)

test/without.js
```diff
@@ -0,0 +1,31 @@
...
+    };
+
+    eq(R.without([0], [-0]).length, 1);
+    eq(R.without([-0], [0]).length, 1);
+    eq(R.without([NaN], [NaN]).length, 0);
+    eq(R.without([[1]], [[1]]).length, 0);
+    eq(R.without([new Just([42])], [new Just([42])]).length, 0);
+  });
+});
+
```
davidchambers: :space_invader:

src/without.js
CrossEye: That's interesting.  While you're probably right, I originally created `useWith` in order to elegantly write `project`.

src/without.js
```diff
@@ -0,0 +1,25 @@
...
+ * Returns a new list without values in the first argument.
+ * `R.equals` is used to determine equality.
+ *
+ * Acts as a transducer if a transformer is given in list position.
+ * @see R.transduce
```
davidchambers: Please move this line to between the `@return` and `@example` annotations (see #1458).

CrossEye:
This has been sitting a while.  Any objections?

davidchambers:
:deciduous_tree:

Issue #1366, opened 2015-09-02
How to apply exact arguments for variadic function?


anton-rudeshko:
I need to call variadic or curried functions with all arguments provided *before* call.

```js
var log = console.log.bind(console);
var promise = Q('promised');

var applyExactArgs = compose(R.nAry(0), R.partial);

promise.then(R.add(1, 2)); // this is calling result function right away
promise.then(R.partial(R.add, 1, 2)); // this is populating all args, but not calling

// usual R.partial
promise.then(R.partial(log, 'Hello', 'World')); // → outputs "Hello World promised"

// What I need
promise.then(applyExactArgs(log, 'Hello', 'World')) // → outputs "Hello World", ignores all additional not pre-applied args.
```

Is there some function for this already? Or better solution?

raine:
If you happen to have access to prettier lambda syntax, in this case it's quite readable option.

anton-rudeshko:
Even better:

```js
var min = magicFunction(2, Math.min);
// min — curried, has 2 real args, but must be called one more time to get result
var min1 = min(1);
var min1and2 = min1(2);
min1and2(0, -1); // 1, all additional args are ignored
```

How `magicFunction` should be defined?

> If you happen to have access to prettier lambda syntax, in this case it's quite readable option.

What do you mean?

raine:
Is [nAry](http://ramdajs.com/docs/#nAry) what you're looking for?

I meant `promise.then(() => console.log('Hello world'))`

anton-rudeshko:
Yeah, lambda looks pretty good.
nAry limits count of args, but i need to have result of it as callback. I need to have it called *one more time* to have a result.

CrossEye:
Ramda does not have a function for this.  But I'm certain of the name you should use if you create one: [`thunk`][th].  I don't have a particular implementation in mind, though.


  [th]: https://en.wikipedia.org/wiki/Thunk#Functional_programming

davidchambers:
It sounds as though a pull request for `R.thunk` may be accepted.

PR #1424, opened 2015-10-04
Add R.thunk

raine:
Discussed in #1366

What is the correct signature? The current one is same as `R.partial`'s.

### DIFF
--- a/src/thunk.js
+++ b/src/thunk.js
@@ -1,0 +1,31 @@
+var _curry1 = require('./internal/_curry1');
+var _slice = require('./internal/_slice');
+
+/**
+ * Accepts as its arguments a function and any number of values and returns a function that,
+ * when invoked, calls the original function with only the values originally provided.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig (a -> b -> ... -> i -> j) -> a -> b -> ... -> i -> (() -> j)
+ * @param {Function} fn The function to invoke.
+ * @param {...*} [args] Arguments to invoke `fn` with.
+ * @return {Function} A new function wrapping `fn`. When invoked, it will call `fn`
+ *         with `args`, while disregarding any new arguments.
+ * @see R.partial
+ * @example
+ *
+ *      var add = (a, b) => a + b;
+ *      var sum = (...args) => args.reduce(add, 0);
+ *
+ *      var sumThunk = R.thunk(sum, 1, 2)
+ *      sumThunk();  //=> 3
+ *      sumThunk(5); //=> 3 // 5 is disregarded
+ */
+module.exports = _curry1(function(fn) {
+  var args = _slice(arguments, 1);
+  return function() {
+    return fn.apply(this, args);
+  };
+});
--- a/test/thunk.js
+++ b/test/thunk.js
@@ -1,0 +1,18 @@
+var R = require('..');
+var eq = require('./shared/eq');
+
+describe('thunk', function() {
+  it('applies function only to arguments given initially', function() {
+    var add = function(a, b) { return a + b; };
+    var sum = function() {
+      return [].reduce.call(arguments, add, 0);
+    };
+
+    eq(R.thunk(sum, 1, 1)(5), 2);
+  });
+
+  it('reports the arity of the new function as zero', function() {
+    var f = R.thunk(R.add(1));
+    eq(f.length, 0);
+  });
+});


src/thunk.js
```diff
@@ -0,0 +1,26 @@
...
+ * @return {Function} A new function wrapping `fn`. When invoked, it will call `fn`
+ *         with `args`, while disregarding any new arguments.
+ * @see R.partial
+ * @example
+ *
+ *      var add = (a, b) => a + b;
+ *      var sum = (...args) => args.reduce(add, 0);
+ *
+ *      var sumThunk = R.thunk(sum, 1, 2)
+ *      sumThunk(2); //=> 3 // 2 is disregarded
```
CrossEye: sumThunk(5); //=> 3 // 5 is disregarded

so as not to confuse the two 2's.

test/thunk.js
```diff
@@ -0,0 +1,18 @@
...
+var eq = require('./shared/eq');
+
+describe.only('thunk', function() {
+  var add = function(a, b) { return a + b; };
+  var sum = function() {
+    return [].reduce.call(arguments, add, 0);
+  };
+
+  it('applies function only to arguments given initially', function() {
+    eq(R.thunk(sum, 1, 1)(2), 2);
```
CrossEye: ditto

CrossEye:
:+1:

src/thunk.js
```diff
@@ -0,0 +1,26 @@
...
+var partial = require('./partial');
+
+/**
+ * Accepts as its arguments a function and any number of values and returns a function that,
+ * when invoked, calls the original function with only the values originally provided.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig (a -> b -> ... -> i -> j -> ... -> m -> n) -> a -> b -> ... -> i -> (j -> ... -> m -> n)
```
scott-christopher: I think the signature should show that this returns a nullary function.
e.g.
```
(a -> b -> ... -> i -> j) -> a -> b -> ... -> i -> (() -> j)
```

src/thunk.js
```diff
@@ -0,0 +1,26 @@
...
+ * @return {Function} A new function wrapping `fn`. When invoked, it will call `fn`
+ *         with `args`, while disregarding any new arguments.
+ * @see R.partial
+ * @example
+ *
+ *      var add = (a, b) => a + b;
+ *      var sum = (...args) => args.reduce(add, 0);
+ *
+ *      var sumThunk = R.thunk(sum, 1, 2)
+ *      sumThunk(5); //=> 3 // 5 is disregarded
```
scott-christopher: While it is good to demonstrate extra arguments are ignored, I think this example should also include the more common use case of evaluating the thunk with no args.

Perhaps including two examples of the same thunk being called could be more informative.
```js
var sumThunk = R.thunk(sum, 1, 2)
sumThunk();  //=> 3
sumThunk(5); //=> 3 // 5 is disregarded
```

raine:
Good points, @scott-christopher and @CrossEye

raine:
The `compose(nAry(0), partial);` implementation wouldn't satisfy the invariants test so I changed the implementation.

```
invariants
  1) -- applying function f with length n (where n > 0) to no arguments gives function with length n
  2) -- applying function f with length n (where n > 0) to R.__ gives function with length n
  ✓ -- applying function f with length n (where n > 1) to any value other than R.__ gives function with length n - 1
```

Before:

```
> var thunk = require('./src/thunk')
> thunk.length
1
> thunk().length
0
```

After:

```
> var thunk = require('./src/thunk')
> thunk.length
1
> thunk().length
1
```

CrossEye:
LGTM

davidchambers:
Let's consider the sequence of steps one must take in order to arrive at `R.thunk`:

1.  One wishes to apply `f` to zero or more arguments. Let's arbitrarily choose two as the number of arguments, to provide an example. Let's name those arguments `a` and `b`. So far so good. Applying functions to arguments is central to functional programming.

2.  There's a catch: `f` has side effects. We wish to avoid triggering the side effects right away, so we can't apply `f` immediately. This raises a red flag, but let's ignore it for now.

3.  There's a function for this: [`R.partial`][1]. `R.partial(f, a, b)` creates a function which, when invoked, will apply `f` to `a` and `b`. We're done!

4.  There's another catch: the function returned by `R.partial(f, a, b)` accepts any number of arguments to which `f` will *also* be applied. For some reason we're providing arguments to our "thunk", and we wish to ignore these arguments. Let's ignore *this* for now as well.

5.  The additional arguments to `f` would not be a problem if `f` were actually a binary function, but there's another catch: `f` is a variadic function!

To summarize, __when working with variadic functions with uncontrolled side effects, we run into problems if we provide more than *n* arguments to an *n*-ary function__. Rare circumstances indeed!

Not only are the circumstances rare, but they only arise if we do naughty things. First, functions should not have *uncontrolled* side effects. Perhaps `f` reads a file from disk, and we don't want to trigger this side effect until later. The correct solution is to use a Task (or similar type, but certainly *not* a Promise) to represent the delayed computation. Secondly, we should never apply an *n*-ary function to more than *n* arguments. I realize this is common practice in some circles, but it's mathematically nonsensical.

Let's not include a function which is only useful when one is naughty. Doing so would implicitly condone the behaviour. In this case, `R.partial(R.binary(f), a, b)` or `R.nAry(0, R.partial(f, a, b))` would be fine. They both look weird, which is a *good* thing: the weirdness may be a :warning: for readers.

:thumbsdown:


[1]: http://ramdajs.com/0.17/docs/#partial

scott-christopher:
This doesn't necessarily need side effects to be useful. Consider a lazy right fold where you don't wish to evaluate the rest of the accumulating value if it is not needed.
```js
function foldRThunk(fn, acc, list) {
  return R.isEmpty(list) ? acc : fn(list[0], thunk(foldRThunk, fn, acc, R.tail(list)));
}
exists = (p, list) => foldRThunk((a, b) => p(a) || b(), false, list);
```
Having said that, I wasn't aware that `partial` already accepts the entire argument list to return a nullary function. I feel that renders the need for a specific `thunk` function a bit moot.

davidchambers:
Thanks for mentioning laziness, @scott-christopher. I had not considered that use case. I agree, though, that `partial` and `() =>` are acceptable solutions.

CrossEye:
It's not just a matter of the sort of examples described earlier, or of lazy data structures either.  There's simply the notion of computationally expensive functions many of which might be prepped ahead of time, but only one of which will end up running.  That, I suppose is another form of laziness, but is somewhat different from the lazy fold example.

This, of course, can be handled by fat arrows or by `partial`, but here is where I think I come down in favor of syntactic sugar when @davidchambers might not:  Even if a thunk can be created by partial application, it's the `thunk` abstraction I'm thinking of, not the `partial application` one.  A `thunk` is a delayed computation.  A `partial application` is the early application of a subset of the parameters of a function.  The fact that one can create a thunk by making that subset the full  complement of parameters does not make a thunk a partial application; it simply means that there is a way to implement it as such.  While it may sound like splitting hairs to some, it's a very real distinction to me.  It's the same reason that I think `head` should not be deprecated in favor of `nth(0)`.  They are separate abstractions which merely happen to coincide.

scott-christopher:
My point was very much inline with using thunks for lazy evaluation (the lazy fold was just an example that I had within arms reach). I agree that `thunk` provides a clearer intent to the reader than `partial`, but I'm personally happy with using fat arrow syntax so I'm impartial as to whether we offer this or not.

davidchambers:
While my preference is not to include this function, I don't mind if we do so. I leave it to you, @CrossEye.

CrossEye:
@buzzdecafe?  @paldepind?  @TheLudd?  Any preferences?

I also don't have particularly strong feelings.

TheLudd:
> I also don't have particularly strong feelings.

Me neither

buzzdecafe:
abstain

paldepind:
> I agree, though, that partial and () => are acceptable solutions.

I agree. `partial` and `() =>` are common and well understood.

raine:
`R.thunk` doesn't seem like a compelling enough addition to Ramda so closing this.

CrossEye:
**ker-Thunk**

PR #1426, opened 2015-10-06
define R.map for Object

davidchambers:
This one's for you, @paldepind! :smile:

Before anyone freaks out, thinking that Ramda is now making functions do all the things à la Underscore, let me remind you that it's possible to arrive at the same conclusion for two very different reasons. This is not reckless overloading: it makes sense, as I hope to demonstrate.

  - In the early days very few Ramda functions were polymorphic. I imagine this was a reaction to the many useful but unprincipled polymorphic functions in Underscore.

  - Over time various Ramda functions have been granted the ability to dispatch to appropriately named methods, allowing Ramda's list functions to be applied to values which are not necessarily array-like.

  - In recent times Haskell and Fantasy Land have influenced the way certain Ramda contributors view polymorphic functions. Rather than view polymorphic functions as complected, this contingent sees them as useful higher-level abstractions over *groups* of types (*type classes* in Haskell terminology), provided they are principled (i.e. satisfy laws).

  - Thinking in terms of groups of types encouraged us to define `R.empty` for Object and String in #1236, and then to define `R.isEmpty` in terms of `R.empty` in #1242.

  - This same thought process led to define `R.map` for Function in #1408.

`R.map` is no longer a function which operates only on array-like objects. It should operate on any functor. User-defined types such as Maybe can simply provide a `map` method, but Ramda must provide the `map` definition for each built-in type which is a functor: Array, Function, and Object.

### DIFF
--- a/src/map.js
+++ b/src/map.js
@@ -1,7 +1,10 @@
 var _curry2 = require('./internal/_curry2');
 var _dispatchable = require('./internal/_dispatchable');
 var _map = require('./internal/_map');
+var _reduce = require('./internal/_reduce');
 var _xmap = require('./internal/_xmap');
+var curryN = require('./curryN');
+var keys = require('./keys');
 
 
 /**
@@ -20,7 +23,7 @@ var _xmap = require('./internal/_xmap');
  * @func
  * @memberOf R
  * @category List
- * @sig (a -> b) -> [a] -> [b]
+ * @sig Functor f => (a -> b) -> f a -> f b
  * @param {Function} fn The function to be called on every element of the input `list`.
  * @param {Array} list The list to be iterated over.
  * @return {Array} The new list.
@@ -29,5 +32,21 @@ var _xmap = require('./internal/_xmap');
  *      var double = x => x * 2;
  *
  *      R.map(double, [1, 2, 3]); //=> [2, 4, 6]
+ *
+ *      R.map(double, {x: 1, y: 2, z: 3}); //=> {x: 2, y: 4, z: 6}
  */
-module.exports = _curry2(_dispatchable('map', _xmap, _map));
+module.exports = _curry2(_dispatchable('map', _xmap, function map(fn, functor) {
+  switch (Object.prototype.toString.call(functor)) {
+    case '[object Function]':
+      return curryN(functor.length, function() {
+        return fn.call(this, functor.apply(this, arguments));
+      });
+    case '[object Object]':
+      return _reduce(function(acc, key) {
+        acc[key] = fn(functor[key]);
+        return acc;
+      }, {}, keys(functor));
+    default:
+      return _map(fn, functor);
+  }
+}));
--- a/src/internal/_map.js
+++ b/src/internal/_map.js
@@ -1,16 +1,10 @@
-var curryN = require('../curryN');
-
 module.exports = function _map(fn, functor) {
-  if (typeof functor === 'function') {
-    return curryN(functor.length, function() {
-      return fn.call(this, functor.apply(this, arguments));
-    });
-  } else {
-    var idx = 0, len = functor.length, result = Array(len);
-    while (idx < len) {
-      result[idx] = fn(functor[idx]);
-      idx += 1;
-    }
-    return result;
+  var idx = 0;
+  var len = functor.length;
+  var result = Array(len);
+  while (idx < len) {
+    result[idx] = fn(functor[idx]);
+    idx += 1;
   }
+  return result;
 };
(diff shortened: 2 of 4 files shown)


CrossEye:
> Before anyone freaks out

You looking at me?  :smile: I'm not freaking out at all.  Now if you start talking `reduceObject`...

I missed @paldepind's [seamless-fantasy][sf] when it first came out (vacation?) but was impressed when I saw it last week and immediately started thinking along these same lines.

It would be useful to deprecate `mapObj` in favor of a lawful `map` on objects.  I certainly have no objections whatsoever.  This is not a matter of complecting, as you say, but of finding a higher abstraction.

:100:


  [sf]: https://github.com/paldepind/seamless-fantasy

paldepind:
I like this :+1: 

Lawful polymorphism is very useful. And I think this is a very sensible place to start.

I especially like how this _both_ makes Ramda more for JavaScript developers and more functional.

buzzdecafe:
:cow2:

svozza:
Does this mean `mapObj` can be deprecated?

svozza:
Ooops, just saw the deprecation notice in `mapObj.js`.

src/map.js
```diff
@@ -29,5 +32,21 @@ var _xmap = require('./internal/_xmap');
...
+        return fn.call(this, functor.apply(this, arguments));
+      });
+    case '[object Object]':
+      return _reduce(function(acc, key) {
+        acc[key] = fn(functor[key]);
+        return acc;
+      }, {}, keys(functor));
+    default:
+      return _map(fn, functor);
+  }
```
jdalton: Any plans to tackle array-like objects?
```js
var children = document.body.children;

Object.keys(children);
// => ['0', '1', 'someName', 'anotherName']
children.length;
// => 2;
```

davidchambers: Not currently. I'm fine with treating `children` as a list of length two (i.e. ignoring the non-"index" property names).

kaw2k:
Will this also impact `mapObjIndexed`?

cscalfani:
Can you please provide an example for using `map` like `mapObj`?

davidchambers:
Here you go, @cscalfani:

```javascript
> R.map(R.negate, {x: 1, y: 2, z: 3})
{x: -1, y: -2, z: -3}
```

cscalfani:
Thanks!! This would be a great addition to the docs. Especially helpful at dispelling any panic over the deprecation of `mapObj`.

davidchambers:
Is the example at <http://ramdajs.com/docs/#map> adequate?

cscalfani:
For `mapObj` it's great. Thanks.

An example on how to create your own Functor and use it with map would be really nice. Or at least a link to a location in the docs that explains this.

Thanks again.

CrossEye:
I think we still need a Types section in the documentation.  I haven't really thought yet about how to go about this.  For the FantasyLand types, these might have some reasonably short description and a link to the FL specification.  For others such as Lenses, I think we need something more.  I would love it if this could also somehow describe the notion of arrays-as-lists that Ramda uses.

If anyone has good ideas for this, I'm all ears.

raine:
It seems like the `src/map.js` documentation should be updated to mention objects.

gilligan:
The category is set to List - i suppose a function can only belong to one category? Maybe there should be a Functor category?

davidchambers:
> Maybe there should be a Functor category?

I think so. Certain functions should live in *multiple* categories.

CrossEye:
> Certain functions should live in *multiple* categories.

Quite a few of them, in fact.  I'm not sure if our documentation template supports that well, but we should get there soon.

netei:
Same question as : @kaw2k 

> Will this also impact mapObjIndexed?

davidchambers:
> Will this also impact `mapObjIndexed`?

No. That function is unaffected.

Xananax:
I know this is very old, and I'm not sure it's the right place to comment, but I just went out a full 20mn feverish halftrip around the internet to finally find this thread. The docs just say that objMap is "deprecated", and there's nothing like "see `map`" or equivalent statement that would tell you what to use
Even if you have the idea, like I did, to check `map`, you'll find it operates on `lists`, as clearly stated by both the category and the first sentence: "Returns a new list, constructed by applying the supplied function to every element of the supplied list." Reading the whole thing doesn't help, as not a single time in the whole paragraph are objects mentioned. I didn't notice the example at all.

I suggest a mention to `map` in the description of `objMap`, as well as changing the first sentence in the description of `map`. Thanks for a great library!

neverfox:
Wait, so what is the new canonical way to iterate over key/value pairs? I tried R.map but only the value is passed to the callback, not the key in second position, as in mapObj.

davidchambers:
`R.mapObjIndexed` isn't the nicest name, but `R.mapIndexed` would be even more confusing, I think.

PR #1427, opened 2015-10-06
do not rely on .constructor.name in R.toString

### LINKED ISSUE #1393: do not rely on .constructor.name in R.toString
As mentioned in https://github.com/ramda/ramda/issues/1140#issuecomment-119953239, the implementation of `_toString` currently contains `x.constructor.name !== 'Object'`. Apparently not every environment provides a `name` property on each function. Instead we could use a regular expression such as `^function (\w*)` to extract a function's name from its string representation.

davidchambers:
Closes #1393

### DIFF
--- a/src/internal/_functionName.js
+++ b/src/internal/_functionName.js
@@ -1,0 +1,5 @@
+module.exports = function _functionName(f) {
+  // String(x => x) evaluates to "x => x", so the pattern may not match.
+  var match = String(f).match(/^function (\w*)/);
+  return match == null ? '' : match[1];
+};
--- a/src/internal/_toString.js
+++ b/src/internal/_toString.js
@@ -1,4 +1,5 @@
 var _contains = require('./_contains');
+var _functionName = require('./_functionName');
 var _map = require('./_map');
 var _quote = require('./_quote');
 var _toISOString = require('./_toISOString');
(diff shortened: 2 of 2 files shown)


src/internal/_functionName.js
```diff
@@ -0,0 +1,5 @@
+module.exports = function _functionName(f) {
+  // String(x => x) evaluates to "x => x", so the pattern may not match.
+  var match = String(f).match(/^function (\w*)/);
+  return match == null ? '' : match[1];
```
buzzdecafe: could you use `constructor.name` _if present_ and otherwise evaluate the stringified function?

davidchambers: Are you concerned about the performance implications?

buzzdecafe: not especially. but if we have it, why not use it?

davidchambers: Given a choice between one code path and two, I'll choose one unless there's a compelling reason not to do so. Were we to have two code paths—one just for old versions of IE—we'd actually need to run the tests in those old browsers. With one code path, running the tests in Node is sufficient.

buzzdecafe: in that case you have dragged all platforms down to IE's level. that hardly seems the preferable choice

davidchambers: > in that case you have dragged all platforms down to IE's level.

It sounds as though you *are* concerned about the performance implications. ;)

[MDN][1] states the following:

> This is a new technology, part of the ECMAScript 2015 (ES6) standard.

It seems the approach above is the "correct" approach, though it's undoubtedly ugly.


[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name

davidchambers:
Is this approach reasonable, or do others share @buzzdecafe's concern?

PR #1429, opened 2015-10-06
define R.filter for Object

davidchambers:
This change is warranted by many of the arguments made in #1426.

We have the elegant `R.map`, which applies the given function to exactly one argument, and the inelegant `R.mapObjIndexed` which applies the given function to `(value, key, object)`. [`R.pickBy`][1] falls between these two stools, applying the given function to *two* arguments.

This pull request:

  - <del>deprecates `R.pickBy`;</del>
  - defines `R.filter` for Object<del>; and</del>
  - <del>adds `R.filterObjIndexed`, which mirrors `R.mapObjIndexed`, to replace `R.pickBy` in contexts in which the key is relevant.</del>

I first suggested this change in https://github.com/ramda/ramda/pull/1265#issuecomment-120538941:

> I think `R.pickBy` and `R.omitBy` should only pass the value to the predicate.
>
> I would then go a step further and support objects in `R.filter` (and by extension `R.reject`). We could then deprecate `R.pickBy` and `R.omitBy`.

The time is now right. :)

[1]: http://ramdajs.com/0.17/docs/#pickBy

### DIFF
--- a/src/filter.js
+++ b/src/filter.js
@@ -1,17 +1,16 @@
 var _curry2 = require('./internal/_curry2');
 var _dispatchable = require('./internal/_dispatchable');
 var _filter = require('./internal/_filter');
+var _isObject = require('./internal/_isObject');
+var _reduce = require('./internal/_reduce');
 var _xfilter = require('./internal/_xfilter');
+var keys = require('./keys');
 
 
 /**
- * Returns a new list containing only those items that match a given predicate
- * function. The predicate function is passed one argument: *(value)*.
- *
- * Note that `R.filter` does not skip deleted or unassigned indices, unlike the
- * native `Array.prototype.filter` method. For more details on this behavior,
- * see:
- * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter#Description
+ * Takes a predicate and a "filterable", and returns a new filterable of the
+ * same type containing the members of the given filterable which satisfy the
+ * given predicate.
  *
  * Dispatches to the `filter` method of the second argument, if present.
  *
@@ -21,15 +20,29 @@ var _xfilter = require('./internal/_xfilter');
  * @memberOf R
  * @since v0.1.0
  * @category List
- * @sig (a -> Boolean) -> [a] -> [a]
- * @param {Function} fn The function called per iteration.
- * @param {Array} list The collection to iterate over.
- * @return {Array} The new filtered array.
+ * @sig Filterable f => (a -> Boolean) -> f a -> f a
+ * @param {Function} pred
+ * @param {Array} filterable
+ * @return {Array}
  * @see R.reject, R.transduce, R.addIndex
  * @example
  *
  *      var isEven = n => n % 2 === 0;
  *
  *      R.filter(isEven, [1, 2, 3, 4]); //=> [2, 4]
+ *
+ *      R.filter(isEven, {a: 1, b: 2, c: 3, d: 4}); //=> {b: 2, d: 4}
  */
-module.exports = _curry2(_dispatchable('filter', _xfilter, _filter));
+module.exports = _curry2(_dispatchable('filter', _xfilter, function(pred, filterable) {
+  return (
+    _isObject(filterable) ?
+      _reduce(function(acc, key) {
+        if (pred(filterable[key])) {
+          acc[key] = filterable[key];
+        }
+        return acc;
+      }, {}, keys(filterable)) :
+    // else
+      _filter(pred, filterable)
+  );
+}));
--- a/src/mapObjIndexed.js
+++ b/src/mapObjIndexed.js
@@ -4,19 +4,19 @@ var keys = require('./keys');
 
 
 /**
- * Like `map` for objects, but passes additional arguments to the predicate function.
- * The predicate function is passed three arguments: *(value, key, obj)*.
+ * An Object-specific version of `map`. The function is applied to three
+ * arguments: *(value, key, obj)*. If only the value is significant, use
+ * `map` instead.
  *
  * @func
  * @memberOf R
  * @since v0.9.0
  * @category Object
- * @sig (v, k, {k: v} -> v) -> {k: v} -> {k: v}
- * @param {Function} fn A function called for each property in `obj`. Its return value will
- *        become a new property on the return object.
- * @param {Object} obj The object to iterate over.
- * @return {Object} A new object with the same keys as `obj` and values that are the result
- *         of running each property through `fn`.
+ * @sig ((*, String, Object) -> *) -> Object -> Object
+ * @param {Function} fn
+ * @param {Object} obj
+ * @return {Object}
+ * @see R.map
  * @example
  *
  *      var values = { x: 1, y: 2, z: 3 };
--- a/src/pickBy.js
+++ b/src/pickBy.js
@@ -15,7 +15,7 @@ var _curry2 = require('./internal/_curry2');
  * @param {Object} obj The object to copy from
  * @return {Object} A new object with only properties that satisfy `pred`
  *         on it.
- * @see R.pick
+ * @see R.pick, R.filter
  * @example
  *
  *      var isUpperCase = (val, key) => key.toUpperCase() === key;
--- a/src/reject.js
+++ b/src/reject.js
@@ -4,9 +4,7 @@ var filter = require('./filter');
 
 
 /**
- * Similar to `filter`, except that it keeps only values for which the given
- * predicate function returns falsy. The predicate function is passed one
- * argument: *(value)*.
+ * The complement of `filter`.
  *
  * Acts as a transducer if a transformer is given in list position.
  *
@@ -14,17 +12,19 @@ var filter = require('./filter');
  * @memberOf R
  * @since v0.1.0
  * @category List
- * @sig (a -> Boolean) -> [a] -> [a]
- * @param {Function} fn The function called per iteration.
- * @param {Array} list The collection to iterate over.
- * @return {Array} The new filtered array.
+ * @sig Filterable f => (a -> Boolean) -> f a -> f a
+ * @param {Function} pred
+ * @param {Array} filterable
+ * @return {Array}
  * @see R.filter, R.transduce, R.addIndex
  * @example
  *
  *      var isOdd = (n) => n % 2 === 1;
  *
  *      R.reject(isOdd, [1, 2, 3, 4]); //=> [2, 4]
+ *
+ *      R.reject(isOdd, {a: 1, b: 2, c: 3, d: 4}); //=> {b: 2, d: 4}
  */
-module.exports = _curry2(function reject(fn, list) {
-  return filter(_complement(fn), list);
+module.exports = _curry2(function reject(pred, filterable) {
+  return filter(_complement(pred), filterable);
 });
--- a/test/filter.js
+++ b/test/filter.js
@@ -17,6 +17,15 @@ describe('filter', function() {
     eq(R.filter(function(x) { return x > 100; }, []), []);
   });
 
+  it('filters objects', function() {
+    var positive = function(x) { return x > 0; };
+    eq(R.filter(positive, {}), {});
+    eq(R.filter(positive, {x: 0, y: 0, z: 0}), {});
+    eq(R.filter(positive, {x: 1, y: 0, z: 0}), {x: 1});
+    eq(R.filter(positive, {x: 1, y: 2, z: 0}), {x: 1, y: 2});
+    eq(R.filter(positive, {x: 1, y: 2, z: 3}), {x: 1, y: 2, z: 3});
+  });
+
   it('dispatches to passed-in non-Array object with a `filter` method', function() {
     var f = {filter: function(f) { return f('called f.filter'); }};
     eq(R.filter(function(s) { return s; }, f), 'called f.filter');
--- a/test/reject.js
+++ b/test/reject.js
@@ -25,6 +25,14 @@ describe('reject', function() {
     eq(R.reject(function(x) { return x > 100; }, []), []);
   });
 
+  it('filters objects', function() {
+    eq(R.reject(R.equals(0), {}), {});
+    eq(R.reject(R.equals(0), {x: 0, y: 0, z: 0}), {});
+    eq(R.reject(R.equals(0), {x: 1, y: 0, z: 0}), {x: 1});
+    eq(R.reject(R.equals(0), {x: 1, y: 2, z: 0}), {x: 1, y: 2});
+    eq(R.reject(R.equals(0), {x: 1, y: 2, z: 3}), {x: 1, y: 2, z: 3});
+  });
+
   it('dispatches to `filter` method', function() {
     function Nothing() {}
     Nothing.value = new Nothing();


CrossEye:
Two points:

1. I want to hold you to this bit from #1426:  Making these sort of changes involves seeing the functions

  > as useful higher-level abstractions over *groups* of types (*type classes* in Haskell terminology), provided they are principled (i.e. satisfy laws).

  What sorts of laws would you provide for `filter`?


2. Perhaps a bit of a devil's advocate here, but something I brought up in https://github.com/ramda/ramda/issues/1067#issuecomment-135923353 is why the strong focus on the values in an object at the expense of the keys?  Why not

  ```js
  filter(key => key < 'e', {foo: 2, bar: 3, baz: 5, quux: 7}); //=> {bar: 3, baz: 5}
  ```

 I don't actually disagree, but I haven't seen the compelling argument that says that this is the only reasonable version:

  ```js
  filter(val => val < 4, {foo: 2, bar: 3, baz: 5, quux: 7}); //=> {foo: 2, bar: 3}
  ```

davidchambers:
> What sorts of laws would you provide for `filter`?

  - __Identity:__ For all filterable values `x`,

        R.filter(R.T, x) ≍ x

  - __Emptiness:__ For all filterable values `x`,

        R.filter(R.F, x) ≍ R.empty(x)

  - __Associativity:__ For all filterable values `x`,

        R.filter(q, R.filter(p, x)) ≍ R.filter(p, R.filter(q, x))

> [W]hy the strong focus on the values in an object at the expense of the keys?

Let's first consider `map`. Why apply transformations to the values rather than the keys? Two reasons:

  - it's not clear how key collisions should be handled; and
  - `map` claims to accept an `a -> b`, but transforming keys would necessitate a `String -> String`.

Now let's consider `filter`. Why filter based on values rather than keys? Because `map` does! :stuck_out_tongue_winking_eye:

We could, of course, also provide a `filterKeys` function.

---

Emptiness does not current hold for Arguments objects, which suggests to me that `R.empty((function() { return arguments; }(1, 2, 3)))` should evaluate to `[]`.

CrossEye:
>> What sorts of laws would you provide for `filter`?

>   - __Identity:__ For all filterable values `x`,
> 
>         R.filter(R.T, x) ≍ x
> 
>   - __Emptiness:__ For all filterable values `x`,
> 
>         R.filter(R.F, x) ≍ R.empty(x)
> 
>   - __Associativity:__ For all filterable values `x`,

This one is actually __Commutativity__.

>         R.filter(q, R.filter(p, x)) ≍ R.filter(p, R.filter(q, x))

I can buy these.  It's not clear to me that these are useful in terms of _defining_ `filter`, but they are certainly at least _a propos_.

I did think that there were likely to be a good set of laws for such functions, but as you were summoning the rationale of #1426, it seemed incumbent upon you to provide reasonable laws when proposing such an extension of this function.  :smile:

As to the keys/values question, of course I'm not making a serious proposal to use the keys.  But there's a very serious point about what's meant by the notion of extending list functions to object ones. The same question could be raised regarding `map` too.  But there we already had an existing implementation to replace.   Moreover, in the case of `map`, there is a built-in bias too, since the problem with key collision make it unclear if there is a useful functor built around the keys or around `(key, value)` pairs.

There's some temptation, when working in JS, to take what's almost a pun, the implementation of Arrays atop Objects as something fairly fundamental, and to use that to build analogies between mapping lists and mapping object values.  But this is not truly fundamental.  For us, arrays are just the closest things to lists available, and these have little to do with dictionaries, which is all Objects would be for us in an ideal world.

And note that if we were to use a key-based implementation such as this:

```js
(pred, obj) => 
  R.fromPairs(R.filter(
    pair => pred(R.head(pair)), R.toPairs(obj)
  ));
```

this would clearly fulfill the Identity, Emptiness, and Commutativity laws described above.

So, while I do believe that filtering on the values is likely the most useful scenario, this is more like the `reduce` case than the `map` one: there is nothing fundamental driving us to one choice over the other.  And that makes me nervous.  Only slightly nervous, perhaps, but the feeling isn't going away.

davidchambers:
> [This] makes me nervous. Only slightly nervous, perhaps, but the feeling isn't going away.

Is anyone else nervous? (I'm excited!)

CrossEye:
>> [This] makes me nervous. Only slightly nervous, perhaps, but the feeling isn't going away.

> Is anyone else nervous? (I'm excited!)

Please note what it was that makes me nervous: the decision to consider the paramount facets of Objects their values, that we should be dealing with these by default instead of either the keys or the key-value pairs.  I'm nervous because this decision seems arbitrary.

buzzdecafe:
it is arbitrary. unconvinced.

davidchambers:
> I'm nervous because this decision seems arbitrary.

I don't see the decision as arbitrary: isomorphism with `map` is a reason to operate on values only.

This may be a case, as with `propEq`, where you and I weight factors differently. You place slightly more emphasis on doing what is right for each function individually, whereas I place slightly more emphasis on the consistency of Ramda as a whole.

davidchambers:
It's time to consult the Northern Europeans. @paldepind? @raine?

paldepind:
> So, while I do believe that filtering on the values is likely the most useful scenario, this is more like the reduce case than the map one: there is nothing fundamental driving us to one choice over the other. And that makes me nervous. Only slightly nervous, perhaps, but the feeling isn't going away.

This doesn't bother me at all. It often happens that there is more than one way to do something. In such cases choosing the most practical option seems reasonable to me. Lists for instance can be an Applicative in at least two different ways. We have chosen one in `R.ap` but there is nothing fundamental driving us to that choice either.

One thing that makes me a bit uneasy about this is that there is no established abstraction for this in neither Fantasy Land nor Haskell. This may or may not matter though. But I would probably prefer to follow the Fantasy Land structures initially and then consider ´filter´ afterwards. If we implemented ´concat´ for objects they would both be Semigroups and Monoids (since we already implement ´empty´ on objects. Then we could deprecate ´R.merge´.

davidchambers:
What say you, @CrossEye? If your answer is no, close the pull request. I won't be offended. :)

CrossEye:
I'm mostly positive on this, but still a bit nervous.  I'd still like to hear from @scott-christopher and @TheLudd.  Any comments, folks?

scott-christopher:
It would be nice have the key and value passed to the filtering predicate as a tuple, though without a dedicated tuple type I feel it might be a little clumsy to work with. If people do want that functionality, they can still do so by either using `filterObjIndexed` or piping the object through `toPairs` before `filter` (or `R.around(R.toPairs, R.fromPairs, R.filter)` if #1356 lands :wink: ).

So I would be happy with `filter` simply operating on the values of objects.

src/filterObjIndexed.js
```diff
@@ -0,0 +1,33 @@
+var _curry2 = require('./internal/_curry2');
+var _reduce = require('./internal/_reduce');
+var keys = require('./keys');
+
+
+/**
+ * An Object-specific version of `filter`. The predicate is applied to three
+ * arguments: *(value, key, obj)*. If only the value is significant, use
```
raine: How about using monospace for `(value, key, obj)`. I have a feeling it's going to look weird in docs like that.

raine:
If `filter` was only to pass the value in the predicate function, I think `filterByKeys` would be a reasonable addition as well. `(value, key, object) -> Bool` is not a very nice API to work if you're only interested in the key, and I don't see key as second-class as to only provide `filterObjIndexed`.

p.s. I find [`mapKeys`](https://github.com/ramda/ramda/wiki/Cookbook#map-keys-of-an-object) useful nearly in every project.

paldepind:
@scott-christopher 

> It would be nice have the key and value passed to the filtering predicate as a tuple

I don't think that would work. `filter` would have a type like this

```js
filter :: (a -> Boolean) -> f a -> f a
```

Where `f a` means "a filterable of `a`s". For arrays that would specialize to

```js
filter :: (a -> Boolean) -> [a] -> [a]
```

And for objects it would be

```js
filter :: (a -> Boolean) -> Object k a -> Object k a
```

But with you suggestion `filter` for objects whould have the type

```js
filter :: ((k,a) -> Boolean) -> Object k a -> Object k a
```

And thus the types wouldn't be consistent. I.e. it would be impossible to write code that worked seamlessly with both arrays and objects which is the main reason behind the polymorphism.

Correct me if I'm wrong.

scott-christopher:
Well one way to look at it is to treat an `Object` as being isomorphic to an indexed set of `(String, a)` tuples, where the signature would look like
```
filter :: ((String,a) -> Boolean) -> Object (String,a) -> Object (String,a)
```
However the signature probably makes more sense to use the only type parameter (the value) as the argument to the predicate
```
filter :: (a -> Boolean) -> Object String a -> Object String a
// or more succinctly
filter :: (a -> Boolean) -> {a} -> {a}
```

Where I get a little uncertain is thinking about what would be necessary for `filter` to be generically defined. The type would need to be `Foldable` to change its shape. It would also need to be a `Monoid` to have an `empty` for when `filter` is given a predicate that always returns `false` and `concat` for reconstructing the filtered instance. And it would need to be an `Applicative` (or just `Point`) to create a singleton instance using `of` to combine with `concat`.
e.g. something like:
```
// genericFilter :: Foldable f, Applicative f, Monoid f => Constructor (f a) -> (a -> Boolean) -> f a -> f a
genericFilter = (Type, p, f) => reduce((acc, x) => p(x) ? concat(acc, Type.of(x)) : acc, Type.empty(), f)
```

Now, if we were to just pass the values of an object to the predicate, we wouldn't have enough information to provide to `of` to reconstruct the filtered object, making me think that both the key and value need to be present in some form. Perhaps I'm just overthinking this now?

CrossEye:
> Perhaps I'm just overthinking this now?

Perhaps you are, but it's making me feel that my squeamishness might be justified...

paldepind:
> Well one way to look at it is to treat an `Object` as being isomorphic to an indexed set of `(String, a)` tuples, where the signature would look like
```
filter :: ((String,a) -> Boolean) -> Object (String,a) -> Object (String,a)
```

That still doesn't work. It doesn't type check. `filter` has the parametrically polymorphic type

```js
filter :: (a -> Boolean) -> f a -> f a
```

But yours has type

```js
filter :: ((String,a) -> Boolean) -> Object (String,a) -> Object (String,a)
```

If that should work `filter` would have to be of type

```js
filter :: ((String, a) -> Boolean) -> f a -> f a
```

But that would break the array instance.

> Where I get a little uncertain is thinking about what would be necessary for filter to be generically defined.

It's a very good question to ask. As far as I can tell you need monoid, foldable and `pure`/`of` from applicative. But requiring applicative is too much. Is that what point is? I haven't heard of it. But, with those you could define filter like this:

```js
function filter(pred, f) {
  return reduce((acc, a) -> f(a) ? concat(acc, pure(a)) : empty, empty);
}
```

You can also create filter on something that is a monoid and a monad (MonadPlus). But requiring a monad is also too much.

paldepind:
@CrossEye 

> Perhaps a bit of a devil's advocate here, but something I brought up in https://github.com/ramda/ramda/issues/1067#issuecomment-135923353 is why the strong focus on the values in an object at the expense of the keys?  Why not

  ```js
  filter(key => key < 'e', {foo: 2, bar: 3, baz: 5, quux: 7}); //=> {bar: 3, baz: 5}
  ```

> I don't actually disagree, but I haven't seen the compelling argument that says that this is the only reasonable version:

  ```js
  filter(val => val < 4, {foo: 2, bar: 3, baz: 5, quux: 7}); //=> {foo: 2, bar: 3}
  ```

Well. It _is_ the only reasonable version. A filter function has the type

```js
filter :: (a -> Boolean) -> f a -> f a
```

Where `f` is a filterable of `a`s. The only parametrically polymorphic value that an object contains is its values. And thus it is only possible to define `filter` on its values. If you where to define filter on its keys it would have the type

```js
filter :: (String -> Boolean) -> f a -> f a
```

But that does not match the polymorphic type of `filter`.

scott-christopher:
> If that should work filter would have to be of type
> `filter :: ((String, a) -> Boolean) -> f a -> f a`

`(String, a) -> Boolean` here was intended to represent a unary function accepting a tuple of `(String, a)`, not a binary function.

paldepind:
> `(String, a) -> Boolean` here was intended to represent a unary function accepting a tuple of `(String, a)`, not a binary function.

Yes? That is also how I interpreted it. But the types still don't match.

The type signature `filter :: (a -> Boolean) -> f a -> f a` says: "the only thing the predicate function must receive is the `a` inside the filterable". An `a` is not the same as a `(String, a)`. I.e. the signature you've given your `filter` for objects does not satisfy the polymorphism required by the `filter` defined in the type class.

scott-christopher:
I do agree @paldepind, hence my earlier comments. I'm just thrown off a little by the generic filter example and in the mindset of an object being iso to a list of `(String, a)` pairs (e.g. `fromPairs`/`toPairs`).

I'm still happy for this to land with the restriction of only operating over the values of an object. The `key`/`value` option is still obtainable via `compose(fromPairs, filter(p), toPairs)` if desired, albeit a little less efficient.

paldepind:
> I do agree @paldepind, hence my earlier comments.

Ok. I probably misunderstood you then. We do agree that this imaginative type class

```js
class Filterable f where
  filter :: (a -> Boolean) -> f a -> f a
```

Is forced upon us due to the signature for `filter` on arrays. And due to that we are forced on this instance for objects:

```js
instance Object a where
  filter :: (a -> Boolean) -> Object a -> Object a
```

In other words there is only a single possible way for us to define `filter` on objects. Everything is completely determined by the types and the laws specified by @davidchambers.

> I'm just thrown off a little by the generic filter example and in the mindset of an object being iso to a list of (String, a) pairs (e.g. fromPairs/toPairs).

Why is that throwing you off? A list is isomorphic to a set of `(Int, a)` pairs but filter doesn't receive an index. Our functor `map` for objects also can't receive keys. It seems consistent to me.

CrossEye:
@paldepind: I'm probably missing something fundamental, but I'm not following this objection:

> That still doesn't work. It doesn't type check. `filter` has the parametrically polymorphic type
> 
> ```js
> filter :: (a -> Boolean) -> f a -> f a
> ```
> 
> But yours has type
> 
> ```js
> filter :: ((String,a) -> Boolean) -> Object (String,a) -> Object (String,a)
> ```
> 
> If that should work `filter` would have to be of type
> 
> ```js
> filter :: ((String, a) -> Boolean) -> f a -> f a
> ```
> 
> But that would break the array instance.

What if we give a name to `(String, a)`, call it `KeyValuePair`?

then this is

```js
filter :: (KeyValuePair -> Boolean) -> Object KeyValuePair -> Object KeyValuePair
```

which precisely matches your `filter`, with `KeyValuePair` for `a`, and `Object` for `f`.

The question, it seems to me, is in how one views an object as a container, as a collection.  To me, the most logical way to do so is to view it as a set of key-value pairs, with only one further constraint: that no two elements in the set can have the same key.  There are at least two other ways to think about it: as a multiset of values, or as a set of keys.  In either of these, one needs to add additional concepts to deal with some sort of metadata of keys or values.  While these are quite possible, they lack the elegance of a set of key-value pairs.

And I really don't make sense out of this:
 
> The only parametrically polymorphic value that an object contains is its values. And thus it is only possible to define filter on its values

Are you suggesting that something is filterable only because the container is parametrically polymorphic?  Then [`Int32Array`][ia] would not be filterable, nor would `[String]`?  Obviously that isn't right, but then what is it you're saying?


  [ia]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Int32Array

paldepind:
> What if we give a name to `(String, a)`, call it `KeyValuePair`?

> then this is

> ```js
filter :: (KeyValuePair -> Boolean) -> Object KeyValuePair -> Object KeyValuePair
```

No no. You can't do that. You are not allowed to substitute anything for `a`. That is like saying that

```js
map :: (KeyValuePair -> KeyValuePair) -> Object KeyValuePair -> Object KeyValuePair
```

Is a valid `map` instance for `Object`. But the type has to be

```js
map :: (a -> b) -> Object a -> Object b
```

Because the `Object` must be able to contain _anything_.

> which precisely matches your `filter`, with `KeyValuePair` for `a`, and `Object` for `f`.

No it doesn't. You define `KeyValuePair` as `(String, a)`. And then you want to substitute `KeyValuePair` for `a`. But for that to work `(String, a)` would have to be the same as `a`. But clearly `(String, a) != a`.

I'll explain my point using functors as an example as they are more familiar. I apologize if the following explanation doesn't explain the things that the actual source of confusion. I also apologize that it is probably a bit confusing and overall not very good.

In Haskell the type class definition for functor looks like this:

```haskell
class Functor f where
  fmap :: (a -> b) -> f a -> f b
```

This says, among other things, that a functor must be a "container" of one parametrically polymorphic type. As examples of this we have `[]` and `Maybe`. Both are type constructors. One can pass them any type and they return a concrete type. Both can contain any `a`.

But something like `Pair`

```haskell
data Pair a b = (a,b)
```
takes two types as parameters and returns a concrete type. Thus we can not make it a functor. We can however make `Pair String`, `Pair Int` or in general any `Pair a` a functor. So even though `Pair` is a container of two values we will have to "fix" one of them before we can define a functor instance.

The functor type class also says that the thing you must pass to `f` in `fmap` is _excactly_ the `a` inside you functor. So if we make `Object a` a functor the only thing we can pass to `f` in `fmap` is `a`. So if we have an object of numbers `Object Number` then we must pass `Number` to `f`. We can not pass `(String, Number)`.

It doesn't matter that `Pair` is a container of two values or that `Object` is a list of key value pairs. That is just surrounding stuff around the `a` that the functor type class actually cares about.

CrossEye:
@paldepind:

I'm not sure if we're talking past each other or if you really have some insight I'm not grasping.  I certainly understand how parameterized types work.  I wonder if you're caught up in the JS notion of `Object` a bit much.  The big question here, as I see it, is how to view an object.  I would have no issue reformulating that as:

    filter :: (KeyValuePair -> Boolean) -> SetWithoutDuplicatedKeys KeyValuePair -> SetWithoutDuplicatedKeys KeyValuePair

and if you want something more strongly typed than is usual in JS, you can parameterize the KeyValuePair type.

In other words, an Object simply _is_ a set of key-value pairs, with one additional constraint disallowing duplicated keys.  This can easily be parameterized on the `value` side of the pair if you're looking for stronger typing, always an interesting question in JS.

I'm also confused by what standard you're trying to adhere to.  Are you just generalizing from `List::filter`?  Or is there some specification for `filter` that you think we're not meeting?

paldepind:
> The big question here, as I see it, is how to view an object.

But, that doesn't matter at all. Type classes are a way of recognizing the _common_ properties of certain structures and abstracting over them in a useful way. It doesn't matter what an object is. The thing we care about is _only_ the things it has in common with lists. And it does not have keys in common with lists. That is why the keys can't play a part when you define `map` for objects. And also why they don't matter when you define `filter`.

> I would have no issue reformulating that as:

>     filter :: (KeyValuePair -> Boolean) -> SetWithoutDuplicatedKeys KeyValuePair -> 
              SetWithoutDuplicatedKeys KeyValuePair

We want `R.filter` to work for both arrays and objects? Then you can't do that. It is a less polymorphic type compared to the current type for `R.filter`.

The current type for `R.filter` is:

```js
R.filter :: (a -> Boolean) -> [a] -> [b]
```

We want to generalize that for any `f` that is filterable. So the type becomes:

```js
R.filter :: (a -> Boolean) -> f a -> f a
```

So specifically for `Object` that becomes:

```js
R.filter :: (a -> Boolean) -> Object a -> Object a
```

There is no way we can get to the type in the quoted code above.

This is a very practical issue. _If_ you pass a key value pair to the predicate function when filtering on an object then you've not gained anything in terms of code reuse – you're just overloading a function. I.e. you're doing ad hoc polymorphism instead of parametric polymorphism. Consider this example:

```js
var keepFoo = R.filter(R.equals('foo'));
keepFoo(['bar', 'foo', 'bar', 'foo']);
keepFoo({key: 'bar', key2: 'foo', key3: 'bar'});
```

The only way the above code will work is if you only pass the values from the object to the predicate functions. This is also what the types told us.

Basically, given the above and @davidchambers laws there is only one possible way to implement `R.filter` for objects.

paldepind:
>  > The only parametrically polymorphic value that an object contains is its values. And thus it is only possible to define filter on its values

>Are you suggesting that something is filterable only because the container is parametrically polymorphic? Then Int32Array would not be filterable, nor would [String]? Obviously that isn't right, but then what is it you're saying?

A JavaScript object is only polymorphic in it values. The keys can only contain strings. Thus you can't define a functor instance for objects where maps operates on the keys.

Int32Array nor [String] can be made an instance of Filterable just as they can't be made an instance of Functor. They are concrete types and have kind `*`. But you can only make something of kind `* -> *` a Functor or Filterable.

CrossEye:
>>  Are you suggesting that something is filterable only because the container is parametrically polymorphic?  Then [`Int32Array`][ia] would not be filterable, nor would `[String]`?  Obviously that isn't right, but then what is it you're saying?

> A JavaScript object is only polymorphic in it values. The keys can only contain strings. Thus you can't define a functor instance for objects where map operates on the keys.

> Int32Array nor [String] _**can**_ be made an instance of Filterable just as they can't be made an instance of Functor.

<small>_(emphasis added)_</small>

I'm getting confused enough by this discussion that I'm not sure if the highlighted word is a typo.

If you do think the `Int32Array` or `[String]` can be made Filterable, then I agree, and I don't understand why an type that consists only of `(String, Value)` pairs could not.  If you don't think they can be made filterable, what precisely is wrong with

```js
R.filter(s => /^[a-z]{3}$/.test(s), ['one', 'two', 'three', 'four', 'five', 
                                     'six', 'seven', 'eight', 'nine', 'ten'])
//=> ['one', 'two', 'six', 'ten']
```

Granted, at the moment, that's running on Ramda's generic `list` filter, but it could just as easily be `filter:: (String - Bool) -> [String] -> [String]` and still pass all the laws that we care about.

My thought is that a JS object is not best thought of as a parameterized container type; that leaves too many open questions.  Instead, it simply is something like an `AssocList`.  And we should be able to run a filter against an object such as this:

```js
var obj = {a: 'a', b: 10, c: false, d: 'd'};
```
with a call such as

```js
filter((k, v) => k == v, obj); //=> {a: 'a', d: 'd'}
```

or

```js
filter(pair => pair[0] == pair[1], obj); //=> {a: 'a', d: 'd'}
```

Obviously, from this, we could also specialize to functions that test only against values, or only against keys.  But I can simply not see how this is not a legitimate extension of `filter`.  It seems to me you're saying that it's not legitimate because it doesn't match a certain abstraction.  And my answer is that if that's the case, then it's likely the abstraction that's wrong.  That's why I wanted to know if you were pointing to an existing standard for an abstraction of `filter`.  If we're making this up as we go along, I'm much more likely to assume we're getting the abstraction wrong if things don't seem to be working correctly.  If there's a long-lived standard that many are living by, then I would assume that it's much more likely a problem with our assumptions than with the standard.


  [ia]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Int32Array

paldepind:
> I'm getting confused enough by this discussion that I'm not sure if the highlighted word is a typo.

It's a typo.

> If you do think the Int32Array or [String] can be made Filterable, then I agree, and I don't understand why an type that consists only of (String, Value) pairs could not.

`Int32Array` can't be made a filterable for the same reasons that it can't be made a functor.

This should work:

```js
var keepFoo = R.filter(R.equals('foo'));
keepFoo(['bar', 'foo', 'bar', 'foo']);
keepFoo({key: 'bar', key2: 'foo', key3: 'bar'});
```

Then we've gotten the types right, found a useful abstraction and can reap the benefits. Otherwise we're wandering into Lodash land.

But I can't explain it better than what I've already tried.

joneshf:
# `Object a`
I think what @paldepind is trying to say is, if we view js objects as containing any value indexed by strings, then it's natural that it would have a type like this:

```haskell
data Object a
```

If this is the type given to objects, then it does not mention what the indexing values are in the implementation.

If you wanted to implement this type using strings as the indices, you could do so as an associative list with no duplicated indexes. You might have a type and implementation like this:

```haskell
data Object a = Object [(String, a)]
```

Even though you know what the implementation is, you cannot reach into it to pull out type information. The only information you can get is what's on the left side of the equal sign.

So if you wanted to implement `Functor` for this given data type, you have exactly one implementation:

```haskell
instance Functor Object where
    fmap f (Object xs) = Object (fmap (fmap f) xs)
```

There is no other type safe way to implement the `Functor` instance for this data type and have it be law abiding. If you notice, we don't touch the indexes at all. We cannot by parametricity.

If we wanted to look at what kind of monomorphized type this `fmap` might have, it would look something like this:

```haskell
fmap :: (a -> b) -> Object a -> Object b
```

No where are the indices mentioned.

Assuming there was some "`Filterable`" (hate that name btw) class that had a `filter` method:

```haskell
class Filterable f where
    filter :: (a -> Bool) -> f a -> f a
```

There's again, very limited ways to implement this (maybe only one given those laws @davidchambers posted):

```haskell
instance Filterable Object where
    filter p (Object xs) = Object (filter (p . snd) xs)
```

Once again, you cannot touch the indexes at all, again ensured by parametricity. The only things that you can touch are the values.

If we want to look at what a monomorphized type this `filter` would have for objects, we'd see it was something like:

```haskell
filter :: (a -> Bool) -> Object a -> Object a
```

Again, no mention of the indices, because we cannot.

If I misunderstood where @paldepind was coming from, then ignore all that.

# `Object (String, a)`

That said, if you wanted to represent js objects in some other way, then things get more complex.
If you wanted to be able to say that an object had this type:

```haskell
foo :: Object (String, a)
```

Then you would need the object to be more general than before. You might want something like this if you were using an associative array implementation:

```haskell
data Object a = Object [a]
```

So your `Functor` instance becomes:

```haskell
instance Functor Object where
    fmap f (Object xs) = Object (fmap f xs)
```

And your "`Filterable`" instance becomes:

```haskell
instance Filterable Object where
    filter p (Object xs) = Object (filter p xs)
```

And once again, you have no other type safe law abiding implementation for the `Functor` instance, and probably the same for the `Filter` instance.

Looking at these as monomorphized types you'd have:

```haskell
fmap :: (a -> b) -> Object a -> Object b
filter :: (a -> Bool) -> Object a -> Object a
```

Which you could specialize to:

```haskell
fmap :: ((String, a) -> (String, b)) -> Object (String, a) -> Object (String, b)
filter :: ((String, a) -> Bool) -> Object (String, a) -> Object (String, a)
```

But these things have to go together in unison. You cannot have the "`Filterable`" instance using this more general representation, while the `Functor` instance uses the more specific representation, or vice versa.

Meaning this doesn't work:

```haskell
fmap :: (a -> b) -> Object a -> Object b
filter :: ((String, a) -> Bool) -> Object (String, a) -> Object (String, a)
```

And neither does this:

```haskell
fmap :: ((String, a) -> (String, b)) -> Object (String, a) -> Object (String, b)
filter :: (a -> Bool) -> Object a -> Object a
```

# Problems with `Object (String, a)`

But there are issues with this more general implementation. Mainly that you cannot actually ensure it's an associative array, nor can you ensure that it only has unique keys. Since the `a` can be anything, you could make it `Bool` and then you'd have `Object Bool`, which would have an implementation of `Object [Bool]`, and there's no way to make that an associative array.

# `Object i a`

So you could take the original object thoughts and make them more general: Take the index in at the type level. So you can talk about the index and ensure it's an associative array (or whatever implementation you want). You can also make sure it's ordered or unique or whatever other constraints you want to put on it.

```haskell
data Object i a
```

You might have the implementation like so:

```haskell
data Object i a = Object [(i, a)]
```

Notice that now we can talk about the index on the type level. In other words, we get some information about the implementation at the type level. Though we don't actually know what it's used for.

And when we go to implement the `Functor` instance, we get a very similar implementation:

```haskell
instance Functor (Object i) where
    fmap f (Object xs) = Object (fmap (fmap f) xs)
```

The reason it's `(Object i)` is as @paldepind  explained. You need the kind to be `* -> *`, and this `Object` has kind `* -> * -> *`, whereas `Object i` has kind `* -> *`.

What does this make the monomorphized type?

```haskell
fmap :: (a -> b) -> Object i a -> Object i b
```

If you notice, even though we can see the indices now, we still cannot do anything with them, since `fmap` prevents this via parametricity.

If we implement the "`Filterable`" instance, we get another very similar implementation:

```haskell
instance Filterable (Object i) where
    filter p (Object xs) = Object (filter (p . snd) xs)
```

Looking at this new monomorphized type, we see again, that we cannot talk about the indices:

```haskell
filter :: (a -> Bool) -> Object i a -> Object i a
```

There's no mention of the indices that we can use (again thanks to parametricity).

So this implementation doesn't help any more than the original.

# Summary

Filtering has to work on the values if you want a robust system. This is not an arbitrary decision. Rather, it makes more sense to say, filtering has to work on the polymorphic type variable. If you monomorphize the type variable to be something else, that doesn't change the fact that the implementation must work on whatever that type variable is.

If this doesn't work, then the definition of "`Filterable`" needs to change. I won't explore that here, but given that you want compatibility with other things people create that's probably not a good idea.

If you are willing to change the rules arbitrarily for different functions, then there's no telling what each function will do, and you lose reasoning abilities.

# Addendum

There's a haskell package that has these filter ideas: http://hackage.haskell.org/package/witherable I've not used it so I can't comment on it, but it seems pretty robust, and builds on `Traversable` so I'd say it's probably a good approach.

paldepind:
Thanks a lot for taking the time to write this @joneshf. That was a really great write up and explanation.

> If I misunderstood where @paldepind was coming from, then ignore all that.

Not at all!

> But these things have to go together in unison.

Indeed. And may I point out that Ramda has already picked a functor instance for Object.

You do mention it but I want to emphasize that both `Object (String, a)` and `Object (String, a)` are, in a way, too general. They both makes it possible to express types that we can't represent with a JavaScript object.

I'll have to take a look at Witherable. It looks interesting.

CrossEye:
> But these things have to go together in unison. You cannot have the "Filterable" instance using this more general representation, while the Functor instance uses the more specific representation, or vice versa.

Thank you.  I think this is the point I was missing from @paldepind's explanations.

I'm still not thrilled with this as the abstraction of `Object`, though, even if it did let us easily do something like #1429.  I believe the most helpful abstraction of `Object` is as a collection of `key-value` pairs.  Because of the restriction on duplicated keys, this is not the full collection of `(String, value)`, but only a subset, but it makes clear some nice properties:

  * `fromPairs` is a retraction of `toPairs`, and `toPairs` a section of `fromPairs`
  * there is a clear-cut general notion of `filter

paldepind:
> I'm still not thrilled with this as the abstraction of Object, though, [...]. I believe the most helpful abstraction of Object is as a collection of key-value pairs.

I don't understand what you're saying. Object _was_ defined _excactly_ as a collection of key-value pairs:

```haskell
data Object a = Object [(String, a)]
```

> let's not pretend that there was only one choice for either one.

There was only one _good_ choice. As Jones showed you clearly can't get `Object (String, a)` because then you'd end up defining `Object` as a list:

```haskell
data Object a = Object [a]
```

The above definition says: "An Object is a list of `a`s". But that's not what we want.

`Object i a` would have been a fine type if JavaScript objects could contain polymorphic values as keys. But then we would still be forced to define both `filter` and `map` in the exact same way.

CrossEye:
>> I believe the most helpful abstraction of Object is as a collection of key-value pairs.

> I'm not sure what you mean by this. Object _was_ defined as a collection of key-value pairs:
> 
> ```haskell
> data Object a = Object [(String, a)]
> ```

Definitely some communication issues here.  Where was it defined as such?  I know of no Haskell implementations of JS.  I'm talking about our mental model of `Object`.  When we discuss objects as containers of simple types, I feel that we're missing something essential.  A `Set` or a `Maybe` is clearly a container of elements of certain type.  A `List` is an _ordered_ collection of elements of a certain type.  "Ordered" gives additional context, and means that it contains strictly more information than a Set.  A `Map` contains still more information, associating each value with a key.  I think of `Object` as a specialization of `Map`.  But this is all at the level of our mental models of these objects.  We can see a `Map` as a collection of values somehow indexed by keys, or, as -- and I think this is a simpler picture -- a set of key-value pairs, restricted by a no-key-duplication rule.

I can just about read Haskell.  I certainly cannot write it.  I don't think in those terms.  So this is quite likely incorrect, but I think it comes somewhat close to what I mean:

```haskell
data Object a = Object (Data.Map String a)
```

Would that not give us this from @joneshf?:

```haskell
fmap :: ((String, a) -> (String, b)) -> Object (String, a) -> Object (String, b)
filter :: ((String, a) -> Bool) -> Object (String, a) -> Object (String, a)
```

This _would_ mean that we'd have to rethink `map` for Objects.  But I don't think that's necessarily a bad idea.  I really am wondering if we've been thinking about it wrong.

I'm afraid I don't understand this bit at all:

> But there are issues with this more general implementation. Mainly that you cannot actually ensure it's an associative array, nor can you ensure that it only has unique keys. Since the `a` can be anything, you could make it `Bool` and then you'd have `Object Bool`, which would have an implementation of `Object [Bool]`, and there's no way to make that an associative array.

If our type was _not_ defined as `Object a` but as something that took key-value, or specifically String-value pairs, how could you make it `Object Bool`?

I really would like to understand this before moving ahead with this.  And I'm afraid it's simply me being dense, as there are some very smart people trying hard to tutor me.  But somehow it feels to me that the description of Objects as simply containers of values loses essential information.

paldepind:
> Would that not give us this from @joneshf?:

The polymorphic type for `fmap` is:

```haskell
fmap :: (a -> b) -> f a -> f b
```

Specializing to the type defined by `data Object a = Object (Data.Map String a)` gives:

```haskell
fmap :: (a -> b) -> Object a -> Object b
```

You can't substitute `a` for `(String, a)` since `a != (String, a)`.

CrossEye:
I guess I'm just dense.  I simply don't see why this doesn't make it a Functor:

```haskell
data Object a = Object (Data.Map String a)

instance Functor Object where
  fmap f o = fromList (map modifiy (toList o))
      where modify (s, v) = (s, f v)
```

I don't see how this prohibits a `filter` function from actually _using_ both `s` and `v`, but perhaps that's where the rubber meets the road.

Still, this is not a language I speak fluently.  And I can't get anyone to respond to my questions about whether it seems correct to model an Object as a collection of key-value pairs.  So I don't know if we're even close to coming together or headed in entirely different directions.

paldepind:
> I guess I'm just dense. I simply don't see why this doesn't make it a Functor:

It is a functor and with a few modifications that is a valid functor instance! :smile: This compiles:

```haskell
import Data.Map (Map)
import qualified Data.Map as Map

data Object a = Object (Map String a)

instance Functor Object where
  fmap f (Object o) = Object (Map.fromList (map modify (Map.toList o)))
    where modify (s, v) = (s, f v)
```

I've just corrected a few minor things. But look what you're passing `f`. You're passing it the value from the object. So `fmap` has this type:

```haskell
fmap :: (a -> b) -> Object a -> Object b
```

And not the type you wrote before:

```haskell
fmap :: ((String, a) -> (String, b)) -> Object (String, a) -> Object (String, b)
```

Again we are forced to pass only the values to the function.

>  And I can't get anyone to respond to my questions about whether it seems correct to model an Object as a collection of key-value pairs.

I think that is very much correct. That is what Jones did in the definition that he introduced and that I refered to before:

```haskell
data Object a = Object [(String, a)]
```

Pretty much says "The type `Object a` is a collection of key-value pairs where the values are `a`s". The thing at the left site of the equals sign is the name of the type and the thing on the left is what it actually is – it's value.

CrossEye:
Maybe I'm getting there!  :smile:

But I'm not all the way there yet.

> But that an object is a list of key-value pairs doesn't mean that we can pass key-value pairs to `map` or `filter`. A list is a bunch of `Cons` cells and a `Nil` in the end. But that doesn't mean we can pass `Cons` cells nor `Nil` to `map` or `filter`.

No, of course not.  But if we can model Obects via `Data.Map`, the implementation gives us two different `filter` functions: [`filterWithKey`][fk] and plain [`filter`][fi].

It seems that the argument is that we need to privilege `filter` over `filterWithKey` because we need to provide some uniformity between the Functor view of an Object and the Filterable (agreed, @jonesjf -- there has to be a better name).  So, since we have `map :: (a -> b) -> Object a -> Object b`, we must therefore choose the similar `filter`.

But why not go the other direction and say that we should choose the more general representation for `map` too, [`mapWithKey`][mk]?  This would of course mean that we couldn't use the same functions for them as we do for a container of plain values.  `map(square, obj)` would make no sense.  But if our conception is that the constituents of Objects are key-value pairs, then this should be obvious: _you can't square a key-value pair._

The big advantage this holds for me is that it's a better fit for Objects-viewed-as-records.  One is not likely to be running a simple value function on a heterogeneous set of values; it seems much more likely to run a somewhat more sophisticated function on a set of (String, value) pairs.

Or think of a differences discriminator (untested):

    fieldsChanged = (oldObj, newObj) => filter((key, value) => oldObj[key] != value, newObj)

It's these kinds of cases that make me think that the mental model of an Object as primarily a collection of values is not robust enough.

  [fi]: https://downloads.haskell.org/~ghc/6.12.3/docs/html/libraries/containers-0.3.0.0/Data-Map.html#v:filter
  [fk]: https://downloads.haskell.org/~ghc/6.12.3/docs/html/libraries/containers-0.3.0.0/Data-Map.html#v:filterWithKey
  [mk]: https://downloads.haskell.org/~ghc/6.12.3/docs/html/libraries/containers-0.3.0.0/Data-Map.html#v:mapWithKey

joneshf:
I think some alpha renaming would help in this situation.

If you define your type as.

```haskell
data Object v = Object (Data.Map String v)
```

and assuming you implemented a law abiding instance for `Functor` and "`Filterable`". 

Then the slightly more monorphic types you've given here:
```haskell
fmap :: ((String, a) -> (String, b)) -> Object (String, a) -> Object (String, b)
filter :: ((String, a) -> Bool) -> Object (String, a) -> Object (String, a)
```

are certainly valid, but I can almost assure they're not what you meant.

Remember that when talking at the type level, you can only reference things on the left side of the equal sign. So you have a signature like:

```haskell
fmap :: ((String, a) -> (String, b)) -> Object (String, a) -> Object (String, b)
```

means that your object you're mapping over has this type:

```haskell
Object (String, a)
```

and if it has that type, then it has this implementation:

```haskell
Object (Data.Map String (String, a))
```

You simply substitute the `v` for `(String, a)`

This is what @paldepind means when he says `a != (String, a)`. The type variable `a` in the `Object a` type is not the same as `(String, a)` in the argument to the first argument of `fmap`. 

It's a type error, that would be caught by a compiler, but since js has no compiler, your head becomes the compiler and you have to catch the type error in your head.

Is that any clearer?

joneshf:
> Still, this is not a language I speak fluently. And I can't get anyone to respond to my questions about whether it seems correct to model an Object as a collection of key-value pairs. So I don't know if we're even close to coming together or headed in entirely different directions.

It can be modeled however, just have to keep in mind the limitations of each model.

```haskell
data Object a = Object [a]
```

Does not give you anything stronger than a plain array. And this is ensured by parametricity. The only purpose for something like this is to provide different implementations of something stronger than `Functor` or differentiate things at the type level.

```haskell
data Object a = Object [(String, a)]
```

Gives you the ability to ensure it's an associative array, and that there are no duplicate keys.

```haskell
data Object i a = Object [(i, a)]
```

Gives you the most power since `i` can be anything (you can index on numbers instead of strings, or booleans, or some other object). 

However, in all of these cases, you cannot talk about the indices for the `Functor` implementation. And this is ensured by parametricity. If you do talk about the indices, you've broken all reasoning ability for anything that uses the `Functor` instance.

joneshf:
> But if we can model Obects via Data.Map, the implementation gives us two different filter functions: filterWithKey and plain filter.

Sure, you can have as many of these functions as you want. In fact, there should be at least these two, Or rather implement the latter in terms of the former. That's not the problem though. The problem is when you want to apply laws to these functions. If you want your laws to be coherent, across different data types, you cannot arbitrarily choose to filter by key, or by value (at least with the formulation of "`Filterable`" given above). You **have** to choose the values to filter by.

> It seems that the argument is that we need to privilege filter over filterWithKey because we need to provide some uniformity between the Functor view of an Object and the Filterable

It doesn't have anything to do with those two typeclasses. I apologize if I conflated the explanation. I was providing the `Functor` information because it seems like most of us have a decent grasp on it. 

What it does have everything to do with is the type given to `Object`. If a data type has at least one type variable, then it can be a `Functor` (well mostly that's true...). And the only want to abide by the laws while implementing that `Functor` is to map over the last type variable. This is the reason a compiler lie ghc can automatically derive the functor instance of any data type without you writing anything (again, mostly true...). If you use `Object a`, then whatever that `a` is in the implementation is what gets mapped over **and only that `a`**. map over anything else and you're either causing type errors or you're breaking laws.

> But why not go the other direction and say that we should choose the more general representation for map too, mapWithKey? 

Sure, this can be done as well. Your type becomes something like:

```haskell
data Object a i = Object [(i, a)]
```

But notice, you can't map over the values, nor can you filter over the values. You've stated in the type that you want to map over `i`, and filter over `i`, since it's the last type variable.

> This would of course mean that we couldn't use the same functions for them as we do for a container of plain values. map(square, obj) would make no sense.

Right, that's a type error that a compiler would catch, but your head has to instead in js.

> It's these kinds of cases that make me think that the mental model of an Object as primarily a collection of values is not robust enough.

Agreed, most times you want a better data type when you find situations like this. The reason that discriminator function is hard to look at is because `filter` is the wrong abstraction. What you really want is a data type (notably a `Set`) that supports set operations, in particular symmetric difference (since that's the algorithm you implemented).

joneshf:
Hopefully this doesn't confuse more than it helps but I think this post is very enlightening in explaining why the type variables determine the behavior of a data type: 

http://bitemyapp.com/posts/2015-10-19-either-is-not-arbitrary.html

CrossEye:
> However, in all of these cases, you cannot talk about the indices for the `Functor` implementation. And this is ensured by parametricity. If you do talk about the indices, you've broken all reasoning ability for anything that uses the `Functor` instance.

Let me put it a different way.

What do we gain by making `Object` a `Functor` if that means abandoning the central understanding of the Object type?  Most Objects in my experience are not parameterized over a single type; they are hetereogeneous, essentially Records.  Would it not be better to define a useful interface, even it it can't match certain well-known types, such as Functor?

CrossEye:
> Hopefully this doesn't confuse more than it helps but I think this post is very enlightening in explaining why the type variables determine the behavior of a data type:
> 
> http://bitemyapp.com/posts/2015-10-19-either-is-not-arbitrary.html

Very funny.  I ran across that not an hour ago trying to figure some of this stuff out!

joneshf:
> What do we gain by making Object a Functor if that means abandoning the central understanding of the Object type? Most Objects in my experience are not parameterized over a single type; they are hetereogeneous, essentially Records. Would it not be better to define a useful interface, even it it can't match certain well-known types, such as Functor?

There's the rub. 

We gain the ability to let people be lazy and pass around the wrong data types. If you're trying to `map` and `Object` in js, you're doing the wrong thing. There's no two ways about it. But remember, if you're trying to `map` an `Array` in js, you're doing the wrong thing, since arrays are heterogenous as well. Unless you can ensure that all the values are the same (which you can't), then you're doing the wrong thing.

I do it all the time, **because** I'm lazy. But I'm willing to be lazy and pray to the gods that it doesn't blow up on me down the line.

The same applies to objects. If you just willy nilly try to map over any random object, it's going to blow up on you. If you try to ensure that your object has values that are all the same type, then it most likely won't blow up on you. This is the exact line of reasoning that allows us to think that `map` works properly for arrays.

So is it worthwhile to deprecate effectively all of `ramda` because you can't know for sure that what you pass around actually has the type given? I'd say no. I'd hope most people would say no.

Let us be lazy and assume that the object we're given has values that are all numbers or strings or arrays of things.

CrossEye:
> There's the rub.

> We gain the ability to let people be lazy and pass around the wrong data types. If you're trying to map and Object in js, you're doing the wrong thing. There's no two ways about it. But remember, if you're trying to map an Array in js, you're doing the wrong thing, since arrays are heterogenous as well. Unless you can ensure that all the values are the same (which you can't), then you're doing the wrong thing.

I don't really have a problem with "you're doing the wrong thing."  This is Javascript.  If you want to work in a type-safe language, you'll have to look elsewhere.  So Ramda is going to do its best to give you useful tools, and assume that you're a grown-up and can handle them without hurting yourself.  (This is not to belittle any of the efforts to add type-checking; whatever we do add will never be able to do what a fully type-safe static compiler can do, though.)

I was happy enough when we defined a `Functor` instance for `Object`.  It didn't really seem to do any harm.  We'd already published the functionality, just under a less convenient name.  And even now, I'm not really unhappy with it; it seems only occasionally useful, but I can certainly contrive examples where it would make sense.

I do find a difference between an `Array` and an `Object`; while it's true that either _can_ be heterogeneous, I usually use `Array`s to represent homogeneous lists;  I rarely use `Object`s that way.  I think this is common.  So, while, yes, `Arrays` can store heterogeneous tuples and `Objects` can store homogeneous mappings,  `Functor` well captures the most common behavior of `Arrays` but not of `Objects`.  Nonetheless, no other use of `map` for objects is significantly more compelling.  Yes, there is a point to be made for `map :: ((String, a) -> b) -> Object (String, a) -> Object (String b)` (or whatever), it's not a strong one..

But then we come to `filter`.  This one is different to me.  First of all, `Functor` is familiar to me.  My graduate work in Mathematics was a long time ago, and Category Theory was my least favorite course, but this rings a bell from years ago, and because of what little Haskell and what little Scala I know, and because of FantasyLand, `Functor` is now commonplace to me.  But a general-purpose abstraction for `filter` is not.  It seems to have been created on the fly in the comments here, and then through the magic of types and some isomorphisms with the existing `Functor` implementation, been forces in a direction that seems less correct for what one would want `filter` to do on `Object`s.  No one has answered my question about whether this is some well-known type that we're trying to match.

I'm 80 - 90% of the way, I think, to fully understanding why these two types are so linked according to the descriptions here.  But instead of my vague acceptance of `Functor-for-Objects` rubbing off onto `Filter-for-Objects`, instead, the connection is making doubt the `Functor-for-Objects`.  It's not all that useful, and if it's leading us away from the most useful definitions of commonly understood functions, perhaps it's time to jettison it. 

And even now, this is more to me an interesting open question than something I have settled in my head.  As I said early on, this makes me a bit nervous.

@paldepind, @joneshf: thank you very much for taking the time to explain this so thoroughly.  Thank you, too, for your patience.

joneshf:
I think it's important to reiterate a point you made:

> But a general-purpose abstraction for filter is not. It seems to have been created on the fly in the comments here, and then through the magic of types and some isomorphisms with the existing Functor implementation, been forces in a direction that seems less correct for what one would want filter to do on Objects. No one has answered my question about whether this is some well-known type that we're trying to match.

This is very important. The reason it's so tough to meld an object to this definition is because it's almost assuredly the wrong abstraction (as I believe you stated earlier). Since there aren't any canonical representations for filter, why not make a better one? Along the way, we needn't restrict ourselves to haskell98 implementations.

We know that you can't filter an array and an object the same way, so you might assume that for any given data type, the way you filter might depend on the data type.

A first pass might give something like this:

```haskell
class Filterable f k where
    filter :: (k -> Bool) -> f a -> f a
```

In the case of array, we'd want the `f` to be `[]`, and the `k` to be `a`. For Objects we'd want the `f` to be `Object` and the `k` to be `(String, a)`.
So we get something like:

```haskell
instance Filterable [] a where
    -- filter :: (a -> Bool) -> [a] -> [a]
    filter p []     = []
    filter p (x:xs) = if p x then x:filter p xs else filter p xs

instance Filterable Object (String, a) where
    -- filter :: ((String, a) -> Bool) -> Object a -> Object a
    filter ...
```

Unfortunately, this causes a problem in the compiler. Since it is not stated anywhere that `k` and `a` are the same, this is a type error. So we need to have the two values be the same for arrays, we might want to change the definition to something like this:

```haskell
class Filterable f k where
    filter :: (k -> Bool) -> f k -> f k
```

So we'd have an implementation like this:

```haskell
instance Filterable [a] a where
    -- filter :: (a -> Bool) -> [a] -> [a]
    filter ...
```

And this works great for arrays, but puts us back in the same spot as before for objects.

What we need is to not talk about a data type that needs an argument, and instead we want to talk about the whole data type while still keeping polymorphism. We want filter to have a type like: `(k -> Bool) -> f -> f` Well, we can do this, but we need some way to say that a given data type has a specific way to filter it.

```haskell
class Filterable f k where
  filter :: (k -> Bool) -> f -> f
```

So now we could implement a filter for both arrays and objects.

```haskell
instance Filterable [a] a where
    -- filter :: (a -> Bool) -> [a] -> [a]
    filter ...

instance Filterable (Object a) (String, a) where
    -- filter :: ((String, a) -> Bool) -> Object a -> Object a
    filter ...
```

Of course, this is not entirely safe, since you can have multiple instance for a given data type, so there's ways to make it safer on the type level with like functional dependencies.

And yet, there are probably other ways to do the same thing.  For one, you could get away with not having a type class at all, and just using `(Alternative f, Foldable f) => (a -> Bool) -> f a -> f a`

The point is, there are other ways to formulate this idea. And going with the formulation originally given is what is the main issue here.

paldepind:
> What do we gain by making Object a Functor if that means abandoning the central understanding of the Object type? Most Objects in my experience are not parameterized over a single type; they are hetereogeneous, essentially Records. Would it not be better to define a useful interface, even it it can't match certain well-known types, such as Functor?

I don't think abandoning is the right word. We are not loosing anything. But, when you make something a functor you must identify a thing that the something has in common with all other functors. So we must ask ourselves: what does an object have in common with lists, sets, trees, etc? It does not have keys in common with sets. So we can't use the keys when we define `fmap`. Because `fmap` must work on all functors.

I think there is a difference between asking "what is an object?" and asking "what makes an object a functor?". An object is a collection of key-value pais. An object is a functor because it can contain any type as its values. So when we make it a functor we can only use this property of objects. But we can still go ahead and write all sorts of object-specific functions if we want.

> But why not go the other direction and say that we should choose the more general representation for map too, mapWithKey? This would of course mean that we couldn't use the same functions for them as we do for a container of plain values.

Because then you're not satisfying the functor laws. The functor abstraction is exactly about using the same functions for all things that can be seen as containers of plain values. If you break that then you're not defining a functor.

@joneshf 

> The point is, there are other ways to formulate this idea. And going with the formulation originally given is what is the main issue here.

I agree that it is a better `Filterable` definition. It seemed wrong to require that instances be `* -> *` since monomorphic containers are filterable as well. But I don't like the instance for `Object`. Now it's impossible to reuse predicate function for filtering both `List` and `Object`.

CrossEye:
> I think there is a difference between asking "what is an object?" and asking "what makes an object a functor?". 

I'm taking one step further back, though, and asking if we should even bother making this a `Functor`.  That requires that we consider our type to be in some important way a collection of values.  I'm wondering if the reasonable abstraction of `Object` does not include the notion of "collection of values" necessary to being a `Functor`.  I really liked this when I first saw it, but I am growing much less convinced, especially as we look at what `filter` might mean.

I'm not at all set on removing it, but I'm less comfortable.  It's a similar twinge to what I feel when people start talking about `foldObj` -- something fundamental is off, even if it's hard to explain what.

algesten:
for me object as a `functor` is not a benefit. objects are records, and if ramda lets me filter on key-value pairs, I can decide what to do using the key and deal with the fact that the value can be anything. granted, any notion of type stringency is out the window, but that was the case already when I chose javascript.

and the reverse. defining a filter predicate that I can reuse for both lists and objects seems just weird (since my mental model of object is record). I would struggle to ever benefit from this reuse.

davidchambers:
> objects are records

That is one of two roles objects serve in JavaScript. The other is mapping values of type `String` to values of type `a` (`StrMap a` in PureScript).

It would be hard to argue that this is a record rather than a dictionary:

```json
{
  "AL": "Alabama",
  "AK": "Alaska",
  ...
  "WI": "Wisconsin",
  "WY": "Wyoming"
}
```

CrossEye:
>> objects are records

> That is one of two roles objects serve in JavaScript. The other is mapping values of type `String` to values of type `a` (`StrMap a` in PureScript).

I think this is the fundamental tension here, though.  There is a reasonable `filter` implementation on objects-as-records, but it's not particularly close to the one that works for objects-as-dictionaries.

Most of the discussion above was based on the assumption that one would want to work with objects-as-dictionaries.  While I certainly do use them as such, for me, they're much more commonly used as records.

I'm in favor of this PR, though -- even with all my debate above -- because in those cases I'd want to filter an object, I'm almost always using it as a dictionary.  I think for those rare cases I'd want to filter a record, I could live with the `toPairs` - `filter` - `fromPairs` shuffle.  If we really want to support that with a dedicated function, we can simply give it a different name.

But we should not use this as an excuse to start thinking about JS Objects as primarily dictionaries.  They have both uses, and I'm guessing that records are more common.

paldepind:
> There is a reasonable filter implementation on objects-as-records, but it's not particularly close to the one that works for objects-as-dictionaries.

I don't understand. What would `filter` mean for records?

> But we should not use this as an excuse to start thinking about JS Objects as primarily dictionaries. They have both uses, and I'm guessing that records are more common.

I agree that objects are most often used as records. Actually I think they are most often used as objects (as we know them from OOP, records with methods). But still see them as dictionaries used as records. Because the underlying data structures is a dictionary.

```js
delete obj.key
if ('key' in obj)
```

Both of the above are things you can do to dictionaries. Not records. Dictionaries can represent records. Not so the other way around.

CrossEye:
Perhaps I'm not communicating well.  The major distinction is whether the values of the object have heterogeneous or homogeneous types.  If they're homogeneous, then a generalization of the `filter` from list makes sense.  If they're heterogeneous, then it's more likely to want a filter that operates on the keys or the key-value pairs.

scott-christopher:
> If we really want to support that with a dedicated function, we can simply give it a different name.

Doesn't `pickBy` already satisfy this?

algesten:
> It would be hard to argue that this is a record rather than a dictionary

didn't put that very well. i was talking about my mental model of objects. obviously i do occasionally use them as dictionaries too.

> But we should not use this as an excuse to start thinking about JS Objects as primarily dictionaries. They have both uses, and I'm guessing that records are more common.

this is exactly my thinking. personally i'd favour the filter predicate having a signature key-value since that makes more sense to me – the hardline fantasyland-functor-stringent typing arguments does nothing for me. but it's not a point worth fighting for more than already been done by @CrossEye.

CrossEye:
>> If we really want to support that with a dedicated function, we can simply give it a different name.

> Doesn't `pickBy` already satisfy this?

Yes, of course.  I was misremembering.

davidchambers:
Let me attempt to summarize the situation.

  - We agree that objects can be used as records (e.g. `{ name :: String, age :: Number }`) or as dictionaries (as in the US state code to state name mapping).

  - We agree that objects are used as records far more frequently than they are used as dictionaries.

  - We agree that dictionaries are functors.

The question is whether `R.filter` should operate on dictionary-like objects. The consensus appears to be that it should, though the usefulness of this behaviour has been called into question.

CrossEye:
> Let me attempt to summarize the situation.

Well done.

joneshf:
Maybe another option is to suggest to people that they use `Map` or something similar when they want the appropriate behavior? I don't believe there's any confusion as to what you want when you have a `Map` (even though the values are heterogeneous). Then we could take a stand and say that objects are records, and this whole issue is solved?

davidchambers:
I like your suggestion, @joneshf.

CrossEye:
> Maybe another option is to suggest to people that they use `Map` or something similar when they want the appropriate behavior?

Certainly.

I don't know that this will satisfy people looking for this behavior, though.

Mostly in Ramda I've been happy to work with the native types as subtly warped versions of useful types.  Arrays only represent lists, and I happily ignore the possibility of sparse arrays.  You supply one, the system could do harmful things.  Oh well.  

While, I have no issue thinking of Objects as mostly records, there is some difference, because everyone does sometimes use them also as dictionaries.  Yes, Maps might eventually supplant them, but it won't be quick, especially with a nice literal syntax for objects.

davidchambers:
This pull request no longer introduces `R.filterObjIndexed`, nor does it deprecate `R.pickBy`. It does, though, define `R.filter` for dictionary-like objects. There's some support for this, and those who use objects as records but never as dictionaries will be unaffected (since `R.pickBy` is still available).

PR #1433, opened 2015-10-07
make R.partial and R.partialRight binary functions

davidchambers:
Let's say goodbye to two more variadic functions.


davidchambers:
I assume there's no opposition to this change. It's just another step in a direction we've already agreed to take.

CrossEye:
Sorry, I've had house-guests and haven't been very active.

This looks good to me.  I'm not thrilled with the signatures, but I don't have any great suggestions.

PR #1434, opened 2015-10-10
add @since annotations

### LINKED ISSUE #1381: include "added in vX.Y.Z" for each function
When working on a project which specifies a Ramda version other than the most recent, one should be able to determine from the documentation whether a particular function is available. This need would be satisfied by ramda/ramda.github.io#10, but I'm interested in seeing each function's "age" regardless.

davidchambers:
Closes #1381

Many thanks to @branneman and @kedashoe for their contributions to this pull request.

This is the script I used to add these annotations:

```
#!/usr/bin/env bash
set -e

next_version=v0.18.0
n=$'\n'

tags() {
  git tag --list --sort=version:refname
}

view() {
  git show "refs/tags/$1:$2"
}

for tag in $(tags) ; do
  view "$tag" "$(node -p "($(view "$tag" package.json)).main")" >"ramda-$tag.js"
done

for name in $(node -p "Object.keys(require('./ramda-$(tags | tail -n 1)')).sort().join('\n')") ; do
  for tag in $(tags) ; do
    if [[ "$(node -p "Object.prototype.hasOwnProperty.call(require('./ramda-$tag'), '$name')")" == true ]] ; then
      echo "R.$name was added in $tag"
      if [[ -e "src/$name.js" ]] ; then
        sed -i "" "s/@memberOf R/@memberOf R\\$n * @since $tag/" "src/$name.js"
      fi
      break
    fi
  done
done

for filename in $(find src -depth 1 -name "*.js" | xargs grep --files-without-match "@since v") ; do
  echo "R.$(basename -s .js "$filename") will be added in $next_version"
  sed -i "" "s/@memberOf R/@memberOf R\\$n * @since $next_version/" "$filename"
done

rm ramda-*.js
```

Do you think it's okay to state `@since v0.1.0` for `R.and`, for example, even though the meaning of `R.and` changed at some point between then and now?

### DIFF
--- a/src/F.js
+++ b/src/F.js
@@ -6,6 +6,7 @@ var always = require('./always');
  *
  * @func
  * @memberOf R
+ * @since v0.9.0
  * @category Function
  * @sig * -> Boolean
  * @param {*}
--- a/src/T.js
+++ b/src/T.js
@@ -6,6 +6,7 @@ var always = require('./always');
  *
  * @func
  * @memberOf R
+ * @since v0.9.0
  * @category Function
  * @sig * -> Boolean
  * @param {*}
--- a/src/__.js
+++ b/src/__.js
@@ -16,6 +16,7 @@
  *
  * @constant
  * @memberOf R
+ * @since v0.6.0
  * @category Function
  * @example
  *
--- a/src/add.js
+++ b/src/add.js
@@ -6,6 +6,7 @@ var _curry2 = require('./internal/_curry2');
  *
  * @func
  * @memberOf R
+ * @since v0.1.0
  * @category Math
  * @sig Number -> Number -> Number
  * @param {Number} a
--- a/src/addIndex.js
+++ b/src/addIndex.js
@@ -15,6 +15,7 @@ var curryN = require('./curryN');
  *
  * @func
  * @memberOf R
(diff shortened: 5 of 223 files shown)


branneman:
Awesome! :+1:

One question though, while I think this step is the most important one, I do not think it actually closes #1381, since the issue states:
> be able to determine from the documentation whether a particular function is available

While the jsdoc contains the new information, the actual documentation on the website probably doesn't update automatically to show this? Or does it?

> Do you think it's okay to state `@since v0.1.0` for `R.and`, for example, even though the meaning of `R.and` changed at some point between then and now?

I think it is technically correct to state this. The [specification of the `@since` tag](http://usejsdoc.org/tags-since.html) mentions availability since a specific version.

I do think we could enhance later on to have additional information for examples like `R.and` as well. But there's more questions to answer. What are the other cases? Are they all of the same nature? What would be the correct spot to put this? The `@since` tag probably doesn't support this format.

davidchambers:
> While the jsdoc contains the new information, the actual documentation on the website probably doesn't update automatically to show this? Or does it?

You're quite right. I'm currently working on overhauling the build process for the website. Once I've simplified the process we should be able to update the template to render this new information.

PR #1435, opened 2015-10-11
use @typedefn rather than @typedef

davidchambers:
[`@typedef`][1] is a real thing. We should use a different name to avoid confusing JSDoc.

[1]: http://usejsdoc.org/tags-typedef.html


CrossEye:
:herb:

PR #1437, opened 2015-10-12
implemented splitAt

niloy:
Implemented `splitAt`.

Works on `Array` and `String`.

Example:
```js
R.splitAt(1, [1, 2, 3]);          //=> [[1], [2, 3]]
R.splitAt(5, 'hello world');      //=> ['hello', ' world']
```

### DIFF
--- a/src/splitAt.js
+++ b/src/splitAt.js
@@ -1,0 +1,25 @@
+var _curry2 = require('./internal/_curry2');
+var length = require('./length');
+var slice = require('./slice');
+
+
+/**
+ * Splits a given list or string at a given index.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig Number -> [a] -> [[a], [a]]
+ * @sig Number -> String -> [String, String]
+ * @param {Number} index The index where the array/string is split.
+ * @param {Array|String} array The array/string to be split.
+ * @return {Array}
+ * @example
+ *
+ *      R.splitAt(1, [1, 2, 3]);          //=> [[1], [2, 3]]
+ *      R.splitAt(5, 'hello world');      //=> ['hello', ' world']
+ *      R.splitAt(-1, 'foobar');          //=> ['fooba', 'r']
+ */
+module.exports = _curry2(function splitAt(index, array) {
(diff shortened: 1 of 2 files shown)


davidchambers:
We should support negative offsets. For example:

```javascript
> R.splitAt(-1, 'foobar')
["fooba", "r"]
```

src/splitAt.js
```diff
@@ -0,0 +1,21 @@
...
+ * @param {Number} a
+ * @param {Array|String} b
+ * @return {Array}
+ * @example
+ *
+ *      R.splitAt(1, [1, 2, 3]);          //=> [[1], [2, 3]]
+ *      R.splitAt(5, 'hello world');      //=> ['hello', ' world']
+ */
+module.exports = _curry2(function splitAt(index, array) {
+  return [array.slice(0, index), array.slice(index)];
```
davidchambers: We should use `R.slice` here to support array-like objects which don't provide a `slice` method.

src/splitAt.js
```diff
@@ -0,0 +1,21 @@
+var _curry2 = require('./internal/_curry2');
+
+
+/**
+ * Splits a given list or string at a given index.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig Number -> [a] -> [[a], [a]]
```
davidchambers: Let's include a second signature for the string case, as we do for `R.take`.

src/splitAt.js
```diff
@@ -0,0 +1,25 @@
...
+ * @sig Number -> String -> [String, String]
+ * @param {Number} a
+ * @param {Array|String} b
```
davidchambers: I believe JSDoc expects a parameter name between `@param` and `{…}`. See one of the other files in __src__ directory for an example.

niloy:
@davidchambers I checked other files, name comes after `{...}`. Added name and description of parameters.

davidchambers:
LGTM. Could you squash the commits?

niloy:
@davidchambers Squashing done.

CrossEye:
:herb:

PR #1438, opened 2015-10-12
Adding component.json

i-am-tom:
Adding component.json to allow, most immediately,
the use of the Duo package manager.

### DIFF
--- a/component.json
+++ b/component.json
@@ -1,0 +1,9 @@
+{
+  "name": "ramda",
+  "author": "Scott Sauyet <scott@sauyet.com> (scott.sauyet.com)",
+  "repository": "ramda/ramda",
+  "description": "A practical functional library for JavaScript programmers.",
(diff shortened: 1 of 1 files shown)


jdalton:
I believe Duo will work without a component.json as it can consume other packagemanifests like bower.json.

i-am-tom:
This doesn't seem to be the case. It doesn't really 'work' with Bower; it just allows for explicit paths to local bower packages (unless I'm doing something wrong!)

jdalton:
\cc @kevva

i-am-tom:
Having done some digging, I've found that Duo use can be achieved by requiring `ramda/ramda:dist/ramda` (because Duo looks for an `index.js` and, in this case, doesn't find one). This isn't exactly a lot of extra effort, so it's probably not something worth doing if no one else has a strong opinion :)

PR #1439, opened 2015-10-12
assert targets of type function are compared with R.identical

syves:
Added two tests for asserting that functions are compared using R.identical. Test names may need editing? Perhaps the adjustments to docs might need tweaking.

### DIFF
--- a/test/indexOf.js
+++ b/test/indexOf.js
@@ -83,4 +83,20 @@ describe('indexOf', function() {
     var curried = R.indexOf(3);
     eq(curried(list), 2);
   });
+
+  it('finds function, compared by identity', function() {
+    var f = function() {};
+    var g = function() {};
+    var list = [g, f, g, f];
+    eq(R.indexOf(f, list), 1);
+  });
+
+  it('does not find function, compared by identity', function() {
+    var f = function() {};
+    var g = function() {};
+    var h = function() {};
+    var list = [g, f];
+    eq(R.indexOf(h, list), -1);
+  });
+
 });
--- a/test/lastIndexOf.js
+++ b/test/lastIndexOf.js
@@ -80,4 +80,20 @@ describe('lastIndexOf', function() {
     var curried = R.lastIndexOf('a');
     eq(curried(list), 2);
   });
+
+  it('finds function, compared by identity', function() {
+    var f = function() {};
+    var g = function() {};
+    var list = [g, f, g, f];
+    eq(R.lastIndexOf(f, list), 3);
+  });
+
+  it('does not find function, compared by identity', function() {
+    var f = function() {};
+    var g = function() {};
+    var h = function() {};
+    var list = [g, f];
+    eq(R.lastIndexOf(h, list), -1);
+  });
+
 });


src/indexOf.js
```diff
@@ -6,7 +6,8 @@ var _isArray = require('./internal/_isArray');
 /**
  * Returns the position of the first occurrence of an item in an array,
  * or -1 if the item is not included in the array. `R.equals` is used to
- * determine equality.
+ * determine equality in most cases. When target is a function `R.identical`
+ * is used to determine if the target is equal to an element in the array.
```
davidchambers: This isn't quite right. `R.equals` is used to determine equality in all cases, but `R.equals` itself compares functions by identity. For example:

```javascript
var f = function() {};

R.equals(f, f);  // => true
R.equals(function() {}, function() {});  // => false
```

Perhaps we should clarify the description of `R.equals` instead.

test/indexOf.js
```diff
@@ -83,4 +83,19 @@ describe('indexOf', function() {
...
+    var f = function() {};
+    var g = function() {};
+    var list = [g, f]
+    eq(R.indexOf(f, list), 1);
+  });
+
+    it('does not find function, compared by identity', function() {
+    var f = function() {};
+    var g = function() {};
+    var e = function() {};
```
davidchambers: It's conventional to name the third function `h`. ;)

test/indexOf.js
```diff
@@ -83,4 +83,19 @@ describe('indexOf', function() {
...
   });
+
+  it('finds function, compared by identity', function() {
+    var f = function() {};
+    var g = function() {};
+    var list = [g, f]
+    eq(R.indexOf(f, list), 1);
+  });
+
+    it('does not find function, compared by identity', function() {
```
davidchambers: :arrow_left: :arrow_left:

davidchambers:
The tests for `R.indexOf` look good. Could you add similar tests for `R.lastIndexOf`?

test/indexOf.js
syves: you are so conventional ;)

test/indexOf.js
syves: what does arrow arrow mean?

syves:
added tests for lastIndexOf

test/indexOf.js
scott-christopher: I believe @davidchambers was suggesting the indentation is not aligned.

davidchambers: Thanks for translating from Emoji to English, @scott-christopher. :)

src/indexOf.js
```diff
@@ -21,6 +21,12 @@ var _isArray = require('./internal/_isArray');
  *
  *      R.indexOf(3, [1,2,3,4]); //=> 2
  *      R.indexOf(10, [1,2,3,4]); //=> -1
+ *
+ *      var f = function() {};
+ *      var g = function() {};
+ *      var h = fucntion() {};
```
davidchambers: `s/fucntion/function/`

I wonder, though, whether we should include these examples on the website. They're very much worth including in the test suite, but the examples on the website should demonstrate common usage. Finding the index of a function in a list of functions is not a common operation, I would say. What do others think?

test/indexOf.js
```diff
@@ -83,4 +83,19 @@ describe('indexOf', function() {
...
+    var f = function() {};
+    var g = function() {};
+    var h = function() {};
+    var list = [g, f];
+    eq(R.indexOf(h, list), -1);
+  });
 });
```
davidchambers: Could you include an empty line between lines 100 and 101? :)

src/indexOf.js
syves: I'll default to you on this.

syves:
@davidchambers can I have a tree?

src/indexOf.js
```diff
@@ -21,6 +21,7 @@ var _isArray = require('./internal/_isArray');
  *
  *      R.indexOf(3, [1,2,3,4]); //=> 2
  *      R.indexOf(10, [1,2,3,4]); //=> -1
+ *
```
davidchambers: :space_invader:

test/indexOf.js
```diff
@@ -83,4 +83,20 @@ describe('indexOf', function() {
     var curried = R.indexOf(3);
     eq(curried(list), 2);
   });
+
+  it('finds function, compared by identity', function() {
+    var f = function() {};
+    var g = function() {};
+    var list = [g, f];
```
davidchambers: You could make this `[g, f, g, f]` (as you do in the equivalent test for `R.lastIndexOf`) to assert that the first occurrence is matched.

davidchambers:
> can I have a tree?

I left two comments about minor things. Address these and a `:deciduous_tree:` will be forthcoming. :)

src/indexOf.js
syves: @davidchambers does octopus mean remove space? Is there a guide to ramda emoji?

davidchambers: There's no guide, I'm sorry! Hover over the emoji, though, and you will see its name – `:space_invader:`, in this case. ;)

davidchambers:
:deciduous_tree:

syves:
yay!

PR #1442, opened 2015-10-13
make R.equals return true for equivalent Error objects

davidchambers:
`R.equals` will never be *done*. :stuck_out_tongue_closed_eyes:


CrossEye:
:herb:

PR #1443, opened 2015-10-14
Adds mergeWith and mergeWithKey

scott-christopher:
Given the recent discussions again in gitter about merging objects, I thought it might help to throw this up as a concrete example to discuss, if nothing else.

Two new public functions are introduced to allow a custom merging strategy to be provided for merging objects with intersecting keys:
```
// provided with values of each object only
mergeWith :: (a -> a -> a) -> {a} -> {a}
// provided with the key along with the values of each object
mergeWithKey :: (String -> a -> a -> a) -> {a} -> {a} -> {a}
```

The main points of potential contention with this implementation:
* ~~`internal/_mergeWith` provides the non-intersecting keys as objects to the provided `fnL` and `fnR` functions (e.g. `fnL :: {a} -> {a}`) to allow filtering of the resulting object – should these instead operate on individual key/pairs like `fn` (e.g. `(String, a) -> a`)? Doing so would prevent reuse for potential hypothetical functions like `mergeIntersection`.~~

* ~~to support filtering, the implementation currently excludes keys from the resulting object whose values are `undefined` – this feels like it overloads the behaviour of `mergeWith`/`mergeWithKeys` a little.~~

_edit_: I've since removed `internal/_mergeWith`, moving the implementation into `mergeWithKeys` and I've also removed the behaviour of filtering `undefined` values.

### DIFF
--- a/src/mergeWith.js
+++ b/src/mergeWith.js
@@ -1,0 +1,32 @@
+var _curry3 = require('./internal/_curry3');
+var mergeWithKey = require('./mergeWithKey');
+
+
+/**
+ * Creates a new object with the own properties of the two provided objects. If
+ * a key exists in both objects, the provided function is applied to the values
+ * associated with the key in each object, with the result being used as the
+ * value associated with the key in the returned object. The key will be
+ * excluded from the returned object if the resulting value is `undefined`.
+ *
+ * @func
+ * @memberOf R
+ * @category Object
+ * @sig (a -> a -> a) -> {a} -> {a} -> {a}
+ * @param {Function} fn
+ * @param {Object} l
+ * @param {Object} r
+ * @return {Object}
+ * @see R.merge, R.mergeWithKey
+ * @example
+ *
+ *      R.mergeWith(R.concat,
+ *                  { a: true, values: [10, 20] },
+ *                  { b: true, values: [15, 35] });
+ *      //=> { a: true, b: true, values: [10, 20, 15, 35] }
+ */
+module.exports = _curry3(function mergeWith(fn, l, r) {
+  return mergeWithKey(function(_, _l, _r) {
+    return fn(_l, _r);
+  }, l, r);
+});
--- a/src/mergeWithKey.js
+++ b/src/mergeWithKey.js
@@ -1,0 +1,47 @@
+var _curry3 = require('./internal/_curry3');
+var _has = require('./internal/_has');
+
+
+/**
+ * Creates a new object with the own properties of the two provided objects. If
+ * a key exists in both objects, the provided function is applied to the key
+ * and the values associated with the key in each object, with the result being
+ * used as the value associated with the key in the returned object. The key
+ * will be excluded from the returned object if the resulting value is
+ * `undefined`.
+ *
+ * @func
+ * @memberOf R
+ * @category Object
+ * @sig (String -> a -> a -> a) -> {a} -> {a} -> {a}
+ * @param {Function} fn
+ * @param {Object} l
+ * @param {Object} r
+ * @return {Object}
+ * @see R.merge, R.mergeWith
+ * @example
+ *
+ *      let concatValues = (k, l, r) => k == 'values' ? R.concat(l, r) : r
+ *      R.mergeWithKey(concatValues,
+ *                     { a: true, thing: 'foo', values: [10, 20] },
+ *                     { b: true, thing: 'bar', values: [15, 35] });
+ *      //=> { a: true, b: true, thing: 'bar', values: [10, 20, 15, 35] }
+ */
+module.exports = _curry3(function mergeWithKey(fn, l, r) {
+  var result = {};
+  var k;
+
+  for (k in l) {
+    if (_has(k, l)) {
+      result[k] = _has(k, r) ? fn(k, l[k], r[k]) : l[k];
+    }
+  }
+
+  for (k in r) {
+    if (_has(k, r) && !(_has(k, result))) {
+      result[k] = r[k];
+    }
+  }
+
+  return result;
+});
--- a/test/mergeWith.js
+++ b/test/mergeWith.js
@@ -1,0 +1,43 @@
+var assert = require('assert');
+
+var R = require('..');
+var eq = require('./shared/eq');
+
+
+describe('mergeWith', function() {
+  function last(x, y) { return y; }
+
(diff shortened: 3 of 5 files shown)


joneshf:
:+1: This is definitely needed!  Thanks!

* I think giving the whole function is good. The caller can iterate the object in their own way if that's what they want.

* I don't think this function should remove undefined values. It looks like it comes from these lines: https://github.com/scott-christopher/ramda/blob/merge-with/src/internal/_mergeWith.js#L46-L56. Could the resulting objects just be merged instead of filtering? Then the onus is to the caller to filter with a function like `omit`, `pick`, or something else.
If the behavior is wanted, another function could be made that does what you're doing there, and given a new name to keep things straight. However, I don't think this function should make that choice.

test/mergeWith.js
```diff
@@ -0,0 +1,50 @@
...
+    eq(res, {w: 1, x: 5});
+  });
+
+  it('reports only own properties', function() {
+    var a = {w: 1, x: 2};
+    function Cla() {}
+    Cla.prototype.x = 5;
+    eq(R.mergeWith(last, new Cla(), a), {w: 1, x: 2});
+    eq(R.mergeWith(last, a, new Cla()), {w: 1, x: 2});
+  });
```
CrossEye: Would probably also be useful to have a test/demo which shows what happens with `mergeWith(last, new Cla(), {w: 1})` (i.e. without the `x` property.)

CrossEye:
:herb:

I haven't been on Gitter much lately, and haven't seen these discussions, but it's certainly come up before.  Our attempt at a deep merge eventually punted on lists as we really needed hints from the user about how to go about it, and that API didn't allow them.  This makes it explicit, and more general.

> I don't think this function should remove undefined values.

I agree.

davidchambers:
I like it!

src/mergeWith.js
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Creates a new object with the own properties of the two provided objects. If
+ * a key exists in both objects, the provided function is applied to the values
+ * associated with the key in each object, with the result being used as the
+ * value associated with the key in the returned object. The key will be
+ * excluded from the returned object if the resulting value is `undefined`.
+ *
+ * @func
+ * @memberOf R
+ * @since x.y.z
```
scott-christopher: Do we have a convention around the value of `@since` for new functions yet?

test/mergeWith.js
scott-christopher: Agreed. I've pushed up a change to include this scenario.

scott-christopher:
If we don't want the ability to filter in the internal `_mergeWith`, then I'd suggest `_mergeWith` should be removed and its implementation moved into `mergeWithKey`, as there isn't a whole lot more that it offers.

src/mergeWith.js
davidchambers: No. I see two options:

  - specify the next minor version; or
  - leave it out and rely on a prepublish script to add it.

The first option is easy to understand, but will potentially require a much-debated pull request to be updated several times as releases are made while it is waiting to be merged.

The second option is arguably more reliable, but I haven't written the script yet. ;)

src/internal/_mergeWith.js
```diff
@@ -0,0 +1,59 @@
...
+module.exports = curryN(5, function _mergeWith(fn, fnL, fnR, l, r) {
+  var lNotR = {};
+  var rNotL = {};
+  var skipped = {};
+  var result = {};
+  var k, v, lMerged, rMerged;
+
+  for (k in l) {
+    if (l.hasOwnProperty(k)) {
+      if (r.hasOwnProperty(k)) {
```
davidchambers: We should use `Object.prototype.hasOwnProperty.call` (or `_has`) to handle objects with own `hasOwnProperty` properties. ;)

src/mergeWith.js
scott-christopher: I opted for the second option for the exact the reason you listed :)

src/internal/_mergeWith.js
scott-christopher: :zap:

scott-christopher:
I've just updated the implementation to no longer remove `undefined` values when merging.

CrossEye:
It looks as though there are some conflicts here... But once those are cleaned up, anyone have objections to merging this?

davidchambers:
These functions seem useful. I'm in favour of merging this pull request. :)

buzzdecafe:
:cow2:

scott-christopher:
I've rebased onto master, so it should be good to go now.

One question prior to merging: Should we remove the `@since` tag for new functions and then add them prior to release, or mark them as `@since x.y.z`/`@since HEAD` which would allow for some script to run with the release script to update it automatically?

davidchambers:
Let's omit the `@since` annotation. We can reuse the script I wrote for #1434 to add the missing annotations as part of `make release-[major|minor|patch]`.

scott-christopher:
* `@since` removed

src/mergeWith.js
```diff
@@ -0,0 +1,32 @@
...
+ * value associated with the key in the returned object. The key will be
+ * excluded from the returned object if the resulting value is `undefined`.
+ *
+ * @func
+ * @memberOf R
+ * @category Object
+ * @sig (a -> a -> a) -> {a} -> {a} -> {a}
+ * @param {Function} fn
+ * @param {Object} lObj
+ * @param {Object} rObj
```
davidchambers: We use `l` and `r` in `R.merge`. Should we use the same parameter names here?

scott-christopher: * updated

src/mergeWithKey.js
```diff
@@ -0,0 +1,47 @@
...
+  var k;
+
+  for (k in l) {
+    if (_has(k, l)) {
+      result[k] = _has(k, r) ? fn(k, l[k], r[k]) : l[k];
+    }
+  }
+
+  for (k in r) {
+    if (_has(k, r) && !(k in result)) {
```
davidchambers: We should use `_has(k, result)` rather than `k in result`, to avoid getting fooled by inherited properties (`'toString' in {}` evaluates `true`, for example).

scott-christopher: * updated

davidchambers:
:deciduous_tree:

ccorcos:
Awesome! I totally made this function myself a few days ago. :+1:

Issue #1301, opened 2015-07-22
Add a list XOR function


trbrc:
There is currently no simple way to do an XOR between two lists, i.e. take two lists and return a list of the elements that occur in only one of the lists. Like this:
```javascript
R.xor([0, 1, 2, 3, 4], [6, 4, 2]) // => [0, 1, 3, 6]
```
Lodash [has a method like this](https://lodash.com/docs#xor).

As suggested by @davidchambers on Gitter, here's a way to do it currently:

```javascript
const xor = R.converge(R.difference, R.union, R.intersection);
```
But that's unlikely to be efficient.

CrossEye:
:+1: for the idea.  We can play around with implementations.  And of course we'd probably want a `xorWith` version as well.

Here's an alternative, just for comparison.  I haven't thought at all about what would be most efficient:

```js
const xor2 = R.converge(R.union, R.difference, R.flip(R.difference));
```

davidchambers:
I'm in favour of this function. `R.xor` is not a good name for it, though, as we already have `R.or` which does something entirely different. I suggest `R.symmetricDifference` or `R.symDifference`.

buzzdecafe:
once again i agree with @davidchambers . fins with the function, but not the name. to my ear, `xor` is a boolean or bitwise operation.

CrossEye:
`diffɟɟıp` :smile:

PR #1448, opened 2015-10-17
Add symmetricDifference, symmetricDifferenceWith

jimf:
Add `R.symmetricDifference` and `R.symmetricDifferenceWith` for finding the [symmetric difference](https://en.wikipedia.org/wiki/Symmetric_difference) between two lists.

Addresses #1301

If interested, perf results for 3 different implementations:
- http://jsperf.com/ramda-r-symmetricdifference-candidates/2

### DIFF
--- a/src/symmetricDifference.js
+++ b/src/symmetricDifference.js
@@ -1,0 +1,25 @@
+var _curry2 = require('./internal/_curry2');
+var concat = require('./concat');
+var difference = require('./difference');
+
+
+/**
+ * Finds the set (i.e. no duplicates) of all elements contained in the first or
+ * second list, but not both.
+ *
+ * @func
+ * @memberOf R
+ * @category Relation
+ * @sig [*] -> [*] -> [*]
+ * @param {Array} list1 The first list.
+ * @param {Array} list2 The second list.
+ * @return {Array} The elements in `list1` or `list2`, but not both.
+ * @see R.symmetricDifferenceWith
+ * @example
+ *
+ *      R.symmetricDifference([1,2,3,4], [7,6,5,4,3]); //=> [1,2,7,6,5]
+ *      R.symmetricDifference([7,6,5,4,3], [1,2,3,4]); //=> [7,6,5,1,2]
+ */
+module.exports = _curry2(function symmetricDifference(list1, list2) {
+  return concat(difference(list1, list2), difference(list2, list1));
+});
--- a/src/symmetricDifferenceWith.js
+++ b/src/symmetricDifferenceWith.js
@@ -1,0 +1,29 @@
+var _curry3 = require('./internal/_curry3');
+var concat = require('./concat');
+var differenceWith = require('./differenceWith');
+
+
+/**
+ * Finds the set (i.e. no duplicates) of all elements contained in the first or
+ * second list, but not both. Duplication is determined according to the value
+ * returned by applying the supplied predicate to two list elements.
+ *
+ * @func
+ * @memberOf R
+ * @category Relation
+ * @sig (a -> a -> Boolean) -> [a] -> [a] -> [a]
+ * @param {Function} pred A predicate used to test whether two items are equal.
+ * @param {Array} list1 The first list.
+ * @param {Array} list2 The second list.
+ * @return {Array} The elements in `list1` or `list2`, but not both.
+ * @see R.symmetricDifference
+ * @example
+ *
+ *      var eqA = R.eqBy(R.prop('a'));
+ *      var l1 = [{a: 1}, {a: 2}, {a: 3}, {a: 4}];
+ *      var l2 = [{a: 3}, {a: 4}, {a: 5}, {a: 6}];
+ *      R.symmetricDifferenceWith(eqA, l1, l2); //=> [{a: 1}, {a: 2}, {a: 5}, {a: 6}]
+ */
+module.exports = _curry3(function symmetricDifferenceWith(pred, list1, list2) {
+  return concat(differenceWith(pred, list1, list2), differenceWith(pred, list2, list1));
+});
--- a/test/symmetricDifference.js
+++ b/test/symmetricDifference.js
@@ -1,0 +1,52 @@
+var R = require('..');
+var eq = require('./shared/eq');
+
+
+describe('symmetricDifference', function() {
+  var M = [1, 2, 3, 4];
+  var M2 = [1, 2, 3, 4, 1, 2, 3, 4];
+  var N = [3, 4, 5, 6];
+  var N2 = [3, 3, 4, 4, 5, 5, 6, 6];
+  var Z = [3, 4, 5, 6, 10];
+  var Z2 = [1, 2, 3, 3, 4, 4, 5, 5, 6, 6, 7, 8];
+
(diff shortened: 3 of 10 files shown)


src/symmetricDifferenceWith.js
```diff
@@ -0,0 +1,31 @@
...
+var converge = require('./converge');
+var differenceWith = require('./differenceWith');
+var flip = require('./flip');
+var unionWith = require('./unionWith');
+
+
+/**
+ * Finds the set (i.e. no duplicates) of all elements contained in the first or second list, but not both.
+ * Duplication is determined according to the value returned by applying the supplied predicate to two list
+ * elements.
```
davidchambers: Let's wrap at 79 characters, to keep the neckbeards happy. ;)

davidchambers:
I'm pleased to see these functions!

src/symmetricDifferenceWith.js
jimf: @davidchambers Will do. Just so I'm clear, you're only talking about the docstrings, correct?

davidchambers: Yes. In my own projects I enforce a strict 79-character limit for all lines, but I don't think Ramda should go this far. Wrapping prose is straightforward and non-controversial, though. :)

jimf:
:droplet:

CrossEye:
:herb:

src/symmetricDifference.js
```diff
@@ -0,0 +1,27 @@
...
+ * @param {Array} list2 The second list.
+ * @return {Array} The elements in `list1` or `list2`, but not both.
+ * @see R.symmetricDifferenceWith
+ * @example
+ *
+ *      R.symmetricDifference([1,2,3,4], [7,6,5,4,3]); //=> [1,2,7,6,5]
+ *      R.symmetricDifference([7,6,5,4,3], [1,2,3,4]); //=> [7,6,5,1,2]
+ */
+module.exports = _curry2(function symmetricDifference(list1, list2) {
+  return converge(union, [difference, flip(difference)])(list1, list2);
```
davidchambers: I'd prefer not to use `converge` here. Could we use the expanded form instead?

```javascript
union(difference(list1, list2), difference(list2, list1))
```

CrossEye: Agreed.  Explicit is good!

jimf:
@davidchambers @CrossEye updated. Good suggestion!

src/symmetricDifference.js
```diff
@@ -0,0 +1,25 @@
...
+
+
+/**
+ * Finds the set (i.e. no duplicates) of all elements contained in the first or
+ * second list, but not both.
+ *
+ * @func
+ * @memberOf R
+ * @category Relation
+ * @sig [a] -> [a] -> [a]
```
davidchambers: Perhaps this should be `[*] -> [*] -> [*]`, since there's no requirement that the lists' values be of a single type.

src/symmetricDifference.js
```diff
@@ -0,0 +1,25 @@
...
+ * @param {Array} list2 The second list.
+ * @return {Array} The elements in `list1` or `list2`, but not both.
+ * @see R.symmetricDifferenceWith
+ * @example
+ *
+ *      R.symmetricDifference([1,2,3,4], [7,6,5,4,3]); //=> [1,2,7,6,5]
+ *      R.symmetricDifference([7,6,5,4,3], [1,2,3,4]); //=> [7,6,5,1,2]
+ */
+module.exports = _curry2(function symmetricDifference(list1, list2) {
+  return union(difference(list1, list2), difference(list2, list1));
```
davidchambers: Could we use `concat` here in place of `union`?

src/symmetricDifferenceWith.js
```diff
@@ -0,0 +1,29 @@
+var _curry3 = require('./internal/_curry3');
+var differenceWith = require('./differenceWith');
+var unionWith = require('./unionWith');
+
+
+/**
+ * Finds the set (i.e. no duplicates) of all elements contained in the first or
+ * second list, but not both.  Duplication is determined according to the value
```
davidchambers: Nit: Single space after full stop (`.`), please. :)

src/symmetricDifference.js
jimf: Makes sense. Should I update difference/intersection/union as well for consistency?

src/symmetricDifferenceWith.js
```diff
@@ -0,0 +1,29 @@
...
+ * returned by applying the supplied predicate to two list elements.
+ *
+ * @func
+ * @memberOf R
+ * @category Relation
+ * @sig (a,a -> Boolean) -> [a] -> [a] -> [a]
```
davidchambers: I'd prefer to write `a -> a -> Boolean`. As written, it's unclear whether the `,` or the `->` binds more tightly.

test/symmetricDifference.js
```diff
@@ -0,0 +1,51 @@
...
+  it('returns an empty array if there are no different elements', function() {
+    eq(R.symmetricDifference(M2, M), []);
+    eq(R.symmetricDifference(M, M2), []);
+  });
+
+  it('is curried', function() {
+    eq(typeof R.symmetricDifference([1, 2, 3]), 'function');
+    eq(R.symmetricDifference([1, 2, 3])([1, 3]), [2]);
+  });
+});
```
davidchambers: Nit: Empty line between lines 50 and 51, please.

src/symmetricDifferenceWith.js
jimf: Word. There's a lot of this in the existing docstrings. I can fix those as well if preferred.

src/symmetricDifference.js
davidchambers: That sounds like a good idea to me.

src/symmetricDifferenceWith.js
jimf: +1. I'll update the related sigs as well.

src/symmetricDifferenceWith.js
davidchambers: I suggest leaving that for a separate pull request. Also, let's see whether @CrossEye prefers one space or two. I switched from two spaces to one about ten years ago and haven't looked back. ;)

src/symmetricDifference.js
jimf: Yup. Will change.

src/symmetricDifferenceWith.js
CrossEye: I never switched, but don't really care at all.

src/symmetricDifference.js
CrossEye: Obviously we should use `concat` for performance reasons, but it seems so much less elegant that way.  :frowning:

jimf:
@davidchambers @CrossEye updated. Let me know if you'd like those commits squashed.

davidchambers:
Yes, please squash the commits. :)

jimf:
Squashed!

CrossEye:
Thanks, @jimf!

jimf:
:beers:

ccorcos:
Sweet! I was wondering what this was called... https://github.com/ramda/ramda/issues/1532

PR #1450, opened 2015-10-18
Corrected dispatching docstrings

branneman:
Fixes:

- [`R.clone()`](http://ramdajs.com/docs/#clone) has a different codestyle for the 'dispatches to...' docstring
- [`R.equals()`](http://ramdajs.com/docs/#equals) has a duplicate sentence

### DIFF
--- a/src/equals.js
+++ b/src/equals.js
@@ -4,10 +4,9 @@ var _equals = require('./internal/_equals');
 
 /**
  * Returns `true` if its arguments are equivalent, `false` otherwise.
- * Dispatches to an `equals` method if present. Handles cyclical data
- * structures.
+ * Handles cyclical data structures.
  *
- * Dispatches to the `equals` method of both arguments, if present.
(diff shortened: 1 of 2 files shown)


src/equals.js
```diff
@@ -4,8 +4,7 @@ var _equals = require('./internal/_equals');
 
 /**
  * Returns `true` if its arguments are equivalent, `false` otherwise.
- * Dispatches to an `equals` method if present. Handles cyclical data
- * structures.
+ * Handles cyclical data structures.
  *
  * Dispatches to the `equals` method of both arguments, if present.
```
CrossEye: This might (or quite well might not!) be clearer if it included the word "symmetrically".

    * Dispatches symmetrically to the `equals` methods of both arguments, if present

Either way, it should definitely read "method**s**".

branneman:
@CrossEye: Updated! I think it makes sense to say there's something special going on here.

CrossEye:
LGTM

Issue #1069, opened 2015-05-06
zipN


ccorcos:
`zipN` is zip, but with N inputs are arguments and the first argument being N.

CrossEye:
If we do anything like this (and I think there's a good case to be made for this one and one for zipWith) then I'd rather it simply accept a list of lists to zip together.

buzzdecafe:
i agree, rather than a variadic function, i'd rather see `zipAll([[xs], [ys], [zs], ...])` and `zipAllWith(f, [[xs], [ys], [zs], ...])`

ccorcos:
I totally agree.  I think transpose if actually a better name for it... @CrossEye wrote this in another issue ticket:

```coffee
transpose = (xy) ->
  R.mapIndexed(R.pipe(R.nthArg(1), R.nth, R.map(R.__, xy)), R.head(xy))
```

CrossEye:
The name is an interesting question.  As an extension of 'zip', clearly 'zipAll' would be most obvious.  If we think of these arrays as matrices, 'transpose' is clearly better.

ccorcos:
definitely...

buzzdecafe:
4 weeks with no activity, so closing without prejudice. if you would like to submit a PR on this please do so

PR #1452, opened 2015-10-20
adds zipAll method with tests

ConorLinehan:
Adds `R.zipAll` for using `zip` on N number of lists.

Addresses #1069

### DIFF
--- a/src/zipAll.js
+++ b/src/zipAll.js
@@ -8,12 +8,12 @@ var _curry1 = require('./internal/_curry1');
 * @func
 * @memberOf R
 * @category List
-* @sig [a] -> [a]
+* @sig [[*]] -> [[*]]
 * @param {Array} List of arrays to consider
 * @return {Array} The list made by paring up same indexed elements of supplied list.
 * @example
 *
-* R.zipAll([[1, 2, 3], ['a', 'b', 'c'], ['A', 'B', 'C']]) //=> [[1, 'a', 'A'], [2, 'b', 'B'], [3, 'c', 'C']]
+* R.zipAll([[1, 2, 3], ['a', 'b', 'c'], [true, true, false]]) //=> [[1, 'a', true], [2, 'b', true], [3, 'c', false]]
 */
 module.exports = _curry1(function zipAll(list) {
   var rv = [];
--- a/src/zipAll.js
+++ b/src/zipAll.js
@@ -18,7 +18,7 @@ var _curry1 = require('./internal/_curry1');
 module.exports = _curry1(function zipAll(list) {
   var rv = [];
   var idx = list.length - 1;
-  var len = list[0].length || 0;
+  var len = list[0].length;
   while (list[idx]) {
     len = Math.min(len, list[idx].length);
     idx -= 1;
--- a/src/zipAll.js
+++ b/src/zipAll.js
@@ -1,0 +1,40 @@
+var _curry1 = require('./internal/_curry1');
+
+/**
+* Creates a new list out of the lists supplied by pairing up
+* equally positioned items from both lists. The returned list is
+* truncated to the length of the shorter of the two input lists.
+*
+* @func
+* @memberOf R
+* @since v.0.18
+* @category List
+* @sig [a] -> [a]
+* @param {Array} List of arrays to consider
+* @return {Array} The list made by paring up same indexed elements of supplied list.
+* @example
+*
+* R.zipAll([[1, 2, 3], ['a', 'b', 'c'], ['A', 'B', 'C']]) //=> [[1, 'a', 'A'], [2, 'b', 'B'], [3, 'c', 'C']]
+*/
+module.exports = _curry1(function zipAll(list) {
+  var rv = [];
+  var idx = list.length - 1;
+  var len = list[0].length || 0;
+  while (list[idx]) {
+    len = Math.min(len, list[idx].length);
+    idx -= 1;
+  }
+
+  idx = 0;
+  var j = 0;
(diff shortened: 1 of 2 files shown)


src/zipAll.js
```diff
@@ -0,0 +1,40 @@
+var _curry1 = require('./internal/_curry1');
+
+/**
+* Creates a new list out of the lists supplied by pairing up
+* equally positioned items from both lists. The returned list is
+* truncated to the length of the shorter of the two input lists.
```
raine: > **two** input lists.

That doesn't sound right if this function takes n lists.

test/zipAll.js
```diff
@@ -0,0 +1,15 @@
...
+  it('returns an array of tuples', function() {
+    var arr = [[1, 2, 3], [10, 20, 30], [100, 200, 300]];
+    eq(R.zipAll(arr), [[1, 10, 100], [2, 20, 200], [3, 30, 300]]);
+  });
+
+  it('returns a list as long as the shorter of the lists input', function() {
+    var arr = [[1, 2, 3], [4, 5, 6, 7], [8, 9, 10, 11, 12]];
+    eq(R.zipAll(arr), [[1, 4, 8], [2, 5, 9], [3, 6, 10]]);
+  });
+
```
cullophid: :+1:  I was literally just about to create a PR for this.
This is essentially a transpose function for matrices, which is very useful.

src/zipAll.js
cullophid: it doesn't it takes a list of n lists (matrix).

This is a transpose function for matrices, but for ramda naming convention its named zipAll
:+1:

cullophid:
I was literally just about to commit a PR for this.
Nice work :+1:

ConorLinehan:
@cullophid Thanks :smile: .

@raine Yeah your right I take it you were on about the use of `pairing`?
How about this instead?
```
* Creates a new list out of the lists supplied by matching
* equally positioned items from both lists. The returned list is
* truncated to the length of the shorter of the two input lists.
```

cullophid:
@conorLinehan  Alternatively you can describe it as a matrix transpose operation where each row is swapped  with its corresponding column

ConorLinehan:
@cullophid I suppose my only worry about calling it transpose is a transpose on a matrix should have the property (A<sup>T</sup>)<sup>T</sup> = A while this won't for differing length lists? @CrossEye mentions naming in the issue, if we prohibit uneven length lists then I think `R.transpose` is better, while if we allow them and truncate like `R.zip` then `R.zipAll` would be preferable IMO ?

src/zipAll.js
```diff
@@ -0,0 +1,40 @@
...
+*
+* @func
+* @memberOf R
+* @since v.0.18
```
davidchambers: Let's remove this line.

src/zipAll.js
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+* Creates a new list out of the lists supplied by matching
+* equally positioned items from both lists. The returned list is
+* truncated to the length of the shorter of the two input lists.
+*
+* @func
+* @memberOf R
+* @category List
+* @sig [a] -> [a]
```
davidchambers: Should this be `[[a]] -> [[a]]`?

src/zipAll.js
```diff
@@ -0,0 +1,39 @@
...
+* @param {Array} List of arrays to consider
+* @return {Array} The list made by paring up same indexed elements of supplied list.
+* @example
+*
+* R.zipAll([[1, 2, 3], ['a', 'b', 'c'], ['A', 'B', 'C']]) //=> [[1, 'a', 'A'], [2, 'b', 'B'], [3, 'c', 'C']]
```
davidchambers: I'd prefer not to mix types in the example. I suggest this instead:

```javascript
R.zipAll([[1, 2], [3, 4], [5, 6], [7, 8]]) //=> [[1, 3, 5, 7], [2, 4, 6, 8]]
```

Also, the `@example` block should be indented as in other files.

CrossEye: > I'd prefer not to mix types in the example.

I disagree.  I think that part of the point of this function is that it can be used to create arbitrary tuples.  I would actually prefer to have the third list contain something like `[true, true, false]`

src/zipAll.js
CrossEye: Actually `[[*]] -> [[*]]`

I don't think there's any way in the type signature to denote that the lists ideally should have the same length, though.

src/zipAll.js
davidchambers: I'm fine with this if we change the signature to `[[*]] -> [[*]]` as you suggest.

kedashoe:
Does this implementation give us what we want if one of the lists is of length 0?

```js
R.zipAll([[1], [3, 4], [5, 6]]) //=> [[1], [2], [3]]
R.zipAll([[], [3, 4], [5, 6]]) //=> []
```

Also, what does the `||` do for us in `var len = list[0].length || 0;`? If `list[0]` is not a proper list, don't we want everything to blow up?

buzzdecafe:
> `R.zipAll([[], [3, 4], [5, 6]]) //=> []`

IMO that is correct behavior. We should have a test to show that.

> Also, what does the || do for us in `var len = list[0].length || 0;`? If `list[0]` is not a proper list, don't we want everything to blow up?

I agree, that should be a show-stopper.

CrossEye:
Closing as stale -- the main case is covered by `transpose`.  If we want the truncation aspects, we can revisit later.

PR #1453, opened 2015-10-20
Add pathSatisfies

gilligan:
pathSatisfies works analogous to propSatisfies but applies
the predicator on a path instead of a property name.

### DIFF
--- a/src/pathSatisfies.js
+++ b/src/pathSatisfies.js
@@ -1,0 +1,24 @@
+var _curry3 = require('./internal/_curry3');
+var path = require('./path');
+
+
+/**
+ * Returns `true` if the specified object property at given path satisfies the
+ * given predicate; `false` otherwise.
+ *
+ * @func
+ * @memberOf R
+ * @category Logic
+ * @sig (a -> Boolean) -> [String] -> Object -> Boolean
+ * @param {Function} pred
+ * @param {Array} propPath
+ * @param {*} obj
+ * @return {Boolean}
+ * @see R.propSatisfies, R.path
+ * @example
+ *
+ *      R.pathSatisfies(y => y > 0, ['x', 'y'], {x: {y: 2}}); //=> true
+ */
+module.exports = _curry3(function pathSatisfies(pred, propPath, obj) {
+  return propPath.length > 0 && pred(path(propPath, obj));
+});
--- a/test/pathSatisfies.js
+++ b/test/pathSatisfies.js
@@ -1,0 +1,25 @@
+var R = require('..');
+var eq = require('./shared/eq');
+
+
+describe('pathSatisfies', function() {
+
+  var isPositive = function(n) { return n > 0; };
+
+  it('returns true if the specified object path satisfies the given predicate', function() {
+    eq(R.pathSatisfies(isPositive, ['x', 'y'], {x: {y: 1}}), true);
+  });
+
+  it('returns false if the specified path does not exist', function() {
+    eq(R.pathSatisfies(isPositive, ['x', 'y'], {x: {z: 42}}), false);
+  });
+
+  it('returns false if the path is empty', function() {
+    eq(R.pathSatisfies(isPositive, [], {x: {z: 42}}), false);
+  });
+
+  it('returns false otherwise', function() {
+    eq(R.pathSatisfies(isPositive, ['x', 'y'], {x: {y: 0}}), false);
+  });
+
+});


test/pathSatisfies.js
```diff
@@ -0,0 +1,17 @@
...
+  var isPositive = function(n) { return n > 0; };
+
+  it('returns true if the specified object path satisfies the given predicate', function() {
+    eq(R.pathSatisfies(isPositive, ['x', 'y'], {x: {y: 1}}), true);
+  });
+
+  it('returns false otherwise', function() {
+    eq(R.pathSatisfies(isPositive, ['x', 'y'], {x: {y: 0}}), false);
+  });
+
```
CrossEye: I think this could use another test demonstrating the behavior on

    R.pathSatisfies(isPositive, ['x', 'y'], {x: {z: 42}}

CrossEye:
This looks reasonable to me.  I think `path` and `prop` are clearly parallel in most everything.

The only question is how far we want to carry this, and how important people find such a function.  Is it useful and important enough for the core, or is it more suited for the Cookbook?

gilligan:
@CrossEye I added it to our code base as a utility function early on and I thought it would make sense to have it as built-in function in Ramda.

Of course you could then begin to wonder if there should also be a `lensPath` analogous to `lensProp` ...

test/pathSatisfies.js
gilligan: Indeed. Will add it.

src/pathSatisfies.js
```diff
@@ -0,0 +1,33 @@
...
+ * @memberOf R
+ * @since v0.18.1
+ * @category Logic
+ * @sig (a -> Boolean) -> [String] -> {String: a} -> Boolean
+ * @param {Function} pred
+ * @param {Array} propPath
+ * @param {*} obj
+ * @return {Boolean}
+ * @see R.propSatisfies
+ * @see R.path
```
davidchambers: Multiple `@see` annotations may appear on one line, separated by commas (e.g. `@see R.foo, R.bar`).

src/pathSatisfies.js
```diff
@@ -0,0 +1,33 @@
...
+var path = require('./path');
+
+
+/**
+ * Returns `true` if the specified object property at given path satisfies the given
+ * predicate; `false` otherwise.
+ *
+ * @func
+ * @memberOf R
+ * @since v0.18.1
```
davidchambers: Let's remove this line.

test/pathSatisfies.js
```diff
@@ -0,0 +1,21 @@
...
+describe('pathSatisfies', function() {
+
+  var isPositive = function(n) { return n > 0; };
+
+  it('returns true if the specified object path satisfies the given predicate', function() {
+    eq(R.pathSatisfies(isPositive, ['x', 'y'], {x: {y: 1}}), true);
+  });
+
+  it('returns false if the specified path does not exist', function() {
+    R.pathSatisfies(isPositive, ['x', 'y'], {x: {z: 42}});
```
davidchambers: There's no assertion here. ;)

CrossEye: sez you!

gilligan:
@CrossEye @davidchambers any further objections on this? ;)

src/pathSatisfies.js
```diff
@@ -0,0 +1,31 @@
+var _curry3 = require('./internal/_curry3');
+var path = require('./path');
+
+
+/**
+ * Returns `true` if the specified object property at given path satisfies the given
+ * predicate; `false` otherwise.
```
davidchambers: Let's wrap this at 79 characters (see #1461).

src/pathSatisfies.js
```diff
@@ -0,0 +1,31 @@
...
+ * @category Logic
+ * @sig (a -> Boolean) -> [String] -> {String: a} -> Boolean
+ * @param {Function} pred
+ * @param {Array} propPath
+ * @param {*} obj
+ * @return {Boolean}
+ * @see R.propSatisfies, R.path
+ * @example
+ *
+ *      R.propSatisfies(y => y > 0, ['x', 'y'], {x: {y: 2}}); //=> true
```
davidchambers: `s/prop/path/`

src/pathSatisfies.js
```diff
@@ -0,0 +1,31 @@
...
+
+
+/**
+ * Returns `true` if the specified object property at given path satisfies the given
+ * predicate; `false` otherwise.
+ *
+ * @func
+ * @memberOf R
+ * @category Logic
+ * @sig (a -> Boolean) -> [String] -> {String: a} -> Boolean
```
davidchambers: `{String: a}` doesn't capture nesting. Perhaps `Object` would be better.

test/pathSatisfies.js
davidchambers: We also need a test for the empty path.

src/pathSatisfies.js
```diff
@@ -0,0 +1,31 @@
...
+ * @example
+ *
+ *      R.pathSatisfies(y => y > 0, ['x', 'y'], {x: {y: 2}}); //=> true
+ */
+module.exports = _curry3(function pathSatisfies(pred, propPath, obj) {
+  switch (propPath.length) {
+    case 0:
+      return false;
+    case 1:
+      return pred(obj[propPath[0]]);
```
davidchambers: This case is unnecessary. The case below can cover `propPath.length === 1` as well.

gilligan:
@davidchambers thanks for the thorough review, fixed everything accordingly

gilligan:
@davidchambers alright, premature optimization I suppose. Use a simple if/else instead of the switch statement then I guess ?

gilligan:
Or even 
```js
module.exports = _curry3(function pathSatisfies(pred, propPath, obj) {                                                                                         
  return (propPath.length === 0) ? false : pred(path(propPath, obj));                                   
});
```

davidchambers:
Even simpler:

```javascript
propPath.length > 0 && pred(path(propPath, obj))
```

gilligan:
ah, so now tests passed - any other objections to this? :)

test/pathSatisfies.js
davidchambers: Could you add a test for the empty path, @gilligan?

CrossEye: Outside that, this LGTM.

gilligan: Sorry, was away - I added the empty path test.

CrossEye:
LGTM.  Anyone else with thoughts?

davidchambers:
:deciduous_tree:

PR #1454, opened 2015-10-20
refactor(useWith): use curryN instead of curry(_arity)

gonzaloruizdevilla:
simple refactor to reduce complexity


CrossEye:
Yes, thank you.  This is clearly an improvement.

davidchambers:
:deciduous_tree:

PR #1455, opened 2015-10-20
Uses regex match instead of constructor.name in _toString()

ConorLinehan:
Address issue #1393


davidchambers:
See also #1427

ConorLinehan:
@davidchambers Looks cool :) will close

PR #1459, opened 2015-10-21
Add log method

Youpinadi:
I know this is not very functional, but i find the method pretty handy. 
It's easy to get lost when you compose or pipe a lot. R.log just logs the current value and returns it.
When you're new in the FP world (as i am), it can help a lot.

It also affects the output of the tests (i don't know what to do about it):
screenshot 2015-10-21 21 41 03

If you don't like it, just close the pull request, no bad feelings ;)

### DIFF
--- a/src/log.js
+++ b/src/log.js
@@ -1,0 +1,21 @@
+var _curry2 = require('./internal/_curry2');
+
+
+/**
+ * Logs the current result in the console
+ *
+ * @func
+ * @memberOf R
+ * @since v1.8.0
+ * @category List
+ * @sig a -> a
+ * @return {Object|List|String} The argument that was passed.
+ * @example
+ *
+ *      R.pipe(R.toUpper, R.log, R.toLower, R.log)('abc') // logs 'ABC' and then 'abc'
+ */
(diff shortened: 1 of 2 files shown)


davidchambers:
Do you know about `R.tap(console.log)`? It's handy when debugging pipelines. In fact, I often write this:

```javascript
R.tap(R.compose(console.log, R.toString))
```

`console.log` needs to be bound in browser environments:

```javascript
R.tap(R.compose(R.bind(console.log, console), R.toString))
```

I'm strongly against adding this to Ramda, though, as it violates referential transparency.

Youpinadi:
Thanks for the answer! I didn't know tap. I will use your technique!

davidchambers:
No problem! :)

raine:
> It's easy to get lost when you compose or pipe a lot.

I usually do this: http://i.imgur.com/qjxvdHK.gif

davidchambers:
Very cool, @raine! I did spot an abuse of `R.add`, though. That should really be `R.concat`. :stuck_out_tongue_winking_eye:

raine:
Good call, I didn't know `concat` works with Strings.

PR #1462, opened 2015-10-23
remove functions deprecated in v0.18.0

davidchambers:


asaf-romano:
Probably not the right place for such a rant, but I must say that I'm sad to see mapObj going away. We upgraded to 0.18 right on launch, and as usual, tried to get rid of deprecated stuff while we're at it. At the end, I decided not to do mapObj->map change for now, as it made very clear pipelines somewhat confusing, as one can no longer - at a glance - know if she's working with an object or an array (think of: ``pipe(map(...), values)``).

</rant>

CrossEye:
@asaf-romano:

```js
const mapObj = R.map;
```

Or if you don't mind a little mutation, assign it back to `R.mapObj`.

test/whereEq.js
```diff
@@ -45,7 +45,6 @@ describe('whereEq', function() {
 
   it('is true for an empty spec', function() {
     eq(R.whereEq({}, {a: 1}), true);
-    eq(R.whereEq(null, {a: 1}), true);
```
raine: Should this test be removed from `test/where.js` as well?

davidchambers: Yes, I'd say so. I removed this assertion because it was failing, and we shouldn't support `null` here.

PR #1463, opened 2015-10-24
Sorting keys

CrossEye:
Sorting the output of `keys` and related calls.

This is certain to be controversial.  I'm not going to go over all the rationale for again; we've discussed it a few times, most recently in #1067, especially starting with my long comment, https://github.com/ramda/ramda/issues/1067#issuecomment-135923353.  This at least gives us something concrete to discuss.

There is a perhaps peculiar design decision here.  I'll understand if there are objections; I'm certainly not wedded to it.  Namely, `keys` is simply a reference to `internal/_sortedKeys`.  It adds nothing whatsoever.  The reason for this is that I want to have the `sortedKeys` name used very explicitly.  I want it to be quite clear when we are using that rather than the unsorted version.  But we still want the public API to be `keys`.  There was also one other concern that turned out not to be relevant, but which I always keep in the back of my head.  I really prefer not to have internal modules depend upon the public ones.  I want the dependencies to always run in one direction, the way one might think about layering an application.  It turned out not to be an issue here, as there was no internals that depended on `keys`.  (And of course this same decision applies to `keysIn` / `sortedKeysIn`.)

### DIFF
--- a/src/internal/_keysIn.js
+++ b/src/internal/_keysIn.js
@@ -1,4 +1,4 @@
-var _sortedKeysIn = require('./internal/_sortedKeysIn');
+var _curry1 = require('./internal/_curry1');
 
 
 /**
@@ -21,4 +21,10 @@ var _sortedKeysIn = require('./internal/_sortedKeysIn');
  *      var f = new F();
  *      R.keysIn(f); //=> ['x', 'y']
  */
-module.exports = _sortedKeysIn;
+module.exports = _curry1(function keysIn(obj) {
+  var prop, ks = [];
+  for (prop in obj) {
+    ks[ks.length] = prop;
+  }
+  return ks;
+});
--- a/src/internal/_sortedKeys.js
+++ b/src/internal/_sortedKeys.js
@@ -1,0 +1,7 @@
+var _curry1 = require('./_curry1');
+var _keys = require('./_keys');
+
+
+module.exports = _curry1(function(obj) {
+  return _keys(obj).sort();
+});
--- a/test/keys.js
+++ b/test/keys.js
@@ -10,7 +10,16 @@ describe('keys', function() {
   var cobj = new C();
 
   it("returns an array of the given object's own keys", function() {
-    eq(R.keys(obj).sort(), ['a', 'b', 'c', 'd', 'e', 'f']);
+    eq(R.keys(obj), ['a', 'b', 'c', 'd', 'e', 'f']);
+  });
+
+
+  it('lists the keys in their natural sort order', function() {
+    eq(R.keys({a: 1, d: 2, c: 3, e: 4, b: 5}), ['a', 'b', 'c', 'd', 'e']);
+  });
+
+  it('places numeric indices first, but in lexicographic order', function() {
+    eq(R.keys({a: 1, 20: 2, c: 3, 9: 4, b: 5}), ['20', '9', 'a', 'b', 'c']);
   });
 
   it('works with hasOwnProperty override', function() {
@@ -30,6 +39,6 @@ describe('keys', function() {
   });
 
   it("does not include the given object's prototype properties", function() {
-    eq(R.keys(cobj).sort(), ['a', 'b']);
+    eq(R.keys(cobj), ['a', 'b']);
   });
 });
--- a/test/toPairs.js
+++ b/test/toPairs.js
@@ -6,6 +6,12 @@ describe('toPairs', function() {
   it('converts an object into an array of two-element [key, value] arrays', function() {
     eq(R.toPairs({a: 1, b: 2, c: 3}), [['a', 1], ['b', 2], ['c', 3]]);
   });
+
+  it('sorts the results according to natural key-order', function() {
+    eq(R.toPairs({c: 3, a: 1, b: 2}), [['a', 1], ['b', 2], ['c', 3]]);
+  });
+
+
   it("only iterates the object's own properties", function() {
     var F = function() {
       this.x = 1;
--- a/test/values.js
+++ b/test/values.js
@@ -9,6 +9,14 @@ describe('values', function() {
   C.prototype.y = 'y';
   var cobj = new C();
 
+  it('lists the values of an object in the natural sort order of the related keys', function() {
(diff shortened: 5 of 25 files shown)


src/internal/_sortedKeys.js
```diff
@@ -0,0 +1,7 @@
+var _curry1 = require('./_curry1');
+var _keys = require('./_keys');
+
+
+module.exports = _curry1(function(obj) {
```
davidchambers: Why are we applying `_curry1` to an internal function?

test/keys.js
```diff
@@ -10,7 +10,16 @@ describe('keys', function() {
   var cobj = new C();
 
   it("returns an array of the given object's own keys", function() {
-    eq(R.keys(obj).sort(), ['a', 'b', 'c', 'd', 'e', 'f']);
+    eq(R.keys(obj), ['a', 'b', 'c', 'd', 'e', 'f']);
+  });
+
+
```
davidchambers: :space_invader:

test/toPairs.js
```diff
@@ -6,6 +6,12 @@ describe('toPairs', function() {
   it('converts an object into an array of two-element [key, value] arrays', function() {
     eq(R.toPairs({a: 1, b: 2, c: 3}), [['a', 1], ['b', 2], ['c', 3]]);
   });
+
+  it('sorts the results according to natural key-order', function() {
+    eq(R.toPairs({c: 3, a: 1, b: 2}), [['a', 1], ['b', 2], ['c', 3]]);
+  });
+
+
```
davidchambers: :space_invader:

test/keys.js
```diff
@@ -10,7 +10,16 @@ describe('keys', function() {
...
+  it('lists the keys in their natural sort order', function() {
+    eq(R.keys({a: 1, d: 2, c: 3, e: 4, b: 5}), ['a', 'b', 'c', 'd', 'e']);
+  });
+
+  it('places numeric indices first, but in lexicographic order', function() {
```
davidchambers: This is misleading, since numeric property names are not given special treatment:

```javascript
> R.keys({'0': true, '/': true})
['/', '0']
```

I suggest we remove the reference to numeric indices.

test/values.js
```diff
@@ -9,6 +9,14 @@ describe('values', function() {
   C.prototype.y = 'y';
   var cobj = new C();
 
+  it('lists the values of an object in the natural sort order of the related keys', function() {
+    eq(R.values({a: 1, d: 2, c: 3, e: 4, b: 5}), [1, 5, 3, 2, 4]); // keys: [a, b, c, d, e]
+  });
+
+  it('places numeric indices first, but in lexicographic order', function() {
```
davidchambers: See comment above.

asaf-romano:
Quick question... are you also changing the behavior or toPairs. My project (unfortunately) relies on the current not-sorting behavior.

src/internal/_sortedKeys.js
CrossEye: Ahh, because this was originally the public `keys` function until I realized that I wanted to make a point of being explicit in calling sort.  I'll move this to the public function, and then my caveat above is no longer necessary.  It'll have to wait until I'm on a computer, though.

test/keys.js
CrossEye: Will fix.

test/keys.js
CrossEye: Actually, I'd rather use this, and include it in the example, dropping "numeric", but keeping "lexicographic" (or "Unicode", or some other functionally equivalent wording).  I think it's useful to note.

This also reminds me that I didn't update any docstrings.  Obviously at least one more commit to come.

test/toPairs.js
CrossEye: yup

test/values.js
CrossEye: ditto

CrossEye:
@asaf-romano:

> are you also changing the behavior or `toPairs`?

Yes, it does.  The public functions changed by this are

  * `functions`
  * `functionsIn`
  * `invert`
  * `invertObj`
  * `keys`
  * `keysIn`
  * `toPairs`
  * `toPairsIn`
  * `values`

Although I should remove `functions`, `functionsIn` as those are being removed.

How deep is your dependency on object order right now?

asaf-romano:
I'm relying on it in for a very particular use case: mongodb compound indexes (awfully!) rely on json properties ordering, i.e. the index { field1: 1, filed2: 1 } is not the same as { field2: 1, field1: 1 }. So, in our db-migration system, when we need to decide whether an index should be removed/added, we ``eqBy(toPairs)``, because mongodb's behavior "accidentally" matches v8's behavior.

Also keep in mind that the current

buzzdecafe:
doesn't the sorting mean that if `x` is a list of tuples (pairs), then `toPairs(fromPairs(x)) != x`?

CrossEye:
> doesn't the sorting mean that if x is a list of tuples (pairs), then toPairs(fromPairs(x)) != x in some cases?

That's true, regardless of sorting.

`fromPairs` is a left inverse (_aka_ 'retraction') of `toPairs`, which also means that `toPairs` is a right inverse (_aka_ 'section') of `fromPairs`, but they are not full inverses of one another.

```js
fromPairs(toPairs(x))  ≍ x, // ∀ x
```

But 

```js
R.toPairs(R.fromPairs([['a', 1], ['a', 2]])); //=> ['a', 2]
```

So this is not an identity we could even try to maintain.

jdalton:
:deciduous_tree:?

CrossEye:
Not ready yet.  I've dropped the ball.  But will be back on it in a few days.

src/internal/_keysIn.js
```diff
@@ -0,0 +1,30 @@
...
+ *      var F = function() { this.x = 'X'; };
+ *      F.prototype.y = 'Y';
+ *      var f = new F();
+ *      R.keysIn(f); //=> ['x', 'y']
+ */
+module.exports = _curry1(function keysIn(obj) {
+  var prop, ks = [];
+  for (prop in obj) {
+    ks[ks.length] = prop;
+  }
```
megawac: this will run into the enum bug addressed in keys

CrossEye: I don't know if we should address that in this PR.  I'm trying not to change behavior, except that for certain public functions, the results will now be sorted.  This was the existing behavior.

PR #1464, opened 2015-10-24
Fix reduce sig.

### LINKED ISSUE #1445: Reconsider the signature of reduce
I'm not at a computer now, or this would be a PR.  A [recent StackOverflow question][so] noted that that the current signature for `reduce` looks like this:

    (a,b -> a) -> a -> [b] -> a

which can easily be mis-parsed as

    (a, (b -> a)) -> a -> [b] -> a

We should make it unambiguous with

    ((a,b) -> a) -> a -> [b] -> a

  [so]: http://stackoverflow.com/questions/33142252

CrossEye:
Closes #1445

### DIFF
--- a/src/reduce.js
+++ b/src/reduce.js
@@ -20,7 +20,7 @@ var _reduce = require('./internal/_reduce');
  * @memberOf R
  * @since v0.1.0
  * @category List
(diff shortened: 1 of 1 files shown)


davidchambers:
:deciduous_tree:

Issue #1367, opened 2015-09-02
ImmutableJS support


DaleJefferson:
I'm currently working on an existing project that uses [immutableJS](https://facebook.github.io/immutable-js/) for all data structures, i've been trying to adopt ramda. Obviously all of the array and object functions don't work and I have to convert in and out of plainJS objects.

```js
var personObject = {firstName: "Dale"}
var personImmutable = immutable.Map(personObject)

R.prop("firstName", personObject) // "Dale"

R.prop("firstName", personImmutable) // undefined
```

I've started a repo to add ImmutableJS support to Ramda at https://github.com/DaleJefferson/ramda-immutable
I've completed a couple of methods as a proof of concept and it works nicely.

```js
function prop(p, obj) {
    if (immutable.Map.isMap(obj)) {
        return obj.get(p)
    }

    return R.prop(p, obj)
}
```

I wonder if this should be part of Ramda or sit as a separate project? It could be done in a generic way like equals.

```js
// This already works with immutableJS because all immutableJS objects implement equals.
function equals(a, b) {
  return _hasMethod('equals', a) ? a.equals(b) : _hasMethod('equals', b) ? b.equals(a) : _equals(a, b, [], []);
})

// immutableJS Maps have a get method to access props
function prop(p, obj) {
    if (_hasMethod("get", obj)) {
        return obj.get(p)
    }

    return obj[p]
}

// immutableJS Maps have a first method to access the first item
function head(list) {
    if (_hasMethod("first", list)) {
        return list.first()
    }

    return list[0]
}
```

CrossEye:
See also #1037, which discussed integration with Highland; it has not had any activity in quite a while.

My favorite idea for this would be to allow the user to supply a number of delegation aliases and get back an instance of Ramda with those aliases applied.  A new instance of Ramda is pretty important in order to maintain referential integrity as discussed [in that thread][th].  Another possibility would be to add these at build time, offering a configuration file that listed delegate aliases, and then the user could just build a version that made sense for her environment.  Ramda could publish likely candidates like `ramda.immutablejs.js` or `ramda.highland.js` as we saw fit.


  [th]: https://github.com/ramda/ramda/issues/1037#issuecomment-96050496

DaleJefferson:
I like the aliases idea but it would require a lot of changes since Ramda internally inconsistently delegates to other Ramda functions. I assume this is for performance reasons?

```js
function propSatisfies(pred, name, obj) {
  return pred(obj[name]);
});

[...]

Issue #441, opened 2014-10-12
commute (fka traverse)


buzzdecafe:
this is shamelessly "adapted" from @drboolean 's https://github.com/DrBoolean/pointfree-fantasy/blob/master/instances/array.js : 

``` javascript
R.traverse = curry3(function(f, point, ls) {
    var cons_f = function(ys, x){
        return R.ap(R.map(function(x){ return function(y){ return y.concat(x); } }, f(x)), ys);
    };
    return R.reduce(cons_f, point([]), ls);
});
```

This would enable reorganizing Functors, e.g an Array of Maybes into a Maybe of an Array, i.e:

``` javascript
var xs = [1,2];
R.traverse(Maybe.of, Maybe.of, xs); //=>  Maybe([1,2])

// as opposed to:
R.map(Maybe.of, xs); //=> [Maybe(1), Maybe(2)];
```

useful? keep it in the `types` extension? thoughts?

PR #1467, opened 2015-10-25
add R.sequence and R.traverse

davidchambers:
This pull request deprecates `R.commute` and `R.commuteMap` in favour of `R.sequence` and `R.traverse` respectively.

Reasons to rename the functions:

  - My understanding is that `sequence` and `traverse` are the most common names for these functions. I read and reread [Understanding traverse and sequence][1] in order to come to terms with this concept, so I may be biased by the names used there. Is `commute` the name used anywhere but in Ramda?

  - Fantasy Land defines [Traversable][2], which requires that a Traversable provide a `sequence` method. It also shows that `traverse` can be derived from `map` and `sequence`.

  - Currently each Ramda functions which dispatches has the same name as the method to which it dispatches. Having `R.commute` dispatch to `sequence` would break this convention.

Changes:

  - `R.sequence` dispatches; `R.commute` does not.

  - `R.traverse` takes `(of, f, traversable)` rather than `(f, of, traversable)`. It's more likely one would apply several different `f` transformations to a single :twisted_rightwards_arrows: type inversion than apply the same transformation to several different type inversions. I also like having `of` off to the side as it conveys type information that would be unnecessary in the presence of type inference.

I'm particularly interested in hearing from @scott-christopher, who has a better grasp of this than I do. We have discussed these changes (though Scott is happy to keep the names) but I'm not sure that I've done everything correctly.

I'm about to :airplane: from Auckland to San Francisco, so I'll be offline for the next 12 hours or so.

[1]: http://fsharpforfunandprofit.com/posts/elevated-world-4/
[2]: https://github.com/fantasyland/fantasy-land#traversable

### DIFF
--- a/src/commute.js
+++ b/src/commute.js
@@ -13,7 +13,8 @@ var identity = require('./identity');
  * @param {Function} of A function that returns the data type to return
  * @param {Array} list An array of functors of the same type
  * @return {*}
- * @see R.commuteMap
+ * @see R.sequence
+ * @deprecated since v0.19.0
  * @example
  *
  *      R.commute(R.of, [[1], [2, 3]]);   //=> [[1, 2], [1, 3]]
--- a/src/commuteMap.js
+++ b/src/commuteMap.js
@@ -18,7 +18,8 @@ var reduceRight = require('./reduceRight');
  * @param {Function} of A function that returns the data type to return
  * @param {Array} list An array of functors of the same type
  * @return {*}
- * @see R.commute
+ * @see R.traverse
+ * @deprecated since v0.19.0
  * @example
  *
  *      var add10 = R.map(R.add(10));
--- a/src/sequence.js
+++ b/src/sequence.js
@@ -1,0 +1,37 @@
+var _curry2 = require('./internal/_curry2');
+var ap = require('./ap');
+var map = require('./map');
+var prepend = require('./prepend');
+var reduceRight = require('./reduceRight');
+
+
+/**
+ * Transforms a [Traversable](https://github.com/fantasyland/fantasy-land#traversable)
+ * of [Applicative](https://github.com/fantasyland/fantasy-land#applicative) into an
+ * Applicative of Traversable.
+ *
+ * Dispatches to the `sequence` method of the second argument, if present.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig (Applicative f, Traversable t) => (a -> f a) -> t (f a) -> f (t a)
+ * @param {Function} of
+ * @param {*} traversable
+ * @return {*}
+ * @see R.traverse
+ * @example
+ *
+ *      R.sequence(Maybe.of, [Just(1), Just(2), Just(3)]);   //=> Just([1, 2, 3])
+ *      R.sequence(Maybe.of, [Just(1), Just(2), Nothing()]); //=> Nothing()
+ *
+ *      R.sequence(R.of, Just([1, 2, 3])); //=> [Just(1), Just(2), Just(3)]
+ *      R.sequence(R.of, Nothing());       //=> [Nothing()]
+ */
+module.exports = _curry2(function sequence(of, traversable) {
+  return typeof traversable.sequence === 'function' ?
+    traversable.sequence(of) :
+    reduceRight(function(acc, x) { return ap(map(prepend, x), acc); },
+                of([]),
+                traversable);
+});
--- a/src/traverse.js
+++ b/src/traverse.js
@@ -1,0 +1,33 @@
+var _curry3 = require('./internal/_curry3');
+var map = require('./map');
+var sequence = require('./sequence');
+
+
+/**
+ * Maps an [Applicative](https://github.com/fantasyland/fantasy-land#applicative)-returning
+ * function over a [Traversable](https://github.com/fantasyland/fantasy-land#traversable),
+ * then uses [`sequence`](#sequence) to transform the resulting Traversable of Applicative
+ * into an Applicative of Traversable.
+ *
+ * Dispatches to the `sequence` method of the third argument, if present.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig (Applicative f, Traversable t) => (a -> f a) -> (a -> f b) -> t a -> f (t b)
+ * @param {Function} of
+ * @param {Function} f
+ * @param {*} traversable
+ * @return {*}
+ * @see R.sequence
+ * @example
+ *
+ *      R.traverse(Maybe.of, R.negate, [Just(1), Just(2), Just(3)]);   //=> Just([-1, -2, -3])
+ *      R.traverse(Maybe.of, R.negate, [Just(1), Just(2), Nothing()]); //=> Nothing()
+ *
+ *      R.traverse(R.of, R.negate, Just([1, 2, 3])); //=> [Just(-1), Just(-2), Just(-3)]
+ *      R.traverse(R.of, R.negate, Nothing());       //=> [Nothing()]
+ */
+module.exports = _curry3(function traverse(of, f, traversable) {
+  return sequence(of, map(f, traversable));
+});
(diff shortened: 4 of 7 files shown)


scott-christopher:
> R.sequence(R.of, []) evaluates to [], whereas R.commute(R.of, []) evaluates to [[]]

I don't think this is correct. Going down this path would also require the applicative instance to have a `Monoid` instance for `empty`. Calling `sequence` on an empty traversable should result in a pure instance of the applicative type containing the empty traversable.
e.g.
```js
sequence(R.of, []) //=> [[]]
sequence(Maybe.of, []) //=> Just ([])
sequence(Maybe.of, Nothing) //=> Just (Nothing)
```

There are [a few laws](https://hackage.haskell.org/package/base-4.8.1.0/docs/Data-Traversable.html#interface) that should confirm this.

For example:
```
var identityL = compose(sequence(Identity.of), map(Identity));
var identityR = Identity;

identityL([]) == identityR([]); //=> false

identityL([]); //=> []
identityR([]); //=> Identity([])
```

scott-christopher:
I'm happy with this PR pending the comments above. The name change doesn't bother me too much, providing the behaviour doesn't drift too much in the process.

CrossEye:
I don't have any objections to the name change, so long as we understand the reasons (#441) why we chose not to use `traverse` in the first place, and are certain they no longer hold.

I will have to read through that article or something similar to get enough of a feel to comment on the content and the fundamentals of the API change.  Maybe I'll have time tomorrow.

davidchambers:
> > `R.sequence(R.of, [])` evaluates to `[]`, whereas `R.commute(R.of, [])` evaluates to `[[]]`.
>
> I don't think this is correct. Going down this path would also require the applicative instance to have a `Monoid` instance for `empty`. Calling `sequence` on an empty traversable should result in a pure instance of the applicative type containing the empty traversable.
> e.g.
> ```javascript
> sequence(R.of, []) //=> [[]]
> sequence(Maybe.of, []) //=> Just ([])
> sequence(Maybe.of, Nothing) //=> Just (Nothing)
> ```

I've removed the special case for `[]`, but I'm still uncertain. Can you explain the following?

```javascript
> R.sequence(R.of, [])
[[]]
> R.sequence(R.of, [[]])
[]
```

Haskell and `R.commute`/`R.sequence` appear not to agree:

```
Prelude Data.Traversable> sequenceA []
[]
Prelude Data.Traversable> sequenceA [[]]
[]
Prelude Data.Traversable> sequenceA Nothing
Nothing
```

/cc @DrBoolean

scott-christopher:
I'm assuming it's something funky going on with GHCi and `IO` there.
```
Prelude> let a = [] :: [IO a]
Prelude> sequenceA a
[]
```

If you explicitly declare their types they behave as I would expect.
```
Prelude> let b = [] :: [[a]]
Prelude> sequenceA b
[[]]
Prelude> let c = [] :: [Maybe a]
Prelude> sequenceA c
Just []
Prelude> let d = Nothing :: Maybe [a]
Prelude> sequenceA d
[Nothing]
```

davidchambers:
> I'm assuming it's something funky going on with GHCi and `IO` there.

Oh! This would never have occurred to me! I'm pleased Haskell agrees with us after all. :)

Do you have an explanation as to why `[]` becomes `[[]]` but `[[]]` becomes `[]`? I'm now convinced there *is* an explanation; I'd love to learn of it.

scott-christopher:
Probably the best intuition I could try to offer (it still isn't entirely intuitive to me) is that if the inner applicative exists, the result will be based on how that applicative behaves with `ap`.

```
> sequence(of, [[1, 2], [3, 4]]);
[ [ 1, 3 ], [ 1, 4 ], [ 2, 3 ], [ 2, 4 ] ]

// compare to

> var a = R.ap(R.map(R.prepend, [3, 4]), [[]])
[ [ 3 ], [ 4 ] ]
> R.ap(R.map(R.prepend, [1, 2]), a)
[ [ 1, 3 ], [ 1, 4 ], [ 2, 3 ], [ 2, 4 ] ]
```

```
> sequence(of, [[1, 2], []]);
[]

// compare to

> R.ap(R.map(R.prepend, []), [[]])
[]
```

```
> sequence(of, [[], [3, 4]]);
[]

// compare to

> R.ap(R.map(R.prepend, []), [[3], [4]])
[]
```

davidchambers:
> it still isn't entirely intuitive to me

I'm pleased to hear this, in a way. :)

Perhaps this is a concept that reveals itself over time rather than being learnable in one sitting.

Have you any thoughts, @CrossEye, on dispatching to `sequence` or on taking `of` before `f`?

CrossEye:
I will try to look at it this evening.  I'm not comfortable enough with any of this yet to comment.

buzzdecafe:
:-1: on the name `traverse` for this. would be nice to reserve `traverse` for, you know, tree traversal et al.

davidchambers:
> :-1: on the name `traverse` for this. would be nice to reserve `traverse` for […] tree traversal et al.

If we use the name at all we should give it to the `traverse` function described in the Fantasy Land spec. One option is to provide `R.sequence` only. Those who want the `map` + `sequence` behaviour can write:

    R.sequence(of, R.map(f, traversable))

DrBoolean:
I see it as the same issue as always: common names across FP libraries or more expressive names in general. I really like the name `commute`, with a possible alias of `sequence` for fantasyland compliance.

CrossEye:
@DrBoolean: Unfortunately, or more likely, fortunately, Ramda has forsworn aliases altogether.

davidchambers:
We could keep the name `commute` (which I also like) but have `R.commute` dispatch to `sequence`.

DrBoolean:
I'm anti-alias anyways :)

buzzdecafe:
> I'm anti-alias anyways :)

that explains why you're so smooth :stuck_out_tongue:

DrBoolean:
zing!

CrossEye:
#1367 might be coming back to life, which could mean all sorts of extensions to our delegation techniques.  But I would rather not by default delegate to a different name, even if that turns out to be a part of an integration mechanism.

I originally raised an objection to "`Traverable`" in #441, saying much as @buzzdecafe does here that it really should be reserved for something else.  But I don't see us actually including that something else, and this is now part of FantasyLand, so I'm all for including it here.  :herb:

I still haven't read that article, and once again, won't have time this evening.  But at first glance, this implementation looks sensible to me.

DrBoolean:
Maybe we propose fantasyland update to `commute` and drop the mention of traverse. I'm down to do that.

davidchambers:
> Maybe we propose fantasyland update to `commute` and drop the mention of traverse. I'm down to do that.

Would the type class still be named Traversable or would we change it to Commutable at the same time? Perhaps such a discussion should happen in fantasy land. Fancy raising the issue over there, Brian?

DrBoolean:
Certainly

DrBoolean:
https://github.com/fantasyland/fantasy-land/issues/115

CrossEye:
Thanks, DrB!

joneshf:
> :-1:  on the name traverse for this. would be nice to reserve traverse for, you know, tree traversal et al.

What is a list if not a degenerate tree? All the branches are on the right.

```
[1,2,3] == 1:[2,3] == 1:2:[3] == 1:2:3:[] ==   :
                                              / \
                                             1   :
                                                / \
                                               2   :
                                                  / \
                                                 3  []
```

Looks like a tree to me. :)

davidchambers:
@joneshf, are you in favour of this pull request as it stands?

buzzdecafe:
> What is a list if not a degenerate tree? 

@joneshf i can't argue with that. since fantasyland/fantasy-land#115 was closed, i think it makes sense to go with the prevailing wind in this case, IOW :cow2:

CrossEye:
And I'll say again, :herb:.  My original arguments are not relevant any more, especially as this version does delegate to the type's versions of `traverse` / `sequence`.

PR #1471, opened 2015-10-27
fix signature of R.times

davidchambers:


scott-christopher:
:+1:

fyyyyy:
I was just wondering, why is R.times categorized under List ? I would expect it under function ?
Cheers

buzzdecafe:
> why is `R.times` categorized under List ?

`times` generates a list. Seems a reasonable category to me.

CrossEye:
:herb:

fyyyyy:
Fair enough though `R.of` or `R.op` create Lists. And they dont necessarily accept functions yet they are categorized under `Function`. I thought categorization would be either by argument type or return type.
Or both :D. But maybe thats not practicable.

CrossEye:
> [ ... ] though `R.of` or `R.ap` create Lists.

Only in certain circumstances.  Those are both specified in [FantasyLand][fl].  `of` returns an [Applicative][of] and `ap` returns an [Apply][ap].  We treat a list as both.

I would like to expand the categories to note the types of any dominant inputs and the output.  (By dominant, I mean to exclude, for instance, `Number` from the list of categories for `slice`.  Even though the majority of the parameters are numbers, this function isn't _about_ numbers, unlike, `add` or `mean`.)  But this is not near the top of my list of priorities.

  [ap: https://github.com/fantasyland/fantasy-land#apply
  [of]: https://github.com/fantasyland/fantasy-land#applicative

PR #1473, opened 2015-10-27
Remove excess words in Map documentation

fix-fix:
'Map' word was removed to keep it in the same style as in `ap` docs


CrossEye:
Thanks!

PR #1475, opened 2015-10-27
Revert lift-based implementation of 'both' and 'either'

### LINKED ISSUE #1466: 0.18 regression: R.both does not short circuit
```js

R.both(R.complement(R.isNil), R.whereEq({ a: 1 }))(null)
// => 0.18: TypeError: n is null
// => 0.17: false
```

asaf-romano:
Closes #1466

### DIFF
--- a/src/both.js
+++ b/src/both.js
@@ -1,13 +1,12 @@
-var and = require('./and');
-var lift = require('./lift');
+var _curry2 = require('./internal/_curry2');
 
 
 /**
  * A function wrapping calls to the two functions in an `&&` operation,
  * returning the result of the first function if it is false-y and the result
- * of the second function otherwise.
- *
- * `R.both` will work on all other applicatives as well.
+ * of the second function otherwise. Note that this is short-circuited,
+ * meaning that the second function will not be invoked if the first returns a
+ * false-y value.
  *
  * @func
  * @memberOf R
@@ -26,4 +25,8 @@ var lift = require('./lift');
  *      f(100); //=> true
  *      f(101); //=> false
  */
-module.exports = lift(and);
+module.exports = _curry2(function both(f, g) {
+  return function _both() {
+    return f.apply(this, arguments) && g.apply(this, arguments);
+  };
+});
--- a/src/either.js
+++ b/src/either.js
@@ -1,13 +1,12 @@
-var lift = require('./lift');
-var or = require('./or');
+var _curry2 = require('./internal/_curry2');
 
 
 /**
  * A function wrapping calls to the two functions in an `||` operation,
  * returning the result of the first function if it is truth-y and the result
- * of the second function otherwise.
- *
- * `R.either` will work on all other applicatives as well.
+ * of the second function otherwise. Note that this is short-circuited,
+ * meaning that the second function will not be invoked if the first returns a
+ * truth-y value.
  *
  * @func
  * @memberOf R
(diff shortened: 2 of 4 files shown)


asaf-romano:
I had an interesting half an hour with git squash; had to kill the branch and start over.

CrossEye:
:herb:

But also :cry: to see that elegant implementation not work for us.

@benperez: Any response to this?

asaf-romano:
ping, pong.

CrossEye:
Good point.

Any comments, @buzzdecafe, @davidchambers, @paldepind, @raine, @scott-christopher, @TheLudd?  Or again, @benperez?

raine:
I am very happy to have the pre v0.18 behavior back.

buzzdecafe:
:cow2:

scott-christopher:
:mushroom:

davidchambers:
:deciduous_tree:

benperez:
@CrossEye Sorry for missing this one, I was out of the country for a while. Doesn't seem like there is any other way to get the short circuiting behavior in JS, although it is a bit of a shame.

CrossEye:
I'm out of the country now.  ;-)

It really does seem a shame.  But we do stress that we won't let implementation elegance dictate our API.  I am sorry to see it go, though.

benperez:
I thought about this today after a good night's sleep. The reason I preferred the lift based implementation wasn't just about implementation elegance, but rather that it was a better abstraction for what these functions are meant to do.

The current behavior of threading the results of applying 2 functions through `&&` and `||` is only one interpretation of lifted `&&` and `||` functions — for *functions* as applicative functors.

What's nice about the lift based implementations is that they actually work for *all* `fantasy-land` compatible applicative functors.
```javascript
> const R = require('ramda');
> const S = require('sanctuary');
> R.both(S.Nothing(), S.Just(true));     // => S.Nothing()
> R.either(S.Nothing(), S.Just(true));   // => S.Nothing()
> R.both(S.Just(false), S.Just(true));   // => S.Just(false)
> R.either(S.Just(false), S.Just(true)); // => S.Just(true)
```

Short circuiting is an important feature, and as was pointed out [here](https://github.com/ramda/ramda/issues/1466#issuecomment-151468622), it's not really possible to do in an eagerly evaluated language, unless we were to use something like a generator in the implementation of `liftN`, which I'm guessing is a non-starter.

Short circuiting only really seems important in the case of functions, since `Maybe`, `Either`, etc are just values already. I propose that we take the best of both worlds by special casing for functions in `R.both` and `R.either`
```javascript
R.both = _curry2(function both(f, g) {
  return f instanceof Function ?
    function _both() { return f.apply(this, arguments) && g.apply(this, arguments); } :
    lift(and)(f, g);
});
```

davidchambers:
> I propose that we take the best of both worlds by special casing for functions in `R.both` and `R.either`

We can have our :cake: and eat it too? Sounds good!

benperez:
@CrossEye I'll open a PR if there are no hesitations from you.

CrossEye:
@benperez:

This sounds great.  I hadn't seen this side of it at all when you brought up the original implementation.  To me these simply _were_ operations on functions.  I like the higher abstraction a lot, though.

I haven't wrapped my head around _how_ they apply to arbitrary FL-compatible applicative functors.  I understood how our `and` and `or` once worked with Sanctuary's versions, but that was through delegation of these functions.

Definitely :+1: on a PR for this.

PR #1482, opened 2015-10-30
Implementing Generators support

niloy:
I've modified a few list based functions to work with ES6 Generators as a PoC. 
This enables _lazy_ evaluation of list. Also, this makes possible to work with 
_infinite_ list similar to other functional languages.

New API to create a generator:

``` js
xrange:: Number -> Number -> Number -> Generator
xrange(step, start, stop) -> Generator
```

A Infinite list can be created by passing `stop = Infinity`. Example:

``` js
const natural = R.xrange(1, 1, Infinity);
```

Plain JS array can be extracted from a generator based list using `R.take`. Example:

``` js
R.take(10, natural); // [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
```

Composition also works.

``` js
// List first 10 even squares
R.take(10, R.filter(x => x % 2 === 0, R.map(x => x * x, natural)));
```

`R.zip` can be used to zip 2 infinite list or (1 infinite with 1 finite list):

``` js
const even = R.xrange(2, 2, Infinity);
const natural = R.xrange(1, 1, Infinity);

R.take(3, R.zip(natural, even)); // [[1, 2], [2, 4], [3, 6]]
R.zip(natural, ['a', 'b', 'c']); // [[1, 'a'], [2, 'b'], [3, 'c']]
```

`R.repeat` can be used to create infinitely repeating values:

``` js
R.take(5, R.repeat(1, Infinity)); // [1, 1, 1, 1, 1]
```

`R.times` can be used to create an infinite list. Example, an infinite stream
of random values:

``` js
R.take(5, R.times(Math.random, Infinity));
```

The following functions have been modified to work with generators:
- head, nth
- tail, slice
- uniq, uniqWith
- reduce
- reverse
- length
- splitEvery
- forEach
- filter
- map
- zip
- take

Will work on other functions if people see value in this PR.


CrossEye:
My first thought is that it's great to implement some form of laziness again.  We pulled out the little-used lazy list extension in #857, but I've always felt a bit of regret over it.  It would be really nice to manage to get it back.  It would be useful to support a [wider range][gt] of collections.

I'm hesitant about this approach for several reasons, though.

First of all, Ramda has been fairly careful about supporting all sorts of old environments.  We don't really test thoroughly in old browsers, but generally we try to write in a way that makes no major assumptions about the user's JS implementation.  Generally, Ramda should work in any ES3+ environment.  We certainly don't want to insist on it being ES6+-only.  This approach would limit us that way.

Second, I think the more general approach would be to support the iteration protocols.  We could do that like we do other things, in a backward-compatible manner, using a Symbol for modern environments and a '`@@iterator`' string for older ones.  We could accept them in many of the same functions you've altered, and use them for internal implementations of lazy lists.

Finally, although there is a simple sense in which a sequence of integers can serve as a basis for any such system, that sense is only the most trivial mathematical one, that all finite or countably infinite sets can be mapped to a subset of the natural numbers.  It doesn't capture the essence of many things we might want to `map`, `filter`, `reverse`, etc.  So using as a basis only a generator for lists of natural numbers seems to be too limiting.

So, I think there are great ideas here, and I think this could be the basis for some useful changes to Ramda, I think it will need some substantial rework before we can include it.



  [gt]: https://github.com/kriskowal/gtor

niloy:
>  We certainly don't want to insist on it being ES6+-only.

I've used babel to transpile the generators, so this version of Ramda should
still work for pre ES6 environments. Without babel, uglifyjs refuses to produce
the minifed js since it cannot compile ES6 generator syntax yet. The `const` and
`let` are also converted to `var` by babel. I can do it manually if you like.

> Second, I think the more general approach would be to support the iteration protocols.

The iteration protocol is already supported. All the changes that I've made,
I have used `.next()` to fetch the next item. So, any plain JS object that implements
`.next` and returns `{value, done}` as result would work, not just ES6 iterators.

>  So using as a basis only a generator for lists of natural numbers seems to be too limiting.

I agree with you. The `xrange` is not really limited to natural numbers, real
numbers can be generated using it. I've used natural numbers to write the test
cases with. The `xrange` is similar to `[1..]` syntax of Haskell, that it can be
used to quickly generate a infinite list.

Perhaps you are saying, people dont want to map over numbers in real life situations.
I agree with you on that also. But we need to start somewhere and an infinite
list of natural numbers can serve as that starting point. If it works, we can
go ahead and create generators that would yield on `keypress`, `click` and
`read/write`.

> I think it will need some substantial rework before we can include it

Great, this is really just a PoC. I am not expecting it to get accepted as-is.
I dont mind a rework.

CrossEye:
I haven't dug into the implementation at all yet.  I will do so as soon as I can.

I wouldn't mind having Babel in our tool-chain, as long as the generated code is quite readable.

> Perhaps you are saying, people dont want to map over numbers in real life situations.  I agree with you on that also. But we need to start somewhere and an infinite list of natural numbers can serve as that starting point.

Well, I meant something more general.  I don't think that lists of numbers are at all the correct portal into lazy lists.  I would rather use a better version of `unfold` -- which really could use a boost -- or some simple iterations.  I had only read the Issue, and not the code, so I was assuming that this was the specified to initialize lazy lists with your new system.  If there are other equally simple ways, I apologize.  I will look into the details soon(ish).

dwsmorris:
I've been trying to find library support for infinite sequences in a javascript library.
There is some support in certain libraries but none of them feels quite right to me.

Highland.js makes you retrieve the result in a callback so its api can fit over asynchronous collections.
Immutable.js has basic support but it doesn't seem like you can use unfold or some kind of iterating generator.
Same with lazy.js, and it's data-first api just seems wrong having seen ramda and functional.js.
lodash-fp looks good - and I'm impressed by lodash's performance credentials - but doesn't look like there's a lodash-fp-contrib project, so all the extended functional operations like generators are presumably missing.
There is lazy-array, which does exactly what I was looking for, but its api is directly ported from clojure and so it doesn't fit so well in javascript.
There's also wu.js, but it uses ES6 syntax heavily. Call me old-fashioned, but I don't want to use these ES6 compilers - and I want the output library (eg. dist/ramda.js) to be debuggable and written in understandable javascript.

Ramda feels exactly right, so I wanted to be able to do these kind of things in it. I heard you can create infinite lists with transducers. There seems to be a dearth of comprehensible information on them, but I finally uncovered (undocumented?) support for feeding generators into ramda transducers to achieve infinite sequences.

Apparently, it's not the same as lazy lists, but the effect seems similar, at least from the perspective of creating unbounded generated lists with chained sequence operations that don't create intermediate collections. 

You seem to be able to chain sequence operations without explicitly wrapping them in some transformer object - nice - but there's a problem with support for sequence ordering that I'll put a pull request in for.

Great work on all this stuff - the generator support in transducers seems very powerful and it would be great to have this documented (unless I missed it!) 

I saw some discussion a while back about making the main api work equally well with arrays as with generators - not sure if this proposed work would be a stepping stone to that - but it would be great if so!

PR #1493, opened 2015-11-02
Documentation fixes

raine:


src/concat.js
```diff
@@ -13,8 +13,8 @@ var invoker = require('./invoker');
  * @category List
  * @sig [a] -> [a] -> [a]
  * @sig String -> String -> String
- * @param {Array|String} a
- * @param {Array|String} b
+ * @param {Array|String|Object} a
+ * @param {Array|String|Object} b
```
davidchambers: This doesn't seem right to me. Are you including `Object` because the function dispatches?

davidchambers:
Great stuff, @raine!

src/concat.js
raine: Yeah, because of this https://github.com/ramda/ramda/blob/7e25f1f6e75cb55d9653ebb6ba46e415e2edcff3/test/concat.js#L33-L35.

Do you think it's a bad idea?

davidchambers: The problem with including `Object` here is we lose the ability to indicate that a dispatching function such as `R.map` also supports plain objects. I don't feel strongly about this, though, as with JSDoc there'll always be some ambiguity.

scott-christopher: I believe there is a standard for describing duck-typed objects in JSDoc as `{a: number}`, which would describe an object with a property `a` whose value is a `number`.

So this could be described as `{Array|String|{concat: function}}`.

davidchambers: Thanks for the info, @scott-christopher! Works for me. :)

CrossEye: > So this could be described as `{Array|String|{concat: function}}`.

That sounds perfect!

src/invert.js
```diff
@@ -12,7 +12,7 @@ var keys = require('./keys');
  * @since v0.9.0
  * @category Object
  * @sig {s: x} -> {x: [ s, ... ]}
- * @param {Object} obj The object or array to invert
+ * @param {Object|*} obj The object or array to invert
```
CrossEye: `*` here to me implies that this will take _anything_, including a boolean, a number, etc.  While these won't throw errors, I'm not sure we want to document accidents of implementation this way.

src/invertObj.js
```diff
@@ -12,7 +12,7 @@ var keys = require('./keys');
  * @since v0.9.0
  * @category Object
  * @sig {s: x} -> {x: s}
- * @param {Object} obj The object or array to invert
+ * @param {Object|*} obj The object or array to invert
```
CrossEye: Ditto

CrossEye:
Very nice.  I think I take issue, though with some of your usage of `*`.  To me, that's to be used when a function really takes _anything_.  In many of these cases, I thought we wanted to handle, say `Object|Array` or `Object|Array|Function`.  I'd rather be explicit with these, and make it clear that `Number` or `Bool` is specifically not included.

src/invert.js
raine: My reasoning were these tests which seemed to me like we're specifically supporting those kinds of values. But I agree with you that `*` doesn't make sense for `invert` or `invertObj`.

```js
  it('returns an empty object when applied to a primitive', function() {
    eq(R.invert(42), {});
    eq(R.invert('abc'), {});
  });

  it('returns an empty object when applied to null/undefined', function() {
    eq(R.invert(null), {});
    eq(R.invert(undefined), {});
  });
```

src/concat.js
raine: What about the second argument that can be an object if used with a dispatching first argument?

Like in this test:

```js
var z1 = { x: 'z1', concat: function(that) { return this.x + ' ' + that.x; } };
var z2 = { x: 'z2' };
eq(R.concat(z1, z2), 'z1 z2');
```

src/length.js
```diff
@@ -3,19 +3,20 @@ var is = require('./is');
...
+ * Returns the length of an `Array`, `String`, or arity in case of a `Function`.
  *
  * @func
  * @memberOf R
  * @since v0.3.0
  * @category List
  * @sig [a] -> Number
- * @param {Array} list The array to inspect.
- * @return {Number} The length of the array.
+ * @param {*} list The value to inspect.
```
raine: Be explicit about the values here?

src/path.js
```diff
@@ -10,7 +10,7 @@ var _curry2 = require('./internal/_curry2');
  * @category Object
  * @sig [String] -> {k: v} -> v | Undefined
  * @param {Array} path The path to use.
- * @param {Object} obj The object to retrieve the nested property from.
+ * @param {Object|*} obj The object to retrieve the nested property from.
```
raine: `{?Object}` is potentially better here.

> This indicates that the type is either the specified type, or null

src/pathOr.js
```diff
@@ -14,7 +14,7 @@ var path = require('./path');
  * @sig a -> [String] -> Object -> a
  * @param {*} d The default value.
  * @param {Array} p The path to use.
- * @param {Object} obj The object to retrieve the nested property from.
+ * @param {Object|*} obj The object to retrieve the nested property from.
```
raine: `{?Object}` is potentially better here.

> This indicates that the type is either the specified type, or null

src/propOr.js
```diff
@@ -14,7 +14,7 @@ var _has = require('./internal/_has');
  * @sig a -> String -> Object -> a
  * @param {*} val The default value.
  * @param {String} p The name of the property to return.
- * @param {Object} obj The object to query.
+ * @param {Object|*} obj The object to query.
```
raine: Same thing.

raine:
I added some commits.

src/concat.js
scott-christopher: I'd suggest using `{concat: function}` for both params to indicate that the types of each param should be the same (even if `concat` isn't actually used in the second param). Furthermore, the return type should be the same as both the params for `concat` too.

That particular test could be updated to something like the following to remove some of the ambiguity:
```js
function a(x) {
  return {
    x: x,
    concat: that => a(x + ' ' + that.x)
  };
}
eq(R.concat(a('z1'), a('z2')).x, 'z1 z2');
```

raine: Is this what you're suggesting, @scott-christopher?

```
@param {Array|String|{concat: function}} a
@param {Array|String|{concat: function}} b
@return {Array|String|{concat: function}} a
```

scott-christopher: Correct (just without the trailing `a` in the `@return` line). It would be nice if JSDoc allowed us to indicate that the other types should be the same as whatever is chosen for the first param, but we'll have to rely on our `@sig` tags for that.

src/path.js
davidchambers: I like `{?Object}`, if it's a real thing. The description you quoted sounds perfect.

davidchambers:
Do you plan to squash these commits, @raine?

davidchambers:
Could you resolve the merge conflicts, @raine? I'm keen to merge this pull request. :)

buzzdecafe:
((:bell:))

CrossEye:
Closing as stale.  This has sat around ignored for far too long as the library progressed.  Let's not ever do this again!

PR #1495, opened 2015-11-03
Add juxt function

samueller:
Adding a `juxt` function as discussed in issue #986. This supports passing variadic functions in by passing in the right number of arguments in an array as the second parameter:

```
juxt([noArityFunc], [6])
juxt([arity1Func], [6])
juxt([arity2Func], [6, 7])
juxt([arity3Func], [6, 7, 8])
```

### DIFF
--- a/src/juxt.js
+++ b/src/juxt.js
@@ -1,0 +1,25 @@
+var __ = require('./__');
+var _curry1 = require('./internal/_curry1');
+var apply = require('./apply');
+var map = require('./map');
+
+
+/**
+ * juxt applies a list of functions to a list of values.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig [(a, b, ..., m) -> n] -> ((a, b, ..., m) -> [n])
+ * @param {Array} fns An array of functions
+ * @return {Function} A function that returns a list of values after applying each of the original `fns` to its parameters.
+ * @example
+ *
+ *      var range = R.juxt([Math.min, Math.max]);
+ *      range(3, 4, 9, -3); //=> [-3, 9]
+ */
+module.exports = _curry1(function juxt(fns) {
+  return function() {
+    return map(apply(__, arguments), fns);
+  };
+});
--- a/test/juxt.js
+++ b/test/juxt.js
@@ -1,0 +1,41 @@
+var R = require('..');
+var eq = require('./shared/eq');
+
+
+describe('juxt', function() {
+  function hello() { return 'hello'; }
+  function bye() { return 'bye'; }
+
+  it('works with no functions and no values', function() {
+    eq(R.juxt([])(), []);
+  });
+
+  it('works with no functions and some values', function() {
+    eq(R.juxt([])(2, 3), []);
+  });
+
+  it('works with 1 function and no values', function() {
+    eq(R.juxt([hello])(), ['hello']);
+  });
+
+  it('works with 1 function and 1 value', function() {
+    eq(R.juxt([R.add(3)])(2), [5]);
+  });
+
+  it('works with 1 function and some values', function() {
+    eq(R.juxt([R.multiply])(2, 3), [6]);
+  });
+
+  it('works with some functions and no values', function() {
+    eq(R.juxt([hello, bye])(), ['hello', 'bye']);
+  });
+
+  it('works with some functions and 1 value', function() {
+    eq(R.juxt([R.multiply(2), R.add(3)])(2), [4, 5]);
+  });
+
+  it('works with some functions and some values', function() {
+    eq(R.juxt([R.add, R.multiply])(2, 3), [5, 6]);
+  });
+
+});


CrossEye:
There's an interesting API question here.  Which is better,

```js
    juxt([arity2Func], [6, 7])
```

or

```js
    juxt([arity2Func])(6, 7)
```

Obviously in favor of the first one is that it is not just a function-returning function, but one which also allows you to pass the data as well.  It's just not clear to me how often one would want to do this; it doesn't seem a likely case to me, and it adds the additional overhead of wrapping simple parameters in an array.  The advantage of the second one is the simplicity of dealing with straightforward functions that don't require additional array wrappers.

I discussed both possibilities in #986, and I think I'd prefer the second one.  What do others think?

One other minor note.  The signature seems wrong.  Unfortunately, I don't think there is a simple signature for this, not with polyadic functions allowed.  Perhaps this would do:

```js
   juxt :: [(a, b, ..., m) -> n] -> (a, b, ..., m) -> [n]
```

Or if we change to the version I'm suggesting, then

```js
   juxt :: [(a, b, ..., m) -> n] -> ((a, b, ..., m) -> [n])
```

samueller:
@CrossEye, my thinking around `juxt` being a 2-arity function that takes an array as its second argument is:
1. Ramda's `ap` function behaves this way
2. You mentioned is issue #986 "we've been moving towards a goal of having no variadic functions at all in Ramda."

I do like your suggestion here though. So I'd love to be able to compose `juxt`:
```
var magnitudeRange = R.compose(R.juxt([R.min, R.max]), R.map(Math.abs))
magnitudeRange([-9, 3]) // [3, 9]
magnitudeRange([3, 9]) // [3, 9]
```
That would require the 2nd argument of `juxt` to be an array. However, we could simply wrap `juxt` with `R.apply` to allow this instead and have `juxt` itself be a variadic function:
```
var magnitudeRange = R.compose(R.apply(R.juxt([R.min, R.max])), R.map(Math.abs))
magnitudeRange([-9, 3]) // [3, 9]
magnitudeRange([3, 9]) // [3, 9]
```
And conversely, a variadic `juxt` wouldn't have to wrap an array around arguments if we actually want to pass an array to our juxtaposed functions. So my vote goes to a variadic `juxt`.

Regarding the function signature. I don't know the exact syntax for these function signatures, but does this make sense?
```
juxt :: [(a, b, ..., m) -> n] -> a -> b -> ... -> m -> [n]
```

CrossEye:
My alternative API was not a variadic `juxt` function.  We definitely have been moving away from that.  I was suggesting that `juxt` simply accepts a list of functions, and returns a new (possibly variadic) function.  It wouldn't allow you to pass the data to operate upon as a parameter in the first call.  It would work the same way functions like `curry` do.

In this version,

```js
var magnitudeRange = R.compose(R.juxt([R.min, R.max]), R.unapply(R.map(Math.abs)))
magnitudeRange(-9, 3) // [3, 9]
magnitudeRange(3, 9) // [3, 9]
```

The sig is somewhat tricky because we don't know if the input functions are curried.  So the input would probably have to remain `[(a, b, ..., m) -> n]`.  But we probably could choose to curry the resulting function.  So instead of 

```js
   juxt :: [(a, b, ..., m) -> n] -> ((a, b, ..., m) -> [n])
```

we might have

```js
   juxt :: [(a, b, ..., m) -> n] -> ( a -> b -> ... -> m -> [n])
```

The first one reads, essentially, "accepts a list of variadic functions of the same type and returns a function that accepts the same types of parameters as those functions and returns a list of values of the same return type as those functions."  The second one is very similar except that it starts "accepts a list of variadic functions of the same type and returns a **curried** function that..."

The question of whether to curry the resulting function is interesting, and maybe difficult, but it's definitely secondary to the question of whether to make this function require two passes, the way, say, `compose` does, or whether to include the arguments that we operate on via currying.  My preference is for the former.  Can you think of good uses for the latter?

samueller:
I understand your API dilemma now. Should `juxt` **only** return a function that will then take in the parameters to be passed to the original list of functions *or* should `juxt` also be able to immediately calculate the result by accepting both the list of functions and the list of parameters to pass to them in one call.

I had assumed you meant the alternative was to make `juxt` variadic because your example of `juxt([arity2Func])(6, 7)` looked like a curried variadic `juxt`.

Because of Ramda's philosophy of currying every function, it seems natural to allow `juxt` to have 2 parameters, the list of functions and the list of parameters to pass in to those functions. You get the ability to get a result from `juxt` with just 1 pass virtually for free. And it behaves very similarly to existing Ramda functions like `ap`.

A use case for why we'd want a one-time call to `juxt`, off the top of my head, might be when a user selects some operations they want performed on some existing data:
```
const basicStatFunctions = [range, mean, median, mode, expectedValue, standardDeviation]
R.juxt(R.map(R.nth(R.__, basicStatFunctions), userSelectedStatFunctions), data)
```

CrossEye:
> I understand your API dilemma now. Should `juxt` **only** return a function that will then take in the parameters to be passed to the original list of functions *or* should `juxt` also be able to immediately calculate the result by accepting both the list of functions and the list of parameters to pass to them in one call.

Yes.  The advantage to only returning a function.is that such a function can simply accept values and not require them to be wrapped in an array.  I would expect the majority of uses of `juxt` would be to create unary functions, and forcing the user to wrap such unary parameters in an array is unfortunate.

> I had assumed you meant the alternative was to make `juxt` variadic because your example of `juxt([arity2Func])(6, 7)` looked like a curried variadic `juxt`.

We've found no way to to curry variadic functions.  We don't even think about trying it any more.

> Because of Ramda's philosophy of currying every function, it seems natural to allow `juxt` to have 2 parameters, the list of functions and the list of parameters to pass in to those functions. You get the ability to get a result from `juxt` with just 1 pass virtually for free. And it behaves very similarly to existing Ramda functions like `ap`.

A number of Ramda's functions return functions without applying parameters, though.  Some that come immediately to mind are `compose*`, `pipe*`, `converge`, `useWith`, `both`, `either`, `complement`, `allPass`, `anyPass`, `nAry`, `binary`, `unary`, `cond`, `ifElse`, `construct`, `invoker`.  I'm sure there are others.

So to me the essential question is really whether, when I call `fn = juxt([arit2fns...])`, whether the correct way to call it is `fn([param1, param2])` or as `fn(param1, param2)`.  I definitely prefer the latter.  If it's the former, then we might as well go ahead and curry in the list of parameters as a small improvement.  If it's the latter, we can't, but I don't find it a great loss.

samueller:
Is the issue of making `juxt` a curried variadic function that we don't know how many parameters each function passed in as the first parameter to `juxt` takes? And calling `.length` on 1 or more of those functions isn't reliable? What about simply currying `juxt` if only an array is passed to `juxt`? If even a single extra parameter is passed in, then `juxt` is not curried.

This would seem to get us the benefits of both approaches, except I understand you want to move away from variadic functions in Ramda.

CrossEye:
I would have no problem if 

```js
R.length(R.juxt(fns)) ≍R.apply(Math.max)(R.map(R.length, fns))
```

but that when we call it, we pass all parameters supplied to every function in case one of the functions reported its arity poorly.

But I don't know whether I would curry the resulting function.  Probably, because I can't see any harm, but I would want to think about it a bit more.  Still, that function would would not actually be variadic.  It would have a specific arity.

> What about simply currying `juxt` if only an array is passed to `juxt`? If even a single extra parameter is passed in, then `juxt` is not curried.

I'm afraid I don't follow.  My suggested API is an uncurried `juxt` function that takes only a list of functions.  (There is no reason to curry unary functions.)  It returns a new function with an arity equal to the maximum arity of the supplied functions.  The returned function is probably curried (although I'd want to think about that a bit.)  We don't generally decide to curry or not based on the data supplied; why would you choose to do so here?

samueller:
My thinking was to have the benefits of getting a returned function from `juxt` that can be called like any other function with multiple parameters. Unary juxtaposed functions get a really simple call syntax, `juxtaposedFunc(value)`. And also have the benefits of being able to call the `juxt` function in 1 pass.

It does seem like the simplest solution, from the user's perspective, is to treat `juxt` like `compose`.

CrossEye:
@samueller:

I'm afraid I didn't quite follow your last.  The final sentence,

> It does seem like the simplest solution, from the user's perspective, is to treat juxt like compose.

definitely sounds as though it agrees with what I've been saying, but I'm not so sure about the rest.

samueller:
Yes @CrossEye my last sentence was agreeing with what you've been saying. As I think about the options discussed here, there are benefits to each. But then I look at the description of `compose`, a similar function in that it takes a list of functions and returns a function which takes in a list of parameters. `compose` is so simple, straightforward and easy to use. `juxt` and `compose`, in Clojure, are implemented as only taking functions and returning a function (the way you've been suggesting).

CrossEye:
@samueller: Do you want to update the PR, then?

samueller:
Yes, I'll update the PR right now. So I'm going to make `juxt` a unary function which only takes as input an array of functions. This is a departure from Ramda's `compose` and `pipe` functions as they simply take functions as input through normal parameters instead of an array. I understand this change is towards the goal of no variadic functions in Ramda.

samueller:
Having a problem with running `sudo npm test`. Everything worked perfect when I submitted this pull request. I modified `src/juxt.js` and `test/juxt.js` just now, except this time I got 2 assertion failures:
> 1) invariants -- applying function f with length n (where n > 0) to no arguments gives function with length n:
>       AssertionError: '0' === '1'
> 2) invariants -- applying function f with length n (where n > 0) to R.__ gives function with length n:
>       AssertionError: '0' === '1'

So I pulled in the latest changes from the upstream ramda/ramda repository and rebased. I still get the assertion failures even though I made no changes to anything except the above `juxt`s.

I'll have to look into this later, but any suggestions are certainly welcome. I can try to just redo the fork as well.

CrossEye:
Although there's talk in several recent issues that might mean removing it, one of the invariants we support is that for any Ramda function, `fn`,

    fn(R.__) ≍ fn

We do this by applying `_curry1` to unary functions.   Perhaps that's what's missing?

buzzdecafe:
> Having a problem with running `sudo npm test`

you shouldn't have to be root to run the test suite

samueller:
wrapping the `juxt` function with `_curry1` worked, thanks. I didn't realize this was necessary. @buzzdecafe, I previously needed to use `sudo`, but I just tried simply `npm test` now and that worked.

CrossEye:
Thanks.  Looking good!

Could you squash the commits?

Also, we don't generally include `dist/ramda.js` or `dist/ramda.min.js` in our commits.

samueller:
I didn't explicitly include `dist/ramda.*` in my commits. Not sure how to squash my commits or if I did it properly. I read the best way to squash commits is with `git reset --soft HEAD~3 && git commit`. I did that and GitHub showed 11 commits. I then read to squash with:
```
git reset --hard HEAD~12
git merge --squash HEAD@{1}
git commit
```
Now GitHub is showing 13 commits... and Travis CI is failing. I can look into this later tonight, but any tips on this would be helpful.

raine:
Sync `master` with `origin/master`, i.e. pull. Switch to `juxt` and rebase it on master with `git rebase master`. Then do an interactive rebase with `git rebase -i master` and when you get:

```
pick b6d7413 Add juxt function
pick 3710212 Updated juxt to only take an array of functions
pick 3cfd87f Updated juxt to only take an array of functions
pick c436026 Wrapped juxt function with _curry1
pick c452d7c Removed ramda.js and ramda.min.js
```

Change it like so:

```
pick b6d7413 Add juxt function
s 3710212 Updated juxt to only take an array of functions
s 3cfd87f Updated juxt to only take an array of functions
s c436026 Wrapped juxt function with _curry1
s c452d7c Removed ramda.js and ramda.min.js
```

Save and exit twice, push with `--force` and you're done.

davidchambers:
@samueller, since you have merge commits on your branch you won't have much luck rebasing (unless you are a Git master).

I suggest this approach:

```console
$ git checkout master
$ git pull upstream master  # assuming "upstream" is alias for ramda/ramda
$ git branch -m juxt juxt-backup
$ git checkout -b juxt  # create a new "juxt" branch from master
$ curl --location https://github.com/ramda/ramda/pull/1495.diff | patch -p1  # retrieve and apply patch
$ git checkout -- dist  # revert the unwanted changes to "dist"
$ git add {src,test}/juxt.js
$ git commit --message 'Add juxt function'
$ git push origin juxt --force  # update this pull request
$ git branch -D juxt-backup  # delete "backup" branch if everything went well
```

In the future, I suggest the following workflow to incorporate upstream changes:

```console
$ git checkout master
$ git pull upstream master
$ git checkout my-feature-branch
$ git rebase master
$ # resolve merge conflicts
$ git push origin my-feature-branch --force  # update pull request without introducing merge commit
```

samueller:
@davidchambers thank you for the git help, looks like your sequence of commands worked perfect. Time for me to learn git better. @raine thanks for your help too.

src/juxt.js
```diff
@@ -0,0 +1,25 @@
...
+var _curry1 = require('./internal/_curry1');
+var apply = require('./apply');
+var map = require('./map');
+
+/**
+ * juxt applies a list of functions to a list of values.
+ *
+ * @func
+ * @memberOf R
+ * @since v0.18.1
```
davidchambers: Please remove this line. We'll include the `@since` annotation when we include this file in a release.

src/juxt.js
```diff
@@ -0,0 +1,25 @@
+var __ = require('./__');
+var _curry1 = require('./internal/_curry1');
+var apply = require('./apply');
+var map = require('./map');
+
+/**
```
davidchambers: Two empty lines after imports, please. :)

samueller:
Just curious about the status of this. We'd like to write curriculum for our students that would depend on the `juxt` function.

CrossEye:
Sorry, just merged.  Most of the regulars haven't been all that regular lately.

Thanks for the ping.  And thanks for the PR!

samueller:
No problem @CrossEye. I love Ramda and now about 600 kids are going to be learning it in-depth in the context of functional programming in JavaScript thanks to you guys. Kids are going to be growing up with point-free style in no small part because of Ramda :).

davidchambers:
Very cool, @samueller!

svozza:
That is seriously cool.

PR #1498, opened 2015-11-04
doc: add @see R.filter, R.reject on R.partition

Delapouite:
Hi

As this function offers both filtered and rejected elements at the same time,  I think it makes sense to mention them in the doc.


davidchambers:
:deciduous_tree: Very nice! Thank you, @Delapouite.

Issue #1487, opened 2015-11-01
Add @see tag that references R.addIndex to appropriate List functions


Andrew-Webb:
Origin of this issue:- #1478 

I have identified the following List functions as being most worthy of the addition of an `@see` tag that references `R.addIndex` in order to clue people in to the ability to add an index (and the whole list) to the callback function:-
- `R.forEach`
- `R.map`
- `R.filter`
- `R.reject`
- `R.dropLastWhile`
- `R.takeLastWhile`
- `R.dropWhile`
- `R.takeWhile`

Criteria for inclusion:-
1. I understand what the function does(!)
2. A similar function in .NET's LINQ libraries and in F#'s List and Seq modules also provides the ability to specify the index (via method overloads in LINQ and via separately-named functions in F#).
3. A reciprocal function should be included (e.g. if `R.dropWhile` is included then so should `R.takeWhile`).
4. I have found from my own experience that knowing the index in these kinds of operations is occasionally very useful.

Criteria for exclusion:-
1. I am unfamiliar with / do not understand the function (e.g. `R.transduce`).
2. The function's callback already takes the index.
3. Unclear benefits of including the index, often coupled with lack of index param for that operation in .NET LINQ and F# List and Seq modules.  An example is `R.reduce`.

Please do review the list and suggest any additions / removals - after which I will open a PR.

Thanks.

CrossEye:
I would suggest that we also include `reduce` and `reduceRight`, so that we capture at least those things changed in #1061.  One of the main reasons to have this function is that people are used to having these parameters available on `Array.prototype` functions.

> A reciprocal function should be included (e.g. if R.dropWhile is included then so should R.takeWhile).

Sure, but don't let the absence of a reciprocal function stop us from adding the @see tag.

Andrew-Webb:
> I would suggest that we also include reduce and reduceRight

If we include `R.reduce` and `R.reduceRight`, and we already have `R.map`, then I guess it makes sense to also include `R.mapAccum` and `R.mapAccumRight`.

> Sure, but don't let the absence of a reciprocal function stop us from adding the @see tag.

No way!

tycho01:
I noticed `addIndex` can also be used for `groupBy`; who knows what else still... it looks like this may also affect the other `*by` functions minus `pickBy`: that'd seem to be `countBy`, `eqBy`, `groupBy`, `indexBy`, `maxBy`, `minBy`, `pickBy`, `reduceBy`, `sortBy`, and `uniqBy`.

PR #1499, opened 2015-11-06
Add @see ref to R.addIndex to 12 List functions

Andrew-Webb:
Added @see doc tag referencing R.addIndex function to 12 identified List
functions whose callback function might reasonably take extra args
specifying the current index and the whole list.  #1487


davidchambers:
:deciduous_tree: Thanks!

PR #1504, opened 2015-11-08
Fix param/return documentation for internal curryN

davezuko:
I know this is beyond low-priority, but I caught it while digging through the source and figured it wouldn't hurt.

### DIFF
--- a/src/internal/_curryN.js
+++ b/src/internal/_curryN.js
@@ -7,8 +7,9 @@ var _arity = require('./_arity');
  * @private
  * @category Function
  * @param {Number} length The arity of the curried function.
- * @return {array} An array of arguments received thus far.
+ * @param {Array} An array of arguments received thus far.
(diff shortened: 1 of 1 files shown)


src/internal/_curryN.js
```diff
@@ -7,8 +7,9 @@ var _arity = require('./_arity');
  * @private
  * @category Function
  * @param {Number} length The arity of the curried function.
- * @return {array} An array of arguments received thus far.
+ * @param {Array} An array of arguments received thus far.
  * @param {Function} fn The function to curry.
+ * @return {Function} the curried function.
```
davidchambers: `s/the/The/`

davezuko:
@davidchambers Sorry about that, I could have sworn I copied it from _curry2 but must not have. Fixed.

davidchambers:
:deciduous_tree: Thanks!

PR #1506, opened 2015-11-10
Add R.lensPath

gilligan:
This adds a lens focusing on an object property specified by a path. This addition would make the `prop` and `path` set of functions fully "analogous":
 
* `R.prop` - `R.path` 
* `R.propSatisfies` - `R.pathSatisfies` 
* `R.lensProp` - `R.lensPath`.

### DIFF
--- a/src/lensPath.js
+++ b/src/lensPath.js
@@ -1,0 +1,28 @@
+var _curry1 = require('./internal/_curry1');
+var assocPath = require('./assocPath');
+var lens = require('./lens');
+var path = require('./path');
+
+
+/**
+ * Returns a lens whose focus is the specified path.
+ *
+ * @func
+ * @memberOf R
+ * @category Object
+ * @typedefn Lens s a = Functor f => (a -> f a) -> s -> f s
+ * @sig [String] -> Lens s a
+ * @param {Array} path The path to use.
+ * @return {Lens}
+ * @see R.view, R.set, R.over
+ * @example
+ *
+ *      var xyLens = R.lensPath(['x', 'y']);
+ *
+ *      R.view(xyLens, {x: {y: 2, z: 3}});            //=> 2
+ *      R.set(xyLens, 4, {x: {y: 2, z: 3}});          //=> {x: {y: 4, z: 3}}
+ *      R.over(xyLens, R.negate, {x: {y: 2, z: 3}});  //=> {x: {y: -2, z: 3}}
+ */
+module.exports = _curry1(function lensPath(p) {
+  return lens(path(p), assocPath(p));
+});
--- a/test/lenses.js
+++ b/test/lenses.js
@@ -7,26 +7,34 @@ var they = it;
 
 var alice = {
   name: 'Alice Jones',
-  address: ['22 Walnut St', 'San Francisco', 'CA']
+  address: ['22 Walnut St', 'San Francisco', 'CA'],
+  pets: {dog: 'joker', cat: 'batman'}
 };
 
 var nameLens = R.lens(R.prop('name'), R.assoc('name'));
 var addressLens = R.lensProp('address');
 var headLens = R.lensIndex(0);
+var dogLens = R.lensPath(['pets', 'dog']);
 
 
 describe('view, over, and set', function() {
+
+  they('may be applied to a lens created by `lensPath`', function() {
+    eq(R.view(dogLens, alice), 'joker');
+  });
 
   they('may be applied to a lens created by `lensProp`', function() {
     eq(R.view(nameLens, alice), 'Alice Jones');
 
     eq(R.over(nameLens, R.toUpper, alice),
        {name: 'ALICE JONES',
-        address: ['22 Walnut St', 'San Francisco', 'CA']});
+        address: ['22 Walnut St', 'San Francisco', 'CA'],
+        pets: {dog: 'joker', cat: 'batman'}});
 
     eq(R.set(nameLens, 'Alice Smith', alice),
        {name: 'Alice Smith',
-        address: ['22 Walnut St', 'San Francisco', 'CA']});
+        address: ['22 Walnut St', 'San Francisco', 'CA'],
+        pets: {dog: 'joker', cat: 'batman'}});
   });
 
   they('may be applied to a lens created by `lensIndex`', function() {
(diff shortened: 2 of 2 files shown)


gilligan:
Oops, there are some mistakes in the code examples block. Will fix that asap. On vacation and thus not in front of laptop as often ;)

CrossEye:
Thank you.

We generally have a single test file for each public function.

src/lensPath.js
```diff
@@ -0,0 +1,27 @@
...
+ *
+ *      var xyLens = R.lensPath(['x', 'y']);
+ *
+ *      R.view(xyLens, {x: {y: 2, z: 3}});            //=> 2
+ *      R.set(xyLens, 4, {x: {y: 2, z: 3}});          //=> {x: 4, y: 2}
+ *      R.over(xyLens, R.negate, {x: {y: -2, z: 3}}); //=> {x: -1, y: 2}
+ */
+module.exports = _curry1(function lensPath(p) {
+  return lens(path(p), assocPath(p));
+});
```
CrossEye: Just curious whether you considered the alternative implementation:

```js
module.exports = compose(apply(compose), map(lensProp));
```

I'm not saying it's any better.  It's just an interesting possibility, building these by composing `lensProp` instances.

gilligan:
@CrossEye That sure makes sense but does not seem to be the case for any of the lens functions. I extended `test/lenses.js` because there are no separate tests for any of the predefined lenses. I don't much like the `lenses.js` tests because it tests too many things at once but i would want to such a refactoring as a separate PR.

test/lenses.js
```diff
@@ -41,16 +56,24 @@ describe('view, over, and set', function() {
 
   they('may be applied to composed lenses', function() {
     var streetLens = R.compose(addressLens, headLens);
+    var dogLens = R.compose(R.lensPath(['pets']), R.lensPath(['dog']));
+
+    eq(R.view(dogLens, alice), R.view(R.lensPath(['pets', 'dog']), alice));
```
CrossEye: This is not a test of the public interface of `lensPath`, only of how it relates to another function.  Much more appropriate would be to note

```js
eq(R.view(dogLens, alice), 'joker');
```

gilligan: But why aren't there any separate tests for lensIndex or lensProp ? The general idea of `lenses` seems to be to test lens interop.

gilligan: @CrossEye actually `test/lenses.js` is about `view`, `over`, and `set`. I am very willing to try to do some refactoring and add specific tests for lensIndex, lensProp and lensPath outside of `test/lenses.js` but preferably not as part of this PR.

Would you prefer if I did that first ?

CrossEye: No, I hadn't caught that when lenses were introduced, or when lensProp was introduced.  We do need to fix this, I believe.  I think that the unit test suite is also part of our documentation.

But don't bother with it in this PR.  I hadn't noticed it before now.

gilligan:
@CrossEye check #1510 :)

buzzdecafe:
LGTM :cow2:

test/lenses.js
```diff
@@ -41,16 +56,24 @@ describe('view, over, and set', function() {
...
 
     eq(R.view(streetLens, alice), '22 Walnut St');
 
     eq(R.over(streetLens, R.toUpper, alice),
        {name: 'Alice Jones',
-        address: ['22 WALNUT ST', 'San Francisco', 'CA']});
+        address: ['22 WALNUT ST', 'San Francisco', 'CA'],
+        pets: {
+          dog: 'joker',
+          cat: 'batman'}});
```
davidchambers: I'd prefer to see `{dog: 'joker', cat: 'batman'}` here. :)

test/lenses.js
```diff
@@ -41,16 +56,24 @@ describe('view, over, and set', function() {
...
 
     eq(R.set(streetLens, '52 Crane Ave', alice),
        {name: 'Alice Jones',
-        address: ['52 Crane Ave', 'San Francisco', 'CA']});
+        address: ['52 Crane Ave', 'San Francisco', 'CA'],
+        pets: {
+          dog: 'joker',
+          cat: 'batman'}});
   });
-
```
davidchambers: Can we keep this empty line? :)

src/lensPath.js
```diff
@@ -0,0 +1,27 @@
+var _curry1 = require('./internal/_curry1');
+var assocPath = require('./assocPath');
+var lens = require('./lens');
+var path = require('./path');
+
+/**
```
davidchambers: Two empty lines after imports, please.

davidchambers:
LGTM, aside from the few style issues I mentioned. :)

gilligan:
Great, thanks. Will fix those asap. Wonder if we can have a linting rule for the spaces after imports. I will check.

test/lenses.js
```diff
@@ -41,16 +49,21 @@ describe('view, over, and set', function() {
...
 
     eq(R.set(streetLens, '52 Crane Ave', alice),
        {name: 'Alice Jones',
-        address: ['52 Crane Ave', 'San Francisco', 'CA']});
+        address: ['52 Crane Ave', 'San Francisco', 'CA'],
+        pets: {dog: 'joker', cat: 'batman'}});
   });
-
 });
+
```
davidchambers: :space_invader:

test/lenses.js
```diff
@@ -41,16 +49,21 @@ describe('view, over, and set', function() {
...
+        address: ['22 WALNUT ST', 'San Francisco', 'CA'],
+        pets: {dog: 'joker', cat: 'batman'}});
 
     eq(R.set(streetLens, '52 Crane Ave', alice),
        {name: 'Alice Jones',
-        address: ['52 Crane Ave', 'San Francisco', 'CA']});
+        address: ['52 Crane Ave', 'San Francisco', 'CA'],
+        pets: {dog: 'joker', cat: 'batman'}});
   });
-
```
davidchambers: Can we keep this line? I'm sorry about the pedantry, but I do appreciate a tidy diff. :smile:

gilligan: D'oh. I am sorry. Again it would be nice if this could be catched by linting/style-checker so I don't have to bore /you/ with it :)

PR #1510, opened 2015-11-11
Split lens tests

### LINKED ISSUE #1509: No dedicated tests for lens functions
As mentioned in #1506 there are no separate tests for the individual lens functions, instead there is only .`test/lenses.js` which is mostly about the functionality of view/set/over and combining lenses. 

There should be separate tests for `lensProp`, `lensIndex` (and if #1506 gets merged also `lensPath`) each testing its functionality in regard to view/over/set.

gilligan:
This fixes #1509 replacing the combined `lenses.js` test with separate tests for `lensProp` and `lensIndex` each testing the correct behavior for 
* `R.view`
* `R.set`
* `R.over`
* composability
* lens laws

### DIFF
--- a/test/lensIndex.js
+++ b/test/lensIndex.js
@@ -1,6 +1,7 @@
 var R = require('..');
 var eq = require('./shared/eq');
 
+
 var testList = [{a: 1}, {b: 2}, {c: 3}];
 
 describe('lensIndex', function() {
--- a/test/lensIndex.js
+++ b/test/lensIndex.js
@@ -22,4 +22,12 @@ describe('lensIndex', function() {
       eq(R.over(R.lensIndex(2), R.keys, testList), [{a: 1}, {b: 2}, ['c']]);
     });
   });
+  describe('composability', function() {
+    it('can be composed', function() {
+      var nestedList = [0, [10, 11, 12], 1, 2];
+      var composedLens = R.compose(R.lensIndex(1), R.lensIndex(0));
+
+      eq(R.view(composedLens, nestedList), 10);
+    });
+  });
 });
--- a/test/lensIndex.js
+++ b/test/lensIndex.js
@@ -1,0 +1,25 @@
+var R = require('..');
+var eq = require('./shared/eq');
+
+var testList = [{a: 1}, {b: 2}, {c: 3}];
+
+describe('lensIndex', function() {
+  describe('view', function() {
+    it('focuses list element at the specified index', function() {
+      eq(R.view(R.lensIndex(0), testList), {a: 1});
+    });
+    it('returns undefined if the specified index does not exist', function() {
+      eq(R.view(R.lensIndex(10), testList), undefined);
+    });
+  });
+  describe('set', function() {
+    it('sets the list value at the specified index', function() {
+      eq(R.set(R.lensIndex(0), 0, testList), [0, {b: 2}, {c: 3}]);
+    });
+  });
+  describe('over', function() {
+    it('applies function to the value at the specified list index', function() {
+      eq(R.over(R.lensIndex(2), R.keys, testList), [{a: 1}, {b: 2}, ['c']]);
+    });
+  });
+});
--- a/test/lensIndex.js
+++ b/test/lensIndex.js
@@ -30,4 +30,16 @@ describe('lensIndex', function() {
       eq(R.view(composedLens, nestedList), 10);
     });
   });
+  describe('well behaved lens', function() {
+    it('set s (get s) === s', function() {
+      eq(R.set(R.lensIndex(0), R.view(R.lensIndex(0), testList), testList), testList);
+    });
+    it('get (set s v) === v', function() {
+      eq(R.view(R.lensIndex(0), R.set(R.lensIndex(0), 0, testList)), 0);
+    });
+    it('get (set(set s v1) v2) === v2', function() {
+      eq(R.view(R.lensIndex(0), R.set(R.lensIndex(0), 11, R.set(R.lensIndex(0), 10, testList))),
+         11);
+    });
+  });
 });
--- a/test/lensPath.js
+++ b/test/lensPath.js
@@ -1,0 +1,60 @@
+var R = require('..');
+var eq = require('./shared/eq');
+
+
+var testObj = {
+  a: {
+    b: 2,
+  },
+  d: 3
+};
+
+describe('lensPath', function() {
+  describe('view', function() {
+    it('focuses the specified object property', function() {
+      eq(R.view(R.lensPath(['d']), testObj), 3);
+      eq(R.view(R.lensPath(['a', 'b']), testObj), 2);
+    });
+  });
+  describe('set', function() {
+    it('sets the value of the object property specified', function() {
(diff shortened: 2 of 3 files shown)


buzzdecafe:
:cow2:
thanks

CrossEye:
Although I think this is worth doing, the code so far is unfortunately only the start.  There are four other functions that are not tested: `lens`, `view`, `over`, and `set`.

The problem, and this has been my general concern with lenses, is that these functions are very difficult to test in isolation, especially `lens` and its children such as `lensProp`, `lensIndex`, and now `lensPath`.  The others are made testable, if a bit obscure, by passing them something like `f => s => map(v => setter(v, x), f(getter, s))` for some `setter` and `getter`.

But looking at this, it's starting to dawn on me why this separation was done.

gilligan:
yeah so this is only a pragmatic first jab at the problem. Still I feel it is better than what we have right now because lenses.js does not do such a great job at documenting the behavior of the individual lenses.

Maybe the lens laws should also be tested?

Not so sure about lens/set/view/over - but they are not currently being tested either.

test/lensIndex.js
```diff
@@ -0,0 +1,45 @@
...
+      eq(R.set(R.lensIndex(0), 0, testList), [0, {b: 2}, {c: 3}]);
+    });
+  });
+  describe('over', function() {
+    it('applies function to the value at the specified list index', function() {
+      eq(R.over(R.lensIndex(2), R.keys, testList), [{a: 1}, {b: 2}, ['c']]);
+    });
+  });
+  describe('composability', function() {
+    it('can be composed', function() {
```
davidchambers: There's some redundancy here. Let's remove the nested `describe` blocks (so we're left with a top-level `describe` block containing `it` blocks).

CrossEye: I disagree.  I prefer this nesting at least when the names are written so the output looks like this:

    lensIndex
        view
            focuses list element at the specified index
            returns undefined if the specified index does not exist
        set
            sets the list value at the specified index
        over
            applies function to the value at the specified list index

The later ones are less appealing, though.

davidchambers: Fair enough. I was focused on the test _file_ rather than the test _output_. I don't have strong feelings on this matter.

gilligan:
@davidchambers @CrossEye In my opinion the describe blocks do a good at adding structure especially to the output. Of course I would change that if there is real opposition. And you are right of course that its less interesting in the cases where there is only a single assertion in a describe block. 

If #1506 gets merged i will update this PR with an appropriate test for lensPath

buzzdecafe:
#1506 is merged

gilligan:
@buzzedcafe aye, maybe i can adapt this tonight then!

gilligan:
@davidchambers @buzzdecafe @CrossEye I updated the PR with a test for lensPath - please have a look :)

davidchambers:
It appears we're missing a test for `R.lensPath([])`.

gilligan:
@davidchambers good catch, thanks. I added some.

test/lensPath.js
```diff
@@ -0,0 +1,63 @@
...
+      eq(R.view(R.lensPath(['d']), testObj), 3);
+      eq(R.view(R.lensPath(['a', 'b']), testObj), 2);
+      eq(R.view(R.lensPath([]), testObj), testObj);
+    });
+  });
+  describe('set', function() {
+    it('sets the value of the object property specified', function() {
+      eq(R.set(R.lensPath(['d']), 0, testObj), {a: {b: 2}, d: 0});
+      eq(R.set(R.lensPath(['a', 'b']), 0, testObj), {a: {b: 0}, d: 3});
+      eq(R.set(R.lensPath([]), 0, testObj), testObj);
```
davidchambers: Should this evaluate to `0` rather than `{a: {b: 2}, d: 3}`? It seems to me that `R.lensPath([])` should create a lens whose focus is the entire data structure. Setting the value of the focus should therefore replace the entire value.

gilligan: Yeah, i was wondering what you would say about this. The thing is: `R.path([], {a:1})` returns `{a: 1}` and since lensPath is just `lens(R.path(p), R.assocPath(p))` we get the current behaviour.

I suppose I should change the lensPath implementation then.

gilligan: Then again what does it actually mean to focus on an empty path, doesn't that mean we are not actually focusing anything? Not obvious to me that the empty path would yield the whole object.

davidchambers: I'm interested to know whether we'd get the behaviour I'm proposing were we to define `lensPath` via composition of lenses: `compose(lensProp('x'), compose(lensProp('y'), lensProp('z')))`.

davidchambers: > Then again what does it actually mean to focus on an empty path, doesn't that mean we are not actually focusing anything? Not obvious to me that the empty path would yield the whole object.

Imagine standing in the entrance of a building. You are told to cross the foyer, ascend the stairs to the first floor, turn left, walk down the hallway and enter the second door on the right. You've been given a "path" to a "focus". Your view of the building is now focused on the room to which you were directed. Had the "path" been instructions to remain standing in the building's entrace, your view would be of the entire building. This is how I conceptualize `[]`.

CrossEye: I think `assocPath` gets it wrong here.  And since it's handling for the empty path is an arbitrary decision, it would be easy enough to reverse.

test/lensProp.js
```diff
@@ -0,0 +1,56 @@
...
+  });
+  describe('composability', function() {
+    it('can be composed', function() {
+      var nestedObj = {a: {b: 1}, c:2};
+      var composedLens = R.compose(R.lensProp('a'), R.lensProp('b'));
+
+      eq(R.view(composedLens, nestedObj), 1);
+    });
+  });
+  describe('well behaved lens', function() {
```
boxofrox: Does it make sense to refactor a `var lens = R.lensProp('a');` at the top of this `describe` context and use `lens` below in each `it` block?

gilligan: I dunno.. `R.lensProp('a')` is not *that* bad and If I  only used `lens` you wouldn't immediately see what it actually is/does so I prefer to be explicit here.

gilligan:
Any other thoughts on this ? @buzzdecafe or @CrossEye ?

CrossEye:
I should have time to look at it tomorrow.

buzzdecafe:
LGTM

test/lensProp.js
CrossEye: I'm generally in favor of being more explicit in tests, even at the cost of being somewhat less DRY.  But I would actually prefer a third path, which is more verbose, perhaps, than either of these:

```js
it('set s (get s) === s', function() {
  var lens = R.lensProp('a');
  eq(R.set(lens, R.view(lens, testObj), testObj), testObj);
});
```

The point here, is that the test is showing how a specific `lens` operates, and not how several different `lens` instances with similar definitions interact.

I prefer this to the suggestion from @boxofrox because when I'm testing a specific function, I want the individual tests to explicitly **use*** that function.

Other than that I'm :+1: on this change.

gilligan:
@CrossEye I just realized this one was still open. Can you have a look? :)

CrossEye:
Sorry, not been very engaged in this.

PR #1511, opened 2015-11-11
refactor example of liftN

safareli:
- in order to use arguments as expected we need to use normal function: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Lexical_arguments
- use R.sum for simplicity

### DIFF
--- a/src/liftN.js
+++ b/src/liftN.js
@@ -20,7 +20,7 @@ var map = require('./map');
  * @see R.lift
  * @example
  *
- *      var madd3 = R.liftN(3, R.curryN(3, () => R.reduce(R.add, 0, arguments)));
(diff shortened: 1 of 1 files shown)


safareli:
anything on this?

iofjuupasli:
Why not rest?

```js
(...args) => R.sum(args)
```

safareli:
@iofjuupasli +1 actualy yap why not, did not thought about that at first

davidchambers:
:deciduous_tree:

CrossEye:
Somehow I missed this when it came through.  Thanks!

safareli:
Welcome :3

Issue #1508, opened 2015-11-11
R.contains is quite slow


greggigon:
[https://jsperf.com/ramda-contains-vs-indexof](https://jsperf.com/ramda-contains-vs-indexof) simple test of looking up for a single char in the Array of chars.

Used **R.contains** agains **Array.indexOf > -1**

I can see some performance issues raised in the last few months. Not sure they are related.

buzzdecafe:
probably related. see https://github.com/ramda/ramda/issues/1449 we may be seeing some improvements presently

CrossEye:
Note that part of the reason for the difference is the difference in the APIs:

```js
var haystack = [{a: 1}, {b: 1}, {c: 1}, {d: 1}, {e: 1}, 
                {f: 1}, {g: 1}, {h: 1}, {i: 1}, {j: 1}, 
                {k: 1}, {l: 1}, {m: 1}, {n: 1}, {o: 1}, 
                {p: 1}, {q: 1}, {r: 1}, {s: 1}, {t: 1}, 
                {u: 1}, {v: 1}, {w: 1}, {x: 1}, {y: 1}, 
                {z: 1}];

var needles = [{a: 1}, {b: 1}, {c: 1}, {d: 1}, {e: 1}, 
               {f: 1}, {g: 1}, {h: 1}, {i: 1}, {j: 1}];

var nativeContains = function(needle, haystack) {
  return haystack.indexOf(needle) > -1
};

R.map(needle => nativeContains(needle, haystack), needles);
//=> [false, false, false, false, false, false, false, false, false, false]

R.map(needle => R.contains(needle, haystack), needles);
//=> [true, true, true, true, true, true, true, true, true, true]
```

`Array.prototype.indexOf` does not recognize that `{a: 1}` is the same as `{a: 1}`.  When doing functional programming, reference equality should mean little.  It's value equality that's important.

Of course it we were using shared references, there is no difference:

```js
var needles2 = haystack.slice(0, 10);

R.map(needle => nativeContains(needle, haystack), needles2);
//=> [true, true, true, true, true, true, true, true, true, true]

R.map(needle => R.contains(needle, haystack), needles2);
//=> [true, true, true, true, true, true, true, true, true, true]
```
  
Nonetheless, we do have to find a way to note when we're searching a list of value types, and not attempt a more complex `equals` operation on them.  Any suggestions of what the API for that would look like?

scott-christopher:
[...]

Issue #1503, opened 2015-11-07
R.curry slows down resulting function way too much


asaf-romano:
Altering @niloy's test on #1500, I found out that R.curry has a dramatic influence on performance, even when the result curried function is called with all of the arguments at once:

``` js
var arr = R.times(Math.random, 20000);

console.time("Object.is");
R.uniqWith(Object.is, arr);
console.timeEnd("Object.is");

console.time("R.curry(Object.is)");
R.uniqWith(R.curry(Object.is), arr);
console.timeEnd("R.curry(Object.is)");
```

results in latest Firefox nightlty:
Object.is: timer started repl.js line 123 > eval:2
Object.is: **276.76ms** repl.js line 123 > eval:4
R.curry(Object.is): timer started repl.js line 123 > eval:5
R.curry(Object.is): **68341.41ms**

asaf-romano:
REPL link (remove ``throw 1`` and get some coffee): http://bit.ly/1MPu10K

houd1ni:
Why you curry it every time before calling? Try to save result function before :)

asaf-romano:
I don't curry it each time... the expression is evaluated right away, i.e. before it's passed to R.uniqWith. It makes no difference if you set in on a variable.

houd1ni:
I see. just deleted the comment.

asaf-romano:
Sigh, I just realized the same benchmark was done independently on #1449

houd1ni:
Yeah, that issue is a starting point to think about (maybe) some compromises between reference and value equity. Looks, like sometimes it isn't necessary to compare values. But I'm going to deeply learn that thread.

scott-christopher:
Some improvements to the currying functions when dealing with primitive arguments in #1512 have just landed in master which will give this a slight improvement.

Using an `arr` size of 5000 in the example above results in:
```
Object.is: timer started
Object.is: 140.62ms
R.curry(Object.is): timer started
R.curry(Object.is): 1081.84ms
```

Unfortunately, while support for placeholders exist in the currying functions there will continue to be a slight hit in performance. On a positive note, using `R.uniq` from the current HEAD of master for the above example now results in:
```
R.uniq: timer started
R.uniq: 2.54ms
```

buzzdecafe:
[...]

Issue #1500, opened 2015-11-06
uniq is slow


krzychukula:
We had to replace:
```
var getIds = R.map(R.path(['id']));
var oneIds = getIds(listOne);
var twoIds = getIds(listTwo);
var ids = R.uniq(R.concat(oneIds, twoIds));
```
with:
```
var idsHash = {};

listOne.forEach(function(item) {
          idsHash[item.id] = 1;
});

listTwo.forEach(function(item) {
          idsHash[item.id] = 1;
});

var ids = Object.keys(idsHash);
```

The former for big amounts of data blocked node event loop for ridiculous amouts of time (something like 25minutes).

Did I misuse uniq?

honzabrecka:
Yes, it's slow because it uses many while cycles O(n) whereas you are using a hash map O(1).

krzychukula:
@honzabrecka I realized that :) Imho it's a bug, but maybe there is a reason that it's not possible to fix without breaking api? 
I would expect uniq to be O(1)...

iofjuupasli:
> I would expect uniq to be O(1)...

But it just impossible for two lists.

O(1) possible only for checking one value in hash map.
When you have many values and hash - you have O(n), where n is length of values list.
With two lists you should compare each value O(n1) from list two with list one O(n2), and only if its uniq insert it O(1). So resulting complexity is O(n1 * n2).
In case when you can use hash map (you have uniq string keys), then you can insert O(1) each value from list one O(n1) to hash and then same for list two O(n2). Resulting complexity is O(n1 + n2). Which is much better.

But uniq doesn't know that you can use hash keys.

However it seems possible to create such method.

krzychukula:
Sorry, I don't get why you have to get O(n1 * n2) complexity :(
Why it can't always treat values as hashable? Can you give me an example?

CrossEye:
I've known for a while that `uniq` is slow.  It's been mentioned before, and I've been bitten by it myself.  It's on my list.  The real trouble is that Ramda is checking for value equality, not reference equality.  Even its `O(n^2)` algorithm would be substantially faster if we would allow ourselves the luxury of comparing references only.  But that's not the Ramda way.

Probably the best bet would be to find a good way to hash values.  The most likely candidate is our `toString` implementation, but I don't know just how robust it is.  Perhaps @davidchambers can tell us if he thinks it's up to this job.

[...]

Issue #1449, opened 2015-10-18
R.uniqWith(R.identical) is slow


raine:
```js
const taim = require('taim')
const R = require('ramda')
const _ = require('lodash')

const N = 20000;
const RAND_RANGE = [0, 5000];
const rand = R.partial(_.random, RAND_RANGE)
const arr = R.map(R.call, R.repeat(rand, N))

const uniqWithIdentical = R.uniqWith(R.identical)

taim('uniqWithIdentical', uniqWithIdentical)(arr)
taim('_uniq', _.uniq)(arr)
```

```
uniqWithIdentical took 15 s
_uniq took 5.31 ms
```

The larger the max end of `RAND_RANGE` is, i.e. the bigger the variability in values, the slower it gets, while lodash's `uniq` performs much the same.

CrossEye:
Obviously it's getting to be time we take up performance.

I still have one major API change I'm looking to introduce -- didn't get to it over the weekend, but should have time in the next few evenings.  After that, I'd like to start looking at where we are slow in comparison to other similar libraries or simply to our own expectations.

There will be places where we will never meet the speed of something like lodash, as our emphasis on value equality rather than reference equality implies significantly more processing, but this case should not cause huge slow-downs, which makes me thing that we have an algorithmic issue.

I had another recently.  I did an `invert` on an object with about 45,000 keys mapping to about 125 unique values.  That took tens of seconds to run -- don't remember the exact number.  But it was far higher than I believe it should have been.  Since it was a one-off, I didn't bother investigating too deeply at the time, but it's clearly doing something `O(n2)` when there should be significantly faster methods available.

I think we're hitting the time when we need to address these issues.

scott-christopher:
We previously had a crack at using `Set` where possible, but it proved troublesome maintaining value-based equality semantics (see #1142 & #1188).

We _could_ optimise the case where `R.identical` is provided as the equality predicate, but it's a route we haven't traditionally gone down.
e.g.
```js
_curry2(function uniqWith(pred, list) {
  var idx = 0,
      len = list.length,
      result = [],
      item, set;

  if (pred === identical && typeof Set == 'function') {
    set = new Set();
    while (idx < len) {
      item = list[idx];
      if (set.size !== set.add(item).size) {
        result[result.length] = item;
      }
      idx += 1;
    }
  } else {
    while (idx < len) {
      item = list[idx];
      if (!_containsWith(pred, item, result)) {
        result[result.length] = item;
      }
[...]

PR #1512, opened 2015-11-11
Perfomance improvements to `uniq` and `contains`

scott-christopher:
Addresses #1449, #1500, #1503 & #1508 regarding performance concerns with `uniq` and `contains`.

The main changes include:
* Refactoring the various internal currying functions to ensure the args are `Object`s before checking for the placeholder property, preventing unnecessary boxing/unboxing of primitives.
* Some tweaks to the conditional layout of the internal currying functions to reduce redundant conditional tests.
* `_contains` and `uniqBy` now make use of a native `Set` where possible – given the history of this, I'd appreciate people paying close attention here to ensure the existing semantics of `equals` are upheld.

I have also included a simple benchmark script for `uniq`, which produced the following results:

Current HEAD of master:
```
│ uniq                   │ Hertz                  │ Margin of Error        │
│ uniqBy(id, arr1)       │ 21.52                  │ 0.60%                  │
│ uniqWith(equals, arr1) │ 21.81                  │ 0.68%                  │
│ uniq(arr1)             │ 21.66                  │ 0.81%                  │
│ uniqBy(id, arr2)       │ 8.73                   │ 0.72%                  │
│ uniqWith(equals, arr2) │ 8.72                   │ 0.58%                  │
│ uniq(arr2)             │ 8.67                   │ 0.83%                  │
│ uniqBy(id, arr3)       │ 692                    │ 0.55%                  │
│ uniqWith(equals, arr3) │ 695                    │ 0.60%                  │
│ uniq(arr3)             │ 686                    │ 0.72%                  │
```
This branch:
```
│ uniq                   │ Hertz                  │ Margin of Error        │
│ uniqBy(id, arr1)       │ 94.13                  │ 0.57%                  │
│ uniqWith(equals, arr1) │ 40.21                  │ 0.55%                  │
│ uniq(arr1)             │ 87.88                  │ 0.62%                  │
│ uniqBy(id, arr2)       │ 21,299                 │ 0.45%                  │
│ uniqWith(equals, arr2) │ 25.60                  │ 0.46%                  │
│ uniq(arr2)             │ 18,967                 │ 0.65%                  │
│ uniqBy(id, arr3)       │ 787                    │ 0.51%                  │
│ uniqWith(equals, arr3) │ 808                    │ 0.53%                  │
│ uniq(arr3)             │ 783                    │ 0.61%                  │
```

### DIFF
--- a/src/internal/_curry1.js
+++ b/src/internal/_curry1.js
@@ -1,3 +1,6 @@
+var _isPlaceholder = require('./_isPlaceholder');
+
+
 /**
  * Optimized internal one-arity curry function.
  *
@@ -8,9 +11,7 @@
  */
 module.exports = function _curry1(fn) {
   return function f1(a) {
-    if (arguments.length === 0) {
-      return f1;
-    } else if (a != null && a['@@functional/placeholder'] === true) {
+    if (arguments.length === 0 || _isPlaceholder(a)) {
       return f1;
     } else {
       return fn.apply(this, arguments);
--- a/src/internal/_curry2.js
+++ b/src/internal/_curry2.js
@@ -1,4 +1,5 @@
 var _curry1 = require('./_curry1');
+var _isPlaceholder = require('./_isPlaceholder');
 
 
 /**
@@ -11,22 +12,17 @@ var _curry1 = require('./_curry1');
  */
 module.exports = function _curry2(fn) {
   return function f2(a, b) {
-    var n = arguments.length;
-    if (n === 0) {
-      return f2;
-    } else if (n === 1 && a != null && a['@@functional/placeholder'] === true) {
-      return f2;
-    } else if (n === 1) {
-      return _curry1(function(b) { return fn(a, b); });
-    } else if (n === 2 && a != null && a['@@functional/placeholder'] === true &&
-                          b != null && b['@@functional/placeholder'] === true) {
-      return f2;
-    } else if (n === 2 && a != null && a['@@functional/placeholder'] === true) {
-      return _curry1(function(a) { return fn(a, b); });
-    } else if (n === 2 && b != null && b['@@functional/placeholder'] === true) {
-      return _curry1(function(b) { return fn(a, b); });
-    } else {
-      return fn(a, b);
+    switch (arguments.length) {
+      case 0:
+        return f2;
+      case 1:
+        return _isPlaceholder(a) ? f2
+             : _curry1(function(_b) { return fn(a, _b); });
+      default:
+        return _isPlaceholder(a) && _isPlaceholder(b) ? f2
+             : _isPlaceholder(a) ? _curry1(function(_a) { return fn(_a, b); })
+             : _isPlaceholder(b) ? _curry1(function(_b) { return fn(a, _b); })
+             : fn(a, b);
     }
   };
 };
--- a/src/internal/_curry3.js
+++ b/src/internal/_curry3.js
@@ -1,5 +1,6 @@
 var _curry1 = require('./_curry1');
 var _curry2 = require('./_curry2');
+var _isPlaceholder = require('./_isPlaceholder');
 
 
 /**
@@ -12,43 +13,26 @@ var _curry2 = require('./_curry2');
  */
 module.exports = function _curry3(fn) {
   return function f3(a, b, c) {
-    var n = arguments.length;
-    if (n === 0) {
-      return f3;
-    } else if (n === 1 && a != null && a['@@functional/placeholder'] === true) {
-      return f3;
-    } else if (n === 1) {
-      return _curry2(function(b, c) { return fn(a, b, c); });
-    } else if (n === 2 && a != null && a['@@functional/placeholder'] === true &&
-                          b != null && b['@@functional/placeholder'] === true) {
-      return f3;
-    } else if (n === 2 && a != null && a['@@functional/placeholder'] === true) {
-      return _curry2(function(a, c) { return fn(a, b, c); });
-    } else if (n === 2 && b != null && b['@@functional/placeholder'] === true) {
-      return _curry2(function(b, c) { return fn(a, b, c); });
-    } else if (n === 2) {
-      return _curry1(function(c) { return fn(a, b, c); });
-    } else if (n === 3 && a != null && a['@@functional/placeholder'] === true &&
-                          b != null && b['@@functional/placeholder'] === true &&
-                          c != null && c['@@functional/placeholder'] === true) {
-      return f3;
-    } else if (n === 3 && a != null && a['@@functional/placeholder'] === true &&
-                          b != null && b['@@functional/placeholder'] === true) {
-      return _curry2(function(a, b) { return fn(a, b, c); });
-    } else if (n === 3 && a != null && a['@@functional/placeholder'] === true &&
-                          c != null && c['@@functional/placeholder'] === true) {
-      return _curry2(function(a, c) { return fn(a, b, c); });
-    } else if (n === 3 && b != null && b['@@functional/placeholder'] === true &&
-                          c != null && c['@@functional/placeholder'] === true) {
-      return _curry2(function(b, c) { return fn(a, b, c); });
-    } else if (n === 3 && a != null && a['@@functional/placeholder'] === true) {
-      return _curry1(function(a) { return fn(a, b, c); });
-    } else if (n === 3 && b != null && b['@@functional/placeholder'] === true) {
-      return _curry1(function(b) { return fn(a, b, c); });
-    } else if (n === 3 && c != null && c['@@functional/placeholder'] === true) {
-      return _curry1(function(c) { return fn(a, b, c); });
-    } else {
-      return fn(a, b, c);
+    switch (arguments.length) {
+      case 0:
+        return f3;
+      case 1:
+        return _isPlaceholder(a) ? f3
+             : _curry2(function(_b, _c) { return fn(a, _b, _c); });
+      case 2:
+        return _isPlaceholder(a) && _isPlaceholder(b) ? f3
+             : _isPlaceholder(a) ? _curry2(function(_a, _c) { return fn(_a, b, _c); })
+             : _isPlaceholder(b) ? _curry2(function(_b, _c) { return fn(a, _b, _c); })
+             : _curry1(function(_c) { return fn(a, b, _c); });
+      default:
+        return _isPlaceholder(a) && _isPlaceholder(b) && _isPlaceholder(c) ? f3
+             : _isPlaceholder(a) && _isPlaceholder(b) ? _curry2(function(_a, _b) { return fn(_a, _b, c); })
+             : _isPlaceholder(a) && _isPlaceholder(c) ? _curry2(function(_a, _c) { return fn(_a, b, _c); })
+             : _isPlaceholder(b) && _isPlaceholder(c) ? _curry2(function(_b, _c) { return fn(a, _b, _c); })
+             : _isPlaceholder(a) ? _curry1(function(_a) { return fn(_a, b, c); })
+             : _isPlaceholder(b) ? _curry1(function(_b) { return fn(a, _b, c); })
+             : _isPlaceholder(c) ? _curry1(function(_c) { return fn(a, b, _c); })
+             : fn(a, b, c);
     }
   };
 };
--- a/src/uniq.js
+++ b/src/uniq.js
@@ -1,5 +1,5 @@
-var equals = require('./equals');
-var uniqWith = require('./uniqWith');
+var identity = require('./identity');
+var uniqBy = require('./uniqBy');
 
 
 /**
@@ -19,4 +19,4 @@ var uniqWith = require('./uniqWith');
  *      R.uniq([1, '1']);     //=> [1, '1']
  *      R.uniq([[42], [42]]); //=> [[42]]
  */
-module.exports = uniqWith(equals);
+module.exports = uniqBy(identity);
--- a/lib/bench/uniq.bench.js
+++ b/lib/bench/uniq.bench.js
@@ -1,0 +1,64 @@
+var R = require('../..');
+
+var arr1 = R.times(function() {
+  var n = (Math.random() * 100) >>> 0;
+  if (n < 20) {
+    // integer
+    return n;
+  } else if (n < 40) {
+    // boolean
+    return n < 30;
+  } else if (n < 60) {
+    // string
+    return n.toString(10);
+  } else if (n < 80) {
+    // array
+    return [n % 3, n % 7, n];
+  } else {
+    // object
+    return {
+      a: 'foo',
+      n: n
+    };
+  }
+}, 1000);
+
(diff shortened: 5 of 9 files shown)


src/internal/_curry1.js
```diff
@@ -10,7 +10,7 @@ module.exports = function _curry1(fn) {
   return function f1(a) {
     if (arguments.length === 0) {
       return f1;
-    } else if (a != null && a['@@functional/placeholder'] === true) {
+    } else if (a instanceof Object && a['@@functional/placeholder'] === true) {
```
davidchambers: Could we use `a != null && typeof a === 'object'` here instead? I'm wary of introducing an `instanceof` check.

scott-christopher: I don't mind changing it, though I'm curious to know what the particular concerns of using `instanceof` here are.

CrossEye:
Can I suggest that this is more informative?:

```
│ uniq                   │ Hertz - Before         │ Hertz - After          │
│ uniqBy(id, arr1)       │ 21.52 ± 0.60%          │ 94.13 ± 0.57%          │
│ uniqWith(equals, arr1) │ 21.81 ± 0.68%          │ 40.21 ± 0.55%          │
│ uniq(arr1)             │ 21.66 ± 0.81%          │ 87.88 ± 0.62%          │
│ uniqBy(id, arr2)       │ 8.73 ± 0.72%           │ 21,299 ± 0.45%         │
│ uniqWith(equals, arr2) │ 8.72 ± 0.58%           │ 25.60 ± 0.46%          │
│ uniq(arr2)             │ 8.67 ± 0.83%           │ 18,967 ± 0.65          │
│ uniqBy(id, arr3)       │ 692 ± 0.55%            │ 787 ± 0.51%            │
│ uniqWith(equals, arr3) │ 695 ± 0.60%            │ 808 ± 0.53%            │
│ uniq(arr3)             │ 686 ± 0.72%            │ 783 ± 0.61%            │
```

The **after** number for several `uniqBy` variants are surprisingly stronger than the others.  Are these accurate?  I suppose I'll download and test for myself... but it won't be until tomorrow.

src/internal/_curry2.js
```diff
@@ -11,22 +11,21 @@ var _curry1 = require('./_curry1');
...
+      case 0:
+        return f2;
+      case 1:
+        if (a instanceof Object && a['@@functional/placeholder'] === true) {
+          return f2;
+        }
+        return _curry1(function(_b) { return fn(a, _b); });
+      default:
+        if (a instanceof Object && a['@@functional/placeholder'] === true) {
+          return b instanceof Object && b['@@functional/placeholder'] === true ? f2 : _curry1(function(_a) { return fn(_a, b); });
```
davidchambers: Can we wrap this line? :)

src/internal/_curry1.js
davidchambers: Many of the concerns raised in plaid/sanctuary#100 do not apply here, but it *is* possible to concoct a scenario we wouldn't handle correctly:

```javascript
> require('vm').runInNewContext('({"@@functional/placeholder": true})') instanceof Object
false
```

scott-christopher: Cool, makes sense. I'll push up some changes shortly.

scott-christopher: I've created a new `_isPlaceholder` function to reduce the duplication and included the suggested change there.

scott-christopher:
@CrossEye: Thanks, that does convey the information much more effectively (I just lazily copy/pasted the results from the benchmark script :smile:)

Those improvements should be legitimate, as it effectively becomes a `Set` of primitive integers which I'm sure the engines can optimise quite well.

asaf-romano:
Does this also affect difference and union?

CrossEye:
:100:

This is great!  I still have to go through the details more carefully, but I do like what I see.

scott-christopher:
@asaf-romano
> Does this also affect difference and union?

It should provide a boost for these too, as they depend on `_contains` and `uniq`.

scott-christopher:
One question I do have regarding this approach is how do we not neglect the environments without a `Set` implementation when testing to ensure we don't introduce any regressions?

Perhaps some env var or config should be set along with the feature detection to indicate whether to use `Set` or not? This would allow certain tests to override and cover the original code path too.

davidchambers:
> One question I do have regarding this approach is how do we not neglect the environments without a `Set` implementation when testing to ensure we don't introduce any regressions?

Would this work?

```javascript
if (typeof Set !== 'undefined') {
  var _Set = global.Set;
  delete global.Set;
  // run tests...
  global.Set = _Set;
}
```

CrossEye:
> Perhaps some env var or config should be set along with the feature detection to indicate whether to use `Set` or not? This would allow certain tests to override and cover the original code path too.

My problem with that is that we'd then have to introduce some mechanism to run those tests twice.   I'd rather get our browser test store suite running again if we can.

asaf-romano:
Please see  #1518. I think you should consider utilizing ``Map`` rather than ``Set``. Even without the valueOf support suggested there, you could optimize ``Date`` collections.

CrossEye:
@asaf-romano:

> I think you should consider utilizing `Map` rather than `Set`.

Could you explain the advantages in `uniq*`, `difference*`, etc?  I'm not following.

asaf-romano:
uniq and friends are going to be faster when they use native Set implementation (assuming we return~~ Array.from(set.values()) at the end). That limits us to optimizing cases where the initial value for comparison is a primitive. However, for an array of Dates object (or anything with friendly valueOf implementation), a Set won't do, because we need to uniq-by valueOf return values, but return the initial values. Here a Map could help because we can use the keys side for comparison and the values side for the return value.

CrossEye:
>  Here a Map could help because we can use the keys side for comparison and the values side for the return value.

Unless I'm grossly misunderstanding you, that's what we already did in `uniqBy`, and @scott-christopher kept that in place while adding a much more performant `Set` implementation whenever possible.  He also set the base `uniq` to build upon this.

asaf-romano:
You're right. I didn't realize we don't use the set values for the return value. I guess the implementation could be a little more elegant with Map, but the functionality is pretty much the same. Sorry for the false alarm.

src/internal/_curry3.js
```diff
@@ -12,43 +13,51 @@ var _curry2 = require('./_curry2');
...
+        } else if (_isPlaceholder(b)) {
+          if (_isPlaceholder(c)) {
+            return _curry2(function(_b, _c) { return fn(a, _b, _c); });
+          } else {
+            return _curry1(function(_b) { return fn(a, _b, c); });
+          }
+        } else if (_isPlaceholder(c)) {
+          return _curry1(function(_c) { return fn(a, b, _c); });
+        }
+        return fn(a, b, c);
```
davidchambers: I do find this more difficult to follow than the existing implementation (which itself is pretty scary).

davidchambers:
>   - Refactoring the various internal currying functions to ensure the args are `Object`s before checking for the placeholder property, preventing unnecessary boxing/unboxing of primitives.
>
>   - Some tweaks to the conditional layout of the internal currying functions to reduce redundant conditional tests.

@scott-christopher, could you provide some indication of the extent to which each of these contributes to the overall performance improvements? The former is a clear improvement; the latter compromises the readability of the internal curry functions (which are already convoluted). If the changes to the conditional logic make only a small difference I'd prefer to leave them out.

jdalton:
I'm kind of surprised you think a wall of `['@@functional/placeholder']` checks is more readable than `_isPlaceholder`

davidchambers:
I've no issue at all with `_isPlaceholder`, @jdalton. It's a nice function.

My issue is with changing code with this shape…

```javascript
if (XXXXXXX) {
  return XX;
} else if (XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX) {
  return XX;
} else if (XXXXXXX) {
  return XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX;
} else if (XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
                      XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX) {
  return XX;
} else if (XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX) {
  return XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX;
} else if (XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX) {
  return XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX;
} else if (XXXXXXX) {
  return XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX;
} else if (XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
                      XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
                      XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX) {
  return XX;
} else if (XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
                      XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX) {
  return XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX;
} else if (XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
                      XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX) {
  return XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX;
} else if (XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
                      XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX) {
  return XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX;
} else if (XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX) {
  return XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX;
} else if (XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX) {
  return XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX;
} else if (XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX) {
  return XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX;
} else {
  return XXXXXXXXXXX;
}
```

to code with this shape…

```javascript
switch (XXXXXXXXXXXXXXXX) {
  case X:
    return XX;

  case X:
    if (XXXXXXXXXXXXXXXXX) {
      return XX;
    }
    return XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX;

  case X:
    if (XXXXXXXXXXXXXXXXX) {
      if (XXXXXXXXXXXXXXXXX) {
        return XX;
      } else {
        return XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX;
      }
    } else if (XXXXXXXXXXXXXXXXX) {
      return XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX;
    }
    return XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX;

  default:
    if (XXXXXXXXXXXXXXXXX) {
      if (XXXXXXXXXXXXXXXXX) {
        if (XXXXXXXXXXXXXXXXX) {
          return XX;
        } else {
          return XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX;
        }
      } else if (XXXXXXXXXXXXXXXXX) {
        return XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX;
      } else {
        return XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX;
      }
    } else if (XXXXXXXXXXXXXXXXX) {
      if (XXXXXXXXXXXXXXXXX) {
        return XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX;
      } else {
        return XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX;
      }
    } else if (XXXXXXXXXXXXXXXXX) {
      return XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX;
    }
    return XXXXXXXXXXX;
}
```

Readability is subjective, of course, but I find the flat structure easier to grok.

jdalton:
So an indentation gripe.

scott-christopher:
@davidchambers 
I've pushed up a change to hopefully improve the readability of the conditional logic in the internal curry functions, which conveniently gained a very slight nudge in performance in my testing too (though with these benchmarks, that could be put down to a variation in solar winds and price of eggs).

davidchambers:
I :heart: the update!

buzzdecafe:
:cow2:
:egg:

src/uniq.js
```diff
@@ -19,4 +19,4 @@ var uniqWith = require('./uniqWith');
  *      R.uniq([1, '1']);     //=> [1, '1']
  *      R.uniq([[42], [42]]); //=> [[42]]
  */
-module.exports = uniqWith(equals);
+module.exports = uniqBy(identity);
```
davidchambers: This is a nice tweak. :)

davidchambers:
:deciduous_tree:

Merge at will, @scott-christopher. I leave it to you to decide whether or not to squash before doing so.

buzzdecafe:
:cow2:

 thanks for this

CrossEye:
Woo hoo!

PR #1513, opened 2015-11-12
Add JSDoc named parameter to functions using argumens object

jussiarpalahti:
JSDoc needs unused parameter for function declaration and documentation string when functions use the arguments object.

See more: http://usejsdoc.org/tags-param.html#multiple-types-and-repeatable-parameters

### DIFF
--- a/src/compose.js
+++ b/src/compose.js
@@ -20,7 +20,7 @@ var reverse = require('./reverse');
  *
  *      f(3, 4); // -(3^4) + 1
  */
-module.exports = function compose() {
+module.exports = function compose(functions) {
   if (arguments.length === 0) {
     throw new Error('compose requires at least one argument');
   }
--- a/src/composeK.js
+++ b/src/composeK.js
@@ -16,7 +16,7 @@ var prepend = require('./prepend');
  * @since v0.16.0
  * @category Function
  * @sig Chain m => ((y -> m z), (x -> m y), ..., (a -> m b)) -> (m a -> m z)
- * @param {...Function}
+ * @param {...Function} functions
  * @return {Function}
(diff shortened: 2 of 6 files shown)


jussiarpalahti:
From my post in Gitter:

Since arguments object is used in only compose and pipe family of functions, I went ahead and added the necessary "fake" argument into function declarations... Apparently this breaks the build since jshint doesn't like this unused argument :(

Apparently there is an option to silence this warning: jshint/jshint@4ac15f6
Let me know if you like the pull request and I can fix this problem

CrossEye:
Note that we do [already use][ar] this JSHint override elsewhere.


  [ar]: https://github.com/ramda/ramda/blob/master/src/internal/_arity.js#L2

CrossEye:
I know that this has something to do with IDEs, from a brief glance at Gitter, but a more complete explanation here as to why would be welcome.

jussiarpalahti:
Well, I noticed the problem when using an IDE, but it's affects any programmatic analysis of JSDoc. Since there is no other way to differentiate between functions taking no arguments and functions that use arguments object, JSDoc uses a hack where there is unused argument in the function declaration and corresponding @param in the docstring.

JSDoc requiring this solution is not clean, since it clutters the code and requires hints for linting. I don't like it that much now that I've thought about it, but so called correct JSDoc requires this.

Also, I noticed that in some cases there is named parameter "functions" in the docstring and in others its missing, which is a bit inconsistent.

This is a relatively small pull request to make my usage a bit better, but not a strong requirement by any means.

jussiarpalahti:
Closing, have no time to make it right (or good)...

PR #1514, opened 2015-11-12
Change Functor to Apply in documentation of R.flip

safareli:
R.flip uses `ap` method of object  instead of just `map` so we need to update documentation

### DIFF
--- a/src/lift.js
+++ b/src/lift.js
@@ -4,7 +4,7 @@ var liftN = require('./liftN');
 
 /**
  * "lifts" a function of arity > 1 so that it may "map over" an Array or other
- * Functor.
+ * Apply.
  *
  * @func
  * @memberOf R
@@ -12,7 +12,7 @@ var liftN = require('./liftN');
  * @category Function
  * @sig (*... -> *) -> ([*]... -> [*])
  * @param {Function} fn The function to lift into higher context
- * @return {Function} The function `fn` applicable to mappable objects.
+ * @return {Function} The function `fn` applicable to appable objects.
  * @see R.liftN
  * @example
  *
--- a/src/lift.js
+++ b/src/lift.js
@@ -4,7 +4,7 @@ var liftN = require('./liftN');
 
 /**
  * "lifts" a function of arity > 1 so that it may "map over" an Array or other
- * Apply.
+ * object that satisfies the FantasyLand Apply spec.
  *
  * @func
  * @memberOf R
@@ -12,7 +12,7 @@ var liftN = require('./liftN');
  * @category Function
  * @sig (*... -> *) -> ([*]... -> [*])
  * @param {Function} fn The function to lift into higher context
- * @return {Function} The function `fn` applicable to appable objects.
+ * @return {Function} The lifted function.
  * @see R.liftN
  * @example
  *
--- a/src/lift.js
+++ b/src/lift.js
@@ -4,7 +4,7 @@ var liftN = require('./liftN');
 
 /**
  * "lifts" a function of arity > 1 so that it may "map over" an Array or other
- * object that satisfies the FantasyLand Apply spec.
+ * object that satisfies the [FantasyLand Apply spec](https://github.com/fantasyland/fantasy-land#apply).
  *
  * @func
  * @memberOf R
--- a/src/liftN.js
+++ b/src/liftN.js
@@ -8,7 +8,7 @@ var map = require('./map');
 
 /**
  * "lifts" a function to be the specified arity, so that it may "map over" that
- * many lists (or other Applys).
+ * many lists (or other objects that satisfies the FantasyLand Apply spec).
  *
  * @func
  * @memberOf R
@@ -16,7 +16,7 @@ var map = require('./map');
  * @category Function
  * @sig Number -> (*... -> *) -> ([*]... -> [*])
  * @param {Function} fn The function to lift into higher context
- * @return {Function} The function `fn` applicable to appable objects.
+ * @return {Function} The lifted function.
  * @see R.lift
  * @example
  *
--- a/src/liftN.js
+++ b/src/liftN.js
@@ -8,7 +8,7 @@ var map = require('./map');
 
 /**
  * "lifts" a function to be the specified arity, so that it may "map over" that
- * many lists (or other objects that satisfies the FantasyLand Apply spec).
+ * many lists (or other objects that satisfies the [FantasyLand Apply spec](https://github.com/fantasyland/fantasy-land#apply)).
  *
  * @func
  * @memberOf R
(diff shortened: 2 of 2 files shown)


safareli:
if fording is ok  i would update `lift.js` to

src/liftN.js
```diff
@@ -8,7 +8,7 @@ var map = require('./map');
 
 /**
  * "lifts" a function to be the specified arity, so that it may "map over" that
- * many lists (or other Functors).
+ * many lists (or other Applies).
```
joneshf: `Apply` is a proper noun, so you'd want to say Applys, not Applies.

src/liftN.js
```diff
@@ -16,7 +16,7 @@ var map = require('./map');
  * @category Function
  * @sig Number -> (*... -> *) -> ([*]... -> [*])
  * @param {Function} fn The function to lift into higher context
- * @return {Function} The function `fn` applicable to mappable objects.
+ * @return {Function} The function `fn` applicable to applicable objects.
```
joneshf: If we're portmanteauing this, you probably want to say "appable".

src/liftN.js
safareli: fixed +1

src/liftN.js
safareli: fixed +1

CrossEye:
I have a hard time using `Apply` here (regardless of whether it's plurarlized as "Applies" or "Applys".)

I know a lot of people here are comfortable with the Fantasy-land specification, but I think for many readers it will be obscure.  Still more, some of its terms are absurd.  A few are common: `Monoid`, `Functor`, and `Monad`, are relatively common. `Setoid` and `Semigroup` have clear justification from Algebra.  `Applicative` is a little harder to accept, but possible because of its use within Haskell.  `Comonad` is a fairly obscure term from Category Theory, but also gets some lift from Haskell.  `Foldable` and `Traversable` are travesties that sound like they belong in Java, and `Apply` and `Chain` make absolutely no sense to me: types named for methods.

I'm not sure of the best phrasing, but perhaps

    (or other objects with `ap` and `map` methods.)

would do better.

src/lift.js
```diff
@@ -4,7 +4,7 @@ var liftN = require('./liftN');
 
 /**
  * "lifts" a function of arity > 1 so that it may "map over" an Array or other
- * Functor.
+ * Apply.
```
buzzdecafe: perhaps: "... or other object that satisfies the FantasyLand Apply spec."

src/lift.js
```diff
@@ -12,7 +12,7 @@ var liftN = require('./liftN');
  * @category Function
  * @sig (*... -> *) -> ([*]... -> [*])
  * @param {Function} fn The function to lift into higher context
- * @return {Function} The function `fn` applicable to mappable objects.
+ * @return {Function} The function `fn` applicable to appable objects.
```
buzzdecafe: i think this would be sufficient:

`@return {Function} the lifted function`

safareli:
@buzzdecafe +1 sounds better, would commit tomorrow

joneshf:
It's not any more absurd to say `Apply` than other terms we use in programming like `Array`, `Object`, `Observable`, or `Promise`.

It's been used in some libraries for more than a few years: https://github.com/ekmett/semigroupoids/blob/6f974f984570f6fdb85428d618edbb21e1f7b5e7/Data/Functor/Apply.hs

However, fantasy land was influenced by more than just haskell. The awesome scalaz library (with way more adoption than fantasy land) provided inspiration as well. `Apply` has been there for almost seven years: https://github.com/scalaz/scalaz/blob/d87a28b3081b2ed69da5fb364d44fd6398fda9f6/src/main/scalaz/control/Apply.scala and that's just the history that's available, might actually have been longer. `Bind` has been in there for just as long: https://github.com/scalaz/scalaz/blob/d87a28b3081b2ed69da5fb364d44fd6398fda9f6/src/main/scalaz/control/Bind.scala They might have been arbitrary or puns initially, but they are well accepted terms now. Let's take a little perspective on this: node.js wasn't even released, we were still on es3 (while es4 was dieing), and jquery was barely two years old at this point in history.

Ramda doesn't have to be **the** fantasy land library, nor does it have to use the appropriate terms, but there's history behind these terms.

CrossEye:
Whoa!  No offence intended!

I really don't have a serious problem here with the FantasyLand specification choosing to use this name.  But for Ramda's documentation, this is assuming far too much knowledge on the part of the
user:

    "lifts" a function of arity > 1 so that it may "map over" an Array or other Apply.

Although Ramda users should be able to works well with FantasyLand implementations, it shouldn't be necessary to know anything about them in order to understand the documentation.

I see that the latest commit updates this to a somewhat more palatable

    ... or other object that satisfies the FantasyLand Apply spec.

But even this is not really enough for me.  I'd want that to link to the specification itself or even better, to the hypothetical description of types I've suggested we should include in our documentation.

buzzdecafe:
> somewhat more palatable

thanks, i guess .... "other object that satisfies the FantasyLand Apply spec" is the intent of the function. That is what we should document that we support. 

> I'd want that to link to the specification

Agreed.

src/lift.js
```diff
@@ -4,7 +4,7 @@ var liftN = require('./liftN');
 
 /**
  * "lifts" a function of arity > 1 so that it may "map over" an Array or other
- * Functor.
+ * object that satisfies the FantasyLand Apply spec.
```
buzzdecafe: per the comments, please add a link to "FantasyLand Apply spec": https://github.com/fantasyland/fantasy-land#apply

src/liftN.js
```diff
@@ -8,7 +8,7 @@ var map = require('./map');
 
 /**
  * "lifts" a function to be the specified arity, so that it may "map over" that
- * many lists (or other Functors).
+ * many lists (or other objects that satisfies the FantasyLand Apply spec).
```
buzzdecafe: likewise, link here please

CrossEye:
>  "other object that satisfies the FantasyLand Apply spec" is the intent of the function. 

I see it differently.  But what's correct is too verbose for this sort of documentation, I believe: Something to the effect of, "Objects with a `map` function that satisfies the appropriate identity and composition laws and an `ap` function that satisfies the appropriate composition law."  While these laws do happen to be specified by FantasyLand, the user should not have to know about that.

src/lift.js
```diff
@@ -4,7 +4,7 @@ var liftN = require('./liftN');
 
 /**
  * "lifts" a function of arity > 1 so that it may "map over" an Array or other
- * Functor.
+ * object that satisfies the [FantasyLand Apply spec](https://github.com/fantasyland/fantasy-land#apply).
```
CrossEye: I can live with this.  If I ever get around to adding a Types section, I'll think about adding it there.

CrossEye:
:herb:

Issue #1384, opened 2015-09-14
Memoize should probably use some internal hashing function rather than toString


asaf-romano:
See #1383 comment 3. I think that using toString limits the flexibility here, as it requires an `eval` representation of an object, which may not always contain everything necessary to base an equality check on.

Powerful enough, such a function could be used for implementing a much more efficient version of `uniq`, `union` and other functions currently relying on `equals`.

CrossEye:
I'm all ears.

I believe that part of the design of `R.toString` (please correct me if I'm wrong, @davidchambers) was that it was supposed to be good enough for this task, that two equal  values would return the same `toString` value, and no two that were not equal would do so.  I would think that if we were to go down the path of #1383, then we would also update `toString`.

If we did do a hash, what's your proposal for dealing with potential hash collisions?

asaf-romano:
I actually didn't suggest a proper hashcode function, and considered toString as some sort of pseudo hashing. The problem with toString here isn't that it returns a string rather than a number, but rather that because it's a public API, you have to be "nice";  that's how we got ``R.toString([...]) == R.toString(/.../.exec("...")``. A proper hashcode is the more robust option anyway though, and collisions could be handles the same way they are handled in Java for example, i.e. utilize ``equals``.

Regarding toString limitations I know of:
* "Overriding" ``equals`` but not ``toString`` is a problem (btw, the same way overriding ``equals`` but not ``hashcode`` in Java is a problem; except here the contract is an implementation detail that is easy to overlook).
* the Object prototype is ignored (see david's comment in the other bug).
* If we do address Symbols in ``equals``, I don't think there's a way ``toString`` them nicely, needless to say in an ``eval``-able way.

```js
var _a = Symbol(), _b = Symbol();
R.toString({ [_a]: 1, [_b]: 2 }); // Currently throws "TypeError: can't convert symbol to string", but would probably be {} to keep eval happy.
```

CrossEye:
> I actually didn't suggest a proper hashcode function

Sorry, I just assumed from the title.

> * "Overriding" ``equals`` but not ``toString`` is a problem (btw, the same way overriding ``equals`` but not ``hashcode`` in Java is a problem; except here the contract is an implementation detail that is easy to overlook).

This is an interesting point, but it's really odd with our dispatching to think of _overriding_ anything, but only of allowing the user to choose her own function rather than our own if she knows better than us how to handle her data.


> * the Object prototype is ignored (see david's comment in the other bug).

[...]

PR #1516, opened 2015-11-13
respect an object's custom "toString" method in R.toString

davidchambers:
Current behaviour:

```javascript
> R.toString({toString: function() { return 'x'; }})
'{"toString": function () { return \'x\'; }}'
```

New behaviour:

```javascript
> R.toString({toString: function() { return 'x'; }})
'x'
```

### DIFF
--- a/src/internal/_toString.js
+++ b/src/internal/_toString.js
@@ -1,5 +1,4 @@
 var _contains = require('./_contains');
-var _functionName = require('./_functionName');
 var _map = require('./_map');
 var _quote = require('./_quote');
 var _toISOString = require('./_toISOString');
@@ -36,9 +35,12 @@ module.exports = function _toString(x, seen) {
     case '[object Undefined]':
       return 'undefined';
     default:
-      return (typeof x.constructor === 'function' && _functionName(x.constructor) !== 'Object' &&
-              typeof x.toString === 'function' && x.toString() !== '[object Object]') ?
-             x.toString() :  // Function, RegExp, user-defined types
-             '{' + mapPairs(x, keys(x)).join(', ') + '}';
+      if (typeof x.toString === 'function') {
+        var repr = x.toString();
+        if (repr !== '[object Object]') {
+          return repr;
+        }
+      }
+      return '{' + mapPairs(x, keys(x)).join(', ') + '}';
   }
 };
--- a/src/internal/_functionName.js
+++ b/src/internal/_functionName.js
@@ -1,5 +1,0 @@
-module.exports = function _functionName(f) {
-  // String(x => x) evaluates to "x => x", so the pattern may not match.
-  var match = String(f).match(/^function (\w*)/);
-  return match == null ? '' : match[1];
-};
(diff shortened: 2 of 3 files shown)


buzzdecafe:
:cow2:

asaf-romano:
Wait, what? I thought R.toString always dispatched.

src/internal/_toString.js
```diff
@@ -36,9 +35,12 @@ module.exports = function _toString(x, seen) {
     case '[object Undefined]':
       return 'undefined';
     default:
-      return (typeof x.constructor === 'function' && _functionName(x.constructor) !== 'Object' &&
-              typeof x.toString === 'function' && x.toString() !== '[object Object]') ?
-             x.toString() :  // Function, RegExp, user-defined types
-             '{' + mapPairs(x, keys(x)).join(', ') + '}';
+      if (typeof x.toString === 'function') {
+        var repr = x.toString();
+        if (repr !== '[object Object]') {
```
CrossEye: Is this better than simply checking whether `x.toString === Object.prototype.toString`?

CrossEye:
:herb:  But all of this makes me realize that `toString` will probably **not** be a reasonable hash for an object.  I'm glad @scott-christopher seems to have found another useful approach for some of the perf problems.

src/internal/_toString.js
asaf-romano: Yes, #1140

CrossEye: Yes, of course.  Ugly, though.

asaf-romano:
Wouldn't this change also take into account String.prototype.toString though, such that ``"foo".toString == "foo"```?

davidchambers:
@asaf-romano, string values are handled earlier in the function:

> ```javascript
> case '[object String]':
>   return typeof x === 'object' ? 'new String(' + recur(x.valueOf()) + ')' : _quote(x);
> ```

asaf-romano:
Ah, I see. JS primitives boxing can be very confusing.

This somewhat relates to @CrossEye's comment regarding toString-as-hash (also see #1384). Now that R.toString dispatches, it's not safe to rely on eval(toString-result) to work (most custom toString implementations will not meet that requirement, pretty much any implementation apart those delegating to JSON.stringify). So if R.toString is to remain the de-facto hashing function in the library, it might be a good time to do away with that in favor of a more robust implementation (in terms of performance and edge cases coverage).

davidchambers:
> Now that R.toString dispatches

To be clear, `R.toString` has always dispatched for objects whose constructor name is not `'Object'`. This pull request just removed the `'Object'` exception.

> it's not safe to rely on eval(toString-result) to work

We certainly shouldn't rely on this property in code, but it is a useful guiding principle (which serves us as we provide implementations for built-in constructors, and third parties defining their own constructors).

> most custom toString implementations will not meet that requirement

In the JavaScript ecosystem as a whole, the `eval(R.toString(x)) = x` property will rarely hold. In the Ramda ecosystem, though, it should hold. Sanctuary and ramda-fantasy both provide compliant `toString` methods. For example:

```javascript
const R = require('ramda');
const S = require('sanctuary');

const Just = S.Just;

eval(R.toString(eval(R.toString(Just(42)))));
// => Just(42)
```

> pretty much any implementation apart those delegating to JSON.stringify

`JSON.stringify` is unlikely to be useful as it loses track of the constructor. For example:

```javascript
JSON.stringify(Just(42));
// => '{"value":42}'
```

PR #1517, opened 2015-11-13
Include browser tests in the build again

scott-christopher:
- Removes redundant build scripts
- Generates test bundle for browser using browserify
- Refactor `clone` implementation to use `valueOf` for `Date` objects to address issue in Firefox (~version 35)
- Refactor `_equals` for `Error` instances to no longer compare properties other than `name` and `message`
  - Safari's `Error` has `line` and `column` properties which will rarely be equal unless they're identical
- ~~Enable strict mode in `isEmpty(arguments)` test to workaround bug in Safari~~
- Refactor `keys` to workaround Safari bug with enumerable `length` prop on `arguments` objects


davidchambers:
Thank you for putting this pull request together, @scott-christopher. It was clearly a lot of work!

scott-christopher:
I've given up trying to get testem running on Travis. First it required PhantomJS v2 to be installed due to a large number of failing tests on pre-v2, then after getting that installed it then failed to run the Firefox and Chrome instances so the builds were failing.

We can investigate getting the testem tests running on Sauce at some point soon, but at least these tests can be manually run for now via `npm run browser_test`, which will run on all available local browsers.

lib/test/index.html
```diff
@@ -1,21 +0,0 @@
...
-  <script src="mocha.js"></script>
-  <script src="assert-shim.js"></script>
-  <!--[if lt IE 9]>
-    <script src="../lib/sauce/ie8-shim.js"></script>
-  <![endif]-->
-  <script>mocha.setup('bdd')</script>
-  <script src="bundle.js"></script>
-  <script src="../sauce/mocharun.js"></script>
-</body>
-</html>
```
CrossEye: While this whole change is clearly moving in the right direction, I've got to admit to some sadness at seeing this file go.  This was how I started the library, long before it was called Ramda, a few tests that ran in Mocha and in the browser.  And if I had problems, I would debug it in Chrome or Firefox in the browser.  This existed before almost any of the other infrastructure of the library.

So :+1: , but :cry:

CrossEye:
> Thank you for putting this pull request together, @scott-christopher. It was clearly a lot of work!

Hear, hear!

test/isEmpty.js
```diff
@@ -28,6 +28,10 @@ describe('isEmpty', function() {
   });
 
   it('returns true for empty arguments object', function() {
+    'use strict';
```
jdalton: you could fix it in ramda.

scott-christopher: I've updated the PR to include a workaround for this bug in `R.keys`.

Makefile
```diff
@@ -38,10 +38,3 @@ setup:
 .PHONY: test
 test: dist/ramda.js
 	npm test
-	find src -name '*.js' -not -path 'src/internal/*' \
-	| sed 's:src/\(.*\)[.]js:exports.\1 = require("./src/\1");:' >index.js
-	sed '/"main":/d' package.json >tmp
-	mv tmp package.json
-	$(MOCHA) -- $(shell find test -name '*.js' -not -name 'test.examplesRunner.js' -not -path 'test/helpers/*')
```
davidchambers: Since this was the only reference to `MOCHA`, we can delete line 1. :)

src/times.js
```diff
@@ -21,9 +21,13 @@ var _curry2 = require('./internal/_curry2');
  *      R.times(R.identity, 5); //=> [0, 1, 2, 3, 4]
  */
 module.exports = _curry2(function times(fn, n) {
-  var len = Number(n);
-  var list = new Array(len);
-  var idx = 0;
+  var len = Number(n),
+      idx = 0,
+      list;
```
davidchambers: @buzzdecafe, @CrossEye, @raine: Do you have a preferred `var` style?

src/times.js
```diff
@@ -21,9 +21,13 @@ var _curry2 = require('./internal/_curry2');
...
+  if (len < 0 || isNaN(len)) {
+    throw new RangeError('n must be a non-negative number');
+  }
```
davidchambers: Should we also assert that `len` is finite?

src/times.js
buzzdecafe: i prefer 1:1 over 1:n. but not a deal-breaker

CrossEye: I've also slowly been converted to 1:1.  But the more important thing is to keep consistent.  Definitely not a deal-breaker.

src/times.js
buzzdecafe: i can vaguely imagine a scenario where this is getting evaluated lazily (or transduced?) where it would be nice to keep going one-by-one. ... vaguely. 

On another topic, `times` feels pretty imperative, much like `forEach`.

davidchambers: I remain in favour of removing `times`, despite my concession to @TheLudd in #1264. ;)

buzzdecafe: hmmm. compromise solution is a `ramda-contrib` package. maybe worth doing? move `times` and `forEach` and maybe some other stuff out of core. idk, sounds like work.

davidchambers: > idk, sounds like work.

That's the problem. I suggest we let these functions stick around for the time being. :)

lib/test/index.html
scott-christopher: I might not be able to help with the nostalgia, but with this in place you'll still be able to run `node_modules/.bin/testem` and then effectively have access to the same content in this file via `http://local host:8080` to poke and prod at it in a browser environment.

CrossEye: Understood, and I don't use it that way much any more.  It's just a little nostalgic tug.  But there is something that I do actually regret.  For some time a user could clone or download Ramda HEAD and simply use `ramda.js` or `ramda.min.js`.  She could also see the tests in action by opening that browser test file.  She didn't have to have any tools beyond git (or even unzip) to be able to use Ramda.  Granted she could not run a build herself.  But it was easily available to unsophisticated developers.

We broke some of this when we stopped including these files in the commits.  While I agreed to that we should do so, I still hope one day to get Travis to add this; I've dropped the ball on this, though.

scott-christopher: Perhaps a REPL like the one available in ramdajs.com/repl could be bundled to better serve this use case?

We could also add a dedicated npm script that runs `testem` in dev-mode with a mention in the README to reduce the level of assumed knowledge.

CrossEye: I've worked with a number of very junior JS developers over the years.  And even today, NPM is a mystery to many of them.  But a script that they can use via `<script src='ramda'></script>` is always familiar.  A test suite that runs as a web page is familiar.  Don't get me wrong.  I don't want to spend to much effort catering to them, but it would be great if we can support them without a lot of effort.  An included REPL sounds like a good idea.

buzzdecafe: FWIW, i prefer the repl @davidchase is developing: http://davidchase.github.io/ramtuary/
I am hoping we can replace the existing repl with that one

davidchambers:
Could you rebase this branch, @scott-christopher?

scott-christopher:
:zap: rebased

davidchambers:
:deciduous_tree: Thanks!

PR #1521, opened 2015-11-16
extract lensIndex test from lenses.js

DanH91:
Hi according to discussion [#1509](https://github.com/ramda/ramda/issues/1509) I made a lensIndes dedicated test file.

### DIFF
--- a/test/lensIndex.js
+++ b/test/lensIndex.js
@@ -1,0 +1,27 @@
+var R = require('..');
+var eq = require('./shared/eq');
+
+var they = it;
+
+
+var alice = {
+  name: 'Alice Jones',
+  address: ['22 Walnut St', 'San Francisco', 'CA']
+};
+
+var headLens = R.lensIndex(0);
+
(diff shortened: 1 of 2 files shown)


buzzdecafe:
i think this is already underway in https://github.com/ramda/ramda/pull/1510/files ?

DanH91:
OK. Thanks. Do you have advice on where I should check to avoid dups efforts?

buzzdecafe:
i don't know if there is a better way than scanning the issue/PR list.

CrossEye:
But also note that #1509 has a section starting "This was referenced `n` days ago", with a pointer (now) to #1510 and this issue.  That's always one place to look.

DanH91:
Ok. Thanks

CrossEye:
@DanH91:

Thank you, in any case, for taking the time to work on this.  I'm sorry it was duplicated effort.  But we appreciate your help regardless.

PR #1524, opened 2015-11-18
R.scan now supports transducers, iterators, and delegates

### LINKED ISSUE #1179: xScan
Right now R.scan does not transduce. It should do so for consistency. Scan is much more commonly used for streams so it's also important for compatibility with stream libraries like flyd.

jethrolarson:
Fixes #1179
Implementation now matches R.reduce and should be compatible with other
types that implement `type.scan(reducer, accumulator)` like flyd.


jethrolarson:
I could swear I ran tests before pushing. /sigh.

Issue #96, opened 2014-06-27
R.map doesn't support Sparse Arrays like the standard map


cscalfani:
I think it would be best to have another function, e.g. R.map.sparse, that does the standard map behavior.

Here is some code that will demonstrate the issue:

``` sh
var R = require('ramda');
var sparse = [, 2, 3];
var mapResults = sparse.map(function(item) {
    return item;
});
console.log(mapResults);
var RmapResults = R.map(function(item) {
    return item;
}, sparse);
console.log(RmapResults);
```

PR #1525, opened 2015-11-19
Add splitWhen function

miwillhite:
Inspired by Haskell's `break`:
http://hackage.haskell.org/package/base-4.8.1.0/docs/Prelude.html#v:break

`R.splitWhen(R.equals(2), [1, 2, 3]);   //=> [[1], [2, 3]]`

### DIFF
--- a/src/splitWhen.js
+++ b/src/splitWhen.js
@@ -1,0 +1,32 @@
+var _curry2 = require('./internal/_curry2');
+var _slice = require('./internal/_slice');
+
+
+/**
+ * Takes a list and a predicate and returns a pair of lists with the following properties:
+ *
+ *  - the result of concatenating the two output lists is equivalent to the input list;
+ *  - none of the elements of the first output list satisfies the predicate; and
+ *  - if the second output list is non-empty, its first element satisfies the predicate.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig (a -> Boolean) -> [a] -> [[a], [a]]
+ * @param {Function} pred The predicate that determines where the array is split.
+ * @param {Array} list The array to be split.
+ * @return {Array}
+ * @example
+ *
+ *      R.splitWhen(R.equals(2), [1, 2, 3, 1, 2, 3]);   //=> [[1], [2, 3, 1, 2, 3]]
+ */
+module.exports = _curry2(function splitWhen(pred, list) {
+  var idx = 0, len = list.length, prefix = [];
+
+  while (idx < len && !pred(list[idx])) {
+    prefix.push(list[idx]);
+    idx += 1;
+  }
+
+  return [prefix, _slice(list, idx)];
+});
--- a/test/splitWhen.js
+++ b/test/splitWhen.js
@@ -1,0 +1,22 @@
+var R = require('..');
+var eq = require('./shared/eq');
+
+
+describe('splitWhen', function() {
+  it('splits an array at the first instance to satisfy the predicate', function() {
+    eq(R.splitWhen(R.equals(2), [1, 2, 3]), [[1], [2, 3]]);
+  });
+
+  it('retains all original elements', function() {
+    eq(R.splitWhen(R.T, [1, 1, 1]), [[], [1, 1, 1]]);
+  });
+
+  it('is curried', function() {
+    var splitWhenFoo = R.splitWhen(R.equals('foo'));
+    eq(splitWhenFoo(['foo', 'bar', 'baz']), [[], ['foo', 'bar', 'baz']]);
+  });
+
+  it('only splits once', function() {
+    eq(R.splitWhen(R.equals(2), [1, 2, 3, 1, 2, 3]), [[1], [2, 3, 1, 2, 3]]);
+  });
+});


src/splitBy.js
```diff
@@ -0,0 +1,24 @@
...
+ * @param {Function} fn The predicate that determines where the array is split.
+ * @param {Array} array The array to be split.
+ * @return {Array}
+ * @example
+ *
+ *      R.splitBy(R.equals(2), [1, 2, 3]);   //=> [[1], [2, 3]]
+ */
+module.exports = _curry2(function splitBy(predicate, array) {
+  var xs = takeWhile(complement(predicate), array);
+  return [xs, difference(array, xs)];
```
davidchambers: We can't use `difference` here, as it discards duplicates. Try something like `R.splitBy(R.T, [1, 1, 1])`.

It would be great to use [JSVerify][1] to test these properties:

  - `forall p :: a -> Boolean, xs :: [a]. length(splitBy(p, xs)) = 2`
  - `forall p :: a -> Boolean, xs :: [a]. unnest(splitBy(p, xs)) = xs`


[1]: https://github.com/jsverify/jsverify

CrossEye: You could probably use `drop(xs.length, array)` instead of the `difference` call.

But speaking of that, I try to think of [Ramda's fundamental data type][fd] as the _list_, and the _array_ as an implementation detail.  So I would prefer to name the parameter `list`.
 

  [fd]: https://github.com/ramda/ramda/issues/96#issuecomment-47304374

raine: More importantly, `list` is used everywhere currently.

miwillhite: Thanks for the `list` suggestion, I had pulled the parameter naming from `splitAt`, which looks like it diverges from the norm. Maybe that'll be my next PR ;)

miwillhite: In terms of naming, it occurred to me this morning that maybe `splitWith` would be better than `splitBy`…any preference?

src/splitWith.js
```diff
@@ -0,0 +1,23 @@
...
+var takeWhile = require('./takeWhile');
+
+
+/**
+ * Splits a given list at the first occurance of a satisfied predicate.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig Number -> [a] -> [[a], [a]]
```
raine: Type of the first parameter should indicate a function. `(a -> Boolean)`?

miwillhite: Oh of course! Thank you.

src/splitBy.js
CrossEye: Actually, see #65 to see why neither might be appropriate.  `splitWhen`?

miwillhite: I like `splitWhen`, I'll make the update.

miwillhite:
Anything I can do to move this along? I'm happy to help however I can.

src/splitWhen.js
```diff
@@ -0,0 +1,23 @@
...
+var takeWhile = require('./takeWhile');
+
+
+/**
+ * Splits a given list at the first occurance of a satisfied predicate.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig (a -> Boolean)t  -> [[a], [a]]
```
davidchambers:
```diff
- * @sig (a -> Boolean)t  -> [[a], [a]]
+ * @sig (a -> Boolean) -> [a] -> [[a], [a]]
```

src/splitWhen.js
```diff
@@ -0,0 +1,23 @@
...
+ * @param {Function} fn The predicate that determines where the array is split.
+ * @param {Array} list The array to be split.
+ * @return {Array}
+ * @example
+ *
+ *      R.splitWhen(R.equals(2), [1, 2, 3]);   //=> [[1], [2, 3]]
+ */
+module.exports = _curry2(function splitWhen(predicate, list) {
```
davidchambers: Let's name the first parameter `pred` rather than `predicate`, for consistency with existing functions.

src/splitWhen.js
```diff
@@ -0,0 +1,23 @@
...
+
+
+/**
+ * Splits a given list at the first occurance of a satisfied predicate.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig (a -> Boolean)t  -> [[a], [a]]
+ * @param {Function} fn The predicate that determines where the array is split.
```
davidchambers: `s/fn/pred/`

test/splitWhen.js
```diff
@@ -0,0 +1,19 @@
...
+  it('retains all original elements', function() {
+    eq(R.splitWhen(R.T, [1, 1, 1]), [[], [1, 1, 1]]);
+  });
+
+  it('is curried', function() {
+    var splitWhenFoo = R.splitWhen(R.equals('foo'));
+    eq(splitWhenFoo(['foo', 'bar', 'baz']), [[], ['foo', 'bar', 'baz']]);
+  });
+});
+
```
davidchambers: :space_invader:

src/splitWhen.js
```diff
@@ -0,0 +1,23 @@
...
+ * @param {Array} list The array to be split.
+ * @return {Array}
+ * @example
+ *
+ *      R.splitWhen(R.equals(2), [1, 2, 3]);   //=> [[1], [2, 3]]
```
davidchambers: Let's use `[1, 2, 3, 1, 2, 3]` as the input list, to demonstrate that splitting occurs in one place only.

src/splitWhen.js
```diff
@@ -0,0 +1,23 @@
+var _curry2 = require('./internal/_curry2');
+var complement = require('./complement');
+var dropWhile = require('./dropWhile');
+var takeWhile = require('./takeWhile');
+
+
+/**
+ * Splits a given list at the first occurance of a satisfied predicate.
```
davidchambers: We could be more descriptive. I suggest something along the following lines:

> Takes a list and a predicate and returns a pair of lists with the following properties:
>
>   - the result of concatenating the two output lists is equivalent to the input list;
>   - none of the elements of the first output list satisfies the predicate; and
>   - if the second output list is non-empty, its first element satisfies the predicate.

test/splitWhen.js
davidchambers: Could you delete this trailing empty line?

src/splitWhen.js
```diff
@@ -0,0 +1,23 @@
...
+ * @param {Function} pred The predicate that determines where the array is split.
+ * @param {Array} list The array to be split.
+ * @return {Array}
+ * @example
+ *
+ *      R.splitWhen(R.equals(2), [1, 2, 3, 1, 2, 3]);   //=> [[1], [2, 3, 1, 2, 3]]
+ */
+module.exports = _curry2(function splitWhen(pred, list) {
+  return [takeWhile(complement(pred), list), dropWhile(complement(pred), list)];
+});
```
davidchambers: How do others feel about this implementation? I'd like to see an imperative implementation, to reduce the number of internal dependencies.

miwillhite: A first draft imperative implementation:
```
module.exports = _curry2(function splitWhen(pred, list) {
  var idx = 0, len = list.length, prefix = [];

  while (idx < len && !pred(list[idx])) {
    prefix.push(list[idx]);
    idx += 1;
  }

  return [prefix, _slice(list, idx)];
});
```

CrossEye: This looks good to me.

Sorry we dropped the ball earlier.  I've been mostly disengaged for a few weeks, and I think others mostly have too.  I probably will be so for a few more, but I will try to watch what's happening.

miwillhite: @CrossEye @davidchambers Is the imperative solution favored? I'll update the commit if so…

davidchambers: Yes, let's go with your imperative implementation. :)

miwillhite: Excellent, thank you all for the help. I really appreciate the attention you guys put into making this a great library.

davidchambers:
:deciduous_tree: Thank you, @miwillhite!

megawac:
Is there a reason `splitWhen` is a better name than `partitionBy`? In any case I think `*By` would be a better name

miwillhite:
@megawac Please see the prior (now hidden) discussion: https://github.com/ramda/ramda/pull/1525#discussion_r45403355

megawac:
Ah, I'm not a fan of the name. 

Also on second take, any reason to limit it to two groups. I think I'd prefer a `partitionBy` method which creates a group whenever `pred` is true. This method would be easily derived from such a method

miwillhite:
@megawac What you describe with `partitionBy` is a different use case. I created the `splitWhen` function because it is what I needed…and I feel it is a valid operation that others can take advantage of (hence the PR).

megawac:
I gotcha, I just feel `partitionBy` would cover both use cases as `splitWhen` can easily implemented (either as core or by user) by `partitionBy`

miwillhite:
Interesting to think about…I see something like this?

```javascript
const list = [1, 2, 3, 4, 1, 2, 3, 4, 1, 2];
partitionBy(equals(2), list);
//=> [[1], [2, 3, 4, 1], [2, 3, 4, 1], [2]] -- right?

// splitWhen:
splitWhenWithPartitionBy = compose((xs) => [head(x), flatten(tail(xs))], partitionBy)(list);
```

Though I've learned in this PR that core prefers to avoid using internal dependencies (which makes sense when thinking about API updates)…so this implementation becomes moot.

miwillhite:
@megawac Actually this definition of `partitionBy` competes/conflicts directly with Ramda's `partition`, so I guess I don't have a clear idea of your vision…

megawac:
Whoops, I've been out of the loop for a bit. The method in ramda that I was referring to would be appropriately named `splitBy` currently. I was mixing up `partition` with `chunk` in lodash/underscore

CrossEye:
@megawac:  A long time ago (#65) we established a convention for Ramda (which follow _almost_ faithfully) that `xxxBy` means that we accept a unary function which generates a gloss on each item, and that gets used to handle `xxx` appropriately.  A good example would be a sort key: `sortBy(person => person.last + ',' + person.first, people)`.  On the other hand, `xxxWith` accepts a binary function which combines or compares two elements.  `zipWith((a, n) => a + '-' + n, ['a', 'b', 'c'], [1, 2, 3]); //=> ['a-1', 'b-2', 'c-3']` or `uniqWith((a, b) => a.id === b.id, items)`.

So the name "`splitBy`" would take on a different connotation than you might want.

Now that we merged this one, I'm wondering if "`splitAt`" would be a better name for this one, reserving "`splitWhen`" for this alternate suggestion.  But I'd like to hear more justification for this style splitting.  The only case I can see really wanting is control-break processing<sup>1</sup>, and that would likely involve passing the previous value and the current value.  While that might be useful, it's still a different API.

----------
<sup>1</sup>By control-break processing, I mean splitting a list ordered on some key fields into sublists every time those keys change.  (This is typical in report processing, for instance.)

miwillhite:
There is a `splitAt` in `master` already, the first parameter being the index.

I can't find the source that inspired this PR, but a naïve implementation would be something like this:

```javascript
const fields = ['field1', 'field2', 'step', 'field3', 'field4', 'step', 'field5'];

const buildSteps = ifElse(
  isEmpty,
  identity,
  (fs) => {
    const [ step, fields ] = splitWhen(equals('step'), fs);
    return compose(prepend(step), buildSteps, tail)(fields);
  }
);

buildSteps(fields)
// => [ [ 'field1', 'field2' ], [ 'field3', 'field4' ], [ 'field5' ] ]
```

Where I have a list of _Field_ objects with a special _Step_ field mixed in. The product allows people to build custom forms and I was roughing out a multi-step form feature. This allows them to insert a _Step_ field where they'd like the form to "break".

Issue #81, opened 2014-06-22
zip should use the length of the largest array


megawac:
Most other libraries use a `zip` implementation that will apply the length of the largest array for the result.

```
zip([1, 2, 3], ['a', 'b', 'c', 'd']); // => [[1,"a"],[2,"b"],[3,"c"],[undefined,"d"]]
```

kwijibo:
+1 for R.unzip

CrossEye:
@kwijibo: Interested in creating a Pull Request?

kwijibo:
@CrossEye I'll give it a go

Issue #1068, opened 2015-05-06
Transposing an array, similar to partition or unzip


ccorcos:
I have an array like this `[[0, 'chet'], [2, 'charlie'], [10, 'joe']]` and I want to transpose the array.

The result looks like this `[R.map(R.nth(0))(array), R.map(R.nth(1))(array)]` but obviously, I'd like a point free way of doing this. I'm not sure what the most general form of this would be either.

CrossEye:
So, something like this?:

```js
var transpose = function(a) {
    return R.mapIndexed(function(col, i) { 
        return R.map(R.nth(i), a)
    }, R.head(a));
};

transpose([[0, 'chet'], [2, 'charlie'], [10, 'joe']]); //=> [[0,2,10],["chet","charlie","joe"]]
```

There are three different versions at http://bit.ly/1P17BKx; none manages to get quite points-free.

ccorcos:
Ah, yes. Well I guess the problem I'm having is when there are multiple arguments to different functions within the map... I tried doing this with `nthArg`, but the problem is we're mapping over two values. I'm realizing the same thing on a similar problem where I want to go from `[[0, 'chet'], [2, 'charlie'], [10, 'joe']]` to `[{id:0, username:'chet'}, ...]`. Ideally, something like `R.map(R.zipObj)` should work, but the problem is that whatever you pass will be mapped immediately. I also tried `R.compose(R.map, R.zipObj)` which doesnt work either...

ccorcos:
This is what I settled with for the second example:
```
zip = R.curry (keys, values) ->
    R.map(R.zipObj(keys))(R.apply(R.zip, values))
```

CrossEye:
Oh, we have two functions that help with things like this, `converge` and `useWith`.  Here `useWith` [will help][re]:

    var zip = R.useWith(R.map, R.zipObj, R.identity);

  [re]: http://bit.ly/1Qo7N42

ccorcos:
Whoa. Thats exactly what I wanted! Its pretty hard getting used to thinking like that, but I like it!

PR #1526, opened 2015-11-19
R.transpose (was R.unzip)

kwijibo:
As mentioned in #81

### DIFF
--- a/src/transpose.js
+++ b/src/transpose.js
@@ -1,0 +1,41 @@
+var _curry1 = require('./internal/_curry1');
+
+
+/**
+ * Transposes the rows and columns of a 2D list.
+ * When passed a list of `n` lists of length `x`,
+ * returns a list of `x` lists of length `n`.
+ *
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig [[a]] -> [[a]]
+ * @param {Array} list A 2D list
+ * @return {Array} A 2D list
+ * @example
+ *
+ *      R.transpose([[1, 'a'], [2, 'b'], [3, 'c']]) //=> [[1, 2, 3], ['a', 'b', 'c']]
+ *      R.transpose([[1, 2, 3], ['a', 'b', 'c']]) //=> [[1, 'a'], [2, 'b'], [3, 'c']]
+ *
+ * If some of the rows are shorter than the following rows, their elements are skipped:
+ *
+ *      R.transpose([[10, 11], [20], [], [30, 31, 32]]) //=> [[10, 20, 30], [11, 31], [32]]
+ */
+module.exports = _curry1(function transpose(outerlist) {
+  var i = 0;
+  var result = [];
+  while (i < outerlist.length) {
+    var innerlist = outerlist[i];
+    var j = 0;
+    while (j < innerlist.length) {
+      if (typeof result[j] === 'undefined') {
+        result[j] = [];
+      }
+      result[j].push(innerlist[j]);
+      j += 1;
+    }
+    i += 1;
+  }
+  return result;
+});
--- a/test/transpose.js
+++ b/test/transpose.js
@@ -1,0 +1,22 @@
+var R = require('..');
+var eq = require('./shared/eq');
+
+describe('transpose', function() {
+  it('returns an array of two arrays', function() {
+    var input = [['a', 1], ['b', 2], ['c', 3]];
+    eq(R.transpose(input), [['a', 'b', 'c'], [1, 2, 3]]);
+  });
+  it('skips elements when rows are shorter', function() {
+    var actual = R.transpose([[10, 11], [20], [], [30, 31, 32]]);
+    var expected = [[10, 20, 30], [11, 31], [32]];
+    eq(actual, expected);
+  });
+  it('copes with empty arrays', function() {
+    eq(R.transpose([]), []);
+  });
+  it('copes with true, false, null & undefined elements of arrays', function() {
+    var actual = R.transpose([[true, false, undefined, null], [null, undefined, false, true]]);
+    var expected = [[true, null], [false, undefined], [undefined, false], [null, true]];
+    eq(actual, expected);
+  });
+});


kwijibo:
FWIW, while writing the docs, I noticed that the docs contain reference to both `tuple` and `pair`, and `element` and `item`; not a big deal, but might be worth standardising for internal consistency.

CrossEye:
Agreed.  It would be good to standardize.

test/unzip.js
```diff
@@ -0,0 +1,10 @@
+var R = require('..');
+var eq = require('./shared/eq');
+
+describe('unzip', function() {
+  it('returns an array of two arrays', function() {
+    var input = [[1, 2], ['pot', 'kettle']];
+    eq(R.unzip(input), [[1, 'pot'], [2, 'kettle']]);
```
CrossEye: I think this would be more demonstrative if the list had more than two elements.  The behavior of this example is indistinguishable from `zip`.  Moreover, I would find it more appropriate if the resulting lists are more homogeneous.  Something like:

```js
R.unzip([[1, 'a'], [2, 'b'], [3, 'c'], [4, 'd']]); //=> [[1, 2, 3, 4], ['a', 'b', 'c', 'd']]
```

I suppose the same length comment applies to the code documentation too.

kwijibo:
@CrossEye I've made the changes you mentioned. Actually an alternative to `unzip` could be a more general pair of functions, `pivot` / `unpivot`.
```
pivot([['a', 'b', 'c'], [1, 2, 3], [true, false, true]]) //=> [['a', 1, true], ['b', 2, false], ['c', 2, true]]
unpivot([['a', 1, true], ['b', 2, false], ['c', 2, true]]) //=>  [['a', 'b', 'c'], [1, 2, 3], [true, false, true]]
```
(However, working with pairs seems the most common use case and has most synergy with the existing Ramda functions)

kwijibo:
(also, is `master` the correct branch to do a PR against? what's the Ramda release process?)

buzzdecafe:
what is the difference between `pivoy` and `unpivot`? seems to me that `pivot(pivot(lists)) ≍
 lists`

CrossEye:
`master` is the correct branch.  So far we've done nothing tricky with releases.

CrossEye:
I think I would leave `zip*` in place as it's a well-known interface, but perhaps instead of `unzip`, we could use `pivot` or `transpose` for arbitrary-length lists of lists.  And, as @buzzdecafe points out, we only need the one.

buzzdecafe:
> I think I would leave zip* in place as it's a well-known interface, but perhaps instead of unzip, we could use pivot or transpose for arbitrary-length lists of lists. 

sounds good to me.

kwijibo:
@buzzdecafe well spotted.

I only meant that `pivot` could replace `unzip`, as it only takes one list (of lists) as argument;  `zip` takes two lists, which is probably useful when curried, and `zipWith` and `zipObject` only really make sense on tuples/pairs.

CrossEye:
I prefer `transpose` to `pivot` but could live with either.  It would be nice to see what other langauges/libraries use here.  This is also discussed in #1452, and I presented some implementations in #1068 (although all of them would have to updated to take into account the deprecation of `mapIndexed` in favor of `addIndex` in #1061.)

CrossEye:
Updated those implementations:

http://bit.ly/1PR1NBA

kwijibo:
`transpose` is fine with me; I think `PIVOT` is from MS SQL, while Excel has a `TRANSPOSE` function. Haskell also has `transpose`.

buzzdecafe:
this is fun -- here is a different impl just for fun:

http://bit.ly/1NgIt1O

kwijibo:
How about this?
The behaviour when rows are of unequal length is lifted  from http://hackage.haskell.org/package/base-4.8.1.0/docs/Data-List.html#v:transpose

kwijibo:
Also, can anyone suggest a more informative type signature than [nx] -> [xn] ?

buzzdecafe:
maybe just crib it from Haskell:

```hs
transpose :: [[a]] -> [[a]] 
```

test/transpose.js
```diff
@@ -0,0 +1,15 @@
...
+  it('returns an array of two arrays', function() {
+    var input = [['a', 1], ['b', 2], ['c', 3]];
+    eq(R.transpose(input), [['a', 'b', 'c'], [1, 2, 3]]);
+  });
+  it('skips elements when rows are shorter', function() {
+    var actual = R.transpose([[10, 11], [20], [], [30, 31, 32]]);
+    var expected = [[10, 20, 30], [11, 31], [32]];
+    eq(actual, expected);
+  });
+});
```
buzzdecafe: i think we should have a test for `eq(transpose([]), [])`, otherwise LGTM

kwijibo:
@buzzdecafe cool. Test added.

buzzdecafe:
just needs squashing, then :cow2:

test/transpose.js
```diff
@@ -0,0 +1,18 @@
...
+describe('transpose', function() {
+  it('returns an array of two arrays', function() {
+    var input = [['a', 1], ['b', 2], ['c', 3]];
+    eq(R.transpose(input), [['a', 'b', 'c'], [1, 2, 3]]);
+  });
+  it('skips elements when rows are shorter', function() {
+    var actual = R.transpose([[10, 11], [20], [], [30, 31, 32]]);
+    var expected = [[10, 20, 30], [11, 31], [32]];
+    eq(actual, expected);
+  });
```
CrossEye: This is the behavior as described by another language/library?  It seems ugly.  But I'm not sure I have anything better at the moment.

kwijibo: If you have a table like:
```
10 11   
20      
30 31 32
```
and you change rows to columns, you get:
```
10  20   30
11         31
             32
```
Which would be 
```javascript
[[10,20,30], [11, undefined, 31], [undefined, undefined, 32]]
```

kwijibo: it seems like, say you were rendering such a table, and you wanted to transpose rows and columns, those `undefined`s are necessary information. However if you padded your initial rows with `undefined` so that they are all the same length, you would get a lossless transformation.
ie: 
```javascript
transpose([[10, 11, undefined], [20, undefined, undefined], [undefined,undefined,undefined], [30, 31, 32]]) //=>
[[10,20,30], [11, undefined, 31], [undefined, undefined, 32]]
```

CrossEye: That's the line I was considering.  But it doesn't really seem to work.  I think perhaps the only law we could support is something to the effect of 

```
// if each list in lists has the same length, then
transpose(transpose(lists))  ≍  lists
```

And if the lists are of different lengths, you're on your own.

CrossEye:
This looks good to me.  Any other feedback, anyone?

Could you squash the commits?

kwijibo:
OK doke. I was a bit clumsy with rebase, but I think the commits are squashed ok now...

buzzdecafe:
LGTM :cow2: 

thanks for persevering

davidchambers:
Please run `git commit --amend` and rewrite the commit message to summarize the patch, rather than give a blow-by-blow account. :)

src/transpose.js
```diff
@@ -0,0 +1,41 @@
+var _curry1 = require('./internal/_curry1');
+/**
+ * Transposes the rows and columns of a 2D list.
+ * When passed a list of `n` lists of length `x`,
+ * returns a list of `x` lists of length `n`.
+ *
+ *
+ * @func
+ * @memberOf R
+ * @since v0.19.0
```
davidchambers: Please remove this line. We'll add it when we publish the next version.

src/transpose.js
```diff
@@ -0,0 +1,41 @@
+var _curry1 = require('./internal/_curry1');
+/**
```
davidchambers: Two empty lines after imports, please.

```javascript


/**
 * ...
 */
```

src/transpose.js
```diff
@@ -0,0 +1,41 @@
...
+    while (y < inlen) {
+      result[y] = result[y] || [];
+      result[y].push(innerlist[y]);
+      y = y + 1;
+    }
+    x = x + 1;
+  }
+  return result;
+});
+
```
davidchambers: :space_invader:

src/transpose.js
```diff
@@ -0,0 +1,41 @@
...
+  var x = 0;
+  var result = [];
+  while (x < outlen) {
+    var innerlist = outerlist[x];
+    var y = 0;
+    var inlen = innerlist.length;
+    while (y < inlen) {
+      result[y] = result[y] || [];
+      result[y].push(innerlist[y]);
+      y = y + 1;
```
davidchambers:
```javascript
y += 1;
```

src/transpose.js
```diff
@@ -0,0 +1,41 @@
...
+    }
+    x = x + 1;
```
davidchambers:
```javascript
x += 1;
```

src/transpose.js
```diff
@@ -0,0 +1,41 @@
...
+ *
+ *      R.transpose([[1, 'a'], [2, 'b'], [3, 'c']]) //=> [[1, 2, 3], ['a', 'b', 'c']]
+ *      R.transpose([[1, 2, 3], ['a', 'b', 'c']]) //=> [[1, 'a'], [2, 'b'], [3, 'c']]
+ *
+ * If some of the rows are shorter than the following rows, their elements are skipped:
+ *
+ *      R.transpose([[10, 11], [20], [], [30, 31, 32]]) //=> [[10, 20, 30], [11, 31], [32]]
+ */
+module.exports = _curry1(function transpose(outerlist) {
+  var outlen = outerlist.length;
```
davidchambers: We needn't cache this value: we can rely on JS engines to perform this optimization.

src/transpose.js
```diff
@@ -0,0 +1,41 @@
...
+  var x = 0;
+  var result = [];
+  while (x < outlen) {
+    var innerlist = outerlist[x];
+    var y = 0;
+    var inlen = innerlist.length;
+    while (y < inlen) {
+      result[y] = result[y] || [];
```
davidchambers: Could you replace the `||` with an explicit check? Using `||` with non-Boolean operands is a common source of bugs (and has been responsible for half a dozen Ramda bugs).

test/transpose.js
```diff
@@ -0,0 +1,18 @@
...
+  it('copes with empty arrays', function() {
+    eq(R.transpose([]), []);
+  });
+});
+
```
davidchambers: :space_invader:

src/transpose.js
kwijibo: @davidchambers this is a good point. You may well need to have `false`, `null`, `undefined` elements in an array (eg `undefined` elements in a transposed column). I have added a test to check for this.

kwijibo:
Done. Is there  a style guide document somewhere that CONTRIBUTING.md  could point to?
And a key for the emoji semantics... :thought_balloon:

src/transpose.js
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Transposes the rows and columns of a 2D list.
+ * When passed a list of `n` lists of length `x`,
+ * returns a list of `x` lists of length `n`.
+ *
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig transpose :: [[a]] -> [[a]]
```
davidchambers: The `@sig` needn't include the function name. `[[a]] -> [[a]]` would suffice.

davidchambers:
> Is there a style guide document somewhere that CONTRIBUTING.md could point to?

That would be very useful. Such a document does not yet exist.

> And a key for the emoji semantics...

`:space_invader:` (:space_invader:) indicates a stray empty line (in this case at the end of a file).

src/transpose.js
```diff
@@ -0,0 +1,42 @@
...
+ *      R.transpose([[10, 11], [20], [], [30, 31, 32]]) //=> [[10, 20, 30], [11, 31], [32]]
+ */
+module.exports = _curry1(function transpose(outerlist) {
+  var x = 0;
+  var result = [];
+  while (x < outerlist.length) {
+    var innerlist = outerlist[x];
+    var y = 0;
+    while (y < innerlist.length) {
+      if (!result.hasOwnProperty(y)) {
```
davidchambers: I haven't quite wrapped my head around this implementation, but I wonder whether we could use a length check of some sort here. It seems odd to treat a list as a dictionary.

src/transpose.js
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * If some of the rows are shorter than the following rows, their elements are skipped:
+ *
+ *      R.transpose([[10, 11], [20], [], [30, 31, 32]]) //=> [[10, 20, 30], [11, 31], [32]]
+ */
+module.exports = _curry1(function transpose(outerlist) {
+  var x = 0;
+  var result = [];
+  while (x < outerlist.length) {
+    var innerlist = outerlist[x];
```
davidchambers: I would prefer to use `i` and `j` rather than `x` and `y`.

CrossEye: Agreed.

kwijibo:
Space invaders removed, `x`s changed for `i`s.

PR #1527, opened 2015-11-20
Change assocPath behavior for empty path

gilligan:
R.assocPath([], 0, {a:1}) should return 0 not {a:1}

As discussed in #1510 the previous behavior of returning the full object does not quite match the desired logic that would also be required for lensPath. From a lens point of view the focus is on the empty path and thus on the whole object, meaning the whole object is being overwritten and not "nothing is being ovewritten"


davidchambers:
:deciduous_tree:

CrossEye:
:herb:

PR #1529, opened 2015-11-21
Replace jshint and jscs with eslint

gilligan:
### what
As mentioned on gitter I would like to replace `jshint` and `jscs` with `eslint`. 

### why
eslint is flexible and powerful enough to replace the functionality of both tools. Even if rules beyond current capabilities might be required at some point there is a nice plugin architecture which would allow for custom rules. Apart from that jshint is somewhat archaic.

### status
I created a `.eslintrc` based on the existing linting configurations. The configuration was extended with rules such as http://eslint.org/docs/rules/wrap-iife and http://eslint.org/docs/rules/one-var plus a various rules for catching obvious mistakes such as dangling commas or duplicate function arguments. Obviously all sources were fixed to adhere to the rules and `npm run lint` reports no errors.


buzzdecafe:
fine with me, i have no great love for jscs or jshint. so if we can use 1 tool, and reduce our conf-file count, then let's.

src/internal/_flatCat.js
```diff
@@ -29,4 +29,4 @@ module.exports = (function() {
       }
     };
   };
-}());
+})();
```
CrossEye: I'm curious as to the rationale for this.  Is this simply the eslint house style?  Is it configurable?

I admit to a minor preference for the other style, `(function() {/* ... */}())` over `(function() {/* ... */})()`, but it's a fairly minor things.  Still, I don't want to change this for no good reason.  And I think "that's what the linter prefers" is no good reason.  If someone can make a compelling argument, or if eslint doesn't even allow my preferred syntax, then I could see just going with the flow.

CrossEye: Never mind.  Looked it up and see this non-default preference in the config file above.  Will add a comment there.

.eslintrc
```diff
@@ -0,0 +1,86 @@
...
+      2,
+      "last"
+    ],
+    "curly": [
+      2,
+      "all"
+    ],
+    "dot-notation": 0,
+    "eol-last": 2,
+    "wrap-iife": [2, "inside"],
```
CrossEye: That's causing a number of tweaks that seem to have little reason.  Were there more if it was the other way?

I'd rather do "outside", although it's a minor preference.

CrossEye:
I'd be quite happy to have this.  I do have a caveat about one of the flags, but I think this is worth doing, overall.

src/uniqBy.js
```diff
@@ -36,12 +36,12 @@ module.exports = _curry2(/* globals Set */ typeof Set === 'undefined' ?
...
-        result        = [],
-        nullExists    = false,
-        negZeroExists = false,
-        idx           = 0;
+      applied       = [], appliedItem, item,
+      prevSetSize   = 0,  newSetSize,
+      result        = [],
+      nullExists    = false,
+      negZeroExists = false,
+      idx           = 0;
```
scott-christopher: The `=` here either needs realigning or the extra whitespace used for alignment should be removed.

scott-christopher:
It might be worth standardising on the use of one var per block vs one var per variable while we're at it too. It looks like the `one-var` eslint config covers this, which I'd propose using:
```
"one-var": [2, {
    "uninitialized": "always", // Exactly one declaration for uninitialized variables per function
    "initialized": "never" // Exactly one declarator per initialized variable declaration per function
}]
```

.eslintrc
gilligan: Neither inside nor outside was used consistently so I basically just picked one. I'm totally fine with changing it to "outside" if that is preferred.

src/uniqBy.js
gilligan: Everything is indented by 2 spaces.

scott-christopher: Sorry, I meant the `=` before `new Set()` is no longer vertically aligned with the other `=`.

gilligan:
I now switched the `iife` config so the expected syntax is now : 
```js
(function () { ... }()); 
```
As you can see there were number of files where it was used the other way round.

gilligan:
@scott-christopher I added the `one-var` rule as suggested by you.

.eslintrc
```diff
@@ -0,0 +1,87 @@
+{
+  "env": {
+      "node": true,
+      "mocha": true
```
davidchambers: :triangular_ruler: We should lint __.eslintrc__. ;)

src/empty.js
```diff
@@ -40,7 +40,7 @@ module.exports = _curry1(function empty(x) {
     _isObject(x) ?
       {} :
     _isArguments(x) ?
-      (function() { return arguments; }()) :
+      (function() { return arguments; })() :
```
davidchambers: This change should be reverted, shouldn't it?

gilligan:
While adjusting the sources for `one-var` I realised that sometimes there is a blank line after the var statements block and sometimes there is not. Is this something you would like to see standardised the one or the other way @davidchambers @scott-christopher ?

davidchambers:
This needn't be standardized.

gilligan:
I think the following rules could be enabled because they are pretty much uncontroversial and just find silly mistakes : 

http://eslint.org/docs/rules/comma-dangle
http://eslint.org/docs/rules/no-debugger 
http://eslint.org/docs/rules/no-dupe-args 
http://eslint.org/docs/rules/no-dupe-keys 
http://eslint.org/docs/rules/no-duplicate-case 
http://eslint.org/docs/rules/no-extra-semi
http://eslint.org/docs/rules/no-unreachable

There is no good reason to allow any of what the above checks I suppose ?

davidchambers:
Go ahead and enable those rules, @gilligan. :)

gilligan:
Added the aforementioned rules. There were some dangling commas at least ;--} Are there any other modifications anyone would like to see to the rules ? Otherwise I'd squash commits and make this a proper PR.

gilligan:
stashed the commits, updated title and description. The changes are of course pretty much all over the place but all of them very minimal.

.eslintrc
```diff
@@ -0,0 +1,51 @@
...
+    "quotes": [ 2, "single", "avoid-escape" ],
+    "comma-dangle": [ 2, "never" ],
+    "no-debugger": 2,
+    "no-dupe-args": 2,
+    "no-dupe-keys": 2,
+    "no-duplicate-case": 2,
+    "no-extra-semi": 2,
+    "no-unreachable": 2,
+  }
+}
```
davidchambers: Could we make the formatting of this file consistent? We use `[a, b, c]` in some places and `[ a, b, c ]` in others. You could use `JSON.stringify(x, null, 2)` to normalize the formatting.

davidchambers:
:deciduous_tree:

CrossEye:
:fireworks:

davidchambers:
@gilligan, can we remove __.jscsrc__, __.jshintignore__, and __.jshintrc__?

CrossEye:
I believe we can.

gilligan:
Oh, yes we can/should indeed. Forgot that :/

davidchambers:
Mind opening a pull request, @gilligan?

gilligan:
@davidchambers #1559 

I just now realised that there were also still plenty of jshint switches across various test files ;(
The PR removes all of them.

davidchambers:
Thanks!

PR #1530, opened 2015-11-21
[perf] Replace Set with indexOf in _contains

scott-christopher:
Addresses #1508

While the use of `Set` in `_contains` provided a nice boost in performance, after further thought I have come to realise that it could be replaced with a simpler `Array.prototype.indexOf` for the instances where `Set` was being utilised (though `NaN` needs to now be handled explicitly), removing the extra insertion cost and memory allocation to create the `Set`. This should remove some extra GC thrashing when used in a tight loop too, such as those often in benchmark code.

I've included a new benchmark file for `R.contains` and run it against the previous implementations, resulting in the following:

Implementation    | Hz         | Margin of Error
------------------|------------|----------------
v0.8              |     73,693 | ±1.31%
Set               |    341,266 | ±0.66%
indexOf (this PR) | 11,372,373 | ±0.84%
nativeIndexOf     | 13,820,992 | ±0.63%

### DIFF
--- a/lib/bench/contains.bench.js
+++ b/lib/bench/contains.bench.js
@@ -1,0 +1,23 @@
+var contains = require('../..').contains;
+
+var alphabet = 'abcdefghijklmnopqrstuvwxyz'.split('');
+
+function randomLowerAlpha() {
+  return String.fromCharCode(97 + (Math.random() * 26) >>> 0);
+}
+
+function nativeIndexOf(x, xs) {
+  return xs.indexOf(x) >= 0;
+}
+
+module.exports = {
+  name: 'contains',
+  tests: {
+    'contains(c, cs)': function() {
+      return contains(randomLowerAlpha(), alphabet);
+    },
+    'native indexOf': function() {
+      return nativeIndexOf(randomLowerAlpha(), alphabet);
+    }
+  }
+};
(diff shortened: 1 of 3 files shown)


scott-christopher:
Update: I've moved the optimisations out of `_contains` and into `_indexOf`. This unfortunately drops the performance down from `11,372,373 Hz` to `8,443,488 Hz` in the included benchmark, but will give a nice boost to other functions that depend on `_indexOf`.

I've also added a check to ensure `Array.prototype.indexOf` exists for older versions of IE, falling back to the previous implementation where necessary.

CrossEye:
This is great stuff, @scott-chistopher!

lib/bench/contains.bench.js
```diff
@@ -0,0 +1,23 @@
+var contains = require('../..').contains;
+
+var alphabet = 'abcdefghijklmnopqrstuvwxyz'.split('');
+
+function randomLowerAlpha() {
+  return String.fromCharCode(97 + (Math.random() * 26) >>> 0);
+}
+
+function nativeIndexOf(x, xs) {
+  return xs.indexOf(x) > 0;
```
davidchambers: Should this be `>= 0`?

davidchambers:
Impressive!

lib/bench/contains.bench.js
scott-christopher: Yep. :see_no_evil: :zap:

davidchambers:
:deciduous_tree:

PR #1533, opened 2015-12-03
R.scan now dispatches to scan method.

jethrolarson:
Fixes half of #1179

### DIFF
--- a/src/scan.js
+++ b/src/scan.js
@@ -1,10 +1,12 @@
 var _curry3 = require('./internal/_curry3');
+var _isArray = require('./_isArray');
 
 
 /**
  * Scan is similar to reduce, but returns a list of successively reduced values
  * from the left
  *
+ * Dispatches to the `scan` method of the third argument, if present.
  * @func
  * @memberOf R
  * @since v0.10.0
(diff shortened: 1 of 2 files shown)


jethrolarson:
Ping

src/scan.js
```diff
@@ -21,7 +23,12 @@ var _curry3 = require('./internal/_curry3');
  *      var factorials = R.scan(R.multiply, 1, numbers); //=> [1, 1, 2, 6, 24]
  */
 module.exports = _curry3(function scan(fn, acc, list) {
-  var idx = 0, len = list.length, result = [acc];
+  if (!_isArray(list) && typeof list.scan === 'function') {
```
davidchambers: Is `!_isArray(list)` necessary?

src/scan.js
```diff
@@ -1,10 +1,12 @@
 var _curry3 = require('./internal/_curry3');
+var _isArray = require('./_isArray');
 
 
 /**
  * Scan is similar to reduce, but returns a list of successively reduced values
  * from the left
  *
+ * Dispatches to the `scan` method of the third argument, if present.
  * @func
```
davidchambers:
```diff
  * Dispatches to the `scan` method of the third argument, if present.
+ *
  * @func
```

src/scan.js
jethrolarson: I suppose not, if people are modifying the array prototype they might be doing so knowingly. I know we intercept it this way for reduce

davidchambers: I'd prefer to remove the guard, in that case. It strikes me as an unnecessary exception to this rule:

> Dispatches to the `scan` method of the third argument, if present.

CrossEye:
While this might be worth doing, we would need to revisit the implementation these days.  Closing without prejudice.

Issue #1536, opened 2015-12-07
Update path documentation


steida:
Because this check https://github.com/ramda/ramda/blob/cffff75259a5b251811bc5197ea6ec41b33731e2/src/path.js#L26
both null and undefined are checked. Or fix code and use !== instead.

`R.path(['a', 'b'], {c: {b: 2}}); //=> undefined or null`

svozza:
Do you mean something like this is incorrect?

```javascript
R.path('a', 'b', {a: 1, b: null}) => // undefined
```

iofjuupasli:
```js
R.path(['foo', 'bar'], {foo: null}) === null
```

It returns value of `obj.foo`. But as I understand it should return value of `obj.foo.bar` or `undefined`

svozza:
Ah yes, I get what you mean now. Yeah, I'd expect `undefined` there too.

CrossEye:
I see @svozza's fine fix in #1537.

My only question is whether we want to fix the code, or whether we're better adjusting the documentation.  Which is the better response for

```js
R.path(['foo', 'bar'], {foo: null}) 
```

`undefined`?  or `null`?

steida:
I think documentation should be updated. We need to test both values.
undefined is absence of evidence, while null is evidence of absence.

svozza:
I'm conflicted about this, on the one hand I wish JavaScript only had one `null` value (or `Maybe` even none :wink:) but we are where we are. I've used APIs where `null` had meaning as opposed to `undefined`and a user doing so is not necessarily doing anything incorrect by respecting the distinction the language makes. End of the day, this is just wrong:

```js
R.path(['a', 'b', 'c'], {a: null}); // => null
```

PR #1537, opened 2015-12-07
Add explicit guards in R.path for null and undefined.

svozza:
Fixes issue #1536.

### DIFF
--- a/src/path.js
+++ b/src/path.js
@@ -23,7 +23,10 @@ module.exports = _curry2(function path(paths, obj) {
   } else {
     var val = obj;
     var idx = 0;
-    while (val != null && idx < paths.length) {
+    while (idx < paths.length) {
+      if (val == null) {
+        return;
+      }
       val = val[paths[idx]];
       idx += 1;
     }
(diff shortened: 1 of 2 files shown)


src/path.js
```diff
@@ -23,7 +23,10 @@ module.exports = _curry2(function path(paths, obj) {
...
     var val = obj;
     var idx = 0;
-    while (val != null && idx < paths.length) {
+    while (val !== void 0 && idx < paths.length) {
+      if (val === null) {
+        return;
+      }
       val = val[paths[idx]];
       idx += 1;
     }
```
davidchambers: It seems we could do this instead:

```diff
@@ -23,8 +23,8 @@ module.exports = _curry2(function path(paths, obj) {
   } else {
     var val = obj;
     var idx = 0;
-    while (val !== void 0 && idx < paths.length) {
-      if (val === null) {
+    while (idx < paths.length) {
+      if (val == null) {
         return;
       }
       val = val[paths[idx]];
```

This would avoid separate code paths for `null` and `undefined`. :)

svozza: Nice. I much prefer that.

davidchambers:
:deciduous_tree:

PR #1538, opened 2015-12-07
make R.path([k], o) equivalent to R.prop(k, o)

### LINKED ISSUE #1522: R.prop and R.path inconsistency
As [mentioned in gitter](https://gitter.im/ramda/ramda?at=5647fa688b242470793dc6c8), I’ve found a scenario where the behavior of  `R.path` differs from `R.prop`:

```js
R.prop('foo', undefined); // throws 
R.path(['foo'], undefined); // -> undefined
```

IMHO if the path array of `R.path` contains only one element the function should be equivalent to `R.prop`.

davidchambers:
Fixes #1522

Set the query string to __?w=__ when viewing the diff to ignore the indentation change.

### DIFF
--- a/src/path.js
+++ b/src/path.js
@@ -18,18 +18,14 @@ var _curry2 = require('./internal/_curry2');
  *      R.path(['a', 'b'], {c: {b: 2}}); //=> undefined
  */
 module.exports = _curry2(function path(paths, obj) {
-  if (obj == null) {
-    return;
-  } else {
-    var val = obj;
-    var idx = 0;
-    while (idx < paths.length) {
-      if (val == null) {
-        return;
-      }
-      val = val[paths[idx]];
-      idx += 1;
+  var val = obj;
+  var idx = 0;
+  while (idx < paths.length) {
+    if (val == null) {
+      return;
     }
-    return val;
+    val = val[paths[idx]];
+    idx += 1;
   }
+  return val;
 });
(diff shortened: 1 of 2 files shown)


gilligan:
👍

CrossEye:
:herb:, although I must admit I've only looked at it briefly.

richardscarrott:
Hi @davidchambers, from what I can tell this PR intended to align `R.prop` and `R.path` by letting `R.path` throw an error if the type passed in is null or undefined?

If so I'm not sure this works because of the null check in the while loop, i.e. the tests that were removed actually still pass.

```
R.path(['a'], null); //=> undefined
R.path(['a'], void 0); //=> undefined
```

Would it make sense for the check in the while loop to be removed too so it'd result in:

```
R.path(['a'], null); //=> TypeError
R.path(['a'], undefined); //=> TypeError
R.path(['a', 'b'], { a: null }); //=> TypeError
R.path(['a', 'b'], { a: { } }); //=> undefined
R.path(['a', 'b'], { a: { b: 'c' } }); //=> 'c'
```

Then if you needed to handle possible null values you'd probably compose `R.prop` using a Maybe monad instead of `R.path`?

davidchambers:
You're quite right, @richardscarrott!

The position we take with Ramda is that the result of applying a function to arguments of the wrong types is undefined (in the English sense of the word rather than the JavaScript one).

For *valid* inputs the equivalence holds:

    forall k :: String, o :: Object. R.path([k], o) = R.prop(k, o)

We tend not to include code to handle invalid inputs explicitly (though [Sanctuary][1] is a Ramda-inspired library which does so). This pull request removed explicit handling of `null`/`undefined`, which I consider invalid values for the second argument. As you have rightly pointed out, though, the `== null` check in the `for` loop catches these values anyway.

Given our stance a proposal to add code to throw an exception if the second argument is `null`/`undefined` is unlikely to be accepted.


[1]: http://sanctuary.js.org/

PR #1539, opened 2015-12-10
`isLt`, `isLte`, `isGt`, `isGte` implemented

niloy:
Implemented `isLt`, `isLte`, `isGt`, `isGte`.

``` js
R.map(R.isLte(3), [1, 2, 3, 4]); // [true, true, true, false]
R.map(R.isLt(3), [1, 2, 3, 4]); // [true, true, false, false]
R.map(R.isGt(3), [1, 2, 3, 4]); // [false, false, false, true]
R.map(R.isGte(3), [1, 2, 3, 4]); // [false, false, true, true]
```

### DIFF
--- a/src/isGt.js
+++ b/src/isGt.js
@@ -1,0 +1,21 @@
+var flip = require('./flip');
+var gt = require('./gt');
+
+/**
+ * Returns `true` if the second argument is greater than the first; `false`
+ * otherwise. This function is meant to be used in a curried fashion.
+ *
+ * @func
+ * @memberOf R
+ * @category Relation
+ * @sig Ord a => a -> a -> Boolean
+ * @param {*} a
+ * @param {*} b
+ * @return {Boolean}
+ * @see R.gt
+ * @example
+ *
+ *      R.map(R.isGt(3), [1, 2, 3, 4]); // [false, false, false, true]
+ *      R.ifElse(R.isGt(3), doSomething, doSomethingElse);
+ */
+module.exports = flip(gt);
--- a/src/gt.js
+++ b/src/gt.js
@@ -5,6 +5,11 @@ var _curry2 = require('./internal/_curry2');
  * Returns `true` if the first argument is greater than the second; `false`
  * otherwise.
  *
+ * Note: `R.gt(18)` does not create a function which checks if a number is
+ * greater than 18. This is because `R.gt(18)` is equivalent to `R.gt(18, R.__)`.
+ * If you want to create a function to check if a number is greater than
+ * 18, use `R.isGt(18)` instead.
+ *
  * @func
  * @memberOf R
  * @since v0.1.0
@@ -13,7 +18,7 @@ var _curry2 = require('./internal/_curry2');
  * @param {*} a
  * @param {*} b
  * @return {Boolean}
- * @see R.lt
+ * @see R.lt, R.isGt
  * @example
  *
  *      R.gt(2, 1); //=> true
--- a/src/gte.js
+++ b/src/gte.js
@@ -5,6 +5,11 @@ var _curry2 = require('./internal/_curry2');
  * Returns `true` if the first argument is greater than or equal to the second;
  * `false` otherwise.
  *
+ * Note: `R.gte(18)` does not create a function which checks if a number is greater
+ * than or equal to 18. This is because `R.gte(18)` is equivalent to `R.gte(18, R.__)`.
+ * If you want to create a function to check if a number is greater than or equal
+ * to 18, use `R.isGte(18)` instead.
+ *
  * @func
  * @memberOf R
  * @since v0.1.0
@@ -13,7 +18,7 @@ var _curry2 = require('./internal/_curry2');
  * @param {Number} a
  * @param {Number} b
  * @return {Boolean}
- * @see R.lte
+ * @see R.lte, R.isGte
  * @example
  *
  *      R.gte(2, 1); //=> true
--- a/src/isGte.js
+++ b/src/isGte.js
@@ -1,0 +1,21 @@
+var flip = require('./flip');
+var gte = require('./gte');
+
+/**
+ * Returns `true` if the second argument is greater than or equal to the first;
+ * `false` otherwise. This function is meant to be used in a curried fashion.
+ *
+ * @func
+ * @memberOf R
+ * @category Relation
+ * @sig Ord a => a -> a -> Boolean
+ * @param {*} a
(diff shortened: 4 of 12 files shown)


buzzdecafe:
lgtm. :cow2:

Side note: to my mind, this makes the placeholder redundant.

src/isGt.js
```diff
@@ -0,0 +1,22 @@
+var flip = require('./flip');
+var gt = require('./gt');
+
+/**
+ * Returns `true` if the second argument is greater than the first; `false`
+ * otherwise. This function is meant to be used in a curried fashion.
+ *
+ * @func
+ * @memberOf R
+ * @since v0.1.8
```
svozza: Is that version number right?

davidchambers: Please don't include `@since` when defining new functions.

asaf-romano: R.prop(__, OBJ) is quite useful (same goes for contains).

asaf-romano:
@buzzdecafe  R.prop(__, OBJ) is quite useful (same goes for contains).

buzzdecafe:
>  R.prop(__, OBJ) is quite useful (same goes for contains).

maybe so. object mutation can also be useful.

CrossEye:
> Side note: to my mind, this makes the placeholder redundant. 

It doesn't to me -- not quite yet -- as there are still a few non-commutative binary operators that probably also need similar glosses.  `subtract`, `divide`, and `modulo` come immediately to mind.  I might also agree with @asaf-romano about wanting something for `prop` and `contains`, although I'm less certain of those.

I'm not bothered by having to apply `flip` to functions when needed.  It's the binary operators that really _look_ to be lying to you such as `lt(5)` and `subtract(42)` which concern me.

niloy:
@davidchambers Removed `@since`

svozza:
I can count on one hand the amount of times I've used the placeholder for the `gt` or `lt` functions. The main times I use it are for `assoc*`, `difference` and`prop` but mostly `contains`. That's not to mention various custom functions that have been defined with `curry`.

davidchambers:
I'm not sure about this change. Currently we can write `R.gte(_, 18)`, which is fairly clear. With this change we'd be able to write `R.isGte(18)`, which is certainly slightly clearer.

But `R.gte(_, 18)` is *not* the concern, as it's not a source of confusion. `R.gte(18)` is the problem. The existence of `R.isGte` would not prevent `R.gte(18)` from being written, nor from being misread. Instead of a situation in which one might write `R.gte(18)` when one meant to write `R.gte(_, 18)`, we'd be in a situation in which one might write `R.gte(18)` when one meant to write `R.isGte(18)`.

niloy:
@davidchambers Should `R.gte` throw when only 1 argument is passed and suggest to use `R.isGte` instead?

buzzdecafe:
> `R.gte(18)` is the problem.

indeed. this all feels very familiar. This is why placeholders for these non-commutative infix ops make some sense. But since it is difficult to distinguish/communicate which functions were non-commutative the placeholder was generalized. IMO that should be reverted, but i understand i am in the minority in that view. 

> Should `R.gte` throw when only 1 argument is passed and suggest to use `R.isGte` instead?

pardon my butting in, but my $0.02: no, it absolutely should not throw. there is nothing wrong with `R.gte(x)`, and we should give the user the benefit of the doubt.

buzzdecafe:
... and i don't have a better solution for converting these infixes to prefix. I suspect there is not a good, comprehensive solution.

davidchambers:
> Should `R.gte` display console warning when only 1 argument is passed and suggest to use `R.isGte` instead?

That would require currying `R.gte` manually, which would be no fun at all (assuming we'd still support the placeholder).

I wouldn't want the function to throw, in any case. As @paldepind pointed out, one can already write `const gte = R.lt;`, and then use `gte(18)`.

CrossEye:
I'm generally in favor of this, with the caveat that the documentation for `lt`, `gte`, etc. will need some significant warnings about the potential issues and links to `isLt`, `isGte`, etc.

The current situation has the same issue that @davidchambers describes, but I believe that `isGte(18)` is substantially clearer than `gte(_, 18)`, not just slightly clearer.

I'm not even close to ready to give up our "curry everything" ideal, but this is tempting me to push for allowing _only_ these signatures:

```js
    isGte(18)(someVal); // only unary, returns unary
    gte(someVal, 18);  // only binary
```

niloy:
I think unary only `isGte` is a good idea and would like to implement it if others agree. In the meantime, I will update the documentation about the hazards of using single argument to `lt`, `gt` etc.

paldepind:
>  I'm not even close to ready to give up our "curry everything" ideal, but this is tempting me to push for allowing _only_ these signatures:

> ```js
    isGte(18)(someVal); // only unary, returns unary
    gte(someVal, 18);  // only binary
```

I can see the temptation. It is pretty much the same as making `R.gte` throw when given only a single argument and making `R.isGte` throw when given two.

Solving it through documentation is probably a better approach.

niloy:
Added documentation about possible incorrect usage of `R.lt(18)` etc.

CrossEye:
> Solving it through documentation is probably a better approach.

It definitely is.  But the temptation is real, as it would in some way rid us of this problem.

CrossEye:
I believe this didn't get resolved.  Reopening to complete the discussion.

niloy:
Kewl :+1: :smiley:

kedashoe:
I have had `isGt` and `isLt` in my ramda wrapper for forever. I don't mind the placeholder, so it doesn't really bother me to write `R.gt(R.__, x)`. The thing is, I literally never use the comparison functions without the placeholder as the first argument. If I already know all the arguments for my comparison, I just write `x > 10` rather than `R.gt(x, 10)`. Otherwise, I need them as part of some composition pipeline.. which means I never actually use `gt` and `lt` at all, only `isGt` and `isLt`.

As ramda is still pre 1.0, I would rather we just switch the functions we already have than include all 8 comparison functions. But I do use these, so if there is not the stomach for that, I would be +1 for this.

tycho01:
On the placeholder, most examples in favor of it here seem binary functions where `flip` might do?
(I don't have a strong stance in that discussion, was just trying to understand as it seems I'm missing something.)

CrossEye:
Yes, `flip` would do for many cases where the placeholder is used.  But it's not applicable to all.

Closing again.  We still need to address this, but obviously I was overeager in reopening it when I did.  :frowning_face:

PR #1540, opened 2015-12-10
RFC : property based tests using jsverify

gilligan:
As mentioned on gitter I would like to look into providing property tests using jsverify (https://github.com/jsverify/jsverify) 

For now this PR only contains a tiny example which tests `R.add` for 

* commutativity
* associativity 
* neutral element 

Writing the tests will in many cases probably not be too hard. I am having more difficulties designating the properties that should hold. I would thus like to ask for some input : What are some functions that you would like to see some properties tested for and which are they. I am happy to implement and then we can see if the outcome is something that can be considered a useful addition to Ramda or not.

### DIFF
--- a/test/add.js
+++ b/test/add.js
@@ -1,5 +1,6 @@
 var R = require('..');
 var eq = require('./shared/eq');
+var jsc = require('jsverify');
 
 
 describe('add', function() {
@@ -10,6 +11,20 @@ describe('add', function() {
   it('is curried', function() {
     var incr = R.add(1);
     eq(incr(42), 43);
+  });
+
+  describe('properties', function() {
+    jsc.property('commutative', 'integer', 'integer', function(a, b) {
+      return R.add(a,b) === R.add(b,a);
+    });
+
+    jsc.property('associative', 'integer', 'integer', 'integer', function(a, b, c) {
+      return R.add(a, R.add(b, c)) === R.add(R.add(a, b), c);
+    });
+
+    jsc.property('identity element', 'integer' , function(a) {
+      return R.add(a, 0) === a && R.add(0, a) === a;
+    });
   });
 
 });
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "eslint": "^1.10.1",
     "handlebars": "3.0.x",
     "js-yaml": "^3.2.5",
+    "jsverify": "^0.7.1",
     "mocha": "2.x.x",
     "q": "^1.1.1",
     "ramda": "0.17.x",
--- a/test/adjust.js
+++ b/test/adjust.js
@@ -1,5 +1,6 @@
 var R = require('..');
 var eq = require('./shared/eq');
+var jsc = require('jsverify');
 
 describe('adjust', function() {
   it('applies the given function to the value at the given index of the supplied array', function() {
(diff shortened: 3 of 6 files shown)


test/add.js
```diff
@@ -1,5 +1,6 @@
 var R = require('..');
 var eq = require('./shared/eq');
+var jsc = require("jsverify");
```
davidchambers: Nit: Single quotes, please. :)

buzzdecafe:
:heart: this idea, would be nice to have some evidence that things behave as we expect

test/add.js
gilligan: Eeek, where did that come from.
Also: i will add an eslint rule for that ;)

CrossEye:
I love the idea as an addition to the more specific testing we do, rather than as a replacement for it.  I've always thought of unit tests serving a dual role: both the obvious one of testing features and the equally important one of serving as additional API guides for users.  So I don't really want to lose the sorts of tests we've been doing.  But done this way as an addition, I think it's fantastic.  I will look around for specific properties to test sometime.  It might not be this coming week, though.  Overseas travel may leave me with more time for this, or less; I can't tell yet.

Very nice!

test/add.js
```diff
@@ -12,4 +13,18 @@ describe('add', function() {
...
+      return R.add(a,b) === R.add(b,a);
+    });
+
+    jsc.property("associative", "integer", "integer", "integer", function (a, b, c) {
+      return R.add(a, R.add(b, c)) === R.add(R.add(a, b), c);
+    });
+
+    jsc.property("neutral element", "integer" , function (a) {
+      return R.add(a, 0) === a;
+    });
```
CrossEye: Probably "identity element" instead.

But also this should either be described here as the "right identity element" with a comparable "left identity element", or they should be combined with something like

```js
  R.add(a, 0) === a && R.add(0, a) === 0;
```

davidchambers:
This is tremendously exciting!

gilligan:
I am currently trying to cover all properties defined by the fantasyland spec followed by functions where the properties appear obvious to me.

gilligan:
I will close this for now since it has been sitting here for way too long and I don't find time to finish it. I should rather create a separate PR later that adds smaller sets of property tests instead of going for "PROPERTY TEST ALL THE THINGS1!1!!" ;-) thanks for the interest.

davidchambers:
I look forward to seeing a smaller pull request at some point. :)

dubzzz:
Hi there,

I wanted to know what was the status of property based testing on **ramda** side.
Are you going to keep it? Are you going to increase its coverage? Or more generally what do you think of it?

Basically my question comes from the fact that more than a year ago **jsverify** owner announced that he stopped the maintenance of the library. Moreover, the most recent commit was done the 6th of September 2019.

As core contributor and author of [**fast-check**](https://github.com/dubzzz/fast-check/) - _the new library to do property based testing in JavaScript_ - I wanted to suggest you a migration to **fast-check**.

Basically I already worked on migrating the tests you had but I was a bit shy to open a PR for that. _By the way, I initially did that to check that **fast-check** was able to answer all the classical needs of pbt._

Would you be interested by such migration?
Code here: https://github.com/ramda/ramda/compare/master...dubzzz:pbt-fast-check

@davidchambers @CrossEye @buzzdecafe

Bradcomp:
We have a number of PBTs in Ramda, but there hasn't been a concerted effort to increase coverage beyond a few explorations. I would welcome any additional ventures into that area.

CrossEye:
Absolutely.  I would love to see some PRs to extend PBTs in JS.  And I'm not fussy about the library we use to do it.  :smile:

dubzzz:
Thanks a lot for your quick answers, I opened a Pull Request related to this topic, see #3095

Issue #931, opened 2015-03-17
sister to groupBy


fyyyyy:
From remote servers we often get arrays of objects. These objects typically contain id fields which are unique. What i would like to do is convert the array of objects in to a hashmap with the object's ids so they are quickly accessible by id.

Currenty i use groupBy, however it creates a hashmap with an array storage that wraps the object. I would like to return the object directly under the specified key.

```js

var tracks = [ {id: 1, name: "track1"}, {id: 3, name.... ]
R.indexBy('id')(tracks) //  => {1: {id: 1, name: "track1"}, 3: {id: 3, name.... }}

fyyyyy:
This is the points free way i guess
 R.mapObj(R.head)(R.groupBy(R.get('id))(tracks));

kedashoe:
Soon you will be able to use [into](https://github.com/ramda/ramda/blob/master/src/into.js), then you can write something like

`var indexed = into({}, tuple(get('id'), identity), tracks);`

kedashoe:
Oops.. maybe ramda does not have a tuple-like function? I find creating a `double` is often quite handy for just this sort of situation.. so maybe `createMapEntry` with a `pipe`?

buzzdecafe:
would `tuple` be `keyFn -> valFn -> {k: v}`  ? or `keyFn -> valFn -> [k, v]`?

kedashoe:
> would tuple be keyFn -> valFn -> {k: v} ? or keyFn -> valFn -> [k, v]?

Maybe `tuple` does not fit ramda since it does not convey size, but in general a tuple is an ordered list, so the latter signature I think? Maybe `couple`? http://en.wikipedia.org/wiki/Tuple#Names_for_tuples_of_specific_lengths

buzzdecafe:
you could `nTuple(n:int, [vals])`, but how is that any better than just `[vals]`?

kedashoe:
ya I guess you could just use `Array`?

CrossEye:
If you want it entirely points-free, and to accept `'id'` as a parameter, this might work:

    var indexBy = R.curry(R.pipe(R.useWith(R.groupBy, R.prop, R.identity), R.mapObj(R.head)));
    indexBy('id', tracks);

(I'm bothered by that explicit `curry` call.  We've let #737 linger, and I think we really need to figure it out.  But there's another issue with `compose`/`pipe` we also want to address before 1.0: their variadic nature.  So I think we're going to have another pass at that.)

[...]

PR #1546, opened 2015-12-14
Add `indexBy` function

### LINKED ISSUE #1545: Support for `indexBy`
There's an example in the wiki on how to index an array of objects by a property: https://github.com/ramda/ramda/wiki/Cookbook#index-objects-by-property-name

Is there a reason why this isn't a "native" Ramda function? Would you accept a PR that introduces this function?

Chris911:
Resolves #1545

This adds a new `indexBy` function. Given a property name to index-by and a list of objects returns an object indexing the objects by the given property. Highly inspired by this [cookbook example](https://github.com/ramda/ramda/wiki/Cookbook#index-objects-by-property-name).

### DIFF
--- a/src/indexBy.js
+++ b/src/indexBy.js
@@ -1,0 +1,30 @@
+var _curry2 = require('./internal/_curry2');
+var _reduce = require('./internal/_reduce');
+
+
+/**
+ * Given a function that generates a key, turns a list of objects into an
+ * object indexing the objects by the given key. Note that if multiple
+ * objects generate the same value for the indexing key only the last value
+ * will be included in the generated object.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig (a -> String) -> [{k: v}] -> {k: {k: v}}
+ * @param {Function} fn Function :: a -> String
+ * @param {Array} array The array of objects to index
+ * @return {Object} An object indexing each array element by the given property.
+ * @example
+ *
+ *      var list = [{id: 'xyz', title: 'A'}, {id: 'abc', title: 'B'}];
+ *      R.indexBy(R.prop('id'), list);
+ *      //=> {abc: {id: 'abc', title: 'B'}, xyz: {id: 'xyz', title: 'A'}}
+ */
+module.exports = _curry2(function indexBy(fn, list) {
+  return _reduce(function(acc, elem) {
+    var key = fn(elem);
+    acc[key] = elem;
+    return acc;
+  }, {}, list);
+});
--- a/test/indexBy.js
+++ b/test/indexBy.js
@@ -1,0 +1,25 @@
+var R = require('..');
+var eq = require('./shared/eq');
+
+
+describe('indexBy', function() {
+  it('indexes list by the given property', function() {
+    var list = [{id: 'xyz', title: 'A'}, {id: 'abc', title: 'B'}];
+    var indexed = R.indexBy(R.prop('id'), list);
+    eq(indexed, {abc: {id: 'abc', title: 'B'}, xyz: {id: 'xyz', title: 'A'}});
+  });
+
+  it('indexes list by the given property upper case', function() {
+    var list = [{id: 'xyz', title: 'A'}, {id: 'abc', title: 'B'}];
+    var indexed = R.indexBy(R.compose(R.toUpper, R.prop('id')), list);
+    eq(indexed, {ABC: {id: 'abc', title: 'B'}, XYZ: {id: 'xyz', title: 'A'}});
+  });
+
+
+  it('is curried', function() {
+    var list = [{id: 'xyz', title: 'A'}, {id: 'abc', title: 'B'}];
+    var indexed = R.indexBy(R.prop('id'))(list);
+    eq(indexed, {abc: {id: 'abc', title: 'B'}, xyz: {id: 'xyz', title: 'A'}});
+  });
+
+});


src/indexBy.js
```diff
@@ -0,0 +1,30 @@
+var _curry2 = require('./internal/_curry2');
+var groupBy = require('./groupBy');
+var head = require('./head');
+var mapObjIndexed = require('./mapObjIndexed');
+var prop = require('./prop');
+
+
+/**
+ * Given a property name, turns a list of objects into an object indexing
+ * the objects by then given property. Note that uf multiple objects have
```
scott-christopher: You've got a typo here:
> Note that _uf_ multiple objects

src/indexBy.js
```diff
@@ -0,0 +1,30 @@
...
+ * the same value for the indexing property only one object is collected,
+ * and there's no guarantee as to which one would it be.
+ *
+ * @func
+ * @memberOf R
+ * @since v0.19.0
```
scott-christopher: For new functions, the `@since` annotation will get added when the next release is tagged so this line should be removed. We should probably make a note of this in `CONTRIBUTING.md`.

src/indexBy.js
Chris911: Fixed. Thanks.

src/indexBy.js
Chris911: First contribution, didn't know. Fixed :+1:

src/indexBy.js
```diff
@@ -0,0 +1,30 @@
...
+ * @category List
+ * @sig String -> [a] -> a
```
scott-christopher: We should probably indicate that `a` should be an object in this signature.
e.g.
```
String -> [{k: v}] -> {k: {k: v}}
```

Chris911: That makes sense and sums up the function pretty nicely. Is your example a valid signature for Ramda doc?

scott-christopher:
This looks good to me, though I wonder whether this should handle the case where the given property name doesn't exist in any of the objects, as currently I assume they'll just end up in an `"undefined"` key of the resulting object.

Chris911:
It will result in an `undefined` key. This could be documented and/or tested if you think it's needed.

src/indexBy.js
scott-christopher: We really need to adopt some clear guidelines for defining these signatures, as there are currently no hard and fast rules. The docs have a few variations for representing an object, e.g. `{*}`, `{k: v}` and `{a}`, so I'm not too concerned among the choice of one of those, with a preference toward either `{k: v}` or `{a}`. Once we standardise on this at some point in the future, we can do a sweep through and consolidate all the existing definitions.

Chris911: Alright. I think what you suggested above sums up the function pretty good so I'll use that for now.

scott-christopher:
If we're happy to keep the `undefined` key, then mentioning it in the docs would be sufficient. The other option would be to not include the object without the key in the resulting object. I would probably lean towards the latter to prevent some potential edge cases introduced by including an extra key in the resulting object, but I'm not too fussed either way if others feel differently.

Chris911:
Ignoring the objets without the given property is a good option. I don't really have a strong opinion on this so maybe wait for more people to comment.

asaf-romano:
Personally, I find indexBy(prop(..)) much more useful.

kedashoe:
It is a nice implementation, but for something going in the core lib what about an procedural solution or using `reduce` so we only need to make one pass over the data?

Chris911:
Yeah this can also be done with reduce using something like: 
```js
R.reduce(function(acc, elem) {
  acc[elem.id] = elem;
  return acc;
}, {}, list)
```

Also works for me.

CrossEye:
I [still have][ob] a minor objection: 

> The trouble is that it makes an assumption about the data that's difficult to document, annoying to test when running, and without an obviously correct fall-back if the assumption fails: the assumption that there is only one item in the list with a given `id`.

This is clearly not a show-stopper, but it's more than enough to give me pause.

  [ob]: https://github.com/ramda/ramda/issues/931#issuecomment-82505534

scott-christopher:
> @asaf-romano: Personally, I find `indexBy(prop(..))` much more useful.

Good point, and it would then fall in line with the convention of Ramda's other `*By` functions.

> @CrossEye: This is clearly not a show-stopper, but it's more than enough to give me pause

Unless I've misunderstood the concern, I don't think this is such an issue if we can agree on a preference for keeping the first or last of any duplicate key values, then document and test accordingly.

I also agree that this could benefit from a lower-level implementation to better control these invariants.

Chris911:
> I also agree that this could benefit from a lower-level implementation to better control these invariants.

I'm happy to re-implement if you can settle on what the expected behaviour should be when the data is not correct. I see 2 options as you mentioned above: ignore or use `"undefined"` key.

Chris911:
> @asaf-romano: Personally, I find indexBy(prop(..)) much more useful.

The only problem I see with using `prop` here is that now `indexBy` expects a function as it's first argument. As far as I can tell `R.prop` is the only function that really makes sense in this context. It is however more in inline with other `*By` functions but in this case I think a string makes more sense.

iofjuupasli:
> As far as I can tell R.prop is the only function that really makes sense in this context.

How about `v => v.firstName + v.lastName` or `R.path([..., ...])`?

Chris911:
Changed the implement to use `_reduce` so it only goes through the list once (cc: @kedashoe, @scott-christopher).

Also changed to the `property` field to now use a function (cc: @iofjuupasli, @asaf-romano)

CrossEye:
@scott-christopher:

> Unless I've misunderstood the concern, I don't think this is such an issue if we can agree on a preference for keeping the first or last of any duplicate key values, then document and test accordingly.

As I said, this is a minor objection.  But it's a real one. A phrase like "if we can agree on a preference" worries me.  I _think_ there is a best choice for this: deciding on the last value.  But it's not obvious, and it's not clear to me if such an API decision is being guided by implementation ideas.  (This method would be simpler than using the first value.)  I've never liked having the implementation guide the API.

What makes this awkward, I think, is the notion that we're converting to an unordered data type, an Object, but depending upon the specific ordering of our input data to determine precisely how this output should be constructed.  There's some sort of mismatch happening.

davidchambers:
> What makes this awkward, I think, is the notion that we're converting to an unordered data type, an Object, but depending upon the specific ordering of our input data to determine precisely how this output should be constructed.

I find a transformation from ordered collection to unordered collection less worrying than a transformation from unordered collection to ordered collection (see `R.keys`, `R.values`, and `R.toPairs`). Yes, we must choose an arbitrary policy for handling collisions, but the common use case—converting something like `[{id: '1', ...}, {id: '2', ...}, ...]` into `{'1': {id: '1', ...}, '2': {id: '2', ...}, ...}`—will be unaffected by this policy.

scott-christopher:
>  I think there is a best choice for this: deciding on the last value.

I'm also happy for the last value to be retained in the case of duplicates.

> A phrase like "if we can agree on a preference" worries me.

"if we can agree on a preference" was intended to indicate that the decision seems largely arbitrary and not such an issue in my opinion as long as it is documented and tested.

I would imagine the common use for `indexBy` would not be to use it with lists containing duplicate index keys, however if they do and for some reason want the first one to be retained instead of the last then they still have the ability to reverse the list prior to calling `indexBy`.

src/indexBy.js
```diff
@@ -0,0 +1,30 @@
...
+ * @memberOf R
+ * @category List
+ * @sig (a -> String) -> [{k: v}] -> {k: {k: v}}
+ * @param {Function} fn Function :: a -> String The indexing property.
+ * @param {Array} array The array of objects to index
+ * @return {Object} An object indexing each array element by the given property.
+ * @example
+ *
+ *      var list = [{id: 'xyz', title: 'A'}, {id: 'abc', title: 'B'}];
+ *      R.indexBy('id', list);
```
Chris911: Update example.

src/indexBy.js
```diff
@@ -0,0 +1,30 @@
+var _curry2 = require('./internal/_curry2');
+var _reduce = require('./internal/_reduce');
+
+
+/**
+ * Given a property name, turns a list of objects into an object indexing
```
Chris911: Update description.

asaf-romano:
I have found myself using my local implementation of indexBy with both R.path and R.pipe(R.prop, somethingBetterThanToString)

Chris911:
Started this PR because I looked at our code base and have 3 different implementations laying around. I think the current one that takes a function to generate the id is the best solution and should be included in the core.

gilligan:
I'n confused by the example
```R.indexBy('id', list);
```
The first parameter is a string but it should be a function instead I think?

Chris911:
@gilligan yes I forgot to update the doc. I'll do it right now.

CrossEye:
> I find a transformation from ordered collection to unordered collection less worrying than a transformation from unordered collection to ordered collection 

I think most everyone here knows how I feel about _that_!  :wink:

And as I said, I don't have any real objection.  I'm going to merge this now.  But we shouldn't let decisions like this pass as though they aren't decisions.  There is something slightly troubling here, and we should at least recognize that.

Chris911:
Thanks :+1: 

How often are Ramda packages released?

davidchambers:
> How often are Ramda packages released?

Sporadically. We're overdue for a release. There's one pull request I need to open and merge before we release v0.19.0. I hope to work on it at some point this week. :)

ivan-kleshnin:
Looks like it should be `Array` -> `Map` conversion rather than `Array` -> `Object`
But Ramda does not support `Map`, right?

davidchambers:
> Ramda does not support `Map`, right?

Correct. It could, though. See #1056 for an early attempt.

PR #1547, opened 2015-12-14
Add node 4.2 (LTS) to travis CI

Chris911:
The node 4.2.* branch is LTS. Should be important that Ramda supports these releases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,6 @@ language: node_js
 node_js:
 - '0.10'
 - '0.12'
-- 'iojs-v1.5.1'
+- '4.2'
 before_script:
 - make clean && make dist/ramda.js


.travis.yml
```diff
@@ -3,6 +3,7 @@ language: node_js
 node_js:
 - '0.10'
 - '0.12'
+- '4.2'
 - 'iojs-v1.5.1'
```
davidchambers: Shall we remove this while we're at it? It seems redundant now that the two projects have merged.

CrossEye: :+1:

CrossEye:
:herb:

gilligan:
I would certainly also suggest to get rid of iojs - what about nodejs v5 though ?

Chris911:
Removed iojs build. I can add v5 if you want.

davidchambers:
:deciduous_tree:

PR #1557, opened 2015-12-22
prepublish: add @since annotations

davidchambers:
This is a fragment of the script I wrote for #1434. Adding it to the prepublish script means that each time we run `make release-[major|minor|patch]` we'll add the appropriate `@since` annotation for each function which lacks one.


davidchambers:
It worked as expected: 6141a3040bfec6b2588b80276161b7a03f03380e.

CrossEye:
Nice!

PR #1559, opened 2015-12-22
Remove jshint/jscsrc leftovers

gilligan:
Remove config files and jshint switches in
the sources.


test/bind.js
```diff
@@ -1,4 +1,3 @@
-/* jshint -W053 */
```
davidchambers: :space_invader:

Siilwyn:
What's up with this?

davidchambers:
We should merge this, I would say. :)

CrossEye:
yup

Issue #1563, opened 2015-12-26
Strangle


ccorcos:
A nice easy function for bounding a number's minimum and maximum.

```js
const strangle = curry((min, max, value) => {
  return value < min ? min :
         value > max ? max :
         value
})
```

niloy:
Nice function, but `strangle` is such a weird name. `bound` or `clamp` sounds nice.

ccorcos:
clamp -- I think thats more commonly used...

CrossEye:
: herb:

I can definitely see this being useful.

PR #1564, opened 2015-12-26
Clamp

ccorcos:
https://github.com/ramda/ramda/issues/1563

```js
R.clamp(1, 10, -1) // => 1
R.clamp(1, 10, 11) // => 10
R.clamp(1, 10, 4)  // => 4
```

### DIFF
--- a/src/clamp.js
+++ b/src/clamp.js
@@ -1,0 +1,29 @@
+var _curry3 = require('./internal/_curry3');
+
+/**
+ * Restricts a number to be within a range.
+ *
+ * Also works for other ordered types such as Strings and Dates.
+ *
+ * @func
+ * @memberOf R
+ * @category Relaction
+ * @sig Ord a => a -> a -> a -> a
+ * @param {Number} minimum number
+ * @param {Number} maximum number
+ * @param {Number} value to be clamped
+ * @return {Number} Returns the clamped value
+ * @example
+ *
+ *      R.clamp(1, 10, -1) // => 1
+ *      R.clamp(1, 10, 11) // => 10
+ *      R.clamp(1, 10, 4)  // => 4
+ */
+module.exports = _curry3(function clamp(min, max, value) {
+  if (min > max) {
+    throw new Error('min must not be greater than max in clamp(min, max, value)');
+  }
+  return value < min ? min :
+         value > max ? max :
+         value;
+});
--- a/test/clamp.js
+++ b/test/clamp.js
@@ -1,0 +1,29 @@
+var eq = require('./shared/eq');
+var R = require('..');
+
+describe('clamp', function() {
+  it('clamps to the lower bound', function() {
+    eq(R.clamp(1, 10, 0), 1);
+    eq(R.clamp(3, 12, 1), 3);
+    eq(R.clamp(-15, 3, -100), -15);
+  });
+
+  it('clamps to the upper bound', function() {
+    eq(R.clamp(1, 10, 20), 10);
+    eq(R.clamp(3, 12, 23), 12);
+    eq(R.clamp(-15, 3, 16), 3);
+  });
+
+  it('leaves it alone when within the bound', function() {
+    eq(R.clamp(1, 10, 4), 4);
+    eq(R.clamp(3, 12, 6), 6);
+    eq(R.clamp(-15, 3, 0), 0);
+  });
+
+  it('works with letters as well', function() {
+    eq(R.clamp('d', 'n', 'f'), 'f');
+    eq(R.clamp('d', 'n', 'a'), 'd');
+    eq(R.clamp('d', 'n', 'q'), 'n');
+  });
+
+});


davidchambers:
Should we throw if `min > max`?

ccorcos:
good idea

ccorcos:
done

src/clamp.js
```diff
@@ -0,0 +1,25 @@
...
+ * @param {Number} value to be clamped
+ * @return {Number} Returns the clamped value
+ * @example
+ *
+ *      R.clamp(1, 10, -1) // => 1
+ *      R.clamp(1, 10, 11) // => 10
+ *      R.clamp(1, 10, 4)  // => 4
+ */
+module.exports = _curry3(function clamp(min, max, value) {
+  if (min > max) throw new Error('min must be greater than max in clamp(min, max, value)')
```
davidchambers: This should read _min must **not** be greater than max_. ;)

ccorcos: done

ccorcos:
hmm. not sure whats going on here...

src/clamp.js
```diff
@@ -0,0 +1,25 @@
...
+ * @param {Number} value to be clamped
+ * @return {Number} Returns the clamped value
+ * @example
+ *
+ *      R.clamp(1, 10, -1) // => 1
+ *      R.clamp(1, 10, 11) // => 10
+ *      R.clamp(1, 10, 4)  // => 4
+ */
+module.exports = _curry3(function clamp(min, max, value) {
+  if (min > max) throw new Error('min must not be greater than max in clamp(min, max, value)')
```
davidchambers: In order to appease the linter, you'll need to write this like so:

```javascript
if (...) {
  ...
}
```

src/clamp.js
```diff
@@ -0,0 +1,25 @@
...
+  return value < min ? min :
+         value > max ? max :
+         value;
```
davidchambers: Here's another implementation:

```javascript
return Math.max(min, Math.min(max, value));
```

I prefer your implementation, though. ;)

src/clamp.js
ccorcos: I see. done

src/clamp.js
ccorcos: Yeah, I agree. Its a little easier to understand.

src/clamp.js
```diff
@@ -0,0 +1,27 @@
+var _curry3 = require('./internal/_curry3');
+
+/**
+ * Clamps a number within a range.
+ *
+ * @func
+ * @memberOf R
+ * @category Relaction
+ * @sig Ord a => a -> a -> a -> a
```
davidchambers: We should have a test case for non-number ordinal values. For example:

```javascript
eq(R.clamp('a', 'f', 'z'), 'f');
```

test/clamp.js
```diff
@@ -0,0 +1,24 @@
+
```
davidchambers: :space_invader:

test/clamp.js
```diff
@@ -0,0 +1,24 @@
...
+var assert = require('assert');
```
davidchambers: Let's replace this with `var eq = require('./shared/eq');`, to allow us to use `eq` in place of `assert.strictEqual`.

src/clamp.js
ccorcos: good idea

test/clamp.js
ccorcos: sounds good -- I just copied it from some other file.

ccorcos:
ok @davidchambers, added some tests for letters as well. I'm not sure what you want to do about `./shared/eq` though since it doesnt exist. I could import `../src/equals` instead.

davidchambers:
@ccorcos, see [__test/shared/eq.js__][1]. ;)


[1]: https://github.com/ramda/ramda/blob/v0.19.0/test/shared/eq.js

ccorcos:
Hmm. I'm confused. Did I fork off of 0.14?! https://github.com/ccorcos/ramda/blob/master/package.json#L27

I'll have to rebase when I get home.

davidchambers:
You'll need to update your fork's __master__ branch:

```console
$ git checkout master
$ git pull upstream master
$ git push origin master
```

ccorcos:
awesome. done.

test/clamp.js
davidchambers: Let's delete this line. :)

src/clamp.js
```diff
@@ -0,0 +1,27 @@
+var _curry3 = require('./internal/_curry3');
+
+/**
+ * Clamps a number within a range.
```
davidchambers: Shall we mention that this function applies to any type which supports ordering?

ccorcos: Sure: how about this?

> Clamps any type that supports ordering within a range.

davidchambers: In light of #1578 we should probably devote one sentence to the common case and one sentence to the abstraction. What do you think, @CrossEye?

CrossEye: Yes, I agree.

    Restricts a number to be within a range.

    Also works for other ordered types such as Strings and Dates.  (see Type::Ord)

That last bit it just a bit of fantasy, but something I'd love to be able to support.

ccorcos: done

src/clamp.js
```diff
@@ -0,0 +1,29 @@
+var _curry3 = require('./internal/_curry3');
+
+/**
+ * Restricts a number to be within a range.
+ *
+ * Also works for other ordered types such as Strings and Dates.  (see Type::Ord)
```
davidchambers: I don't think @CrossEye was suggesting to actually include the parenthesized text. I believe he was saying it would be nice if we could do something like this and have it link to an explanation of Ord.

ccorcos: Oh. I thought maybe it would :)

ccorcos: How about this:

> Restricts a number to be within a range. Also works for other ordered types such as Strings and Dates.

CrossEye: Looks good to me.  Sorry about the confusion earlier.

ccorcos: No worries :)

ccorcos:
Alright, so what do you guys think? My first Ramda contribution? :)

davidchambers:
Looks good, aside from the two outstanding issues:

  - `(see Type::Ord)` should be removed; and
  - the empty line at the top of __test/clamp.js__ should be removed.

Once you've addressed these minor issues I'll happily merge this pull request. :)

CrossEye:
:herb:, once those minor changes are made.

ccorcos:
Boom

CrossEye:
Thank you very much!

davidchambers:
:deciduous_tree: Thanks for persevering. :)

ccorcos:
Of course :+1: I look forward to contributing more now that I got the first one out of the way.

xgrommx:
About `clamp` maybe it will be more general? 
```js
const clamp = curry((min, max, value) => Math.max(min, Math.min(max, value)));
```
 Also if you add `clamp` why not `lerp`? 
```js
const lerp = curry((start, end, alpha) => start*(1-alpha)+end*alpha
``` 
It is linear interpolation.

src/clamp.js
xgrommx: Maybe we need transform chars to ascii code and compare by code.

CrossEye:
@xgrommx:

> About `clamp` maybe it will be more general? 

Perhaps I'm misunderstanding.  The current implementation handles that case, but should work with any ordered type.

>  Also if you add `clamp` why not `lerp`? 

While that is certainly a useful function, my first thought is that it belongs in [The Cookbook][cb] instead.  I certainly would not object to hearing arguments to the contrary though if you disagree.


src/clamp.js
CrossEye: We have (working) tests already.  Because we use `a < b` it will work with Numbers, Strings, Dates, or any type that has a proper `valueOf` method.

PR #1112, opened 2015-05-26
remove functions deprecated in v0.14.0

davidchambers:
I'm sad to see `R.isNaN` go. :cry:

PR #1565, opened 2015-12-28
Remove R.isNaN leftover in documentation

lo1tuma:
The examples for `R.none` are still using `R.isNaN` which was removed in #1112. I’ve replace the examples with actual use-cases from the tests.


CrossEye:
Thank you.

PR #1568, opened 2015-12-30
`range` returns numbers in descending order if `from` is greater than `to`

niloy:
```js
R.range(5, 1);    //=> [5, 4, 3, 2]
```

Clojure's [range](http://clojuredocs.org/clojure.core/range) works similar.

### DIFF
--- a/src/range.js
+++ b/src/range.js
@@ -11,12 +11,16 @@ var _isNumber = require('./internal/_isNumber');
  * @category List
  * @sig Number -> Number -> [Number]
  * @param {Number} from The first number in the list.
- * @param {Number} to One more than the last number in the list.
- * @return {Array} The list of numbers in tthe set `[a, b)`.
+ * @param {Number} to The exclusive bounds of the sequence. The last number of
+ *                    the returned list will be one less than or more than the
+ *                    `to` parameter for incrementing and decrementing ranges
+ *                    respectively.
+ * @return {Array} The list of numbers in the set `[a, b)`.
  * @example
  *
  *      R.range(1, 5);    //=> [1, 2, 3, 4]
  *      R.range(50, 53);  //=> [50, 51, 52]
+ *      R.range(5, 1);    //=> [5, 4, 3, 2]
  */
 module.exports = _curry2(function range(from, to) {
   if (!(_isNumber(from) && _isNumber(to))) {
@@ -24,9 +28,10 @@ module.exports = _curry2(function range(from, to) {
   }
   var result = [];
   var n = from;
-  while (n < to) {
+  var dx = from < to ? 1 : -1;
+  while (n !== to) {
     result.push(n);
-    n += 1;
+    n += dx;
   }
   return result;
 });
--- a/test/range.js
+++ b/test/range.js
@@ -11,21 +11,17 @@ describe('range', function() {
     eq(R.range(4, 7), [4, 5, 6]);
   });
 
-  it('returns the empty list if the first parameter is not larger than the second', function() {
-    eq(R.range(7, 3), []);
+  it('returns numbers in descending order if from is greater than to', function() {
+    eq(R.range(5, 1), [5, 4, 3, 2]);
+  });
+
+  it('returns the empty list if both parameters are equal', function() {
     eq(R.range(5, 5), []);
   });
 
   it('is curried', function() {
     var from10 = R.range(10);
     eq(from10(15), [10, 11, 12, 13, 14]);
-  });
-
-  it('returns an empty array if from > to', function() {
-    var result = R.range(10, 5);
-    eq(result, []);
-    result.push(5);
-    eq(R.range(10, 5), []);
   });
 
   it('terminates given bad input', function() {


src/range.js
```diff
@@ -12,11 +12,12 @@ var _isNumber = require('./internal/_isNumber');
  * @sig Number -> Number -> [Number]
  * @param {Number} from The first number in the list.
  * @param {Number} to One more than the last number in the list.
```
scott-christopher: It would probably be worth updating the description for the `to` param to recognise that it now supports both directions.

Perhaps something like:
```
The exclusive bounds of the sequence. The last number of the returned list will be one less than or more than the `to` parameter for incrementing and decrementing ranges respectively.
```

niloy:
@scott-christopher Done

scott-christopher:
This looks good to me. :mushroom:

buzzdecafe:
http://fr.umio.us/ranging-near-and-far

niloy:
Nice read :+1:

davidchambers:
To create `[5, 4, 3, 2]`, I'd prefer to require `R.reverse(R.range(2, 6))` than allow `R.range(5, 1)`. Assuming one knows that `R.range(0, 3)` evaluates to `[0, 1, 2]`, one knows the value to which `R.range(2, 6)` will evaluate and the reverse of that value.

There are three values to which `R.range(5, 1)` could evaluate: `[5, 4, 3, 2]`, `[4, 3, 2, 1]`, and `[]`.

I don't think `R.range(0, n)` should create an incrementing range in some cases and a decrementing range in others. Reversing a range as a separate step seems fine to me, but if this is unacceptable for some reason I'd prefer to define `R.rangeDec` for creating decrementing ranges.

scott-christopher:
I don't have such an issue with supporting both incrementing and decrementing sequences in `range`, however it seems like I'm in the minority.

Perhaps we could look at including something similar to Haskells `enumFromThenTo` to support this instead; something like:
```js
//:: Int -- from: Inclusive start of sequence
//-> Int -- then: Next step in sequence
//-> Int -- to:   Inclusive end of sequence
//-> [Int]
fromThenTo = (from, then, to) => {
  const step = then - from;
  const inBounds = step > 0 ? R.lte : R.gte;
  const result = [from];
  if (from === then) {
    return result;
  }
  if ((from < then && to < from) || (from > then && to > from)) {
    return [];
  }
  for (let next = then; inBounds(next, to); next = next + step) {
    result.push(next);
  }
  return result;
};

fromThenTo(1, 2, 10); // [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
fromThenTo(2, 4, 10); // [2, 4, 6, 8, 10]
fromThenTo(1, 3, 10); // [1, 3, 5, 7, 9]
fromThenTo(2, 2, 10); // [2]
fromThenTo(2, 1, 10); // []
```

That would leave `range` to operate as a simple incrementing function to use where the extra complexity isn't needed.

davidchambers:
I like the idea of defining `R.fromThenTo` and leaving `R.range` as is. :)

CrossEye:
@niloy: Are you finding a common need for reversed ranges that make doing a `reverse` after a range impractical?  Or does this just seem like a nice-to-have?

@scott-christopher: Same question.  How often do you need to create [24, 21, 18, 15, 12, 9], and is it common enough to build in a function for it rather than to just use `R.map(n = > 24 - 3 * n, R.range(0, 6))`?  The reason for the article @buzzdecafe cited was that such flexibility seemed to run strongly counter to the YAGNI philosophy of Ramda.

niloy:
I don't have a common need, its a nice-to-have. Recently, I needed to create a year dropdown with values 2015, 2014.... I did it using `R.range(0, 10).map(R.subtract(2015))` and thought it would have been cooler if I could write `R.range(2015, 2005)`. But I can live without it.

scott-christopher:
@CrossEye: TBH, if I needed `[24, 21, 18, 15, 12, 9]` and a stepping sequence function wasn't available, I'd probably just use `[24, 21, 18, 15, 12, 9]` rather than leaving `R.map(n => 24 - 3 * n, R.range(0, 6))` for the next person to find.

I rarely use `R.range` at all so I'm not such a good candidate to say whether this would be useful to me or not, but it was deemed useful enough for it to be included in many other language's stdlibs in this space.

I'll let others decide the fate of this one :)

CrossEye:
@niloy: I would not have a big concern with having reversed ranges as well as forward ranges.  It does not seem to a great deal of significant conceptual complexity to the function.  This concerns me less than adding `steps`.  But I would only want to do it if there is a real need, as it clearly does add some.

So if you want to argue for it, please feel free.  I might argue back.  Or I might not.  :smile:

@scott-christopher:  I do use `range`, but almost exclusively as `range(0, n)` or `range(1, n+ 1)`.  Of course my example was easy to hard-code, but the point should be clear too for the equivalent of `R.map(n => 2400 - 3 * n, R.range(0, 687))`, which you we probably wouldn't want to hard-code.  But let's not worry about `enumFromThenTo` unless there's a real demand for it.

scott-christopher:
> @CrossEye: But let's not worry about enumFromThenTo unless there's a real demand for it.

I've added a `rangeStep` function to the cookbook for anyone who may be interested in the future.

CrossEye:
> I've added a `rangeStep` function to the cookbook for anyone who may be interested in the future.

Perfect.  Perhaps @raine can tell us how to update the TOC.

raine:
I use a script where the critical part is:

```sh
doctoc --title '## Recipes' Cookbook.md
```

CrossEye:
Do you think we could add it to `package.json/scripts`?

raine:
I added it to the `ramda.wiki.git` repo. Only owners have push access to wiki repos though.

CrossEye:
oh, ok.  Didn't even know we had that repo.  :blush:

raine:
Yeah, every wiki on Github has its own repo which basically consists of markdown files.

scott-christopher:
> Yeah, every wiki on Github has its own repo which basically consists of markdown files.

TIL

niloy:
@CrossEye Yeh, reversed ranges are required every once in a while, since its implementation is trivial, it would be cool if Ramda supported it out-of-the box :smiley: .

src/range.js
```diff
@@ -24,9 +28,10 @@ module.exports = _curry2(function range(from, to) {
   }
   var result = [];
   var n = from;
-  while (n < to) {
+  var dx = from < to ? 1 : -1;
+  while (n !== to) {
```
CrossEye: As discussed on Gitter, this would fail if `to` is not an integer (or more precisely if `to - from` when cast to a number is not an integer... gotta love JS.)

CrossEye:
We discussed this on Gitter.

Although I'm happy to hear an updated PR, I'm still not convinced that the increased complexity of the API is worth it, when `compose(reverse, range)` serves perfectly well.  I'd still like to hear the arguments for the other side of it.  Mine are mostly made at http://fr.umio.us/ranging-near-and-far.

jackfirth:
Just a note: if range allowed descending sequences, I'd expect the law `range(a, b) == reverse(range(b, a))` to hold. Or more accurately, I'd assume it holds knowing that's a mistake and trip myself up somewhere.

CrossEye:
@jackfirth:

That's a good point.  While that would be a useful law, it's also likely that we'd want to assert that `range` is inclusive in the first parameter and exclusive in the second.  Obviously these conflict.

buzzdecafe:
i still prefer keeping `range` as is and composing with `reverse` when desired. Current `range` implementation is *simple*; let's keep it that way.

davidchambers:
> i still prefer keeping `range` as is and composing with `reverse` when desired. Current `range` implementation is *simple*; let's keep it that way.

:+1:

niloy:
:+1:

PR #1571, opened 2015-12-31
Set arity and curry juxted functions

### LINKED ISSUE #1569: juxt reports wrong arity
```js
R.juxt([R.prop("a"), R.path(["a", "b"])]).length; // => 0
```

asaf-romano:
Closes #1569

### DIFF
--- a/src/converge.js
+++ b/src/converge.js
@@ -33,7 +33,8 @@ var pluck = require('./pluck');
  *      R.converge(add3, [multiply, add, subtract])(1, 2); //=> 4
  */
 module.exports = _curry2(function converge(after, fns) {
-  return curryN(Math.max.apply(Math, pluck('length', fns)), function() {
+  var arity = fns.length ? Math.max.apply(Math, pluck('length', fns)) : 0;
+  return curryN(arity, function() {
     var args = arguments;
     var context = this;
     return after.apply(context, _map(function(fn) {
--- a/test/converge.js
+++ b/test/converge.js
@@ -41,4 +41,10 @@ describe('converge', function() {
     eq(f3(R.__).length, 3);
   });
 
+  it('works with empty functions list', function() {
+    var fn = R.converge(function() { return arguments.length; }, []);
+    eq(fn.length, 0);
+    eq(fn(), 0);
+  });
+
 });
--- a/src/internal/_arrayOf.js
+++ b/src/internal/_arrayOf.js
@@ -1,0 +1,3 @@
+module.exports = function _arrayOf() {
+  return Array.prototype.slice.call(arguments);
+};
--- a/src/juxt.js
+++ b/src/juxt.js
@@ -1,7 +1,6 @@
-var __ = require('./__');
+var _arrayOf = require('./internal/_arrayOf');
 var _curry1 = require('./internal/_curry1');
-var apply = require('./apply');
-var map = require('./map');
+var converge = require('./converge');
 
 
 /**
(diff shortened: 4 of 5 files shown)


asaf-romano:
This is somewhat subjective, but the more I thought about it the more it made sense, to me, for these two functions to work on empty list of functions even though the use case is questionable.

davidchambers:
Please rebase to remove the merge commits. :)

src/converge.js
```diff
@@ -33,7 +33,8 @@ var pluck = require('./pluck');
  *      R.converge(add3, [multiply, add, subtract])(1, 2); //=> 4
  */
 module.exports = _curry2(function converge(after, fns) {
-  return curryN(Math.max.apply(Math, pluck('length', fns)), function() {
+  var arity = fns.length ? Math.max.apply(Math, pluck('length', fns)) : 0;
```
davidchambers: What is the purpose of the `fns.length` check?

asaf-romano: Math.max returns -Infinity for zero arguments, which is why converge didn't work for an empty functions list,

asaf-romano:
@davidchambers I had a hard time trying to get rid of those zombie merge commit (couldn't squash them, at least). I'll probably just generate the branch from scratch once the review is done (well, unless you know a particular set of magic commands :) ).

davidchambers:
> unless you know a particular set of magic commands

I have a relatively straightforward solution:

```console
$ git checkout juxt-converge
$ git reset --hard $(git merge-base master juxt-converge)  # reset to common ancestor
$ curl --location https://github.com/ramda/ramda/pull/1571.diff | patch -p1  # apply patch
$ git add src test
$ git commit -m '...'
$ git push origin juxt-converge --force  # update pull request
```

While you're at it, I suggest:

```console
$ git checkout master
$ git pull upstream master  # update local master branch (assumes "upstream" is an alias for "git@github.com:ramda/ramda.git")
$ git checkout juxt-converge
$ git rebase master  # incorporate upstream changes into feature branch
$ git push origin juxt-converge --force
```

By following the steps above you'll avoid ever introducing a merge commit on a feature branch. :)

CrossEye:
David _always_ seems to know the right set of magic commands!  :smile:

test/converge.js
```diff
@@ -41,4 +41,8 @@ describe('converge', function() {
     eq(f3(R.__).length, 3);
   });
 
+  it('works with empty functions list', function() {
+    eq(R.converge(R.always(0), [])(), 0);
+  });
+
```
CrossEye: I wonder if we want to be a little more explicit here:

```js
var fn = R.converge(function() {return arguments.lengths;}, []);
eq(fn.length, 0);
eq(fn(), 0);
```

CrossEye:
Overall, this looks great!

asaf-romano:
Many many thanks David!

davidchambers:
In your commit message and in this pull request's title you have "artity" when I believe you mean "arity". Also, I think "juxed" is missing a "t". :)

src/converge.js
davidchambers: I'd like to use `reduce(..., 0, fns)` here to avoid a special case.

asaf-romano:
artity, yes, oops. but juxed was already fixed :)

davidchambers:
@asaf-romano, would you mind updating this pull request's title to match the updated commit message?

src/converge.js
davidchambers: Thoughts on using `reduce` here, @asaf-romano? We do so in #1572.

CrossEye:
Thanks for your patience.

src/converge.js
davidchambers: #1587

PR #1572, opened 2015-12-31
Retain arity in cond

### LINKED ISSUE #1570: cond reports wrong arity
```js
R.cond([[R.propEq("a", "b")]]).length; // 0
```

asaf-romano:
Closes #1570

### DIFF
--- a/src/cond.js
+++ b/src/cond.js
@@ -1,4 +1,8 @@
+var _arity = require('./internal/_arity');
 var _curry1 = require('./internal/_curry1');
+var map = require('./map');
+var max = require('./max');
+var reduce = require('./reduce');
 
 
 /**
@@ -28,7 +32,10 @@ var _curry1 = require('./internal/_curry1');
  *      fn(100); //=> 'water boils at 100°C'
  */
 module.exports = _curry1(function cond(pairs) {
-  return function() {
+  var arity = reduce(max,
+                     0,
+                     map(function(pair) { return pair[0].length; }, pairs));
+  return _arity(arity, function() {
     var idx = 0;
     while (idx < pairs.length) {
       if (pairs[idx][0].apply(this, arguments)) {
@@ -36,5 +43,5 @@ module.exports = _curry1(function cond(pairs) {
       }
       idx += 1;
     }
-  };
+  });
 });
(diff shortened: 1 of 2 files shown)


src/cond.js
```diff
@@ -28,7 +30,9 @@ var _curry1 = require('./internal/_curry1');
  *      fn(100); //=> 'water boils at 100°C'
  */
 module.exports = _curry1(function cond(pairs) {
-  return function() {
+  var arity = pairs.length ?
+    Math.max.apply(Math, map(function(pair) { return pair[0].length; }, pairs)) : 0;
```
davidchambers: We could express this as a fold to obviate the need for the `pairs.length` check.

asaf-romano: ``reduce(function(accum, pair) { return Math.max(accum, pair[0].length); }, 0, pairs)``?

davidchambers: Yes, or this:

```javascript
reduce(max,
       0,
       map(function(pair) { return pair[0].length; }, pairs))
```

asaf-romano:
Heh, one of these days I'll figure out to what extent public api could/should be used. Because outside of ramda I'd go ahead and also do ``pipe(head, length)``.

src/cond.js
asaf-romano: Heh, one of these days I'll figure out to what extent public api could/should be used. Because outside of ramda I'd go ahead and also do pipe(head, length).

davidchambers: I copied the lambda from your code above. ;)

asaf-romano: I was referring to the use of both map and reduce, which implies two passes over the same array. I'm fine with that, of course, but as I understand things, the code in src/ looks the way it does because it tries to be as efficient as possible.

CrossEye: >  [A]s I understand things, the code in src/ looks the way it does because it tries to be as efficient as possible.

Yes, but it's a balancing act.  We don't have it all figured out, and every new commit changes the balance a bit.

asaf-romano: I'll go with David's suggested (iterate twice) because in-practice this code never runs (successfully) for arity > 9.

davidchambers:
:deciduous_tree:

CrossEye:
Thank you.  A nice fix!

Issue #1560, opened 2015-12-22
Merge breaking change


davidebianchi:
I noticed a (undocumented) breaking change in the `R.merge` function between `v0.18.0` and `v0.19.0`. 

Reproduction:

``` js
R.merge({a: 5}, undefined)
```
- [Behaviour in `v0.18.0`](http://bit.ly/1YyX7la)
- [Behaviour in `v0.19.0`](http://bit.ly/1YyWJTU)

As soon as I have a bit of time I'll try to figure out the cause of the regression and prepare a PR for it. For now, I'll just open the issue. :smiley:

davidchambers:
Though this is a behaviour change, I don't consider it a regression. `R.merge` was not and is not documented to accept `null` or `undefined`: it operates on two objects. The previous implementation *happened* to treat `undefined` as equivalent to `{}` due to its use of `keys`, which for some reason does not throw when applied to `undefined`.

This is a situation where [sanctuary-def][1]-style type checking—discussed in #1125—would have prevented a "regression" by not allowing the invalid use in the first place.


[1]: https://github.com/plaid/sanctuary-def

davidebianchi:
Hi @davidchambers,

good point. Without type-checking it is indeed difficult to know/predict whether or not the function will be called correctly (unfortunately `undefined` is often slip through).

Right now though `R.merge` has an [inconsistent behaviour when it comes to non objects](http://bit.ly/1MuyTS8), which makes using it fairly confusing.

I would suggest either adding a runtime type check - like in the case of `R.test` - or defaulting the input parameters to the empty object in case they are not objects to begin with, sort of how `R.keys` defaults to the empty array if its input is not an object.

davidchambers:
I'll let @CrossEye respond to this. My position on run-time type checking is well documented. ;)

CrossEye:
Sorry, just catching up on some old issues.

While I agree that this is not technically a regression, as we didn't document any behavior for non-objects, I believe the 0.18 behavior is what we would prefer.  I especially like this:

> defaulting the input parameters to the empty object in case they are not objects to begin with, sort of how `R.keys` defaults to the empty array if its input is not an object.

davidchambers:
> I especially like this:
>
> > defaulting the input parameters to the empty object in case they are not objects to begin with, sort of how `R.keys` defaults to the empty array if its input is not an object.

I'm not fond of ad hoc polymorphism. `R.keys(null)` is a type error, as far as I'm concerned. ;)

[...]

PR #1573, opened 2016-01-03
ad hoc type checking for R.add and R.subtract

davidchambers:
See https://github.com/ramda/ramda/issues/1560#issuecomment-168455854

Current behaviour:

```javascript
R.add(1, 'XXX');
// => '1XXX'

R.subtract(1, 'XXX');
// => NaN

R.inc('XXX');
// => '1XXX'

R.dec('XXX');
// => '-1XXX'
```

New behaviour:

```javascript
R.add(1, 'XXX');
// ! TypeError: ‘add’ expected a value of type Number as its second argument; received "XXX"

R.subtract(1, 'XXX');
// ! TypeError: ‘subtract’ expected a value of type Number as its second argument; received "XXX"

R.inc('XXX');
// ! TypeError: ‘add’ expected a value of type Number as its second argument; received "XXX"

R.dec('XXX');
// ! TypeError: ‘add’ expected a value of type Number as its second argument; received "XXX"
```

I far prefer the new behaviour. I'd like `R.inc` and `R.dec` to throw more appropriate errors, but that change could easily be made in a subsequent pull request.

### DIFF
--- a/src/add.js
+++ b/src/add.js
@@ -1,4 +1,6 @@
 var _curry2 = require('./internal/_curry2');
+var _isNumber = require('./internal/_isNumber');
+var toString = require('./toString');
 
 
 /**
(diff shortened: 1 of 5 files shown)


CrossEye:
Why does this not surprise me?  :smile: 

:herb:

PR #1574, opened 2016-01-03
package.json: remove reference to defunct scripts/testfiles

davidchambers:
We removed __scripts/testfiles__ in #1517, but forgot to delete this reference.

/cc @scott-christopher


scott-christopher:
:mushroom:

PR #1575, opened 2016-01-03
package.json: update "homepage"

davidchambers:
If we'd prefer to stick with the GitHub URL, I insist we remove the subdomain. :stuck_out_tongue_winking_eye:


CrossEye:
Absolutely.

PR #1579, opened 2016-01-05
ad hoc type checking for R.multiply and R.divide

davidchambers:
Current behaviour:

```javascript
R.multiply(1, 'XXX');
// => NaN

R.divide(1, 'XXX');
// => NaN
```

New behaviour:

```javascript
R.multiply(1, 'XXX');
// ! TypeError: ‘multiply’ expected a value of type Number as its second argument; received "XXX"

R.divide(1, 'XXX');
// ! TypeError: ‘divide’ expected a value of type Number as its second argument; received "XXX"
```

### DIFF
--- a/src/divide.js
+++ b/src/divide.js
@@ -1,4 +1,6 @@
 var _curry2 = require('./internal/_curry2');
+var _isNumber = require('./internal/_isNumber');
+var toString = require('./toString');
 
 
 /**
@@ -23,4 +25,14 @@ var _curry2 = require('./internal/_curry2');
  *      var reciprocal = R.divide(1);
  *      reciprocal(4);   //=> 0.25
  */
-module.exports = _curry2(function divide(a, b) { return a / b; });
+module.exports = _curry2(function divide(a, b) {
+  if (!_isNumber(a)) {
+    throw new TypeError('‘divide’ expected a value of type Number ' +
+                        'as its first argument; received ' + toString(a));
+  }
+  if (!_isNumber(b)) {
+    throw new TypeError('‘divide’ expected a value of type Number ' +
+                        'as its second argument; received ' + toString(b));
+  }
+  return a / b;
+});
--- a/src/multiply.js
+++ b/src/multiply.js
@@ -1,4 +1,6 @@
 var _curry2 = require('./internal/_curry2');
+var _isNumber = require('./internal/_isNumber');
+var toString = require('./toString');
 
 
 /**
@@ -21,4 +23,14 @@ var _curry2 = require('./internal/_curry2');
  *      triple(4);       //=> 12
  *      R.multiply(2, 5);  //=> 10
  */
-module.exports = _curry2(function multiply(a, b) { return a * b; });
+module.exports = _curry2(function multiply(a, b) {
+  if (!_isNumber(a)) {
+    throw new TypeError('‘multiply’ expected a value of type Number ' +
+                        'as its first argument; received ' + toString(a));
+  }
+  if (!_isNumber(b)) {
+    throw new TypeError('‘multiply’ expected a value of type Number ' +
+                        'as its second argument; received ' + toString(b));
+  }
+  return a * b;
+});
--- a/test/divide.js
+++ b/test/divide.js
@@ -1,3 +1,5 @@
+var assert = require('assert');
+
 var R = require('..');
 var eq = require('./shared/eq');
 
@@ -5,6 +7,25 @@ var eq = require('./shared/eq');
 describe('divide', function() {
   it('divides two numbers', function() {
     eq(R.divide(28, 7), 4);
+  });
+
+  it('type checks its arguments', function() {
+    assert.throws(
+      function() { R.divide('1', '2'); },
+      function(err) {
+        return err.constructor === TypeError &&
+               err.message === '‘divide’ expected a value of type Number ' +
+                               'as its first argument; received "1"';
+      }
+    );
+    assert.throws(
+      function() { R.divide(1, '2'); },
+      function(err) {
+        return err.constructor === TypeError &&
+               err.message === '‘divide’ expected a value of type Number ' +
+                               'as its second argument; received "2"';
+      }
+    );
   });
 
   it('is curried', function() {
--- a/test/multiply.js
+++ b/test/multiply.js
@@ -1,3 +1,5 @@
+var assert = require('assert');
+
 var R = require('..');
 var eq = require('./shared/eq');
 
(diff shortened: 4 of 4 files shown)


CrossEye:
I'm not yet convinced.

Why shouldn't `1 * 'XXX'` result in `NaN`?  If you try to multiply a number by a non-number, that seems quite _a propos_.  And a throwing an exception again has the problems with composability that always bother me.  It strikes me as very much a last resort.

In #1573, we had a different situation, where our implementation was reporting an inappropriate value.  Here this starts to stray into the territory of Sanctuary.  As far as I'm concerned, Ramda isn't in the business of making Javascript into a more type-safe language.  If the user wants to use it in with her more type-safe code, it should work just fine, but trying to impose that much additional discipline on her is the province of a higher-level library, or even better, of a language that has better type tools built in.

davidchambers:
If we want `R.multiply('foo', 'bar')` to evaluate to `NaN`, shouldn't `R.add('foo', 'bar')` evaluate to `NaN` also? It seems to me the behaviour should at least be consistent.

niloy:
I agree with @CrossEye . 

JS allows multiplication between non-numeric types if the object defines the `valueOf` method. For example, the following code works just fine in `Chrome 46.0.2490.80 (64-bit)`.

```js
var obj = {valueOf: () => 2};
obj * 2 // 4
```

davidchambers:
@niloy, your argument applies equally to addition and subtraction.

niloy:
@davidchambers Which begs the question, why did you choose `multiply` and `divide`. There are better candidates. In `multiply` and `divide`, the input can come from user, the user is free to input non-numeric data.

Consider a scenario where a list of numbers is being pulled from server and I need to map over them to generate their doubles. `getDataFromServer().then(R.map(R.multiply(2)))`. Lets say from whatever reason, the list contains some non-numeric data. If `multiply` throws, mapping if other valid numbers in the list is also halted. I would rather prefer that `map` produces `NaN` for the invalid values rather than a complete exit.

CrossEye:
This just makes me think we were too hasty in #1573.

I just saw the problems with `R.inc` and realized that they were traceable to `R.add` and didn't see any better solution than throwing.  But the point from @niloy makes me realize that there's more going on.

buzzdecafe:
makes me wonder if a function enabling composition with exceptions could be useful. 

Consider: `tryWith :: (x -> a) -> (e -> a) -> a` where the first function is evaluated and returns `a` unless it throws; in which case the second function is evaluated and returns `a`. 

Not as elegant as it is in OCaml, but maybe better than nothing. Because this is JS, this would rely on the good behavior of developers to ensure that the return types of the two functions line up.

davidchambers:
> makes me wonder if a function enabling composition with exceptions could be useful.

We have [`encase`][1] and [`encaseEither`][2]. :)


[1]: https://github.com/plaid/sanctuary#encase
[2]: https://github.com/plaid/sanctuary#encaseEither

CrossEye:
> makes me wonder if a function enabling composition with exceptions could be useful. 

Recently [discussed on Gitter][gi].

  [gi]: https://gitter.im/ramda/ramda?at=568959604f069158055cd722

buzzdecafe:
> We have `encase` and `encaseEither`. :)

yes, similar. `encase` guarantees a `'a Maybe`, what i am suggesting would leave the choice of return value up to the user.

svozza:
>Consider a scenario where a list of numbers is being pulled from server and I need to map over them to generate their doubles. getDataFromServer().then(R.map(R.multiply(2))). Lets say from whatever reason, the list contains some non-numeric data. If multiply throws, mapping if other valid numbers in the list is also halted. I would rather prefer that map produces NaN for the invalid values rather than a complete exit.

If you know that your computation requires a list of `Number`s and you can't trust your input, why are you not beginning your pipeline with `R.filter(R.is(Number))`? It's good practice to sanitise your inputs when you have strict constraints like that.

CrossEye:
> If you know that your computation requires a list of `Number`s and you can't trust your input, why are you not beginning your pipeline with `R.filter(R.is(Number))`? It's good practice to sanitise your inputs when you have strict constraints like that.

For the four years this system has been running, the list has always supplied numbers.  When you wrote the code, it was clear and obvious that the generator could only supply numbers.  But a confluence of changes made by the junior developer who was around two years ago and made some changes that seemed innocuous to the code reviewer, a little optimization made by the summer intern, and a update to an external library that a brand new hire implemented as her first project before leaving for greener pastures means that in that unlikely -- but unfortunately possible -- case that a user deletes her age on the final submission screen, there could be an undefined value in what would usually be a list of numbers.  Again, this was not possible when the code was written four years ago, or you would have guarded against it.  Now however, even though this is an entirely optional field, the application crashes on the user, with an obscure message and a very difficult-to-recreate situation.

Not being able to trust the data is very much a sliding scale, and an ever-changing one.

svozza:
I understand what you're saying but my response was to the particular scenario presented there, i.e., for some reason allowing a pipeline to spew out `NaN`s is preferable to a function within it throwing with a descriptive error message. Surely those `NaN`s will just go on to cause an exception in an even more obscure and hard to track down place unless you filter them out, which brings us back to the original point that it's better not to let the situation arise in the first place.

CrossEye:
> it's better not to let the situation arise in the first place

That's certain.  But the question to us as a library is how to handle it if they do.  My take is that throwing an exception should be extremely rare, and only if there is no other reasonable option.  Because a function that throws is not composable, and, in my example above, even though the error would not cause any further issues, we've stopped a customer from completing her purchase on our site because of this hard-to-debug scenario.  Who knows how much business we'll lose before some kind-hearted soul bothers to report this to us?

davidchambers:
I hope your back end knows how to process `NaN`-dollar payments. :stuck_out_tongue_winking_eye:

CrossEye:
Presumably some system somewhere will have to fail, and there is certainly a problem in separating the cause of the failure from where it's articulated.  But I don't think it's our responsibility as a library to interject ourselves and point out the user's already-existing failures.  Bad data passed to the library is not the library's fault, and interjecting our possibly unwanted corrective action just says that the user needs her hand held.  Doing it by throwing an exception means that she can no longer use functional composition to build this part of the application but must wrap everything in some `try...catch` logic.

I would rather let her compose functions as she chooses, and decide where to handle such potential failures on her own.  My hope is that she would correct the data at its source so there's no chance of passing something bad into Ramda, but if not, I would rather pass out a `NaN` and let her decide whether that's a problem on her own.

It's not always possible, of course.  But often we can give a reasonable result even with bad data; when we can, I would prefer that we do, reserving exceptions for when we have no reasonable choice.

To me, `encase` or `tryWith` are mostly Band-Aids on this problem.  They don't solve the basic issue, but allow some chance of composing the system out of pieces in places where its clear an exception is a real possibility.  They're good to have, but I wouldn't want to rely on them except in quite unusual circumstances.

PR #1580, opened 2016-01-05
coerce arguments to R.add and R.subtract

davidchambers:
See #1579

### DIFF
--- a/test/add.js
+++ b/test/add.js
@@ -1,5 +1,3 @@
-var assert = require('assert');
-
 var R = require('..');
 var eq = require('./shared/eq');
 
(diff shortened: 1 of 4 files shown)


CrossEye:
This is still missing the case under discussion from #1579:

```js
R.subtract({valueOf: () => 42}, 2); //=> will throw, but should return 40
```

davidchambers:
```javascript
> Number({valueOf: function() { return 42; }})
42
```

CrossEye:
Weird.  I tested that in some console and it failed...

Or I thought I did.  I just retested in every console I have and it works, so I don't know what I tried.

Sorry about that.

I want to think about this a bit more, but my first (second I guess!) impression is favorable.

davidchambers:
> I want to think about this a bit more, but my first (second I guess!) impression is favorable.

Okay. We should either merge this or merge #1579, and close the other either way.

test/add.js
```diff
@@ -9,23 +7,12 @@ describe('add', function() {
...
-               err.message === '‘add’ expected a value of type Number ' +
-                               'as its second argument; received "2"';
-      }
-    );
+  it('coerces its arguments to numbers', function() {
+    eq(R.add('1', '2'), 3);
+    eq(R.add(1, '2'), 3);
+    eq(R.add(true, false), 1);
+    eq(R.add(null, null), 0);
+    eq(R.add(undefined, undefined), NaN);
```
svozza: Could also add a test for dates here too.

davidchambers: :zap:

niloy:
Just something to consider, not sure if this is relevant:

http://jsperf.com/numeric-boxing

buzzdecafe:
whatever happened to "let it fail"? This is exactly the type of hand-holding the lib was always trying to avoid.

davidchambers:
> whatever happened to "let it fail"? This is exactly the type of hand-holding the lib was always trying to avoid.

The question is whether it's more appropriate for `R.inc('XXX')` to evaluate to `'1XXX'` or `NaN`. To me, the answer is clear, since the former is not a member of the function's supposed return type.

Of course, my strong preference is to throw when the caller breaches the contract, but Ramda is clearly not the library for this approach. So, our only choice is to return a bogus value. In this case the Number type contains a bogus value for such situations: `NaN`.

CrossEye:
:herb:

I've been thinking about this a lot.  And I'm :+1: on this version, and obviously for doing the same for multiply and divide.

The perf numbers from @niloy gave me some pause, but this version reassured me that it's probably an artifact of the test: http://jsperf.com/numeric-boxing/3.  Things are probably being inlined somewhere in the original.

I am still in favor of a let-it-fail philosophy.  But when we can coerce the output to the right type, or in this case, when we can work with the grain of JS and use its automatic conversions to avoid returning something of the wrong type, I can't see anything better.  Certainly `NaN` is a better response than `'1XXX'`.

PR #1581, opened 2016-01-05
tryCatch for composing with exceptions

buzzdecafe:
see #1579

### DIFF
--- a/src/tryCatch.js
+++ b/src/tryCatch.js
@@ -1,0 +1,34 @@
+var _arity = require('./internal/_arity');
+var _concat = require('./internal/_concat');
+var _curry2 = require('./internal/_curry2');
+
+
+/**
+ * `tryCatch` takes two functions, a `tryer` and a `catcher`. The returned
+ * function evaluates the `tryer`; if it does not throw, it simply returns the
+ * result. If the `tryer` *does* throw, the returned function evaluates the
+ * `catcher` function and returns its result. Note that for effective
+ * composition with this function, both the `tryer` and `catcher` functions
+ * must return the same type of results.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig (...x -> a) -> ((e, ...x) -> a) -> (...x -> a)
+ * @param {Function} tryer The function that may throw.
+ * @param {Function} catcher The function that will be evaluated if `tryer` throws.
+ * @return {Function} A new function that will catch exceptions and send then to the catcher.
+ * @example
+ *
+ *      R.tryCatch(R.prop('x'), R.F, {x: true}); //=> true
+ *      R.tryCatch(R.prop('x'), R.F, null);      //=> false
+ */
+module.exports = _curry2(function _tryCatch(tryer, catcher) {
+  return _arity(tryer.length, function() {
+    try {
+      return tryer.apply(this, arguments);
+    } catch (e) {
+      return catcher.apply(this, _concat([e], arguments));
+    }
+  });
+});
--- a/test/tryCatch.js
+++ b/test/tryCatch.js
@@ -1,0 +1,67 @@
+var R = require('..');
+var eq = require('./shared/eq');
+
+
+describe('tryCatch', function() {
+
+  function headX(ls) {
+    return ls[0].x;
+  }
+
+  function catcher() {
+    return 10101;
+  }
+
+  it('takes two functions and return a function', function() {
+    var mayThrow = R.tryCatch(headX, catcher);
+    eq(typeof mayThrow, 'function');
+  });
+
+  it('returns a function with the same arity as the `tryer` function', function() {
+    function a1(a) { return a; }
+    function a2(a, b) { return b; }
+    function a3(a, b, c) { return c; }
+    function a4(a, b, c, d) { return d; }
+
+    eq(R.tryCatch(a1, catcher).length, a1.length);
+    eq(R.tryCatch(a2, catcher).length, a2.length);
+    eq(R.tryCatch(a3, catcher).length, a3.length);
+    eq(R.tryCatch(a4, catcher).length, a4.length);
+  });
+
+  it('returns the value of the first function if it does not throw', function() {
+    var mayThrow = R.tryCatch(headX, catcher);
+    eq(mayThrow([{x: 10}, {x: 20}, {x: 30}]), 10);
+  });
+
+  it('returns the value of the second function if the first function throws', function() {
+    function throw10() {
+      throw new Error(10);
+    }
+
+    function eCatcher(e) {
+      return Number(e.message);
+    }
+
+    var mayThrow = R.tryCatch(headX, catcher);
+    eq(mayThrow([]), 10101);
+
+    var willThrow = R.tryCatch(throw10, eCatcher);
+    eq(willThrow([]), 10);
+    eq(willThrow([{}, {}, {}]), 10);
+  });
+
+  it('the second function gets passed the error object and the same arguments as the first function', function() {
+    function thrower(a, b, c) {
+      void c;
+      throw new Error('throwerError');
+    }
+
+    function catch3(e, a, b, c) {
+      return [e.message, a, b, c].join(' ');
+    }
+
+    var mayThrow = R.tryCatch(thrower, catch3);
+    eq(mayThrow('A', 'B', 'C'), 'throwerError A B C');
+  });
+});


asaf-romano:
Shouldn't this work more like ``ifElse`` (or ``converge``), which always returns a function (with the arity set to to maximum possible arity)? While not supporting a call like this:
``tryWith(f1, f2, ...args)``

It'll support non unary functions, which is a more interesting use case, imho.

I'm not asking this just because of the practical use cases, but also because the documentation here ("the first and second functions should take the same types of arguments") doesn't fit the normal use case, in which the tryer function is unary, whereas the catcher function is, very often, something like "always" (or R.T in your example).

niloy:
Agree with @asaf-romano . No reason to limit `tryWith` with just unary functions.

buzzdecafe:
> Shouldn't this work more like ifElse (or converge), which always returns a function

I could get behind that idea. However: 1) The direction of the lib is moving away from variadic functions; and 2) this is explicitly for composition, in which case you would want a unary function.

>  (with the arity set to to maximum possible arity)

Something like this?

```js
_curry2(function tryWith(tryer, catcher) {
  return arity(Math.max(tryer.length, catcher.length), function() {
    try { return tryer.apply(this, arguments); }
    catch (_) { return catcher.apply(this, arguments); }
  });
});
```

I have to say I prefer the simplicity of the original, but would go along with the second.

src/tryWith.js
```diff
@@ -0,0 +1,30 @@
...
+ * @category Function
+ * @sig (x -> a) -> (x -> a) -> x -> a
+ * @param {Function} tryer The function that may throw.
+ * @param {Function} catcher The function that will be evaluated if `tryer` throws.
+ * @param {*} x
+ * @return {*} The result of evaluating `tryer` or possibly `catcher`.
+ * @example
+ *
+ *      R.tryWith(R.head, R.T, [false, false, false]); //=> false
+ *      R.tryWith(R.head, R.T, []); //=> true
```
buzzdecafe: actually this doesn't throw, just returns `undefined`. need a better example.

asaf-romano: > Something like this?

That's what I would expect, yes, but mostly because that's how all similar functions in the library behave (including the new ``juxt`` function). Regarding composition, I can see how this function could be used as the first step in a pipeline, and then a non-unary case is as valid as it is for ``ifElse`` and friends.

> actually this doesn't throw, just returns undefined. need a better example.

``R.prop("foo", null)`` would do; only problem is that we already have ``propOr``.

buzzdecafe:
updated per notes. also passing error to catcher at @CrossEye 's suggestion

src/tryCatch.js
```diff
@@ -0,0 +1,34 @@
...
+ *
+ *      R.tryCatch(R.prop('x'), R.F, {x: true}); //=> true
+ *      R.tryCatch(R.prop('x'), R.F, null);      //=> false
+ */
+module.exports = _curry2(function _tryCatch(tryer, catcher) {
+  return _arity(tryer.length, function() {
+    try {
+      return tryer.apply(this, arguments);
+    } catch (e) {
+      return catcher.apply(this, _concat([e], arguments));
```
niloy: Would it be better to put the `exception` object at the end of the argument array? So that a `catcher` with the same arity as the `tryer` can be used without modification. For example:

```js
const getX = tryCatch(R.prop("x"), $ => "Unable to find 'x' in " + $);
```

buzzdecafe: i went back and forth on where to put the exception. I opted for first because 

1. the fact that something exceptional has happened is the most important thing. the original arguments are somewhat of a "by the way" concern;
2. error-first means you can re-use the same `catcher` in a generic way, i.e. without regard to the number of arguments.

CrossEye: I agree with error-first.  If it wasn't for the chance of throwing an error here, we wouldn't have this function.  So, it's the most significant part of this branch, and I would expect that often enough it would be all the function uses.

But I don't feel strongly about this at all.

test/tryCatch.js
```diff
@@ -0,0 +1,68 @@
...
+  });
+
+  it('returns the value of the second function if the first function throws', function() {
+    function throw10() {
+      throw new Error(10);
+    }
+
+    function eCatcher(e, ls) {
+      var n = Number(e.message);
+      return ls.length === 0 ? n : n * 1000;
```
CrossEye: I would think `return Number(e.message)` would be plenty.  The rest seems just distraction.

CrossEye:
:herb:, modulo one minor suggestion on the test above.

I don't know if this is something I would use much, but it is something I would like to know is available when needed.

test/tryCatch.js
buzzdecafe: sure -- the whole example is completely contrived. I just want to show that the both the error and the original args are present.

CrossEye: Oh, I get it.  Then I think the description should be updated. "'Returns the value of the second function if the first function throws," doesn't capture this.

buzzdecafe: changed this one to just return the number. i cover "all the args are passed" in the next test.

buzzdecafe:
anyone else? or are we good to go here?

davidchambers:
> anyone else?

I have `S.encase` and `S.encaseEither`, so I won't be using `R.tryCatch`. ;)

buzzdecafe:
> I have `S.encase` and `S.encaseEither`, so I won't be using `R.tryCatch`. 

Understood, but note that the behavior is significantly different. For a contrived example, suppose I am consuming a stream of ints and multiplying them together. I have no idea how long this stream may go on. But if I encounter a zero, obviously, the whole product will be `0`. In OCaml, this is nice to read & write:

```ml
exception Zero;;
let rec stream_product acc = function
  | [] -> acc
  | x::xs -> if x = 0 then raise Zero else (stream_product (acc * x) xs)
;; 
(* why not just return 0? because you may have built up a lot of stack *)

try stream_product 1 some_stream with Zero -> 0;;
```

In this case I just want to get an int out; wrapping that int in a `Maybe` and `map`-ping down the rest of the way might be excessive. The computation has not *failed*; it has just exited early. 

As usual, the example is contrived. Nonetheless, I think there is room for both approaches.

davidchambers:
Interesting. Thanks for the example, @buzzdecafe!

safareli:
@buzzdecafe I have just read your article about ["exceptional composition"](http://buzzdecafe.github.io/code/2016/02/05/exceptional-composition) and came to this issue. Here are some of my thoughts on this topic:


This is minimalistic linked list and a function to generate list of `n` elements to work with something:

```js
// type list a = [] | [a, list a]

const makeList = (n) => {
  let res = []
  for (let i = 0; i < n; i++) {
    res = [i, res]
  }
  return res
}
```

And this recursive product function working on that list (as one in ocaml):

```js
const productUnsafe = (xs) => xs.length == 0 ? 1 : xs[0] * productUnsafe(xs[1])
```

i'm not familiar with ocaml but in js it has a big issue, using that function on big inputs will blow the stack:

```js
try{
  productUnsafe(makeList(100000))
} catch(e) {
  console.log(e) // RangeError: Maximum call stack size exceeded
}
```

So it's only working for small inputs (which don't need big stack size) and because of that the optimisation would not be significant enough.

To make recursion version work with big inputs we need to rewrite it so such that VMs which support proper tail call optimisation could optimise it and run it using constant stack size:

```js
const product_ = (xs, acc) => xs.length == 0 ? acc : productUnsafe_(xs[1], xs[0] * acc)
const product = (xs) => product_(xs, 1)
```

So now it would work with big inputs, but as it's using constant stack we could just `return 0` :

```js
const product_ = (xs, acc) => {
  if(xs.length == 0) return acc
  if(acc === 0) return 0
  return product_(xs[1], xs[0] * acc)
}
```



Unfortunatly as  tail call is not widely supported  (current status in [node](http://node.green/#proper-tail-calls--tail-call-optimisation-)) we would need to rewerite it in while loop (it could be done mechanically). Again in that case as it's using constant stack, there is no benefit of throwing anything when we can just return:

```js
const product = (xs) => {
  let acc = 1
  while (true) {
    if(xs.length == 0) return acc
    if(acc === 0) return 0
    acc = xs[0] * acc
    xs = xs[1]
  }
}
```

And btw after #1908 is merged we could use `chainRec` (monadic recursion) and still get same results:

```js
const product = (xs) => chainRec(Either, (next, done, {acc,xs}) => {
    if(xs.length == 0) return Right(done(acc))
    if(acc === 0) return Left(Zero)
    return Right(next({acc: acc + xs[0], xs: xs[1]}))
}, {acc:1, xs}).cata({
  Left: (e) => 0,
  Right: (res) => res,
})
```

Actually here `Either` is not needed and we could do same with `Identity`  :

```js
const product = (xs) => chainRec(Identity, (next, done, {acc,xs}) => {
  if(xs.length == 0) return Identity(done(acc))
  if(acc === 0) return Identity(done(0))
  return Identity(next({acc: acc + xs[0], xs: xs[1]}))
}, {acc:1, xs}).value
```

Now consider this example:
```js
const searchTree_ = (node, data) => {
  if (equals(node.data, data)) {
    throw new ResultException(node);
  } else {
    if (!node.leftChild.isEmpty) searchTree_(node.leftChild, data);
    if (!node.leftChild.isEmpty) searchTree_(node.rightChild, data);
  }
}

const searchTree = (root, data) => {
  try {
    searchTree_(root, data)
  } catch (e) {
    if (e.isResult) return e.result
    throw e
  }
}
```

Even that could be rewritten using Maybe.chainRec and as a result it's stack safe and types are fair:
```js
// searchTree :: (Tree a, a) -> Maybe a
const searchTree = (root, data) => chainRec(Maybe, (done, next, { node, stack }) => {
  if (equals(node.data, data)) return Just(done(node))
  if (!node.leftChild.isEmpty) return Just(next({
    node:node.leftNode,
    stack: node.rightChild.isEmpty? stack : [node.rightChild, stack]
  }))
  if (node.rightChild.isEmpty) return Just(next({ node:node.rightChild, stack }))
  if (stack.length == 0) return Nothing
  return Just(next({ node: stack[0], stack: stack[1] }))
}, { node:root, stack: [] /* list of node*/ })
```

To sum up my thoughts: `tryCatch` could be useful for "fixing" APIs of some libraries which have exceptions as a part of their normal API, but in general I don't think expectations are good for control flow.

buzzdecafe:
> To sum up my thoughts: `tryCatch` could be useful for "fixing" APIs of some libraries which have exceptions as a part of their normal API, 

Yes, that is the intent; exceptions are a fact of life in JS, it seems reasonable we should have a simple way to trap them.

> but in general I don't think expectations are good for control flow.

For JavaScript, I agree. In OCaml, maybe not.

And I acknowledge your point that the fold/multiply example is not ideal, because of stack limitations, and it could be rewritten in tail-call style (at least in OCaml; as you note, that wouldn't help in JS).

safareli:
In JS it would help (with --harmony flag). What I was saying is that if recursion is using constant stack size, than we can `return` instead of `throw` as there is no huge stack to unwind and throwing has no benefit when stack size is constant. In example so far (product, search in tree) it was used for control flow which could be avoided and still have same, if not better, results. Maybe in some cases exceptions could be useful for control flow, but I don't know any of such examples and also as you mentioned `there is nothing in the signature to indicate that this function might raise an exception`

Issue #1585, opened 2016-01-11
Feature request: easier to read source


Sequoia:
Reading [the source](https://github.com/ramda/ramda/blob/v0.19.0/src/lens.js) when...
```js
module.exports = _curry2(function lens(getter, setter) {
  return function(f) {
    return function(s) {
      return map(function(v) { return setter(v, s); }, f(getter(s)));
    };
  };
});
```
`f`? `s`? `v`? `s` is... I think the property name? Or wait, maybe `v` is the propname and `s` is . . . well `f` if "function", that's clear, but what is the function's purpose here?

### Benefit
1. Easier to read source
2. Makes Ramda a better teaching tool
3. Makes it easier for newbs to understand & contribute

### Cost
1. ???

CrossEye:
Are you looking for names for those three properties in this function, or do you think it's an issue endemic to Ramda in general?

Ramda does try to keep reasonably readable source-code.  I picked another function at random.  `anyPass` is defined like this:

```js
module.exports = _curry1(function anyPass(preds) {
  return curryN(reduce(max, 0, pluck('length', preds)), function() {
    var idx = 0;
    var len = preds.length;
    while (idx < len) {
      if (preds[idx].apply(this, arguments)) {
        return true;
      }
      idx += 1;
    }
    return false;
  });
});
```

Here we do abbreviate a number of variables, `'idx'` for `'index'`, `'len'` for `'length'`, and `'preds'` for `'predicates'`, but to me its quite readable.  I _believe_ the vast majority of our source code is of this type.

Lenses are a particularly abstract data structure.  Perhaps we could replace `'f'`, `'s'`, and `'v'` -- if my tenuous grasp of lenses is correct (pinging @davidchambers or @scott-christopher!)  -- with `'fn'`, `'structure'`, and `'value'`.  Note that `'fn'` is what we generally use for arbitrary functions.  But `'structure'` and `'value'` are not really meaningful names; they're as bad as calling a variable `'data'` or `'info'`.  I think making a change like that just generates a false sense of understanding: "Oh so **this** is the structure, and **that** is the value, I get it!"  But I can certainly be convinced otherwise.

Do you have other suggested names for these?  Or any way to write the code that's more clear?

CrossEye:
... and I forgot to say that "property name" is not the correct abstraction.  Lenses are applicable to all sorts of data structures, not simply Objects.

CrossEye:
[...]

PR #1588, opened 2016-01-12
Improve comments around lens functions

scott-christopher:
In response to #1585, this change adds some more descriptive argument names along with some code comments around the various lens functions.

### DIFF
--- a/src/lens.js
+++ b/src/lens.js
@@ -26,9 +26,14 @@ var map = require('./map');
  *      R.over(xLens, R.negate, {x: 1, y: 2});  //=> {x: -1, y: 2}
  */
 module.exports = _curry2(function lens(getter, setter) {
-  return function(f) {
-    return function(s) {
-      return map(function(v) { return setter(v, s); }, f(getter(s)));
+  return function(toFunctorFn) {
+    return function(target) {
+      return map(
+        function(focus) {
+          return setter(focus, target);
+        },
+        toFunctorFn(getter(target))
+      );
     };
   };
 });
--- a/src/over.js
+++ b/src/over.js
@@ -24,11 +24,16 @@ var _curry3 = require('./internal/_curry3');
  *      R.over(headLens, R.toUpper, ['foo', 'bar', 'baz']); //=> ['FOO', 'bar', 'baz']
  */
 module.exports = (function() {
+  // `Identity` is a functor that holds a single value, where `map` simply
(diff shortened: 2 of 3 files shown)


CrossEye:
:herb:

This looks great!  I'm not generally a fan of source-code comments, but this material is difficult enough to warrant them.

davidchambers:
:deciduous_tree: I *wrote* this code yet I very much appreciate these comments!

Sequoia:
:palm_tree: +1 to the comments & descriptive names! I still don't fully grok lenses, but this might help make it easier :smile_cat: 

One style question: why are `Const` and `Ident` capitalized? In Javascript there is a very strong convention of capitalizing only "constructor" functions, i.e. it signals that the function should be called with `new MyConstructor`. Speaking as someone from the JS world rather than FP world, I can confidently say that this is likely to cause at least confusion for some people.

jackfirth:
@Sequoia In the FP JS world, type constructors are meant to be normal functions - you should never need to call anything with `new`. Shoving functions that require `new` to call them into `compose` chains for example, is quite annoying and bug prone. In Ramda specifically, capitalized functions are usually constructors for things that implement the Javascript equivalent of Haskell's typeclasses (I think). For instance, `Const` and `Ident` specifically refer to the `Const` and `Identity` types from Haskell, which are both examples of very simple Functors. In Javascript and in Ramda, typeclasses are [defined in terms of the Fantasy Land spec](https://github.com/fantasyland/fantasy-land).

Disclaimer: I am but a humble lurker, and many more knowledgeable people than me can contribute better explanations than this.

davidchambers:
> Disclaimer: I am but a humble lurker, and many more knowledgeable people than me can contribute better explanations than this.

The comments you have posted recently have been excellent! You undersell yourself (which I appreciate, as a New Zealander).

To add one point to @jackfirth's comment, the names `id`/`identity` and `const`/`constant` are commonly given to functions quite different from those in question:

```haskell
id :: a -> a
const :: a -> b -> a
```

In this case, we have:

```haskell
Identity :: a -> Identity a
Const :: a -> Const a
```

`Identity` is a constructor (in the Haskell sense) for an Identity value. In Haskell, data constructors must be capitalized. There's no such requirement in JavaScript, but it's convenient to be able to define both `Identity` and `identity` so we've adopted the Haskell approach as a convention.

Sequoia:
Fair enough!

Issue #1505, opened 2015-11-08
Make ramda more modular


luizbills:
my proposal

``` js
var R = require('ramda'); // import ALL ramda functions
var compose = require('ramda/compose'); // import JUST the compose function
...
```

davidchambers:
`require('ramda/src/compose')` currently works, but it would be nice to use the cleaner path.

luizbills:
That's the point. :+1:

CrossEye:
:+1:

We've discussed this before, but no one's gotten around to doing it.  It should just be a matter of how the NPM package is built.

@luizbills: Do you feel up to taking a swing at it?

luizbills:
This allows custom builds easy. Ramda is very big. Hardly ever someone will use the entire library.

Hypercubed:
Any thoughts on converting to ES6 modules and using rollup?  Tree-shaking Ramda with rollup would be amazing.

CrossEye:
@Hypercubed: Could you describe the advantages?  I haven't tried `rollup` yet.

Although it's not well-documented, it's fairly simple to build your own version of Ramda right now.  This doesn't start with static analysis of a codebase, though, but with a list of public functions you want to include.  The build does the recursive dependency management and builds as small a package as possible.

Hypercubed:
Using ES6 modules would improve static analysis of the code base.... something that is difficult with CJS modules.  On your end Rollup would do pretty much what you described... take an index file and build the smallest a package as possible (maybe replacing your build script).

I guess the advantage of using ES6 modules would be to allow users to do tree shaking on their end.  They can import ramda methods individually, without the need to know the internal paths, then use rollup tree shaking to get a minimal build.

For example in my project I can do:

```
import {compose, isNil, equals} from 'ramda';
```

to get only the relevant portions of the ramda codebase.

Because Ramda is already pretty modular it's probably not to hard to convert to ES6 modules.  The hard part would be converting the build process.

CrossEye:
[...]

PR #1597, opened 2016-01-15
Optimize intersection and intersectionWith to filter the smaller list

### LINKED ISSUE #1595: Optimize R.interesction to lookup the smaller list
http://bit.ly/1RJVPp4

asaf-romano:
Closes #1595

``intersection`` performs much much better. ``intersectionWith`` is almost unchanged (and wasn't as affected to begin with) because most of the time is spent on the predicate execution, apparently.

### DIFF
--- a/src/intersection.js
+++ b/src/intersection.js
@@ -23,5 +23,13 @@ var uniq = require('./uniq');
  *      R.intersection([1,2,3,4], [7,6,5,4,3]); //=> [4, 3]
  */
 module.exports = _curry2(function intersection(list1, list2) {
-  return uniq(_filter(flip(_contains)(list1), list2));
+  var lookupList, filteredList;
+  if (list1.length > list2.length) {
+    lookupList = list1;
+    filteredList = list2;
+  } else {
+    lookupList = list2;
+    filteredList = list1;
+  }
+  return uniq(_filter(flip(_contains)(lookupList), filteredList));
 });
--- a/src/intersectionWith.js
+++ b/src/intersectionWith.js
@@ -39,11 +39,19 @@ var uniqWith = require('./uniqWith');
  *      //=> [{id: 456, name: 'Stephen Stills'}, {id: 177, name: 'Neil Young'}]
  */
 module.exports = _curry3(function intersectionWith(pred, list1, list2) {
+  var lookupList, filteredList;
+  if (list1.length > list2.length) {
+    lookupList = list1;
+    filteredList = list2;
+  } else {
+    lookupList = list2;
+    filteredList = list1;
+  }
   var results = [];
   var idx = 0;
-  while (idx < list1.length) {
-    if (_containsWith(pred, list1[idx], list2)) {
-      results[results.length] = list1[idx];
+  while (idx < filteredList.length) {
+    if (_containsWith(pred, filteredList[idx], lookupList)) {
+      results[results.length] = filteredList[idx];
     }
     idx += 1;
   }


CrossEye:
:herb:

src/intersection.js
```diff
@@ -23,5 +23,7 @@ var uniq = require('./uniq');
  *      R.intersection([1,2,3,4], [7,6,5,4,3]); //=> [4, 3]
  */
 module.exports = _curry2(function intersection(list1, list2) {
-  return uniq(_filter(flip(_contains)(list1), list2));
+  var lookupList = list1.length > list2.length ? list1 : list2;
+  var filteredList = lookupList == list1 ? list2 : list1;
```
davidchambers: I'd prefer to branch once rather than twice:

```javascript
if (list1.length > list2.length) {
  // ...
} else {
  // ...
}
return ...;
```

CrossEye: Ahh, for the days when we can count on destructuring...

```js
let [lookupList, filteredList] = list1.length > list2.length ? [list1, list2] : [list2, list1];
```

asaf-romano: @CrossEye it's all a matter of integrating with babel ;)

@davidchambers so,
```js
if (...) {
  return ...uniq...filter..
} else {
  return ...uniq...filter..
}
```
or
```js
var lookupList, filteredList;
if (...) {
 // set them
}
else { ... }
return ...uniq...filter;
}
```
Because I honestly don't like either options, and would also not like to depend on ``var`` hoisting.

davidchambers: There's one more option:

```javascript
var pair = list1.length > list2.length ? [list1, list2] : [list2, list1];
return uniq(_filter(flip(_contains)(pair[0]), pair[1]));
```

asaf-romano: I could do that, but it will probably hurt readability in ``intersectionWith``

CrossEye: @asaf-romano:
>> Ahh, for the days when we can count on destructuring...

> it's just a matter of integrating with babel ;)

Yes, if we want this in our source:

```js
var _slicedToArray = (function () { function sliceIterator(arr, i) { var _arr = []; var _n = true; var _d = false; var _e = undefined; try { for (var _i = arr[Symbol.iterator](), _s; !(_n = (_s = _i.next()).done); _n = true) { _arr.push(_s.value); if (i && _arr.length === i) break; } } catch (err) { _d = true; _e = err; } finally { try { if (!_n && _i["return"]) _i["return"](); } finally { if (_d) throw _e; } } return _arr; } return function (arr, i) { if (Array.isArray(arr)) { return arr; } else if (Symbol.iterator in Object(arr)) { return sliceIterator(arr, i); } else { throw new TypeError("Invalid attempt to destructure non-iterable instance"); } }; })();

var _ref = list1.length > list2.length ? [list1, list2] : [list2, list1];

var _ref2 = _slicedToArray(_ref, 2);

var lookupList = _ref2[0];
var filteredList = _ref2[1];
```
(http://bit.ly/1U2w2pB)  I have no idea why it needs that `_slicedToArray` monstrosity here.  It's _creating_ a clearly iterable instance the line above.

asaf-romano: FWIW, I think that for Ramda, which will never use ES6 classes and all that jazz, "loose mode" is good enough (you can try that there). But are you worried about readability or performance? Because I think it's reasonable to expect consumers to read the pre-compilation source code.

CrossEye: I would not want to ship with Babel as a dependency. We would have to build the output ourselves.

I very much like having the shipped source being as clean as possible, noting that `ramda.js` is our flagship project and should always be as readable and clean as possible.

I wasn't worried about performance.

@scott-christopher [recently demonstrated][sc] that at least with rollup, we can build very clean packages, one of the things that has bothered me about babel and webpack.


  [sc]: https://github.com/ramda/ramda/issues/1505#issuecomment-165610371

davidchambers:
The change to `intersectionWith` is harmful, I believe.

I just tried to use the function to answer a [question][1] on Stack Overflow. The goal is to filter a list of items by a list of ids.

```javascript
const items = [
  {id: 1, name: 'a'},
  {id: 2, name: 'b'},
  {id: 3, name: 'c'},
  {id: 4, name: 'd'},
  {id: 5, name: 'e'},
  {id: 6, name: 'f'},
];

const ids = [1, 3, 5];
```

I thought it would be neat to use `intersectionWith` to solve this. I tried this:

```javascript
R.intersectionWith((item, id) => item.id === id, items, ids)
```

I was surprised to discover that the function received `(id, item)` even though I provided `items, ids`. Stranger still, switching to `ids, items` had no effect on the order!

This made it clear to me that the change to `intersectionWith` was not *just* an optimization: it makes the function unreliable. Were one to provide `[1, 2, 3, 4, 5, 6]` rather than `[1, 3, 5]`, the arguments provided to the predicate would flip! Additionally, the order in which the lists are provided must determine the list to be filtered. The user should be able to provide `[a]` and `[b]` lists and know that she'll get back an `[a]` list rather than a `[b]` list or a mixed list.

I suggest we revert the change to `intersectionWith`. Do others agree?


[1]: http://stackoverflow.com/q/37091448/312785

CrossEye:
I don't know.  You're clearly not supplying a predicate of the form `:: a -> a -> Boolean`.  (BTW, presumably `intersectionWith` should be `(a -> a -> Boolean) -> [a] -> [a] -> [a]` rather than the currently documented `(a -> a -> Boolean) -> [*] -> [*] -> [*]`.)

I don't think we could expand `interesectionWith` to cover the case of the SO post, where `list1` and `list2` contain different types.  Remember that the predicate is essentially saying that the two things passed are equal as far as you're concerned.  But `{id: 3, name: 'c'}` is probably not equal to `3`.

I had looked at the post and came up with http://goo.gl/d4fqoJ, but didn't think it was worth posting, as that added nothing to the two decent vanilla JS solutions already posted.

davidchambers:
I like this signature:

```javascript
intersectionWith :: (a -> b -> Boolean) -> [a] -> [b] -> [a]
```

We could easily support this more general type. Why do you favour the less general type, Scott?

buzzdecafe:
intersection of two sets of different types should be the empty set. The user indicates that both sides are the same by passing in the equality function. So I agree with @CrossEye that `(a -> a -> Boolean) -> [a] -> [a] -> [a])` makes more sense.

CrossEye:
Yes, that.  `intersection` should be about equality, even when generalized to `intersectionWith`.  I could see another function that does precisely what the SO question asked for, though, and what you were reaching for with `(a -> b -> Boolean) -> [a] -> [b] -> [a]`.

davidchambers:
Scott, could you provide a usage example for `intersectionWith`? The one in the docstring isn't helpful as `intersection` would work equally well. It's not obvious to me why an equality function is necessary to compare values of the same type. Are you thinking of row polymorphism?

```javascript
const points2D = [{x: 1, y: 1}, {x: 2, y: 2}, {x: 3, y: 3}];
const points3D = [{x: 1, y: 1, z: 1}, {x: 2, y: 2, z: 2}];

R.intersectionWith((p, q) => R.equals(R.pick(['x', 'y'], p),
                                      R.pick(['x', 'y'], q)),
                   points2D,
                   points3D);
```

In this case `a` is `{ x :: Number, y :: Number }`.

Should the result be `[{x: 1, y: 1}, {x: 2, y: 2}]` or `[{x: 1, y: 1, z: 1}, {x: 2, y: 2, z: 2}]`, or should I not care since both satisfy `Array { x :: Number, y :: Number }`?

It's worth noting that even in the example above, the predicate is complicated by the fact that we don't know the order in which the arguments will be provided. We're forced to apply `R.pick(['x', 'y'])` to *both* arguments rather than just the argument of type `{ x :: Number, y :: Number, z :: Number }`.

CrossEye:
> Are you thinking of row polymorphism?

Yes, although that phrase was escaping me.  Another example would be `intersectionWith(eqProps('id'), list1, list2)`.

> It's worth noting that even in the example above, the predicate is complicated by the fact that we don't know the order in which the arguments will be provided.

Perhaps, although I would generally find it odd to define my equality predicate directly in terms of `equals`.  I would probably handle that one as `(p, q) => p.x == q.x && p.y == q.y` or perhaps `both(eqProps('x'), eqProps('y'))`.  I'm not necessarily arguing that you're wrong in 

> The change to `intersectionWith` is harmful,

only that this function shouldn't reasonably handle the instance you describe.

davidchambers:
I understand your point of view with respect to input types. Where do you stand with respect to output type? Does this bother you?

```javascript
> R.intersectionWith(R.eqProps(['id']), [{id: 1}, {id: 2}], [{id: 1, name: 'one'}])
[{id: 1, name: 'one'}]

> R.intersectionWith(R.eqProps(['id']), [{id: 1}, {id: 2}], [{id: 1, name: 'one'}, {id: 2, name: 'two'}])
[{id: 1}, {id: 2}]
```

If the answer is no, we needn't change anything. :)

CrossEye:
I can see how it might bother someone.  @buzzdecafe and I discussed this when the PR first came out.  It doesn't bother me, personally, since the user has already made it clear that to her mind, `{id: 1, name: 'one'}` and `{id: 1}` are equal.  Again, I can easily see the value of something like  `(a -> b -> Boolean) -> [a] -> [b] -> [a]`, though.  I just would use a different name for it.

buzzdecafe:
`innerJoin :: (a -> b -> Boolean) -> [a] -> [b] -> [a]` ?

CrossEye:
Yes.

buzzdecafe:
I think we should rename `intersectionWith` to `innerJoin` and use the `(a -> b -> Boolean) -> [a] -> [b] -> [a]` signature. That seems clearest to me.

davidchambers:
I agree, @buzzdecafe! The more general type signature covers both use cases. :)

CrossEye:
Yes, I agree too, although perhaps some more bikeshedding on the name is in order.

Bradcomp:
For what it's worth, I like the _name_ `innerJoin`, but I would prefer a more general signature to go with that name.

`(a -> b -> c) -> (a -> b -> Boolean) -> [a] -> [b] -> [c]`

One could also flip the first two arguments if that made more sense. 

While it's more complex, it encapsulates what I would want if I were to use such a function.  The signature above falls out of the more complex one as `innerJoin(nthArg(0))`, or `innerJoin(always)`.

The benefit would be the ability to do stuff like `innerJoin(merge, eqProps('id'))` which is more similar to how I view a SQL join.

CrossEye:
Yes, that is more general still, of course.  And not doing that was the reason I didn't want to call it `innerJoin`.

I could see including both this one and `innerJoin(always)` under some name.

PR #1598, opened 2016-01-16
Add function reduceBy (a more general groupBy)

jirutka:
### Motivation

I very often need to group elements of a list and immediately apply some transformation to the grouped sub-lists or their elements. It’s usually quite ineffective to do it in two steps (requires another loop with nested loop and intermediate arrays).

I’m quite sure that it’s very common need so there _must be_ some “standard” name for such function. I’ve tried to find something in Haskell, but without success. Eventually I found a function named `reduceBy` which does what I need in these two libs:

* [itertoolz.reduceby](http://toolz.readthedocs.org/en/latest/api.html#toolz.itertoolz.reduceby) (Python), 
* [CloudFlow.reduceBy](https://github.com/mbraceproject/MBrace.Core/blob/7de07474e7ed32adc90a66c93c6e8fa279577a14/src/MBrace.Flow/Combinators.fs#L646-L666) (F#).

#### Example 1

```js
const students = [
  { name: 'Lucy', score: 92 },
  { name: 'Drew', score: 85 },
  // ...
  { name: 'Bart' score: 62 },
]
const grade = (n) => (n < 65) ? 'F' : (n < 70) ? 'D' : (n < 80) ? 'C' : (n < 90) ? 'B' : 'A'
const byGrade = R.pipe(R.prop('score'), grade)

// A. Using R.groupBy and nested pair of R.map:
const namesByGrade = R.pipe(
  R.groupBy(byGrade),
  R.map(R.map(R.prop('name')))
)

// B. Using new R.reduceBy:
const namesByGrade = R.reduceBy(byGrade, (acc, stud) => acc.concat(stud.name), [])

namesByGrade(students)
// {
//   'A': ['Lucy'],
//   'B': ['Drew'],
//   // ...
//   'F': ['Bart'],
// }
```

#### Example 2

```js
const tweets = [
  { author: '@drboolean', text: 'Lorem ipsum', favs: 60 },
  { author: '@dan_abramov', text: 'dolor sit amet', favs: 80 },
  { author: '@drboolean', tex: 'consectetur', favs: 40 },
  // ...
]

// A. Using R.groupBy, R.map and nested R.reduce:
const sumFavsByUser = R.pipe(
  R.groupBy(R.prop('author')),
  R.map(R.reduce((acc, tw) => acc + tw.favs))
)

// B. Using new R.reduceBy:
const sumFavsByUser = R.reduceBy(R.prop('author'), (acc, tw) => acc + tw.favs, 0)

sumFavsByUser(tweets)
// {
//    '@drboolean': 100,
//    '@dan_abramov': 80,
// }
```

### Transducer?

I’ve implemented also `_xreduceBy` transducer, but I’m not sure how to use it (and so how to test it). How can be `groupBy` used with `R.transduce` or `R.into`?

<a name="poll" />
### Poll

| Name               | Vote                                                                       |
|--------------------|----------------------------------------------------------------------------|
| @davidchambers     | [:+1:](https://github.com/ramda/ramda/pull/1598#issuecomment-173035182)    |
| @buzzdecafe        | [:open_hands:](https://github.com/ramda/ramda/pull/1598#issuecomment-175007799) (neutral) |
| @CrossEye          | [:+1:](https://github.com/ramda/ramda/pull/1598#issuecomment-177531934)    |
| @scott-christopher | [:+1:](https://github.com/ramda/ramda/pull/1598#issuecomment-178314222)                                 |
| @raine             | :question:                                                                 |

### DIFF
--- a/src/reduceBy.js
+++ b/src/reduceBy.js
@@ -1,0 +1,55 @@
+var _curryN = require('./internal/_curryN');
+var _has = require('./internal/_has');
+var _reduce = require('./internal/_reduce');
+
+
+/**
+ * Groups the elements of the list according to the result of calling
+ * the String-returning function `keyFn` on each element and reduces the elements
+ * of each group to a single value via the reducer function `valueFn`.
+ *
+ * This function is basically a more general `groupBy` function.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig (b -> String) -> ((a, b) -> a) -> a -> [b] -> {String: a}
+ * @param {Function} keyFn The function that maps the list's element into a key.
+ * @param {Function} valueFn The function that reduces the elements of each group to a single
+ *        value. Receives two values, accumulator for a particular group and the current element.
+ * @param {*} acc The (initial) accumulator value for each group.
+ * @param {Array} list The array to group.
+ * @return {Object} An object with the output of `keyFn` for keys, mapped to the output of
+ *         `valueFn` for elements which produced that key when passed to `keyFn`.
+ * @see R.groupBy, R.reduce
+ * @example
+ *
+ *      var byGrade = R.reduceBy(function(student) {
+ *        var score = student.score;
+ *        return score < 65 ? 'F' :
+ *               score < 70 ? 'D' :
+ *               score < 80 ? 'C' :
+ *               score < 90 ? 'B' : 'A';
+ *      });
+ *      var namesByGrade = byGrade(function(acc, student) {
+ *        return acc.concat(student.name);
+ *      }, []);
+ *      var students = [{name: 'Lucy', score: 92},
+ *                      {name: 'Drew', score: 85},
+ *                      // ...
+ *                      {name: 'Bart', score: 62}];
+ *      namesByGrade(students);
+ *      // {
+ *      //   'A': ['Lucy'],
+ *      //   'B': ['Drew']
+ *      //   // ...,
+ *      //   'F': ['Bart']
+ *      // }
+ */
+module.exports = _curryN(4, [], function reduceBy(keyFn, valueFn, valueAcc, list) {
+  return _reduce(function(acc, elt) {
+    var key = keyFn(elt);
+    acc[key] = valueFn(_has(key, acc) ? acc[key] : valueAcc, elt);
+    return acc;
+  }, {}, list);
+});
--- a/test/reduceBy.js
+++ b/test/reduceBy.js
@@ -1,0 +1,57 @@
+var R = require('..');
+var eq = require('./shared/eq');
+
+var byType = R.prop('type');
+var sumValues = function(acc, obj) {return acc + obj.val;};
+
+describe('reduceBy', function() {
+  it('splits the list into groups according to the grouping function', function() {
+    var grade = function(score) {
+      return (score < 65) ? 'F' : (score < 70) ? 'D' : (score < 80) ? 'C' : (score < 90) ? 'B' : 'A';
+    };
+    var students = [
+      {name: 'Abby', score: 84},
+      {name: 'Brad', score: 73},
+      {name: 'Chris', score: 89},
+      {name: 'Dianne', score: 99},
+      {name: 'Eddy', score: 58},
+      {name: 'Fred', score: 67},
+      {name: 'Gillian', score: 91},
+      {name: 'Hannah', score: 78},
+      {name: 'Irene', score: 85},
+      {name: 'Jack', score: 69}
+    ];
+    var byGrade = function(student) {return grade(student.score || 0);};
+    var collectNames = function(acc, student) {return acc.concat(student.name);};
+    eq(R.reduceBy(byGrade, collectNames, [], students), {
+      A: ['Dianne', 'Gillian'],
+      B: ['Abby', 'Chris', 'Irene'],
+      C: ['Brad', 'Hannah'],
+      D: ['Fred', 'Jack'],
+      F: ['Eddy']
+    });
+  });
+
+  it('returns an empty object if given an empty array', function() {
+    eq(R.reduceBy(byType, sumValues, 0, []), {});
+  });
+
+  it('is curried', function() {
+    var splitByType = R.reduceBy(byType);
+    var sumByType = splitByType(sumValues, 0);
+    eq(sumByType([
+      {type: 'A', val: 10},
+      {type: 'B', val: 20},
+      {type: 'A', val: 30},
+      {type: 'A', val: 40},
+      {type: 'C', val: 50},
+      {type: 'B', val: 60}
+    ]), {A: 80, B: 80, C: 50});
+  });
+
+  it('correctly reports the arity of curried versions', function() {
+    var inc = R.reduceBy(byType, sumValues)(0);
+    eq(inc.length, 1);
+  });
+
+});


jirutka:
@buzzdecafe @davidchambers Any opinion about this PR…?

davidchambers:
Is your primary concern readability or efficiency?

I tend to view `reduce` as a measure of last resort (and potentially an indicator of code smell). I prefer this to your `reduceBy` example:

```javascript
R.pipe(R.groupBy(R.prop('author')),
       R.map(R.map(R.prop('favs'))),
       R.map(R.sum));
```

If efficiency is a concern, I realize that the code above is not ideal.

jirutka:
Both of them. Your code is indeed very inefficient and I don't consider it much readable either.

davidchambers:
I'll let others weigh in. I'm neutral because the specialized function seems more complex to me than a composition of simpler functions. Here's the type signature:

```haskell
reduceBy :: (b -> String) -> ((a, b) -> a) -> a -> [b] -> {String: a}
```

I won't oppose this addition to the library, so if others like it, let's add it. :)

src/reduceBy.js
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Groups the elements of the list according to the result of calling
+ * the String-returning function `keyFn` on each element and reduces the elements
+ * of each group to a single value via the reducer function `valueFn`.
+ *
+ * This function is basically a more general `groupBy` function.
+ *
+ * @func
+ * @memberOf R
+ * @since v0.20.0
```
davidchambers: Please remove this line. It'll get added automatically at release time.

jirutka:
Actually it's a more general function, you can consider `groupBy` as a specialisation of `reduceBy`:

```js
const groupBy = (fn) => reduceBy(fn, R.flip(R.append), [])
```

davidchambers:
You know the way to my :heart:, @jirutka! Since we already provide `groupBy`, I'm in favour of also providing the more basic building block, `reduceBy`. :)

src/reduceBy.js
```diff
@@ -0,0 +1,55 @@
...
+ *      //   'A': ['Lucy'],
+ *      //   'B': ['Drew']
+ *      //   // ...,
+ *      //   'F': ['Bart']
+ *      // }
+ */
+module.exports = _curryN(4, [], function reduceBy(keyFn, valueFn, valueAcc, list) {
+  return _reduce(function(acc, elt) {
+    var key = keyFn(elt);
+    acc[key] = valueFn(acc[key] || valueAcc, elt);
```
davidchambers: I believe this use of `||` to be unsound (as it so often is in JavaScript). If `key` is a value such as `'toString'` or `'valueOf'` this check will erroneously pass. Let's use our internal `_has` function to perform this check instead.

CrossEye:
> Any opinion about this PR…?

Sorry, I've been a bit under the weather, and this one slipped by entirely.

I have mixed feelings here.  I'd really like to hear from folks who know the tranducer stuff the best.  Would this be easily handled with our current transducers?  Would a simple extension to a few new functions help?  Pinging @buzzdecafe, @kevinbeaty, @kedashoe .  If this is achievable with transducers, that seems like the best way to handle it.

If it's not, then I would have to think a bit more.  Sure it's an abstraction of `groupBy`.  But that does not mean it's the definitive abstraction of it, or even a particularly great one.

It's not that I object specifically.  But if the strongest argument for it is an argument for doing things more efficiently in a single pass, then the same argument essentially applies to the way much code using Ramda (at least code without tranducers) already works.

I'm trying to get a feel for the other argument, that of a useful general abstraction.  That may just take another reading or two.  Maybe when I can breathe a little better.  :frowning:

kedashoe:
Transducers would not particularly help with the performance of `groupBy` (it uses an intermediate collection https://github.com/ramda/ramda/blob/master/src/internal/_xgroupBy.js#L30-L31 and any subsequent xforms would not be called until all the input is processed https://github.com/ramda/ramda/blob/master/src/internal/_xgroupBy.js#L18)

This looks like a useful function, but for me it is not common enough to go in ramda core vs the cookbook.

We might define a function `filterMap`, `(a -> Boolean) -> (a -> b) -> [a] -> [b]`

```javascript
var filterMap = R.curry(function filterMap(pred, mapFn, data) {
  var idx, d;
  var result = [];
  for (idx = 0; idx < data.length; ++idx) {
    d = data[idx];
    if (pred(d)) {
      result.push(mapFn(d));
    }
  }
  return result;
});
```
Now, transducers can help performance in the case of a `filter` and then a `map`, but they will never be as fast as this function. And I think you can pretty safely make the case we filter then map a lot more often then we group by and then do something else. So should we add `filterMap` to ramda?

CrossEye:
I guess I was wondering if `groupBy` could be transducerified (??) the way a number of other functions have been.

kedashoe:
> I guess I was wondering if groupBy could be transducerified (??) the way a number of other functions have been.

Sorry not sure I follow.. are you still wondering? Or you were wondering but aren't wondering anymore? We have transducerified :zap: `groupBy`, but it doesn't have the same properties as `xmap`, `xfilter` etc as all the input must be processed before it can `step` to the next `xform` in the transducer stack.

CrossEye:
> We have transducerified :zap: `groupBy`, but it doesn't have the same properties as `xmap`, `xfilter` etc as all the input must be processed before it can `step` to the next `xform` in the transducer stack.

Oh, of course.  I clearly wasn't thinking straight (read: at all!)

Sorry, I don't think I should be making too many decisions this week.  :smile:

But I think I'm overall :+1:  on this.

:herb:

jirutka:
I’ve implemented `_xgroupBy` for `_xreduceBy`:

```js
var _curryN = require('./_curryN');
var _has = require('./_has');
var _xfBase = require('./_xfBase');


  function XReduceBy(keyFn, valueFn, valueAcc, xf) {
    this.xf = xf;
    this.keyFn = keyFn;
    this.valueFn = valueFn
    this.valueAcc = valueAcc
    this.inputs = {};
  }
  XReduceBy.prototype['@@transducer/init'] = _xfBase.init;
  XReduceBy.prototype['@@transducer/result'] = function(result) {
    var key;
    for (key in this.inputs) {
      if (_has(key, this.inputs)) {
        result = this.xf['@@transducer/step'](result, this.inputs[key]);
        if (result['@@transducer/reduced']) {
          result = result['@@transducer/value'];
          break;
        }
      }
    }
    this.inputs = null;
    return this.xf['@@transducer/result'](result);
  };
  XReduceBy.prototype['@@transducer/step'] = function(result, input) {
    var key = this.keyFn(input);
    var group;
    this.inputs[key] = group = _has(key, this.inputs) ? this.inputs[key] : [key, this.valueAcc];
    this.inputs[key][1] = this.valueFn(group[1], input);
    return result;
  };

  return _curryN(4, [], function _xreduceBy(keyFn, valueFn, valueAcc, xf) {
    return new XReduceBy(keyFn, valueFn, valueAcc, xf);
  });
}());
```

The problem is that I’m not sure how to properly test it (existing tests for `_xgroupBy` are IMHO insufficient, so it’s not a good inspiration).

I’ll add some benchmarks to the discussion later, I think that it’ll be good argument. :)

jirutka:
Sorry for long delay, here are benchmarks.

[reduceBy vs. groupBy |> map(map)](http://jsperf.com/ramda-reduceby): `reduceBy` is 10x faster on Firefox 43, 20x faster on Firefox 45.

@kedashoe, [filterMap vs. filter |> map](http://jsperf.com/ramda-filtermap/3): only 2x faster on Firefox. As you see, it’s not a valid comparison.

kedashoe:
> @kedashoe, filterMap vs. filter |> map: only 2x faster on Firefox. As you see, it’s not a valid comparison.

I wasn't trying to say `filterMap` will see similar speed gains (and why isn't 2x vs 10x valid? what is the cutoff for something being valid?), simply that this seems rare to me, I'm worried about ramda growing too big, and we could include similar functions that are more common (ie, doesn't matter if you have 10x speed if you don't use it, 2x is nice if you use it all the time). In any event, I don't have a vote and even if I did I have been outvoted! :smile: Nice work and I appreciate the perf stuff!

davidchambers:
We don't exactly operate by voting, but your views are certainly important to me, @kedashoe. I agree we need to keep an eye on the size of the library.

jirutka:
Well, then we may consider splitting Ramda into few modules. For example lenses can be separated into ramda-lenses. WDYT?

jirutka:
BTW, there's one real-world usage of `reduceBy`: https://github.com/cvut/fittable/blob/ce378a38eb46cfb4d5b55592f70c5c54b9c79513/src/semesterWeeks.js#L102

CrossEye:
@davidchambers 
> We don't exactly operate by voting, but your views are certainly important to me, @kedashoe.

I think they are to everyone here.  Yours have been invaluable contributions to our community.

> I agree we need to keep an eye on the size of the library.

Of course.  It's one more of the balancing acts we have to perform.  I believe the larger the library grows, the more new functions will need to prove their worth before being included.  That fact alone makes it more challenging to accept new functions, because they will themselves make it harder for further new ones.  But that's all right.  Most likely, as we go along, the new functions that will be proposed will be more specialized than the ones we already have..


@jirutka:
> Well, then we may consider splitting Ramda into few modules. For example lenses can be separated into ramda-lenses. WDYT?

That effort is [already underway][rl] for lenses.  But I don't think we'll find many easy places like that to split the library.  Although we can build smaller versions, we haven't promoted this at all, and we don't have any suggested smaller builds available.


  [rl]: https://github.com/ramda/ramda-lens

CrossEye:
My feeling is :+1: on including `reduceBy`.  What do others think?

buzzdecafe:
I am with @kedashoe on this one; useful sure, but common enough for core? But I agree with @davidchambers above: 

> I won't oppose this addition to the library, so if others like it, let's add it. :)

jirutka:
It's two weeks since I've opened this PR, could you please merge it, or close it (I prefer former :smiley_cat:)?

CrossEye:
Sorry, we seem to be in one of our slow periods.

@davidchambers, @scott-christopher, @buzzdecafe, @raine, what say you?

I'm in favor.

jirutka:
No problem. :) Ok, let’s sum up:

| Name               | Vote                                                                       |
|--------------------|----------------------------------------------------------------------------|
| @davidchambers     | [:+1:](https://github.com/ramda/ramda/pull/1598#issuecomment-173035182)    |
| @buzzdecafe        | [:open_hands:](https://github.com/ramda/ramda/pull/1598#issuecomment-175007799) (neutral) |
| @CrossEye          | [:+1:](https://github.com/ramda/ramda/pull/1598#issuecomment-177531934)    |
| @scott-christopher | :question:                                                                 |
| @raine             | :question:                                                                 |

buzzdecafe:
no change for me. still :open_hands:

scott-christopher:
While I don't have any immediate use for this myself, I could see this being reasonably useful to others.

The only point I would make is the default accumulating value is arguably unnecessary as the first value for a new key could be used instead, effectively becoming a `reduce1`. Alternatively, the values could be mapped to some monoid type and `concat`ed together to achieve the aggregate.

jirutka:
> …default accumulating value is unnecessary as the first value for a new key could be used instead…

So you suggests this change:

```diff
+module.exports = _curryN(4, [], function reduceBy(keyFn, valueFn, valueAcc, list) {
-module.exports = _curry3(function reduceBy(keyFn, valueFn, list) {
   return _reduce(function(acc, elt) {
     var key = keyFn(elt);
+    acc[key] = valueFn(_has(key, acc) ? acc[key] : valueAcc, elt);
+    acc[key] = valueFn(acc[key], elt);
     return acc;
   }, {}, list);
 });
```

And the above mentioned examples would change to:

```diff
-const namesByGrade = R.reduceBy(byGrade, (acc, stud) => acc.concat(stud.name), [])
+const namesByGrade = R.reduceBy(byGrade, (acc, stud) => (acc || []).concat(stud.name))

-const sumFavsByUser = R.reduceBy(R.prop('author'), (acc, tw) => (acc || 0) + tw.favs)
+const sumFavsByUser = R.reduceBy(R.prop('author'), (acc, tw) => (acc || 0) + tw.favs)
```

You’re right, the default value is kinda unnecessary. This change will also make the function more efficient, because then we can use optimized `_curry3` instead of general `_curryN`. :smile_cat:

> …the values could be mapped to some monoid type and concated together to achieve the aggregate.

I’m not sure if I understand that correctly, could you please provide some example?

benperez:
I think what @scott-christopher is suggesting is that the type signature of `R.reduceBy` should change:
```diff
-(b -> String) -> ((a, b) -> a) -> a -> [b] -> {String: a}
+(a -> String) -> ((a, a) -> a) -> [a] -> {String: a}
```
Such that you don't need an initial accumulating value anymore, since the list fed to your reducing function is always guaranteed to have at least one element. It's analogous to the difference between [foldl](http://hackage.haskell.org/package/base-4.8.2.0/docs/Prelude.html#v:foldl) and [foldl1](http://hackage.haskell.org/package/base-4.8.2.0/docs/Prelude.html#v:foldl1).

I would actually argue against that, since in the second formulation, the reduced value always has to have the same type as the elements in the list. I don't think you could achieve the current example with that formulation of the function.

scott-christopher:
@jirutka: Removing the initial accumulator value would look something similar to:
```js
_curry3(function reduceBy(keyFn, valueFn, list) {
    var key = keyFn(elt);
    acc[key] = _has(key, acc) ? valueFn(acc[key], elt) : elt;
    return acc;
  }, {}, list);
});
```

@benperez: You're right. Removing the initial accumulator value does restrict the aggregated value to be of the same type as the list elements.

@jirutka: My comment about mapping to some monoid type was just me thinking out loud about using something similar to Haskell's [foldMap](hackage.haskell.org/package/base/docs/Data-Foldable.html#v:foldMap) which could look like:
```js
//:: Semigroup m => (a -> String) -> (a -> m) -> [a] -> {m}
_curry3(function reduceBy(keyFn, toMonoidFn, list) {
    var key = keyFn(elt);
    acc[key] = _has(key, acc) ? R.concat(elt, toMonoidFn(acc[key])) : toMonoidFn(elt);
    return acc;
  }, {}, list);
});
```
This could however just be implemented with your existing `reduceBy` implementation.
```js
//:: Monoid m => Type m -> (a -> String) -> (a -> m) -> [a] -> {m}
reduceByM = R.curry((M, toKeyFn, toMonoidFn, list) =>
  reduceBy(toKeyFn, (m, a) => R.concat(m, toMonoidFn(a)), M.empty(), list);
```

So after all that, I think I'm happy for the existing function to land as is. :mushroom:

jirutka:
> Removing the initial accumulator value would look something similar to…

This modification completely misses my main motivation for `reduceBy`. It can’t be used to implement any of two examples in my initial comment.

> …using something similar to Haskell's foldMap which could look like…

Thanks for the example.

What do you think about the modification I’ve described in the [previous comment](https://github.com/ramda/ramda/pull/1598#issuecomment-177644497)?

scott-christopher:
> What do you think about the modification I’ve described in the previous comment?

I'd prefer to explicitly pass the initial value to `reduceBy` rather than having to check whether the argument is `undefined` in the accumulator function.

jirutka:
It’s one month since I’ve opened this PR. Could you please make a decision?

According to [our poll](https://github.com/ramda/ramda/pull/1598#poll) 3 people are for merging, 1 is neutral, 1 didn’t respond.

Issue #1599, opened 2016-01-19
R.compose not curried by default


tcoats:
When the arity of the rightmost function in a compose takes more than one argument it does not wait to be curried before called.

```js
var valueRequired = R.compose(R.not, R.either(R.isEmpty, R.isNil));
var propRequired = R.compose(valueRequired, R.prop);

var emailRequired = propRequired('email'); // is instantly run and returns true
```

Instead the output needs to be manually curried.

```js
var valueRequired = R.compose(R.not, R.either(R.isEmpty, R.isNil));
var propRequired = R.curry(R.compose(valueRequired, R.prop)); // note: R.curry

var emailRequired = propRequired('email'); // is a function
console.log(emailRequired({ email: '' })); // false
console.log(emailRequired({ email: 'something' })); // true
```

Is this expected behaviour?

davidchambers:
Which Ramda version are you using, @tcoats? The behaviour changed in v0.16.0 (via #1260) and was changed again in v0.18.0 (via #1391).

tcoats:
Latest on npm - v0.19.1.

Reading that second pull request it seems currying compose had been removed to support optional parameters and other complexity.

Perhaps that could be added to the documentation - that compose and pipe do not curry?

davidchambers:
> Perhaps that could be added to the documentation - that compose and pipe do not curry?

Good idea. Fancy opening a pull request?

tcoats:
Functionality has been documented.

PR #1600, opened 2016-01-19
Doc update noting that pipe and compose do not automatically curry

tcoats:
As discussed in #1599 these functions are not curried by default and it's helpful to explain this in the documentation .

### DIFF
--- a/src/compose.js
+++ b/src/compose.js
@@ -6,6 +6,8 @@ var reverse = require('./reverse');
  * Performs right-to-left function composition. The rightmost function may have
  * any arity; the remaining functions must be unary.
  *
+ * **Note:** The result of compose is not automatically curried.
+ *
  * @func
  * @memberOf R
  * @since v0.1.0
--- a/src/pipe.js
+++ b/src/pipe.js
@@ -10,6 +10,8 @@ var tail = require('./tail');
  *
  * In some libraries this function is named `sequence`.
  *
+ * **Note:** The result of pipe is not automatically curried.
+ *
  * @func
  * @memberOf R
  * @since v0.1.0


davidchambers:
Thanks, @tcoats!

src/compose.js
```diff
@@ -6,6 +6,9 @@ var reverse = require('./reverse');
  * Performs right-to-left function composition. The rightmost function may have
  * any arity; the remaining functions must be unary.
  *
+ * The result of compose is not automatically curried so functions with optional
+ * arguments can be used as the rightmost function and are called immediately.
+ *
```
davidchambers: I suggest removing "automatically" and "and are called immediately".

CrossEye: I'd suggest it be even more bare-bones.  As the discussion in #1391 showed, the biggest problem was not optional arguments (or at least not as we typically think of them) but with passing partially configured functions along to another function.

Perhaps just

    **Note**: the result of compose is not automatically curried.

tcoats: In fitting with the bare-bones documentation approach I've used your example @CrossEye

tcoats:
Updated with the text "**Note**: the result of compose is not automatically curried."

src/compose.js
```diff
@@ -6,6 +6,8 @@ var reverse = require('./reverse');
  * Performs right-to-left function composition. The rightmost function may have
  * any arity; the remaining functions must be unary.
  *
+ * **Note**: the result of compose is not automatically curried.
```
davidchambers: I suggest two final tweaks:

```diff
- * **Note**: the result of compose is not automatically curried.
+ * **Note:** The result of compose is not automatically curried.
```

I'm sorry to be pedantic. :)

tcoats: No worries. Done.

tcoats:
Moved colon so "Note:" is bold from an observation from @davidchambers

src/compose.js
```diff
@@ -6,6 +6,8 @@ var reverse = require('./reverse');
  * Performs right-to-left function composition. The rightmost function may have
  * any arity; the remaining functions must be unary.
  *
+ * **Note:** the result of compose is not automatically curried.
```
davidchambers: Could we also `s/the/The/`? I recall reading that if there's just one word (or perhaps just a few words) before the colon, the word after the colon should be capitalized.

CrossEye:
Thank you @tcoats for putting this together... and for putting up with us!  :smiling_imp:

tcoats:
Happy to help. Latest change `s/the/The/

davidchambers:
:deciduous_tree: Thanks!

CrossEye:
Wow, I went to merge this and the button disappeared.  Timing, Chambers, timing!  :smile:

CrossEye:
Thanks again, @tcoats!

Issue #1327, opened 2015-08-04
reduceObj


gyzerok:
How can I achieve something like reduce for objects in ramda?

paldepind:
This question seems ambiguous to me. How would you like to reduce objects? Reduce an array of objects? Reduce the keys of an object? The values? You can use `R.toPairs` or `R.values` and use `R.reduce` on the result?

gyzerok:
AFAIK array is an object with decimal keys. I'm talking about reducing object in the same way that arrays. Check https://lodash.com/docs#reduce.

svozza:
There's an example of getting the same functionality with `useWith` here:

https://github.com/ramda/ramda/issues/1067

paldepind:
> AFAIK array is an object with decimal keys.

True. But that is more like an implementation detail in JavaScript. Reduce as a concept works on lists where a list is an ordered collection of values.

But something like this
```js
_.reduce({ 'a': 1, 'b': 2 }, function(result, n, key) {
  result[key] = n * 3;
  return result;
}, {});
```

Could be done like
```js
R.reduce((acc, [val, key]) => R.assoc(val, key, acc), {}, R.toPairs({'a': 1, 'b': 2}), 
```

And that is, IMO, as much saner way to do it.

Note that the above example from the Lodash documentation is actually a really poor example of `_.reduce`. We could achieve the same thing with Ramda like this:

```js
R.mapObj(R.multiply(3), {'a': 1, 'b': 2})
```

gyzerok:
@paldepind, @svozza Ok, thank you for an implementation example. It feels like I'm not the only person who is challenged with this. Mb you can provide docs for such cases?

svozza:
Could go in the [cookbook](https://github.com/ramda/ramda/wiki/Cookbook) I guess.

anton-rudeshko:
Second ago I came here to post exact same issue, so strongly voting up!
Proposed solution is great... but in ES5 looks way much uglier.

anton-rudeshko:
BTW, what do you thing about implementing [`_.transform`](https://lodash.com/docs#transform)?

megawac:
`_.transform` isn't a good canidate for ramda. It was discussed in  <insert issue here>

anton-rudeshko:
The only difference between them is that `return` is automatic in transform.

megawac:
It also detects and choses the return type implicitly

buzzdecafe:
[...]

PR #1601, opened 2016-01-19
`reduce` now works on objects

niloy:
Now that simple JS objects are treated as functors, `map` and `filter` works over it, `reduce` should also work over objects.

### DIFF
--- a/src/internal/_reduce.js
+++ b/src/internal/_reduce.js
@@ -1,6 +1,8 @@
+var _isObject = require('./internal/_isObject');
 var _xwrap = require('./_xwrap');
 var bind = require('../bind');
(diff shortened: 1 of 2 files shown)


buzzdecafe:
:-1: for me. this has come up over and over.

niloy:
Aww okay.

davidchambers:
See #1327 for earlier discussion.

CrossEye:
> See #1327 for earlier discussion.

And #1067.

And #257, #364, #546, #625, #656, #920, and probably others.

Yes, this has come up a few times.

But this reminds me of a PR I started and then didn't follow up on.

<scott wanders off mumbling...>

PR #1603, opened 2016-01-21
`isBetween` implemented

niloy:
```js
R.isBetween(1, 3, 2);     //=> true
R.isBetween(1, 3, 4);     //=> false
```

### DIFF
--- a/src/isBetween.js
+++ b/src/isBetween.js
@@ -1,0 +1,22 @@
+var _curry3 = require('./internal/_curry3');
+
+
+/**
+ * Checks if a given number is between min and max values.
+ *
+ * @func
+ * @memberOf R
+ * @category Relation
+ * @sig Number -> Number -> Number -> Boolean
+ * @param {Number} min The min value.
+ * @param {Number} min The max value.
+ * @param {Number} min The value to be checked.
+ * @return {Boolean}
+ * @example
+ *
+ *      R.isBetween(1, 3, 2);     //=> true
+ *      R.isBetween(1, 3, 4);     //=> false
+ */
+module.exports = _curry3(function isBetween(min, max, num) {
+  return num >= min && num <= max;
+});
--- a/test/isBetween.js
+++ b/test/isBetween.js
@@ -1,0 +1,10 @@
+var R = require('..');
+var eq = require('./shared/eq');
+
+
+describe('isBetween', function() {
+  it('reports if a number is between min and max values', function() {
+    eq(R.isBetween(1, 3, 4), false);
+    eq(R.isBetween(1, 3, 2), true);
+  });
+});


jirutka:
You should explicitly document in the jsdoc that the range is inclusive on both sides.

src/isBetween.js
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Checks if a given number is between min and max values.
+ *
+ * @func
+ * @memberOf R
+ * @category Relation
+ * @sig Number -> Number -> Number -> Boolean
+ * @param {Number} min The min value.
+ * @param {Number} min The max value.
+ * @param {Number} min The value to be checked.
```
CrossEye: Typo in name.  Should be `num` -- although IMHO that's not a great name.  Perhaps `test`

CrossEye:
I'm not sure I really like this.  Why are we privileging 

```js
min ≤ test && test ≤ max
```

over any of these?:

```js
min ≤ test && test < max
min < test && test ≤ max
min < test && test < max
```

davidchambers:
I share Scott's concern.

buzzdecafe:
@CrossEye i agree; seems like this might be better expressed as `(a -> bool) -> (a -> bool) -> a -> bool`. But why this particular pair of constraints? And why just a pair of constraints? We have functions for expressing constraints like this already, e.g. `R.any` or `R.all`

jirutka:
Just my two cents; Ruby privileges the same variant (both ends inclusive) in [Comparable#between?](http://ruby-doc.org/core-2.3.0/Comparable.html#method-i-between-3F). Also the range operator `..` is inclusive on both ends.

jirutka:
This function is indeed trivial to implement; I think that the purpose is to provide simple convenient function for the most common variant.

buzzdecafe:
> This function is indeed trivial to implement; I think that the purpose is to provide simple convenient function for the most common variant.

Understood. IMO the lib is already too large. So we should be very circumspect about adding functions.

jirutka:
> So we should be very circumspect about adding functions.

Totally agree. However, this is function is very tiny, just 86 characters unminimized.

I personally don’t need this function (yet), but I think that it may be useful for many people. ¯\_(ツ)_/¯

CrossEye:
> Just my two cents; Ruby privileges the same variant (both ends inclusive) in [Comparable#between?](http://ruby-doc.org/core-2.3.0/Comparable.html#method-i-between-3F). Also the range operator `..` is inclusive on both ends.

And note that _our_ Range is inclusive in the first parameter, exclusive in the second, which would add some inconsistency to the library.

But I'm also with @buzzdecafe, in asking simply _why bother_?  Is this significant?  The fact that it's so easily expressible, and that it captures a function that doesn't seem all _that_ common makes me doubt it's really worth it.

jirutka:
> And note that our Range is inclusive in the first parameter, exclusive in the second, which would add some inconsistency to the library.

Oh, right, I totally forgot that! Then it’s simple, `isBetween` must be consistent with Range IMO.

niloy:
> I'm not sure I really like this. Why are we privileging

No special reason. Having 4 variants doesn't seem like a good idea, so implemented based on what I thought would be useful most of the times.

> We have functions for expressing constraints like this already, e.g. R.any or R.all

True. I was writing game code where I was frequently writing code like `x >= left && x <= right` for collision detection. And thought it would be nicer to write code like `isBetween(left, right, x)`.

> Maybe it’d be better to use tuple (array of two values) instead of two arguments to define the range;

Good idea :+1: 

> IMO the lib is already too large. So we should be very circumspect about adding functions.

Yes, I understand. Let me know if I should close the PR, its cool :smile:

davidchambers:
The most *consistent* function would be `R.inRange`, but I wouldn't use this often. I think we should stick with lambdas such as `x => x >= min && x <= max`, so there's no doubt as to whether the bounds are included.

CrossEye:
> True. I was writing game code where I was frequently writing code like `x >= left && x <= right` for collision detection. And thought it would be nicer to write code like `isBetween(left, right, x)`.

This is an eminently reasonable function, and I think it should go in the cookbook.  But for the reasons already discussed, it doesn't look as though it belongs in Ramda.

PR #1605, opened 2016-01-22
fix docs

### LINKED ISSUE #1604: R.concat dispatches to the FIRST argument's concat method
Not the second (per docs)

https://github.com/ramda/ramda/blob/v0.19.0/src/concat.js

```javascript
var a = {value: 'abc', concat: function (v) {return this.value + v}};

R.concat(a, 'def');
//=> 'abcdef'

R.concat('def', a);
//=> 'def[object Object]'
```

ecwyne:
Fixes #1604

### DIFF
--- a/src/concat.js
+++ b/src/concat.js
@@ -5,7 +5,7 @@ var invoker = require('./invoker');
 /**
  * Returns the result of concatenating the given lists or strings.
  *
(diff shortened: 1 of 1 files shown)


davidchambers:
:thumbsup: Thanks!

CrossEye:
:herb:

Issue #1554, opened 2015-12-21
consistent type signatures


davidchambers:
Questions to be answered:

  - __How should pairs be represented?__ `[a, b]` and `(a, b)` are leading contenders.

  - __How should binary functions be represented?__ `(a, b) -> c` is one option, though this may conflict with the representation of pairs.

  - __How should "dictionaries" be represented?__ `{a}`, `{k: v}`, and `StrMap a` are possibilities.

  - __How should "records" be represented?__ `{ age :: Number, name :: String }` makes sense when the fields are known, but `Object` might be all we can say for functions such as `R.prop` which operate on any record.

scott-christopher:
My thoughts:
* **How should pairs be represented?**
   * My preference would be for `[a, b]`, as it better reflects that it just represents a two element array. The downside however would be that there could be some confusion in thinking `[a]` then represents a single element array, rather than a list of any size. We could consider something like `[...a]` to represent lists of any size.

* **How should binary functions be represented?**
   * If the binary function should be curried, then `a -> b -> c`, however I'd opt for `(a, b) -> c` in other cases assuming we went with `[a, b]` for pairs. 

* **How should "dictionaries" be represented?**
   * My preference is probably `{a}` here, as `k` is always going to be `String` for native JS objects and the shorthand is nicer in my opinion.

* **How should "records" be represented?**
   * I agree that the keys should be present if the keys help to define the type. I'm somewhat inclined to think we should just represent "records" in the same way as "dictionaries", as `{a}` allows for `a` to be referenced elsewhere in the type signature, even if it just represents a type in the set of all types used for values in the object. For example what would the return type of `prop` look like if `Object` were used in the type signature rather than `{a}`?

svozza:
If we're worried about confusion arising from using brackets for pairs and functions then we might want to do something about the two different ways they're used in the type sig of the compose* functions.

CrossEye:
Another one I forgot to respond to.  (This one I actually saw when it first came in, too.  Shame on me!)

* **How should pairs be represented?**
   * While I prefer `[a, b]`, over `(a, b)`, I do wonder if we might not be still better off by considering `<a, b>`, which wouldn't allow the confusion with `[a]`.

* **How should binary functions be represented?**
   * I agree with @scott-christopher: `a -> b -> c` for curried function, `(a, b) -> c` for non-curried ones.

* **How should "dictionaries" be represented?**
[...]

PR #1609, opened 2016-01-24
Replace * with Any in all type signatures

svozza:
This was discussed in #1554 and seemed to have a reasonable bit of support. I'm happy to do some of the other suggestions made in that issue, like a consistent way to represent dictionaries - as `{a}` - but that seemed to get a more lukewarm reception.

### DIFF
--- a/src/F.js
+++ b/src/F.js
@@ -8,8 +8,8 @@ var always = require('./always');
  * @memberOf R
  * @since v0.9.0
  * @category Function
- * @sig * -> Boolean
- * @param {*}
+ * @sig Any -> Boolean
+ * @param {Any}
  * @return {Boolean}
  * @see R.always, R.T
  * @example
--- a/src/T.js
+++ b/src/T.js
@@ -8,8 +8,8 @@ var always = require('./always');
  * @memberOf R
  * @since v0.9.0
  * @category Function
- * @sig * -> Boolean
- * @param {*}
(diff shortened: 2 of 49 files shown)


buzzdecafe:
LGTM

davidchambers:
:deciduous_tree:

CrossEye:
I think the point that @joneshf make [on Gitter][gi] is pretty relevant.  Many of our uses of `*` should not actually be replaced with `Any` but with a simple type variable.  `Any` is mostly useful for the construct `[Any]` to cover such monstrosities that JS allows as `['abc', 42, false, {}, new Date()]`

For instance, `tap` should just be

```haskell
tap:: (a -> b) -> a -> a
```

Yes, `b` is an unknown type, but it's fixed for a particular invocation of `tap` so it might as well be represented by a simple type variable such as `a`, `b`, `c`, etc.

That the type is irrelevant and simply discarded is evident from the signature.  We don't need a token such as `*` or `Any` to tell us that.

  [gi]: https://gitter.im/ramda/ramda?at=56a5047084cccde9258ae6b3

svozza:
Fair enough, if I can get a steer on what functions will require the `Any` type then I can update the ones that don't with simple type variables.

CrossEye:
Dropped and forgotten ages ago.  This may still be worth doing, but the library has moved too far to use this implementation.

PR #1618, opened 2016-01-30
Make where recursive

asaf-romano:


src/where.js
```diff
@@ -35,11 +36,27 @@ var _has = require('./internal/_has');
  *      pred({a: 'foo', b: 'bar', x: 11, y: 19}); //=> false
  *      pred({a: 'foo', b: 'xxx', x: 10, y: 19}); //=> false
  *      pred({a: 'foo', b: 'xxx', x: 11, y: 20}); //=> false
+ *
+ *      // Recursive behavior
+ *      var pred = R.where({
+ *        a: {
+ *          x: R.equals("foo")
```
davidchambers: Let's `s/"/'/g` to be consistent. :)

src/where.js
```diff
@@ -35,11 +36,27 @@ var _has = require('./internal/_has');
...
+ *        },
+ *        b: R.complement(R.equals('bar'))
+ *      });
+ *
+ *      pred({a: { b: 'foo' }, b: 'xxx'}); //=> true
+ *      pred({a: { b: 'xxx' }, b: 'xxx'}); //=> false
+ *      pred({a: { b: 'foo' }, b: 'bar'}); //=> false
```
davidchambers: Let's stick with `{k: v}` for the formatting of object literals (rather than `{ k: v }`).

CrossEye:
I'd like for @buzzdecafe to weigh in on this one, as I think `where` has a special place in his heart.  If nothing else, it was a focus of the [very first article][ir] on Ramda, which he published nearly two years back.

But I'm happy to see this.  It adds a bit of both power and complexity to this function, but make the library as a whole more consistent.

:herb:


  [ir]: http://buzzdecafe.github.io/code/2014/05/16/introducing-ramda/

buzzdecafe:
i'm happier with `whereEq` and `where` as is. The current definitions are simple, respectively: specify equality constraints as values on object properties; and specify arbitrary constraints as functions on object properties. This was an improvement over the original `where` because it removed the run-time type-checking: `whereEq` is a `String -> value` mapping, `where` is `String -> (Any -> Boolean)`. Simple.

i expect (although i have no data) that those two functions cover the overwhelming majority of uses. i certainly have not  heard any rumbling of discontent that the function are insufficient. 

Cases where recursive descent is desired can be implemented by the user, possibly nesting `where*` within the predicate. I strongly prefer that to adding complexity to the implementation and the overhead of type-checking all the keys of the spec.

CrossEye:
@buzzdecafe:

I suspected that might be your reaction.  I should have also pointed to the motivation in #1592.

Perhaps we should be looking at restricting `evolve` instead.

What do others think of this?  @davidchambers, you're our strongest advocate for library-wide consistency.  How do you think this should play out?  (Or is `evolve` simply off your radar since lenses should cover it anyway?)

davidchambers:
I agree with @buzzdecafe.

Ad hoc polymorphism provides convenience at the cost of simplicity. Adding a special case for one particular data type inevitably leads to difficult questions as to whether other data types deserve special treatment as well. In this case I can imagine someone requesting a special case for Array, so one could write `R.where({address: {street: [R.equals('123 Main Street')]}})`.

PR #1619, opened 2016-01-30
compare Promise objects by identity in R.equals

### LINKED ISSUE #1616: R.equals considers any two promises to be equal
This issue was raised by @zhangchiqing in davidchambers/doctest#70.

```javascript
> R.equals(Promise.resolve(2), Promise.resolve(3))
true
```

Value-based equality may not be applicable to promises, but we could at least fall back to identity.

davidchambers:
Fixes #1616

### DIFF
--- a/src/internal/_equals.js
+++ b/src/internal/_equals.js
@@ -1,4 +1,5 @@
 var _arrayFromIterator = require('./_arrayFromIterator');
+var _functionName = require('./_functionName');
 var _has = require('./_has');
 var identical = require('../identical');
 var keys = require('../keys');
@@ -27,6 +28,10 @@ module.exports = function _equals(a, b, stackA, stackB) {
     case 'Arguments':
     case 'Array':
     case 'Object':
+      if (typeof a.constructor === 'function' &&
+          _functionName(a.constructor) === 'Promise') {
+        return a === b;
+      }
       break;
     case 'Boolean':
     case 'Number':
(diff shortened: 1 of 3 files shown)


src/internal/_functionName.js
```diff
@@ -0,0 +1,5 @@
+module.exports = function _functionName(f) {
+  // String(x => x) evaluates to "x => x", so the pattern may not match.
+  var match = String(f).match(/^function (\w*)/);
+  return match == null ? '' : match[1];
+};
```
davidchambers: This function was added in #1427 then removed in #1516.

src/internal/_equals.js
```diff
@@ -27,6 +28,10 @@ module.exports = function _equals(a, b, stackA, stackB) {
     case 'Arguments':
     case 'Array':
     case 'Object':
+      if (typeof a.constructor === 'function' &&
+          _functionName(a.constructor) === 'Promise') {
```
davidchambers: Unfortunately `Object.prototype.toString.call(Promise.resolve(42))` evaluates to `'[object Object]'` rather than `'[object Promise]'`, so we're forced to do something nasty.

davidchambers:
How does this look, @CrossEye?

CrossEye:
Sorry, missed it earlier.  This looks as good as we're going to get.

:herb:

I still hate Promises.

Issue #1447, opened 2015-10-15
deprecate R.converge


davidchambers:
[`R.lift`][1] covers all my uses of [`R.converge`][2].

```javascript
> R.converge(R.subtract, [R.inc, Math.sqrt])(9)
7
> R.lift(R.subtract)(R.inc, Math.sqrt)(9)
7
```

Is `R.converge` only necessary to support non-unary "transformation" functions?

[1]: http://ramdajs.com/docs/#lift
[2]: http://ramdajs.com/docs/#converge

buzzdecafe:
LGTMt. This still works:

```js
R.converge(R.subtract, [R.multiply, R.add])(4, 3); //=> 5
R.converge(R.subtract, [R.multiply, R.add])(4)(3); //=> 5
R.lift(R.subtract)(R.multiply, R.add)(4, 3); //=> 5
R.lift(R.subtract)(R.multiply, R.add)(4)(3); //=> 5
```

CrossEye:
This looks _fantastic_!  Once stated, it seems obvious, but I've been thinking about this one and `useWith` a lot, and didn't manage to come up with it.

I'd like to spend a little time looking at the various uses I've had for `converge`, but I'm not coming up with any cases in the abstract that this wouldn't cover.

paldepind:
I like this :+1: 

The only issue it that to those unfamiliar with the applicative functor instance for functions it isn't obvious how this works.

davidchambers:
Would you like to open a pull request for this, @benperez?

benperez:
I actually didn't see this until now. `lift` won't respect context or variadic functions for branching functions at the moment. We could special case it in `R.lift`if it's important to keep that behavior. Let me know what you guys think.

davidchambers:
I don't enjoy thinking about variadic functions. ;)

CrossEye:
Variadic functions are always a problem.  But I think we need to make a decision about respecting context.  I've always been of the mind that as long as it's reasonably easy to do and doesn't get in the way of anything else, continuing to respect context was a good practice.

But @asaf-romano pointed out in #1592 that this has been broken for a while.  (I'm not sure, but I'm guessing it was broken during @scott-christophers performance enhancements in @1512.)  We've received no complaints about this at all.  @asaf-romano only noticed it because a line of existing code looked odd.

[...]

PR #1621, opened 2016-01-31
amended docs for lift and liftN 

kwijibo:
Amended docs for lift and liftN to indicate support for functions-as-applicatives via R.ap
see https://github.com/ramda/ramda/issues/1447

I have also changed  *list* to *Array* in the `liftN` docs; this is consistent with the `lift` docs.

### DIFF
--- a/src/lift.js
+++ b/src/lift.js
@@ -3,7 +3,7 @@ var liftN = require('./liftN');
 
 
 /**
- * "lifts" a function of arity > 1 so that it may "map over" an Array or other
+ * "lifts" a function of arity > 1 so that it may "map over" a list, Function or other
  * object that satisfies the [FantasyLand Apply spec](https://github.com/fantasyland/fantasy-land#apply).
  *
  * @func
@@ -13,7 +13,7 @@ var liftN = require('./liftN');
  * @sig (*... -> *) -> ([*]... -> [*])
  * @param {Function} fn The function to lift into higher context
  * @return {Function} The lifted function.
- * @see R.liftN
+ * @see R.liftN, R.lift
  * @example
  *
  *      var madd3 = R.lift(R.curry((a, b, c) => a + b + c));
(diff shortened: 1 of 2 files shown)


buzzdecafe:
LGTM, thanks
:cow2:

CrossEye:
Thank you.

> I have also changed list to Array in the liftN docs; this is consistent with the lift docs.

I would rather go the other way.  We try to use "list" where we can in the documentation, under the idea that lists are our essential type, and that (dense) arrays are simply the best approximation offered by the language.

src/liftN.js
```diff
@@ -8,7 +8,7 @@ var map = require('./map');
 
 /**
  * "lifts" a function to be the specified arity, so that it may "map over" that
- * many lists (or other objects that satisfies the [FantasyLand Apply spec](https://github.com/fantasyland/fantasy-land#apply)).
+ * many Arrays, Functions (or other objects that satisfies the [FantasyLand Apply spec](https://github.com/fantasyland/fantasy-land#apply)).
```
svozza: s/satisfies/satisfy

src/lift.js
```diff
@@ -14,6 +14,7 @@ var liftN = require('./liftN');
  * @param {Function} fn The function to lift into higher context
  * @return {Function} The lifted function.
  * @see R.liftN
+ * @see R.ap
```
davidchambers: Let's write these on one line: `@see R.liftN, R.ap`.

kwijibo:
All good points, thanks everyone. @CrossEye I mistakenly thought `lift` would only work on Arrays, but I see it works on other list-like things  (eg `arguments` and strings), so now both docs use `list`

kwijibo:
(It would also be nice if the docs could define the terminology, ie, `List` and whatever else; I'm not sure where this would go though. Perhaps a section at the top could briefly explain the cornerstones of Ramda's philosophy - Lists, Fantasy-land support, type signatures, currying, etc )

kwijibo:
(squashed the commits)

CrossEye:
> It would also be nice if the docs could define the terminology, ie, List and whatever else

Yes, we've been talking about that for a while.  No one has stepped forward to do it, though.

PR #1625, opened 2016-02-03
remove functions deprecated in v0.19.0

davidchambers:


buzzdecafe:
i still think these are better, more descriptive names than `sequence` and `traverse`. But what are you gonna do. :cow2:

davidchambers:
I was happy with the old names, but it's nice to have `R.sequence` :articulated_lorry: to `sequence` and `R.traverse` :articulated_lorry: to `traverse`.

buzzdecafe:
yep, i agree. we can just point people to the Haskell docs

CrossEye:
Yes, I definitely preferred `commute`, but complying with standards is generally a good thing.

PR #1632, opened 2016-02-06
fix typo in `dropRepeatsWith`'s documentation

benjycui:
It seems that it is a typo, but I am not sure :-)

### DIFF
--- a/src/dropRepeatsWith.js
+++ b/src/dropRepeatsWith.js
@@ -17,7 +17,7 @@ var last = require('./last');
  * @memberOf R
  * @since v0.14.0
  * @category List
- * @sig (a, a -> Boolean) -> [a] -> [a]
+ * @sig (a -> a -> Boolean) -> [a] -> [a]
  * @param {Function} pred A predicate used to test whether two items are equal.
  * @param {Array} list The array to consider.
  * @return {Array} `list` without repeating elements.


CrossEye:
Actually, I don't think this is a typo.

I'm not sure we've got a great syntax for this yet (see #1554) but the predicate supplied here is never going to be treated as a curried function, even if you supply one that happens to be curried.  It will always be called with two parameters.

buzzdecafe:
if it's not a typo, it's still not correct:

`@sig (a, a -> Boolean) -> [a] -> [a]`

should be

`@sig ((a, a) -> Boolean) -> [a] -> [a]`

CrossEye:
That's why I brought up #1554.  There is some debate about how we should represent pairs/tuples in these signatures.  But certainly we need something to group them, or there's no way to distinguish between `((a, a) -> Boolean)` and `(a, (a -> Boolean))`.

CrossEye:
But agreed we should fix this, and probably to the version @buzzdecafe suggests.

niloy:
`a * a` is used to represent tuple in F#.

https://en.wikipedia.org/wiki/Product_type

buzzdecafe:
> a * a is used to represent tuple in F#.

right, that traces its ancestry to ML. Unfortunately, I don't think it fits in this case, because we are just trying to indicate that the function takes two arguments, but may or may not be curried. And it does not *really* take a tuple for an argument.

PR #1639, opened 2016-02-13
Move Set logic in uniqBy into custom _Set type

scott-christopher:
This change moves the logic for handling the equality semantics from within `R.uniqBy` into a separate `internal/_Set` type, cleaning up the internals of `R.uniqBy` while making something of a reusable Set type that honours `R.equals` semantics and can also be used in environments where a native `Set` doesn't exist.

Performance gets a nice little boost on earlier versions on Node too :wink:
```
> process.version
'v0.10.36'
> a = R.range(0, 1e6);
> c.time(''); _.uniq(a); c.timeEnd('');
: 156ms
> c.time(''); R.uniq(a); c.timeEnd('');
: 235ms

> process.version
'v5.5.0'
> a = R.range(0, 1e6);
> c.time(''); _.uniq(a); c.timeEnd('');
: 315.964ms
> c.time(''); R.uniq(a); c.timeEnd('');
: 250.782ms
```

### DIFF
--- a/src/internal/_Set.js
+++ b/src/internal/_Set.js
@@ -1,0 +1,162 @@
+var _contains = require('./_contains');
+
+
+// A simple Set type that honours R.equals semantics
+module.exports = (function() {
+  function _Set() {
+    /* globals Set */
+    this._nativeSet = typeof Set === 'function' ? new Set() : null;
+    this._items = {};
+  }
+
+  _Set.prototype.add = function(item) {
+    return hasOrAdd(item, true, this);
+  };
+
+  _Set.prototype.has = function(item) {
+    return hasOrAdd(item, false, this);
+  };
+
+  /**
+   * Combines the logic for checking whether an item is a member of the set and
+   * for adding a new item to the set.
+   *
+   * @param item       The item to check or add to the Set instance.
+   * @param shouldAdd  If true, the item will be added to the set if it doesn't
+   *                   already exist.
+   * @param set        The set instance to check or add to.
+   * @return {boolean} When shouldAdd is true, this will return true when a new
+   *                   item was added otherwise false. When shouldAdd is false,
+   *                   this will return true if the item already exists, otherwise
+   *                   false.
+   */
+  function hasOrAdd(item, shouldAdd, set) {
+    var type = typeof item;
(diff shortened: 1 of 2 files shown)


buzzdecafe:
nice. :cow2:

svozza:
Impressive!

src/internal/_Set.js
```diff
@@ -0,0 +1,167 @@
+var _contains = require('./_contains');
+
+
+// A simple Set type that honours R.equals semantics
+module.exports = (function() {
+  function _Set() {
+    /* globals Set */
+    this._nativeSet = typeof Set === 'undefined' ? null : new Set();
```
davidchambers: Is there a reason to favour this negatory assertion over `typeof Set === 'function'`?

src/internal/_Set.js
```diff
@@ -0,0 +1,167 @@
...
+    this._items = {};
+    this._negZeroExists = false;
+    this._nullExists = false;
+    this._undefExists = false;
+    this._falseExists = false;
+    this._trueExists = false;
```
davidchambers: Why is it necessary to treat `true` and `false` specially?

CrossEye:
Very nice!

src/internal/_Set.js
scott-christopher: Not necessary for native Sets. The same would have to exist for environments that don't support native Sets though, so I figured we could reuse it to prevent an extra lookup in the native Set.

src/internal/_Set.js
scott-christopher: Nope, good point. :zap:

src/internal/_Set.js
davidchambers: > The same would have to exist for environments that don't support native Sets

Couldn't we simply have an `'[object Boolean]'` bucket?

scott-christopher: Yeah now that we have an `_items` bucket, we might as well store the others in there too – updated :zap:

davidchambers: Much cleaner! :thumbsup:

davidchambers:
:deciduous_tree:

PR #1648, opened 2016-02-18
Defaulting r so that there is not a failure if r is undefined.

ExploreMqt:


davidchambers:
I'm opposed to this change for several reasons:

  - It's the caller's responsibility to provide arguments of the correct types. Having functions guess what the caller may have intended when a type error is encountered leads to a system of poorly defined functions which swallow errors.

  - Why treat the second argument differently from the first argument?

  - Should the many other functions which accept objects as arguments also treat `undefined` as `{}`?

  - If we *are* to treat some values specially, which values would we choose? Not `0`, `-0`, `null`, `false`, `NaN`, `undefined`, and `''` as we do in this pull request (likely unintentionally due to the use of `||`). I consider treating `null` and `undefined` differently poor API design, so I'd suggest only treating `null` and `undefined` specially.

ExploreMqt:
I most often use merge to deal with dependency injection. Therefore, I have designed a system that intentionally has undefined as the second parameter in a production situation.
Maybe I'm wrong but this seems to be a "falling into a pit of success" kind of topic.

Why is the second argument different? Well for starters, there is no failure if you pass in undefined for the first parameter! I can certainly add tests for that case, but ad-hoc testing determined that for me before I decided to submit a pull request. 

You are the managers of the library and determining the consistency of behavior is your job and I'm glad you care. I will not get into those discussions.

davidchambers:
> Why is the second argument different? Well for starters, there is no failure if you pass in undefined for the first parameter!

D'oh! This is unintentional, I'm sure. This being the case, though, your proposal arguably makes the function *more* consistent, so I could get behind this change. My strong preference, of course, would be for Ramda to perform type checking to avoid such inconsistencies throughout the library, but this has been discussed elsewhere.

customcommander:
@CrossEye I think #3222 is somewhat related to this pull request. Shall we close this one?

PR #1655, opened 2016-02-20
Implement partition in terms of filter and reject

### LINKED ISSUE #1651: Should R.partition work for objects?
Now that R.filter and R.reject work with objects, it might make sense to make package-deal R.partition (in contrast to R.splitWhen, btw) to do so as well.

asaf-romano:
Closes #1651

### DIFF
--- a/src/partition.js
+++ b/src/partition.js
@@ -1,30 +1,29 @@
-var _curry2 = require('./internal/_curry2');
-var _reduce = require('./internal/_reduce');
+var filter = require('./filter');
+var juxt = require('./juxt');
+var reject = require('./reject');
 
 
 /**
- * Takes a predicate and a list and returns the pair of lists of elements which
- * do and do not satisfy the predicate, respectively.
+ * Takes a predicate and a list or other "filterable" object and returns the
+ * pair of filterable objects of the same type of elements which do and do not
+ * satisfy, the predicate, respectively.
  *
  * @func
  * @memberOf R
  * @since v0.1.4
  * @category List
- * @sig (a -> Boolean) -> [a] -> [[a],[a]]
- * @param {Function} pred A predicate to determine which array the element belongs to.
- * @param {Array} list The array to partition.
- * @return {Array} A nested array, containing first an array of elements that satisfied the predicate,
- *         and second an array of elements that did not satisfy.
+ * @sig Filterable f => (a -> Boolean) -> f a -> [f a, f a]
+ * @param {Function} pred A predicate to determine which side the element belongs to.
+ * @param {Array} filterable the list (or other filterable) to partition.
+ * @return {Array} An array, containing first the subset of elements that satisfy the
+ *         predicate, and second the subset of elements that do not satisfy.
  * @see R.filter, R.reject
  * @example
  *
  *      R.partition(R.contains('s'), ['sss', 'ttt', 'foo', 'bars']);
- *      //=> [ [ 'sss', 'bars' ],  [ 'ttt', 'foo' ] ]
+ *      // => [ [ 'sss', 'bars' ],  [ 'ttt', 'foo' ] ]
+ *
+ *      R.partition(R.contains('s'), { a: 'sss', b: 'ttt', foo: 'bars' });
+ *      // => [ { a: 'sss', foo: 'bars' }, { b: 'ttt' }  ]
  */
-module.exports = _curry2(function partition(pred, list) {
-  return _reduce(function(acc, elt) {
-    var xs = acc[pred(elt) ? 0 : 1];
-    xs[xs.length] = elt;
-    return acc;
-  }, [[], []], list);
-});
+module.exports = juxt([filter, reject]);
--- a/test/partition.js
+++ b/test/partition.js
@@ -1,3 +1,5 @@
+var S = require('sanctuary');
+
 var R = require('..');
 var eq = require('./shared/eq');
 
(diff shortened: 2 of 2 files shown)


src/partition.js
```diff
@@ -1,30 +1,29 @@
...
  * @func
  * @memberOf R
  * @since v0.1.4
  * @category List
- * @sig (a -> Boolean) -> [a] -> [[a],[a]]
- * @param {Function} pred A predicate to determine which array the element belongs to.
- * @param {Array} list The array to partition.
- * @return {Array} A nested array, containing first an array of elements that satisfied the predicate,
- *         and second an array of elements that did not satisfy.
+ * @sig Filterable f => (a -> Boolean) -> f -> [f,f]
```
buzzdecafe: should have a space after the comma? this is more @davidchambers 's domain

buzzdecafe:
> 'works with other filterables'

that's what makes this worthwhile IMO. apart from one teensy formatting thing, :cow2:

src/partition.js
asaf-romano: These signatures are somewhat hard for me to follow (I admittedly just adjusted the current signature to the new ``filter``), so I'll wait for David's input.

CrossEye: I would have thought

```js
Filterable f => (a -> Boolean) -> f a -> [f a, f a]
```

When changing from a List of type `a` (`[a]`) to a container type matching type `Foo`, I think we would always write `Foo f =>` and  `f a`.  But I may still be confused.

CrossEye:
:herb:, modulo the signature decision.

src/partition.js
davidchambers: We certainly want `f a`, as Scott suggested. It may be helpful to work backwards from a concrete type:

```haskell
(Number -> Boolean) -> Array Number -> Pair (Array Number) (Array Number)
```

Replace `Number` with `a`:

```haskell
(a -> Boolean) -> Array a -> Pair (Array a) (Array a)
```

Replace `Array` with `Filterable f => f`:

```haskell
Filterable f => (a -> Boolean) -> f a -> Pair (f a) (f a)
```

The remaining question is how to represent a pair. I favour `Pair a b` as it's unambiguous, whereas `(a, b)` collides with our representation of binary functions and `[a, b]` reuses our array notation in a conflicting manner. I'm okay with `[f a, f a]`, though.

src/partition.js
```diff
@@ -1,30 +1,29 @@
...
+ * @param {Function} pred A predicate to determine which side the element belongs to.
+ * @param {Array} filterable the list (or other filterable) to partition.
```
davidchambers: `{Array}` is no longer accurate.

asaf-romano:
As long as ``pair`` claims to return an array I'll follow that.

src/partition.js
CrossEye: I don't think `Pair` makes much sense here, as we will always actually return a 2-element array from this, correct?

davidchambers: > I don't think `Pair` makes much sense here, as we will always actually return a 2-element array from this, correct?

I'm suggesting `Pair` could be defined as exactly that. In sanctuary-def terms:

```javascript
//    Pair :: Type
const Pair = $.BinaryType(
  'ramda/Pair',
  x => R.type(x) === 'Array' && x.length === 2,
  R.nth(0),
  R.nth(1)
);

['foo', 42] :: Pair String Number
```

davidchambers:
> I thought {Array} to help IDEs code analysis for the primary usage of these functions.

Seems reasonable. Either we're too restrictive (`{Array}`) or too vague (`{*}`). On balance, `{Array}` seems the better option.

src/partition.js
```diff
@@ -1,30 +1,29 @@
...
- * @sig (a -> Boolean) -> [a] -> [[a],[a]]
- * @param {Function} pred A predicate to determine which array the element belongs to.
- * @param {Array} list The array to partition.
- * @return {Array} A nested array, containing first an array of elements that satisfied the predicate,
- *         and second an array of elements that did not satisfy.
+ * @sig Filterable f => (a -> Boolean) -> f a -> [f a, f a]
+ * @param {Function} pred A predicate to determine which side the element belongs to.
+ * @param {Array} filterable the list (or other filterable) to partition.
+ * @return {Array} An array, containing first the subset of elements that satisfied the
+ *         predicate, and second the subset of elements that did not satisfy.
```
davidchambers: Let's use "satisfy"/"do not satisfy" rather than "satisfied"/"did not satisfy". Since we're programming with values, we don't expect the result to change over time. :)

src/partition.js
```diff
@@ -1,30 +1,29 @@
...
+ *      // => [ { a: 'sss', foo: 'bars' }, { b: 'ttt' }  ]
  */
-module.exports = _curry2(function partition(pred, list) {
-  return _reduce(function(acc, elt) {
-    var xs = acc[pred(elt) ? 0 : 1];
-    xs[xs.length] = elt;
-    return acc;
-  }, [[], []], list);
-});
+module.exports = juxt([filter, reject]);
```
davidchambers: I :heart: this definition. :)

CrossEye: And given the current definition of `juxt` as essentially `converge(Array.of)`, we're well on our way to elegance.

...

But I am nervous still, as Ramda started as a rewrite of some really elegant code which fundamentally could not perform.  This was not an abstract, perhaps-some-day-it-might-be-an-issue sort of problem but a serious problem that meant that we couldn't use the predecessor for serious work.  Of course that meant that we did manage to [throw away the first version][ta], perhaps a good thing.


  [ta]: http://c2.com/cgi/wiki?PlanToThrowOneAway

davidchambers: Traversing the list twice rather than once is unlikely to be an issue. It's the O(n^2) stuff we really need to avoid. :)

PR #1657, opened 2016-02-20
Optimize R.into to perform mutable reductions

### LINKED ISSUE #1656: in-house transducer accumulators should use mutable reductions
Last week I tired to optimize some slow, heavy multi-map-inversion-kind-of-code to use transducers, only to find out they made the situation even worse. Curios, I went ahead and read the implementation, and was surprised to see that the in-house transducer accumulators ([] and {}) do not perform what's, in Java 8's lingo, called "mutable reductions" (https://docs.oracle.com/javase/8/docs/api/java/util/stream/package-summary.html#MutableReduction ; apart the combiner element and the naming conventions, Java's Collector is Clojure's transducer). Rather, they clone the result object *every step of the way* (see https://github.com/ramda/ramda/blob/master/src/internal/_stepCat.js). I don't know what Cloujre is doing in their analogous APIs (comparing to java's mutable reductions, or to Ramda's implementation of transducers), but regardless, it seems very strange for the "default" implementation of into([]) and into({}) not to take advantage of the functionality provided by transducers themselves, esp. given that all those intermediate objects are not visible to the consumer.

Semi-benchmark:
http://goo.gl/xn24gq

asaf-romano:
Closes #1656

### DIFF
--- a/src/internal/_clone.js
+++ b/src/internal/_clone.js
@@ -9,9 +9,10 @@ var type = require('../type');
  * @param {*} value The value to be copied
  * @param {Array} refFrom Array containing the source references
  * @param {Array} refTo Array containing the copied source references
+ * @param {Boolean} deep Whether or not to perform deep cloning.
  * @return {*} The copied value.
  */
-module.exports = function _clone(value, refFrom, refTo) {
+module.exports = function _clone(value, refFrom, refTo, deep) {
   var copy = function copy(copiedValue) {
     var len = refFrom.length;
     var idx = 0;
@@ -24,7 +25,8 @@ module.exports = function _clone(value, refFrom, refTo) {
     refFrom[idx + 1] = value;
     refTo[idx + 1] = copiedValue;
     for (var key in value) {
-      copiedValue[key] = _clone(value[key], refFrom, refTo);
+      copiedValue[key] = deep ?
+        _clone(value[key], refFrom, refTo, true) : value[key];
     }
     return copiedValue;
   };
--- a/src/internal/_objectAssign.js
+++ b/src/internal/_objectAssign.js
@@ -1,0 +1,24 @@
+var _has = require('./_has');
+
+// Based on https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/assign
+module.exports = function _objectAssign(target) {
+  if (target == null) {
+    throw new TypeError('Cannot convert undefined or null to object');
+  }
+
+  var output = Object(target);
+  var idx = 1;
+  var length = arguments.length;
+  while (idx < length) {
+    var source = arguments[idx];
+    if (source != null) {
+      for (var nextKey in source) {
+        if (_has(nextKey, source)) {
+          output[nextKey] = source[nextKey];
+        }
+      }
+    }
+    idx += 1;
+  }
+  return output;
+};
--- a/src/internal/_stepCat.js
+++ b/src/internal/_stepCat.js
@@ -1,15 +1,17 @@
-var _concat = require('./_concat');
 var _identity = require('./_identity');
 var _isTransformer = require('./_isTransformer');
+var _objectAssign = require('./_objectAssign');
 var isArrayLike = require('../isArrayLike');
-var merge = require('../merge');
 var objOf = require('../objOf');
 
 
 module.exports = (function() {
   var _stepCatArray = {
     '@@transducer/init': Array,
-    '@@transducer/step': function(xs, x) { return _concat(xs, [x]); },
+    '@@transducer/step': function(xs, x) {
+      xs.push(x);
+      return xs;
+    },
     '@@transducer/result': _identity
   };
   var _stepCatString = {
@@ -20,7 +22,7 @@ module.exports = (function() {
   var _stepCatObject = {
     '@@transducer/init': Object,
     '@@transducer/step': function(result, input) {
-      return merge(
+      return _objectAssign(
         result,
         isArrayLike(input) ? objOf(input[0], input[1]) : input
       );
--- a/src/clone.js
+++ b/src/clone.js
@@ -25,5 +25,5 @@ var _curry1 = require('./internal/_curry1');
 module.exports = _curry1(function clone(value) {
   return value != null && typeof value.clone === 'function' ?
     value.clone() :
-    _clone(value, [], []);
+    _clone(value, [], [], true);
 });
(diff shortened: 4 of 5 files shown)


CrossEye:
I'm hoping for reviews from @kevinbeaty, @kedashoe, and @buzzdecafe, who probably know the most about our Transducer implementation.

I'm wondering if `into` is enough.  Is this just the start?

src/internal/_clone.js
```diff
@@ -9,9 +9,10 @@ var type = require('../type');
  * @param {*} value The value to be copied
  * @param {Array} refFrom Array containing the source references
  * @param {Array} refTo Array containing the copied source references
+ * @param {Boolean} deep Whether or not to perform deep cloning.
  * @return {*} The copied value.
```
CrossEye: Not related to this PR, but why in the world do we have JSDoc tags on an internal function?

asaf-romano:
Nothing else is using _stepCat, which is, as far as I know, the sole producer of these "basic" transducers.

src/internal/_objectAssign.js
```diff
@@ -0,0 +1,24 @@
+var _has = require('./_has');
+
+// Based on https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/assign
+module.exports = function _objectAssign(target) {
+  if (target === undefined || target === null) {
+    throw new TypeError('Cannot convert undefined or null to object');
+  }
```
CrossEye: As this is entirely internal, I don't think we need this guard.

asaf-romano:
We can talk elsewhere about adding utility like Collector.of, which would make transducers much more approachable:
https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collector.html#of-java.util.function.Supplier-java.util.function.BiConsumer-java.util.function.BinaryOperator-java.util.function.Function-java.util.stream.Collector.Characteristics...-

(Ignore combiner and characteristics)

src/internal/_clone.js
asaf-romano: The implementation is not so trivial so I'm happy there is some documentation there.

src/internal/_objectAssign.js
asaf-romano: Thought you'd say that :) I left it in because that's the exception native Object.assign throws. I can leave it out of course.

src/internal/_stepCat.js
```diff
@@ -20,7 +22,7 @@ module.exports = (function() {
   var _stepCatObject = {
     '@@transducer/init': Object,
     '@@transducer/step': function(result, input) {
-      return merge(
+      return _objectAssign(
```
CrossEye: What's the advantage of `objectAssign` over `merge`?

src/internal/_clone.js
CrossEye: No real objection.  It's just odd.  We don't usually do that in Ramda.

src/internal/_stepCat.js
asaf-romano: merge clones every time, which is the root cause of this issue. _objectAssign mutates the first argument.

CrossEye: Oh, of course.  Interestingly, way back in Ramda's prehistory, the predecessor to `merge` was built on top of an internal version which mutated the first argument, to which we simply supplied `{}` in the public API.  Ah well, _plus ça change..._.

asaf-romano: Prehistory as in Eweda? I'm sure there was also some shallow cloning utility at some point...

Funny enough, I was about to change ``merge`` to use ``_objectAssign`` before I posted this patch. Then I saw there are few levels of abstractions there and it's not worth it. On the other hand, it's probably a good idea to that in ``mergeAll``.

asaf-romano: Oh boy http://goo.gl/2cFsoO

CrossEye: > Prehistory as in Eweda?

Prehistory as in I'm too lazy to dig up when it changed.  :smiling_imp: 

> Oh boy http://goo.gl/2cFsoO

That's interesting.  We'd have to look to see if there's significant difference in API, but if not, it looks like there's some real optimization possible...

src/internal/_objectAssign.js
```diff
@@ -0,0 +1,24 @@
...
+
+  var output = Object(target);
+  var idx = 1;
+  var length = arguments.length;
+  while (idx < length) {
+    var source = arguments[idx];
+    if (source !== undefined && source !== null) {
```
davidchambers: `source != null` handles both cases. :)

asaf-romano: I'll change that, but note that polyfills tend to explicitly favor this kind of verbosity.

davidchambers: I mentally expand `x != null` into `x !== null && x !== undefined` so to me, both approaches are equally clear. I don't feel strongly about this, though.

kedashoe:
:+1: looks great!

kevinbeaty:
Did not review in depth, but looks good to me.  Funny enough, I think we were using `push` in early implementations but that must have been changed as part of review.

> I'm wondering if into is enough. Is this just the start?

You'll also want to look at `reduce` and `transduce`.   May want to shallow `_clone` the `acc` there as well.

asaf-romano:
I shallow-clone in ``into`` because ``_stepCat`` cannot do it anymore. ``transduce`` on the other hand doesn't clone at all: it just calls ``_reduce`` with the supplied accumulator and performs no magic around the accumulator itself. It's the caller's job, in the case of ``transduce``, to supply the reducer function, and this also means it is responsible for choosing between mutable or immutable reduction. The same applies to ``reduce``.

kevinbeaty:
Works for me.  Thanks for the explanation.

CrossEye:
:herb:

buzzdecafe:
LGTM too :cow2:
@davidchambers if your notes are satisfied, please go ahead and merge

asaf-romano:
I'll fix the != null issue, as I already diverged from the polyfill on code style (for -> while), but let's keep the exception here, because that's the expected behavior of Object.assign.

src/internal/_objectAssign.js
```diff
@@ -0,0 +1,24 @@
+var _has = require('./_has');
+
+// Based on https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/assign
+module.exports = function _objectAssign(target) {
+  if (target === undefined || target === null) {
```
davidchambers: Let's use `target == null` here. :)

davidchambers:
:deciduous_tree:

Issue #1667, opened 2016-02-25
Should zipObj truncate if values array is shorter than keys array?


davezuko:
Looking at the documentation for `zip` and `zipWith`:

> Creates a new list out of the two supplied by pairing up equally-positioned items from both lists. The returned list is truncated to the length of the shorter of the two input lists. Note: zip is equivalent to zipWith(function(a, b) { return [a, b] }).

However, `zipObj` will happily implicitly set keys with values of `undefined` if the keys array is longer than the values array. Example:

```js
const _keys = ['foo', 'bar', 'baz']
const _vals = [1]

const _zipped = R.zipObj(_keys, _vals)

// produces
{
  foo: 1,
  bar: undefined,
  baz: undefined,
}
```

I was going to submit a PR for this but wanted to verify because you may very well have a good reason for this that I'm just not aware of. If it is in fact unintentional, I will happily submit a followup PR.

CrossEye:
Agreed.  It seems that, however it's implemented, this identity should hold: 

```js
 zipObj(keys, vals) ≍ pipe(zipWith(pair), fromPairs)(keys, vals)
```

Do you feel like creating a PR for this?

davezuko:
Awesome, glad to see I'm not crazy. Yes, I will submit a PR today for this, thanks!

PR #1669, opened 2016-02-26
update zipObj to truncate to shorter list

davezuko:
Resolves https://github.com/ramda/ramda/issues/1667

### DIFF
--- a/src/zipObj.js
+++ b/src/zipObj.js
@@ -3,6 +3,8 @@ var _curry2 = require('./internal/_curry2');
 
 /**
  * Creates a new object out of a list of keys and a list of values.
+ * Key/value pairing is truncated to the length of the shorter of the two lists.
+ * Note: `zipObj` is equivalent to `pipe(zipWith(pair), fromPairs)`.
  *
  * @func
  * @memberOf R
@@ -18,7 +20,7 @@ var _curry2 = require('./internal/_curry2');
  */
 module.exports = _curry2(function zipObj(keys, values) {
   var idx = 0;
-  var len = keys.length;
+  var len = Math.min(keys.length, values.length);
   var out = {};
   while (idx < len) {
     out[keys[idx]] = values[idx];
--- a/test/zipObj.js
+++ b/test/zipObj.js
@@ -11,9 +11,8 @@ describe('zipObj', function() {
     eq(R.zipObj(['a', 'b', 'c'], [1, 2, 3, 4, 5, 6, 7]), {a: 1, b: 2, c: 3});
   });
 
-  it('extra keys are undefined', function() {
-    eq(R.zipObj(['a', 'b', 'c', 'd', 'e', 'f'], [1, 2, 3]),
-       {a: 1, b: 2, c: 3, d: undefined, e: undefined, f: undefined});
+  it('ignores extra keys', function() {
+    eq(R.zipObj(['a', 'b', 'c', 'd', 'e', 'f'], [1, 2, 3]), {a: 1, b: 2, c: 3});
   });
 
   it('last one in wins when there are duplicate keys', function() {


test/zipObj.js
```diff
@@ -11,9 +11,8 @@ describe('zipObj', function() {
     eq(R.zipObj(['a', 'b', 'c'], [1, 2, 3, 4, 5, 6, 7]), {a: 1, b: 2, c: 3});
   });
 
-  it('extra keys are undefined', function() {
```
davezuko: @CrossEye noticed this test here, which implies that this functionality was intended. Just want to point it out in case I'm jumping the gun.

src/zipObj.js
```diff
@@ -2,7 +2,10 @@ var _curry2 = require('./internal/_curry2');
 
 
 /**
- * Creates a new object out of a list of keys and a list of values.
+ * Creates a new object out of a list of keys and a list of values. The
+ * returned list is truncated to the length of the shorter of the two input
```
davezuko: Minor doc issue, will rebase in a second to fix.

src/zipObj.js
```diff
@@ -18,7 +20,7 @@ var _curry2 = require('./internal/_curry2');
  */
 module.exports = _curry2(function zipObj(keys, values) {
   var idx = 0;
-  var len = keys.length;
+  var len = Math.min(keys.length, values.length);
```
davezuko: https://github.com/ramda/ramda/blob/master/src/zipWith.js#L30

test/zipObj.js
CrossEye: That means that we should leave it open a bit longer for discussion, and maybe dig a little into the history and talk to whoever implemented it.  I don't remember.  For all I know it was me.

I definitely prefer the altered behavior, but perhaps there was a good reason for the original one.  I'll try to look tonight.

CrossEye: @buzzdecafe: You implemented this in #200.

Was there a particular reason for supporting `undefined` values, or did this simply fall out of the implementation?

davezuko: Sounds reasonable, thanks. I started digging through the history of the function and it looks like it's _always_ been as it is currently (pre-fix), for what that's worth.

CrossEye: Yes.  And I don't remember any particular discussion.  Obviously the function is useful, but I don't recall if there was anything in particular that drove us to add it.  I am guessing that the current behavior fell out of the implementation rather than from considered design, but I could be entirely wrong.

buzzdecafe: > Was there a particular reason for supporting undefined values, or did this simply fall out of the implementation?

I don't recall. So it may be "Development driven tasting". 

Some things bug me about this:

* if were to implement a safe version of this, it would guarantee a value for each key; the value would simply be wrapped in a `Maybe`. 
* i'm afraid of generating different inconsistency w/r/t `for-in` tests with this implementation. But as we've discussed before, what exactly is the difference between a key that isn't there and a key that is mapped to `undefined`?

davidchambers:
:deciduous_tree: This change looks great to me!

CrossEye:
@buzzdecafe:

> Some things bug me about this:
> 
> * if were to implement a safe version of this, it would guarantee a value for each key; the value would simply be wrapped in a `Maybe`. 

I'm not so sure.  Would we want to do something similar with a safe version of `zip` or `zipWith`?  And if not, why the discrepancy?  I like the idea of treating the keys and values the same way in all of them.

> * i'm afraid of generating different inconsistency w/r/t `for-in` tests with this implementation. 

I don't follow.

> But as we've discussed before, what exactly is the difference between a key that isn't there and a key that is mapped to `undefined`?

I was assuming that we'd do as @davezuko has done and simply use the `length` as our guide.  So 

```js
zipObj(['a', 'b', 'c'], [1, 2]); //=> {a: 1, b: 2}
// but
zipObj(['a', 'b', 'c'], [1, 2, undefined]); //=> {a: 1, b: 2, c: undefined}
```

There would be problems with sparse arrays, especially in the keys, but as always with those, _caveat emptor_.

davezuko:
Not that I have too much to add, but the rationale for the fix was simply to enforce consistency with both `zip` and `zipWith`. If there is a larger discussion to be had around how to treat lists with mismatched lengths then I'll leave that to the experts (you guys), but it just struck me as strange that `zipObj` treated the array pairing differently. I especially appreciated @CrossEye 's example that `zipObj` should essentially be sugar for ` pipe(zipWith(pair), fromPairs)(keys, vals)`.

asaf-romano:
In any event I don't think zipObj should have different semantics. 🌷

Trying to track down what might have have happened here, though, I checked what lodash does. Indeed:
```js
require("lodash").zipObject(["a", "b", "c"], [1,2]);
// => { a: 1, b: 2, c: undefined }
```
But it's not just zipObj(ect); this is how normal zip behaves in lodash:

```js
require("lodash").zip([1,2,3], ["a", "b"]);
// => [ [ 1, 'a' ], [ 2, 'b' ], [ 3, undefined ] ]
```
This is different from Haskell's zip, of course, which takes the shorter list.

So R.zip follows Haskell's zip, whilst R.zipObj follows lodah.

Issue #1481, opened 2015-10-30
Bound function equality


ccorcos:
I think it would be really sweet if all partially applied functions in Ramda returned bound functions with a .equals method that allows for a deep comparison between functions that are effectively the same. Additionally, I propose `R.bind` to help with this as well.

```coffee
R = require 'ramda'

bind = (f, args=[]) ->
  bound = (moreArgs...) -> f.apply(null, R.concat(args, moreArgs))
  bound.func = (f.func or f)
  bound.args = R.concat((f.args or []), args)
  bound.equals = (g) ->
    R.equals((g?.args or []), bound.args) and R.equals((g?.func or g), bound.func)
  return bound

add = (a, b) -> a + b
_add = bind add, []

_add1 = bind add, [1]
add1 = bind add, [1]

three = bind add1, [2]
_three = bind _add1, [2]

console.log add1.equals(_add1), R.equals(add1, _add1), _add1.equals(add1), R.equals(_add1, add1), add1(10) is _add1(10)
console.log three.equals(_three), R.equals(three, _three)
console.log _add.equals(add), R.equals(add, _add), R.equals(_add, add)
```

The main reason I want to do this is so I can implement more effective memoization. Often times, I have to map over a list of items, bind a function to remove that item with the item id, and pass that function on to a view component (React, for example). The view has to recompute every time because it always things the remove function is different even thought it might be the same function with the same bound arguments.

jdalton:
Kind of sounds like you should create a memoized copy of `bind` instead.

ccorcos:
Ha! Thats an interesting idea...

Imagine you have some kind of feed. You'll have different items with different id's. You can bind a function to that id (and memoize the result) and pass it down. But now you need to worry about invalidating the memoized cache. Otherwise, over time it will explode. So maybe you can use an LRU cache with a well thought out limit. But this is an imperfect solution.

ccorcos:
Wouldnt it be sweet if:

```
R.equals(R.assoc('a': 2), R.assoc('a': 2))
// => true
```

jdalton:
Naw, because those methods can be extended augmented and sharing them seems odd. I think it's better to simply memoize the methods you're wanting to use.

davidchambers:
> Wouldnt it be sweet if:
>
> ```javascript
> R.equals(R.assoc('a', 2), R.assoc('a', 2))
> // => true
> ```

[...]

Issue #1650, opened 2016-02-19
Functors, Applicative Functors, and Monads support in Ramda - documentation regarding current support?


arcseldon:
This week I have been swatting up on Haskell, and believe I have a pretty good grasp on the fundamentals constituting the fantasy land types - certainly as they relate to Haskell (i think!). However, I am now trying to relate what I now understand in Haskell to what has been implemented in Ramda in terms of algebraic type support.  Will start looking at source code, but I thought it would be worth raising an issue that the docs appear unclear on what kind of support each function offers. For instance, this [test](https://github.com/ramda/ramda/blob/master/test/liftN.js#L42-L45) suggests LiftN works with arbitrary applicative functors (in this case a Monad too). But I am unclear in general what support is available for which functions - I understand that Ramda should work well with Folktale or other Fantasy-land implementations, just be great to get some commentary on which and how if you know what I mean. 

For instance:  Ramda's map function, operates similarly to the one on Array.prototype, so R.map(square, [1, 2, 3, 4]); //=> [1, 4, 9, 16]. But because Folktale's / Ramda-Fantasy's Maybe implements the Fantasy-land Functor spec, which also specifies map, you can also use Ramda's map with it. And I have seen this first-hand too, working with Ramda and Maybe, and third party libs like ImmutableJs, and RxJs over the past week.

Right, I get that but would like more information around "dispatch" and which functions it applies too etc without trawling the code base where possible. Can anyone offer some quick pointers here, it may be I am raising a non-issue and just need someone to point out something fundamental.

Thanks in advance.

arcseldon:
Just going to add a visual here too, to further illustrate what i think might be helpful. Comments welcome.

![ramdadocs](https://cloud.githubusercontent.com/assets/2599332/13181770/05bd7ff0-d772-11e5-8607-c6e61d01517c.jpg)

CrossEye:
Of course you can get the same thing with a text search of the page, but I think it might be useful if our search bar filtered the list by doing the equivalent of a text search of the sections for each named function.  That could well be more useful.

arcseldon:
@CrossEye - right, something like search of sections.  The search might address questions like - "Which functions in Ramda have dispatch capabilities that allow them to work with Objects that conform to the Fantasy Land apply specification?"

[...]

Issue #1282, opened 2015-07-16
The Ramda Manual


CrossEye:
We've discussed it a number of times, most recently in #1269: there is a level of documentation simply missing for Ramda: the tutorial-style description of many of its core capabilities.

I believe it's time to start building this.  It seems to me that the most logical place to do this is on the  [GitHub wiki][wi].  I'm envisioning a series of brief descriptions of capabilities of of related functions -- something that would take 5 - 10 minutes each to read, organized into a coherent logical picture of Ramda.

This really should be a team writing effort though, so I want to know if the idea seems worthwhile, and if there are people willing to contribute.  Below is an (almost-complete) rough draft of a table of contents for such a manual.  I figure people could take a section at a time, and we don't have to build it in order.  Whatever people find interesting to work on would be helpful.

This would not be a comprehensive guide to every function.  The API documentation is meant to do that.  This is meant to give a more in-depth view of how you would use various (groups of) functions from Ramda, and how they all hang together.

----------

Formatting:
* [ ] Unassigned
* [x] Assigned
* [x] <del>Completed<del>

----------

Ramda Manual
============

* [x] <del>Introduction: Core features in Ramda</del>
  * [x] <del>Small functions against common data structures</del>
  * [x] <del>Immutability of user data</del>
  * [x] Avoiding side-effects
  * [x] Composing functions
  * [x] Currying everywhere
  * [x] Dispatching
* [ ] Working with functions
  * [ ] Currying
  * [ ] Combining functions: `compose*`, `pipe*`, `converge`, `useWith`
  * [ ] Changing signatures: `flip`, `partial`, `lift`, `nAry`, `apply`, `unapply`
  * [ ] Calling functions: `apply`, `call`, `bind`, `memoize`
  * [ ] Adding additional functionality: `tap`, `wrap`, `invoker` 
* [ ] Working with lists
    * [ ] What is a list?
    * [ ] Arrays as lists
    * [ ] Strings as (sometimes) close-enough
  * [ ] Building lists: `range`, `repeat`, `times`, `keys`, `toPairs`
  * [ ] Slicing and dicing lists: `take`, `drop`, `append`, `prepend`, `concat`
  * [ ] Tweaking lists: `adjust`, `update`
  * [ ] List to list transformations: `map`, `pluck`, `filter`, `aperture`, `uniq`, `zip`, `partition`
  * [ ] Transforming lists to single values: `reduce`, `find`, `nth`, `all`/`any`
* [x] Working with objects
[...]

PR #1672, opened 2016-03-02
Curried Function Equality

ccorcos:
This is PR is just an experiment at this point. But I really think it would be useful -- or maybe there's a better way of dealing with my issues. [This discussion](https://github.com/ramda/ramda/issues/1481) didnt get too far, so let me rephrase some things.

In the big picture, I want to find some way of [optimizing a "listOf" higher-order component **in a pure functional way**.](https://github.com/ccorcos/elmish/blob/master/src/ui/listOf.js)  (it may make more sense if you just [browse through these](https://github.com/ccorcos/elmish/tree/master/tutorial)). This consists of (1) being able to lazily build, difference, and evaluate trees in an idiomatic fashion and (2) being able to compare bound functions as if they're referentially transparent to satisfy the first point.

Lets start with the second point. Here's a simple example using React:

```js
const Counter = React.createClass({
  render() {
    const inc = () => this.props.update(this.props.value + 1)
    return (
      <div>
        <span>{this.props.value}</span>
        <button onClick={inc}>inc</button>
      </div>
    )
  }
})

const ListOfCounters = React.createClass({
  getInitialState() {
    return {numbers: [1,2,3,4,5,6,7,8,9,10]}
  },
  update(i, value) {
    this.setState(R.evolve({
      numbers: R.update(i, value)
    }, this.state))
  },
  render() {
    const update = (i) => (value) => this.update(i, value)
    const counter = (value, i) => <Counter value={value} update={update(i)}/>
    return (
      <div>
        {this.state.numbers.map(counter)}
      </div>
    )
  }
})
```

The issue here is that whenever a single counter changes, all counters need to re-render. Thats because the `update` prop to each counter has a new reference. One solution is to make a custom `shouldComponentUpdate` that ignores the comparison of the `update` prop. But there are more complicated scenarios where this isn't a good thing to do, for example, if the list could add and remove counters so the indexes change. What would be amazing though, is if you could use `R.equals` to compare new props and `R.equals` can compare functions *as if they're referentially transparent*. And the only change that would need to be made to the logic (other than changing `shuoldCompnentUpdate` to use `R.equals`) above is:

```js
const update = R.curry(this.update)
```

Thats because we're able to keep track of the original function and any bound arguments thus far. 

Now, `R.curry` isn't the only place you'd want this kind of functionality. Maybe you just want a thunk to call a function with some arguments. For example:

```js
const ListOfCounters = React.createClass({
  getInitialState() {
    return {numbers: [1,2,3,4,5,6,7,8,9,10]}
  },
  update(i) {
    this.setState(R.evolve({
      numbers: R.adjust(i, R.inc)
    }, this.state))
  },
  render() {
    const counter = (value, i) => {
      const inc = () => this.update(i)
      return (
        <div>
          <span>{value}</span>
          <button onClick={inc}>inc</button>
        </div>
      )
    }
    return (
      <div>
        {this.state.numbers.map(counter)}
      </div>
    )
  }
})
```

In this case, you don't exactly want to curry, but you want to return a `thunk`. So perhaps we could have a function called `R.thunk` or maybe we could just use `R.partial`. But the point is that we have a function that is bound to an argument. And so long as we're using pure functional programming patterns, then there should be no harm in partially applied functions being referentially transparent. That is:

```js
R.equals(R.partial(f, 1), R.partial(f, 1))
```

Obviously they won't truely be referentially transparent in JavaScript, but that's why we have Ramda, right?

This leads me back to the first point... React lazily evaluates the virtual dom tree because when you create a component like `<Counter value={value} update={update}/>`, the JSX compiles into `React.createElement(Counter, {value, update})` which in turn returns some sort of object like `{type: Counter, props: {value, update}}`. When React is evaluating the vdom to see if there are any updates, it first sees if we want to render the same component (Counter), and then compares the props to see if a re-render is needed. The thing that bugs me about this is that this format is so specific to React. Sure, we could copy it, but I think theres a much more idiomatic way of accomplishing all of this.

Suppose our components are "stateless components" as React would call it, or just functions. What we really want here is a thunk like we talked about before.

```
const thunk = R.partial(render, props)
```

If we call `thunk()` we'll get back the vdom tree thats returned from render. But if we want to lazily evaluate the vdom tree, we can simply use `R.equals` to compare thunks to determine if anything has changed. 

What I really like about this approach is its more generic feeling--its just a function with bound arguments. In an academic, functional, pure world, there's no reason bound functions shouldnt be referentially transparent. As I've been getting more and more familiar with functional programming, this has always been the limiting factor as far as complexity/performance in javascript, esp in my [elmish](https://github.com/ccorcos/elmish) project where I'm building similar lazy trees for things like http requests and side-effects other than just the vdom.

Maybe I'm just too far down the rabbit hole, but I hope that all makes sense. I'm very interested to hear some of your thoughts.

Cheers

### DIFF
--- a/src/curryN.js
+++ b/src/curryN.js
@@ -1,4 +1,3 @@
-var _arity = require('./internal/_arity');
 var _curry1 = require('./internal/_curry1');
 var _curry2 = require('./internal/_curry2');
 var _curryN = require('./internal/_curryN');
@@ -50,5 +49,5 @@ module.exports = _curry2(function curryN(length, fn) {
   if (length === 1) {
     return _curry1(fn);
   }
-  return _arity(length, _curryN(length, [], fn));
+  return _curryN(length, [], fn);
 });
--- a/src/internal/_curryN.js
+++ b/src/internal/_curryN.js
@@ -1,6 +1,8 @@
 var _arity = require('./_arity');
+var _complement = require('./_complement');
+var _equals = require('./_equals');
+var _filter = require('./_filter');
 var _isPlaceholder = require('./_isPlaceholder');
-
 
 /**
  * Internal curryN function.
@@ -13,7 +15,8 @@ var _isPlaceholder = require('./_isPlaceholder');
  * @return {Function} The curried function.
  */
 module.exports = function _curryN(length, received, fn) {
-  return function() {
+  var arity = length - _filter(_complement(_isPlaceholder), received).length;
+  var _fn = _arity(arity, function() {
     var combined = [];
     var argsIdx = 0;
     var left = length;
@@ -35,6 +38,17 @@ module.exports = function _curryN(length, received, fn) {
       combinedIdx += 1;
     }
     return left <= 0 ? fn.apply(this, combined)
-                     : _arity(left, _curryN(length, combined, fn));
+                     : _curryN(length, combined, fn);
+  });
+  // keep track of the original function and the combined args
+  _fn.curried = true
+  _fn.fn = fn.curried ? fn.fn : fn
+  _fn.args = received;
+  // compare the original function with the combined args
+  _fn.equals = function(fn2) {
+    return fn2.curried &&
+           _equals(fn2.fn, _fn.fn, [], []) &&
+           _equals(fn2.args, _fn.args, [], []);
   };
+  return _fn
 };
--- a/test/curryN.js
+++ b/test/curryN.js
@@ -105,4 +105,19 @@ describe('curryN', function() {
     eq(g(1)(2)(3, 4), [1, 2, 3, 4]);
   });
 
+  it('equals another curried function', function() {
+    var f = function() { return Array.prototype.slice.call(arguments); };
+    var g = R.curryN(3, f);
+
+    eq(true, R.equals(g(1), g(1)));
+    eq(true, R.equals(g(1, 2), g(1, 2)));
+    eq(true, R.equals(g(), g()));
+
+    eq(false, R.equals(g(1), g(2)));
+    eq(false, R.equals(g(1, 2), g(2)));
+    eq(false, R.equals(g(1, 2), g(2, 1)));
+
+    eq(true, R.equals(g(g(1,2)), g(g(1,2))));
+    eq(false, R.equals(g(g(1,2)), g(g(2,1))));
+  });
 });


ccorcos:
[Here's a simple example](https://github.com/ccorcos/elmish/blob/147ac1730a6fe700cae85db478e7d651b6073730/lazy/main.js) demonstrating a real use case. I've called my function `ƒ`. If you run the example (`node server.js`), you'll see that it optimally re-renders.

Maybe this is outside the scope of Ramda, but I'd still like to hear what you guys think.

CrossEye:
I saw [your article on Medium][cc] this weekend and am intrigued by it.  I will look at this more carefully later, but my initial impression hasn't changed much.

I suppose one could argue that this reduces the incidence of false negatives in `equals`, but because what it does is so restricted, I feel as though it's setting up a false expectation.  I was just discussing this with a colleague, who pointed out that it's never going to return true even for this:

```js
equals((a, b) => a + b, (a, b) => a + b)
```

That makes it hard to really believe it's going to add much value.  See also [Rice's Theorem][rt].

But I will look it over more carefully when I have a bit more time.

  [cc]: https://medium.com/@chetcorcos/elmish-functional-programming-in-javascript-50995f1d4b9e#.6p3klm8m3
  [rt]: https://en.wikipedia.org/wiki/Rice%27s_theorem

ccorcos:
I have more ideas for elmish too. I also just published 2 days ago ["functional programming for javascript people"](https://medium.com/@chetcorcos/functional-programming-for-javascript-people-1915d8775504#.fxesdegdw) -- and it really made me think a lot about persistent data structures and lazy evaluation as the next *big step* for a functional programming utility library for something like Ramda.

To your point, `equals((a, b) => a + b, (a, b) => a + b)` will never be something we can do in JS. If you were working in Clojure, you could create a macro that would be able to successfully evaluate this though. But the point isn't to get some generic referential transparency in JS because that wont happen. I would, however, like if we could maintain some sanity without our own ecosystem though. 

Rather than think of this as comparing functions (and all the issues and inconsistencies associated with that), lets think of it as comparing thunks.

```js
const thunk = (fn, ...args) => {
  let _fn = (...more) => {
    return R.apply(fn, R.concat(args, more))
  }
  _fn.fn = fn
  _fn.args = args
  _fn.equals = (fn2) => {
    return R.equals(fn2.fn, _fn.fn) &&
           R.equals(fn2.args, _fn.args)
  }
  return _fn
}
```

In this world, we can show that `equals(thunk(f, 1), thunk(f, 1))`. But we must always be comparing thunks with thunks and not thunks with functions or functions with functions. You still run into the issue of `equals(thunk((a, b) => a + b), thunk((a, b) => a + b))` but i think its reasonable to force the developer to define the pure function as its own reference so that this thunk equality can work.

So at the end of the day (1) to avoid confusion, lets think about comparing thunks (or maybe theres a more proper word for this) as opposed to comparing functions because function equality will never be a thing in JS and (2) it would be awesome if Ramda understood thunks and used thunks for curried and partially applied functions -- thus if you dont care about thunks, you never have to think about it, and everything works as usual, but if you know what youre doing and want to squeeze some more performance our of your pure functional program, now you can!

ccorcos:
I also want to say, that this could become a more generic standard for thunks. Perhaps we could submit a PR to React which would allow thunks for stateless components as opposed to React.createElement.

CrossEye:
> I also just published 2 days ago ["functional programming for javascript people"][cc]

Nice!  Made a comment with two minor corrections.  But a nice introduction.


> and it really made me think a lot about persistent data structures and lazy evaluation as the next *big step* for a functional programming utility library for something like Ramda.

There's a lot to be said for lazy evaluation, I believe.  But I don't know what sort of persistent data structure you might be considering.  Ramda's insistence on immutability often puts it at odds with many people's ideas of persistent structures.

> To your point, `equals((a, b) => a + b, (a, b) => a + b)` will never be something we can do in JS. If you were working in Clojure, you could create a macro that would be able to successfully evaluate this though.

Well, no, Rice's Theorem says that we cannot do so consistently.  For instance, would a macro also note that this is equivalent to those?

```js
(x, y) => {
    const z = x - y;
    return x < y ? 2* x - z : 2 * y + z;
}
```

> Rather than think of this as comparing functions [ ... ], lets think of it as comparing thunks.
 
Whether we call them "thunks" or "closures", we would have to think this through very carefully.  I started work some time ago -- for entirely different reasons -- on some code that stored much of this information in metadata.  That metadata is in a local closure rather than in property expados on the function.  It strikes me as significantly safer that way.  I'm afraid the data hiding might then make it impossible to use these for what you're doing.  I'm not entirely wedded to the closures, though.  Nor have I had time to finish what I was working on, a somewhat radical rethinking of all the function wrapping we do in Ramda.  But if it switched to expandos, this could conceivably work in.

But I'm still not sure I want it to.  Yes, right now `equals` works with functions through reference equality, but that's because we've come to terms with the notion that _there is no better notion of equality for functions_.  What you're suggesting contradicts that, but says that it's only good for some functions built in a certain way, when the moon is just so.  And that scares me.

My initial opposition is now tempered by some interest and some confusion.  But I'm far from convinced.

  [cc]: https://medium.com/@chetcorcos/functional-programming-for-javascript-people-1915d8775504#.fxesdegdw

arcseldon:
@ccorcos - Hey, good post on Medium. Enjoyed the read, and left you brief feedback.

> Two of my absolute favorite JavaScript libraries are Ramda and Flyd.  But Ramda isn’t lazy and doesn’t play nice with Immutable.js. 

Can you please elaborate on what aspects of Ramda doesn't play nice with Immutable.js. Not trying to defend ramda, just be really interested to explore further and understand where you had cases in which it came short - still not seen much concrete code contributions from anyone with regards to Immutable.js and Ramda inter-op. For instance, just because ramda isn't lazy, I don't think that disqualifies it from inter-op due to transducers. Take a look at a (very, very) limited snippet I wrote on this [issue](https://github.com/ramda/ramda/issues/1282#issuecomment-184695573)


> I’d really like to see a library that combines all of these concepts — persistent / shared / immutable data structures with curried, lazily evaluated, composable utility functions.

Yes! Really like your thinking here.

Btw, flyd looks really interesting, thanks for the mention. This snippet on [ramda transducer](https://github.com/paldepind/flyd/blob/master/examples/ramda-transducer/script.js) caught my attention.

arcseldon:
> I’d really like to see a library that combines all of these concepts — persistent / shared / immutable data structures with curried, lazily evaluated, composable utility functions.

Perhaps with the fantasy land algebraic data types baked into the APIs too (Maybe, Either, IO, and Future at least).

Finally, I saw this which might interest people - [transducers.js](https://github.com/jlongster/transducers.js)  If nothing else, the documentation provided in describing transducers is better than that of Ramda's ?  Perhaps we could lift some of the prose from that readme, and add to ramda docs (with necessary modifications of course).

For instance, it clearly lays out which transformations are availble (ie. which functions are transducer friendly). I still struggle to understand what's what with Ramda function support atm...

ccorcos:
@CrossEye in terms of persistent data structures, that quote sums it up pretty well:

> I’d really like to see a library that combines all of these concepts — persistent / shared / immutable data structures with curried, lazily evaluated, composable utility functions.

In an ideal world, I could swap out a native javascript array for whatever other library's list/array/vector equivalent and all the data transformations with Ramda would still work. I'm not sure @arcseldon's comment on using transducers would help here because (1) they're not lazy and (2) you'd still have to re-create utility functions for whatever implementations of list/array/vector etc. I think the only way for this to really work is if theres some consensus on what sort of prototype methods should be associated with what types. I think `prototype.map` is pretty understandable, and I'd imagine you could use R.map with an Immutable.js Map. But it would be great if there were standards for things like `obj.get(key)` or `array.get(3)` instead of `obj[key]` and `array[3]` because I don't think we can make custom functions for grabbing those indexes. And thus `R.prop` wouldnt work Immutable.js -- so in short, it would be awesome if we had some kind of standard for what prototype functions mean/do what so ramda could be the common data transformation library. 

On another note, it would be awesome if ramda had full support for generators so we could lazily evaluate things like `R.pipe(R.filter(R.gt(10)), R.take(4))`. It would be awesome if ramda returned a generator if you pass it a generator. That way you get all the laziness you want, if you want it.

But now back to the real point of this PR:

---

@CrossEye, you make a good point about Rice's theorem, and I'd love to entertain that concept some more, but I'm afraid it would be too distracting from the point. 


> that's because we've come to terms with the notion that there is no better notion of equality for functions

I don't want to compare functions though. I want to compare specific functions that I'm calling *thunks* which are partially applied functions. And that means if you partially apply *the exact same function* with the same arguments, the they're effectively the same thing.

> What you're suggesting contradicts that, but says that it's only good for some functions built in a certain way, when the moon is just so. And that scares me.

This is true, you definitely need to be conscious of how you're writing your program. If you define the same pure function twice with two different references, then they're never going to be equal. And if you have a function that isnt pure, then you might have a similar problem as well. 

This concept of *pseudo-referentially-transparent thunks* is very specific and pertains to a certain functional style of programming. Its not perfect, and thats because Javascript isnt perfect, but we're trying to make due with what we have. And when you're building user interfaces with a pure functional style, then you end up in situations when you want to be lazy, but you cant because you have new functions each time so you can never be sure if you need to re-render or not. For example, here, we can diff the props for MyButton and we dont have to rerender every time because the insert function won't have changed based on R.equals. Otherwise, you'd have to re-render every single time.

```js
    view: (dispatch, state) => {
      const insert = thunk(dispatch, {type: 'insert'})
      return h('div.list-of', [
        h(MyButton, {onClick: insert}, '+')
      ])
    }
```

The alternative is a very object oriented style of programming:

```js
React.createClass({
  insert() {
    this.props.dispatch({type: 'insert'})
  },
  render() {
    return h('div.list-of', [
      h(MyButton, {onClick: this.insert}, '+')
    ])
  }
})
```

But now the function / component is stateful and we don't want that. I've had to write `this` mutliple times. Those functions are no longer pure. These are the compromises I'm faced with all the time, and **the only way of creating efficient versions of the same UI using only pure functions** is with some sort of psuedo-referentially-transparent way of comparing functions that I've demonstrated.

I understand your concern from a broader persepctive about programming in general. But think about it within the context of how its being used and when. If this were incorporated into Ramda (and if definitely could), then you get all these other bonuses as well:

```js
R.equals(R.map(f), R.map(f))
R.equals(R.pipe(R.prop('count'), R.inc), R.pipe(R.prop('count'), R.inc))
```

My argument for referential transparency in my article was thinking like a mathematician. If those functions above return false, then you can't prove anything and you dont know how to reason about the world! It just makes so much sense that they should be equal. We already arent using `===` and `==` and are instead using `R.equals` because it just makes more sense. So we've already made the compromise that JavaScript is terrible sometimes. And I think this is another great example. And not only that, its very useful when you build an entire application out of pure functions so you can lazily evaluate things and be just as performant as an object oriented approach, and this seems to align with Ramda's mission.

arcseldon:
@ccorcos - good to hear your thoughts. There has been quite a bit of discussion already on generators introduction in Ramda - a search through the issues archive should bring up quite a bit of commentary. I was recently enquiring about that too. Still not 100% certain I understand the conclusion though, was it performance ?

> In an ideal world, I could swap out a native javascript array for whatever other library's list/array/vector equivalent and all the data transformations with Ramda would still work. I'm not sure @arcseldon's comment on using transducers would help here because (1) they're not lazy and (2) you'd still have to re-create utility functions for whatever implementations of list/array/vector etc. 

I hear you!  What you are saying resonates. It would indeed be good if we could swap out native JS array with equivalents. I may well have been off the mark with regards to laziness and transducers. My understanding is not that clear, but i did think they could be used in such a way that intermediate data structures were avoided - ie. much like a pipelined iterator where each element gets passed right through the pipeline - basically just like ES6 generator / iterator behaviour. And if you stick say a take(10) on the end of the composition then you've only lazily taken 10 regardless of the length of the original input Seq.
Hence you get all the efficiency gains and reduced garbage collector usage. 

I'd really appreciate some guidance here, and also which functions this applies to in Ramda 19.1.

arcseldon:
Here is a small example using Ramda and ImmutableJs, creating a lazy pipeline:

```js
 it('should behave lazily...', () => {
      let range = Immutable.Range(1, Infinity);
      expect(range.size).to.equal(Infinity);
      const mul2 = x => {
        console.log(`mul2 ${x}`);
        return x * 2;
      };
      const add3 = x => {
        console.log(`add3 ${x}`);
        return x + 3;
      };
      const calc = R.compose(R.take(10), R.map(add3), R.map(mul2));
      const result = calc(range);
      console.log(result);
      expect(result.size).to.equal(10);
    });
```

Results output:

```
mul2 1
add3 2
mul2 2
add3 4
mul2 3
add3 6
mul2 4
add3 8
mul2 5
add3 10
mul2 6
add3 12
mul2 7
add3 14
mul2 8
add3 16
mul2 9
add3 18
mul2 10
add3 20
Seq [ 5, 7, 9, 11, 13, 15, 17, 19, 21, 23 ]
```

ccorcos:
@arcseldon interesting. just dove through a bunch source code -- Immutable.js seems to do a great job of supporting prototype functions. They implement `.slice()` which is actually lazy, and thus `R.take` works on it! They also implement `.map`. So the example you gave works. However, `R.prop` does not check for the `.get` method and thus that doesn't work. Sounds like a good PR though!

Transducers require you to specify how a transformation and a reduction. It still traverses the entire list though, so it isnt exactly lazy. Transducers are interesting, but I'm starting to think that generators are a much better way of doing this stuff....

The problem I'm seeing is that native generators don't have a the prototype methods I'd like to see. For example, since it doesnt implement slice, you're kind of at a loss.

```js
> let R = require('ramda')
undefined
> function* counter() { var i = 0; while(true) { yield i++ } }
undefined
> let c = counter()
undefined
> c.next()
{ value: 0, done: false }
> c.next()
{ value: 1, done: false }
> c.next()
{ value: 2, done: false }
> R.take(10, c)
[]
> R.map(R.inc, c)
[ undefined ]
```

I tried making some patches to make all this work, but I couldnt because (1) I dont know if its possible to modify function generator's prototype methods, and (2) its not possible to access a generators constructor functoin and construct a fresh identitcal generator so you're not performing side-effects. But that said, I've been able to get some similar functionality:

```js
function Gen(fn) {
  let g = fn()
  g.drop = function drop(n) {
    // it would be great if we could somehow construct a new identical generator
    for(let i = 0; i < n; i++) { this.next() }
    let self = this;
    return Gen(function* _drop() {
      yield self.next()
    })
  }

  g.take = function take(n) {
    // it would be great if we could somehow construct a new identical generator
    let self = this
    return Gen(function* _take() {
      for(let i = 0; i < n; i++) { yield self.next() }
    })
  }

  g.slice = function slice(fromIndex, toIndex) {
    // it would be great if we could somehow construct a new identical generator
    return self.drop(fromIndex).take(toIndex-fromIndex)
  }

  g.values = function values() {
    let list = []
    for (let item of this) {
      list.push(item.value)
    }
    return list
  }
  return g
}

> c = Gen(function* counter() { var i = 0; while(true) { yield i++ } })

> R.take(10, c).values()
[ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9 ]
> R.take(10, c).values()
[ 10, 11, 12, 13, 14, 15, 16, 17, 18, 19 ]
```

Notice that its lazy--we're dealing with an infinite list here. But also notice that there are side-effects and the generator is mutated because we get the next set of 10 when we call it again. Thats no good. However, its interesting to see that both Immutable.js and Ramda are working towards implementing some sort common algrebraic language so they can be interoperable. .. perhaps this is worth opening a new PR for to explore some more.

arcseldon:
@ccorcos - Thanks for the reply! That definitely helped my understanding. 

Going back to your original comment around having a `library with  persistent / shared / immutable data structures with curried, lazily evaluated, composable utility functions`

Perhaps a PR that focuses on driving further ImmutableJs and Ramda interoperability gets us one step closer. When I get some time I shall try researching further.  If you decide to go ahead and raise a PR, then I'll actively try to contribute / comment / research based on discussions etc.

CrossEye:
Note that Ramda will work with this generator-based list through its transducers, so long as the pipeline ends in something that reduces it to a finite list:

```js
function* counter() { var i = 0; while(true) { yield i++ } }
R.into([], R.compose(R.take(5), R.map(n => n * n)), counter());    
//=> [0, 1, 4, 9, 16]
```

CrossEye:
Is it time to resurrect the [LazyList][ll] extension?


  [ll]: https://github.com/ramda/ramda/pull/857

ccorcos:
[ramda-fantasy](https://github.com/ramda/ramda-fantasy) -- nice!

@CrossEye thats not lazy though -- looks like it turns it into an array and uses all the normal stuff... ideally, it should return a generator if you pass a generator..

LazyList looks interesting, but I think its overkill. If Ramda could somehow support generators (or a better solution is if generators support Ramda/fantasyland) then we could convert a list into a generator and we're all good.

CrossEye:
While we probably could create an `into` that would in fact return a generator (`into` can simply adhere to the [transducer protocol][tp]), it still would not get us the laziness that you're looking for.

Two things about supporting generators: First of all, it seems like it would be difficult.  Probably half of our functions deal with lists.  Every one of those would need to add branches to deal with generators, and there would be the further problem of recognizing generators: do we use duck typing?  If it has a `next` method it's a generator?  What if that returned something without `done` / `value` properties?  It's not that one couldn't write code for this.  Braithwaite's [JavaScript Allongé "Six"][a6] demonstrates very clean code for this.  But hooking it into Ramda seems challenging.

But the second question is why just generators?  Wouldn't it be better to deal with all iterables?  That seems to be the proper abstraction of lists to potentially infinite sequences.  Is there any reason to focus solely on generators?


  [a6]: https://leanpub.com/javascriptallongesix
  [tp]: https://github.com/cognitect-labs/transducers-js#transformer-protocol

arcseldon:
> But the second question is why just generators? Wouldn't it be better to deal with all iterables? That seems to be the proper abstraction of lists to potentially infinite sequences. Is there any reason to focus solely on generators?

For small list sizes laziness may be more expensive, but overall my take is it would be great if ramda users could opt into laziness at will. For me, generators seem to present the idiomatic way of achieving this and are now part of the language syntax. If there is a way to achieve the same effect as generators, such as passing in ImmutableJs List types and everything "just works" then great. But right now, we seem to be slightly stuck in "no mans land" with no built-in ramda laziness (as LazyList was relegated), partial (undocumented?) support for libraries such as RxJs and ImmutableJs, and (dare I say it) slightly circular debate over how to proceed. 

> persistent / shared / immutable data structures with curried, lazily evaluated, composable utility functions.

If we can achieve that using a solution that works with all iterables, then fantastic. If it requires making concessions, then what would those concessions be?

CrossEye:
> [It] would be great if ramda users could opt into laziness at will. For me, generators seem to present the idiomatic way of achieving this and are now part of the language syntax.

Our current delegation mechanism was at least the first pass at something similar.  The idea was that if you implement, say, `map`, we can simply call it.  That's not enough, because things we might want to work with _don't_ implement such functions, or -- in cases like mori -- implement them in a way that won't work with our delegation.

The lazy list extension was the very first pass at doing any sort of delegation, and it was extremely naive.  But it did lead to something that at least works with types that implement various of the FantasyLand specifications.  My thoughts were always a bit broader, though.  Although lazy lists were the first implementation, they were never my main idea.  What I really wanted to do was what's now fairly common in FRP libraries, treating streams and arrays almost alike, but really dealing with any sequential collection in a uniform manner.

I haven't seen any reason to think of laziness as anything very special.  To me, generators are simply one more way to create iterators.  I do recognize that there's more to them than that, but that's all they've ever meant to me.

But, philosophy aside, does anyone have a practical suggestion for how to make this work in Ramda?  Updating every list function to optionally deal with arbitrary generators or itarators would be a massive undertaking.  Does anyone felel like taking a crack at it?

ccorcos:
It looks like [native iterators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_Generators) are pretty much defined by having a Symbol.iterator property.

> But the second question is why just generators? Wouldn't it be better to deal with all iterables?

Definitely! I'm just not entirely sure how that would be accomplished -- is there some kind of fantasyland spec for iterables? I'm 100% on board with this btw.

> Our current delegation mechanism was at least the first pass at something similar.

I figured this is probably the best way of doing it. I'm curious whats the best way to create "type classes" in JS. Immutable seems to be doing a decent job using ES6 classes, but it would be great if you could do the same things without using `this`. And Mori seems to be similar to ramda in that it has its own `mori.map` function, but it would be much nicer if there was some sort of interface for the data structures so you could use a more generic R.map which maps over *anything that is mappable*.

> Does anyone felel like taking a crack at it?

@CrossEye, I'm happy to. Maybe this weekend if I have time, maybe in the coming weeks. But before I do, I probably need to learn some more stuff. Idk how to accomodate genericly for "iterables". If you have any articles or videos to share, I'd like to check them out.

ccorcos:
[here's a cool way of using type classes in JS](https://github.com/loop-recur/typeclasses/blob/master/functor.js#L27-L32) and its not OO

CrossEye:
>> Wouldn't it be better to deal with all iterables?

> Definitely! I'm just not entirely sure how that would be accomplished -- is there some kind of fantasyland spec for iterables? I'm 100% on board with this btw.

It's actually a [language spec][it], probably more readable on [MDN][md].

> I'm curious whats the best way to create "type classes" in JS.

There are many [implementations][fi] of [FantasyLand][fl], including [Ramda's own][rf], all of which implement various types.  I would hope that Ramda doesn't try to do this itself, or doesn't do much of it.

>> Does anyone felel like taking a crack at it?

> I'm happy to. [ ... ] But before I do, I probably need to learn some more stuff. Idk how to accomodate genericly for "iterables". If you have any articles or videos to share, I'd like to check them out.

I don't have anything in particular, except to point again to Braithwaite's [JavaScript Allongé "Six"][a6], [the documentation][md], and to note that to some degree Ramda [already][r1] [does this][r2] in it's internal [`_reduce`][re] function.

  [a6]: https://leanpub.com/javascriptallongesix
  [fi]: https://github.com/fantasyland/fantasy-land/blob/master/implementations.md
  [it]: http://www.ecma-international.org/ecma-262/6.0/#sec-iteration
  [md]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols
  [re]: https://github.com/ramda/ramda/blob/master/src/internal/_reduce.js
  [r1]: https://github.com/ramda/ramda/blob/master/src/internal/_reduce.js#L38
  [r2]: https://github.com/ramda/ramda/blob/master/src/internal/_reduce.js#L49-L51

ccorcos:
Hmm. Well the issue I'm having with the iterator spec is that it doesn't play nice with immutability because you can't clone a generator since in the general case, there may be all kinds of complicated internal state going on. So how can you create a generic "drop" function?

```js
> function* counter() { var i = 0; while(true) { yield i++ } }
undefined
> c = counter()
{}
> c.next()
{ value: 0, done: false }
> c.next()
{ value: 1, done: false }
>
> function drop(n, iter) {
...   for(let i = 0; i < n; i++) { iter.next() }
...   return iter
... }
undefined
>
> d = drop(10, c)
{}
> d.next()
{ value: 12, done: false }
> c.next()
{ value: 13, done: false }
```

Maybe I just don't know how to read the fantasyland spec, but where the hell does it talk about iterables, lists, associative maps? Maybe I'm misunderstanding the point of fantasyland...

arcseldon:
> Maybe I just don't know how to read the fantasyland spec, but where the hell does it talk about iterables, lists, associative maps? Maybe I'm misunderstanding the point of fantasyland...

@ccorcos - the fantasyland spec is a specification for interoperability of common algebraic structures in JavaScript akin inspired from Haskell (although now many languages have similar implementations). It helpfully spells out what an implementation should adhere to in order to achieve interoperability with other implementation of the same algebraic type.

Hesitate to offer recommended reading as you clearly have a good grasp of the basics of FP, but here are couple anyhow that I would pretty much recommend to anyone new to Ramda:

[Learn  you a Haskell](http://learnyouahaskell.com)

[Mostly adequate guide](https://www.gitbook.com/book/drboolean/mostly-adequate-guide/details)

The second book is by Brian Lonsdorf (dr boolean) the same guy that wrote that snippet of code you mentioned above on types.

Essentially, the algebraic types are implementations of functors, pointed functors, applicative functors, and monads, each of which has a context when it is most appropriate to use. This [issue](https://github.com/ramda/ramda/issues/1650) might be helpful - but maybe not..

Support for these is already part of Ramda due to its dispatch mechanism, but the APIs don't expose any of these as return types of their own accord.  head([]) does not return a Maybe etc - so to achieve this you have to either compose with a Maybe next in the pipelline or else wrap in a Maybe constructor etc. Personally, I would be in favour of going down that road, but my views here only recently changed after weeks of daily usage to understand if they made sense.  

[Sanctuary](https://github.com/plaid/sanctuary) has gone all in (incidentally the lead dev on that project is also a core team member of Ramda), but has a dependency on Ramda internally. @davidchambers - quick question please - does Sanctuary offer much in terms of laziness support?  For instance, if I pass in Immutable List, will it pipeline lazily etc, or is it very similar to Ramda in that regards?

What I most interested in seeing is laziness totally supported out of the box. Now if that is achievable by ensuring we have fluent support for the fantasy spec (?) then great, but I thought actually there were some orthogonal concerns here.. Sorry was I wrong in thinking that?  As @CrossEye pointed out we already do `map` dispatch etc but there are still interoperability issues, and laziness support is sketchy at best.

Separately, and sorry to be a scatter brain, I'd also be interested to know whether when JS has proper support for tail recursion, that might have any effect on how a modern FP library could be written internally too, or whether performance would still be a major problem.

davidchambers:
> does Sanctuary offer much in terms of laziness support?

Not at this stage.

ccorcos:
Sanctuary looks awesome! I really like that idea. Although thats going to scare a lot of people away from Ramda... We essentially get Maybes with `undefined`... Santuary is a very cool project though.

So to my question, where in the fantasyland spec does it specify indexing into arrays or associative maps? `obj[key]`, `list[index]`?

arcseldon:
> So to my question, where in the fantasyland spec does it specify indexing into arrays or associative maps? obj[key], list[index]?

@ccorcos - would also welcome some commentary here. 

@TheLudd @buzzdecafe - both contributed heavily to impl. of [ramda-fantasy](https://github.com/ramda/ramda-fantasy) and likely know more.  From what I read of the [spec](https://github.com/fantasyland/fantasy-land) although it does talk about definitions of plain JavaScript objects in terms of equivalence

> Two lists are equivalent if they are equivalent at all indices.
> Two plain old JavaScript objects, interpreted as dictionaries, are equivalent when they are equivalent for all keys.
> Two promises are equivalent when they yield equivalent values.
> Two functions are equivalent if they yield equivalent outputs for equivalent inputs.

It really doesn't go into the lower level implementation details of specifying indexing into arrays or associative maps. All the prose is at a much higher level of abstraction.  But at this point, I am going to step aside and hope to hear from others who have already been through the process of building such an implementation and actually know what they are talking about! :smile:

arcseldon:
@davidchambers - thank you for clarification.

arcseldon:
Final post.. The more I read about [ImmutableJs](https://facebook.github.io/immutable-js), the more I think this sounds like what we are all discussing here. Going right back to an earlier thread in this discussion, perhaps achieving inter-op with ImmutableJs, is one way to provide a facility for those that want that kind of behaviour (on the understanding you are working with ImmutableJs data structures throughout, we don't provide for implicit conversions between plain old JS and ImmutableJS types etc. 

In terms of dispatch support - which functions currently do this already?  Mentioned above:

>  just dove through a bunch source code -- Immutable.js seems to do a great job of supporting prototype functions. They implement .slice() which is actually lazy, and thus R.take works on it! They also implement .map. So the example you gave works. However, R.prop does not check for the .get method and thus that doesn't work. 

So is ImmutableJs following any kind of standards / specification, or would providing support for that library be akin to hard-coding in conditional logic / checks that apply just to that library, and might even conflict with and break alternate libraries that clash with similar function names, but different intentions.

arcseldon:
Well, I think this is what needs to be implemented to gain support (iterable dispatch from ramda at least for ImmutableJs iterables) - [iterable](https://facebook.github.io/immutable-js/docs/#/Iterable)  - PR anyone?

CrossEye:
@ccoros:
> Well the issue I'm having with the iterator spec is that it doesn't play nice with immutability because you can't clone a generator since in the general case

I think we would have to think through what sorts of functions make sense.  Clearly it does not actually make sense to try to clone a generator or another iterator, any more than it makes sense to clone a stream.  How in the world would you expect to clone this generator?:

```js
var myGen = function * () {while (true) yield Math.random();};
```

Perhaps one good way to go about this is to look at the overlap between Ramda and one of the standard pure lazy libraries, Lazy, or lz, or both.  Maybe those functions would be the ones we would try to port to iterators.

Or maybe this really is a pipe dream, since iterators are so stateful.

----------

@arcseldon:
> Separately, and sorry to be a scatter brain, I'd also be interested to know whether when JS has proper support for tail recursion

It's [specified by ES6][tc], but as of now [almost no one has implemented it][ka].  (Note, at the time of this writing, the only environments in Kangax's excellent table listed as supporting it are Webkit r19750 and Kinoma X56.  I'm pretty sure at one point Babel was on that list.  I don't know what happened.)

> The more I read about [ImmutableJs], the more I think this sounds like what we are all discussing here. Going right back to an earlier thread in this discussion, perhaps achieving inter-op with ImmutableJs, is one way to provide a facility for those that want that kind of behaviour.

I think doing so would be excellent.  The most recent discussion on this was in #1367.  It fizzled out.  I'd love to see something that looks like progress here.  Any suggestions would be welcome.


  [kc]: https://kangax.github.io/compat-table/es6/#test-proper_tail_calls_%28tail_call_optimisation%29
  [im]: https://facebook.github.io/immutable-js
  [tc]: http://www.ecma-international.org/ecma-262/6.0/#sec-tail-position-calls

ccorcos:
> Two functions are equivalent if they yield equivalent outputs for equivalent inputs.

Now that fantasyland spec is the real motivation for this PR.

As for Immutable, I asked them about their motivation for names and types and any standards or specs they're going off of:

https://github.com/facebook/immutable-js/issues/802

ccorcos:
this whole writeup about react performance is addressed with this:

http://benchling.engineering/deep-dive-react-perf-debugging/

ccorcos:
Just wanted to mention that I created a little library for creating class types in JS.

https://github.com/ccorcos/js-type

CrossEye:
@ccoros:

Interesting.  I've seen several similar libraries.  Have you worked with [daggy][da] or @paldepind's [union-types][ut]?

There is one thing about your API that bothers me, and it's from your own description:

> The second argument is either an array, an object, or a function. If its an array, it needs to be an array of strings. [ ... ] If its an object, then we're creating a sum-type [ ... ]. If its a function, then its a constructor function [ ... ].

I would personally prefer having separate functions to handle these cases than a single function which acts in all these different ways.

Nonetheless, very interesting!

  [da]: https://github.com/puffnfresh/daggy/
  [ut]: https://github.com/paldepind/union-type

ccorcos:
I took inspiration from both daggy and union-type. What I really wanted to do with this library was allow you to create prototype methods without having to deal with `this` everywhere.

> I would personally prefer having separate functions to handle these cases than a single function which acts in all these different ways.

That's actually how I wrote it! Then I had some fun making a polymorphic function :)

https://github.com/ccorcos/js-type/blob/master/src/type.js#L159-L166

Still dont like it?

CrossEye:
>> I would personally prefer having separate functions to handle these cases than a single function which acts in all these different ways.

> That's actually how I wrote it! Then I had some fun making a polymorphic function :)
> 
> https://github.com/ccorcos/js-type/blob/master/src/type.js#L159-L166
> 
> Still dont like it?

Afraid not, or that part of it at least.  The signature is too complex for me.  I really prefer my functions to do just one thing.  I would like this better if it exposed `tagged`, `construct`, and `taggedSum` individually.

I'm still trying to figure out if this is something I would use.  I don't do much with constructed types outside a few utility ones right now, but this certainly seems a decent approach for when I do.  But I also don't share your concerns, I believe, for `this`.  If I need it, I don't worry about using it.  So I'm not sure if this style is overkill for me.

ccorcos:
> I really prefer my functions to do just one thing. I would like this better if it exposed `tagged`, `construct`, and `taggedSum` individually.

I'll do that then ;) I just think of it as one function that dispatches to one of three functions based on the type signature. or one of 6 in this case. And thats just more convenient. It would be annoying to write it like this, wouldnt it?

```js
Maybe = withProto(taggedSum)('Maybe, ['value'], {
  map: {
    Just: (fn, x) => Maybe.Just(fn(x)),
    Nothing: (fn, x) => x
  }
})
```

> I'm still trying to figure out if this is something I would use. [...] I also don't share your concerns, I believe, for `this`

I came up with this as I was learning all about category theory. I found it quite awkward that the most idiomatic way of creating types in JS was to use object-oriented style, so in the spirit of functional programming, I wanted to create types that are defined using pure functions. I also wanted to take care of the boilerplate involved in dealing with dependent types, e.g. `Just instanceof Maybe === true`, and automatically generating a `.equals` function.

I havent implemented it yet, but I thought it would be pretty neat to automatically derive missing prototype methods from the available ones, for example, we could recognize if you specify `of` and `chain` and drive `map` and `ap` for you!

```
ap: (mfn, m) => mfn.chain(f => m.map(f)),
map: (f, m) => m.chain(a => m.of(f(a))),
```

Or maybe this could be done better by explicitly defining the inheritance.

> So I'm not sure if this style is overkill for me.

Might definitely be overkill. ;)

paldepind:
> And so long as we're using pure functional programming patterns, then there should be no harm in partially applied functions being referentially transparent. That is:
```js
R.equals(R.partial(f, 1), R.partial(f, 1))
```
> Obviously they won't truely be referentially transparent in JavaScript, but that's why we have Ramda, right?

I might be misunderstanding something but it seems to me like this issue implies that in a purely functional setting we should be allowed to compare functions for equality. It would be wonderful if we could. But actually comparing functions for equality always _breaks_ referential transparency. As an example take a look at:

```js
R.equals(function(x) { return x; }, function(x) { return x; });
```

These functions are identical an yet `R.equals` returns `false`. That is because `R.equals` ends up comparing pointers but pointer equality makes no sense in a functional setting. But you note that yourself:

> To your point, equals((a, b) => a + b, (a, b) => a + b) will never be something we can do in JS. If you were working in Clojure, you could create a macro that would be able to successfully evaluate this though.

You are indeed right that you in a Lisp language could probably use a macro to solve the above problem. But how would you solve this problem:

```js
R.equals(
  (x: number) => x - 2 == x ? x - 2 : x;
  (x: number) => x;
);
```

These functions are also equal but detecting that with a macro is much harder. In general this problem is unsolvable (as @CrossEye has mentioned) because two functions `f` and `g` are equal if for any `x` it is the case that `f(x)` is equal to `g(x)`. But in any Turing complete language answering that question is impossible.

Thus if we write functional programs and care about referential transparency we should just forget about comparing functions for equality.

ccorcos:
@paldepind you're completely right. But think about pure functional user interfaces. You have this lazy tree and you need to pass this dispatch function down the tree, mapping over it along the way to contextualize the actions of each component. 

In order to lazily parse through and render this tree, you need to determine if certain branches havent changed. And since each branch is represented by a pure function and associated arguments (a thunk), you can simply compare the function and its arguments to see if its the same as before. Since one of the arguments is going to be a `dispatch` function, you'll need to be able to compare it with previous one. However, you're going to generate a new dispatch function every time when you map over it.

So I want to be clear. I hear your points, and I'm not really considering this something to merge into ramda at this point. But I do think there is a some specific value of being able to compare partially applied functions, i.e. thunks, so we can create pure lazy trees that can dispatch actions. That is, unless you know of a better way... But am I at least making sense?

CrossEye:
> But am I at least making sense?

Absolutely.  It would be awesome to be able to achieve this.

davidchambers:
Rather than partially apply a function, could one represent partial application as a value such as this?

```javascript
{function: R.add, arguments: [1]}
```

This could be compared to other `{ function :: Function, arguments :: Array Any }` values via `R.equals`. `r.function.apply(null, r.arguments)` would allow us to perform the (partial) function application when required.

CrossEye:
I am actually feeling reasonably well, and have decent internet connection.  If both hold up over the weekend, perhaps I'll try to get something in that offers some possibilities here.

ccorcos:
Good to hear :)

I suppose you could do that. Here's a go at it:

```js
function PAF(f, args...) {
  this.f = f
  this.args = args
}

function mapper(g, f, args) {
  return g(f(...args))
}

PAF.prototype.map = function(g) {
  return new PAF(mapper, g, this.f, this.args)
}

PAF.prototype.partial = function(more...) {
  return new PAF(this.f, this.args.concat(more))
}

PAF.prototype.equals = function(other) {
  return R.equals(other.f, self.f) && R.equals(other.args, self.args)
}

PAF.prototype.call = function(more...) {
  return this.f(...args)
}
```

I guess the main reason behind my original approach was so that you could do that stuff under the hood without the developer using the library having to know -- they can just call it like a real function rather than using `.call`

paldepind:
> So I want to be clear. I hear your points, and I'm not really considering this something to merge into ramda at this point.

Great. Then I think we agree :)

I think @davidchambers's idea is very good. However, I just had another though. I think what might actually solve @ccorcos problem is if our `R.memoize` function returned a curried function. Then the below would work:

```js
const fn = R.memoize((a, b) => {a, b});
fn(1)(2) === fn(1)(2) == fn(1, 2) // true
```

Then instead of using the `curryN` in this PR `R.memoize` would achieve the same result.

ccorcos:
In your example, I think you meant something more like this:

```js
const fn = R.memoize((a, b) => ({a, b}));
fn(1) === fn(1) // true
```

I tried playing around with WeakMap and memorization but I couldn't get it to work. The problem is that you end up mapping over a function multiple times as you pass dispatch down to deeper and deeper components, and if our views are pure, theres no way to save a reference of the previous memoized function. If I recall correctly, I started looking into making a WeakMap of WeakMaps was which point, I figured I was going in the wrong direction...

paldepind:
> In your example, I think you meant something more like this:

Except a missing `=` I wrote what I meant.

> The problem is that you end up mapping over a function multiple times as you pass dispatch down to deeper and deeper components, and if our views are pure, theres no way to save a reference of the previous memoized function.

Do you map over the view functions as well? Isn't it the result of the view functions you#re interested in comparing for pointer equality? I might be misunderstanding something.

> If I recall correctly, I started looking into making a WeakMap of WeakMaps was which point, I figured I was going in the wrong direction...

That doesn't sound bad for memorizing curried functions of multiple arguments?

ccorcos:
> Except a missing = I wrote what I meant.

Ah I see what you mean, since its memoized...

```js
const fn = R.memoize((a, b) => ({a, b}));
fn(1)(2) === fn(1)(2) == fn(1, 2) // true
fn(1) === fn(1) == f(1) // true
```

> Isn't it the result of the view functions you're interested in comparing for pointer equality? I might be misunderstanding something.

No, at least that wasnt what I was thinking. I so the goal is to create a thunk that can be lazily evaluated. For example, in React, they use `React.createElement(Component, props)` to create a lazy element -- they aren't actually rendering all the way down until they need to. They compare component and props and if they're all the same, then React knows its doesnt need to touch that entire branch of the tree since its all going to be the same.

So my idea was to so something similar, but rather than use something like `React.createElement`, i.e. `PAF(counter.view, PAF(dispatch).map(toCounterAction), props)`. The reason I didnt use `PAF` though was because I didnt want to have to belabor the developer with having to deal with all this stuff...

> That doesn't sound bad for memorizing curried functions of multiple arguments?

That's the idea. Maybe I should give this another shot then...

PR #1689, opened 2016-03-17
Adding index to iterator function

raminbp:
i think it's very helpful if we pass the index


davidchambers:
Use `R.addIndex(R.forEach)` to get this behaviour. ;)

See [`R.addIndex`][1] for details.


[1]: http://ramdajs.com/docs/#addIndex

PR #1692, opened 2016-03-18
until

### LINKED ISSUE #1690: until
Like `R.unless` but applies function until predicate is satisfied and passes new values to predicate each step.

Or `while` which is the same but with `complement` pred

```js
R.until(
    R.gt(R.__, 100),
    R.multiply(2)
)(1) === 128
```

iofjuupasli:
fixes #1690

English is not my native, and so I'm not sure what is clear description for this function.

### DIFF
--- a/src/until.js
+++ b/src/until.js
@@ -1,0 +1,28 @@
+var _curry3 = require('./internal/_curry3');
+
+
+/**
+ * Takes a predicate, a transformation function, and an initial value,
+ * and returns a value of the same type as the initial value.
+ * It does so by applying the transformation until the predicate is satisfied,
+ * at which point it returns the satisfactory value.
+ *
+ * @func
+ * @memberOf R
+ * @category Logic
+ * @sig (a -> Boolean) -> (a -> a) -> a -> a
+ * @param {Function} pred A predicate function
+ * @param {Function} fn The iterator function
+ * @param {*} init Initial value
+ * @return {*} Final value that satisfies predicate
+ * @example
+ *
+ *      R.until(R.gt(R.__, 100), R.multiply(2))(1) // => 128
+ */
+module.exports = _curry3(function until(pred, fn, init) {
+  var val = init;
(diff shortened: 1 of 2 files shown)


src/until.js
```diff
@@ -0,0 +1,25 @@
...
+ * @category Logic
+ * @sig (a -> Boolean) -> (a -> a) -> a -> a
+ * @param {Function} pred A predicate function
+ * @param {Function} fn The iterator function
+ * @param {*} init Initial value
+ * @return {*} Final value that satisfies predicate
+ * @example
+ *
+ *      R.until(R.gt(R.__, 100), R.multiply(2))(1) // => 128
+ *      R.until(R.none(R.isArrayLike), R.unnest)([1, [2], [[3]]]) // => [1, 2, 3]
```
davidchambers: Let's remove this example as [`R.flatten`][1] would be much clearer. ;)

The first example is great!


[1]: http://ramdajs.com/docs/#flatten

davidchambers:
For the description, I suggest something like this:

_Takes a predicate, a transformation function, and an initial value, and returns a value of the same type as the initial value. It does so by applying the transformation until the predicate is satisfied, at which point it returns the satisfactory value._

CrossEye:
This is great.  Thank you.

I like David's description, except that I would substitute "Accepts" for "Takes".  I always wonder where functions are taking things and if thievery is involved.  :smile:

I don't know if a additional sentence is in order warning about the possibility of infinite loops.  I know it's possible in other Ramda functions, but this one seems more likely.

iofjuupasli:
Does it need anything else?

CrossEye:
:herb:  I'm happy with this, unless someone wants to do some bikeshedding on the name.

buzzdecafe:
gotta squash the commits then, please

CrossEye:
Thanks, @iofjuupasli!

PR #1693, opened 2016-03-18
Update example for REPL usage per #70 on docs

GantMan:
as indicated in https://github.com/ramda/ramda.github.io/pull/70


buzzdecafe:
LGTM. This is to avoid a collision between the example variable name (`append`) colliding with a const variable from `R`.

CrossEye:
Thank you!  Nice catch.

:herb:

Issue #1582, opened 2016-01-06
Inconsistency between assoc and merge


ivan-kleshnin:
Starting with usecase... I need to remove trailing hash from URL (for scraping). 
Node `url` module is able to parse and format urls but gives no way to immutably nullify hash. 
Let's assume I need parsed URL down the code so I won't use regexes.

---

Let's try `R.assoc`

``` js
import Url from "url";

let url = "https://github.com/somepage#trash"
R.assoc("url", null, Url.parse(url)).format();
// => https://github.com/somepage
```

**It works** because `assoc` copies all props including functions.
`.format` method was copied to new object.

``` js
// === assoc source code ===
for (var p in obj) {
  result[p] = obj[p];
}
// === === === === === ===
```

---

Let's try `R.merge`

``` js
import Url from "url";

let url = "http://github.com/somepage#trash";
R.merge(Url.parse(url), {url: null}).format();
// => **format is not a function**
```

**It does not work** because `merge` copies only `enumerable` props omitting functions.
`.format` method was not copied to new object.

``` js
// === merge source code ===
var result = {};
var ks = keys(a); // a is original obj
// === === === === === ===
```

---

I think we should apply same politics to both functions and they should give equal results in this case.
Other Ramda functions may also be affected.

But I'm not sure which behavior is "correct" and which one should be "fixed". 
In this case `assoc` turned out to be more useful as `merge` failed to patch Object-Oriented code. 
But we surely need to account more use cases.

CrossEye:
> I think we should apply same politics to both functions and they should give equal results in this case.
Other Ramda functions may also be affected.

Why do you think these should act the same?  I see no strong arguments either for or against this.  If I were to choose, I'd rather stick with the properties delineated by `keys`, although that would not work for your case.  But I'm not sure I see any great reason yet to make the choice.

asaf-romano:
Note that in your case the reason ``format`` is not copied over isn't because it's set as non-enumerable, but rather because it's part of the object prototype, rather than the object itself:

```js
Object.keys(Object.getPrototypeOf(require("url").parse("http://a.tld"))); // [ 'parse', 'format', 'resolve', 'resolveObject', 'parseHost' ]
```

[...]

PR #1699, opened 2016-03-26
Changes assoc so it does not flatten object prototype properties into…

### LINKED ISSUE #1697: about assoc prototype flattening
As is documented, assoc flattens prototype properties into the new object.

I'm using Ramda in a MeteorJS project. I have objects that have helpers which are assigned as properties in the prototype and a schema validation that checks the structure of the object before saving it to the database. Whenever I do an operation in an object using assoc it becomes invalid because of the prototype properties flattening.

I know this is not really a Ramda issue, but I think this behavior is not really what you would expect when you're just trying to change a property in an object.

Would the performance cost be too high if we were to set the prototype properties on the new object's prototype or is there something I'm missing?

VitorHP:
… the new object. Fixes #1697 

Implemented a version of assoc the does not flatten prototype properties.

It runs 2-3x slower than the original one however it's still pretty fast. 

There's some measuring [here](https://jsfiddle.net/wuqq57mk/2/)


asaf-romano:
Functionality-wise, I think this change is desired (see previous discussion on https://github.com/ramda/ramda/issues/1582), but only if it's applied across-the-board for all cloning operations. If we can abstract the implementation so that it's consolidated in one place, then we can also keep a fast-lane for POJOs.

CrossEye:
I have several issues with the commit as presented, but let's discuss this in terms of _whether_ in #1697 before thinking too much about _how_.

src/assoc.js
```diff
@@ -22,10 +21,14 @@ var _curry3 = require('./internal/_curry3');
  *      R.assoc('c', 3, {a: 1, b: 2}); //=> {a: 1, b: 2, c: 3}
  */
 module.exports = _curry3(function assoc(prop, val, obj) {
-  var result = {};
-  for (var p in obj) {
-    result[p] = obj[p];
+  var result = Object.create(Object.getPrototypeOf(obj))
+  var properties = Object.getOwnPropertyNames(obj)
```
davidchambers: `Object.create`, `Object.getPrototypeOf`, and `Object.getOwnPropertyNames` are not defined in ES3, which Ramda targets.

buzzdecafe: > ES3, which Ramda targets.

all in favor of dropping that constraint? :hand:

jdalton: `Object.getOwnPropertyNames` gets non-enumerable props. If you go down that route you'll want to use `Object.defineProperty` and `Object.getOwnPropertyDescriptor` to ensure consistency.

PR #1700, opened 2016-03-26
Use native Object.assign if available

asaf-romano:
My fault. It turns out that the polyfill isn’t nearly as fast as the native implementation (at least in Firefox).

### DIFF
--- a/src/internal/_assign.js
+++ b/src/internal/_assign.js
@@ -1,0 +1,4 @@
+var _objectAssign = require('./_objectAssign');
+
+module.exports =
+  typeof Object.assign === 'function' ? Object.assign : _objectAssign;
--- a/src/internal/_stepCat.js
+++ b/src/internal/_stepCat.js
@@ -1,6 +1,6 @@
+var _assign = require('./_assign');
 var _identity = require('./_identity');
 var _isTransformer = require('./_isTransformer');
-var _objectAssign = require('./_objectAssign');
 var isArrayLike = require('../isArrayLike');
 var objOf = require('../objOf');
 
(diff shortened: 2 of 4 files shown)


src/internal/_objectAssign.js
```diff
@@ -1,24 +1,28 @@
 var _has = require('./_has');
 
-// Based on https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/assign
-module.exports = function _objectAssign(target) {
-  if (target == null) {
-    throw new TypeError('Cannot convert undefined or null to object');
-  }
+module.exports = (function() {
+  return typeof Object.assign === 'function' ?
```
buzzdecafe: does it make sense to do this conditional logic in its own module and then use `internal/_assign` instead?

asaf-romano: I followed `keys` here. I cannot think of a use case for the standalone module, but it if it helps readability it's probably worth it anyway.

jdalton: Builtin `Object.assign` will also copy symbols over.

CrossEye:
@jdalton:

> Builtin `Object.assign` will also copy symbols over.

That's an interesting point.  We've been mostly ignoring Symbols.  It's hard to come to terms with them from an FP perspective.  At some point we're going to have to figure out what we want to do, though.

jdalton:
While implementation perf of `Object.assign` is improving its baggage has kept me from using it.

CrossEye:
:herb:

src/internal/_assign.js
```diff
@@ -0,0 +1,5 @@
+var _objectAssign = require('./_objectAssign');
+
+module.exports = (function() {
+  return typeof Object.assign === 'function' ? Object.assign : _objectAssign;
+}());
```
davidchambers: Is the IIFE necessary?

asaf-romano: Nope.

Issue #1688, opened 2016-03-16
Why don't we have the Haskell's groupBy function?


kahless:
Use case
========

I have an array of objects. Each object contains a parsed street address like this one:
```
{
   country: 'Russia',
   house_number: '5/6',
   road: "Tverskaya Street",
   ...
   theatre: "Театр им. М.Н. Ермоловой"
}
```

The problem is that some addresses are too specific. For example I don't need to know about the theatre that is located there. From this point of view my array contains duplicates (theatre, bakery, hotel and others that are located in the same building). And I need to replace this duplicates with one, least specific address, with the same house number but without unwanted information about theatre and others. There is always one such address for all the duplicates.

In Haskell we have a wonderful groupBy function that can help me locate a sequence of duplicates and reduce them to one address by a predicate. But Ramda's groupBy returns an object so I lose the original order. But order is important because addresses were sorted by relevance.

I am sorry for my english and possibly excessive verbosity.

TheLudd:
Perhaps these functions will come in handy?
http://ramdajs.com/0.19.1/docs/#uniqBy
http://ramdajs.com/0.19.1/docs/#uniqWith

kahless:
I tried it, but these functions won't let me choose a certain address that will be in place of the duplicates. It would be just the first one. But in my case the desired address usually is the last duplicate (but it's position is not guaranteed).

kahless:
This is the missing groupBy function from Haskell:
http://hackage.haskell.org/package/base-4.8.2.0/docs/Data-List.html#v:groupBy

TheLudd:
It seems to me to have the same signature as `uniqWith`

kahless:
Nop. In Haskell it returns a list of lists. In Ramda uniqWith returns just an array:
```
Ramda's uniqWith  — (a, a → Boolean) -> [a] -> [a]
Haskell's groupBy — (a -> a -> Bool) -> [a] -> [[a]]
```

For example:
```
groupBy (==) "Mississippi"
   -- Returns ["M","i","ss","i","ss","i","pp","i"]
R.uniqWith(R.eq, ['M','i','s','s','i','s','s','i','p','p','i'])
   // Returns ["M", "i", "s", "p"]
```

CrossEye:
[...]

PR #1702, opened 2016-03-28
add R.groupWith

benperez:
Closes https://github.com/ramda/ramda/issues/1688

### DIFF
--- a/src/groupWith.js
+++ b/src/groupWith.js
@@ -1,0 +1,41 @@
+var _curry2 = require('./internal/_curry2');
+
+/**
+ * Takes a list and returns a list of lists where each sublist's elements are
+ * all "equal" according to the provided equality function.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig (a, a -> Boolean) -> [a] -> [[a]]
+ * @param {Function} fn Function for determining whether two given (adjacent)
+ *        elements should be in the same group
+ * @param {Array} list The array to group. Also accepts a string, which will be
+ *        treated as a list of characters.
+ * @return {List} A list that contains sublists of equal elements,
+ *         whose concatenations is equal to the original list.
+ * @example
+ *
+ *    groupWith(R.equals, [0, 1, 1, 2, 3, 5, 8, 13, 21])
+ *    // [[0], [1, 1], [2, 3, 5, 8, 13, 21]]
+ *
+ *    groupWith((a, b) => a % 2 === b % 2, [0, 1, 1, 2, 3, 5, 8, 13, 21])
+ *    // [[0], [1, 1], [2], [3, 5], [8], [13, 21]]
+ *
+ *    R.groupWith(R.eqBy(isVowel), 'aestiou')
+ *    // ['ae', 'st', 'iou']
+ */
+module.exports = _curry2(function(fn, list) {
+  var res = [];
+  var idx = 0;
+  var len = list.length;
+  while (idx < len) {
+    var nextidx = idx + 1;
+    while (nextidx < len && fn(list[idx], list[nextidx])) {
+      nextidx += 1;
+    }
+    res.push(list.slice(idx, nextidx));
+    idx = nextidx;
+  }
+  return res;
+});
--- a/test/groupWith.js
+++ b/test/groupWith.js
@@ -1,0 +1,28 @@
+var R = require('..');
+var eq = require('./shared/eq');
+
+describe('groupWith', function() {
+
+  it('splits the list into groups according to the grouping function', function() {
+    eq(R.groupWith(R.equals, [1, 2, 2, 3]), [[1], [2, 2], [3]]);
+    eq(R.groupWith(R.equals, [1, 1, 1, 1]), [[1, 1, 1, 1]]);
+    eq(R.groupWith(R.equals, [1, 2, 3, 4]), [[1], [2], [3], [4]]);
+  });
+
+  it('returns an empty array if given an empty array', function() {
+    eq(R.groupWith(R.equals, []), []);
+  });
+
+  it('can be turned into the original list through concatenation', function() {
+    var list = [1, 1, 2, 3, 4, 4, 5, 5];
+    eq(R.unnest(R.groupWith(R.equals, list)), list);
+    eq(R.unnest(R.groupWith(R.complement(R.equals), list)), list);
+    eq(R.unnest(R.groupWith(R.T, list)), list);
+    eq(R.unnest(R.groupWith(R.F, list)), list);
+  });
+
+  it('also works on strings', function() {
+    eq(R.groupWith(R.equals)('Mississippi'), ['M','i','ss','i','ss','i','pp','i']);
+  });
+
+});


benperez:
While writing this I wonder whether we should also include a `group` function:
```javascript
R.group = R.groupWith((a, b) => a === b);
```
 since this seems to be the most useful application of groupWith and also a bit tedious to write out.

src/groupWith.js
```diff
@@ -0,0 +1,35 @@
...
+ * @sig (a, a -> Boolean) -> [a] -> [[a]]
+ * @param {Function} fn Function :: a, a -> Boolean
+ * @param {Array} list The array to group
+ * @return {List} A list that contains sublists of equal elements,
+ *         whose concatenations is equal to the original list.
+ * @example
+ *
+ *      var groupWithLetters = R.groupWith(function(letter1, letter2) {
+ *        return letter1 === letter2;
+ *      });
```
davidchambers: Shall we use `R.equals` in place of the lambda?

src/groupWith.js
```diff
@@ -0,0 +1,35 @@
...
+ *      var letters = ["M", "i", "s", "s", "i", "s", "s", "i", "p", "p", "i"];
+ *      groupWithLetters(letters);
+ *      // ["M","i","ss","i","ss","i","pp","i"]
```
davidchambers: The result is actually `[['M'], ['i'], ['s', 's'], ['i'], ['s', 's'], ['i'], ['p', 'p'], ['i']]`.

src/groupWith.js
```diff
@@ -0,0 +1,35 @@
...
+ */
+module.exports = _curry2(function(fn, list) {
+  var result = [];
+  var idx = 0;
+  while (idx < list.length) {
+    var nextidx = idx;
+    while (fn(list[idx], list[nextidx]) && nextidx < list.length) {
```
davidchambers: We should switch the order of these checks. As written, `undefined` will be provided as the second argument to `fn` when the end of the list is reached. The behaviour of `fn` in such cases may be undefined.

src/groupWith.js
```diff
@@ -0,0 +1,35 @@
...
+      nextidx = nextidx + 1;
```
davidchambers: Let's use `+= 1` here.

src/groupWith.js
```diff
@@ -0,0 +1,35 @@
...
+ *      });
+ *      var letters = ["M", "i", "s", "s", "i", "s", "s", "i", "p", "p", "i"];
+ *      groupWithLetters(letters);
+ *      // ["M","i","ss","i","ss","i","pp","i"]
+ */
+module.exports = _curry2(function(fn, list) {
+  var result = [];
+  var idx = 0;
+  while (idx < list.length) {
+    var nextidx = idx;
```
davidchambers: Could we use `idx + 1` here as the initial value?

src/groupWith.js
benperez: Nice one, I forgot about that function...

test/groupWith.js
```diff
@@ -0,0 +1,31 @@
...
+
+
+describe('groupWith', function() {
+  var tripleEquals = function(a, b) {
+    return a === b;
+  };
+
+  it('splits the list into groups according to the grouping function', function() {
+    var list1 = [1, 2, 2, 3];
+    assert.deepEqual(R.groupWith(tripleEquals, list1), [[1], [2, 2], [3]]);
```
davidchambers: Let's use [`eq`][1] here instead.


[1]: https://github.com/ramda/ramda/blob/v0.20.0/test/shared/eq.js

src/groupWith.js
benperez: We could, although we'd have to remember to push `list[idx]`onto the resultant array in that case though.

test/groupWith.js
```diff
@@ -0,0 +1,31 @@
...
+    assert.deepEqual(R.groupWith(tripleEquals, list2), [[1, 1, 1, 1]]);
+
+    var list3 = [1, 2, 3, 4];
+    assert.deepEqual(R.groupWith(tripleEquals, list3), [[1], [2], [3], [4]]);
+  });
+
+  it('is curried', function() {
+    var groupWithTripleEquals = R.groupWith(tripleEquals);
+    assert.deepEqual(groupWithTripleEquals([1, 1, 2, 2]), [[1, 1], [2, 2]]);
+  });
```
davidchambers: I'm in favour of omitting this test. We already have a [test][1] which asserts that every function is curried.


[1]: https://github.com/ramda/ramda/blob/v0.20.0/test/invariants.js#L27-L42

test/groupWith.js
```diff
@@ -0,0 +1,31 @@
+var assert = require('assert');
+
+var R = require('..');
+
+
+describe('groupWith', function() {
+  var tripleEquals = function(a, b) {
+    return a === b;
+  };
```
davidchambers: Let's use `R.equals` instead.

src/groupWith.js
benperez: Actually, we wouldn't, `idx + 1` works well here.

davidchambers: > we'd have to remember to push `list[idx]`onto the resultant array

I don't think so. As written we first check whether `fn(list[idx], list[idx])` evaluates `true`, which should always be the case. Using `idx + 1` as the initial value will skip this step without affecting the result (for any well-behaved `fn`, at least).

davidchambers: We should include tests to assert the following property:

  - `forall f :: (a, b) -> Boolean. R.unnest(R.groupWith(f, xs)) = xs`

Good candidates for `f` are `R.equals`, `R.complement(R.equals)`, `R.T`, and `R.F`.

benperez:
:zap:

Here's the [haskell implementation](https://hackage.haskell.org/package/base-4.8.2.0/docs/src/Data.OldList.html#groupBy) btw
```haskell
groupBy                 :: (a -> a -> Bool) -> [a] -> [[a]]
groupBy _  []           =  []
groupBy eq (x:xs)       =  (x:ys) : groupBy eq zs
                           where (ys,zs) = span (eq x) xs
```

src/groupWith.js
```diff
@@ -0,0 +1,33 @@
...
+ * @param {Function} fn Function :: a, a -> Boolean
+ * @param {Array} list The array to group
+ * @return {List} A list that contains sublists of equal elements,
+ *         whose concatenations is equal to the original list.
+ * @example
+ *
+ *      var groupWithLetters = R.groupWith(R.equals);
+ *      var letters = ["M", "i", "s", "s", "i", "s", "s", "i", "p", "p", "i"];
+ *      groupWithLetters(letters);
+ *      // [['M'], ['i'], ['s', 's'], ['i'], ['s', 's'], ['i'], ['p', 'p'], ['i']]
```
davidchambers: I'd prefer this example:

```javascript
> groupWith((a, b) => a % 2 === b % 2, [0, 1, 1, 2, 3, 5, 8, 13, 21])
[[0], [1, 1], [2], [3, 5], [8], [13, 21]]
```

test/groupWith.js
```diff
@@ -0,0 +1,21 @@
...
+    eq(R.groupWith(R.equals, list1), [[1], [2, 2], [3]]);
+
+    var list2 = [1, 1, 1, 1];
+    eq(R.groupWith(R.equals, list2), [[1, 1, 1, 1]]);
+
+    var list3 = [1, 2, 3, 4];
+    eq(R.groupWith(R.equals, list3), [[1], [2], [3], [4]]);
+  });
+
+  it('returns an empty object if given an empty array', function() {
```
davidchambers: An empty array, do you mean?

test/groupWith.js
```diff
@@ -0,0 +1,21 @@
...
+
+  it('splits the list into groups according to the grouping function', function() {
+    var list1 = [1, 2, 2, 3];
+    eq(R.groupWith(R.equals, list1), [[1], [2, 2], [3]]);
+
+    var list2 = [1, 1, 1, 1];
+    eq(R.groupWith(R.equals, list2), [[1, 1, 1, 1]]);
+
+    var list3 = [1, 2, 3, 4];
+    eq(R.groupWith(R.equals, list3), [[1], [2], [3], [4]]);
```
davidchambers: I'd prefer not to name the lists:

```javascript
eq(R.groupWith(R.equals, [1, 2, 2, 3]), [[1], [2, 2], [3]]);
eq(R.groupWith(R.equals, [1, 1, 1, 1]), [[1, 1, 1, 1]]);
eq(R.groupWith(R.equals, [1, 2, 3, 4]), [[1], [2], [3], [4]]);
```

test/groupWith.js
```diff
@@ -0,0 +1,29 @@
...
+  });
+
+  it('returns an empty object if given an empty array', function() {
+    eq(R.groupWith(R.equals, []), []);
+  });
+
+  it('can be turned into the original list through concatentation', function() {
```
davidchambers: `s/concatentation/concatenation/`

test/groupWith.js
benperez: whoops, yes.

src/groupWith.js
```diff
@@ -0,0 +1,33 @@
+var _curry2 = require('./internal/_curry2');
+
+/**
+ * Takes a list and returns a list of lists where each sublist contains only equal elements.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig (a, a -> Boolean) -> [a] -> [[a]]
+ * @param {Function} fn Function :: a, a -> Boolean
```
davidchambers: I don't think we should include Hindley–Milner notation in `@param` descriptions. Users who rely on these descriptions likely haven't yet learnt to read HM.

src/groupWith.js
```diff
@@ -0,0 +1,33 @@
+var _curry2 = require('./internal/_curry2');
+
+/**
+ * Takes a list and returns a list of lists where each sublist contains only equal elements.
```
davidchambers: It's incorrect to write "equal elements" here, since the grouping function is provided by the caller.

src/groupWith.js
benperez: I think an example that uses `R.equals` is nice since it might be a little more straightforward. I don't feel super strongly about this, though.

src/groupWith.js
benperez: OK, I don't 100% understand which of these `@param` descriptions are required or what's expected in them. I based this docstring off of [`groupBy`](https://github.com/ramda/ramda/blob/35f9094d94865d1b368d4a8399223a0261508a3d/src/groupBy.js#L22)

src/groupWith.js
```diff
@@ -0,0 +1,33 @@
...
+ */
+module.exports = _curry2(function(fn, list) {
+  var result = [];
+  var idx = 0;
+  while (idx < list.length) {
+    var nextidx = idx + 1;
+    while (nextidx < list.length && fn(list[idx], list[nextidx])) {
+      nextidx += 1;
+    }
+    result.push(list.slice(idx, nextidx));
```
davidchambers: Shall we document the fact that this function also works on lists of strings? For example:

```javascript
> R.groupWith(R.useWith(R.equals, [isVowel, isVowel]), 'benperez')
['b', 'e', 'np', 'e', 'r', 'e', 'z']
```

src/groupWith.js
davidchambers: The problem I have with *only* using `R.equals` is that at a glance the function appears to be specialized for grouping contiguous equal values.

Let's have our :cake: and eat it too!

```javascript
> groupWith(R.equals, [0, 1, 1, 2, 3, 5, 8, 13, 21])
[[0], [1, 1], [2, 3, 5, 8, 13, 21]]

> groupWith((a, b) => a % 2 === b % 2, [0, 1, 1, 2, 3, 5, 8, 13, 21])
[[0], [1, 1], [2], [3, 5], [8], [13, 21]]
```

src/groupWith.js
benperez: I struggled with the explanation for this function. I think it's easier to explain as the version of `group` that allows a user-supplied function. How does this sound?
> Takes a list and returns a list of lists where each sublist contains only elements which satisfy the user supplied equality function.

src/groupWith.js
davidchambers: A description such as _function for determining whether two given (adjacent) elements should be in the same group_ would be appropriate.

src/groupWith.js
davidchambers: > contains only elements which satisfy

This is suggestive of *filtering*.

This would be clearer, I think:

_Takes a list and returns a list of lists where each sublist's elements are all "equal" according to the provided equality function._

src/groupWith.js
benperez: I added an example along the lines of the one you've given above, I'm not quite sure where to add it to the _documentation_ though. I added a sentence about strings to the `@param` section of the docstring.

test/groupWith.js
CrossEye: What have you got against tents?

CrossEye:
LVGTM.  :herb:

src/groupWith.js
```diff
@@ -0,0 +1,40 @@
...
+ * @example
+ *
+ *      groupWith(R.equals, [0, 1, 1, 2, 3, 5, 8, 13, 21])
+ *     // [[0], [1, 1], [2, 3, 5, 8, 13, 21]]
+ *
+ *     groupWith((a, b) => a % 2 === b % 2, [0, 1, 1, 2, 3, 5, 8, 13, 21])
+ *     // [[0], [1, 1], [2], [3, 5], [8], [13, 21]]
+ *
+ *    R.groupWith(R.useWith(R.equals, [isVowel, isVowel]), 'aestiou')
+ *    // [['a', 'e'], ['s', 't'], ['i', 'o', 'u']]
```
davidchambers: Doesn't this give `['ae', 'st', 'iou']`?

src/groupWith.js
```diff
@@ -0,0 +1,40 @@
...
+ * @param {Function} fn Function for determining whether two given (adjacent)
+ *        elements should be in the same group
+ * @param {Array} list The array to group. Also accepts a string, which will be
+ *        treated as a list of characters.
+ * @return {List} A list that contains sublists of equal elements,
+ *         whose concatenations is equal to the original list.
+ * @example
+ *
+ *      groupWith(R.equals, [0, 1, 1, 2, 3, 5, 8, 13, 21])
+ *     // [[0], [1, 1], [2, 3, 5, 8, 13, 21]]
```
davidchambers: :triangular_ruler:

src/groupWith.js
CrossEye: Should we be using our internal `_slice` here?

If we did, of course, it would end up with what's in the documentation now, `[['a', 'e'], ['s', 't'], ['i', 'o', 'u']]` rather than what David points out it should now return of `['ae', 'st', 'iou']`.

The question is whether this should work on anything list-like or just lists (and strings).

src/groupWith.js
```diff
@@ -0,0 +1,40 @@
...
+ *         whose concatenations is equal to the original list.
+ * @example
+ *
+ *     groupWith(R.equals, [0, 1, 1, 2, 3, 5, 8, 13, 21])
+ *     // [[0], [1, 1], [2, 3, 5, 8, 13, 21]]
+ *
+ *     groupWith((a, b) => a % 2 === b % 2, [0, 1, 1, 2, 3, 5, 8, 13, 21])
+ *     // [[0], [1, 1], [2], [3, 5], [8], [13, 21]]
+ *
+ *    R.groupWith(R.useWith(R.equals, [isVowel, isVowel]), 'aestiou')
```
asaf-romano: ``R.eqBy(isVowel)``?

src/groupWith.js
```diff
@@ -0,0 +1,40 @@
...
+ *    // [['a', 'e'], ['s', 't'], ['i', 'o', 'u']]
+ */
+module.exports = _curry2(function(fn, list) {
+  var result = [];
+  var idx = 0;
+  while (idx < list.length) {
```
asaf-romano: list.length should be cached (see https://github.com/ramda/ramda/pull/1192)

src/groupWith.js
```diff
@@ -0,0 +1,40 @@
...
+ *         whose concatenations is equal to the original list.
+ * @example
+ *
+ *    groupWith(R.equals, [0, 1, 1, 2, 3, 5, 8, 13, 21])
+ *    // [[0], [1, 1], [2, 3, 5, 8, 13, 21]]
+ *
+ *    groupWith((a, b) => a % 2 === b % 2, [0, 1, 1, 2, 3, 5, 8, 13, 21])
+ *    // [[0], [1, 1], [2], [3, 5], [8], [13, 21]]
+ *
+ *    R.groupWith(R.useWith(R.equals, [isVowel, isVowel]), 'aestiou')
```
asaf-romano: ``R.eqBy(isVowel)``

test/groupWith.js
```diff
@@ -0,0 +1,28 @@
...
+    eq(R.groupWith(R.equals, [1, 1, 1, 1]), [[1, 1, 1, 1]]);
+    eq(R.groupWith(R.equals, [1, 2, 3, 4]), [[1], [2], [3], [4]]);
+  });
+
+  it('returns an empty array if given an empty array', function() {
+    eq(R.groupWith(R.equals, []), []);
+  });
+
+  it('can be turned into the original list through concatenation', function() {
+    var list = [1, 1, 2, 3, 4, 4, 5, 5];
```
asaf-romano: This test is a little strange. It basically tests ``unnest`` (or ``equals``) provided that the previous tests passed.

test/groupWith.js
```diff
@@ -0,0 +1,28 @@
...
+    eq(R.unnest(R.groupWith(R.equals, list)), list);
+    eq(R.unnest(R.groupWith(R.complement(R.equals), list)), list);
+    eq(R.unnest(R.groupWith(R.T, list)), list);
+    eq(R.unnest(R.groupWith(R.F, list)), list);
+  });
+
+  it('also works on strings', function() {
+    eq(R.groupWith(R.equals, 'Mississippi'), ['M','i','ss','i','ss','i','pp','i']);
```
asaf-romano: let's curry R.groupWith(...) at least once; here is probably a good place.

asaf-romano:
Looks good otherwise 🌷

test/groupWith.js
benperez: These tests were added at the request of @davidchambers to demonstrate this property:
> `forall f :: (a, b) -> Boolean. R.unnest(R.groupWith(f, xs)) = xs`

asaf-romano:
By the way, While I see no problem with supporting the string-mode (it's a nice touch), is there a sound use case for it? If not, it might worth omitting it from the documentation (the example itself doesn't hurt); Then the signature becomes a little less complex: list input; list-of-lists output.

test/groupWith.js
asaf-romano: Ah, ok.

benperez:
> If not, it might be worth omitting it from the documentation

Support for strings happens to be an accident of the implementation. Having said that, the only mention in the documentation is in that one example the `@param` section:
> @param {Array} list The array to group. Also accepts a string, which will be treated as a list of characters.

If it simplifies my life in getting this pull request accepted, I'll just leave it out of the documentation.

asaf-romano:
I think you're past that point anyway ;)

davidchambers:
:deciduous_tree:

buzzdecafe:
only reservation is how different this is from `groupBy`. if i just missed that discussion, then nm

benperez:
It was discussed in #1688 .

`R.groupBy` takes a hashing function and a list of elements. It returns a mapping of hashed values to lists of elements.
`R.groupWith` takes a comparison function and a list of elements. It returns a list of lists, where each sublist contains contiguous elements that return true when applied to the comparison function.

buzzdecafe:
ah thanks for the reminder @benperez

ok then, nm

PR #1704, opened 2016-04-01
Optimize groupBy to use mutation in reducer

ThomWright:
Using `R.groupBy` on large arrays can get pretty slow.

Simply using `arr.push()` instead of `R.append` improves performance significantly. Since the accumulator in the reducer is internal, this mutation shouldn't be a problem.

See [this perf test](http://jsperf.com/ramda-groupby-slow-for-large-arrays/2)

All credit to @davidgtonge

Also, I noticed `_xgroupBy.js`. Should this be updated too as part of this PR? I couldn't immediately see if it was safe to mutate instead of using `append` in that case.

### DIFF
--- a/src/groupBy.js
+++ b/src/groupBy.js
@@ -1,9 +1,7 @@
 var _curry2 = require('./internal/_curry2');
 var _dispatchable = require('./internal/_dispatchable');
-var _reduce = require('./internal/_reduce');
 var _xgroupBy = require('./internal/_xgroupBy');
-var append = require('./append');
-
+var reduceBy = require('./reduceBy');
 
 /**
  * Splits a list into sub-lists stored in an object, based on the result of
@@ -45,10 +43,10 @@ var append = require('./append');
  *      //   'F': [{name: 'Eddy', score: 58}]
  *      // }
  */
-module.exports = _curry2(_dispatchable('groupBy', _xgroupBy, function groupBy(fn, list) {
-  return _reduce(function(acc, elt) {
-    var key = fn(elt);
-    acc[key] = append(elt, acc[key] || (acc[key] = []));
-    return acc;
-  }, {}, list);
-}));
+module.exports = _curry2(_dispatchable('groupBy', _xgroupBy, reduceBy(function(acc, item) {
+  if (acc == null) {
+    acc = [];
+  }
+  acc.push(item);
+  return acc;
+}, null)));


buzzdecafe:
LGTM
:cow2:

asaf-romano:
I wonder if it's worth using reduceBy like this:
```js
const groupBy = curry(
  (keyFn, list) => reduceBy(keyFn, (a, c) => { a.push(c); return a; }, [], list));
```
http://goo.gl/5FeSEy

With the reason being that it's better to have one hack per file.

ThomWright:
Defining `groupBy` in terms of `reduceBy` seems like a good idea. I'll update the PR if there's a consensus.

src/groupBy.js
```diff
@@ -48,7 +47,10 @@ var append = require('./append');
 module.exports = _curry2(_dispatchable('groupBy', _xgroupBy, function groupBy(fn, list) {
   return _reduce(function(acc, elt) {
     var key = fn(elt);
-    acc[key] = append(elt, acc[key] || (acc[key] = []));
+    if (acc[key] == null) {
```
davidchambers: This check will fail for `key` values such as `'toString'` and `'valueOf'`. I'll fix this in a separate pull request.

davidchambers: Actually, the existing code *happens* to do the right thing:

```javascript
> R.append(1, Object.prototype.toString)
[1]
```

Let's include an assertion such as `eq(R.groupBy(R.identity, ['toString']), {toString: ['toString']})`, and update the check to use `_has`.

asaf-romano: @davidchambers if we change this to be defined in terms of "reduceBy" it's not going to be an issue, because it uses "_has", which, IIIRC, looks only on the object's own keys.

buzzdecafe:
implementing in terms of `reduceBy` sounds good. if this happens timely i'd like to release it in 0.20.2

ThomWright:
@asaf-romano That implementation using `reduceBy` and `Array.push` doesn't work.

Using `R.append` works, but we're back at square one :smile: 

See [example](http://ramdajs.com/repl/#?code=const%20groupByWithReduceBy%20%3D%20curry%28%0A%20%20%28keyFn%2C%20list%29%20%3D%3E%20reduceBy%28keyFn%2C%20%28a%2C%20c%29%20%3D%3E%20%7B%20return%20append%28c%2C%20a%29%20%7D%2C%20%5B%5D%2C%20list%29%29%3B%0A%0AgroupByWithReduceBy%28modulo%28__%2C%205%29%2C%20range%281%2C%20100%29%29%0A%0A%0A)

asaf-romano:
WFM: http://goo.gl/HTRUlb

Note Array.push doesn't return the array, so it's not a one liner.

davidgtonge:
@asaf-romano that example isn't working, as the same empty array gets re-used.
Looking at the code, the only suggestion that I have is that `reduceBy` clones the `valueAcc` argument.
See this working example: http://goo.gl/KCHFyy

I imagine that this issue with reduceBy could cause issues for other people as well, essentially it only works if the `valueFn` doesn't mutate.

davidgtonge:
Or - another option that doesn't require a change to reduceBy: http://goo.gl/F1F3zL
The valueFn becomes: 
```javascript
(a, c) => {
    (a.length) ? a.push(c) : a = [c]
    return a;
}
```

asaf-romano:
Nice catch.

Another option is to provide ``null`` initial value an then the condition becomes a little simpler. I don't think it's a good idea to clone the initial value in ``reduceBy``.

asaf-romano:
By the way, ``countBy`` and ``indexBy`` could also be defined in ``reduceBy`` terms (and there we won't need any null-initial-value trick). It'd be really nice to get rid of the code duplication between the four methods.

buzzdecafe:
> It'd be really nice to get rid of the code duplication between the four methods.

:+1:

asaf-romano:
However, see #1706.

asaf-romano:
That change landed.

ThomWright:
PR updated with this implementation:

```js
reduceBy(function(acc, item) {
  acc ? acc.push(item) : acc = [item]
  return acc;
}, null)
```

src/groupBy.js
```diff
@@ -45,10 +43,7 @@ var append = require('./append');
...
  *      // }
  */
-module.exports = _curry2(_dispatchable('groupBy', _xgroupBy, function groupBy(fn, list) {
-  return _reduce(function(acc, elt) {
-    var key = fn(elt);
-    acc[key] = append(elt, acc[key] || (acc[key] = []));
-    return acc;
-  }, {}, list);
-}));
+module.exports = _curry2(_dispatchable('groupBy', _xgroupBy, reduceBy(function(acc, item) {
```
asaf-romano: I think you can remove _curry2.

ThomWright:
Of course! Done.

asaf-romano:
Sorry, I think I was wrong. It looks like you've to re-curry because _dispatchable doesn't retain arity. I'll file a separate issue for that, but for now, let's just _curry2.

asaf-romano:
Looks good otherwise.

src/groupBy.js
```diff
@@ -45,10 +42,7 @@ var append = require('./append');
...
  */
-module.exports = _curry2(_dispatchable('groupBy', _xgroupBy, function groupBy(fn, list) {
-  return _reduce(function(acc, elt) {
-    var key = fn(elt);
-    acc[key] = append(elt, acc[key] || (acc[key] = []));
-    return acc;
-  }, {}, list);
-}));
+module.exports = _dispatchable('groupBy', _xgroupBy, reduceBy(function(acc, item) {
+  acc ? acc.push(item) : acc = [item]
```
asaf-romano: So apart that curry thing (sorry for misleading you earlier!), I prefer a proper ``if () { ...; return ...; } return ...`` here.

davidchambers: I'd like to be very clear:

```javascript
if (acc == null) {
  acc = [];
}
acc.push(item);
return acc;
```

I haven't been following this pull request closely, so this may already have been discussed, but why must we use `null` as the initial value rather than `[]`?

asaf-romano: Because the array instance is shared otherwise. That's what you get for mutability. Another option is to special case the empty array case, but I think passing null is cleaner, as far as "clean" can get in this case.

davidchambers: > Because the array instance is shared otherwise.

Ah, yes. Thanks for the explanation.

ThomWright:
:grinning: Done. It does read better this way.

asaf-romano:
🌷

buzzdecafe:
LGTM again :cow2:

davidchambers:
:deciduous_tree:

PR #1707, opened 2016-04-03
Change reduceBy to take the value function and the accumulator value first

### LINKED ISSUE #1706: Reconsider reduceBy arguments order
Naive implementations of the 3 in-house use-cases for reduceBy:
```js
const groupBy = reduceBy(__, append, [], __);
const countBy = reduceBy(__, add, 0, __);
const indexBy = reduceBy(__, (a, b) => b, null, __);
```

However if the signature were reduceBy(valueFn, initialValue, keyFn, list):
```js
const groupBy = reduceBy(append, []);
const countBy = reduceBy(add, 0);
const indexBy = reduceBy((a, b) => b, null);
```
I understand why the current signature makes sense, but the above seems reasonable as well, given that it resembles plain ``reduce`` quite nicely.

asaf-romano:
Closes #1706

### DIFF
--- a/src/reduceBy.js
+++ b/src/reduceBy.js
@@ -14,27 +14,25 @@ var _reduce = require('./internal/_reduce');
  * @memberOf R
  * @since v0.20.0
  * @category List
- * @sig (b -> String) -> ((a, b) -> a) -> a -> [b] -> {String: a}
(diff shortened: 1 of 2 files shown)


davidchambers:
:deciduous_tree:

CrossEye:
:herb:

PR #1711, opened 2016-04-04
eslint config: allow unused args

asaf-romano:
so that ``function(a, b) { return a + 1; }`` is allowed.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -13,7 +13,13 @@
     "new-cap": 0,
     "no-plusplus": 2,
     "no-undef": 2,
-    "no-unused-vars": 2,
+    "no-unused-vars": [
+      2,
+      {
+        "vars": "all",
+        "args": "none"
+      }
+    ],
     "brace-style": [
       2,
       "1tbs",


davidchambers:
:deciduous_tree:

CrossEye:
> so that ``function(a, b) { return a + 1; }`` is allowed.

Where do we want this?  Tests?

asaf-romano:
It's coming.

CrossEye:
[How do you keep a turkey in suspense?][ts]

  [ts]: http://www.reddit.com/r/funny/related/e4d2d/how_do_you_keep_a_turkey_in_suspense/

asaf-romano:
:)

It happened already: https://github.com/ramda/ramda/commit/1921d69c8107e1632e5cb3dfd0a22775f563c580 (in both functions actually)

CrossEye:
Ok, but note that the first one could be replaced by 

```js
reduceBy(inc, 0);
```

and the second one should not trigger this error, since it uses a parameter later in the order than the unused one.

PR #1713, opened 2016-04-04
Define indexBy and countBy in terms of reduceBy

asaf-romano:


davidchambers:
:deciduous_tree: Very nice!

buzzdecafe:
indeed, should also transduce, so that is nice too

asaf-romano:
@buzzdecafe hrm? there's no _xreduceBy (there probably could be, and at that point we would remove _xgroupBy).

buzzdecafe:
sorry my mistake; since it is defined in terms of `reduce` i jumped to the conclusion. but of course we are specifying the output container, so it won't transduce as is.

PR #1723, opened 2016-04-09
Make prepublish script work on GNU

### LINKED ISSUE #1717: prepublish script is platform specific
Doesn't work on GNU because of `-depth` flag to `find`, which should be -`maxdepth` to work on GNU and BSD (OSX). Also, `sed` expression needs to be repaired.

Putting this placeholder issue here to remind me to fix it

buzzdecafe:
Fixes #1717


davidchambers:
Very nice. I just ran this on OS X and can confirm that it works.

🌳

CrossEye:
Yes, it also runs on my Ubuntu 14.04, but that's not a surprise.

PR #1727, opened 2016-04-11
Fix CDN links

cust0dian:
Looks like CDN links were broken since [0.20.1](https://github.com/ramda/ramda/commit/dbc17cc71467daaff0a91ab64599761c80804e65#diff-04c6e90faac2675aa89e2176d2eec7d8). Maybe a bug in `prepublish` script?

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,14 +93,14 @@ or the minified version:
 or from a CDN, either cdnjs:
 
 ```html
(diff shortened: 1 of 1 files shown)


davidchambers:
Thanks for the fix, @cust0dian!

It looks as though the script was run multiple times, but it's not idempotent.

Issue #1732, opened 2016-04-14
Document `R.chain` supports FantasyLand Chain spec


yotamDvir:
I see that functions like [`R.lift`](http://ramdajs.com/docs/#lift) and [`R.map`](http://ramdajs.com/docs/#map) can be used with (appropriate) functors satisfying the specifications in [Fantasy-Land](https://github.com/fantasyland/fantasy-land), yet not [`R.chain`](http://ramdajs.com/docs/#chain). Why is this?

buzzdecafe:
`R.chain` works with FL chains. 

http://goo.gl/18QP4V

yotamDvir:
Oh great. Why isn't this written in the documentation though?

buzzdecafe:
absolutely should be

Issue #1578, opened 2016-01-04
Ramda seems to be spinning off into outer space, getting more abstruse with each release.  Is it intended for normal mortals?


Andrew-Webb:
Today I reached for good old R.filter(...) to filter out some objects from an array.

I fire up the online documentation just to remind me, and it says:-

```
Filterable f => (a → Boolean) → f a → f a
Takes a predicate and a "filterable", and returns a new filterable of the same type
containing the members of the given filterable which satisfy the given predicate.
```

At this point, I'm like: WHAT?!  I don't recognize the signature type, and I don't understand the description.  I then notice that the docs now describe 0.19.0 so I look at 0.18.0 docs and it says:-

```
(a -> Boolean) -> [a] -> [a]
Returns a new list containing only those items that match a given predicate function.
```

Now this is quite understandable and normal.  I look at the 0.19.0 upgrade guide, and there is mention of R.filter, but nothing to suggest that you've broken compatibility here, only that R.filter "may now operate on dictionary-like objects".

My guess is that your new signature and description now cover both lists and dictionary-like objects, in a very clever and terse way.  This is nice from a functionality pov of course, but IMO it's a disaster in terms of keeping Ramda simple, understandable, approachable, friendly, usable.

My experience with R.filter today just brought things sharply into focus... my feeling over the last year has been that, with each new release, Ramda becomes more abstruse, harder to get into, harder to understand.  And that is a shame.

I think that a simpler, less powerful, less clever library could win more hearts and minds, and get more JS programmers using functional techniques.

Ramda has gone from being a library that I've been pushing onto work colleagues ("use this!!  forget about Underscore") to something that I don't think I can recommend any more because it induces too much head scratching.

If you're happy with Ramda's current approach then that's fine, and I don't mean to be critical of it.  But I thought that I should say what I felt.

davidchambers:
Thank you for sharing your concerns, @Andrew-Webb.

I agree that Ramda is less approachable than it once was. This is unfortunate, of course, though we must consider what has been gained as well as what has been lost. One of the wonderful facets of functional programming, in my view, is the clarity it brings to reasoning about data structures. We become aware of underlying truths which unify different data types in useful ways. We realize that "mapping" is not an operation unique to lists, but an operation which dozens of data types can support. With this knowledge in mind, we can write functions which operate on lists, trees, and streams without any special casing.

[...]

PR #1733, opened 2016-04-14
update chain docs to mention FL

benperez:
closes https://github.com/ramda/ramda/issues/1732

### DIFF
--- a/src/chain.js
+++ b/src/chain.js
@@ -9,13 +9,14 @@ var map = require('./map');
  * `chain` maps a function over a list and concatenates the results. `chain`
  * is also known as `flatMap` in some libraries
  *
- * Dispatches to the `chain` method of the second argument, if present.
+ * Dispatches to the `chain` method of the second argument, if present,
+ * according to the [FantasyLand Chain spec](https://github.com/fantasyland/fantasy-land#chain).
  *
  * @func
  * @memberOf R
  * @since v0.3.0
  * @category List
- * @sig (a -> [b]) -> [a] -> [b]
+ * @sig Chain m => (a -> m b) -> m a -> m b
  * @param {Function} fn
  * @param {Array} list
  * @return {Array}


buzzdecafe:
this is great, thanks. 

An aside: I don't like mentioning in the docs that a function dispatches; we should simply point out that it supports FL chains. That FL chains are OO-style methods and so we dispatch is an implementation detail that I don't think we should broadcast.

But that's a bigger issue than this PR; this looks fine to merge to me.

:cow2:

benperez:
> I don't like mentioning in the docs that a function dispatches

Fair enough. It might make sense to remove mention of dispatching in a different PR.

buzzdecafe:
> It might make sense to remove mention of dispatching in a different PR.

Yes, i don't see any reason to hold this up.

davidchambers:
Shall we update the type signature while we're at it?

```haskell
chain :: Chain m => (a -> m b) -> m a -> m b
```

benperez:
:zap:

CrossEye:
I guess this is what I get for going offline for a few days, but I don't agree with that being the only type signature.  I'd like to take some of the lessons from #1578 to heart and make sure that the simplest cases don't get short shrift in our documentation.

I would have no problem at all with

```haskell
chain ::  (a -> [b]) -> [a] -> [b]
      :: Chain m => (a -> m b) -> m a -> m b
```

But let's not lose touch with our simple cases.

benperez:
@CrossEye I also like the idea of keeping 2 signatures: one simple one for the default implementation and one generic for the people interested in FL compatibility. How do you suggest we do this using the existing system? One signature per line?

CrossEye:
Yes, we can already handle this.  See [`take`][ta] for an example.

```js
 * @sig Number -> [a] -> [a]
 * @sig Number -> String -> String
```

screenshot-2016-04-16a

  [ta]: http://ramdajs.com/docs/#take

PR #1738, opened 2016-04-17
Issue#1730 Documentation somewhat unclear for dropLastWhile

bcowgill:
Clarified documentation for dropLastWhile and dropWhile as discussed on issue#1730

### DIFF
--- a/src/dropLastWhile.js
+++ b/src/dropLastWhile.js
@@ -6,10 +6,10 @@ var _xdropLastWhile = require('./internal/_xdropLastWhile');
 
 /**
  * Returns a new list excluding all the tailing elements of a given list which
- * match the supplied predicate function. It passes each value from the right
+ * satisfy the supplied predicate function. It passes each value from the right
  * to the supplied predicate function, skipping elements while the predicate
- * function returns `true`. The predicate function is passed one argument: 
- * (value)*.
+ * function returns `true`. The predicate function is applied to one argument:
+ * *(value)*.
  *
  * @func
  * @memberOf R
--- a/src/dropLastWhile.js
+++ b/src/dropLastWhile.js
@@ -5,10 +5,11 @@ var _xdropLastWhile = require('./internal/_xdropLastWhile');
 
 
 /**
- * Returns a new list containing all but last the`n` elements of a given list,
- * passing each value from the right to the supplied predicate function,
- * skipping elements while the predicate function returns `true`. The predicate
- * function is passed one argument: (value)*.
+ * Returns a new list excluding all the tailing elements of a given list which
+ * match the supplied predicate function. It passes each value from the right
+ * to the supplied predicate function, skipping elements while the predicate
+ * function returns `true`. The predicate function is passed one argument: 
+ * (value)*.
  *
  * @func
  * @memberOf R
--- a/src/dropWhile.js
+++ b/src/dropWhile.js
@@ -5,10 +5,10 @@ var _xdropWhile = require('./internal/_xdropWhile');
 
 
 /**
- * Returns a new list containing the last `n` elements of a given list, passing
- * each value to the supplied predicate function, skipping elements while the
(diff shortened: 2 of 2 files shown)


src/dropLastWhile.js
```diff
@@ -5,10 +5,11 @@ var _xdropLastWhile = require('./internal/_xdropLastWhile');
...
 /**
- * Returns a new list containing all but last the`n` elements of a given list,
- * passing each value from the right to the supplied predicate function,
- * skipping elements while the predicate function returns `true`. The predicate
- * function is passed one argument: (value)*.
+ * Returns a new list excluding all the tailing elements of a given list which
+ * match the supplied predicate function. It passes each value from the right
+ * to the supplied predicate function, skipping elements while the predicate
+ * function returns `true`. The predicate function is passed one argument: 
+ * (value)*.
```
davidchambers: I think we want an asterisk (`*`) immediately before the opening paren (`(`).

src/dropWhile.js
```diff
@@ -5,10 +5,10 @@ var _xdropWhile = require('./internal/_xdropWhile');
 
 
 /**
- * Returns a new list containing the last `n` elements of a given list, passing
- * each value to the supplied predicate function, skipping elements while the
- * predicate function returns `true`. The predicate function is passed one
- * argument: *(value)*.
+ * Returns a new list excluding the leading elements of a given list which
+ * match the supplied predicate function. It passes each value to the supplied
```
davidchambers: I'd like to see "satisfy" in place of "match" (here and in the description of `dropLastWhile`).

src/dropWhile.js
```diff
@@ -5,10 +5,10 @@ var _xdropWhile = require('./internal/_xdropWhile');
...
+ * predicate function, skipping elements while the predicate function returns
+ * `true`. The predicate function is passed one argument: *(value)*.
```
davidchambers: I'd like to see "is applied to" in place of "is passed" (here and in the description of `dropLastWhile`).

bcowgill:
just realised the discussion moved here. will fix in a bit.

bcowgill:
It's all good now.

buzzdecafe:
can you squash?

megawac:
@buzzdecafe https://github.com/blog/2141-squash-your-commits also \o

benperez:
There is an issue for github squashing in sanctuary that's probably also relevant for ramda.

https://github.com/plaid/sanctuary/issues/186

The company I work at just turned it on for our main repo and it's working well for us.

CrossEye:
> There is an issue for github squashing in sanctuary that's probably also relevant for ramda.

It's on as an option already in Ramda (I think that's the default.)  I've used it once.  It works fine, and it's nice to avoid all the "please squash" messages.  Is there a good case for making it always on?

buzzdecafe:
thanks @megawac nice of you to drop in :smile:

benperez:
> Is there a good case for making it always on?

A contributor can have as many commits as they want in their feature branch but only 1 gets added on top of master. Github keeps the individual commits around for historical purposes so you can go back and see them in the PR.

It looks like @buzzdecafe just used it if you want an example of what the commit looks like. Note that it literally only adds a single commit to master as opposed to the `feature commit` + `merge branch commit` of traditional merging.

The benefits of disabling the traditional merge commit is that you don't have to manually select the squash one (the merge one is the default, hence why there are still a lot of recent merge commits in master).

CrossEye:
> The benefits of disabling the traditional merge commit is that you don't have to manually select the squash one (the merge one is the default, hence why there are still a lot of recent merge commits in master).

That's what I was asking about.  I'd already used the squashing one as well.  Actually, what I'd really like is for Github to let us reverse the default, but still keep both options available.  Squashing is definitely more useful.  I'll be very glad to stop asking users to squash their commits in PRs.  I just am not sure if we want to keep both options available or not.

PR #1741, opened 2016-04-19
fix handling of function applicatives in R.ap

davidchambers:
When using [`S.lift2`][1], which is defined in terms of `R.map` and `R.ap`, I expected this to work:

```javascript
lift2(map,
      assoc('name'),
      get(Object, _, newDeps))
```

I was forced to write this instead:

```javascript
lift2(map,
      x => assoc('name', x),
      get(Object, _, newDeps))
```

I'm fairly sure the current definition of `R.ap` is unsound.

Haskell:

```haskell
Prelude Control.Applicative> (<*>) (+) (* 2) 10
30
```

Ramda (before):

```javascript
> R.ap(R.add)(R.multiply(2))(10)
«Function»
```

Ramda (after):

```javascript
> R.ap(R.add)(R.multiply(2))(10)
30
```

As a bonus, the new definition is much clearer.

Please review this change, @benperez.

[1]: http://sanctuary.js.org/#lift2

### DIFF
--- a/src/ap.js
+++ b/src/ap.js
@@ -1,7 +1,6 @@
 var _concat = require('./internal/_concat');
 var _curry2 = require('./internal/_curry2');
 var _reduce = require('./internal/_reduce');
-var curryN = require('./curryN');
 var map = require('./map');
 
 
@@ -9,7 +8,7 @@ var map = require('./map');
  * ap applies a list of functions to a list of values.
  *
  * Dispatches to the `ap` method of the second argument, if present. Also
- * treats functions as applicatives.
+ * treats curried functions as applicatives.
  *
  * @func
  * @memberOf R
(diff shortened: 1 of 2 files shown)


blperez:
```diff
-      curryN(Math.max(applicative.length, fn.length), function() {
-        return applicative.apply(this, arguments)(fn.apply(this, arguments));
-      }) :
+      function(x) { return applicative(x)(fn(x)); } :
```
Beautiful! Functional programming the way it was meant to be written!

This implementation may or may not prevent you from doing some javascript black magic related to variadic functions and preserving context but mentally working through how that is possible is giving me a headache. I'll let somebody who feels more strongly about those issues sort that one out 😉 .

🐪 from me

CrossEye:
I'm pretty sure that section was entirely written on spec.  I really doubt anyone is counting on the possible variadic or `this`-preserving features of the existing code.

So a :herb: from me.

PR #1743, opened 2016-04-20
Fixes equality of neg-zero in _Set

### LINKED ISSUE #1742: R.uniq and -0
```js
R.uniq([-0, -0])
// => [-0, -0]
```

scott-christopher:
Fixes #1742

### DIFF
--- a/src/internal/_Set.js
+++ b/src/internal/_Set.js
@@ -9,10 +9,18 @@ module.exports = (function() {
     this._items = {};
   }
 
+  /**
+   * @param item The item to add to the Set
+   * @returns {boolean} true if the item did not exist prior, otherwise false
+   */
   _Set.prototype.add = function(item) {
-    return hasOrAdd(item, true, this);
+    return !hasOrAdd(item, true, this);
   };
 
+  /**
+   * @param item The item to check for existence in the Set
+   * @returns {boolean} true if the item exists in the Set, otherwise false
+   */
   _Set.prototype.has = function(item) {
     return hasOrAdd(item, false, this);
   };
@@ -25,10 +33,7 @@ module.exports = (function() {
    * @param shouldAdd  If true, the item will be added to the set if it doesn't
    *                   already exist.
    * @param set        The set instance to check or add to.
-   * @return {boolean} When shouldAdd is true, this will return true when a new
-   *                   item was added otherwise false. When shouldAdd is false,
-   *                   this will return true if the item already exists, otherwise
-   *                   false.
+   * @return {boolean} true if the item already existed, otherwise false.
    */
   function hasOrAdd(item, shouldAdd, set) {
     var type = typeof item;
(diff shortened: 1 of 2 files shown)


svozza:
Ha. Beat me to [it](https://github.com/svozza/ramda/blob/uniq-minus-zero/src/internal/_Set.js).

src/internal/_Set.js
```diff
@@ -37,11 +37,15 @@ module.exports = (function() {
...
-            set._items['-0'] = true;
+        if (item === 0 && 1 / item === -Infinity) {
+          if (set._items['-0'] === true) {
+            return !shouldAdd;
+          } else {
+            if (shouldAdd) {
+              set._items['-0'] = true;
+            }
+            return shouldAdd;
           }
```
davidchambers: I find the logic above difficult to follow. The `shouldAdd` parameter indicates that this function has two modes. The logic above is confusing because it references this parameter several times. I'd like to see a single branch which reflects the fact that `shouldAdd` is a switch. I suggest something along these lines:

```javascript
var setContainsNegativeZero = set._items['-0'] === true;
if (shouldAdd) {
  return setContainsNegativeZero ? false : (set._items['-0'] = true);
} else {
  return setContainsNegativeZero;
}
```

I would also like to see `{'-0': false}` rather than `{}` as the initial value of `set._items`. This would remove the need for `=== true`, and would indicate—in the constructor—the purpose of the `_items` property.

scott-christopher: > I find the logic above difficult to follow.

I agree. I think the confusion is particularly caused by the overloaded boolean return value. I've updated the PR such that the function will now just return `true` or `false` based on whether the value previously existed.

> I would also like to see {'-0': false} rather than {} as the initial value of set._items. This would remove the need for === true, and would indicate—in the constructor—the purpose of the _items property.

The original PR had this declared explicitly, though we decided to move them all to the same mechanism for consistency. I can add them all back if desired.

davidchambers:
:deciduous_tree: The changes look great, @scott-christopher!

scott-christopher:
And I think negative zero has become my arch-nemesis.

jdalton:
@scott-christopher 
```js
R.prop(-0, {'-0': 'a'}); 
// => undefined
```

scott-christopher:
Yeah that's not nice either. It will give you the expected response if you pass it a string for the first arg though.
```js
R.prop('-0', {'-0': 'a'}); // "a"
```
It sucks that negative zero exists at all in JS, but I guess it just comes along for the ride by having IEEE-754 floats.

davidchambers:
I'd like `R.prop(-0)` to be a type error as it is in Sanctuary:

```javascript
S.prop(-0);
// ! TypeError: Invalid value
//
//   prop :: Accessible a => String -> a -> b
//                           ^^^^^^
//                             1
//
//   1)  -0 :: Number, FiniteNumber, Integer, ValidNumber
//
//   The value at position 1 is not a member of ‘String’.
```

Failing that, the behaviour of `R.prop` when applied to invalid arguments should remain unspecified.

jdalton:
> Yeah that's not nice either. It will give you the expected response if you pass it a string for the first arg though.

```js
R.sum([-0])
// => 0
```

CrossEye:
The fixes are great.  

I have no issue with this:

```js
R.sum([-0])
// => 0
```

Once you start putting `-0` in an additive context, even if we can imagine an implementation which avoids any addition in certain circumstances, I would actually rather this result.  The alternative, presumably `//=> -0`, might lead to some odd results:

```js
    R.sum(numbers.concat(0)) ≍  R.sum(numbers); // ∀ numbers ∈ [Number]
                                                // Oh, wait, except for numbers = [-0]
```

davidchambers:
Good point, @CrossEye. If I were to :trollface: I'd counter with something like this:

```haskell
forall n :: Number. R.sum([n]) = n
```

Seriously, though, `0` and `-0` are equivalent in the context of addition. The behaviour of `R.sum([-0])` does not concern me.

jdalton:
Sweet.

PR #1748, opened 2016-04-23
improve perf of propEq and propIs

benji6:
I think defining `propEq` and `propIs` in terms of `propSatisfies` is elegant, but having had problems using these functions on large lists I feel performance is probably more important.

I added a benchmark for `propEq` and had these results running locally:

```
# original definition
propEq("value", [1, 2… │ 368,588Hz │ 2.06% error margin

# new definition
propEq("value", [1, 2… │ 438,435Hz │ 1.79% error margin
```
Could add a benchmark for `propIs` too but I expect the results would be quite similar.

Let me know what you think

### DIFF
--- a/lib/bench/propEq.bench.js
+++ b/lib/bench/propEq.bench.js
@@ -1,0 +1,10 @@
+var propEq = require('../..').propEq;
+
+module.exports = {
+  name: 'propEq',
+  tests: {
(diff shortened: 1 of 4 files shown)


davidchambers:
:deciduous_tree: Works for me!

buzzdecafe:
me too :cow2:

buzzdecafe:
thanks

CrossEye:
:herb:

Thanks!

PR #1749, opened 2016-04-23
maybe

buzzdecafe:
this is a strawman proposal to generate some discussion about bringing a Maybe type into the core lib, but without forcing it down users' throats

### DIFF
--- a/src/toMaybe.js
+++ b/src/toMaybe.js
@@ -1,0 +1,68 @@
+var _curry2 = require('./internal/_curry2');
+var curryN = require('./curryN');
+
+
+/**
+ * Takes a predicate and a function to wrap. `toMaybe` evaluates the function.
+ * and passes its result to the predicate. If the predicate evaluates to
+ * `true`, then the value is returned inside a `Just`; otherwise `Nothing`
+ * is returned.
+ *
+ * `Just` and `Nothing` satisfy the Fantasy-Land specs for  Functor,
+ * Applicative, Apply, Chain, and Monad.
+ *
+ *
+ * @func
+ * @memberOf R
+ * @category Type
+ * @sig (a -> Bool) -> (...x -> a) -> (...x -> Maybe a)
+ * @param {Function} pred The predicate
+ * @param {Function} fn The function whose value you want to wrap in a Maybe
+ * @return {Object} Maybe a
+ * @example
+ *
+ *   const safeFind = R.toMaybe(x => x !== void 0, R.find);
+ *   const found = safeFind(R.equals(2), [1, 2, 3,]); //=> R.toMaybe.Just(2)
+ *   const notFound = safeFind(R.equals(4), [1, 2, 3,]); //=> R.toMaybe.Nothing()
+ *
+ */
+module.exports = (function() {
+  function returnThis() { return this; }
+
+  function Maybe() {}
+  Maybe.prototype.of = function _maybeof(x) { return new Just(x); };
+  Maybe.prototype['@@type'] = 'ramda/Maybe';
+
+  function Just(x) { this.value = x; }
+  Just.prototype = new Maybe();
+  Just.prototype.map = function _justmap(f) {
+    return new Just(f(this.value));
+  };
+  Just.prototype.ap = function _justap(m) {
+    return m.map(this.value);
+  };
+  Just.prototype.chain = function _justchain(f) {
+    return f(this.value);
+  };
+  Just.prototype.isNothing = false;
+  Just.prototype.isJust = true;
+
+  function Nothing() {}
+  Nothing.prototype = new Maybe();
+  Nothing.prototype.map = returnThis;
+  Nothing.prototype.ap = returnThis;
+  Nothing.prototype.chain = returnThis;
+  Nothing.prototype.isNothing = true;
+  Nothing.prototype.isJust = false;
+
+  var toMaybe = _curry2(function _toMaybe(pred, fn) {
+    return curryN(fn.length, function() {
+      var result = fn.apply(this, arguments);
+      return pred(result) ? new Just(result) : new Nothing();
+    });
+  });
+  toMaybe.Just = function _just(x) { return new Just(x); };
+  toMaybe.Nothing = function _nothing() { return new Nothing(); };
+
+  return toMaybe;
+}());
--- a/test/toMaybe.js
+++ b/test/toMaybe.js
@@ -1,0 +1,37 @@
+var R = require('..');
+var eq = require('./shared/eq');
+
+
+describe('toMaybe', function() {
+  var isDefined = function(x) { return x !== void 0; }
+  var safeHead = R.toMaybe(isDefined, R.head);
+  var safeFind = R.toMaybe(isDefined, R.find);
+
+  it('returns a function', function() {
+    eq(typeof safeHead, 'function');
+    eq(typeof safeFind, 'function');
+  });
+
+  it('returns a function with the same arity as the function it is wrapping', function() {
+    eq(safeHead.length, 1);
+    eq(safeFind.length, 2);
+  });
+
+  it('returns a function that returns a Just when the predicate is satisfied', function() {
+    eq(safeHead([1, 2, 3]).isJust, true);
+    eq(safeHead([1, 2, 3]), R.toMaybe.Just(1));
+    eq(safeFind(R.equals(2), [1, 2, 3]).isJust, true);
+    eq(safeFind(R.equals(2), [1, 2, 3]), R.toMaybe.Just(2));
+  });
+
+  it('returns a function that returns a Nothing when the predicate is not satisfied', function() {
+    eq(safeHead([]).isNothing, true);
+    eq(safeFind(R.equals(2), []).isNothing, true);
+  });
+
+  it('is curried', function() {
+    eq(typeof R.toMaybe(isDefined), 'function');
+    eq(typeof R.toMaybe(isDefined)(R.head), 'function');
+    eq(R.toMaybe(isDefined)(R.head)([]).isNothing, true);
+  });
+});


CrossEye:
I like this as a start.  It would really be nice if it were an `addIndex` sort of helper.

test/maybe.js
```diff
@@ -0,0 +1,18 @@
...
+    eq(R.maybe(notEmpty, R.head, []).isJust, false);
+    eq(R.maybe(notEmpty, R.head, [1]).isNothing, false);
+    eq(R.maybe(notEmpty, R.head, [1]).isJust, true);
+  });
+
+  it('is curried', function() {
+    eq(typeof R.maybe(notEmpty), 'function');
+    eq(typeof R.maybe(notEmpty, R.head), 'function');
+    eq(R.maybe(notEmpty, R.head)([]).isNothing, true);
+});
```
davidchambers: You're missing a `});`.

buzzdecafe: :zap:

that's it

src/maybe.js
```diff
@@ -0,0 +1,36 @@
...
+
+  function Nothing() {}
+  Nothing.prototype = {
+    of: Just.prototype.of,
+    map: returnThis,
+    ap: returnThis,
+    chain: returnThis,
+    isNothing: true,
+    isJust: false
+  };
```
davidchambers: Could you add `'@@type': 'ramda/Maybe'` for compatibility with [`S.type`][1]?


[1]: http://sanctuary.js.org/#type

buzzdecafe: certainly -- but since this is a strawman, i want to wait and see what direction this whole thing is going to take before buttoning down the details. That's also why the test "suite" is so sparse.

also, i am a bit squeamish about magic string typing; but i can see the case for S compatibility.

jimf:
> but without forcing it down users' throats

I'm all for having a Maybe type in the core lib. However, I'm curious what thoughts are on instead (or even in addition to) going with a "little maybe," more akin to a decorator that abstracts null-checking. Here's a version that comes straight from [JavaScript Allonge](https://leanpub.com/javascriptallongesix/read):

```js
const maybe = (fn) =>
  function (...args) {
    if (args.length === 0) {
      return
    }
    else {
      for (let arg of args) {
        if (arg == null) return;
      }
      return fn.apply(this, args)
    }
  }

maybe((a, b, c) => a + b + c)(1, 2, 3)
  //=> 6

maybe((a, b, c) => a + b + c)(1, null, 3)
  //=> undefined
```

I'd argue that this is a more natural *start* for most JS devs coming in, and it certainly wouldn't preclude a "big Maybe" from being added as well. However, two cons that come to mind:

1. When docs make reference to Maybe types, we'd want to avoid any confusion with the decorator version
1. There's another version of "little maybe," notably from the [Mostly Adequate Guide](https://drboolean.gitbooks.io/mostly-adequate-guide/content/ch8.html) that is a helper for getting the value out of a Maybe:
```js
//  maybe :: b -> (a -> b) -> Maybe a -> b
var maybe = curry(function(x, f, m) {
  return m.isNothing() ? x : f(m.__value);
});
```

This looks great though. Looking forward to the discussion!

gilligan:
I would much prefer a "proper" fantasyland compatible implementation. I think that as long as the data type is introduced on its own it really does not pose a problem to anyone who might not fully grasp all of its uses.

To be honest personally I would rather not have Maybe at all than having a "simplified" version.

buzzdecafe:
i'm inclined to agree with @gilligan that if we're in for a penny we should be in for a pound. 

It sounds like there's some momentum for this. So then I have this question: Should we expose, e.g. `R.Just`, on the public API? @scott-christopher has suggested some interesting stuff making use of types internally (e.g. `Cons`) that we could also expose (or not).

CrossEye:
> However, I'm curious what thoughts are on instead (or even in addition to) going with a "little maybe," more akin to a decorator that abstracts null-checking. Here's a version that comes straight from [JavaScript Allonge][a6]

If there was any one thing that served as a primary inspiration for Ramda it was the first edition of [Allongé][a5].  It is still one of my most-recommended JS books.  But I don't think we would want to incorporate this into Ramda.  It is a very useful function for certain kinds of application code, but it is a little undisciplined for Ramda.  Moreover, the sort of `Maybe` suggested by @buzzdecafe would not work together with it; I think it could just lead to confusion.

Where I absolutely could see using lower-case "`maybe`" would be as a decorator, to wrap an existing function (along with a predicate that has the same input signature) and create a new function that returns a `Maybe`-fied version of the output.

```
// head:: [a] -> a | Undefined
// safeHead :: [a] -> Maybe a
const safeHead = maybe(as -> as.length > 0, head);
```

  [a5]: https://leanpub.com/javascript-allonge
  [a6]: https://leanpub.com/javascriptallongesix/read:

src/maybe.js
CrossEye: I also think there would need to be a `Maybe` type in here that the others extend.  I really, really would not want to promote signatures with types like `Just a | Nothing`.  :smile:

gilligan:
@buzzdecafe how about adding `fromNullable` like in `data.maybe`?

buzzdecafe:
> how about adding fromNullable like in data.maybe?

@gilligan that would be simple enough to build out of what i am proposing, e.g.

```js
const fromNullable = R.maybe(R.complement(R.isNil), R.identity);
```
(cf. https://github.com/folktale/data.maybe/blob/master/lib/maybe.js#L124-L126)

src/maybe.js
```diff
@@ -0,0 +1,36 @@
...
+
+  return _curry3(function maybe(pred, fn, x) {
+    return pred(x) ? new Just(fn(x)) : new Nothing();
+  });
```
CrossEye: One thing that I think is wrong with this POC is that this will only work for unary functions.  I really think we want to support polyadic functions too.  Even internally, we'd probably like to do this with, say, `find`

I don't think that wrapping the third parameter in a list is the right solution.  Instead, we should probably make this a two-step process:

```js
maybe(pred, fn)(some, args, here); //=> Maybe val
```

I know that this would involve a less simple signature than we might like, perhaps something like this:

```haskell
maybe :: (a -> b -> ... -> Boolean) -> (a -> b -> ... -> z) -> (a -> b -> ... -> Maybe z)
```

But this seems worth it for the added power and flexibility.

buzzdecafe:
> I really think we want to support polyadic functions too.

I was hoping that I could slide that by. Oh well.

CrossEye:
>> I really think we want to support polyadic functions too.

> I was hoping that I could slide that by. Oh well.

What?  You don't like `find`?

buzzdecafe:
> What? You don't like `find`?

Well, `find` is `(a -> Any) -> [Any] -> Any | undefined`. So one could simply `safeFindThing = maybe(complement(isNil), find(thingPred))`, to get `safeFindThing :: [Any] -> Maybe Any`, right?

buzzdecafe:
OK, I've updated this PR to include the following:

* change the API to return a safe function, rather than including a single argument, as requested by @CrossEye
* Made `Just | Nothing` inherit from a (private) `Maybe` constructor (also at @CrossEye 's request
* exposed `Just` and `Nothing` as properties on `R.maybe`, i.e. `R.maybe.Just` and `R.maybe.Nothing` 
* added some more tests

Haven't yet, but intend to, pending approval:
* add magic `@@type` property for Sanctuary-compat
* include tests for Functor, Apply, Applicative, Chain to the `Maybe` impl. 
* change the name to `toMaybe`; not as nice, but if we want to add an `Either` type, we already have an `either` function. So .....
* Move `Maybe` implementation to `internal` directory so it may be reused in other functions (or should we start a `types` directory?)

I think this is close to being ready to merge. I am very interested to hear feedback. ((:bell:))

src/maybe.js
```diff
@@ -0,0 +1,44 @@
...
+  };
+  Just.prototype.chain = function _justchain(f) {
+    return f(this.value);
+  };
+  Just.prototype.isNothing = false;
+  Just.prototype.isJust = true;
+
+  function Nothing() {}
+  Nothing.prototype = new Maybe();
+  Nothing.prototype.of = Just.prototype.of;
```
davidchambers: We might as well define this on `Maybe.prototype`.

buzzdecafe: good point

CrossEye:
I like this a lot.  My internet has been down again (the cable company is coming this week to give me an estimate; maybe I'll finally get rid of DSL!) so I haven't been able to play with it, but the code and tests look good.

As to the proposed changes, I think two should probably be done before merging; the other two can wait.  We should probably rename to the less pleasing `toMaybe`.  And we should add the `@@type` annotation.  I'm not a fan of magic strings either, but Sanctuary integration is important... unless David has a different magic trick up his sleeve for that.  The additional tests, and `./internal` versus `./types` can happen now or later; I don't think it matters much.

buzzdecafe:
> We should probably rename to the less pleasing toMaybe. And we should add the @@type annotation. 

By "we" I assume you mean me? :smiling_imp: 

I've updated the PR with those features.

davidchambers:
There's a fundamental problem with this approach: it's impossible to define lossless functions in terms of lossy ones.

If `R.head(xs)` evaluates to `undefined`, is `xs` empty? We don't know.

If `R.tail(xs)` evaluates to `[]`, is `xs` empty? We don't know.

I'd like Ramda to provide safe equivalents of its unsafe functions. Failing that, though, I think it's best to leave things as they are. Providing a function which derives safe functions from unsafe ones would be making a promise we can't keep.

buzzdecafe:
on the contrary. in order to cover cases in a generic fashion, it made sense to branch on the result of the wrapped function. If instead, you want to create a safe `head` based on the list length before evaluating `head`, you can do that for yourself, since `Just` and `Nothing` are exposed on the API: 

`const safeHead2 = xs => ifElse(isEmpty(xs), toMaybe.Nothing(), toMaybe.Just(head(xs)))`

This is consistent with treating these things as lists, where `undefined` as a list element doesn't make sense.

> If `R.tail(xs)` evaluates to `[]`, is `xs` empty? We don't know.

We didn't ask. We asked, "What is the `tail` of `xs`?" If we want to find out if `xs` is empty, we can ask `isEmpty(xs)`. Seems like a separate problem.

davidchambers:
> `undefined` as a list element doesn't make sense

This reminds me of the times `Maybe.of(undefined)` has been discussed. I agree that arrays don't often contain `undefined` as one of their values, but `undefined` is a JavaScript value whether we like it or not (and apparently we like it, since several of our functions use it as a return value).

Which of the following is preferable?

```haskell
Maybe.of :: a -> Maybe a

Maybe.of :: NotUndefined a => a -> Maybe a
```

Similarly, which of the following is preferable?

```haskell
head :: Array a -> Maybe a

head :: NotUndefined a => Array a -> Maybe a
```

For a concrete example of an array containing `undefined`, consider the result of evaluating the following:

```javascript
R.ap([R.head], [[], [1, 2, 3], [], [4, 5, 6]])
```

> > If `R.tail(xs)` evaluates to `[]`, is `xs` empty? We don't know.
>
> We didn't ask. We asked, "What is the `tail` of `xs`?" If we want to find out if `xs` is empty, we can ask `isEmpty(xs)`. Seems like a separate problem.

It's a matter of [data coherence][1]. If we wish to apply a transformation to the head of a list if the list is non-empty, we could write `xs.length > 0 ? f(R.head(xs)) : …`, but this is an unsafe expression guarded by a predicate. We can see that evaluating the unsafe expression will not result in an exception, though a compiler may not be able to prove this.

Similarly, if we wish to apply a transformation to the tail of a list if the list is non-empty, we could write `xs.length > 0 ? g(R.tail(xs)) : …`.

Rich Hickey claims data is always simpler than code. I agree. It's better to represent success/failure *and* the result of the computation (if successful) in a single *value*, rather than with conditional logic. We can then write `S.head(xs).map(f)` and `S.tail(xs).map(g)`.


[1]: https://www.youtube.com/watch?v=gVXt1RG_yN0

buzzdecafe:
Had a very productive back-channel discussion with @davidchambers .

The basic problem we face, is that some functions should be checked at input (`head`), others at output (`find`). I see three ways to deal with this in the context of the current PR:

1. `maybeIn` and `maybeOut` (or some such. But where would `Just` and `Nothing` live?)

2. change `toMaybe` to take an in and an out test, e.g.
`toMaybe :: (...args -> Boolean) -> (a -> Boolean) -> (...args -> a) -> (...args -> Maybe a)`
this feels awkward, because you will typically only need one or the other, not both.

3. pass both the args and the result to the predicate, and let the user decide, i.e.
`toMaybe :: ([...args] -> a -> Boolean) -> (...args -> a) -> (...args -> Maybe a)`
But then you may waste a call to evaluate the function if the `...args` are sufficient to tell you the answer.

Then @davidchambers cut the Gordian Knot: "I think the simplest solution is to provide `R.Nothing` and `R.Just` and call it a day."

I agree -- I think we should provide `Just` and `Nothing` on `R`.

gilligan:
:+1: on just providing Just and Nothing from me ;)

buzzdecafe:
dumping this PR. I'll submit another one as described

CrossEye:
Perhaps this is for the best.  But it does seem sad that these would have to be done through patterns rather than through a dedicated function (or two of them.)

Obviously, we already have `wrap`, which can be used for this, but it's simply not as elegant as I would like:

```js
const safeHead = wrap(head, (head, ls) => 
  isEmpty(ls) ? Nothing() : Just(head(ls))
);

safeHead([1, 2, 3]); //=> Just(1)
safeHead([]); //=> Nothing()

const safeFind = wrap(find, (find, pred, ls) => {
  const result = find(pred, ls);
  return isNil(result) ? Nothing() : Just(result);
})

safeFind(whereEq({a: 4}), [{a: 3, b: 1}, {a: 4, b: 1}, {a: 5, b: 9}]);
//=> Just({a: 4, b: 1})
safeFind(whereEq({a: 8}), [{a: 3, b: 1}, {a: 4, b: 1}, {a: 5, b: 9}]);
//=> Nothing()
```

davidchambers:
I still don't see the point of [`R.wrap`][1], @CrossEye.

> ```javascript
> const safeHead = wrap(head, (head, ls) =>
>   isEmpty(ls) ? Nothing() : Just(head(ls))
> );
> ```

The above strikes me as a roundabout way of writing:

```javascript
const safeHead = xs => isEmpty(xs) ? Nothing() : Just(head(xs))
```


[1]: http://ramdajs.com/docs/#wrap

CrossEye:
Well, it may simply be the last vestiges in my brain of AOP, but I do reach for it whenever I want to wrap a number of functions in the same wrapper, especially if it does processing both on the input and the output.

Issue #230, opened 2014-08-09
testing for Primitives in `is`


buzzdecafe:
at present:

```
R.is(Object, 1); // => true
R.is(Object, "s"); // => true
R.is(Object, true); // => true
```

Should it be true? Do we want to distinguish primitives from their object wrappers, e.g.

```
function isPrimitive(x) { return Object(x) !== x; }
```

PR #1756, opened 2016-05-03
avoid instanceof checks

### LINKED ISSUE #1670: R.join unusable in other contexts due to instanceof check
```console
$ node -p 'require("vm").runInNewContext("R.join(\"-\", [\"foo\", \"bar\", \"baz\"])", {R: require("ramda")})'

path/to/ramda/dist/ramda.js:7927
            throw new TypeError(toString(target) + ' does not have a method named "' + method + '"');
            ^
TypeError: ["foo", "bar", "baz"] does not have a method named "join"
```

This fails due to an `instanceof` check in [__invoker.js__][1]:

```javascript
target != null && is(Function, target[method])
```

We should update [`R.is`][2] to work like [`S.is`][3].

[1]: https://github.com/ramda/ramda/blob/v0.19.1/src/invoker.js#L34
[2]: http://ramdajs.com/docs/#is
[3]: https://github.com/plaid/sanctuary#is

davidchambers:
Fixes #1670

I'd like to "fix" `R.is` and `R.propIs` as well, but that can wait for another day.

See sanctuary-js/sanctuary#100 for context.

### DIFF
--- a/src/length.js
+++ b/src/length.js
@@ -1,5 +1,5 @@
 var _curry1 = require('./internal/_curry1');
-var is = require('./is');
+var _isNumber = require('./internal/_isNumber');
 
 
 /**
@@ -18,5 +18,5 @@ var is = require('./is');
  *      R.length([1, 2, 3]); //=> 3
  */
 module.exports = _curry1(function length(list) {
-  return list != null && is(Number, list.length) ? list.length : NaN;
+  return list != null && _isNumber(list.length) ? list.length : NaN;
 });
--- a/src/invoker.js
+++ b/src/invoker.js
@@ -1,7 +1,7 @@
 var _curry2 = require('./internal/_curry2');
+var _isFunction = require('./internal/_isFunction');
 var _slice = require('./internal/_slice');
 var curryN = require('./curryN');
-var is = require('./is');
 var toString = require('./toString');
 
 
@@ -31,7 +31,7 @@ var toString = require('./toString');
 module.exports = _curry2(function invoker(arity, method) {
   return curryN(arity + 1, function() {
     var target = arguments[arity];
-    if (target != null && is(Function, target[method])) {
+    if (target != null && _isFunction(target[method])) {
       return target[method].apply(target, _slice(arguments, 0, arity));
     }
     throw new TypeError(toString(target) + ' does not have a method named "' + method + '"');
--- a/src/isArrayLike.js
+++ b/src/isArrayLike.js
@@ -1,5 +1,6 @@
 var _curry1 = require('./internal/_curry1');
 var _isArray = require('./internal/_isArray');
+var _isString = require('./internal/_isString');
 
 
 /**
@@ -25,7 +26,7 @@ module.exports = _curry1(function isArrayLike(x) {
   if (_isArray(x)) { return true; }
   if (!x) { return false; }
   if (typeof x !== 'object') { return false; }
-  if (x instanceof String) { return false; }
+  if (_isString(x)) { return false; }
   if (x.nodeType === 1) { return !!x.length; }
   if (x.length === 0) { return true; }
   if (x.length > 0) {
--- a/test/invoker.js
+++ b/test/invoker.js
@@ -1,4 +1,5 @@
 var assert = require('assert');
+var vm = require('vm');
 
 var R = require('..');
 var eq = require('./shared/eq');
@@ -38,6 +39,10 @@ describe('invoker', function() {
                err.message === '[1, 2, 3] does not have a method named "length"';
       }
     );
+  });
+
+  it('does not rely on constructor identity', function() {
+    eq(concat2([2], [3], vm.runInNewContext('[1]')), [1, 2, 3]);
   });
 
   it('curries the method call', function() {


buzzdecafe:
:cow2:

asaf-romano:
🌷

src/length.js
```diff
@@ -1,5 +1,5 @@
 var _curry1 = require('./internal/_curry1');
-var is = require('./is');
+var _isNumber = require('./internal/_isNumber');
```
CrossEye: How did we get ourselves to the state where we proudly promote a public API of `R.is(Type, val)` and internally have a dozen `internal/isTypeA)` calls.  It just feels strange.

davidchambers: [`S.is`][1] does not rely on `instanceof`. If we're okay with losing support for inheritance (I am) we can use the same approach. If we don't want to support `@@type` the implementation is straightforward:

```javascript
function is(ctor, x) {
  var match = String(ctor).match(/function (\w*)/);
  return match != null && match[1] === R.type(x);
}
```


[1]: http://sanctuary.js.org/#is

CrossEye:
>  If we're okay with losing support for inheritance (I am) we can use the same approach.

I don't think I am.  I would like the library to at least work somewhat with OOP JS, and that would certainly mean being able to report `is(Rectangle, mySquare) //=> true`.

As to your proposed implementation, I'm not going  to dig into the specification now, but I believe that the behavior for `String(hostFunction)` is not specified, so even if it works now for the current crop of environments, I'm hesitant to depend upon it.  I think perhaps doing something with `isPrototypeOf` might work.  (Or would that have the same cross-realm problems?  Hmm...)

davidchambers:
> I would like the library to at least work somewhat with OOP JS, and that would certainly mean being able to report `is(Rectangle, mySquare) //=> true`.

"Somewhat" is the operative word here. `is(Rectangle, mySquare)` can evaluate `true` in *some* cases. If `Rectangle` and `mySquare` are from different realms the expression will evaluate `false`.

The problem results from our reliance on object identity. To quote a recent [comment][1]:

> FP is about values, not references. We always try to think about equality in a way that has nothing to do with where something is stored in memory.

I agree. Let's not encourage our users to think about where `Rectangle` is stored in memory. Let's make Ramda work consistently across realms. If a user wishes to opt in to fragile `instanceof` checks, writing `x instanceof y` is not onerous.


[1]: https://github.com/ramda/ramda/issues/1760#issuecomment-217332582

buzzdecafe:
>  Let's not encourage our users to think about where Rectangle is stored in memory. 

@davidchambers I don't follow this point at all. The question is not "where is Rectangle in memory?", the question is "Is a square a Rectangle?" Scott thinks we should answer that question __yes__; you think it should be __no__.  Or is it? `S.is(S.Maybe, S.Just())` is true -- but I assume that is using string-typing?

I am inclined to punt on this. On one hand it seems reasonable to expect that `is` would follow `instanceof`; On the other, I am wary of "inheriting" the problems of JS OOP.

:football: *punt*

davidchambers:
I believe you've misunderstood my position, @buzzdecafe. I don't much mind whether the answer to the question is yes or no, but I demand a *consistent* answer. There's no way to be consistent while relying on `instanceof`, and there's no way to support inheritance *without* using `instanceof` (as far as I'm aware).

I consider this unacceptable:

```javascript
> R.is(Array, [])
true

> R.is(Array, vm.runInNewContext('[]'))
false

> R.is(vm.runInNewContext('Array'), [])
false

> R.is(vm.runInNewContext('Array'), vm.runInNewContext('[]'))
false

> let ctx = vm.createContext({})
undefined

> R.is(vm.runInContext('Array', ctx), vm.runInContext('[]', ctx))
true
```

The only way to understand these inconsistent results is to think about where things live in memory.

For the past year or so I've avoided `R.is`, and have used `Object.prototype.toString`, `R.type`, and `S.is` for type checking. I suggest we do one of the following:

  - make `R.is` behave just like `S.is`;
  - make `R.is` behave like `S.is`, but without support for `@@type`;
  - deprecate `R.is` (and `R.propIs`); or
  - warn users of the dangers of `R.is` (and `R.propIs`) in the documentation.

These options are listed from most favoured to least favoured.

buzzdecafe:
thank you for the clarification. I now understand where you are coming from, and I'm inclined to agree. I'm not a fan of string typing, as you know. So at best I'd go for "make `R.is` behave like `S.is`, but without support for `@@type`".

CrossEye:
I will try testing some of this tomorrow, but I don't want to throw the baby out with the bath water.  I agree absolutely with removing `instanceof`, and would like to go with a `Object.prototype.toString` solution.  I feel as though we keep suggesting this but never quite getting to it.

But I'm not as ready as @davidchambers is to give up on inheritance, at least not without a fight.  I think answering consistently wrong is still a pretty bad idea.  For if the inheritance is set up correctly, a `Square` _is_ a `Rectangle`, and answering `false` is fairly nasty.

There seem to be some echos of #230 / #296 here.

I'm not sure how much consistency of user types across realms should dictate the rest of this.  I might well be able to live with documenting problems in this case, so long as we can properly handle the built-in types.  But I will look to see what I can come up with over the weekend.

svozza:
Apropos of nothing:

```js
R.is(Object, Object.create(null))
// => false
```

CrossEye:
@svozza: Right, that's one great reason to move to kangax's method.

safareli:
just some ideas:
What if `@@type` was an array of its ancestors?
```
Object.prototype['@@type'] = ['Object']
Function.prototype['@@type'] = Object.prototype['@@type'].concat('Function')
Maybe.prototype['@@type'] = Object.prototype['@@type'].concat('org.ramda.Maybe')
Just.prototype['@@type'] = Maybe.prototype['@@type'].concat('org.ramda.Maybe/Just')
```

Maybe even include version in `@@type` too:
```
Maybe.prototype['@@type'] = Object.prototype['@@type'].concat('org.ramda.Maybe#2.1.8')
Just.prototype['@@type'] = Maybe.prototype['@@type'].concat('org.ramda.Maybe/Just#2.1.8')
```

safareli:
Thoughts on this? @CrossEye @davidchambers

CrossEye:
I don't particular like the idea of slapping on a property to describe the type.  I feel as though this should be intrinsic to the object.  I also don't feel as though we should be playing around with the native prototypes to support it.

I know that Sanctuary uses this when it's available, and I suppose I wouldn't have a major objection, but I don't really feel it's worth adding if there's any other option.

buzzdecafe:
> I don't particular like the idea of slapping on a property to describe the type.

+ :100:

davidchambers:
I find `@@type` properties very useful. Here's the current implementation of `equals` from sanctuary-js/sanctuary-type-classes#2:

> ```javascript
>   var equals = function(x, y) {
>     var equalsX = Setoid.getBoundMethod('equals', x);
>     var equalsY = Setoid.getBoundMethod('equals', y);
>     return type(x) === type(y) &&
>            equalsX != null && equalsX(y) &&
>            equalsY != null && equalsY(x);
>   };
> ```

`type(x) === type(y)` ensures we never compare apples and oranges. As a result, each type's `fantasy-land/equals` method is only responsible for comparing two values of the type itself. `Null#fantasy-land/equals`, for example, simply returns `true` since the Null type has only one member.

The other role the `@@type` property serves is indicating that a value should *not* be treated as a string map. This allows `empty(Right(42))` to be a type error since the Either type cannot satisfy the requirements of [Monoid][1]. Were it not for the `@@type` property `Right(42)` would be treated as a string map so the expression would evaluate to the empty string map, `{}`.

I don't think `@@type` should be an array of values (or serialized equivalent). The values are not hierarchical, which turns out to be a useful property. One way to think of `@@type` is as a more descriptive relative of `Object#toString` (but it's still forced to give exactly one answer).

I don't think `@@type` should include the name of the data constructor (unless we're punning<sup>†</sup>). This would make the property's purpose less clear, and force choices between two reasonable behaviours: should `S.type(Right(42))`, for example, evaluate to `'sanctuary/Either'` or `'sanctuary/Either/Right'`? Would we need *two* functions?

Instead, authors of algebraic data types should keep track of the data constructor to which each value belongs in whichever way they choose. This needn't be exposed to the user. In Sanctuary, for example, the internal representation of an Either value is of type `{ isLeft :: Boolean, isRight :: Boolean, value :: Any }`.

I'm opposed to including version information in `@@type` properties for the same reasons I oppose including the name of the data constructor.

---

† Punning is common for types with only one data constructor. The Identity type, for example, has a data constructor named `Identity`.


[1]: https://github.com/fantasyland/fantasy-land#monoid

Issue #1755, opened 2016-05-03
Special case R.concat for Arrays


Bradcomp:
Currently: 
```
R.concat([1, 2, 3], 4);  
=> [1, 2, 3, 4]

R.concat(R.concat([1], 2), 3);
=> [1, 2, 3]

R.concat([1], R.concat(2, 3));
=> Throws!
```
Given that we special case arrays for `map`, `filter`, etc., the same should be done here.  

`R.concat([1], 2)` Should throw an error.

buzzdecafe:
thanks @Bradcomp this needs to be fixed

`R.concat([1, 2, 3], 4);  => [1, 2, 3, 4]` :-1:

`R.concat(R.concat([1], 2), 3); => [1, 2, 3]` :-1: 

`R.concat([1], R.concat(2, 3)); => Throws!` :+1:

Bradcomp:
I would be willing to PR this with possibly a little guidance. Would the rule be something like: 
If the first arg is a JavaScript array, and the second arg is not a JavaScript array, throw an error?

PR #1757, opened 2016-05-04
concat now throws when concatenating array with non-array

Bradcomp:
Fix for https://github.com/ramda/ramda/issues/1755

### DIFF
--- a/src/concat.js
+++ b/src/concat.js
@@ -1,6 +1,6 @@
-var flip = require('./flip');
+var _curry2 = require('./internal/_curry2');
+var _isArray = require('./internal/_isArray');
 var invoker = require('./invoker');
-
 
 /**
  * Returns the result of concatenating the given lists or strings.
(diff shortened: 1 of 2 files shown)


src/concat.js
```diff
@@ -23,4 +23,9 @@ var invoker = require('./invoker');
  *      R.concat([4, 5, 6], [1, 2, 3]); //=> [4, 5, 6, 1, 2, 3]
  *      R.concat('ABC', 'DEF'); // 'ABCDEF'
  */
-module.exports = flip(invoker(1, 'concat'));
+module.exports = _curry2(function concat(a, b) {
+  if (_isArray(a) && !_isArray(b)) {
+    throw new TypeError(toString(b) + ' is not an array');
```
scott-christopher: It looks like we're missing an import of `toString` in here.

Bradcomp: Right you are!

scott-christopher:
Thanks Brad. This looks good to me. 🍄

I'll hold off to give the others a chance to comment before merging.

buzzdecafe:
this is a step in the right direction, because it protects us from the polymorphic native JS concat behavior. 

Should we update the docs to notify that the semantics of `R.concat` are different from `Array.prototype.concat`? 

Otherwise :cow2:

Bradcomp:
Would something like this work (based on note for reduce)? 
"Note: `R.concat` expects both arguments to be of the same type, unlike the native `Array.prototype.concat` method.  It will throw an error if you `concat` an Array with a non-Array value."

buzzdecafe:
LGTM @Bradcomp thx

CrossEye:
:herb:

PR #1758, opened 2016-05-04
added sub equals function

jimmymain:
I have added a partial equals method.
It's useful for filters and such.
```
R.subequals({ id: 4, name: 'fred' }, { id: 4, name: 'fred', surname: 'basset' }) => true
```

What additional items are required for this to be accepted?
I will write tests and documentation for this method if that is likely to result in the change request being accepted. Perhaps it should be named differently?


iplus26:
Simply comment the lines to check the length of two objects as we discussed in [SO](http://stackoverflow.com/questions/37020029/comparing-partial-objects-in-ramda-js/37020772). However I'm not quite sure about the completion and more tests should be done. Good luck man!

buzzdecafe:
I think this can be done with `whereEq`

jimmymain:
@buzzdecafe - thanks, missed that little trick. I will close the pull request.

CrossEye:
The [StackOverflow post][so] that prompted this PR looked for additional capability that I'm not sure can be handled by `whereEq`:

> I would prefer the recursive nature of `equals` remain intact - if that's possible.

and agreed with a comment that this would mean that 

```js
eq({foo: {bar: 'test'}}, {foo: {bar: 'test', additional: 'item'})
```

It's an interesting notion.  I'd be interested in seeing if we could achieve it on top of existing Ramda functions for the Cookbook, rather than as a new function, as I'm not sure it's quite important enough for that.  But it certainly is interesting.


  [so]: http://stackoverflow.com/questions/37020029

buzzdecafe:
for some context, we explicitly opted not to descend recursively in `where*` here https://github.com/ramda/ramda/pull/1618

I also think it's an interesting topic for a cookbook recipe

CrossEye:
> for some context, we explicitly opted not to descend recursively in `where*`

Right, and I certainly would not want to open that up, but there is a reasonable place for `subobject`, if only in the Cookbook.

jimmymain:
the recursion was a nice to have, I thought it might come in handy down the road. for my current purposes `whereEq` works just fine. There's no discontent (from me anyway) that the `where*` functions are not sufficient. I do find @davidchambers proposed 

```R.where({address: {street: [R.equals('123 Main Street')]}})```

appealing, but I am not sure I *actually need it...* :+1:

Issue #1709, opened 2016-04-03
R.sequence Is Really Slow


svozza:
I was messing about with `sequence` today and found that it's pretty much unusable with large arrays. If I change the array size from 10 to 1000 in the link below I get script timeout errors in my browser:

http://goo.gl/mAs6TS

I'm pretty sure I know what's causing it by looking at the implementation:

```js
module.exports = _curry2(function sequence(of, traversable) {
  return typeof traversable.sequence === 'function' ?
    traversable.sequence(of) :
    reduceRight(function(acc, x) { return ap(map(prepend, x), acc); },
                of([]),
                traversable);
});
```

The call to `prepend` there is the culprit because, unlike in a language like Haskell where `prepend` operations on lists are really cheap they're a killer with arrays. To prepend anything to an array is an `O(n)` operation so what's happening is that there are `m x O(n)` operations being called inside the `reduce`, which itself is `O(n)` operation.

CrossEye:
A lot of our latest code has veered from our earlier style of building everything from scratch and not depending on many other of our functions.  This is clearly of the later style.  It probably should be fixed, but it's always a shame to see something based on the FantasyLand-specified functions (see also `liftN`) get altered.

svozza:
I'm not really sure if there's another way of doing it. I did manage to write something in an imperative way that was much quicker but the implementation nearly made me want to cry. I'm almost too embarrassed to post it but I will in the hope that someone else can come up with something better:

http://goo.gl/2vWHfq

scott-christopher:
This appears to be a combination of both the implementation of `sequence` _and_ the performance of `ap` and `map` of the applicative types used.

For example, if we redefine `sequence` to something like the following, your example still halts for long enough for the browser to intercept with 1000 elements:
```js
function sequence_(of, traversable) {
  var res = of([]),
      i = 0;
  while (i < traversable.length) {
    res = ap(map(x => xs => { xs.push(x); return xs }, traversable[i]), res);
    i += 1;
  }
  return res;
}
```

If however we use the same `sequence_` as above using the Maybe type from Ramda Fantasy instead, we can scale this up to more than 1,000,000 elements before it starts to push the friendship.

```js
let xs = [];
for(let i = 0; i < 1000000; i++) {
  xs.push(Maybe.Just(i));
}
Maybe.maybe([], identity, sequence_(Maybe.of, xs)).length;
```

[...]

PR #1759, opened 2016-05-04
Remove dependency on R.invoker in R.concat

svozza:
In [1709](https://github.com/ramda/ramda/issues/1709#issuecomment-205089684) it was mentioned that recently several internal dependencies have crept into the Ramda code base. Looking at the [recent pull request](https://github.com/ramda/ramda/pull/1757) for `R.concat`, I noticed the dependency on `invoker`. While I can appreciate the better error message I can't help but think that a function as fundamental as `concat` should be implemented in its own terms. Also, not that important but on my machine it seems like this version is around 20-30 per cent quicker on arrays with 1000 or less elements:

http://goo.gl/SgU42T

### DIFF
--- a/src/concat.js
+++ b/src/concat.js
@@ -1,6 +1,6 @@
 var _curry2 = require('./internal/_curry2');
 var _isArray = require('./internal/_isArray');
-var invoker = require('./invoker');
+var _isFunction = require('./internal/_isFunction');
 var toString = require('./toString');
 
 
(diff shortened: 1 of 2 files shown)


src/concat.js
```diff
@@ -30,8 +30,11 @@ var toString = require('./toString');
  *      R.concat('ABC', 'DEF'); // 'ABCDEF'
  */
 module.exports = _curry2(function concat(a, b) {
+  if (a == null || !is(Function, a.concat)) {
```
davidchambers: Let's use `_isFunction` rather than `is(Function)` to avoid dependence on `instanceof` (#1756).

CrossEye: I guess, but something needs to be done about `is`.

svozza: Ah yes, I knew there was a better way to check for a function. Also, there was a slight issue where `isFunction` was named `isNumber` so I fixed that too.

davidchambers:
:deciduous_tree:

src/internal/_isFunction.js
```diff
@@ -1,3 +1,3 @@
-module.exports = function _isNumber(x) {
```
buzzdecafe: :laughing: whoops!

PR #1763, opened 2016-05-06
do not filter out invalid values in R.fromPairs

davidchambers:
In Ramda's early days we were guilty of Underscore-like special cases. ;)


buzzdecafe:
wow! nice catch. that is nasty

:cow2:

CrossEye:
Glad we no longer write any bad code.  :smiling_imp: 

:herb:

svozza:
This should be flagged as a breaking change in the upgrade guide.

PR #1764, opened 2016-05-07
add R.typeEq; deprecate R.is and R.propIs

davidchambers:
See #1756 for prior discussion.

### DIFF
--- a/src/is.js
+++ b/src/is.js
@@ -13,6 +13,7 @@ var _curry2 = require('./internal/_curry2');
  * @param {Object} ctor A constructor
  * @param {*} val The value to test
  * @return {Boolean}
+ * @deprecated since v0.22.0
  * @example
  *
  *      R.is(Object, {}); //=> true
--- a/src/propIs.js
+++ b/src/propIs.js
@@ -16,6 +16,7 @@ var is = require('./is');
  * @param {*} obj
  * @return {Boolean}
  * @see R.is, R.propSatisfies
+ * @deprecated since v0.22.0
  * @example
  *
  *      R.propIs(Number, 'x', {x: 1, y: 2});  //=> true
--- a/src/typeEq.js
+++ b/src/typeEq.js
@@ -1,0 +1,26 @@
+var _curry2 = require('./internal/_curry2');
+var type = require('./type');
+
+
+/**
+ * Takes a type name (as returned by [`type`](#type)) and a value of
+ * any type, and returns `true` if the value is a member of the type
+ * corresponding to the type name; `false` otherwise.
+ *
+ * Does not respect inheritance.
+ *
+ * @func
+ * @memberOf R
+ * @category Type
+ * @sig String -> a -> Boolean
+ * @param {String} typeName
+ * @param {*} val
+ * @return {Boolean}
+ * @example
+ *
+ *      R.typeEq('Array', [1, 2, 3]); //=> true
+ *      R.typeEq('Object', [1, 2, 3]); //=> false
+ */
+module.exports = _curry2(function typeEq(typeName, val) {
+  return type(val) === typeName;
+});
--- a/test/typeEq.js
+++ b/test/typeEq.js
@@ -1,0 +1,17 @@
+var R = require('..');
+
+var eq = require('./shared/eq');
+
+
+describe('typeEq', function() {
+
+  it('indicates whether a value is a member of a specified type', function() {
+    function Foo() {}
+    eq(R.typeEq('Array', [1, 2, 3]), true);
+    eq(R.typeEq('Object', [1, 2, 3]), false);
+    eq(R.typeEq('Object', new Foo()), true);
+    eq(R.typeEq('Number', 42), true);
+    eq(R.typeEq('Number', new Number(42)), true);
+  });
+
+});


davidchambers:
((:bell:)) @buzzdecafe, @CrossEye

buzzdecafe:
apologies, i have been absolutely plowed under lately. will try to get back into this tonight

buzzdecafe:
`is` is such a good *name*, but that's not much of a justification for keeping it. is the `toString` solution off the table? (I'm not sure how to get the strings to line up exactly.)

davidchambers:
> is the `toString` solution off the table?

Are you referring to `R.match(/function (\w*)/, String(type))`? I'm willing to take this approach, but I'm not sure @CrossEye is.

buzzdecafe:
right, i think @crosseye may be the one you have to convince -- although if he is not sufficiently prompt i recommend asking forgiveness later. 

and if that rx means we can keep `is`, i am for that. i am also fine with adding `typeEq`.

CrossEye:
I still don't see anything that changes my position from #1756.

I agree that we should avoid `instanceof`, and I have no problem using kangax's solution for the known types.  For custom types, I would be willing to simply document the potential problem with realms.  I find working with inheritance more important than being consistent across realms.  But maybe that's because I almost never do any cross-realm work.  I know we have this issue in our REPL.  What other real-world experience do people have with this?

One example is that although FL's specification is all about duck-typing, there are other ways to build ADTs.  One I've been noodling with might allow you to define:

```js
// ... Define Maybe, Just, Nothing in a typical fashion.
Monad(Maybe, {
  of: //...
  chain: //...
  ap: //...
});
```

And then I really want all of these:

```js
is(Just, Just(42)); //=> true
is(Maybe, Just(42)); //=> true
is(Monad, Just(42)); //=> true
is(Functor, Just(42)); //=> true
```

There might  be some hiccups in this, due to the diamond inheritance between `Apply` and `Monad`, but wherever these are available, I would like to be able to report them correctly.  Of course there's no similar problem with my usual `is(Rectangle, new Square(5))`.

So can someone explain why the cross-realm issue is so important that we can't have useful behavior like that?

davidchambers:
> So can someone explain why the cross-realm issue is so important that we can't have useful behavior like that?

It's not just a cross-realm issue. As mentioned in sanctuary-js/sanctuary#100, it applies to npm packages within a realm. Consider a package which depends on `foo`, `bar`, and `ramda`. `bar` also depends on `foo`. `foo` provides a Foo type. `bar` provides a function of type `a -> Foo a`. Is it safe to apply `R.is(Foo)` to a value returned by this function? No, unless we're willing to rely on npm package deduplication – an optimization which should not affect a program's behaviour. Depending on whether deduplication occurs, our `Foo` and `bar`'s `Foo` may or may not be identical.

> ```javascript
> is(Just, Just(42)); //=> true
> ```

A specialized function—`isJust :: Maybe a -> a`—seems more appropriate to me, as it doesn't require the Maybe type to be implemented with constructor functions and prototypal inheritance. I see `R.is` as a function that we should ideally never need to use, as we should always know our values' types. `isJust`, on the other hand, is useful even in a well-typed program. I'd like to avoid complecting these two things.

CrossEye:
(Once again, I've been off the scene for a while due both to health and connectivity problems.)

>> ```javascript
>> is(Just, Just(42)); //=> true
>> ```

> A specialized function—`isJust :: Maybe a -> a`—seems more appropriate to me, as it doesn't require the Maybe type to be implemented with constructor functions and prototypal inheritance. 

So I'm missing how `isJust` is supposed to work cross-realm and reliably.  ISTM that it would suffer from the same sort of issues, or from similar ones if you start depending upon magic strings to capture type information.


> I see `R.is` as a function that we should ideally never need to use, as we should always know our values' types. `isJust`, on the other hand, is useful even in a well-typed program. I'd like to avoid complecting these two things.

We obviously live in very different worlds.  There are plenty of times when I don't know my value's types.  OTOH, I almost never use a function that tries to distinguish between `Just` and `Nothing`, simply using `map` or `chain` in ways that distinguish the subtypes.

davidchambers:
> So I'm missing how `isJust` is supposed to work cross-realm and reliably.

Here's one option:

```javascript
//    Nothing :: Maybe a
const Nothing = {
  '@@type': 'my-package/Maybe',
  isNothing: true,
  isJust: false,
};

//    Just :: a -> Maybe a
const Just = x => ({
  '@@type': 'my-package/Maybe',
  isNothing: false,
  isJust: true,
  value: x,
});

//    isNothing :: Maybe a -> Boolean
const isNothing = maybe => maybe.isNothing;

//    isJust :: Maybe a -> Boolean
const isJust = maybe => maybe.isJust;
```

> There are plenty of times when I don't know my value's types.

Do you resolve this uncertainty at the edges of your programs, or allow it to permeate? I'm happy to have functions with types such as `Any -> Maybe SomeKnownType` to ingest foreign values, I just wouldn't want to have type-checking logic strewn throughout a program.

> I almost never use a function that tries to distinguish between `Just` and `Nothing`, simply using `map` or `chain` in ways that distinguish the subtypes.

That's the right approach. The implementations of the `map` and/or `chain` methods must differentiate the two, though, so even if the user of the data type is never expected to do so herself she will benefit from a reliable differentiator. The `map` and `chain` methods of an implementation which uses `instanceof` will behave incorrectly in some cases.

CrossEye:
```js
class Character {
  constructor(name, characteristics) {
    t11his.name = name;
    this.characteristics = characteristics || [];
  }
  isLoyal() {
    return this.characteristics.includes('loyal');
  }
  isJust() {
    return this.characteristics.includes('just');
  }
  isBrave() {
    return this.characteristics.includes('brave');
  }
  isNothing() {
    return this.characteristics.length === 0;
  }
}

const fred = new Character('Fred', ['loyal', 'brave']);
```

>> There are plenty of times when I don't know my value's types.

> Do you resolve this uncertainty at the edges of your programs, or allow it to permeate? I'm happy to have functions with types such as `Any -> Maybe SomeKnownType` to ingest foreign values, I just wouldn't want to have type-checking logic strewn throughout a program.

Again, I think we write very different types of programs.  My programs tend to have only small bits of code between places where there's some external input, either from a user or from an outside service.  So the "edges of my programs" are less clear-cut that you seem to expect.  And often these programs need to be reasonably robust in the case of bad input from any of these sources.  If a service I'm calling does not respond, or throws an unexpected error, or if a user enters odd data or clicks buttons in totally ridiculous order, the program should do no worse than move to a friendly-looking, and usually recoverable error state.  So I'm much more loathe to introduce type errors where I know you would prefer to do so.

I still find `is` a fairly compelling function.  You've certainly succeeded in demonstrating that it's not as useful as one would hope.  But it still comes across as a reasonable abstraction.

davidchambers:
> My programs tend to have only small bits of code between places where there's some external input, either from a user or from an outside service. So the "edges of my programs" are less clear-cut that you seem to expect.

This is enlightening! I can see that for programs which are mostly edges, `R.is`, `typeof`, `instanceof`, etc. are invaluable. Our different requirements explain the philosophical differences between Ramda and Sanctuary. I'd love for you to one day spend some time walking me through some of your code, if you're willing and able to share it.

As for your `isNothing` counterexample, one could first checking that the value's `@@type` property is equal to `'my-package/Maybe'`. A value *could* have a `@@type` property whose value happens to exactly match that which we're expecting, so it's not a perfect solution, but the failure condition is hypothetical whereas I've actually been bitten by `instanceof` fragility several times.

CrossEye:
> This is enlightening! [ ... ] Our different requirements explain the philosophical differences between Ramda and Sanctuary. 

I agree.  I use Sanctuary a fair bit for my personal projects, and am considering trying PureScript sometime soon.  But in my day job, I don't even consider it, even though I'm using Ramda fairly heavily now.

I think there is one other difference that also explains a lot: My day job involves working with a large, ever-shifting cast of junior developers.  Sometimes even `map` and `filter` are alien concepts.  Trying to work with more abstract concepts gets to involve too much of my own time in explanations. 

> I'd love for you to one day spend some time walking me through some of your code, if you're willing and able to share it.

Willing, absolutely.  Able, only if I'm willing to do some real work to sterilize it and remove all proprietary stuff.  Even then my company wouldn't like it, but then they'd never know.  Yes, it's that kind of place.  But I'm sure I can find some examples at some point.

PR #1765, opened 2016-05-10
dissoc should only create a new object when necessary

louisremi:
So don't create a new object when the prop isn't found in it.

### DIFF
--- a/src/dissoc.js
+++ b/src/dissoc.js
@@ -11,7 +11,8 @@ var _curry2 = require('./internal/_curry2');
  * @sig String -> {k: v} -> {k: v}
  * @param {String} prop the name of the property to dissociate
  * @param {Object} obj the object to clone
- * @return {Object} a new object similar to the original but without the specified property
+ * @return {Object} a new object similar to the original but without the specified property,
+ *         or the same object if the property isn't found in the object
  * @see R.assoc
  * @example
  *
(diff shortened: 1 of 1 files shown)


louisremi:
Hi guys, not sure if this PR makes sense for this project.
I'm using Ramda with Redux, and try to only "touch" the state when necessary. In the case of dissoc, I'd like the function to return the passed object when the prop isn't found in it.

Let me know if that is an acceptable change of behavior for this function, before I try to see what's wrong with the tests.

Thanks!

davidchambers:
Ramda may not play nicely with code which relies on object identity. Ramda embraces value equality; from this perspective the identity of the value returned by `dissoc` doesn't matter. I don't think we should add complexity to the implementation and documentation of a Ramda function in order to accommodate code which cares about where things live in memory.

CrossEye:
> I don't think we should add complexity to the implementation and documentation of a Ramda function in order to accommodate code which cares about where things live in memory.

Agreed.

louisremi:
On the other hand, avoiding creating new objects can be considered a perf improvement for everyone (as it limits the work of the GC). And the added complexity is very limited.
It's up to you to decide, it saddens me a bit to have to reimplement this method for my project, though.

davidchambers:
> avoiding creating new objects can be considered a perf improvement for everyone

This may be the case. I'm not opposed to optimizations, but I'm opposed to adding tests which *enforce* optimizations. As a result, even if `R.dissoc` were to behave as you would like we would not guarantee that it will continue to do so.

> it saddens me a bit to have to reimplement this method for my project

I understand the feeling. An explicit membership check may be clearest, though:

```javascript
//  Note: Since we wish to preserve object identity, dissociate only if necessary.
if (R.has(k, o)) o = R.dissoc(k, o);
```

CrossEye:
To me it's not just a matter of the tests, but also the documentation.  Right now it's simple: "Returns a new object that..."  Thus if a user doesn't follow our lead and does do some mutation on the returned object, she can count on it not altering the original one.  That seems a very good thing.

acconrad:
We don't even check if this is a property anymore so I think this PR is no longer relevant and can be closed.

PR #1767, opened 2016-05-12
Add ES3 test for R.keys

svozza:
In the discussion that flowed from sanctuary-js/sanctuary#214 we toyed with the idea of testing the fallback for when `Object.keys` is not present. In the end we decided ES3 support was more hassle than it's worth but if Ramda wants to support such ancient browsers then there really ought to be tests for these fallbacks.

### DIFF
--- a/test/keys.js
+++ b/test/keys.js
@@ -1,36 +1,5 @@
 var R = require('..');
-var eq = require('./shared/eq');
+var keys = require('./shared/keys');
 
 
-describe('keys', function() {
-  var obj = {a: 100, b: [1, 2, 3], c: {x: 200, y: 300}, d: 'D', e: null, f: undefined};
-  function C() { this.a = 100; this.b = 200; }
-  C.prototype.x = function() { return 'x'; };
-  C.prototype.y = 'y';
-  var cobj = new C();
-
-  it("returns an array of the given object's own keys", function() {
-    eq(R.keys(obj).sort(), ['a', 'b', 'c', 'd', 'e', 'f']);
-  });
-
-  it('works with hasOwnProperty override', function() {
-    eq(R.keys({
-      /* jshint -W001 */
-      hasOwnProperty: false
-      /* jshint +W001 */
-    }), ['hasOwnProperty']);
-  });
-
-  it('works for primitives', function() {
-    /* jshint elision: true */
-    var result = R.map(function(val) {
-      return R.keys(val);
-    }, [null, undefined, 55, '', true, false, NaN, Infinity, , []]);
-    eq(result, R.repeat([], 10));
-  });
-
-  it("does not include the given object's prototype properties", function() {
-    eq(R.keys(cobj).sort(), ['a', 'b']);
-  });
-
-});
+keys('keys', R.keys);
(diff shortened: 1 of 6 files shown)


test/keys.js
```diff
@@ -13,6 +15,19 @@ describe('keys', function() {
     eq(R.keys(obj).sort(), ['a', 'b', 'c', 'd', 'e', 'f']);
   });
 
+  it('should work when Object.keys is not present', function() {
+    var code =
+      '(function(require) { ' +
+      'var R = require(".."); ' +
+      'var eq = require("./shared/eq"); ' +
+      '' +
+      'Object.keys = void 0; ' +
```
davidchambers: This should come before requiring Ramda, I believe, as Ramda may already have saved a reference to `Object.keys` by this point. Also, I think `delete Object.keys` would be slightly better here, as it would more accurately reflect the fact that the `keys` property is absent in old browsers.

svozza: :zap:

davidchambers:
I don't think this actually tests what we intend it to test.

```console
$ echo 'module.exports = Object.keys;' >/tmp/keys.js

$ node --print $'require("vm").runInNewContext(\'delete Object.keys; require("/tmp/keys")\', {require: require})'
[Function: keys]
```

I believe the problem is that we're proving `require` from *this* context, so `require('..')` will give us the Ramda module from this context, in which `Object.keys` *is* defined.

We may need to do something like `delete require.cache[require.resolve('..')]`, but it's fiddly.

svozza:
Have a look at my last update there, is that what you had in mind?

davidchambers:
> Have a look at my last update there, is that what you had in mind?

I don't think it does what we want it to do.

```console
$ git pull git@github.com:svozza/ramda.git keys-es3-test

$ echo 'module.exports = Object.keys;' >src/keys.js

$ make && node_modules/.bin/mocha --grep Object.keys -- test/keys.js
```

The test still passes for me, suggesting that we're getting Ramda from *this* context rather than the sandbox context.

svozza:
Yeah, I've tried quite a few things, even passing in `rewire` instead of `require` but no luck. Unless someone has a solution I wonder that if it can't be tested then should it be in the code base?

davidchambers:
We could define `_keysFallback` and test this internal function directly. It's not the perfect solution, of course, but if it saves us from running actual browser tests it's worth doing. ;)

svozza:
That's a good idea. I'm happy to update to update the PR with this solution if people are happy to go that oute.

svozza:
Well that was painful, it's quite annoying the way the tests are ran against the file in `dist`, it makes setting breakpoints very tedious. Anyway, turns out none of the tests in `test/internal` were being ran and doing `--recursive` gives this weird error, which again I presume is related to running the tests from `dist`:

```
Error: Cannot find module '..'
```

I can't face fighting with the build process anymore so I've just explicitly added the `internal` folder to `npm test`.

One question, why are functions of arity 1 being curried?

davidchambers:
> One question, why are functions of arity 1 being curried?

For consistent placeholder behaviour. `R.inc(R.__)(R.__)(R.__)(R.__)(R.__)(0) = 1`. ;)

buzzdecafe:
> For consistent placeholder behaviour. 
should #1843 be merged, that constraint would disappear.

customcommander:
@CrossEye Given the roadmap and the plan for v2 this doesn't seem something we'd like to pursue. Close?

PR #1768, opened 2016-05-14
run tests on node 6

svozza:
Node 6 is going to be the new LTS version of Node so probably best to start running the tests against it.


buzzdecafe:
thanks
:cow2:

PR #1771, opened 2016-05-17
fix: add correct comments escaping for R.not example

Delapouite:
The REPL is broken without this fix.


buzzdecafe:
thank you! 
:cow2:

PR #1775, opened 2016-05-20
Improve performance of object-to-object code paths

### LINKED ISSUE #1773: map(object) performance could be improved
http://goo.gl/CIbMkr

I get 0.4 in Chrome ; 0.55 in Firefox.

asaf-romano:
Closes #1773

### DIFF
--- a/src/filter.js
+++ b/src/filter.js
@@ -1,10 +1,9 @@
+var _accumObj = require('./internal/_accumObj');
 var _curry2 = require('./internal/_curry2');
 var _dispatchable = require('./internal/_dispatchable');
 var _filter = require('./internal/_filter');
 var _isObject = require('./internal/_isObject');
-var _reduce = require('./internal/_reduce');
 var _xfilter = require('./internal/_xfilter');
-var keys = require('./keys');
 
 
 /**
@@ -36,12 +35,11 @@ var keys = require('./keys');
 module.exports = _curry2(_dispatchable('filter', _xfilter, function(pred, filterable) {
   return (
     _isObject(filterable) ?
-      _reduce(function(acc, key) {
+      _accumObj(function(key, value, newObject) {
         if (pred(filterable[key])) {
-          acc[key] = filterable[key];
+          newObject[key] = value;
         }
-        return acc;
-      }, {}, keys(filterable)) :
+      }, filterable) :
     // else
       _filter(pred, filterable)
   );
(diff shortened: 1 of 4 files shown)


CrossEye:
:herb:

src/filter.js
```diff
@@ -36,12 +35,11 @@ var keys = require('./keys');
...
-      _reduce(function(acc, key) {
+      _accumObj(function(key, value, newObject) {
         if (pred(filterable[key])) {
-          acc[key] = filterable[key];
+          newObject[key] = value;
         }
-        return acc;
-      }, {}, keys(filterable)) :
+      }, filterable) :
     // else
```
jdalton: This bypasses bug fixes in `keys`.

asaf-romano: hrm?

jdalton: See https://github.com/ramda/ramda/blob/v0.21.0/src/keys.js#L24-L32.

asaf-romano: The same applies to all those functions that already traverse objects this way (e.g. ``evolve``). I guess copying over these properties doesn't break much. Also note that current implementation of ``keys``, in ES5+ environments, only takes care of the Safari enum bug.

svozza: The hoops Ramda jumps through to support ancient versions of IE seems crazy to me.

jdalton: How many hoops?

svozza: The stuff in `R.keys` and `R.trim` are the ones that spring to mind initially. I've better things to be doing on a Saturday night than trawling through the Ramda source code for more examples. ;)

jdalton: It looks like the bulk of the `R.trim` work is less for old IE and more for consistent whitespace removal.

svozza: Or you could just use the native `trim()` function even though it's not supported in IE8.

jdalton: As I wrote above the bulk of `R.trim` is guarding against buggy native `trim` implementations.

asaf-romano:
Regarding the original issue raised by @jdalton, I'll see what it takes to incorporate the workarounds into the new helper function (without hurting performance). I'm personally more worried about the Safari bug, which seem to apply to the current browser version (I'll have to test, but the code says Safari 9). IE8 is a lost cause.

If it plays out, I may also fix ``evolve`` and other functions to use the new helper. Currently they do not apply any of these workarounds.

CrossEye:
Too long abandoned.  If we need this, we should start afresh.

PR #1777, opened 2016-05-23
fix Duplicate declaration "add" error in example in REPL

aonghusonia:


davidchambers:
I prefer to use `(a, b) => a + b` here. The problem with using `R.add` (or any other binary Ramda function) is that it supports both `f(a, b)` and `f(a)(b)`. Crucially, though, `R.reduce` requires the former. The lambda makes this apparent whereas `R.add` would not.

buzzdecafe:
:cow2:

aonghusonia:
I have no problem with (a, b) => a + b
the problem is that R.add  has been added to the global scope on the REPL as add and the REPL is giving a duplicate declaration error. How about:

var plus = (a, b) => a + b

buzzdecafe:
`plus` is fine with me

CrossEye:
`:plus1:` on `plus`.

buzzdecafe:
thanks

PR #1781, opened 2016-05-25
fix signature and example of composeK

benperez:
This looks wrong. From haskell's [Kleisli composition](http://hackage.haskell.org/package/base-4.9.0.0/docs/Control-Monad.html#v:-62--61--62-):

``` haskell
(>=>) :: Monad m => (a -> m b) -> (b -> m c) -> a -> m c
```

This supposed to produce a function which takes as input a normal type and returns a monad. There was also an incorrect function in the example

### DIFF
--- a/src/composeK.js
+++ b/src/composeK.js
@@ -15,7 +15,7 @@ var prepend = require('./prepend');
  * @memberOf R
  * @since v0.16.0
  * @category Function
- * @sig Chain m => ((y -> m z), (x -> m y), ..., (a -> m b)) -> (m a -> m z)
+ * @sig Chain m => ((y -> m z), (x -> m y), ..., (a -> m b)) -> (a -> m z)
  * @param {...Function}
  * @return {Function}
  * @see R.pipeK
@@ -24,9 +24,8 @@ var prepend = require('./prepend');
  *      //  parseJson :: String -> Maybe *
  *      //  get :: String -> Object -> Maybe *
  *
- *      //  getStateCode :: Maybe String -> Maybe String
+ *      //  getStateCode :: String -> Maybe String
  *      var getStateCode = R.composeK(
- *        R.compose(Maybe.of, R.toUpper),
  *        get('state'),
  *        get('address'),
  *        get('user'),


davidchambers:
Did you play around with `R.composeK`? The current type signature seems to match the actual behaviour:

```javascript
> R.composeK(S.Maybe.of)(S.Just(42))
S.Just(42)
```

On a related note, we should define `(>=>)` in Sanctuary. :)

benperez:
> Did you play around with R.composeK

Yes, after looking at the implementation, it seems to be written to match the documentation.

I'd argue that the version I suggested is an improvement, though, for 3 reasons:

1. If you were to write a function for sequencing a series of monadic operations, why would the input need to already be a monad? It seems like it would just lead to a lot of uneccessary wrapping of things with `of` aka `return`. We actually do this in the example as it's written now.
2. It would also simplify the implementation of the function, no need to [prepend an identity function to the front of the list](https://github.com/ramda/ramda/blob/64990c5cc491dfe8e13055b50f67f91c20312911/src/composeK.js#L42).
3. I guess ultimately it feels like you're unnecessarily wrapping/unwrapping the input argument. If you really wanted to use the output of this function with a monad as input, you would just `bind`/`chain` it to that monad, but that feels like a less common use case.
```
# how R.composeK should work
R.composeK(f, g, h)(a) => R.chain(f, R.chain(g, h(a)))
# how R.composeK works right now
R.composeK(f, g, h)(a) => R.chain(f, R.chain(g, R.chain(h, a)))
```
__Bonus reason__: it matches the haskell version, but I know that won't sway you 😝

davidchambers:
I find your reasoning persuasive. :)

I'm not sure `R.pipeK` and `R.composeK` pull their weight, though. I've used `R.pipeK` two or three times. I've never used `R.composeK`. Does anyone use one or both of these functions regularly?

If we're to keep these functions, making the change you suggest sounds good to me.

CrossEye:
I've never used either of them, so I don't have a strong opinion.

Bradcomp:
I've used `pipeK` a few times, and I like that they're there.  The changes seem like they would be better for the majority of use cases.

benperez:
I'll wait until you decide whether or not they are going to stick around before I go through the trouble of updating them. I don't want to brush the cobwebs off these functions if they're just getting thrown out!

davidchambers:
I'm happy to keep them. :)

PR #1782, opened 2016-05-26
changes juxt code example to prevent name conflict with range

thurt:
This looks like it would only be a problem if the ramda functions are not namespaced under `R.` etc. For the Ramda REPL, the original code example does produce an error


buzzdecafe:
yup, seen this sort of thing before. Thanks for the fix
:cow2:

Issue #1497, opened 2015-11-04
make gt, gte, lt, lte, etc... right associative


luizbills:
actual:
```
var gtFive = R.gt(5);
gt10(5); // => false
```

proposal:
```
var gtFive = R.gt(5);
gt10(5); // => true
```

davidchambers:
:+1:

buzzdecafe:
been down this road before. Under this proposal `gte(5)(6) !== gte(5, 6)`. that is surprising. this is where the placeholders came from, the desire to right-section the non-commutative infix operators.

davidchambers:
I'm not suggesting shenanigans. I'm in favour of swapping the order of the arguments. One can easily write `x < y` if both values are known, so these functions are primarily useful in conjunction with partial application. Let's optimize for partial application (as we do elsewhere in the library).

buzzdecafe:
>  I'm in favour of swapping the order of the arguments.

Does this argument order hold for any non-commutative infix operator converted to prefix?

```js
gt(5, 10) === true // ?
divide(20, 2) === 0.1 // ?
subtract(10, 9) === -1 // ?
```

How about `concat`?

Feels very familiar.

davidchambers:
`divide` and `subtract`, absolutely!

`concat` I'm not sure about. I do like the fact that `R.reduce(R.concat, '', ['foo', 'bar', 'baz'])` evaluates to `'foobarbaz'`.

CrossEye:
>  I do like the fact that `R.reduce(R.concat, '', ['foo', 'bar', 'baz'])` evaluates to `'foobarbaz'`.

That's my take on the whole concept.  Although I'm very frustrated by `gt(5)` not feeling correct, enough so that I've [gone searching][so] for solutions that lead us [down inappropriate paths][rt] in the past., equivalences like these seem pretty important.  It's not just `concat`.<sup>1</sup>

```js
R.reduce(R.subtract, 12, [1, 2, 3, 4])
```

should be equivalent to 

```js
12 - 1 - 2 - 3 - 4
```

This proposal would replace that with

```js
4 - (3 - ( 2 - (1 - 12)))
```

which is neither as intuitive nor as understandable.

My biggest difficulty is with the fact that I would want to remain consistent across all our operators.  Here, I think are all the current functions that can reasonably be called operators, including a few more questionable ones (`merge`, `range`, `zip`, `zipWith`), with the non-commutative ones highlighted.


  * add
  * **both**
  * **concat**
[...]

PR #1784, opened 2016-05-28
Updated eslint to v2 

bryanbierce:
I am not sure if the eslint config is meant to be targeting v2.0 yet so this may not be a desired update. I found the warning about the "space-after-keywords" being replaced in v2.0 and replaced it with its up to date rule "keyword-spacing".

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -86,9 +86,9 @@
     ],
     "space-infix-ops": 2,
     "space-return-throw-case": 0,
-    "space-after-keywords": [
+    "keyword-spacing": [
       2,
-      "always"
+      { "after": true, "before": true }
     ],
     "lines-around-comment": 0,
     "space-before-blocks": [
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,7 +1,8 @@
 {
   "env": {
     "node": true,
-    "mocha": true
+    "mocha": true,
+    "es6": true
   },
   "globals": {
     "Symbol": true
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "dox": "latest",
     "envvar": "1.x.x",
     "escodegen": "1.4.x",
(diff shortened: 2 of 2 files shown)


buzzdecafe:
i think we should upgrade to latest eslint. just hasn't gotten to the top of the list yet. if you update eslint version in  package.json, does the build pass? if so, then let's do that

bryanbierce:
Quick update: I tried updating to latest and am running into a problem with eslint not recognizing 'ArrayBuffer', 'Int8Array', and 'Promise' and throwing a `no-undef` error when linting. I'm digging in to see if I can figure out why and get it passing.

bryanbierce:
Well that was easy enough :) es6 environment needed to be explicitly declared for it to permit es6 globals. v2.0 is in and all tests are passing :)

CrossEye:
Then I'm all for this.  Any objections?

buzzdecafe:
thanks @bryanbierce

bryanbierce:
Yeah, glad to contribute something :) Are you all open to/looking for more contributors to the project? If so do you have anywhere in particular I might start lending a hand while I get my head around its scope?

CrossEye:
We certainly welcome new contributors.  We'd love to have your help.

There are plenty of open issues, but we have an unfortunate (perhaps!) tendency to use our issues log as a discussion forum rather than a backlog of problems to fix.  But feel free to poke around there, or even better, find where your own pain points are and suggest a change based on those.

#1497 just got some new attention.  I [suggested][ss] a collection of names that might help settle down the issues with non-commutative binary operators.  This would be a slightly broad, but very shallow change.  I think we'd need to see if people wanted it, first, though.  OTOH, a PR is by far the best way of seeing if an idea really is popular.


  [ss]: https://github.com/ramda/ramda/issues/1497#issuecomment-195644724

PR #1786, opened 2016-05-29
data Maybe a = Just a | Nothing

buzzdecafe:
ok here it is. i think there are still some gaps.

1. it is not possible to invoke `Just.Nothing()` because the placeholder-tail is wagging the ramda-dog. In order to present a consistent API, i have to `_curry1(Nothing)`, effectivekly making `Nothing` a unary function instead of a nullary one. That just doesn't feel right. If I don't do this, then it fails an invariant test. 

~~2. I don't think the tests are correct. It is passing the "foldable" test although `reduce` is not on `Maybe` (yet). So some investigation is required there.~~


CrossEye:
I know we dropped #1749, but would it make sense to create tests that demonstrate how to use this to create a `safeHead`/`safeFind`?  Just dropping them in seems quite unmotivated: Why `Maybe` and not `Future`/`Either`/whatever?

Bradcomp:
Maybe I'm off-base, but I am struggling to see how this fits in with the project.  Why would I use Ramda's constructors instead of just using Folktale, or Sanctuary, or Ramda-Fantasy, or something else.  I could see the usefulness of the helper function that created a Maybe in #1749, but I can also understand the reasons it got nixed.

Now that that is gone, I am left wondering why Ramda should be providing type constructors at all.

> Why Maybe and not Future/Either/whatever?

☝️  I think this is a good question.  `Maybe` is a useful type, but so are the other container types.  Ramda provides functions for manipulating these objects, but does it need to provide the types themselves?  

I could see providing constructors for custom types that get returned from Ramda's functions, but Ramda doesn't return any custom types at the moment due to the prior PR being eliminated.

CrossEye:
@Bradcomp: This is still meant to fill the job of #1749.  The point was the complexity of adding a useful generic `maybe` function didn't seem worth it.  But by simply exposing `Maybe`, we can allow users to do the same thing where it makes sense to them.

My question was rhetorical.  I do understand why we want to add only `Maybe` and its subtypes ... it's simply to make it straightforward for users to add type-safe versions of their own functions or of Ramda functions, something similar to what Sanctuary does automatically, but on the users' own terms.  I asked the question only because without any examples of doing this, the PR seems very unmotivated.

src/internal/_Maybe.js
```diff
@@ -0,0 +1,66 @@
...
+  };
+  Just.prototype.toString = function() {
+    return 'Just(' + toString(this.value) + ')';
+  };
+  Just.prototype.isNothing = false;
+  Just.prototype.isJust = true;
+
+  function returnThis() { return this; }
+
+  function Nothing() {}
```
davidchambers: It would be nice to use `Nothing` rather than `Nothing()`. See sanctuary-js/sanctuary#225.

davidchambers:
The title is missing the type variable. It should read `data Maybe a = Just a | Nothing`.

buzzdecafe:
> The title is missing the type variable. 

:zap:

src/internal/_Maybe.js
buzzdecafe: funny that is the approach ramda-fantasy took; that there is a single `Nothing`. But once it's released in the wild it can get mutated by naughty developers. I feel better guaranteeing that `Nothing()` will return a well-behaved `Nothing` at the risk of many `Nothing`s

CrossEye: Is there any guarantee that a naughty developer won't mutate `Nothing.prototype` or override the `Nothing` constructor?  If not, then I think the simpler single `Nothing` wins.

buzzdecafe: it's a lot easier to glom a flag onto an instance than it is to extract the prototype from the instance and then mutate that. if that fails, then good. and as you know, jr. developers love jamming flags on objects. FOP

joneshf: Would making it a getter work? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get

```js
Object.defineProperty(Maybe.prototype, 'Nothing', {
  get: () => ({
    map: returnThis,
    ...
  }),
});
```

Or whatever syntax magic you need to make that work.

buzzdecafe: that could be fine. i don't have particularly strong feelings one way or the other. One `Nothing` makes sense -- but the possibility of that `Nothing` getting polluted once released into the wild worries me. Many `Nothing`s seems strange, but at least guarantees that the "factory" generates a different reference.Unless of course, @CrossEye 's scenario plays out, where they extract the prototype and muck with that. Ah, javascript. ¯\_(ツ)_/¯

asaf-romano:
Surprised to see most JS implementations of ``Maybe`` don't implement ``filter`` (Sanctuary being the notable exception), and following some Haskell detour, I was wondering if it's worth implementing honoring MonadPlus (chain + of + empty, in short) in ``R.filter`, here's a little demo:

http://goo.gl/V1XAk1

As you may see, I had to extend Maybe.prototype (of ramda-fantasy) to implement ``empty`. I then noticed this implementation of Maybe doesn't implement ``empty`` either. I don't think it would hurt to do so (``Maybe.empty()`` returning ``Nothing()`` seems intuitive). Question is, how much we want in-house types to strictly follow fantasy-land types.

buzzdecafe:
> Question is, how much we want in-house types to strictly follow fantasy-land types.

i think the benefits outweigh the costs.

CrossEye:
> `Maybe.empty()` returning `Nothing()` seems intuitive.

My own intuition here so often leads me astray that I've learned not to trust it too much.  Would we also want to implement `concat` at some point?  And if so, we're presumably trying to live with the `Semigroup` and `Monoid` laws, but I'm not sure if that is the best Monoid instance to work with or not.  There is certainly something to be said, when working on a Monoid of `Maybe String`, to have `empty` return `Just('')`` or for a Monoid of `Maybe Number` over addition to have it return `Just(0)`.  It's not that such is necessarily better, but it seems equally defensible.  The big trouble is that we want both for `Maybe` to be meaningful on its own, and for it to stand for a whole collection of `Maybe a` types.  I'm not sure we can really do that.

asaf-romano:
FWIW
```haskell
λ mplus (Just 4) (Just 0)
Just 4
```

PR #1788, opened 2016-05-31
Fixed typo in tryCatch example

suside:


buzzdecafe:
whoops. looks like the docs didn't keep up with the PR. _mea culpa_. thanks for the fix

PR #1793, opened 2016-06-08
fix R.take extra step

### LINKED ISSUE #1792: Transduce does more work than necessary
Changing the function in the example http://ramdajs.com/docs/#transduce to print each time it's been called:
```
var addOne = x => {
  console.log('add one to', x);
  return x + 1;
}

var numbers = [1, 2, 3, 4];
var transducer = R.compose(R.map(addOne), R.take(2));

R.transduce(transducer, R.flip(R.append), [], numbers); //=> [2, 3]
```
Console prints:
```
add one to
1
add one to
2
add one to
3
```
I see #921 had the same behaviour and was supposed to be closed by #923 , but the extra step persists.

cristianocd:
Fixes #1792 

 - Reintroduce #923 fix that was reverted by #1245 
 - Add test case

### DIFF
--- a/src/internal/_xtake.js
+++ b/src/internal/_xtake.js
@@ -2,21 +2,18 @@ var _curry2 = require('./_curry2');
 var _reduced = require('./_reduced');
 var _xfBase = require('./_xfBase');
 
-
 module.exports = (function() {
   function XTake(n, xf) {
     this.xf = xf;
     this.n = n;
+    this.i = 0;
   }
   XTake.prototype['@@transducer/init'] = _xfBase.init;
   XTake.prototype['@@transducer/result'] = _xfBase.result;
   XTake.prototype['@@transducer/step'] = function(result, input) {
-    if (this.n === 0) {
-      return _reduced(result);
-    } else {
-      this.n -= 1;
-      return this.xf['@@transducer/step'](result, input);
-    }
+    this.i += 1;
+    var ret = this.n === 0 ? result : this.xf['@@transducer/step'](result, input);
+    return this.i >= this.n ? _reduced(ret) : ret;
   };
 
   return _curry2(function _xtake(n, xf) { return new XTake(n, xf); });
--- a/test/take.js
+++ b/test/take.js
@@ -1,4 +1,5 @@
 var assert = require('assert');
+var sinon = require('sinon');
 
 var R = require('..');
 var eq = require('./shared/eq');
@@ -37,6 +38,12 @@ describe('take', function() {
 
   it('handles zero correctly (#1224)', function() {
     eq(R.into([], R.take(0), [1, 2, 3]), []);
+  });
+
+  it('steps correct number of times', function() {
+    var spy = sinon.spy();
+    R.into([], R.compose(R.map(spy), R.take(2)), [1, 2, 3]);
+    sinon.assert.calledTwice(spy);
   });
 
 });
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "ramda": "0.17.x",
     "rimraf": "~2.3.2",
     "sanctuary": "0.7.x",
+    "sinon": "^1.17.4",
     "testem": "0.9.x",
     "uglify-js": "2.4.x",
     "xyz": "0.5.x"


test/take.js
```diff
@@ -39,4 +39,17 @@ describe('take', function() {
     eq(R.into([], R.take(0), [1, 2, 3]), []);
   });
 
+  it('steps correct number of times', function() {
+    var counter = 0;
+    var incCounter = function(x) { counter += 1; return x; };
+    var takeXTimes = function(x) {
+      counter = 0;
+      R.into([], R.compose(R.map(incCounter), R.take(x)), [1, 2, 3]);
```
davidchambers: I think it would be clearer to replace `incCounter` with its definition, so `counter` could be local to `takeXTimes`.

src/internal/_xtake.js
```diff
@@ -15,7 +15,8 @@ module.exports = (function() {
       return _reduced(result);
     } else {
       this.n -= 1;
-      return this.xf['@@transducer/step'](result, input);
+      return this.n === 0 ? _reduced(this.xf['@@transducer/step'](result, input))
+                          : this.xf['@@transducer/step'](result, input);
     }
```
davidchambers: Have both an `n == 0` case and an `n == 1` case is unfortunate. Is there any way to simplify this? There may not be.

cristianocd: I don't think so, but I'm not familiar with the codebase. 

I've tried a couple of things without success.
Adding some kind of init flag might make it less clear.

What do you say? Should we keep it like this?

kedashoe: > Is there any way to simplify this?

Depends what you mean by simplify, but right now we are overloading `n`, so to speak. For me it would be simpler/clearer to add something like `this.idx = 0;` in the constructor and use that to count up towards `this.n` (which I think would also make the first `this.n === 0` check clearer).

Another alternative:

```js
    this.n -= 1;
    return (
      this.n < 0   ? _reduced(result) :
      this.n === 0 ? _reduced(this.xf['@@transducer/step'](result, input)) :
                     this.xf['@@transducer/step'](result, input));
```

davidchambers: > For me it would be simpler/clearer to add something like `this.idx = 0;` in the constructor and use that to count up towards `this.n` (which I think would also make the first `this.n === 0` check clearer).

This sounds good to me. :)

test/take.js
```diff
@@ -39,4 +39,19 @@ describe('take', function() {
     eq(R.into([], R.take(0), [1, 2, 3]), []);
   });
 
+  it('steps correct number of times', function() {
+    var takeXTimes = function(x) {
+      var counter = 0;
+      R.into(
+        [],
+        R.compose(R.map(function(x) { counter += 1; return x; }), R.take(x)),
```
buzzdecafe: running `into` for side-effects and the side-effect inside `map` is not a good example of library usage. is there a more idiomatic way to illustrate this?

cristianocd: I think the best way would be using 3rd party lib like sinon.js. A spy would make it very clear.

buzzdecafe: i agree that sounds like a good solution. let's add `sinon` to `devDependencies`

TheLudd: Note that a spy is just another side effect and what @cristianocd has created here is indeed a spy function already.

buzzdecafe: i understand that, but what bothers me is modeling undesirable practices in the test suite. the test suite should be examples of how the library should be used. It makes sense to me to write idiomatic code for the test, and let the *test framework* handle the side-effects.

cristianocd: sinon bonuses: DRY, clear
```
    var spy = sinon.spy();
    R.into([], R.compose(R.map(spy), R.take(2)), [1, 2, 3]);

    sinon.assert.calledWithExactly(spy, 1)
    sinon.assert.calledWithExactly(spy, 2)
    sinon.assert.neverCalledWith(spy, 3)
```

src/internal/_xtake.js
cristianocd: What else is going on after `this.xf['@@transducer/step'](result, input))` is returned?
Maybe we can achieve the desired behaviour somewhere else, I'm just having a hard time tracing the calls. I don't get why calling `_reduce` on the result of the call, vs on the result on the next iteration would have different behaviour.

kedashoe: Ya, it is quite a bit going on between dispatching for Fantasy Land and the same functions supporting transducers vs their normal behavior, I get lost pretty much every time I come back to it :p

This line `this.xf['@@transducer/step'](result, input))` basically calls the next function in the transducer stack. If we return without calling that, we end up back [in one of our reducing functions (which one depending on the input)](https://github.com/ramda/ramda/blob/master/src/internal/_reduce.js). In that file, you will also see the check for the special `reduced` value, which breaks out of the reducing.

cristianocd: Thanks @kedashoe . So, using your suggestion, what do you guys think about something like:
``` javascript
  XTake.prototype['@@transducer/step'] = function(result, input) {
    this.i += 1;
    var ret = this.n === 0 ? result : this.xf['@@transducer/step'](result, input)
    return this.i === this.n ? _reduced(ret) : ret;
  };
```

CrossEye:
Lots of good discussion here, and useful changes.  It certainly reads better with Sinon.

Any objections to merging as it stands?

buzzdecafe:
do it

PR #1794, opened 2016-06-09
Currently trying where is broken due to _

krainboltgreene:
Fixes a bug in the "try this function" feature.


buzzdecafe:
thank you

PR #1811, opened 2016-06-28
prepublish: support old versions of GNU bash

davidchambers:
See ramda/ramda.github.io#87

### DIFF
--- a/scripts/prepublish
+++ b/scripts/prepublish
@@ -1,9 +1,10 @@
 #!/usr/bin/env bash
 set -e
 
+slash=/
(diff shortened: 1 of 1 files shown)


CrossEye:
LGTM, but I don't really know my makefile from an unmade bed, so someone else should review...

davidchambers:
I was able to reproduce the error by running the old version of the script with an old version of bash, so I believe this fix is sound. :)

buzzdecafe:
finally had a chance to give this a quick spin -- looks good, thanks @davidchambers

PR #1818, opened 2016-07-08
deprecate R.wrap

davidchambers:
I've wanted to deprecate this function for a long time, as I've yet to see an example not better expressed *without* `wrap`. Consider the examples in the documentation:

```javascript
var greet = name => 'Hello ' + name;

//  Option 1: R.wrap
var shoutedGreet1 = R.wrap(greet, (gr, name) => gr(name).toUpperCase());

//  Option 2: R.compose
var shoutedGreet2 = R.compose(R.toUpper, greet);

//  Option 3: lambda
var shoutedGreet3 = name => greet(name).toUpperCase();

//  Option 1: R.wrap
var shortenedGreet1 = R.wrap(greet, function(gr, name) {
  return gr(name.substring(0, 3));
});

//  Option 2: R.compose
var shortenedGreet2 = R.compose(greet, R.take(3));

//  Option 3: lambda
var shortenedGreet3 = name => greet(name.substring(0, 3));
```

`wrap` introduces indirection in these cases and gains us nothing. Note that we're providing `greet` as an argument to `wrap` only for `wrap` to then provide it to us as an argument (which we name `gr`). Why not simply reference `greet` directly? :stuck_out_tongue_winking_eye:

### DIFF
--- a/src/wrap.js
+++ b/src/wrap.js
@@ -16,6 +16,7 @@ var curryN = require('./curryN');
  * @param {Function} fn The function to wrap.
  * @param {Function} wrapper The wrapper function.
  * @return {Function} The wrapped function.
+ * @deprecated since v0.22.0
  * @example
  *
  *      var greet = name => 'Hello ' + name;


buzzdecafe:
i agree the examples in the docs are not persuasive. The only real-world use-cases I can imagine for `wrap` are for adding aspects to functions, i.e., side-effects. So I am fine with removing it.

buzzdecafe:
btw
:cow2:

buzzdecafe:
i do wanna give @CrossEye the opportunity to defend this one

davidchambers:
> The only real-world use-cases I can imagine for `wrap` are for adding aspects to functions, i.e., side-effects.

We don't need a special function for this, though. We can simply use a lambda:

```javascript
x => {
  // Pre-processing
  const x$ = pre(x);
  // Guarding
  const x$$ = pred(x$) ? f(x$) : defaultValue;
  // Post-processing
  return post(x$$);
}
```

In JavaScript decorators such as this are quite natural, unlike in Python, say, whose limited lambdas prompted the inclusion of special *syntax* for decorators to avoid the ugliness of function declarations.

CrossEye:
Big surprise that I'm the holdout here! :smile:  I'm not sure I disagree.  But I have used this, and used it differently than these examples.  Perhaps this could easily be turned into a nice lambda too.  But please recall that this was written with ES3 - ES6 in mind, so having nice lambdas seemed a luxury at one point.

The function is not well designed.  For my uses, the parameters are in the wrong order

I have used it with code something like this:

```js
const secure = wrap(__, (f, user, val) => user.loggedIn ? f(val) : 'security error');

const foo = val => val + 1;
const secureFoo = secure(foo);

secureFoo({loggedIn: true}, 42); //=> 43
secureFoo({loggedIn: false}, 42); //=> "security error"

const bar = val => val * 2;
const secureBar = secure(bar);

secureBar({loggedIn: true}, 42); //=> 84
secureBar({loggedIn: false}, 42); //=> "security error"
```

I use it to create *generic* function wrappers that change the behavior of the wrapped functions in consistent ways, including altering their signatures.  If we are to keep it, I don't know if there's any way to document it better.  To my mind, these are the real point of the function, but you need to show the multiple examples for the point to really be clear; that's a pretty long example for our docs.

@buzzdecafe is right that this is all about side-effects.  We don't have many functions like this, but there are a few, and I think they do serve a purpose, ones like `tap`, `once`, and `memoize`.  If there is a clean way to do the above without `wrap`, then I would have no issue removing it.  But it is something I've used.

davidchambers:
If the goal is to avoid the unnecessary evaluation of an effectful statement or expression, a thunk would make this clearer to me:

```javascript
// Pretend that foo and bar are effectful.

const secure = (user, thunk) => user.loggedIn ? thunk() : 'security error';

const foo = val => val + 1;

secure({loggedIn: true}, () => foo(42));  // => 43
secure({loggedIn: false}, () => foo(42));  // => "security error"

const bar = val => val * 2;

secure({loggedIn: true}, () => bar(42));  // => 84
secure({loggedIn: false}, () => bar(42));  // => "security error"
```

It's up to you to decide whether this function gets to stay, Scott. I'm not fussed either way.

CrossEye:
> If the goal is to avoid the unnecessary evaluation of an effectful statement or expression, a thunk would make this clearer to me

No, the side-effects I'm talking about are in the actual function which is wrapping the one supplied.  So, for instance, one might be looking up the user information from the session instead of a parameter as suggested above.  One might be logging to the console or to a tracking system.  One might be preventing unnecessary execution, or delaying execution, or simply managing a work queue.

Most frequently, we wouldn't change the signature of the underlying function at all, simply AOP-style adjusting the parameters supplied to it or the results it generates.

> It's up to you to decide whether this function gets to stay, Scott. I'm not fussed either way.

You've brought this up a few times, so it seems like something that bothers you at least a bit.  I'm not wedded to it, not at all.  But I think the documentation examples don't do it justice.  If we want to get rid of it, we need to understand how it might really be used.  And I'm not convinced that replacing them with lambdas would always be as easy as suggested.

davidchambers:
> I'm not convinced that replacing them with lambdas would always be as easy as suggested.

I believe this is the main point of disagreement. I don't see how providing a function to `R.wrap` only for it to give it back to us ever achieves anything. In other words, there should be a mechanical transformation from any expression involving `R.wrap` to a simpler one which does not use `R.wrap`. For example:

```javascript
var g = R.wrap(<expr>, function(f, x) { <body> });
```

```javascript
var g = (function() {
  var f = <expr>;
  return function(x) { <body> };
}());
```

If we do the transformation by hand rather than mechanically we can avoid the IIFE by ensuring that `f` does not conflict with another variable of the same name in scope. This gives:

```javascript
var f = <expr>;
var g = function(x) { <body> };
```

This is objectively more direct than the `R.wrap` equivalent. My claim is that such a transformation is possible for every imaginable use of `R.wrap`. I'll be happy to see a counterexample, of course. If a counterexample is provided, let's use it as the example in the documentation.

CrossEye:
I have two problems with that.  First of all, I don't think this transformation would handle the example I used above, with a placeholder for the function name.  Am I missing that?  Or would you then need a second style for those?

Secondly, `var g = R.wrap(<expr>, function(f, x) { <body> });` is not at all exhaustive.  I think about it more like `var g = R.wrap(R.__, functionRef);`  Digging a little I find something only slightly different [from this][rr]:

```js
const logger = function(f) {
  const args = [].slice.call(arguments, 1);
  console.log('Calling function "' + f.name + '"');
  console.log('    with arguments ' + map(JSON.stringify, args));
  const result = f.apply(null, args);
  console.log('    returning ' + result);
  return result;
};

const addLogger = wrap(__, logger);

const plus = addLogger(function plus(a, b) {return a + b});
plus(25, 17); //=> 42
//: Calling function "plus"
//:     with arguments 25,17
//:     returning 42

const get = addLogger(function get(prop, obj) {return obj[prop];});
get('a', {a: 'foo', b: 'bar'}); //=> 'foo'
//: Calling function "get"
//:     with arguments "a",{"a":"foo","b":"bar"}
//:     returning foo
```

While there might be some other simple version of `addLogger`, it's hard to imagine one that's *much* simpler.


  [rr]: http://goo.gl/R1WShh

davidchambers:
Thanks for the concrete example, Scott. I rewrote it using some ES6 features to avoid having to deal with an `arguments` object:

```javascript
const addLogger = R.wrap(R.__, (f, ...args) => {
  console.log('Calling function "' + f.name + '"');
  console.log('    with arguments ' + R.map(JSON.stringify, args));
  const result = f(...args);
  console.log('    returning ' + result);
  return result;
});
```

Here's the equivalent version without `R.wrap`:

```javascript
const addLogger = f => (...args) => {
  console.log('Calling function "' + f.name + '"');
  console.log('    with arguments ' + R.map(JSON.stringify, args));
  const result = f(...args);
  console.log('    returning ' + result);
  return result;
};
```

The diff is small:

```diff
-const addLogger = R.wrap(R.__, (f, ...args) => {
+const addLogger = f => (...args) => {
   console.log('Calling function "' + f.name + '"');
   console.log('    with arguments ' + R.map(JSON.stringify, args));
   const result = f(...args);
   console.log('    returning ' + result);
   return result;
-});
+};
```

> I don't think this transformation would handle the example I used above, with a placeholder for the function name.

Are you referring to your `secure(foo)` example?

> ```javascript
> const secure = wrap(__, (f, user, val) => user.loggedIn ? f(val) : 'security error');
> ```

Here's the `wrap`-less equivalent:

```javascript
const secure = R.curry((f, user, val) => user.loggedIn ? f(val) : 'security error');
```

It turns out `wrap(__, f)` is just an intricate way of writing `curry(f)` in this case. :stuck_out_tongue_winking_eye:

buzzdecafe:
My $0.02: 

IMO @davidchambers arguments carry the day here. i am still :cow2: on whacking this function. Also note that `tap` can do side-effects with the input, and side-effects with the output can be done, well, on the output. So while I concede that `wrap` is more general and powerful than `tap`, I also don't see it adding sufficient value to be in the core lib.

CrossEye:
>  I rewrote it using some ES6 features to avoid having to deal with an `arguments` object:

You also removed an important public function that's used elsewhere: `logger`.  That's a seriously breaking change.

Again, I won't stand in the way of a consensus on removing this.  The version of `wrap` I have used most often is a heavier-weight, more completely AOP function than what we have here.  But I don't think this is useless.  It's simply not very important to me.

davidchambers:
> The version of `wrap` I have used most often is a heavier-weight, more completely AOP function than what we have here.

Since `R.wrap` is not commonly used and—based on your comment above—doesn't necessarily do quite what one would like it to do on the occasions when it is useful, leaving a user to define such a function herself seems right to me.

Thanks for the enjoyable back and forth, @CrossEye. :)

CrossEye:
> Thanks for the enjoyable back and forth, @CrossEye. :)

Actually, I almost always enjoy our exchanges, agree or disagree.  But I was having a really hard time with this one.  I felt I couldn't get my point across, and I really wanted to explain, because, well, you know, 

![Someone is wrong][xk]

And I remained frustrated, fighting my lousy internet connection, typing half what I did online today on a tablet or a phone.

I'm still not sure I ever did show how I use it and why I think it's reasonable, but really it's not that important.


-----

(image from https://xkcd.com/386/)


  [xk]: https://imgs.xkcd.com/comics/duty_calls.png "What do you want me to do?  LEAVE? Then they'll keep being wrong!"

davidchambers:
> I felt I couldn't get my point across

It does seem as though there's a thought in your head which is yet to make its way into my head.

PR #1820, opened 2016-07-09
adjustWhere

ccorcos:
Seems like `evolveWhere` could be relevant too -- perhaps even the same function. If I used R.map rather than list.map then it should work. I just don't know what we'd call it.

### DIFF
--- a/src/adjustWhere.js
+++ b/src/adjustWhere.js
@@ -1,0 +1,30 @@
+var _curry3 = require('./internal/_curry3');
+
+
+/**
+ * Applies a function to the value at the index of an array where the given
+ * condition is true, returning a new copy of the array with the element
+ * replaced with the result of the function application.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig (a -> a) -> (a -> Boolean) -> [a] -> [a]
+ * @param {Function} adj The function to apply.
+ * @param {Function} pred The predicate function.
+ * @param {Array|Arguments} list An array-like object whose value
+ *        where the predicate is true will be replaced.
+ * @return {Array} A copy of the supplied array-like object with
+ *         the element where the predicate is true is replaced with the value
+ *         returned by applying `adj` to the existing element.
+ * @see R.adjust
+ * @example
+ *
+ *      R.adjustWhere(R.add(1), R.equals(1), [0, 1, 1]);    //=> [0, 2, 2]
+ *      R.adjustWhere(R.add(1))(R.equals(1))([0, 1, 1]);    //=> [0, 2, 2]
+ */
+module.exports = _curry3(function adjust(adj, pred, list) {
+  return list.map(function(item) {
+    return pred(item) ? adj(item) : item;
+  });
+});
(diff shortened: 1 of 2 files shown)


CrossEye:
Mostly, this LGTM.

I think it would be more idiomatic if we were to call the `map` function rather than a map method.  It probably wouldn't change much.  It should also note that this is available for other functors besides lists.  So perhaps:


```js
module.exports = _curry3(function adjust(adj, pred, functor) {
  return map(function(item) {
    return pred(item) ? adj(item) : item;
  }, functor);
});
```

With some tests to show these behaviors:

```js
R.adjustWhere(R.add(1), R.equals(2), [0, 1, 2, 2]); //=> [0, 1, 3, 3]
R.adjustWhere(R.add(1), R.equals(2), Just(2)); //=> Just(3)
R.adjustWhere(R.add(1), R.equals(2), Just(5)); //=> Just(5)
R.adjustWhere(R.add(1), R.equals(2), Nothing()); //=> Nothing()
```

My only other concern is that I'm not convinced that the first two parameters are in the right order.  I'm not sure they aren't but I hate to get these things wrong.

buzzdecafe:
i'm not sold on the arg order or the name (sorry for the bikeshedding). 

* On the name: `where` and `whereEq` take a spec object; it'd be nice to keep that name-API linkage consistent. 
* On the argument order: IMO this function is a `map` that only applies the mapping to the elements that satisfy the predicate. So I propose something like:

```
mapWhen :: (a -> Bool) -> (a -> b) -> [a] -> [b]
```

davidchambers:
The function needs to be `a -> a` rather than `a -> b`, @buzzdecafe. The generalized signature you're proposing (which I like) is:

```haskell
mapWhen :: Functor f => (a -> Boolean) -> (a -> a) -> f a -> f b
```

I then wonder why one would use

```javascript
R.mapWhen(pred, f, functor)
```

rather than

```javascript
R.map(R.when(pred, f), functor)
```

I don't consider this worthwhile sugar.

buzzdecafe:
> The function needs to be `a -> a` rather than `a -> b`, @buzzdecafe. 

I guess it depends on what you mean by "needs"; e.g.:

```js
mapWhen(gt(__, 1), toString, [1, 2, 3]) //=> [1, '2', '3']
```

Now you could say that the mixed-type list that comes out is pretty gross. And I would agree. But it is a possible `a -> b` outcome in this dirty js world we find ourselves in. :-)

But I think this quibble is moot anyways, since `R.map(R.when(pred, f), functor)` captures the same thing. So I'm afraid I am a "no" on this PR. 

once again I am agreeing with @davidchambers ! either @davidchambers is getting dumber or i am getting smarter.

CrossEye:
> once again I am agreeing with @davidchambers ! either @davidchambers is getting dumber or i am getting smarter. 

And I'm agreeing with both.  Is the apocalypse upon us?

Although this is potentially useful, as demonstrated it's easily achievable, and a dedicated function could be written like this:

```js
const mapWhen = curry((pred, adj, functor) => map(when(pred, adj), functor));
```

ccorcos:
ah yes. fair enough. i didn't know about `R.when`

buzzdecafe:
yeah i missed it myself. thanks for the pr, i appreciate you taking the time.

PR #1821, opened 2016-07-10
Add propPath function

ryanaghdam:
I've added a new function, `propPath`, which returns a function that returns the value at the supplied path.  It is meant to complement `prop` as `assocPath` does `assoc`.

### DIFF
--- a/src/propPath.js
+++ b/src/propPath.js
@@ -1,0 +1,31 @@
+var _curry2 = require('./internal/_curry2');
+var _slice = require('./internal/_slice');
+
+/**
+ * Returns a function that, when supplied a path, returns the indicated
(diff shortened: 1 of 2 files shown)


asaf-romano:
Isn't this basically path without the second argument?

ryanaghdam:
D'oh!  It is.

I've been using Ramda for almost a year and I missed that one.  I had been using `R.compose` and `R.prop`.

asaf-romano:
We should probably cross-referenece (See ...) the two, if they're not already.

ryanaghdam:
That's a great idea.  I'll open a pull request that does that.

asaf-romano:
👍

Issue #1832, opened 2016-07-19
Unexpected behavior for `evolve` when `transformations` contains a null value.


tansongyang:
The following code produces unexpected output ([repl](http://goo.gl/PgJSY7)):

```JavaScript
var o  = {a: 123, b: 456, c: 789};
var transformations = {
  a: null,
  b: undefined,
  c: 'foo',
};
R.evolve(transformations, o); // -> {"a": {}, "b": 456, "c": 789}
```

The value of `a` became an empty object instead of its original value, `123`.

Looking at the [code](https://github.com/ramda/ramda/blob/master/src/evolve.js#L38), it seems like this could be resolved by changing the check from `type === 'object'` to `type && type === 'object'` (unless this is expected behavior).

buzzdecafe:
yeah that sucks. your solution seems fine to me though. would you like to do a PR?

CrossEye:
Yes, I think that should be fixed.  I think you will have to do `transformation && type === 'object'` instead, though.  Javascript, in its infinite, er, wisdom, has decided that `typeof(null); //=> 'object'`.

tansongyang:
@CrossEye Good catch! @buzzdecafe I'd be glad to do PR. Will let you know when it's done.

tansongyang:
@buzzdecafe Let me know if there are any issues. I took a best guess at what the style should be.

PR #1833, opened 2016-07-19
Issue 1832

tansongyang:
Addresses issue #1832 .

### DIFF
--- a/src/evolve.js
+++ b/src/evolve.js
@@ -34,9 +34,9 @@ module.exports = _curry2(function evolve(transformations, object) {
   for (key in object) {
     transformation = transformations[key];
     type = typeof transformation;
-    result[key] = type === 'function' ? transformation(object[key])
-                : type === 'object'   ? evolve(transformations[key], object[key])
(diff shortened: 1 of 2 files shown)


buzzdecafe:
this is an improvement on what's there. so i am in favor merging this. 

OTOH i don't like silently passing bad data down the line. We were doing that here anyways, so it's beyond the scope of this PR. But deliberately ignoring bad input and even _testing that we will deliberately ignore bad input_ doesn't feel right to me. So we should circle back to this one.

CrossEye:
I'm also in favor, with less reservations than @buzzdecafe expresses.  I feel like Ramda is simply testing that it actually can handle the data supplied, not testing that the data is valid.  It then does the best it can with what it's given: calling a function, recursing on an object, ignoring anything else (including, with this change, null values.)

CrossEye:
Thanks, @buzzdecafe for following up on this.  I'm finally feeling better after several months of you-don't-want-to-know.  Hopefully, I'll be more attentive for a while.

PR #1836, opened 2016-07-25
Add bind example

MattMS:


src/bind.js
```diff
@@ -17,6 +17,11 @@ var _curry2 = require('./internal/_curry2');
  * @param {Object} thisObj The context to bind `fn` to
  * @return {Function} A function that will execute in the context of `thisObj`.
  * @see R.partial
+ * @example
+ *
+ *      var log = R.bind(console.log, console);
+ *      R.pipe(R.assoc('a', 2), R.tap(log), R.assoc('a', 3))({a: 1}); //=> {a: 3}
+ *      //-> {a: 2}
```
CrossEye: I might write something like

```js
// logs {a: 2}
```

instead.  The difference between `//=>` and `//->` is fairly subtle and quite difficult to notice.

CrossEye:
Outside  a minor suggestion made, this looks good to me.

MattMS:
I was just copying the format used in the [tap example](http://ramdajs.com/0.21.0/docs/#tap).

Happy to make the change though.

CrossEye:
I'm perfectly happy to merge as it is.  I didn't remember the `tap` example, and I'm not thrilled with it, but it turns out we also using it in `forEach`, so let's go with this for now.  It's easy enough to change later if decide to.

Thanks for the help!

MattMS:
No worries :smile: 

Did you want me to open an issue to discuss the formatting?

CrossEye:
If you feel like it.  That's smaller than many of our concerns, but I'd still rather not lose sight of it.

PR #1840, opened 2016-07-29
nthArg should return a curried function

buzzdecafe:


CrossEye:
:herb:

asaf-romano:
🌷

harryhazza77:
Hi,

I'm unable to upgrade to Ramda 0.22.0 and I'll try and explain why.

I have this set of utility functions defined in a file called _callbackHandlers_

```javascript
import R from "ramda";`

export const callbackFail = R.pipe(R.nthArg(0), R.isNil, R.not);
export const withError = next => R.pipe(R.nthArg(0), next);
export const withResult = next => R.pipe(R.nthArg(1), next);
```


And these tests (newly failing ones only included)

```javascript
describe("callbackHandlers", () => {
    describe("#withError", () => {
        describe("zero args", () => {
            it("should be undefined", () => {
                let argSet;
                const next = arg => argSet = arg;
                callbackHandlers.withError(next)();
                expect(argSet).to.be.undefined; // with Ramda 0.22.1 expected [Function: f1] to be undefined
            });
        });
    });
    describe("#withResult", () => {
        describe("zero args", () => {
            it("should be undefined", () => {
                let argSet;
                const next = arg => argSet = arg;
                callbackHandlers.withResult(next)();
                expect(argSet).to.be.undefined; // with Ramda 0.22.1 expected [Function] to be undefined
            });
        });
        describe("one arg", () => {
            it("should be undefined", () => {
                let argSet;
                const next = arg => argSet = arg;
                callbackHandlers.withResult(next)("a");
                expect(argSet).to.be.undefined; // with Ramda 0.22.1 expected [Function] to be undefined
            });
        });
    });
});
```

note the comments in line where all 3 tests pass prior to this change.

I've tried all sort of things in the callbackHandlers but can't get these 3 tests to pass. So hoping someone can help please.

CrossEye:
Yes, these are certainly not the sorts of cases expected when we created, or when we modified these (fairly uncommon) functions.

I think the simplest thing here would simply to skip Ramda for these simple versions:

```js
const withError = next => fst => next(fst);
const withResult = next => (fst, snd) => next(snd);
```

(and probably something similar for your other function; there was no test case for it shown.)

Functions like `nthArg` are much less useful in an ES6 environment than they were earlier.  Here it's often easy to simply use arrows like this.

PR #1843, opened 2016-07-29
move placeholder support from curry to partial

### LINKED ISSUE #1363: Drop placeholder support in curry, add placeholder support in `partial`
Quotes (from #1178):

@buzzdecafe 

> i am not thrilled with generalized placeholders, because of the amount of complexity they are baking into fundamental operations of the lib. to my mind they don't give you much you can't get with flip. the cost exceeds the benefit.

@CrossEye 

> I'm starting to join @buzzdecafe's distaste for these things.

Back then I didn't want to remove support for placeholders (I like them). But it just occurred to me that we could drop placeholder support from curry but still have it available by adding support for it to `R.partial` and `R.partialRight`.

Having placeholder support in curry adds an invocation overhead to every single function in Ramda and functions created with `R.curry` and `R.curryN`.

Placeholder syntax is nice but I don't think it's used often enough that having it in curry is warranted.

I suggest dropping placeholder support from all our curry functions and instead adding placeholder support to `R.partial`. Thus people can still get the benefits of placeholders when it's really needed.

I don't have any benchmarks about the actual cost of placeholder support. But I think the simplifications we can get to `_curryN`, `_curry1`, `_curry2` and `_curry3` would be enough in itself.

As far as I'm away our binary math functions are the primary use case for placeholders. But users can `R.flip` them or maybe we could provide flipped versions of them.

Doing this would close #1178 and #1027.

kedashoe:
Closes #1363. WIP.
- placeholder support in `partial`
- no more curry1 :emoji:
- don't need [test/internal](https://github.com/ramda/ramda/tree/master/test/internal) anymore, only files in there were curry{2,3}.js and they were only testing R.__ support
- curry about 10% faster according to benchRunner
- added `rearg`. Simple enough function I think, don't really have a strong opinion one way or the other if it should be included.

todo:
- clean up the rest of the tests using `R.__`
- check doc examples for `R.__`
- `partialRight`
- other?

Let me know how it's looking so far.. son is waking up from nap, gotta run! bbl


buzzdecafe:
wow, this is a lot of work. thanks so much for taking a whack at this. I am very much looking forward to seeing how this proceeds once your son gets back to sleep

src/divide.js
```diff
@@ -17,9 +17,6 @@ var _curry2 = require('./internal/_curry2');
  *
  *      R.divide(71, 100); //=> 0.71
  *
- *      var half = R.divide(R.__, 2);
- *      half(42); //=> 21
- *
```
CrossEye: I think we'd still want to show how to do this, either with `flip` or with `partial`.

(Unless we ever decide to add things like `divideBy` and cross-reference the documentation.)

src/rearg.js
```diff
@@ -0,0 +1,45 @@
...
+ *      var double = R.partial(multiply, [2]);
+ *      double(2); //=> 4
+ *
+ *      var greet = (salutation, title, firstName, lastName) =>
+ *        salutation + ', ' + title + ' ' + firstName + ' ' + lastName + '!';
+ *
+ *      var sayHello = R.partial(greet, ['Hello']);
+ *      var sayHelloToMs = R.partial(sayHello, ['Ms.']);
+ *      sayHelloToMs('Jane', 'Jones'); //=> 'Hello, Ms. Jane Jones!'
+ */
```
CrossEye: Are these the examples for another function?  They don't seem to mention `rearg`.

test/addIndex.js
```diff
@@ -68,7 +68,7 @@ describe('addIndex', function() {
     it('passes the index as an additional parameter', function() {
       eq(mapFilterIndexed(
         R.multiply,
-        R.gt(R.__, 13),
+        R.lte(13),
```
CrossEye: First off, `R.gt(R.__, x)` is equivalent to `R.lt(x)`, not `R.lte(x)`.  It's an easy mistake to make, as that's how the converses of comparison operators work.  But this is not a converse.

But more importantly, I think if we don't introduce some other function such as `isGt` and simply crossreference them in the docs, then we show how to achieve this by using `gt`, either with `flip` or `partial`.

CrossEye:
This is great stuff!  I'm sorry I won't be around over the next two weeks to see how it proceeds.

I really, really, like seeing all those `curry1` calls vanishing!  

:rabbit: :tophat:

src/divide.js
kedashoe: This example still lives, I moved it to `partial`. But we can move it back here if that is preferred.

src/rearg.js
kedashoe: woops yep, I'll update this if we decide to keep rearg

kedashoe:
> wow, this is a lot of work. thanks so much for taking a whack at this. I am very much looking forward to seeing how this proceeds once your son gets back to sleep

won't have much time to code over the weekend, but I'll be sure to keep everyone appraised of what the kids are up to :stuck_out_tongue:

test/addIndex.js
kedashoe: Sounds good, I'll flip it.

kedashoe:
There are several tests like this in various test files

```js
  it('behaves right curried when passed `R.__` for its first argument', function() {
    var lt5 = R.lt(R.__, 5);
    eq(lt5(10), false);
    eq(lt5(5), false);
    eq(lt5(3), true);
});
```

Seems unnecessary to me, shall I just get rid of these? Or use the placeholder with partial for all of them?

buzzdecafe:
IMO they should be removed. If you want to test `partial`/`placeholder` behavior that should be in `partial*` tests, or could be consolidated into a placeholder test file. Then if regressions pop up, we can add tests as needed.

kedashoe:
Ok I think this can be more thoroughly reviewed now

- updated `partialRight`
- tests for `partial` and `partialRight`
- removed `R.__` from examples
- got rid of `rearg`. can always add it later if there is a clamoring for it

let me know how all that looks and if I've forgot anything

buzzdecafe:
I love this, I think it is the right thing to do. I am a bit anxious about regressions this may cause out there; but I would rather err on the side of doing the right thing than legacy support for a lib that is not even at 1.0 yet.

That said, I note that curried functions will throw if the last argument is undefined, but return a reference to the function otherwise:

```js
> R.reduce()();
[Function: f3]
> R.reduce(R.add)();
[Function: f2]
> R.reduce(R.add, 0)();
TypeError: Cannot read property 'reduce' of undefined
```

In my opinion, *all* of these should throw. Invoking a curried function with no arguments makes no sense and should be an error. Others may disagree -- so let's figure it out. And it may be beyond the scope of this PR. 

That said, I am strongly :+1: :cow2: on this PR.

CrossEye:
Just back from vacation, and simply triaging issues at the moment.  This one will take more than that.  I'm putting it at the top of my 'to investigate more thoroughly' list.

buzzdecafe:
((:bell:))

src/internal/_xreduceBy.js
```diff
@@ -1,7 +1,6 @@
-var _curryN = require('./_curryN');
 var _has = require('./_has');
 var _xfBase = require('./_xfBase');
-
+var curryN = require('./curryN');
```
scott-christopher: This should either be `require('../curryN')` or `require('./_curryN')` here.

src/of.js
```diff
@@ -20,4 +19,4 @@ var _of = require('./internal/_of');
  *      R.of(null); //=> [null]
  *      R.of([42]); //=> [[42]]
  */
-module.exports = _curry1(_of);
+module.exports = _of;
```
scott-christopher: We might as well just inline `internal/_of` here now.

src/reduced.js
```diff
@@ -20,9 +19,9 @@ var _reduced = require('./internal/_reduced');
...
  *
  *      R.reduce(
- *        R.pipe(R.add, R.when(R.gte(R.__, 10), R.reduced)),
+ *        R.pipe(R.add, R.when(R.flip(R.gte)(10), R.reduced)),
  *        0,
  *        [1, 2, 3, 4, 5]) // 10
  */
 
-module.exports = _curry1(_reduced);
+module.exports = _reduced;
```
scott-christopher: And `internal/_reduced` could be inlined here.

src/partialRight.js
```diff
@@ -26,4 +30,37 @@ var flip = require('./flip');
  *
  *      greetMsJaneJones('Hello'); //=> 'Hello, Ms. Jane Jones!'
  */
-module.exports = _createPartialApplicator(flip(_concat));
+module.exports = _curry2(function partialRight(fn, partialArgs) {
```
scott-christopher: Is it perhaps worth simplifying the implementation of `partialRight` by deriving it from `partial`?

Something like:
```js
function partialRight(fn, partialArgs) {
    var leftArgs = [];
    var leftN = fn.length - partialArgs.length;
    while (leftN > 0) {
        leftArgs.push(__);
        leftN = leftN - 1;
    }
    return partial(fn, leftArgs.concat(partialArgs));
}
```

scott-christopher:
Aside from that this is looking great, especially the reduced complexity in the various `_curry*` functions.

src/of.js
kedashoe: by inlined do you mean

`module.exports = require('./internal/_of');`

? It doesn't look like our build is equipped to handle that, it leaves the `require` in the output:

`var of = require('./internal/_of');`

Or something else you had in mind?

scott-christopher: Sorry, the suggestions was just to move the contents of `internal/_of` into here and then remove `internal/_of.js`.

kedashoe: Ah! okay, sounds good.

src/partialRight.js
kedashoe: Looks good, I think it is worth it.

kedashoe: :zap:

scott-christopher:
:mushroom: Great work @kedashoe!

(looks like we have a merge conflict that needs to be resolved though)

paldepind:
I think  this is wonderful and whould really like to see it merged!

davidchambers:
I suggest that we leave this pull request open until we agree on a course of action in #1912. We don't want a repeat of the `reduce` → `foldl` → `reduce` fiasco!

CrossEye:
Agreed.

PR #1844, opened 2016-07-30
constrained curry

buzzdecafe:


CrossEye:
I probably shouldn't get into a discussion like this as I'm leaving for a two-week vacation in about 28 hours.  But when has that ever stopped me? :smiling_imp: 

I'm bothered by what seems to me a half-measure here.  It's probably easiest to explain by example.  One of our most useful functions is `map`.  It has the signature `Functor f => (a → b) → f a → f b`.  What would a set of constraints for this look like?  If we wanted to create, say, `mapNumberLists`, we might be able to write a condition for the second argument reasonably well, perhaps `all(is(Number))`.  But how would we be able to deal with the first one?  I don't see any reasonable possibilities.  But saying something like, "Oh, you can't type-check functions," when functions are probably the most important type used in Ramda, seems at most a half-measure.  I really think that if you want this in JS, you should probably work in a compile-to-JS language.

Trying to build static typing on top of dynamic typing seems a lot like trying to create case-sensitive comparisons out of case-sensitive ones: the information necessary is already missing, and there is simply no useful way of getting it back.

buzzdecafe:
> But saying something like, "Oh, you can't type-check functions,"

straw man much? i don't recall anybody suggesting that. You can solve this problem by constraining the function before it is passed to `map`.

```js
const numToStr = curryStrong([R.is(Number)], a => a.toString());
const mapNumberLists = ([R.is(Function), R.all(R.is(Number))], map);
```

I agree that trying to bolt static type-checking on to JavaScript is probably a fool's errand. The environment is completely hostile to it. But I find this approach interesting enough to propose.

CrossEye:
Feel free to ignore this. I'm leaving on vacation in a few hours and likely won't see your response for two weeks.  Maybe I'm just tired, but I'm confused by this:

>  You can solve this problem by constraining the function before it is passed to `map.`

If you're suggesting that Ramda can do that, then I would really like to know how: it strikes me as impossible.  If you mean that the user can do that, of course, but she can do the same for every other parameter as well.   What would be the point?

buzzdecafe:
> If you're suggesting that Ramda can do that, 

no

> If you mean that the user can do that, of course, but she can do the same for every other parameter as well. What would be the point? 

indeed. what is the point of `curry` anyway? the user can pass in all the arguments

davidchambers:
@buzzdecafe, could you explain your motivation? Why would one use `R.curryStrong` rather than `def` (created via [`$.create`][1]). It lacks support for return-type validation, type variables, type-class constraints, and higher-order functions<sup>†</sup>.

Is the goal to provide a lightweight alternative to sanctuary-def which provides some of the benefits with a smaller performance penalty?

---

† It's not sufficient for the caller to provide as arguments functions which themselves perform type checking. In the case of `(a -> b) -> Array a -> Array b`, for example, all the function argument could do is assert that it is applied to one argument. The constraints only exist within the context of an application of the higher-order function.


[1]: https://github.com/sanctuary-js/sanctuary-def

buzzdecafe:
Two motivations, in no particular order:

* Provide a lightweight, non-prescriptive way to add constraints to functions. I see this as a _building block_ for "typing," not a solution.
* Purely exploratory. There is something interesting here, and I want to see if I can get at it.

> Is the goal to provide a lightweight alternative to sanctuary-def which provides some of the benefits with a smaller performance penalty?

No (except for "lightweight"). See above. At the moment, all I want to do is start with adding constraints to functions. Then I want to see what it would take to extend it in other directions. Frankly, I am not familiar enough with sanctuary-def to know if or how much this overlaps.

Point-by-point:

* return-type validation: 
  Agreed. but that is simple enough to add; the final constraint could apply to the return value, and test that before returning.

* type variables, type-class constraints: 
  No indeed. at the moment that is (way) beyond the scope. As noted above, I am more interested in what the building blocks are or could be, rather than a full-blown solution.

* higher-order functions:
  Yup, this is a thorny one. At present my best answer is to constrain all the way down. Maybe eventually this means that functions have to carry some metadata around with them identifying their ins and outs.

davidchambers:
> Frankly, I am not familiar enough with sanctuary-def to know if or how much this overlaps.

Well, sanctuary-def allows us to fully express the type of `R.map`:

```javascript
const a = $.TypeVariable('a');
const b = $.TypeVariable('b');
const f = $.UnaryTypeVariable('f');

//    Functor :: TypeClass
const Functor = ...;

//    map :: Functor f => (a -> b) -> f a -> f b
const map =
def('map',
    {f: [Functor]},
    [$.Function([a, b]), f(a), f(b)],
    (fn, functor) => ...);
```

This pull request is philosophically aligned with sanctuary-def. The overlap is perhaps only 10% at this stage, but the proportion will grow if you continue to work on this. ;)

CrossEye:
@davidchambers:

```js
def('map',
    {f: [Functor]},
    [$.Function([a, b]), f(a), f(b)],
    (fn, functor) => ...);
```

I'm going to have to look at this again at some point.  I'm assuming you haven't even tried to do the impossible here with `$.Function`.

So if `S.map` were to catch an error in this scenario:

```js
// With whatever numeric type annotations are needed to allow first parameter to this map
// call to accept any number and return a number...

S.map(x => x === 0 ? 'Cannot do that' : 120 / x, [20, 15, 8, 3, 0, -4, -12]);
``` 

It would successfully perform the first four iterations before bombing on the fifth one, correct?  There's no magic that lets this fail up front?

davidchambers:
> It would successfully perform the first four iterations before bombing on the fifth one, correct?

Yes, that's right.

For the sake of simplicity, let's assume the environment comprises five types:

    {Array *, Maybe *, Number, PositiveNumber, String}

Initially `a`, `b`, and `f` are unconstrained:

    a = *
    b = *
    f = *

`[20, 15, 8, 3, 0, -4, -12]` is provided as `f a`, so we determine that `f` is `Array`:

    a = *
    b = *
    f = {Array}

We then look at the elements of the array, one at a time, to constrain `a`. The first element, `20`, is a member of Number and PositiveNumber:

    a = {Number, PositiveNumber}
    b = *
    f = {Array}

The fifth element of the array, `0`, is a member of Number but not of PositiveNumber:

    a = {Number}
    b = *
    f = {Array}

Initial type checking passed, so we apply the function (the implementation of `map`). This should result in a function wrapping `x => x === 0 ? 'Cannot do that' : 120 / x` being called once for each element in the array. The function is expecting a value of type `a`, and by this point we've determined that `a` is `Number`. The function is expected to return a value of type `b`, which is still unconstrained. The first return value is `6`, giving:

    a = {Number}
    b = {Number, PositiveNumber}
    f = {Array}

The fifth return value, `'Cannot do that'`, is only a member of String. Since the intersection of `{Number, PositiveNumber}` and `{String}` is `{}` there's no possible type for `b`, so we must throw a TypeError.

Issue #1846, opened 2016-07-31
add the T combinator : R.thrush ?


guillaumearm:
Hey guys, I think add a thrush function in Ramda can be cool, is not it ?
`
const thrush = x => f => f(x);
`

an example here with lenses: http://goo.gl/RgLubh

buzzdecafe:
i'd be happy to see a PR for it

davidchambers:
I'd like to remove `R.T` and `R.F`, to make the single-upper-case-character identifiers available for combinators.

guillaumearm:
@davidchambers : today i use ramda in production, and i have some : 
`
  const otherwise = R.T
`
in my code.
Maybe lots of people use R.T and R.F in their code, and i think R.identity is more expressive than R.I

davidchambers:
> i think R.identity is more expressive than R.I

I'd say `R.always(true)` is more expressive than `R.T` by the same logic. :stuck_out_tongue_winking_eye:

guillaumearm:
Yes, you're totally right.
anyway, i did a PR for R.thrush -> #1847
Maybe we should create an issue for talking about removing R.T and R.F, and another one for adding missing combinators and rename them.
does it would mean than R.compose should be called R.K ? o_Ô

buzzdecafe:
`R.K` would have to be added. `always` is close, but really generates a thunk instead of a unary function. 

You may see some resistance on using `T` for the combinator instead of a constant `true` function.

guillaumearm:
but it means, R.compose will be removed or just R.K = R.compose ?

guillaumearm:
oups sorry, the B combinator is `compose`, and K is `always`

CrossEye:
While I have little problem removing `T` and `F`, although I would like to have simple names for these, and not depend on `R.always(true)`.

But part of the reason I'm in favor of this is that I'm not in favor of such single-character names.

S-K-I, and other combinator-systems are wonderful.  But for a general-purpose library, code that looks like `B(K, T)` looks too much like gibberish, IMHO.

PR #1847, opened 2016-07-31
add R.thrush

guillaumearm:
from #1846,
R.thrush is the T combinator.
examples : http://goo.gl/8tp3jI
with lenses : http://goo.gl/RgLubh

### DIFF
--- a/src/thrush.js
+++ b/src/thrush.js
@@ -19,21 +19,16 @@ var _curry2 = require('./internal/_curry2');
  * @return {*} Tun result of applying f on x.
  * @example
  *
+ *      // First example
  *      var t42 = R.thrush(42);
  *      t42(R.identity); //=> 42
  *      t42(R.add(1)); //=> 43
  *
- *      // with lenses
- *      var state = {
- *        firstName: "Bruce",
- *        lastName: "Wayne",
- *      };
- *      var tState = R.thrush(state);
- *      var lastName = lensProp('lastName');
+ *      // Second example
+ *      var url = { pathname: 'http://localhost/index.html', query: '?firstName=Bruce&lastName=Wayne' };
  *
- *      tState(R.view(lastName)); //=> "Wayne"
- *
- *      var overState = R.compose(tState, over);
- *      overState(lastName, R.toUpper); //=> {"firstName": "Hello", "lastName": "WORLD"}
+ *      R.thrush(url)(
+ *        R.compose(R.join(''), R.values)
+ *      ); //=> http://localhost/index.html?firstName=Bruce&lastName=Wayne
  */
 module.exports = _curry2(function thrush(x, f) { return f(x); });
--- a/src/thrush.js
+++ b/src/thrush.js
@@ -1,0 +1,39 @@
+var _curry2 = require('./internal/_curry2');
+
+/**
+ * When applied fully :
+ * - A function that take a value and a function,
+ *   and apply the function to the value.
+ *
+ * When applied partially :
+ * - A function that just wrap a value into a function.
+ *   It's the T combinator,
+ *   you can retreive this value with I combinator (also R.identity)
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig a -> (a -> b) -> b
+ * @param {*} x The value to wrap.
+ * @param {Function} f The function to apply.
+ * @return {*} Tun result of applying f on x.
+ * @example
+ *
+ *      var t42 = R.thrush(42);
+ *      t42(R.identity); //=> 42
+ *      t42(R.add(1)); //=> 43
+ *
+ *      // with lenses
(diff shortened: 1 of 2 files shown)


guillaumearm:
maybe we should wait R.T name be available according to the issue #1846

src/thrush.js
```diff
@@ -0,0 +1,39 @@
...
+ *        firstName: "Bruce",
+ *        lastName: "Wayne",
+ *      };
+ *      var tState = R.thrush(state);
+ *      var lastName = lensProp('lastName');
+ *
+ *      tState(R.view(lastName)); //=> "Wayne"
+ *
+ *      var overState = R.compose(tState, over);
+ *      overState(lastName, R.toUpper); //=> {"firstName": "Hello", "lastName": "WORLD"}
```
CrossEye: I think the lens examples are overkill here.  Someone wanting to understand `thrush` suddenly has to know how `lensProp`, `view`, and `over` all work.

CrossEye:
I like having the function, but as much as I like Smullyan, I don't agree with using the bird names for our functions.  We should have a more descriptive name.  It's too bad that `apply` is already taken.  I don't have a great one now.  `perform` is the best I have to offer.  But I really think it should be something other than `T` or `thrush`, although I would like notes about those alternative names.

guillaumearm:
I'm totally agree with you, i think `perform` is a good idea ;)
anyway, the exemple is overkill, i think i will rewrite one soon.

davidchambers:
I don't like the name `perform`. It sounds very side-effecty. We could instead:

  - remove `call` and `apply` and name the T combinator `apply`; or
  - name the T combinator `apply1`.

CrossEye:
> I don't like the name `perform`. It sounds very side-effecty. 

Hmm, I don't get that impression at all.  But if you do, probably others would as well.

> We could instead:
>   - remove `call` and `apply` and name the T combinator `apply`; or

While I have no love for `call` -- it seems very rarely useful -- I use `apply` all the time, converting external variadic functions into ones that accept a list instead.  Things like `minimum = R.apply(Math.min)`.

>   - name the T combinator `apply1`.

That sounds like a last resort.  If we can't find anything else that would suit, I would live with that, but we've managed to avoid that sort of wonkiness so far.  It would be nice to continue to do so.

guillaumearm:
About `perform`, I think everyone should now Ramda is free-side-effects, so the confusion doesn't really exist ?

and what about `transform` ?

guillaumearm:
simple example.
```
const transform = x => f => f(x)

const state = {
  firstName: 'Hello',
  lastName: 'World',
}

transform(state)(pipe(
  toPairs,
  flatten,
))
```

buzzdecafe:
`exec` anyone? too java-ish?

kedashoe:
> While I have no love for call -- it seems very rarely useful -- I use apply all the time, converting external variadic functions into ones that accept a list instead.

I also don't use call but use apply quite often. How about

apply -> spread
call -> deprecate/leave it alone
thrush -> apply

although there is also unapply.. unspread?

guillaumearm:
Sure, apply is a good name for T, but what is about backward compatibility ? will there be a Ramda 1.0 ?

buzzdecafe:
are we just stuck on the name? if we can't agree on a name but like the function, let's merge and change the name later

guillaumearm:
@buzzdecafe : I agree with you, but not `R.thrush` ?
and I will update the documentation example with the simple example. are you ok for this ?

CrossEye:
But bikeshedding is such an important part of our culture, @buzzdecafe!   Are you trying to spoil our fun?

guillaumearm:
I know ! 

why not `R.with` ?

```
with (41) (add(1)) //=> 42
```

buzzdecafe:
let's just merge as `thrush` FFS.

CrossEye:
I say merge as any of`perform`, `transform`, or `exec`, not `thrush`.

itaied246:
Any plans on merging it?

buzzdecafe:
i think the only hangup is the name. So let's see if we can get one more round of bikeshedding:

`applyTo` anyone?

If we can't come to some agreement, i'd be happy to merge as is

:cow2:

itaied246:
I find `pipeFirst` very informative and clear, 
but I would use anything just merge this PR.

kedashoe:
I like `applyTo` :+1:

buzzdecafe:
@CrossEye you bless `applyTo`? Let's move this one along

CrossEye:
`applyTo` sounds good to me.

kedashoe:
@guillaumearm looks like the bike is finally in the shed. Could you rebase and call the function `applyTo`?

haskellcamargo:
Yay! Updates on this?

Issue #1776, opened 2016-05-23
Proposal: show how higher order functions perform function applications using symbolic generics


thurt:
original ref #1725 -- I raise the question of whether there are standard terms for describing how a list of functions can be applied to a list of values? It is hard to express some kinds of function applications with just plain English.

Showing a symbolic output for the functions which are performing function applications could be another way to help explain behavior (if plain English is not enough). I am aware of `juxt`, `converge`, `ap`, `scan`, whose documentation could benefit from showing this.

See this [Mathematica](https://reference.wolfram.com/language/guide/FunctionalProgramming.html) page for inspiring examples of what I mean. 
As one example, [`FoldList`](https://reference.wolfram.com/language/ref/FoldList.html) is equivalent to Ramda's `scan`. Here its symbolic output expresses how it performs function applications to list values:

![](https://reference.wolfram.com/language/ref/Files/FoldList.en/I_2.gif)
![](https://reference.wolfram.com/language/ref/Files/FoldList.en/O_1.gif)

CrossEye:
While I'm loathe to add anything to our bloated comment structure,  (to my mind anything more than the sigs starts to feel bloated, so don't mind me) this seems like a good idea.  I'm sure there are others that would benefit as well... `compose`, `pipe`, `useWith` leap to mind immediately.

I would assume that we could handle this with simply another JSDoc tag and a tweak to the template.  @thurt, would you be interested in creating a PR?  This would actually involve a PR in the code to add the actual `@symb` (or whatever) tags, and a second one in the [docs project][dp] to show these in some reasonable way in the documentation.


  [dp]: https://github.com/ramda/ramda.github.io

thurt:
>  While I'm loathe to add anything to our bloated comment structure, (to my mind anything more than the sigs starts to feel bloated, so don't mind me)

To reduce bloat, it may be possible to reduce/eliminate some text explanations with these. I think their purpose is simply as a teaching tool; to show the internal behavior of the function applications and express how the function's name correlates with that.

I can look into getting a PR started more seriously over the weekend. Also interested in hearing any other suggestions/ideas on these topics .

benperez:
This is a really good idea. I'm trying to work through some examples. Let's take compose:
```
# current type signature, alphabet soup:
((y → z), (x → y), …, (o → p), ((a, b, …, n) → o)) → ((a, b, …, n) → z)
# let's make it a little more clear with this new notation
> R.compose(f, g, h)(a)
[...]

PR #1852, opened 2016-08-06
adds-symbols - initial commit

thurt:
original issue: https://github.com/ramda/ramda/issues/1776

Also, here is a single file with all symbols together for easier viewing/editing: https://github.com/thurt/ramda-analysis/blob/master/symbols.js

There will be a second pull request to ramda.github.io repo in order to display these to the web documentation. I will reference it when I make that PR. 

This PR is totally flexible so add suggestions for the changes you would like to see!

### DIFF
--- a/src/adjust.js
+++ b/src/adjust.js
@@ -24,6 +24,8 @@ var _curry3 = require('./internal/_curry3');
  *
  *      R.adjust(R.add(10), 1, [0, 1, 2]);     //=> [0, 11, 2]
  *      R.adjust(R.add(10))(1)([0, 1, 2]);     //=> [0, 11, 2]
+ * @symb R.adjust(f, -1, [a, b]) = [a, f(b)]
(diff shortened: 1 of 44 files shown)


buzzdecafe:
this is really nice. thanks :cow2:

CrossEye:
Very, very nice.  Thanks for all your hard work, and for putting up with my fiddling!

CrossEye:
There's some issue that prevents this from automatically merging.  I'm still triaging issues that have come up over my vacation.  If no one has looked into it before I'm done with that, I'll try to figure it out.

thurt:
I rebased this PR so you should be able to automatically merge now. The conflict was that `src/wrap.js` was deleted in a recent commit and I had a symbol added into that same file.

buzzdecafe:
:tada:

CrossEye:
This is great!  If we can find time, I'd love to publish a new point release just to include this on our site.

Thanks, @thurt!

PR #1858, opened 2016-08-09
deprecate R.unary and R.binary

davidchambers:
:warning: This may be controversial.

I posit that `R.unary(f)` is equally well expressed `x => f(x)`, and that `R.binary(g)` is equally well expressed `(x, y) => g(x, y)`.

Ramda contains several functions for "fixing" inconvenient or poorly designed functions. In most cases a humble wrapper function does the job equally well. When more than one "adjustment" is necessary, a wrapper function is much clearer. Imagine, for example, that we have a function `f` which takes three arguments, the last of which is optional. We'd like to derive from `f` a function, `g`, which takes two arguments and returns the result of applying `f` to these arguments _in reverse order_. Which of the following is clearer?

``` javascript
const g = R.flip(R.binary(f));
```

``` javascript
const g = (x, y) => f(y, x);
```

The `g` defined in the second case is not curried, of course, but this could be achieved by applying `R.curry` to the wrapper function if desired. Still, I find the second case easier to parse.

### DIFF
--- a/src/binary.js
+++ b/src/binary.js
@@ -15,6 +15,7 @@ var nAry = require('./nAry');
  * @param {Function} fn The function to wrap.
  * @return {Function} A new function wrapping `fn`. The new function is guaranteed to be of
  *         arity 2.
+ * @deprecated since v0.23.0
  * @example
  *
  *      var takesThreeArgs = function(a, b, c) {
--- a/src/unary.js
+++ b/src/unary.js
@@ -15,6 +15,7 @@ var nAry = require('./nAry');
  * @param {Function} fn The function to wrap.
  * @return {Function} A new function wrapping `fn`. The new function is guaranteed to be of
  *         arity 1.
+ * @deprecated since v0.23.0
  * @example
  *
  *      var takesTwoArgs = function(a, b) {


buzzdecafe:
lgtm
(need a new emoji)
:camel:

CrossEye:
I'm putting this one on my 'to think about' list.

davidchambers:
@CrossEye: ((:bell:))

CrossEye:
I have thought about it some, and it's the same answer for a number of similar issues.  I want to get a version 1.0 out the door that supports ES5.  (We currently almost support ES3, I think.)

After that, I'm quite interested in reducing the surface area in those places where ES6 makes some of our features less relevant.  However, I'm not sure that I would put these functions on the chopping block.

I think that 

```js
var g = function(x, y) {return f(y, x);}
```

is *less* readable than the current version, and I think that either `flip` or `binary` on their own are quite readable now.  And this:

```js
const g = (x, y) => f(x, y)
```
is much more obscure than this:

```js
const g = binary(f);
```

Looking at that first one, my initial question is "how is that different from `const g = f`?  Why bother?"  There's no such confusion with `binary`.

In other words, I'm not convinced.

davidchambers:
> In other words, I'm not convinced.

That's okay. I'm just going through <https://github.com/pulls> dealing with old pull requests one way or another. I'll close this one. :)

CrossEye:
I was planning on doing a bunch of clean-up myself this weekend, but it looks like that's going on hold again.  :disappointed:

PR #1861, opened 2016-08-11
Update binary predicate signatures as they're not curried.

mjgpy3:
This started with `differenceWith`, I noticed that things like
```
R.differenceWith(a => b => ..., xs, ys);
```
don't work as expected.

I tried to propagate the signature changes to all functions that use `internal/_containsWith`.

If updating the signatures is not the right solution, let me know, I'll work on something better (e.g. allowing predicates to be curried).

### DIFF
--- a/src/differenceWith.js
+++ b/src/differenceWith.js
@@ -11,7 +11,7 @@ var _curry3 = require('./internal/_curry3');
  * @memberOf R
  * @since v0.1.0
  * @category Relation
- * @sig (a -> a -> Boolean) -> [*] -> [*] -> [*]
+ * @sig ((a, a) -> Boolean) -> [a] -> [a] -> [a]
  * @param {Function} pred A predicate used to test whether two items are equal.
  * @param {Array} list1 The first list.
  * @param {Array} list2 The second list.
--- a/src/intersectionWith.js
+++ b/src/intersectionWith.js
@@ -12,7 +12,7 @@ var uniqWith = require('./uniqWith');
  * @memberOf R
  * @since v0.1.0
  * @category Relation
- * @sig (a -> a -> Boolean) -> [*] -> [*] -> [*]
+ * @sig ((a, a) -> Boolean) -> [a] -> [a] -> [a]
  * @param {Function} pred A predicate function that determines whether
  *        the two supplied elements are equal.
  * @param {Array} list1 One list of items to compare
--- a/src/symmetricDifferenceWith.js
+++ b/src/symmetricDifferenceWith.js
@@ -12,7 +12,7 @@ var differenceWith = require('./differenceWith');
  * @memberOf R
  * @since v0.19.0
  * @category Relation
- * @sig (a -> a -> Boolean) -> [a] -> [a] -> [a]
+ * @sig ((a, a) -> Boolean) -> [a] -> [a] -> [a]
  * @param {Function} pred A predicate used to test whether two items are equal.
  * @param {Array} list1 The first list.
  * @param {Array} list2 The second list.


src/differenceWith.js
```diff
@@ -11,7 +11,7 @@ var _curry3 = require('./internal/_curry3');
  * @memberOf R
  * @since v0.1.0
  * @category Relation
- * @sig (a -> a -> Boolean) -> [*] -> [*] -> [*]
+ * @sig (a, a -> Boolean) -> [a] -> [a] -> [a]
```
buzzdecafe: `((a, a) -> Boolean` would be clear. Otherwise it is ambiguous: it could mean a tuple of `a` and a function from `a` to `Boolean`

mjgpy3: That's a good point. Updating.

mjgpy3:
@buzzdecafe updated

src/differenceWith.js
```diff
@@ -11,7 +11,7 @@ var _curry3 = require('./internal/_curry3');
  * @memberOf R
  * @since v0.1.0
  * @category Relation
- * @sig (a -> a -> Boolean) -> [*] -> [*] -> [*]
+ * @sig ((a, a) -> Boolean) -> [a] -> [a] -> [a]
```
davidchambers: The signature should actually be `((a, b) -> Boolean) -> [a] -> [b] -> [a]`, since the two lists needn't be of the same type. For example:

```javascript
> R.differenceWith((r, id) => r.id === id, [{id: 1}, {id: 2}, {id: 3}], [2, 4, 6])
[{id: 1}, {id: 3}]
```

src/intersectionWith.js
```diff
@@ -12,7 +12,7 @@ var uniqWith = require('./uniqWith');
  * @memberOf R
  * @since v0.1.0
  * @category Relation
- * @sig (a -> a -> Boolean) -> [*] -> [*] -> [*]
+ * @sig ((a, a) -> Boolean) -> [a] -> [a] -> [a]
```
davidchambers: As above, this signature should be `((a, b) -> Boolean) -> [a] -> [b] -> [a]`.

mjgpy3: Alright, will fix, thank you!

buzzdecafe: interesting. I see this as a "set" operation, where the user is saying these two sets are the same type based on the predicate. I'm not gonna make a stand here, but i would expect an intersection of `[a]` and `[b]` to be `[ ]`.

src/differenceWith.js
buzzdecafe: as i mentioned below, i think i disagree. Viewed as a "set" relation, I would expect `A - B = A`  where `A` is a set of type `a` and `B` is a set of some other type `B`.

mjgpy3: @buzzdecafe that was my assumption as well when I made the signatures `... -> [a] -> [a] -> [a]`.

src/intersectionWith.js
mjgpy3: I suppose `((a, b) -> Boolean) -> [a] -> [b] -> [a]` is more in line with what the code actually does, correct? `((a, a) -> Boolean) -> [a] -> [a] -> [a]` is more inline with "ideal usage."

Maybe the appropriate question to ask is what's more useful for API consumers?

buzzdecafe: I think that is a fair characterization. I also don't wanna get bogged down on this, but I don't wanna just steamroll @davidchambers either. I prefer the "set" semantics, but if David has a strong argument in favor the "practical" side I wanna hear it.

davidchambers: I'd forgotten that #1597 introduced a regression which means that the more general signature proposed above no longer works reliably:

```javascript
> R.intersectionWith(R.prop('x'), [{x: 1, y: 1}], [{x: 1}])
[{x: 1, y: 1}]

> R.intersectionWith(R.prop('x'), [{x: 1}], [{x: 1, y: 1}])
[{x: 1}]
```

I'll open a pull request to address this. We introduced a regression. We should fix the regression *then* discuss the value of replacing a more general function with a less general one.

buzzdecafe: nice find, @davidchambers; we almost rehashed that thread. shall we merge this one or wait for yours?

davidchambers: Let's see whether #1868 is merged before deciding how to proceed.

davidchambers:
Given the debate in #1868, I think we should merge this pull request as it stands.

:deciduous_tree:

buzzdecafe:
agreed
🐪

mjgpy3:
Thank you!

PR #1864, opened 2016-08-12
xyz@1.0.x

davidchambers:
davidchambers/xyz#22


bower.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "ramda",
   "main": "dist/ramda.js",
-  "version": "0.22.1",
```
davidchambers: davidchambers/xyz#19

buzzdecafe:
:cow2:

buzzdecafe:
:dromedary_camel:

davidchambers:
Anyone who publishes Ramda should run `rm -r node_modules && npm install`. :)

buzzdecafe:
@davidchambers would it be overkill to add that to the publish flow?

davidchambers:
Something like this, do you mean?

```diff
 .PHONY: release-major release-minor release-patch
 release-major release-minor release-patch:
+	@rm -rf node_modules
+	@npm install
 	@$(XYZ) --increment $(@:release-%=%)
```

We could certainly do so, though publishing would become slightly less satisfying. I'm don't mind one way or the other.

buzzdecafe:
yeah that's what i mean. i like it because it's one less thing to remember.

davidchambers:
`xyz@1.0.0` is the first release since November 2014. It may well be another two years before the next release. So running `rm -r node_modules && npm install` is a once-every-year-or-two task rather than a run-before-publishing task. I would not even have mentioned it but for the change which means the `"version"` field in __bower.json__ is no longer updated. Running an older version of xyz would reintroduce the `"version"` field removed in this pull request.

buzzdecafe:
i see. nm then :-)

svozza:
Would `npm prune` achieve the same thing?

davidchambers:
> Would `npm prune` achieve the same thing?

I don't think so. Its [documentation][1] says nothing about updating dependencies.


[1]: https://docs.npmjs.com/cli/prune

svozza:
I meant in conjunction with `npm install `.

davidchambers:
I don't think `npm install` updates dependencies either. [`npm update`][1] does, but I've had problems with that command in the past. npm is a complex, bug-ridden piece of software. I rely on it as little as possible. ;)


[1]: https://docs.npmjs.com/cli/update

svozza:
Haha. Fair enough.

PR #1866, opened 2016-08-16
Tidy the curry'ness of doc examples

olsonpm:
- Remove unnecessary examples that add no value other than show curry'ness
- Change ')(' -> ', ' where the parentheses are unnecessary.  It causes confusion because
  there are examples where the parentheses are necessary (e.g. invoker)

### DIFF
--- a/src/uniqWith.js
+++ b/src/uniqWith.js
@@ -19,10 +19,10 @@ var _curry2 = require('./internal/_curry2');
  * @example
  *
  *      var strEq = R.eqBy(String);
- *      R.uniqWith(strEq)([1, '1', 2, 1]); //=> [1, 2]
- *      R.uniqWith(strEq)([{}, {}]);       //=> [{}]
- *      R.uniqWith(strEq)([1, '1', 1]);    //=> [1]
- *      R.uniqWith(strEq)(['1', 1, 1]);    //=> ['1']
+ *      R.uniqWith(strEq, [1, '1', 2, 1]); //=> [1, 2]
+ *      R.uniqWith(strEq, [{}, {}]);       //=> [{}]
+ *      R.uniqWith(strEq, [1, '1', 1]);    //=> [1]
+ *      R.uniqWith(strEq, ['1', 1, 1]);    //=> ['1']
  */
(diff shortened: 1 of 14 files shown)


olsonpm:
All changes were tested locally.  I am also confident I didn't miss any - I searched for text matching ')(' on the docs page.

CrossEye:
I'd like to think about this a little.  This change went further than I expected.  I was imagining that you would do the ones like `adjust` and `useWith` that seemed to redundantly show how to use the function in a curried and a non-curried manner.

For the remainder, where it's a judgement call.

buzzdecafe:
why did you *close* this, @CrossEye? let's get it into mergeable shape instead

olsonpm:
I definitely explained my reasoning for why I removed the curryness of the others.  I'd like to hear your reasoning for keeping it.

src/uniqWith.js
```diff
@@ -19,10 +19,10 @@ var _curry2 = require('./internal/_curry2');
...
  *
  *      var strEq = R.eqBy(String);
- *      R.uniqWith(strEq)([1, '1', 2, 1]); //=> [1, 2]
- *      R.uniqWith(strEq)([{}, {}]);       //=> [{}]
- *      R.uniqWith(strEq)([1, '1', 1]);    //=> [1]
- *      R.uniqWith(strEq)(['1', 1, 1]);    //=> ['1']
+ *      R.uniqWith(strEq, [1, '1', 2, 1]); //=> [1, 2]
+ *      R.uniqWith(strEq, [{}, {}]);       //=> [{}]
+ *      R.uniqWith(strEq, [1, '1', 1]);    //=> [1]
+ *      R.uniqWith(strEq, ['1', 1, 1]);    //=> ['1']
```
buzzdecafe: in a case like this, it might make sense to define `const uniqByStr = R.uniqWith(R.eqBy(String))` and use that in the examples

olsonpm: Ah - good point.  I'll await further clarifications before modifying the branch

CrossEye:
While the idea may be sound, it's so far behind that we should start over rather than try to fix this PR.

PR #1867, opened 2016-08-17
Add mergeDeep, mergeDeepWith & mergeDeepWithKey

scott-christopher:
Continuing on from #1088, this change introduces deep (recursive) merge equivalents for the existing `merge`, `mergeWith` and `mergeWithKey` functions.


buzzdecafe:
this lgtm -- even handles circular references gracefully, which i imagine falls out of the `mergeWithKey` implementation.
:dromedary_camel:

expelledboy:
Would it be possible to implement `mergeDeepEvolve` as per [this comment](https://github.com/ramda/ramda/pull/1088#issuecomment-237044481) using `mergeDeepWithKey`?

I can't see how, given that merging with the key does not handled duplicate keys anywhere in the object. Not that its an argument for changing the implementation, just curious.

scott-christopher:
> Would it be possible to implement mergeDeepEvolve [...]?

It can be effectively derived from `reduce` and `evolve` already, though I'm not entirely sure there would be enough demand for it to exist as a specialised function in the core library though.

```js
const mergeDeepEvolve = (fnObj, aObj, bObj) => reduce(evolve, fnObj, [aObj, bObj])
mergeDeepEvolve({ a: { b: concat } }, { a: { b: [1], c: true }, d: true }, { a: { b: [2] }, e: true })
//=> { a: { b: [ 1, 2 ] }, e: true }
```

Speaking of `evolve`, I suspect it could now be derived from `mergeDeepWith(call)`, though it would depend on curried functions and some of the edge cases may potentially produce different results so it's probably best to leave it as it is.

expelledboy:
@scott-christopher That would be `mergeEvolve` :) I failed to implement a recursive version in less than 30 lines of code. Still getting used to the library though..

scott-christopher:
@expelledboy I also realised there is a difference between `reduce(evolve)` and `reduce(mergeDeepWith(call))`, such that folding over `evolve` only keeps keys that appear in both objects or the last:

```js
const mergeEvolve = (fnObj, lObj, rObj) => reduce(mergeDeepWith(call), fnObj, [lObj, rObj])
mergeEvolve({ a: { b: concat } }, { a: { b: [1], c: true }, d: true }, { a: { b: [2] }, e: true })
//=> { a: { b: [ 1, 2 ], c: true }, d: true, e: true }
```
```js
const mergeEvolve_ = (fnObj, lObj, rObj) => reduce(evolve, fnObj, [lObj, rObj])
mergeEvolve_({ a: { b: concat } }, { a: { b: [1], c: true }, d: true }, { a: { b: [2] }, e: true })
//=> { a: { b: [ 1, 2 ] }, e: true }
```

CrossEye:
:herb:

@paldepind: Any objections to this version?

src/mergeDeep.js
```diff
@@ -0,0 +1,29 @@
...
+ * @memberOf R
+ * @category Object
+ * @sig {a} -> {a} -> {a}
+ * @param {Object} lObj
+ * @param {Object} rObj
+ * @return {Object}
+ * @see R.merge, R.mergeDeepWith, R.mergeDeepWithKey
+ * @example
+ *
+ *      R.merge({ name: 'fred', age: 10, contact: { email: 'moo@example.com' }},
```
kedashoe: R.mergeDeep?

src/mergeDeepWithKey.js
```diff
@@ -0,0 +1,41 @@
...
+ * @sig (String -> a -> a -> a) -> {a} -> {a} -> {a}
+ * @param {Function} fn
+ * @param {Object} lObj
+ * @param {Object} rObj
+ * @return {Object}
+ * @see R.mergeWithKey, R.mergeDeep, R.mergeDeepWith
+ * @example
+ *
+ *      let concatValues = (k, l, r) => k == 'values' ? R.concat(l, r) : r
+ *      R.mergeWithKey(concatValues,
```
kedashoe: R.mergeDeepWithKey?

kedashoe:
:mans_shoe:

paldepind:
This is really great :smile:

I do however think that the order of the arguments should be reversed as in #1088. The current order is inconsistent with the rest of Ramda where the thing that is being modified is the _last_ argument. With the current `merge` and these deep variants that is not the case. I.e. if I want to set two properties on `o` i have to do `R.merge(o, {foo: 1, bar: 2})`.

src/mergeDeep.js
scott-christopher: Yep, thanks. :zap:

src/mergeDeepWithKey.js
scott-christopher: :zap:

scott-christopher:
I've previously thought of the arguments of `merge` as both being the data, but you make a valid point @paldepind (and it's good to see you round these parts again 😄 )

How do others feel about swapping the argument order of `merge` and `mergeDeep` (the order of the other variants is determined by the function merging duplicate keys)?

My only concern would be that the breaking change could introduce some interesting bugs for anyone that didn't realise the argument order had changed, but we're a bit limited in what we can do to help avoid it. Perhaps we could echo breaking changes as an npm `postinstall` script?

paldepind:
> I've previously thought of the arguments of merge as both being the data, but you make a valid point @paldepind (and it's good to see you round these parts again :smile: )

It's good to be around :smile:

I think that even if we consider both to be data one of the objects is still the data that is applied on top of the other objects data. If that makes sense. But to me it is most useful to think of `R.merge` as a functional version of `Object.assign`.

> My only concern would be that the breaking change could introduce some interesting bugs for anyone that didn't realise the argument order had changed, but we're a bit limited in what we can do to help avoid it. Perhaps we could echo breaking changes as an npm `postinstall` script?

Yes. You are right. That is an important point. The change can be quite subtle. That was actually the biggest reason for the renaming in #1088. A new name means that people will have to notice the breaking change.

davidchambers:
> How do others feel about swapping the argument order of `merge` and `mergeDeep` […] ?

:+1: when I think of the partially applied forms.

:-1: when I think of the fully applied forms.

For me this has the same problem as `R.subtract`. I want `R.subtract(10)` to mean `x => x - 10` but I want `R.subtract(10, 2)` to mean `10 - 2`.

Having `b` "trump" `a` in `R.merge(a, b)` feels right to me.

expelledboy:
@davidchambers I haven't use ramda enough to make a decision based on experience, however `merge` is recursive equivalent to `Object.assign`?

Additionally you know less about arg 2 then 1, so I feel is should trump arg 1.

Besides you can just `flip(merge)(obj)`, so the question is really which version will be most used?

davidchambers:
> Besides you can just `flip(merge)(obj)`, so the question is really which version will be most used?

I don't think that's accurate. `R.subtract(R.__, y)(x)` is used more often than `R.subtract(x)(y)`, but we've chosen to stick with the less useful argument order because `R.subtract(x, y)` meaning `y - x` is too surprising.

expelledboy:
Oh wow, never used `subtract`, but having used `inc` how `subtract` is implemented currently is surprising :D 

If I think about it I cant actually ever imagine using it without reversing the arguments...

davidchambers:
> If I think about it I cant actually ever imagine using it without reversing the arguments...

We've been grappling this issue for quite a while. See #1497, for example.

Bradcomp:
For what it's worth, I like the current ordering of `merge`.  It matches the order of `Object.assign` (not that important), but it also makes sense in the context of `mergeAll` being a `reduce` of `merge`.

From a practical standpoint, one of the more common use cases I have for `merge` is that of providing default values to objects.  It's nice to be able to just `merge(defaults)` and then pass in my object(s).

buzzdecafe:
i'm for merging as is and punting on the infix/prefix problem. any strong objections?

paldepind:
> any strong objections?

Yes :smile: I think we should get things right _before_ we introduce new functions to our API.

If we look at how Ramda was conceived, why it was created in a world that already contained Underscore and Lodash, it seems that there was three distinguishing principles/features:

1. Functions should be pure. Ramda satisfies this goal for _all_ functions.
2. Functions should be curried. Ramda satisfies this for _all_ functions.
3. Arguments to functions should be ordered such that the data to modify is the last argument—because this is most useful for currying. Ramda satisfies this for almost all functions.

I get the above ideas from "Introducing Ramda" and the other blog posts that we link to on the frontpage of the website.

Note that the last of these three principles is sometimes expressed as "data should be last". This is however, a less useful way of saying it as in functional programming everything is data. A function is also a form of data. Additionally functions such as `R.assoc` don't take any functions at all but yet its current argument order seems right. That is because the principle really is more like: The data that is the base for any modifications should be last. This explains both functions like `R.map` and `R.assoc`.

The current `R.merge` breaks this core principle. This is bad for two reasons:

1. It decreases consistency. If I want to set a property on an object with `R.assoc` I do `R.assoc('foo', 1, obj)`. But if I want to set a property with `R.merge` I'll have to do `R.merge(obj, {foo: 1})`. There is no symmetry because `R.merge` breaks the third principle. If I want to set a single property on an object I may use `R.assoc` but if I later want to set two properties I might want to switch my code to using `R.merge`. In that case I will have to be aware of the fact that `R.merge` is inconsistent with the rest of Ramda and I will have to reverse the arguments as part of my refactor.
2. We miss out on the benefits that the third principle gives. `R.merge` is less useful for currying. @Bradcomp already gave a great example on this.

If the above is not convincing enough lets consider a scenario where I am implementing my own map data structure. I call it `PureMap` as it's supposed to be a pure version of JavaScript's `Map`. Because a map is a monoid and I want to support fantasy-land I implement a `concat` method on it. The `concat` method works like this:

```js
const map1 = PureMap.fromObject({foo: 1, bar: 1});
const map2 = PureMap.fromObject({foo: 2, baz: 3});
map1.concat(map2); //=> PureMap.fromObject({foo: 1, foo: 2, baz: 3})
```

In the above example `map2` overrides `map1` when their keys intersect. This is what one would expect from a OO `concat` method. For instance this is what Immutable.js does on their similair `merge` method. If we use this monoid with `R.concat` we get:
```js
const map1 = PureMap.fromObject({foo: 1, bar: 1});
const map2 = PureMap.fromObject({foo: 2, baz: 3});
R.concat(map2, map1); //=> PureMap.fromObject({foo: 1, foo: 2, baz: 3})
```

So now we see that when using `R.concat` with a sensible Map monoid we arrive at the opposite argument order of `R.merge`. This again is because `R.concat` respects the third principle but `R.merge` violates it.

I think it would be a sad to introduce two new functions to Ramda that increases the currently very small number of functions that doesn't take the structure to modify as their last argument and thereby increase inconsistency in the library. I also think that this would be a great occasion to finally fix `R.merge`.

buzzdecafe:
I'm not convinced there is a "correct" argument order here. As @davidchambers succinctly put it:

>:+1: when I think of the partially applied forms.
>
>:-1: when I think of the fully applied forms.

This looks to me like the "infix-to-prefix for non-commutative functions" problem one more time. We have a number of these in the lib. As was also pointed out above, we tend to favor the fully-applied form. 

I think the guiding principle you left out of your list is *simplicity*. A deep recursive merge function is a stretch on that score. On those grounds, I'd be happy to punt on this completely, and leave it out of the lib. But I would like to come to a decision, one way or the other: (1) include as-is; (2) swap args and include; (3) pass. If we cannot agree on 1 or 2, then 3 should win by default.

scott-christopher:
Whether the `merge` operation is left or right biased has no impact on whether it forms a valid monoid instance, it is just a choice of the implementation. Both arguments here are _the data_. It is the use case that determines here whether the choice of a left or right bias is more useful when partially applied. The left-bias is useful to act as a "setter", while the right-bias offers the notion of an initial "default" object that can be extended or overridden.

If a left-biased `merge` is desired, it can still be implemented simply via `R.mergeWith(S.K)`.

paldepind:
@scott-christopher 

> Whether the `merge` operation is left or right biased has no impact on whether it forms a valid monoid instance, it is just a choice of the implementation.

Indeed, yes. Maybe I didn't make my argument clear enough. I am saying is that if you implement a Map monoid and follow the typical convention of OOP methods then you end up with a `R.concat` that works opposite to `R.merge`. You can do it the other way and still have a monoid, but that does not affect my argument.

> Both arguments here are _the data_.

As I mentioned I consider _everything_ to be data. I once saw a splendid video with Brian Beckman in which he made this point. He said that a function, for instance, is isomorphic to some look-up table (potentially infinitely large and/or dense). So a function is just data. That is why we can pass it around like any other value. That is my current view and therefore I don't understand your argument here. What is your notion of data? Are you saying that since both arguments to `R.merge` are data we can arrange them in any way we prefer? What about `R.assoc`? How do you explain it's argument order? Aren't all of it's arguments data as well?

What matters to me is what we do with the data. For instance all the arguments to `R.assoc` are data but the first two describes a modification that is applied to the last argument. Similarly the first argument to `R.map` is data that describes a modification that should be made to the elements in the second argument. To me this explains the ordering of their arguments since the data that is the basis of modifications should be last and descriptions of modifications should be first.

If we consider `R.merge` then one of the arguments is applied on top of the other one. They are not equal since one overwrites the other. Thus one of them describes a number of changes and the other what the changes should be based upon. This is similar to `R.assoc` and `R.map` where the first arguments describe a modification and the last argument what to modify. Therefore I think that the object `R.merge` puts at the bottom should be the last one.

kedashoe:
Another (less important) reason to have the object being modified as the last argument for `merge` is that we can use the word "default" (eg `defaults` or `mergeDefaults` or etc) for the other order which should be pretty good for discoverability and inuition about what it does just from the name.

scott-christopher:
To clarify, by both arguments being considered _the data_ I was referring to your point 3 and the fact that this is a binary operation over identical types where either argument could produce a useful function if partially applied.

An example of a use case for keeping the order as it is in this PR is that it allows me to partially apply `mergeDeep` with a state object within a redux-like reducer function and pass that partially applied function off to other code to update their bits of the state.

The emphasis here being that both orders have valid use-cases, which is why I'm a little hesitant to introduce a breaking change to just switch use cases.

If we reach an impasse here, perhaps an alternative solution would be to deprecate the existing `merge` function, replacing it with an explicit `mergeRight` function and introduce `mergeLeft` as the left-biased alternative.

kedashoe:
> The emphasis here being that both orders have valid use-cases, which is why I'm a little hesitant to introduce a breaking change to just switch use cases.
> 
> If we reach an impasse here, perhaps an alternative solution would be to deprecate the existing merge function, replacing it with an explicit mergeRight function and introduce mergeLeft as the left-biased alternative.

That all seems quite reasonable. I think I still prefer breaking `merge` and adding `mergeLeft`, but I'm okay either way.

expelledboy:
But think about it `deepMergeRightWithKey` 😶  naming conventions aside if an agreement cant be reached I would be cool with explicit implementations. Certainly at a later point when consensus is reached surely you can alias or depreciate accordingly?

expelledboy:
I certainly dont want to entertain the idea this not get included because there is differing opinions on how best to use it. There are two very powerful use-cases 1) applying defaults 2) mutating state. What modern project these days doesnt include atleast one of those?

scott-christopher:
> But think about it deepMergeRightWithKey 😶

We'd only need to be explicit with the naming of `mergeLeft` and `mergeRight`, as the others make use of the provided function to decide whether it is left or right biased (or something else entirely).

davidchambers:
Having both `mergeLeft` and `mergeRight` seems like a good idea to me.

The relationship between `merge` and `concat` is unclear to me. Were we to treat objects as semigroups we could remove `R.merge`, but we'd still need to decide on the associativity: would `R.concat({x: 1}, {x: 2})` give `{x: 1}` or `{x: 2}`?

buzzdecafe:
maybe just `merge` and `mergeRight`, consistent with `reduce`/`reduceRight`, `partial`/`partialRight`, `mapAccum`/`mapAccumRight`

CrossEye:
@paldepind 

> Arguments to functions should be ordered such that the data to modify is the last argument—because this is most useful for currying. Ramda satisfies this for almost all functions.

My take is slightly different.  Parameters should be ordered so that those more likely to change come after those less likely to change -- again because this is most useful for currying.  This mostly overlaps with your suggestion, but in ones like this, it's a little harder.  My most common use-case for `merge`, as others have suggested, is for something like

```js
const options = merge(defaults, opts);
```

which might well lead to currying in the current order, and is probably responsible for the choice of the current order (from the days it was called `mixin`.)  I also like the way this version works with a fold.

I do like @buzzdecafe's `merge` / `mergeRight`, though.

scott-christopher:
The combination of `merge` and `mergeRight` does have a nice synergy with the other functions.

The only concern I have with adopting `merge`/`mergeRight` would be that we would have to flip the existing behaviour of `merge` from right-biased to left-biased, which I imagine would increase the likelihood of introducing subtle bugs into user's applications if they were unaware of the change while bumping the version of Ramda.

Renaming `merge` to `mergeRight` and introducing `mergeLeft` would at least allow for a deprecation cycle and subsequent removal, where discovering that `R.merge` is now `undefined` would at least be somewhat clearer for a user who was unaware of the change to discover the steps necessary to rectify.

buzzdecafe:
OK let's phase in `merge`/`mergeRight` like so:

0.22 (now): deprecate `merge`
0.23: add `mergeRight` and `mergeLeft`
0.24: deprecate `mergeLeft`; add `merge` (fka `mergeLeft`)

paldepind:
@scott-christopher 

> An example of a use case for keeping the order as it is in this PR is that it allows me to partially apply `mergeDeep` with a state object within a redux-like reducer function and pass that partially applied function off to other code to update their bits of the state.

On the other hand redux makes very heavy use of `Object.assign` to set properties on objects. And as we agree the left-biased merge it more useful as a setter.

@CrossEye 

> My take is slightly different.  Parameters should be ordered so that those more likely to change come after those less likely to change -- again because this is most useful for currying.  This mostly overlaps with your suggestion, but in ones like this, it's a little harder.

That is a great way to look at it as well! My only counter argument would be that then one will have to decide the order for each function on a case by case basis. My explanation ensures consistency across the functions to which the reasoning applies.

I think that naming the left-biased version `R.merge` and the right-biased version `R.defaults` would be a fine idea as well.

davidchambers:
@scott-christopher makes a good point: flipping the associativity of `R.merge` would lead to many subtle bugs at upgrade time.

I don't buy the argument that `merge` and `mergeRight` would match `partial` and `partialRight`. Partially application is assumed to occur from the leftmost argument, so readers who see `partial` will always make the right assumption. This thread makes it clear that the same is not true of `merge`: regardless of how we have it behave some readers will assume it has the other behaviour.

I favour:

1.  __v0.23.0:__ Add `mergeLeft` and `mergeRight`; deprecate `merge`.
2.  __v0.24.0:__ Remove `merge`.

paldepind:
I think `mergeLeft` and `mergeRight` are confusing. What does the following do?

```js
mergeLeft({foo: 1}, {foo: 2});
```

My intuition would be that "left" means that it merges from right _to left_, i.e. that the result would be `{foo: 2}`. But per the [previous description](https://github.com/ramda/ramda/pull/1867#issuecomment-242240252) it means that it biases the left object, i.e. that the result will be `{foo: 1}`.

Maybe it's just me, but it feels counter intuitive.

davidchambers:
> My intuition would be that "left" means that it merges from right _to left_, i.e. that the result would be `{foo: 2}`.

Great point! This is a completely reasonable assumption.

Naming things is challenging!

expelledboy:
Perhaps right and left are the wrong words?

- `mergeFirst` `mergeLast`
- `mergeHead` `mergeTail`
- `mergeFrom` `mergeTo`
- `merge` `default`
- `merge` `mergeWith` ~ probably not an option but just throwing things out there
- `merge` `mergeInto` ~ makes the most sense to me
- `mergeIn` `mergeOut`

buzzdecafe:
I agree with @expelledboy 's point that the name itself may be misleading us. I could imagine going even further, e.g.

`override` for `({x: 1}, {x: 2}) -> {x: 2}`
`???` for `({x: 1}, {x: 2}) -> {x: 1}`

CrossEye:
> I agree with @expelledboy 's point that the name itself may be misleading us. I could imagine going even further

Anyone want to bring back `mixin`?  :smile:

expelledboy:
To be honest `merge` sounds simple, whereas `mixin` is probably something I would have to google because it sounds like it contains some voodoo. Just saying 🍠 

I like the sound of `merge` and `mergeTo`. And `mergeDeepTo` almost says exactly what it does, merges recursively to the object I pass it.

``` javascript
let state = {applied: false};
const mutator = mergeTo({applied: true});
mutator(state);
=> {applied: true}

let defaults = merge({applied: false});
const options = {applied: true}
defaults(options);
=> {applied: true}
```

CrossEye:
@expelledboy: To me those would be the best names we could have, **except** for the confusion caused by reversing the current meaning of `merge`.  I don't know if it's worth it, but it might be.

kedashoe:
I guess nobody has had any nomenclature epiphanies over the last two weeks..

I like the plan @scott-christopher / @davidchambers laid out for using `mergeLeft` and `mergeRight` and deprecating `merge`. Agreed that changing what `merge` does is too tricky.

As for what those names mean, my intuition lines up with what @paldepind specified [here](https://github.com/ramda/ramda/pull/1867#issuecomment-242686843), ie `mergeLeft` is merge to the left, so I would vote for that. *But*, if the majority of people watching this issue disagree and think the other way is better, then lets just go with that. I don't think it hugely matters re: left vs right, clearly whatever way we choose would be well documented, and if you search the ramda documentation for `merge` in this world, you would see both `mergeLeft` and `mergeRight` in the results, go on to read the docs, see the examples..

CrossEye:
Well... there was [a sugggestion][dc] that we might be able to forgo `merge` altogether and fold it into `concat`.  Just one more idea to toss around.

  [dc]: https://github.com/ramda/ramda/issues/1897#issuecomment-246200546

scott-christopher:
> @kedashoe: As for what those names mean, my intuition lines up with what @paldepind specified here, ie mergeLeft is merge to the left, so I would vote for that.

Even if I think about it merging from right to left, it doesn't necessarily indicate whether the keys of the left object will be preserved or overridden. We could consider something like `mergeOverL` and `mergeOverR` to help indicate which object will have its keys replaced, but I'm beginning to think that we won't be able to come up with a succinct name that clearly describes what either means to everyone without further assistance from some docs.

> @CrossEye: Well... there was a sugggestion that we might be able to forgo merge altogether and fold it into concat.

We'd lose the ability to help identify the left/right bias by the function name alone, but it would make it possible for us to leave the existing behaviour of `merge` as-is and offer the alternative via `concat`. We'd just need to be quite clear of the behaviours in the docs and identifying the alternatives exist.

expelledboy:
So no one liked my idea of `mergeTo` :(

buzzdecafe:
@expelledboy it can get pretty rough in here sometimes :-)

kedashoe:
How about `sinister` and `adroit` where `sinister` deletes javascript and installs haskell and `adroit` reads the programmer's mind and always merges the way they want :tada:

All seriousness aside, what I was trying to say with all [this](https://github.com/ramda/ramda/pull/1867#issuecomment-246259854) is that I think with `mergeLeft` and `mergeRight`, we have two fine interpretations, so we needn't go in a different direction (or do nothing), because we can explain very clearly what we mean. For example, `mergeLeft` as left biased is *not* how I naturally interpret `mergeLeft`, but because of that word "biased" and its yet-to-be-written accompanying explanation, it would be very easy for me to forever more remember when I need to use a merge what it is going to do. Similarly, I like @scott-christopher's suggestion of the use of the word "over", and I would think that if we go with `mergeLeft` as "merge over the things on the left", users would be able to remember that.

Well.. now that that's settled, the only thing left to decide is when we merge this pull request, is it a `mergeLeft` or a `mergeRight`??

buzzdecafe:
> `adroit` reads the programmer's mind and always merges the way they want 

What does the sig look like?

`merge :: Object -> Mind -> Object -> Object`

kedashoe:
> What does the sig look like?
> 
> merge :: Object -> Mind -> Object -> Object

haha maybe `Mind` should be last as it is most likely to change :thought_balloon:

joaomilho:
I object to including `adroit` in the current proposal because this function would be impure. To solve that, just wrap Mind into a timestamped thingy, like `PlanckTimestamp Mind` so you can always check the expected merged value at any point in time, including, of course, the future.

CrossEye:
Wait, I know @buzzdecafe, and his mind is always pure!

expelledboy:
I vote we leave this as is and merge it now!! 🉑

ScottFreeCode:
Random passerby putting in a vote for `default` as I think that expresses intent excellently.

Not sure what to use for the other direction instead of `merge` to avoid soft errors... Maybe something that suggests it's a form of `evolve` that uses an object of values instead of a tree of transformation functions? Or maybe `assocAll` because it's like `assoc` for multiple properties?

expelledboy:
`mergeDeepRight` and `mergeDeepLeft`  are explicit but lengthy.. is it not possible to just alias them to `override` and `default` and be done with it :-)

Wish Github allowed project to host polls, because this is being blocked by personal preference, rather than functionality, and thats not something which can be resolved by discussion.

Alternatively we can open a new issue to discuss a wiki page regarding naming conventions, parameter order, etc. There seems to be consensus in many respects, hashed out in various PRs I have read through, but its more of a general understanding amongst contributors then a spec to be followed in these situations.

OR 😈  we could add the ability to dictate "Porcelain" wrappers, and and then this would just be the "Plumbing" ela Git, but configurable 🙈  haha jk I just need this again in another project... Perhaps we could add a new staging area for functions whose API is yet to be determined, but the functionality is well defined?

```js
const R = require('ramda')({ staged: true });
R.mergeDeepRight(...)
```

davidchambers:
> Wish Github allowed project to host polls

It does, if one squints: #1958.

joaomilho:
Isn't it time this reaches master? 😄

scott-christopher:
I've rebased against `master` to get it up to date. If we agree that the differing assumptions of what `left` and `right` means can be resolved in documentation, is there anything else missing from this PR?

CrossEye:
LVGTM.

Any other thoughts before we merge?

kedashoe:
So our current plain old `merge` is "right biased", 2nd object wins on collisions? I think we should add `mergeLeft` to this PR for a left biased simple merge.. completes the set and makes it more clear from the names alone what `merge` does.

expelledboy:
Maybe add `mergeRight` as well, and then alias the old `merge`?

scott-christopher:
I'm happy to include changes to `merge` / `mergeLeft` / `mergeRight` now, though I'd prefer we open a separate PR to capture those changes.

expelledboy:
When is the next release? :)

CrossEye:
We don't usually schedule releases, just accumulate enough changes and say, "It's time."  Perhaps we can do so soon.

PR #1868, opened 2016-08-19
add R.innerJoin and deprecate R.intersectionWith

davidchambers:
#1597 introduced a regression. Prior to that change, the _actual_ type of the function was:

``` haskell
intersectionWith :: ((a, b) -> Boolean) -> [a] -> [b] -> [a]
```

Post optimization, the order in which arguments are provided to predicate is determined by the lengths of the input lists rather than the order in which the lists are provided to `intersectionWith`. This means that the function only behaves _reliably_ when treated as:

``` haskell
intersectionWith :: ((a, a) -> Boolean) -> [a] -> [a] -> [a]
```

This makes the function much less useful: if `a` has sensible equality semantics we can simply use `intersection`. The example in the [documentation](http://ramdajs.com/docs/#intersectionWith) is better expressed in terms of `intersection`. :stuck_out_tongue_winking_eye:

Let's set aside the question of the usefulness of the old behaviour, though. There's a more important matter to discuss, which is our process for handling regressions. Our current process seems to be:
1.  Regression.
2.  Discuss the merits of the regression.
3.  Fix the regression, perhaps.

This is not right. All else being equal we rightly favour a function's existing behaviour. The onus is therefore on the contributor proposing a change in behaviour to clearly demonstrate the benefits. Our current process means that when we unintentionally change the behaviour of a function the new behaviour becomes the incumbent and one must make a case for reinstating the old behaviour.

I'd like us to follow this process instead:
1.  Regression.
2.  Fix the regression.
3.  Discuss the merits of the regression, if applicable.

After discussing the merits of a regression, we may decide we actually stumbled upon a better function. We can then make a pull request to formalize the change, making any necessary documentation changes. This will also make it clear to others and to our future selves that the change was intentional, and provide a document to which the upgrade guide can link.


test/intersectionWith.js
```diff
@@ -10,4 +10,16 @@ describe('intersectionWith', function() {
     eq(R.intersectionWith(eqA, Ro, So), [{a: 3}, {a: 4}]);
   });
 
+  it('only returns elements from the first list', function() {
```
buzzdecafe: if this is the contract, then the name *must* change. `innerJoin` is a better fit for this behavior.

buzzdecafe:
>  The example in the documentation is better expressed in terms of intersection. :stuck_out_tongue_winking_eye:

i agree, a better example would show the revised semantics you are proposing

test/intersectionWith.js
davidchambers: Okay. Shall we deprecate `intersectionWith` (and leave its behaviour as it is now) and define `innerJoin`?

buzzdecafe: sounds good to me

davidchambers: :zap:

test/innerJoin.js
```diff
@@ -0,0 +1,25 @@
...
+  it('only returns elements from the first list', function() {
+    eq(f([a, b, c], []), []);
+    eq(f([a, b, c], [1]), [a]);
+    eq(f([a, b, c], [1, 2]), [a, b]);
+    eq(f([a, b, c], [1, 2, 3]), [a, b, c]);
+    eq(f([a, b, c], [1, 2, 3, 4]), [a, b, c]);
+  });
+
+  it('preserves the ordering of the first list', function() {
+    eq(f([c, a, b], [1, 2, 3]), [c, a, b]);
```
buzzdecafe: do we need to guarantee this? or does this just fall out of the implementation?

davidchambers: If we're to use an ordered collection to represent sets I think it's important to state how order is determined. Since we're filtering the first array, preserving the order of the first array seems sensible.

buzzdecafe: that sounds to me like "it falls out of the implementation." we're using an ordered collection because it is convenient. adding a test implies that's all we will ever use

davidchambers: > adding a test implies that's all we will ever use

I don't think so. Tests document and verify the *current* intended behaviour. Were every test case a lifelong commitment we'd be well advised to write no tests. ;)

buzzdecafe: > Tests document and verify the current intended behaviour. 

"intended" is what i am wondering about. is this the *intent* or merely an artifact of the implementation? We seem to agree it is the latter; in that case I don't think this test is warranted. i doubt it's worth fighting for though; should we ever broaden the definition of `innerJoin` we can simply remove the test

davidchambers: It's very much intentional. We're using an ordered collection. Order is significant, since Ramda does not consider `[1, 2]` equal to `[2, 1]`. I can think of three possible rules:

  - sort `xs'` based on the order of `xs`;
  - sort `xs'` based on the order of `ys`; or
  - apply our own sort order, perhaps the default sorting function used by `Array#sort`.

The other option is to shuffle `xs'` before returning it. This would prevent users from relying on a particular ordering. It would also break referential transparency (and possibly hurt performance), so I don't think it's a reasonable option.

Sorting `xs'` based on the order of `xs` seems best to me. Users should be able to rely on this until we change our minds and make a breaking API change (which hopefully won't happen).

buzzdecafe: > We're using an ordered collection. 

but *why* is that? is that fact *intrinsic* to this function? or just we have at hand? if the *intent* of this function is *only ordered lists* then this test makes sense. If the intent is relations generally, but for now all we have to implement it is Arrays, then this test may be unnecessarily prescriptive. That's all. Let's move on.

buzzdecafe:
looks good to me -- 🐪

test/innerJoin.js
davidchambers: We could wait until Ramda provides a Set type to introduce `innerJoin`. I'm in favour of this, actually.

buzzdecafe: i think we can proceed as is. When we have a `Set` we should support that

davidchambers: Works for me. :)

Bradcomp:
Shouldn't an `innerJoin` function allow the user to provide a joining function in addition to the equality check function?

buzzdecafe:
@Bradcomp i don't follow. How would a "joining function" differ from the `(a, b) -> Boolean` that is passed in the first argument? My understanding is that function specifies whether an item should be included in the "joined" result.

Bradcomp:
So for me, an inner join consists of both an equality check 
`(a -> b -> Boolean)` 
and a joining function 
`(a -> b -> c)`
to make the final signature (maybe with the first two switched)
`(a -> b -> c) -> (a -> b -> Boolean) -> [a] -> [b] -> [c]`
So then what is _currently_ being proposed as `innerJoin` would be expressed as 
`innerJoin(identity)`

buzzdecafe:
@Bradcomp thanks for the clarification, I think I see what you are saying now. You want a function that describes the shape of the output (viz `c`); is that a fair statement?

Bradcomp:
@buzzdecafe That's correct.  A function that describes the shape of the output, based on the 'matching' items for each list.

e.g, `innerJoin(merge, eqProps('id'))`

buzzdecafe:
ok; might it be more consistent with the rest of `R` API to expose `innerJoinWith` that takes a "shape" function, and then `innerJoin` could be implemented as `innerJoin(identity)`? 

In any event, I don't want to hold up this PR too much

Bradcomp:
Sorry for the last minute bikeshedding, I don't want to hold it up either!

CrossEye:
@Bradcomp:

This is definitely not bikeshedding: it's a fundamental API question, and one I know you brought up when this [first surfaced][co].  I also think it's quite important, although it might not be an actual deal-breaker, since we can implement `innerJoin` in one pass and the more generic version in a later pass, most likely reimplementing the first on in terms of the more generic one.  But then we do hit the damned bikeshed, because we need to know which function should get the really nice name of `innerJoin`?:

```js
//    :: a -> b -> Boolean) -> [a] -> [b] -> [a]
// or :: (a -> b -> c) -> (a -> b -> Boolean) -> [a] -> [b] -> [c]

That we really would like to get right the first time around.

  [co]: https://github.com/ramda/ramda/pull/1597#issuecomment-218342779

File: test/innerJoin.js (hunk 2 above)
Comment (CrossEye) [inline test/innerJoin.js]: While I agree that we can well proceed, I still share @buzzdecafe's disquiet about this particular test, for the same reasons already expressed.

@davidchambers, do you have a strong objection to removing this assertion?

**Comment by CrossEye** [comment]
> The example in the [documentation][1] is better expressed in terms of `intersection`. :stuck_out_tongue_winking_eye:

A [slight variant][sv] might better capture it.

----------

> This is not right. All else being equal we rightly favour a function's existing behaviour.

That's *almost* right.  But there's a critical word missing: 

> All else being equal we rightly favour a function's **documented** existing behaviour.

We should not let accidents of implementation drive our future API.  If we document the behavior, either in our API docs or our unit test suite, then I believe we have to have a much more systematic process for altering it.  If we don't, we should not feel particularly bound by it.  So if we need to alter `map` in some way, there is no need to be concerned with the fact that the current implementation happens to give a half-way reasonable value for `map(String.charCodeAt, 'foobar')`.

  [1]: http://ramdajs.com/docs/#intersectionWith
  [sv]: http://goo.gl/CKp0s0

**Comment by davidchambers** [comment]
When finding the intersection of two values of type `Array a` we'll usually want the default equality semantics of `a`, in which case `intersectionWith` is unnecessary. If we wish to use custom equality semantics, the lack of constraints on the return type gives the implementation too much freedom in my opinion. Consider case-insensitive string comparison, for example:

```javascript
R.intersectionWith(R.useWith(R.equals, [R.toUpper, R.toUpper]),
                   ['Ramda', 'Sanctuary'],
                   ['ramda', 'sanctuary', 'underscore']);
// => ['Ramda', 'Sanctuary']
```

If we consider both arrays to contain values of type `a`, as `intersectionWith` currently does, there are four valid return values:

  - `['Ramda', 'Sanctuary']`
  - `['Ramda', 'sanctuary']`
  - `['ramda', 'Sanctuary']`
  - `['ramda', 'sanctuary']`

If we do not require the implementation to return values in a specific order there are *eight* valid return values:

  - `['Ramda', 'Sanctuary']`
  - `['Ramda', 'sanctuary']`
  - `['ramda', 'Sanctuary']`
  - `['ramda', 'sanctuary']`
  - `['Sanctuary', 'Ramda']`
  - `['Sanctuary', 'ramda']`
  - `['sanctuary', 'Ramda']`
  - `['sanctuary', 'ramda']`

The definition I'm advocating has exactly one valid return value for each set of inputs.

> We should not let accidents of implementation drive our future API.

I agree. I just happen to believe we accidentally implemented the *right* function despite setting out to implement the wrong one. :stuck_out_tongue_winking_eye:

CrossEye:
I'm suggesting too that `Array`s are mostly an accident.  They're the only thing that was available until `Set`s became fairly standard.  I would really like to get a version 1.0 out the door that lets us consider a ES6/ES2015+ version 2 in which we take things like `Set` for granted.

davidchambers:
We don't want to use the native Set type as it doesn't have value-based equality semantics.

Bradcomp:
For what it's worth, I just ran into another case where the semantics of `intersectionWith` would cause problems.  I still think an `innerJoin` and `innerJoinWith` would very helpful to have in the library.

guidsen:
An innerJoin function is really missing. Coming from Lodash FP, used to the `intersectionBy` method that will also return the values like you passed them. It was pretty confusing how the `intersectionWith` method in Ramda worked. Had to look into the source and was kinda shocked by this logic https://github.com/ramda/ramda/blob/v0.22.1/src/intersectionWith.js#L43.

So a big 👍 !

CrossEye:
I'm all for merging this, if we could agree to remove the assertion that's causing some consternation for @buzzdecafe and me.

test/innerJoin.js
davidchambers: > @davidchambers, do you have a strong objection to removing this assertion?

I do, but the most important thing is to merge this pull request in some form. Perhaps you or @buzzdecafe would like to push a commit to my branch to remove the assertion? That way the Git history will be a testament to my strong objection. :stuck_out_tongue_winking_eye:

guidsen:
Is there any status update?

CrossEye:
I just took David's suggestion from https://github.com/ramda/ramda/pull/1868#r94101128.  If we can get that merged, I'm ready to go.  Sorry this one fell off the radar.

davidchambers:
This pull request is ready to merge now, I believe.

CrossEye:
Merged.   This will be in the next release.

davidchambers:
> Perhaps you or @buzzdecafe would like to push a commit to my branch to remove the assertion? That way the Git history will be a testament to my strong objection.

You did a squash merge, @CrossEye! The change I so opposed is now attributed to me! :persevere:

I don't use squash merging in my projects partly because I don't like an author's work being modified by third parties. Merge commits don't bother me.

kedashoe:
> You did a squash merge, @CrossEye! The change I so opposed is now attributed to me! :persevere:
> 
> I don't use squash merging in my projects partly because I don't like an author's work being modified by third parties. Merge commits don't bother me.

I think squash merges are the default? Is there a way to change the default?

davidchambers:
> I think squash merges are the default? Is there a way to change the default?

Yes. The __Merge button__ section at <https://github.com/ramda/ramda/settings> currently shows:

>   - [x] Allow merge commits
>   - [x] Allow squash merging
>   - [x] Allow rebase merging

For Sanctuary repositories only the first option is permitted.

CrossEye:
> You did a squash merge, @CrossEye! The change I so opposed is now attributed to me! :persevere:

I apologize.  I shouldn't ever merge from the bus.  But it just seemed so easy.

I'll see if I can straighten this out tonight.

davidchambers:
> I'll see if I can straighten this out tonight.

It's not a problem. I can point people to this pull request and protest my innocence. ;)

CrossEye:
> It's not a problem. I can point people to this pull request and protest my innocence. ;)

True.  Let the world know that David wanted to keep that test!

But I really need to improve my Git skills -- and clearly even my GitHub skills.  :cry:

Issue #1528, opened 2015-11-20
pipeP fails when first argument is function that does not return promise


bwestergard:
This seems like obviously pathological behavior. Tested on version 0.18.0.

```javascript
var R = require("ramda")
var Promise = require("bluebird")

var mulslow = R.curry(function (x, y) {
    return new Promise(function (resolve, reject) {
        resolve(x * y)
    })
})

R.pipeP(mulslow(3))(3).then(console.log) // works
R.pipeP(mulslow(3),R.multiply(3))(3).then(console.log) // works
R.pipeP(mulslow(3),R.multiply(3),mulslow(3))(3).then(console.log) // works
R.pipeP(R.multiply(3), mulslow(3))(3).then(console.log).catch(console.log) // does not work.
R.pipeP(R.multiply(3))(3).then(console.log).catch(console.log) // does not work.
```

CrossEye:
I don't have time to look right now, but this might have got lost in the #1260, #1391 shuffle.

bwestergard:
#1314 as well. I'm very worried about this.

asaf-romano:
Yup, this is the same issue I reported in #1314. At the end, we introduced our own version of pipe to workaround this. Here's it:
```js
export function pipeP(first, ...rest)
{
    return R.pipeP(nAry(first.length,
                        (...args) => {
                            try
                            {
                                return Promise.resolve(first(...args));
                            }
                            catch(ex)
                            {
                                return Promise.reject(ex);
                            }
                        }),
                   ...rest);
}
```

scott-christopher:
There are a few issues here:
* If we don't require the first function to return a valid Promise, there is a possibility that none of the functions given to `pipeP` will return a Promise, meaning that `pipeP` as a whole won't return a Promise which makes for inconsistent return types.
* We can avoid this by calling `Promise.resolve` on the output of the first function if it is missing a `then` method, however this then adds a dependency on a Promise implementation.
* We can work around this by hand-rolling our own minimal Promise implementation, but we're then hand-rolling our own Promise implementation.

This can however be avoided currently by users starting the `pipeP` composition with the appropriate function from the Promise implementation that lifts plain values into a Promise instance. This is only necessary if the first function would otherwise not return a Promise.

e.g. for ES6 Promises:
```js
R.pipeP(Promise.resolve, R.add(42), someOtherAsyncFn);
// or if the pipeline needs to accept multiple arguments
[...]

Issue #683, opened 2014-12-09
use maybes for unsafe operations


davidchambers:
Certain Ramda functions are unsafe. Take this quartet, for example:

```
R.head :: [a] -> a | Undefined  -- unsafe
R.last :: [a] -> a | Undefined  -- unsafe
R.tail :: [a] -> [a]            -- safe, but misleading
R.init :: [a] -> [a]            -- safe, but misleading
```

I suggest the following types:

```
R.head :: [a] -> Maybe a
R.last :: [a] -> Maybe a
R.tail :: [a] -> Maybe [a]
R.init :: [a] -> Maybe [a]
```

We could also provide the unsafe versions:

```
R.unsafeHead :: [a] -> a | Undefined
R.unsafeLast :: [a] -> a | Undefined
R.unsafeTail :: [a] -> [a]
R.unsafeInit :: [a] -> [a]
```

We'd also provide the following function which takes a default value as its first argument:

```
R.fromMaybe :: a -> Maybe a -> a
```

Let's consider a simple example: taking the first element of a `[String]` of unknown length and converting it to upper-case.

``` javascript
// Naive solution (will throw if list is empty)
R.toUpper(R.head(list))

// Pre-emptive check
R.ifElse(R.isEmpty, R.always('<default>'), R.pipe(R.head, R.toUpperCase))(list)

// Maybe
R.fromMaybe('<default>', R.map(R.toUpperCase, R.head(list)))
```

It's cleaner and less error-prone to map over a maybe than to rely on checking for the empty list whenever one uses a member of this quartet.

These types are used by [purescript-arrays](https://github.com/purescript/purescript-arrays). I'd like to Ramda do the same.

davidchambers:
These [list functions][1] are available in [Sanctuary][2], a JavaScript library for manipulating values without null checks. Sanctuary plays nicely with Ramda, and takes advantage of `R.map`'s dynamic dispatch.


[1]: https://github.com/plaid/sanctuary/blob/v0.2.0/index.js#L212-L226
[2]: https://github.com/plaid/sanctuary

buzzdecafe:
Sanctuary looks pretty nice

CrossEye:
Very nice!

austbot:
Hey folks, just so I'm on the right track. If not using fantasyland types would this be a safealternative.

```js
R.head = R.curry(function(a){
  let n = R.nth(0);
  return R.isNil(n) ? [] : n;
});
```

[...]

PR #1869, opened 2016-08-19
deprecate R.composeP and R.pipeP

davidchambers:
We often advise Ramda users against using promises, yet we provide functions for working with promises. Moreover, such functions are not necessary when working with tasks/futures, which support the regular `compose`/`pipe`/`map`/`chain` operations.

@CrossEye indicated in https://github.com/ramda/ramda/issues/1528#issuecomment-239966677 that he's on board with deprecating `composeP` and `pipeP`. Others have also shown support for the move.

### DIFF
--- a/src/composeP.js
+++ b/src/composeP.js
@@ -15,6 +15,7 @@ var reverse = require('./reverse');
  * @param {...Function} functions
  * @return {Function}
  * @see R.pipeP
+ * @deprecated since v0.23.0
  * @example
  *
  *      //  followersForUser :: String -> Promise [User]
--- a/src/pipeP.js
+++ b/src/pipeP.js
@@ -17,6 +17,7 @@ var tail = require('./tail');
  * @param {...Function} functions
  * @return {Function}
  * @see R.composeP
+ * @deprecated since v0.23.0
  * @example
  *
  *      //  followersForUser :: String -> Promise [User]


buzzdecafe:
this is fine with me -- although it might be nice to show in the cookbook how one can compose with wrapping promises, e.g. using [futurize](https://github.com/futurize/futurize)

CrossEye:
I'd like to ping @yuri on this, but I'm on-board.

yuri:
@CrossEye @buzzdecafe @davidchambers Whatever their shortcomings, promises are a part of the ES6 spec and are returned by many an increasing range of native web APIs. Additionally, many popular frameworks today expect promises as inputs. And that's before async / await. Any middling JavaScript developer today understands how they work. Removing functions that work with promises would mean telling newcomers that in order to use Ramda with basic web APIs they would need to (1) take time to learn how Futures work and (2) adopt two new libraries that do not currently have adoption or support on the same scale as Ramda itself. They will also then need to figure out how to turn their Future-returning functions into something that returns promises so that they could then use those functions with libraries that rely on promises... And I imagine that debugging such promises-to-futures-back-to-promises code won't be a lot of fun.

asaf-romano:
I'm not on board either, for the same reasons mentioned by yurl. Async-await has just approved for Stage 4 (and native implementations are months-away from shipping). Removing basic promises support at this stage is being hostile to the language ecosystem.

semmel:
Just some browser apis which rely on Promises:
* [Fetch](https://developer.mozilla.org/en-US/docs/Web/API/GlobalFetch/fetch)
* [Web Cryptography](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto)
* [Permissions](https://developer.mozilla.org/en-US/docs/Web/API/Permissions/query)
* [Web Audio](https://developer.mozilla.org/de/docs/Web/API/AudioContext/decodeAudioData)
* [HTMLMediaElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/play)
* [RTCPeerConnection](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/setRemoteDescription)
* [MediaDevices](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia)
* [Storage](https://storage.spec.whatwg.org)
* [Battery Status API](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/getBattery)

I'd like to interact with those well-documented apis easily and friction-free in the next couple of  years. 

Inserting a wrapper like [futurize](https://github.com/futurize/futurize) really obscures the apis and adds to much complexity for our team. I use Ramda as FP utility managing our basic JS data structures ( *promised (?)* K-V-pairs, arrays, etc...). I don't like it to be forced to use a specific island data type. Our business code will *directly* utilise the promises provided by those apis.

davidchambers:
Ramda has a [philosophy][1]. Some things are compatible with the philosophy; others are not. Functions which mutate their arguments, for example, are incompatible. Despite this, people are free to write programs which use Ramda functions and functions which mutate their arguments. Ramda itself, though, will never contain functions which mutate their arguments.

Similarly, people are free to write programs which use Ramda and promises. But because promises violate the philosophy of Ramda<sup>†</sup>, Ramda should not contain *promise-specific* functions.

---

† Promises promote the use of exceptions for error handling. Promises provide a bespoke interface rather than conform to a common, well-defined interface. `Promise#then` conflates `map`, `chain`, and `fork`/`run`.


[1]: http://fr.umio.us/the-philosophy-of-ramda/

buzzdecafe:
TIL [bespoke](http://www.dictionary.com/browse/bespoke); excellent word! 

I'm wavering on this one. I agree wih @davidchambers that Promises are an odd fit _at best_. But I am sympathetic to the points made by @yuri , @asaf-romano , and @semmel

CrossEye:
I'm also wavering, and for the same reasons.

Let's at least think about this for a few more days before making any decisions.

semmel:
> But because promises violate the philosophy of Ramda†, Ramda should not contain promise-specific functions.

I could buy that. However just two remarks:

- It limits the range of problems where I can use Ramda (I need another library for Promise compositions)
- It's a dangerous conclusion. For example: Imperative coding with `if`-`else` statements and *plain JS variables* is not Ramda philosophy. Does this mean that Ramda should just work on `Either`s or `Maybe`s, and implementations working with "regular" functions (x => f(x)) will get removed? Afaik Either or Maybe are not even part of Ramda right now.

davidchambers:
> I need another library for Promise compositions

Yes, this is the problem. We shouldn't need anything special in order to compose promise-returning functions. We should be able to write:

```javascript
//    f :: String -> Promise Number
const f = R.pipe(
  somePromiseReturningFunction1,
  R.chain(somePromiseReturningFunction2),
  R.chain(somePromiseReturningFunction3),
  R.map(someRegularFunction),
  R.chain(somePromiseReturningFunction4)
);
```

My view is that we should encourage Ramda users to treat promise-returning functions as they (hopefully) treat exception-throwing functions and (perhaps) functions which return bogus values such as `NaN`. A promise-returning function is a "foreign" function which should be wrapped in a function of a more composable type.

Perhaps my stance is too principled. Promises are here to stay (sadly). The goal of Ramda, though, is not to provide a function for every piece of functionality provided by the language. The goal is to make a certain style of programming possible (and fun!) in JavaScript. Saying no to functions which do not further this goal (and periodically pruning existing functions) has given us the consistent, fairly opinionated library we have today.

> Imperative coding with `if`-`else` statements and *plain JS variables* is not Ramda philosophy. Does this mean that Ramda should just work on `Either`s or `Maybe`s, and implementations working with "regular" functions (x => f(x)) will get removed?

I'm confused by this point. Sure, `if` statements do not fit with the Ramda philosophy as they're only useful in conjunction with side effects, but we do have `cond ? then : else` which is an expression. Also, side effects which are local to a function are fine. Many Ramda functions reassign to local variables and mutate private objects.

Although I would love Ramda's head function to be of type `[a] -> Maybe a`, I don't think anyone is arguing that somehow all functions should make use of the Maybe type or the Either type. These types do a good job of capturing the possibility of failure, but this is not applicable to all functions.

semmel:
@davidchambers Okay I agree. 
But with my `Either`-`Maybe` argument I just wanted to express my insecurity about the often cited Ramda philosophy: Aren't `Either` and `Maybe` invented to abstract away those ugly imperative `if`-`else` failure checking statements? So when working on "plain JS types" too, don't the Ramda functions encourage that particular *bad* imperative coding style? 
Don't get me wrong - I just wan't to find out in which direction Ramda is going: In the future can I upgrade to a new version of Ramda just by fixing some names and function arguments, or do I have to write parts anew because a future Ramda will just cover fewer application scenarios?
(I hope I made myself clear this time - as always ;-)

yuri:
To get more concrete, can someone provide an example of what the following code would look like using `futurize` + `ramda-fantasy` + regular `pipe()` instead of `pipeP()`:

```js
const composedPromiseReturningFunction = R.pipeP(
  somePromiseReturningFunction1,
  somePromiseReturningFunction2,
  R.map(someRegularFunction)
);
```

Maybe I am overestimating the amount of complexity this would add?

This to me is a very common scenario: we've got some promise-returning functions provided by some 3rd party or native API and we need to get a promise (or a promise-returning function) in the end because that's what is expected by another 3rd party or native API.

davidchambers:
> So when working on "plain JS types" too, don't the Ramda functions encourage that particular *bad* imperative coding style?

Yes, they do. I argued in #683 that Ramda should provide safe versions of `head` and other partial functions. We decided this is a step too far for Ramda, so I created and released [Sanctuary][1].

> To get more concrete, can someone provide an example of what the following code would look like using `futurize` + `ramda-fantasy` + regular `pipe()` instead of `pipeP()`:
>
> ```javascript
> const composedPromiseReturningFunction = R.pipeP(
>   somePromiseReturningFunction1,
>   somePromiseReturningFunction2,
>   R.map(someRegularFunction)
> );
> ```

I imagine something like this:

```javascript
//    convert :: (a -> Promise b c) -> (a -> Future b c)
const convert = f => x => f(x).then(Future.of, Future.reject);

const f = R.pipe(
  convert(somePromiseReturningFunction1),
  convert(somePromiseReturningFunction2),
  R.map(someRegularFunction)
);
```


kedashoe:
@yuri here is a more complete example

```js
let Future = require('fluture');
let Futurize = require('futurize');
let R = require('ramda');

let futurize = Futurize.futurizeP(Future);

let waitPlus1 = (x, cb) => setTimeout(() => cb(x + 1), 123);
let waitTimes2 = (x, cb) => setTimeout(() => cb(x * 2), 123);

let API = {
  
  foo: function(x) {
    return new Promise((good, bad) => waitPlus1(x, good));
  },
  
  bar: function(x) {
    return new Promise((good, bad) => waitTimes2(x, good));
  },
};

let foo = futurize(API.foo);
let bar = futurize(API.bar);

let err = x => console.log('uh oh', x);
let log = x => console.log('great!', x);

let regularFn = R.add(10);

let pipeP = R.pipeP(
  API.foo,
  API.bar,
  regularFn,
  log
);

let pipeF = R.pipe(
  foo,
  R.chain(bar),
  R.map(regularFn),
  Future.fork(err, log)
);

pipeP(1);
pipeF(2);
```

You can copy and paste that into http://requirebin.com/ to run it. Somebody help me out if I've got anything wrong, just getting started using `Future`s. I think the error you'll see in your console is because of https://github.com/futurize/futurize/pull/44.

I'm fine leaving `pipeP` in for now. As @davidchambers mentioned, JS is committed to them, so as developers of JS library, I think it is good to provide functions that users like for dealing with them.

For me this hearkens back to https://github.com/ramda/ramda/issues/1874#issuecomment-242991462. Maybe ultimately `pipeP` doesn't belong in Ramda "Core", but if we have Ramda Fantasy and Ramda Lens and potentially others like Ramda String, Ramda Math, etc, I'd be ok with a Ramda Promise.

yuri:
@kedashoe The example I had in mind would involve converting back to promises, this would be more like:

```js
let foo = futurize(API.foo);
let bar = futurize(API.bar);
let pipeF = R.pipe(
  foo,
  R.chain(bar),
  R.map(regularFn)
);

pipeF(2).promise().then(log, err);
```

Which doesn't work at the moment because of the `futurize` issue that you noted, but presumably will after the fix.

Speaking of which: the fact that `fork()` manages to call the success callback even in case of an error – is that how Futures are supposed to work? This seems very unintuitive coming from the promise-land.

Also, is there a way to do this without having to add the (seemingly unnecessary) `R.chain` and `R.map`? This seems to add a lot of mental clutter.

xgrommx:
By the way, we need only kleisli composition `R.composeK`

kedashoe:
> Speaking of which: the fact that fork() manages to call the success callback even in case of an error – is that how Futures are supposed to work?

That does seem surprising. It is `fork` itself that is throwing the error, so maybe it is running the computation and then throwing. Hard to debug on requirebin, but it was working without errors with all the latest versions on my machine.

> Also, is there a way to do this without having to add the (seemingly unnecessary) R.chain and R.map?

I was hoping someone who knows more about this would chime in here, but from what I understand, `chain` and `map` are necessary and are indicative of why https://github.com/fantasyland/fantasy-land exists in the first place. Rather than the `then` method that promises use which is both `chain` and `map`, `Future`s adhere to well known specifications that allow them to be used like many other data types (Maybe, etc).

yuri:
@kedashoe No problem here is not that there is an exception being thrown, but that this exception seems to then be handled in a somewhat odd way, where the success callback is invoked despite the error.  But this might be just a fluture issue. Ramda-Fantasy's Future seems to handle errors as one would expect.

> Rather than the then method that promises use which is both chain and map, Futures adhere to well known specifications that allow them to be used like many other data types (Maybe, etc).

For 99% of JavaScript programmers, there is no semantic overlap between `map()` and `then()`. The first function transforms every element of a list using the provided function, the second registers a callback. One needs to operate with a fairly abstract notion of `map` (defined in terms of Functors rather than lists) to see this as an overlap. So, I suspect that this use of `.map()` will be extremely confusing to the overwhelming majority of JS developer (including those who have a decent amount of experience with "functional" JavaScript). Which, in turn, would make code written using Ramda + Futures very hard to maintain in the long term, due to the difficulty of onboarding new developers.

I am biased when it comes to `composeP`, obviously, but one of the things I personally like about it is that it allows you to mix functions that return promises with functions that return synchronous values. So, `R.composeP(f, g, h)` is just that, regardless of which of the functions are synchronous. In contrast, with Futures we will need to wrap f, g, and h in either `R.map` or `R.chain` depending not only on what sort of function each of them is, but also depending on whether the functions earlier in the composition return Futures or regular values. Which makes sense at an abstract level, but sounds painful in practice.

kedashoe:
I think you are wildly underestimating JavaScript programmers, both novice (in my experience developers are excited to learn about these concepts) and experienced (I think they can tell a good story about why Futures are nice (maybe not me, you seem unconvinced!)).

In any event, while I may weigh the relevant factors quite a bit differently, as I said earlier, I vote for keeping `pipeP` :)

davidchambers:
> I am biased when it comes to `composeP`, obviously, but one of the things I personally like about it is that it allows you to mix functions that return promises with functions that return synchronous values. So, `R.composeP(f, g, h)` is just that, regardless of which of the functions are synchronous. In contrast, with Futures we will need to wrap f, g, and h in either `R.map` or `R.chain` depending not only on what sort of function each of them is, but also depending on whether the functions earlier in the composition return Futures or regular values.

This is a philosophical issue. Should we provide an API which is simple or one which is easy? Ideally we'd provide an API which is both simple and easy, but it is not possible here. We must choose one or the other.

Given:

```haskell
f :: String -> Maybe Number
g :: Number -> Number
x :: Maybe String
```

Assume we want to calculate roughly `g(f(x))`, but of course we want the types to line up. With the simple API we must fit the pieces together ourselves:

```haskell
chain(f) :: Maybe String -> Maybe Number
chain(f)(x) :: Maybe Number
map(g) :: Maybe Number -> Maybe Number
map(g)(chain(f)(x)) :: Maybe Number

compose :: (b -> c) -> (a -> b) -> (a -> c)
compose(map(g))(chain(f)) :: Maybe String -> Maybe Number
compose(map(g))(chain(f))(x) :: Maybe Number
```

Alternatively we could define a Maybe-specific version of `compose` which interrogates values as they flow through the composition, and:

  - applies `map(f)` to the value if it is a Maybe; otherwise
  - applies `f` to the value directly.

We could then simply write:

```haskell
magicMaybeCompose(g)(f) :: (String | Maybe String) -> Maybe Number
magicMaybeCompose(g)(f)(x) :: Maybe Number
```

The advantage of the magic is one can now throw things at `magicMaybeCompose` and there's a reasonable chance the function will do the right thing. Convenient.

There are two downsides, though. We've lost the ability to infer the types of `f` and `g`. We've also lost the ability to work with certain functions. We cannot include `h :: Maybe Foo -> Maybe Bar` in our composition. Why not? Because given `Just(someFoo)`, `magicMaybeCompose` will erroneously apply `h` to `someFoo` rather than to `Just(someFoo)`. So whenever we work with `magicMaybeCompose` we must keep in mind that it doesn't always do the right thing.

Interestingly, `Array#concat` is similarly magical. `[1, 2, 3].concat(4)` evaluates to `[1, 2, 3, 4]`, but so does `[1, 2, 3].concat([4])`. This means we should always write `.concat([x])` __to disable the magic__.

My final point is that no one is asking for a version of `compose` which allows functions which operate on arrays and functions which operate on individual values to be used interchangeably:

```javascript
magicArrayCompose(R.sum)(R.inc)([1, 2, 3]);
// ≣ R.compose(R.sum)(R.map(R.inc))([1, 2, 3])
// ≣ R.sum(R.map(R.inc)([1, 2, 3]))
// ≣ R.sum([R.inc(1), R.inc(2), R.inc(3)])
// ≣ R.sum([2, 3, 4])
// ≣ 9
```

Many JavaScript programmers are comfortable mapping over arrays. The tragedy of the Promises/A+ spec is that it deprived countless JavaScript programmers of the opportunity to become just as comfortable mapping over futures/tasks (revealing in the process that mapping is a general concept).

semmel:
Well written @davidchambers - that post nails my issues down to the point! 👍 

However I'm still on the `magicArrayCompose` side of the discussion. This is why:
I remember reading the important [Hughes paper "Why Functional Programming Matters"](http://www.cse.chalmers.se/~rjmh/Papers/whyfp.html) where in the introduction he argues that FP should not be characterised by LIMITATIONS and negative statements (no side-effects, no assignments, referential transparency...) 

> The functional programmer sounds rather like a medieval monk, denying himself the pleasures of life in the hope that it will make him virtuous.

but by "PROGRAMMING GLUE":
> ... that functional languages provide two new, very important kinds of glue (glueing functions and programs together). ... So ... programs that can be modularised in new ways, and thereby greatly simplified. This is the key to functional programming’s power - it allows greatly improved modularisation. It is also the goal for which functional programmers must strive - smaller and simpler and more general modules, glued together with the new glues ...

So back to the problem I am facing every day in web programming is dealing with *major JS APIs*  from different sources which are all *far away* from that Ramda philosophy. * What `magicArrayCompose` gives me (and what I want from Ramda) is that GLUE binding those original, *non-wrapped* data structures easily together.

**E.g. 
- Exceptions communicating recoverable errors (`JSON.parse`)
- Great libraries which mutate their data structures (e.g. Moment.js)

CrossEye:
> Well written @davidchambers - that post nails my issues down to the point! 👍 

Yes, @davidchambers, another excellent post.

> However I'm still on the `magicMaybeCompose` side of the discussion. This is why:
I remember reading the important [Hughes paper "Why Functional Programming Matters"][wf] where in the introduction he argues that FP should not be characterised by LIMITATIONS and negative statements (no side-effects, no assignments, referential transparency...) [ ... ] but by "PROGRAMMING GLUE".

I reread that paper several times a year; I find it tremendously insightful and a very useful guide.

> So back to the problem I am facing every day in web programming is dealing with *major JS APIs*  from different sources which are all *far away* from that Ramda philosophy. * What `magicMaybeCompose` gives me (and what I want from Ramda) is that GLUE binding those original, *non-wrapped* data structures easily together.

I'm all for providing that glue, but I can't go along with "*non-wrapped*".

I can't prove it, of course, but I truly believe that a big part of Ramda's success so far has been its insistence on simplicity: functions have a single signature with only a small number of fixed-type parameters.  Parameter order is nearly always obvious; often the only difficult decisions have to do with naming.  Ramda makes no guesses, and tries very hard to avoid ambiguities.  Simplicity rules.

A `magicMaybeCompose` would have to try to infer your intent from the types you supply, and as David points out, that's fraught with complexity.  And it would seem to me something that would quickly spiral out of control:  If you think your `JSON.parse` errors might be recoverable, then you need to trap the specific error, so should it be a `magicEitherCompose`?  Or might there be some of both in a single composition?  Should it be `magicMaybeEitherCompose`?  And so on.

If on the other hand, Ramda simply provides the composition mechanism along with utilities for wrapping, then you can do what you need, and its not overly complex to do so.



  [wf]: http://www.cse.chalmers.se/~rjmh/Papers/whyfp.html

semmel:
> If you think your JSON.parse errors might be recoverable, then you need to trap the specific error, so should it be a magicEitherCompose?

I am not searching for a silver bullet for composition i.e. function chaining. 

Will I need `magicEitherCompose` or more likely just some `magicCompose`? It depends always on the concrete problem like in the next two examples:
*Preliminary note:* There is this great mental model of [Railway oriented programming](https://fsharpforfunandprofit.com/posts/recipe-part2/) (No need to read the text, just the images tell the story)

So when I use a reactive, aka stream-based, aka observable-programming library like BaconJS I can just shove any type into my streams and map (or map and flatten) a chain of functions on that data stream. On that "data railway", if I like, I can anytime fork off a new separate failure track using my own types (mostly just string enums) for failures or I just *abuse* the builtin error track by throwing.

If chaining (composing) promise-returning functions I can do the same thing by *abusing* the builtin rejection track of Promises, even if [somewhere advised differently](http://www.2ality.com/2016/03/promise-rejections-vs-exceptions.html?m=1). Filtering out the operational errors is not so scary:

```javascript
var
  rejected = x => Promise.reject(x),
  fails = isFailure => func => p => Promise.resolve(p).catch( R.ifElse(isFailure, func, rejected) );
// example
var TIMEOUT = 1, WRONG_PASSWORD = 2;
var recovered = fails(R.is(Number))(code => { 
   console.log("recovering from failure:" + code); 
   return 100 + code; 
});
recovered(Promise.reject(TIMEOUT)).then(console.log.bind(console), console.error.bind(console));
// -> [warning] recovering from failure: 1
// -> 1
```

Both cases have in common that I don't need data wrappers while employing just the facilities provided by the libraries.

Bradcomp:
It seems to me that the argument is more related to _signalling_ that Ramda supports promises than it is about actually supporting them.  I use promises all the time with Ramda, but the cases where I've used `composeP` and `pipeP` I can count on one hand.  

What about deprecating these functions and providing a more focused and obvious solution for dealing with promises in function compositions?

I use this: `const then = R.curry((f, p) => p.then(f))`.  Obviously this could be done using `invoker` too.

It doesn't solve the underlying issues with promises, but it has a few benefits - it's explicit, it makes our types clearer, and it can be used as a drop in to replace the functionality of the deprecated functions.
 
`pipeP(f, g, h)` becomes `pipe(f, then(g), then(h))`

It also serves as a signal that Ramda can be used cleanly with promises, and gives some insight into how that can be done cleanly.

olsonpm:
> What about deprecating these functions and providing a more focused and obvious solution for dealing with promises in function compositions?

My understanding (thanks much to James Forbes on gitter) is that promises have properties which make it difficult to use reliably in a functional manner.  Therefore having **any** support for promises in ramda seems better fit for a project outside of core.  I agree there would be utility in supporting promises as they've organically permeated js libs, and there's no getting around the need to consume them.

davidchambers:
> `pipeP(f, g, h)` becomes `pipe(f, then(g), then(h))`

This is definitely an improvement. Replacing `pipeP` with `then` is a reasonable half measure.

buzzdecafe:
i agree; after all this discussion, i am back where i started; i support deprecation here. A wrapper such as @Bradcomp 's `then` would be a nice way to offer promise composition.

🐪

yuri:
I quite like @Bradcomp's solution. Are we talking about including this implementation of `then" in Ramda as `R.then`? If so, then I think this takes care of the issue.

CrossEye:
Yes we are, and it's now available for review in #1906.  I agree that this is a very useful middle ground.

asaf-romano:
Still **very** uneasy about this, but if this happens after all, we should have an implementation in the cookbook of current pipeP/composeP (and in the release notes for the next release).

cristianocd:
Removing `pipeP` and `composeP` is great, but I think `.then(onFulfill, onReject)` `.all` `. catch` `.resolve` `.reject` are promise specific functions that ramda core should not care about.

Maybe adding examples to the cookbook, like the proposed `pipe(f, then(g), catch(h)...)` and something like:
```javascript
const coerceArray = unless(isArrayLike, of)
const openPromiseArray = when(isArrayLike, x => Promise.all(x)) 
const then = f => promise => promise.then(...coerceArray(f)).then(openPromiseArray)

const pipePromise = (...functions) =>
    pipe(
      x => Promise.resolve(x),
      ...map(then, functions)
    )
```
is enough, or am I missing something?

keithamus:
Just thought I'd add my 2c here and demonstrate how trivial it is to set up promise support which composes over the rest of the functions:

```js
const ofP = R.bind(Promise.resolve, Promise); // sad we have to do this one... but hohum
const thenP = R.invoker(1, 'then');
// could also add catch, or even thencatch trivially.
const catchP = R.invoker(1, 'catch');
const thenCatchP = R.invoker(2, 'then');

const promiseAdd2Multiply4 = R.pipe(
  ofP,
  thenP(R.add(2)),
  thenP(R.multiply(4))
)

let logP = thenP(R.tap(console.log)) 

logP(promiseAdd2Multiply4(2)) // logs 16

// alternatively, just wrap `thenP` and pass an `ofP` which
// allows composition of any functions to Promise based ones.
const add2Multiply4 = R.pipe(
  R.add(2),
 R.multiply(4)
);
thenP(add2Multiply4)(ofP(2))

```

Seems like this is easy enough to support in userland.

davidchambers:
I'll close this pull request as it proved contentious.

CrossEye:
I just pinged everyone on #1906.  If people like that, then I would want to resurrect this one.

Bradcomp:
I am in favor of resurrecting this as well.

PR #1870, opened 2016-08-19
Removed semvar lock on travis ci node versions

amilajack:


buzzdecafe:
i don't see any reason why not -- not sure why it was so specific.
🐪

CrossEye:
:herb:

PR #1871, opened 2016-08-20
Compose properties

Bradcomp:
Starting small on property based testing.  Adds a couple property tests for compose based on recent conversations in the gitter.  One just ensures that it's actually performing function, while the other tests for the associative property.

Hopefully this can serve as a starting point for additional property tests in the future.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "handlebars": "3.0.x",
     "istanbul": "^0.4.x",
     "js-yaml": "^3.2.5",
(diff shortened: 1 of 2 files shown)


buzzdecafe:
❤️ this

:100:

davidchambers:
Nice. Could you clean the commit history?

CrossEye:
:fireworks: :balloon: :star2: 

Very nice!

PR #1877, opened 2016-08-22
deprecate R.isArrayLike

davidchambers:
Let's deprecate this function. If a glance at the implementation is not sufficient to persuade you that this function should be removed, I'll spend some time writing a case for the deprecation.

### DIFF
--- a/src/isArrayLike.js
+++ b/src/isArrayLike.js
@@ -14,6 +14,7 @@ var _isString = require('./internal/_isString');
  * @sig * -> Boolean
  * @param {*} x The object to test.
  * @return {Boolean} `true` if `x` has a numeric length property and extreme indices defined; `false` otherwise.
+ * @deprecated since v0.23.0
  * @example
  *
  *      R.isArrayLike([]); //=> true


buzzdecafe:
> If a glance at the implementation is not sufficient to persuade you that this function should be removed, I'll spend some time writing a case for the deprecation.

😆 ouch that hits a little close to home @davidchambers -- although IIRC you were there when it got added to the library.

Anyways, I don't need any persuading on this -- definitely 🐪 for me

davidchambers:
I did not mean to imply that the implementation is of low quality. My point is that the implementation reveals that "array-like" is far from well-defined in JavaScript.

buzzdecafe:
no worries, i agree that it is ill-defined.

CrossEye:
Thanks, folks!  This one has too long been a thorn in our side!

jdalton:
>  My point is that the implementation reveals that "array-like" is far from well-defined in JavaScript.

The internal operation [ToLength](http://www.ecma-international.org/ecma-262/7.0/#sec-tolength) is a good start.

PR #1878, opened 2016-08-23
 Fix duplicate declaration errors in converge example in REPL

aonghusonia:
When you run the converge example in the ramda REPL you get a duplicate declaration error for the add, subtract & multiply functions

### DIFF
--- a/src/converge.js
+++ b/src/converge.js
@@ -24,15 +24,15 @@ var reduce = require('./reduce');
  * @return {Function} A new function.
  * @example
  *
- *      var add = (a, b) => a + b;
- *      var multiply = (a, b) => a * b;
- *      var subtract = (a, b) => a - b;
+ *      var plus = (a, b) => a + b;
+ *      var multiplyBy = (a, b) => a * b;
+ *      var minus = (a, b) => a - b;
  *
- *      //≅ multiply( add(1, 2), subtract(1, 2) );
- *      R.converge(multiply, [add, subtract])(1, 2); //=> -3
+ *      //≅ multiplyBy( plus(1, 2), minus(1, 2) );
+ *      R.converge(multiplyBy, [plus, minus])(1, 2); //=> -3
  *
  *      var add3 = (a, b, c) => a + b + c;
- *      R.converge(add3, [multiply, add, subtract])(1, 2); //=> 4
+ *      R.converge(add3, [multiplyBy, plus, minus])(1, 2); //=> 4
  * @symb R.converge(f, [g, h])(a, b) = f(g(a, b), h(a, b))
  */
 module.exports = _curry2(function converge(after, fns) {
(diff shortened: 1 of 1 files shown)


davidchambers:
Thanks, @aonghusonia. What do you think about using `R.add` and friends in the example, to save us from redefining these functions?

aonghusonia:
that would be the simplest solution but I did not know if add, subtract & multiply had been defined for some illustrative purpose similarly to the reduce example https://github.com/ramda/ramda/pull/1777.
 If not I agree example could simply read:

//≅ R.multiply( R.add(1, 2), R.subtract(1, 2) );
R.converge(R.multiply, [R.add, R.subtract])(1, 2); //=> -3

var add3 = (a, b, c) => a + b + c;
R.converge(add3, [R.multiply, R.add, R.subtract])(1, 2); //=> 4

davidchambers:
Someone recently suggested moving away from arithmetic examples. This is sensible, I think. How about something like this instead?

```javascript
> R.converge((a, b, c) => a + b + c, [R.toUpper, R.toLower, R.identity])('Ramda')
'RAMDAramdaRamda'
```

Using `converge` with binary functions is pretty hard to imagine in real code. I think a single example featuring an array of unary functions (such as the one above) would suffice.

aonghusonia:
how about:

aonghusonia:
how about: 

var capitalizeFirst = R.converge(R.concat, [R.compose(R.toUpper, R.head), R.tail]);

capitalizeFirst("converge a list of branching functions"); //=>"Converge a list of branching functions"

davidchambers:
That's another option. It's certainly a more realistic example. The only problem is that it's more simply expressed without `converge`:

```javascript
//    capitalizeFirst :: String -> String
const capitalizeFirst = R.replace(/./, R.toUpper);
```

aonghusonia:
Courtesy of [tacit Ramda](https://medium.com/@bobiblazeski/tacit-ramda-c914775ff4b1#.n2t5osovn) :

```javascript
var avg = R.converge(R.divide,[R.sum,R.length])
avg([1,2,3,4,5,6,7])//=> 4

//or

var isDuplicate  = R.converge(R.complement(R.equals),[R.indexOf,R.lastIndexOf]);
isDuplicate(2,[2,1,3,1,6]) //=> False
isDuplicate(1,[2,1,3,1,6]) //=> true
```

I'm leaning towards avg as a simple but realistic example but you can still do it more simply as:
avg = lift(divide)(sum,length)

davidchambers:
I like the `avg` example. :)

davidchambers:
:deciduous_tree:

CrossEye:
I'm still catching up on everything.  This is a very nice improvement, thanks!

I would offer a caveat.  If we get around to it, it would be slightly clearer still with this variant:

```js
avg([1, 2, 3, 4, 5, 6]); //=> 3.5
```

which makes it very explicit *which* average we're discussing.  (That is mean, not median or mode.)

But this is far from critical.

PR #1888, opened 2016-09-02
Adding a new resource to introductions section in  README.md

luijar:
I used Ramda extensively in Functional Programming in JavaScript, published this year. I think it's also a good reference for developers wanting to learn more about it. 

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ Introductions
 * [Why Curry Helps](https://hughfdjackson.com/javascript/why-curry-helps/) by Hugh Jackson
 * [Hey Underscore, You're Doing It Wrong!](https://www.youtube.com/watch?v=m3svKOdZijA&app=desktop) by Brian Lonsdorf
 * [Thinking in Ramda](http://randycoulman.com/blog/categories/thinking-in-ramda) by Randy Coulman
+* [Functional Programming in JavaScript (Manning 2016)](https://www.manning.com/books/functional-programming-in-javascript)
 
 
 Philosophy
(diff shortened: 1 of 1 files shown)


buzzdecafe:
cool! so when can we expect our free copies? :stuck_out_tongue_winking_eye: 

congrats on publishing your book.

davidchambers:
I see many occurrences of `_` in chapter 3. :stuck_out_tongue_winking_eye:

I don't think this book belongs in the list. The current items are Ramda-centric and/or curry-centric rather than general FP resources for JavaScript programmers.

luijar:
Yeah there's a bit of Lodash in CH03 lol :-)  after teaching currying it's Ramda from then on for the most part. 
I understand. Thanks.

CrossEye:
I agree that it probably doesn't belong on this list, but I would suggest that you make a PR to

    https://github.com/stoeffel/awesome-fp-js

And I wonder if Ramda should offer a link to that list as well.

I did read the book, BTW.  It was several months ago, and I keep saying that I'm going to get around to writing reviews of the various FP in JS books available, but it never bubbles to the top of my list.  FWIW, this one ended up below Fogus and far above Mantyla.  But so far no one has come close to Lonsdorf and Braithwaite.

luijar:
Thanks for the feedback. Yes, absolutely, I read them all as well and I'm a huge fan of Lonsdorf and Braithwaite. The stuff they write about is so incredibly well articulated. Sorry, didn't intend for this to be a book review lol, certainly that's not ramda-centric. 

Thanks again.

PR #1889, opened 2016-09-03
Property tests

Bradcomp:
I've added some property tests for `add`, `always`, `curry`, and `flip`.

### DIFF
--- a/test/add.js
+++ b/test/add.js
@@ -1,5 +1,6 @@
 var R = require('..');
 var eq = require('./shared/eq');
+var jsv = require('jsverify');
 
 
 describe('add', function() {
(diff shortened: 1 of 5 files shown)


test/add.js
```diff
@@ -22,3 +23,17 @@ describe('add', function() {
   });
 
 });
+
+describe('add properties', function() {
+  jsv.property('commutative', jsv.nat, jsv.nat, function(a, b) {
```
davidchambers: Is there a reason to restrict ourselves to `jsv.nat` here?

Bradcomp: Only that I was conflating `nat` and `number` last night.  Whoops!

buzzdecafe:
:cow2:

PR #1900, opened 2016-09-13
fantasy-land@1.0.x

davidchambers:
There are several breaking changes on the cards for `fantasy-land@1.0.0`:
- [x] fantasyland/fantasy-land#146: prefix method names
- [x] fantasyland/fantasy-land#145: flip `ap` argument order
- [x] fantasyland/fantasy-land#135: define Traversable in terms of `traverse` rather than `sequence`
- [ ] <del>fantasyland/fantasy-land#162: remove support for `.constructor.empty`</del>

I'll keep this pull request up to date as pull requests are merged into the FL repository.


CrossEye:
These look good so far.  I'm looking forward to seeing fantasyland/fantasy-land#135.

I'm not as convinced by fantasyland/fantasy-land#162, though.

buzzdecafe:
The `ap` change is a bit of a PITA. This seems like yet another case of vacillating over which way to convert what is semantically an infix operation to prefix.

davidchambers:
> The `ap` change is a bit of a PITA.

It's a good change: it makes `ap` consistent with `map` and `chain`.

Ramda users won't notice anything as `R.ap` won't change. :)

SimonRichardson:
The `ap` change has been much needed for a _very_ long time.

SimonRichardson:
FYI: https://github.com/fantasyland/fantasy-land/releases/tag/1.0.0

If you spot anything out of the norm, then let me know so we can get a new release out.

davidchambers:
This pull request is now ready for review.

I rewrote the implementations of the Id, Maybe, and Either types which are used in the test suite in various places. As a result, Sanctuary is no longer a dev dependency.

Questions:

  - __Should we support the old method names (e.g. `map`) in addition to the new names (e.g. `fantasy-land/map`) for some period of time?__ I hope the answer is no.

  - __Should we continue to dispatch to `sequence` for some period of time?__ Again, I hope the answer is no.

src/traverse.js
```diff
@@ -30,5 +31,9 @@ var sequence = require('./sequence');
  *      R.traverse(Maybe.of, safeDiv(10), [2, 0, 5]); //=> Nothing
  */
 module.exports = _curry3(function traverse(of, f, traversable) {
-  return sequence(of, map(f, traversable));
+  return typeof traversable['fantasy-land/traverse'] === 'function' ?
+    traversable['fantasy-land/traverse'](f, of) :
+    reduceRight(function(acc, x) { return ap(map(prepend, f(x)), acc); },
+                of([]),
+                traversable);
 });
```
davidchambers: Could someone review this change? I copied the existing `sequence` implementation but replaced `x` with `f(x)` in `ap(map(prepend, x), acc)`. Is that correct?

buzzdecafe: that should be correct AFAICT. That said, FWIW I am certain there is a more performant implementation than this.

buzzdecafe:
this looks good -- better -- to me. I prefer dispatching to the prefixed names, it is much less ambiguous.

🐪

davidchambers:
@CrossEye?

CrossEye:
Sorry, will look this evening.

CrossEye:
> I prefer dispatching to the prefixed names, it is much less ambiguous.

I pushed for this in FL, but now that I see it, I worry about it a bit.  My concern might be mostly a transitional one, but it's real.  Will this break all sorts of existing useful integrations because other libraries are still using, say, `map` and not `fantasy-land/map`?

This is a very serious concern.  And I'm not sure what to do about it.

In every other way, I love this change.  But that makes me squeamish about just doing it.

davidchambers:
> Will this break all sorts of existing useful integrations because other libraries are still using, say, `map` and not `fantasy-land/map`?

Would you like to support both for some period of time? It'll be messy, but shouldn't be difficult.

It's worth noting that not supporting prefixed names would break all sorts of useful integrations. ;)

buzzdecafe:
> Would you like to support both for some period of time? 

No. Tear off that band-aid

safareli:
Also as argument order of `ap` has changed it might be problematic.?

davidchambers:
> Also as argument order of `ap` has changed it might be problematic.?

This isn't a problem since the `ap` change was released in the same version of the spec as the prefixes. So we'd check for `fantasy-land/ap`, and dispatch to it (the new way) if present; otherwise we'd check for `ap` and dispatch to it (the old way) if present.

CrossEye:
My concern is broader, though.

I think I may be alone in this, but I do not think of the dispatching as specific to FantasyLand.  While that's an extremely useful specification, I would like for our dispatching to be broader than that.

We started the dispatching before incorporating FL, and at the time my thoughts were on how it might someday be used to integrate streams and lists -- something similar to what Highland eventually did.  When there were suggestions to integrate with FL, it was a no-brainer to me.  But I never felt as though that was the _raison d'être_ for our dispatching.  I do think that it should have a somewhat privileged place, because it is the only specification out there doing this important work.  However, I would like to be able to work well with other tools, regardless of whether they adhere to the FantasyLand spec.

We haven't been great about this yet, but that doesn't mean that we shouldn't get there.  I think it would be ideal to simply be able to work with libraries like Immutable, like Highland, like Flyd.

buzzdecafe:
> I do not think of the dispatching as specific to FantasyLand. While that's an extremely useful specification, I would like for our dispatching to be broader than that.

I am 100% on the other side. I only want to dispatch to "methods" that have well-defined semantics. It's amusing that Highland is in your example list, because [it is moving toward FL compatibility](https://github.com/caolan/highland/pull/114). And there's this from Flyd's [README](https://github.com/paldepind/flyd): "Complies to the fantasy land applicative specification." So maybe Immutable is out. Oh well. You can always [bridge the gap with lenses.](https://medium.com/@drboolean/lenses-with-immutable-js-9bda85674780#.h70xkoxgd)

CrossEye:
Well, https://github.com/caolan/highland/pull/114 hasn't been updated in 20 months.  I'm not sure that really counts as "moving towards"!  :smile:

But Immutable, Flyd, and Highland were meant to be simply examples.  If you have a data structure with a meaningful method `obj.take :: Int -> a` then I would love it if Ramda offered to make this a pure function for you with `R.take(5, obj); //=> (something of type a)`.

>  I only want to dispatch to "methods" that have well-defined semantics.

Well-defined, but unenforceable, doesn't seem all that powerful to me.  I would offer a somewhat privileged place to FantasyLand, mostly because the types there work together to offer additional power (think `lift`, for instance.)

We've discussed this before (see especially #1173) and it got pretty hot.  But I'm much closer to the view [David expressed][dc] some time ago:

> I'm starting to see that dynamic dispatch is not a nice extra: it's the very core of Ramda (or should be).
> 
> Ideally, every Ramda function would simply dispatch: there would be no code to speak of (except perhaps code to derive functions from other functions).


  [dc]: https://github.com/ramda/ramda/pull/1160#issuecomment-109888577

buzzdecafe:
i remain strongly in favor of merging this and moving away from promiscuous dispatching. i would like to get it in and move toward 1.0

evilsoft:
Are there any plans to still support the old dispatches?
Or is part of upgrading to 1.x mean you will have to wait until all other FL compliant libs in your project are moved over to 1.x of FL?

So for instance if I am using Ramda to provide some point-free goodness on say most.js for my super 🔥 frontend masterpiece. Does that mean I am expected to wait until most.js decides to update to the new FL before I can update to 1.x of Ramda? and I would have to wait for ALL FL-compatible stuff to make the move before I can upgrade to 1.x.

Am I correct in that assumption?

CrossEye:
@evilsoft:

That discussion is part of what's slowing us down here.  I'm in favor of of retaining the old dispatching, either temporarily or permanently, which is a change from what @davidchambers proposed here.  @buzzdecafe is quite opposed.

This is a fairly fundamental disagreement, having to do with what we think our dispatching is about.  @buzzdecafe and I are still trying to schedule the knock-down-drag-out fight over this issue.

evilsoft:
HA. @CrossEye gonna get some 🍿 and have a little sit down.

CrossEye:
Oh, don't worry, I think I can take him.  :smile:

buzzdecafe:
@CrossEye your mouth is writing checks your ass can't cash

evilsoft:
🍿

CrossEye:
I don't know.  There was a proposal to have three televised debates on this issue over several weeks.  We had @paldepind, @scott-christopher, and @davidchambers agreeing to be moderators.

But with the threats from my opponent to seat John-David Dalton and Jeremy Ashkenas in the front row, I'm really not sure he's serious about this.

paldepind:
@buzzdecafe 

> I am 100% on the other side. I only want to dispatch to "methods" that have well-defined semantics.

I very much agree with this.

@CrossEye 

> I do think that it should have a somewhat privileged place, because it is the only specification out there doing this important work.

Well, it's actually not the only spec. I'm working on a fantasy land replacement called [Jabz](https://github.com/Funkia/jabz) :smile: It's very much a work in progress but one key aim is to provide a specification that allows for significantly better performance than FL. I.e. its the same abstractions—but at a lower overhead.

> While that's an extremely useful specification, I would like for our dispatching to be broader than that.

To me it would be very helpful if you could give a practical real-world example of how our current dispatching is beneficial in ways that is unrelated to FL. Creating useful abstractions is tricky. And in other languages using them typically happens in principled ways using interfaces or type classes.

I'm not at all convinced that ad-hoc dispatching through duck-typing results in integration with 3rd party libraries in a meaningful way. We currently do have dispatching but we still don't integrate with Highland nor Immutable. But I'd love to be proven wrong with some actual code of how our dispatching makes life with other libraries easier. But without that it feels like we're keeping the dispatching based on a vague vision about somehow working with many libraries but without a clear idea about how dispatching actually realizes that dream.

CrossEye:
>> I do think that it should have a somewhat privileged place, because it is the only specification out there doing this important work.

> Well, it's actually not the only spec. I'm working on a fantasy land replacement called [Jabz][ja] :smile: It's very much a work in progress but one key aim is to provide a specification that allows for significantly better performance than FL. I.e. its the same abstractions—but at a lower overhead.

I'm very excited to see it!  Are you going to forgo compatibility with Ramda?

>> While that's an extremely useful specification, I would like for our dispatching to be broader than that.

> To me it would be very helpful if you could give a practical real-world example of how our current dispatching is beneficial in ways that is unrelated to FL. 

I will try to do so soon.  Perhaps this evening.


  [ja]: https://github.com/Funkia/jabz

CrossEye:
@paldepind: I haven't forgotten this.  I'm just behind on everything.  I'll try again tomorrow.

safareli:
A bit related [ramda-fantasy#131](https://github.com/ramda/ramda-fantasy/pull/131)

buzzdecafe:
@CrossEye please either state your case or let this one through. Already over a month old, and i think the only  thing holding it up is you?

CrossEye:
I've been talking about this [for a long time][oc]:

> Say we had this tree, `tree1`:
> 
>                    A
>         +----------^----------+
>         |          |          |
>         B          C          D
>     +---^---+  +---^---+  +---^---+
>     |   |   |  |       |  |   |   |
>     1   E   4  6       7  8   F   4
>       +-^-+                   |
>       |   |                   |
>       2   3                   6
>      
>      
> Then if we wanted to map `square` over that tree, the result should be pretty obvious:
> 
>     map(square, tree1); //=>
>     
>                    A
>         +----------^----------+
>         |          |          |
>         B          C          D
>     +---^---+  +---^---+  +---^---+
>     |   |   |  |       |  |   |   |
>     1   E  16 36      49 64   F  16
>       +-^-+                   |
>       |   |                   |
>       4   9                  36
>       
> 
> Similarly if we wanted to run `filter` over the tree:
> 
>     filter(odd, tree1); //=>
>       
>              A      
>         +----^-----+
>         |          |
>         B          C
>     +---^---+      |    
>     |       |      |
>     1       E      7
>             |            
>             |
>             3

I could add that perhaps we would want to do 

    take(5, tree1); //=>
      
            A
        +---^--+
        |      |
        B      C
    +---^---+  +
    |   |   |  |
    1   E   4  6
      +-^-+
      |   |
      2   3

or any of a number of other similar functions.  To me, perhaps the most central point of Ramda is to make it easy to work via functional composition.  If Ramda can take the `map` *method* of `Array.prototype` and of other useful types and turn them into plain functions, it's adding value, because those functions can now be composed.  That's just as true for something that doesn't happen to have exposed `fantasy-land/map` as for something that has.  The only advantage, to my mind of the prefixing is that it makes it clear that someone is claiming to follow certain laws, and so we and others could take advantage of these laws for certain implementations to create useful abstractions such as `lift`.

But there is nothing here that says that `map` is so much more important than `filter` or `take` or many others methods we could lift into functions.

I would delegate to FantasyLand if it's available, but if not, I still like delegating to user-supplied functions.  I'm having a hard time because this seems so obviously useful to me that I really don't understand the objection.

  [oc]: https://github.com/ramda/ramda/issues/729#issuecomment-70351882

buzzdecafe:
this is blocking the `chainRec` PR, so let's come to a decision timely. I am strongly in favor. @CrossEye is opposed unless old dispatches are kept. @davidchambers says it's messy but possible. so let's do that for now and come back later to clean up

davidchambers:
`Z.concat` knows how to merge string maps:

```javascript
> Z.concat({concat: Math.sin}, {concat: Math.cos})
{concat: Math.cos}
```

The problem is that we want to preserve the current Ramda behaviour of dispatching to `concat`:

```javascript
> R.concat({concat: Math.sin}, {concat: Math.cos})
NaN
```

We can't easily use `Z.concat` as it only respects the `fantasy-land/`-prefixed property.

davidchambers:
I've updated this pull request. Shall we update the documentation to indicate that FL functions dispatch to both prefixed and unprefixed methods? It's particularly confusing in the case of `ap` because the argument order differs.

CrossEye:
I didn't really follow the FL discussion around swapping the order of `ap`, but I was assuming that we would follow along with this.  Is there any good reason (besides backward compatibility) to keep a different order for our version?

As to the documentation, I think we first need to settle what we want to do with dispatching.  If this is a short way station on the way to removing it, then there is no need.  If we are going to keep something more substantial, then I think we need to figure out *how* we'd like to document it.

davidchambers:
> I didn't really follow the FL discussion around swapping the order of `ap`, but I was assuming that we would follow along with this.  Is there any good reason (besides backward compatibility) to keep a different order for our version?

The order in which a `fantasy-land/ap` method takes its arguments and the order in which `R.ap` takes its arguments are unrelated concerns. The change to the FL spec was made to bring `ap` into line with `map` and `chain`.

Before:

```purescript
mapMethod   :: Functor f => f a ~> (a -> b) -> f b
apMethod    :: Apply f => f (a -> b) ~> f a -> f b
chainMethod :: Chain f => f a ~> (a -> f b) -> f b
```

After:

```purescript
mapMethod   :: Functor f => f a ~> (a -> b) -> f b
apMethod    :: Apply f => f a ~> f (a -> b) -> f b
chainMethod :: Chain f => f a ~> (a -> f b) -> f b
```

The "correct" argument order for functions which wrap these methods is unaffected:

```purescript
map   :: Functor f => (a -> b) -> f a -> f b
ap    :: Apply f => f (a -> b) -> f a -> f b
chain :: Chain f => (a -> f b) -> f a -> f b
```

This is the confusing code:

> ```javascript
>     typeof applyX['fantasy-land/ap'] === 'function' ?
>       applyX['fantasy-land/ap'](applyF) :
>     typeof applyF.ap === 'function' ?
>       applyF.ap(applyX) :
> ```

We have `` applyX `ap` applyF `` in one case and `` applyF `ap` applyX `` in the other. This looks like a typo but it reflects the change to the FL spec. Were we to dispatch to `ap` assuming the *new* method signature we'd break backwards compatibility for no gain.

CrossEye:
>  Were we to dispatch to `ap` assuming the *new* method signature we'd break backwards compatibility for no gain.

Wouldn't that just correct the mistake that's been in one of most confusing functions for far too long?

| http://ramdajs.com/docs/#ap |
-----------------------------------
| Dispatches to the `ap` method of the second argument, if present. Also treats curried functions as applicatives. |

Well, no, it dispatches to the `ap` method of the *first* argument, if present:

```js
ap(
  {ap: (x) => `ap_first(${x})`, toString: () => 'first'}, 
  {ap: (x) => `ap_second(${x})`, toString: () => 'second'}
); // => "ap_first(second)"
```

It strikes me that this is an opportunity to correct that mistake in a consistent fashion.

syaiful6:
is there any progress with this?

syaiful6:
@davidchambers also consider adding ```bimap```, i don't see it here. But, maybe i am wrong. 👍

davidchambers:
@syaiful6, #1938 will bring us `bimap` and all the other functions defined in [sanctuary-type-classes][1]. Scott needs to decide what to do with this pull request, though, before we move on to #1938.


CrossEye:
I didn't feel that it was particularly waiting on me, although I might be the only one on one side of the discussion.  The main issue is that we need to decide what we're doing with dispatching before we can figure out what to do with this issue, #1938, #1949, #1950.  The last of the discussion seems to have been on #1938.

paldepind:
> I'm very excited to see it! Are you going to forgo compatibility with Ramda?

I've been meaning to respond to this. Mostly because it will lead me to a Ramda-related issue that I'd like to discus. But I wanted to get Jabz to a more presentable state before doing so. Now it has a decent [readme](https://github.com/Funkia/jabz), [API documentation](https://funkia.github.io/jabz/) and a [blog post](http://vindum.io/blog/introducing-jabz/) describing how and why Jabz differs from Fantasy Land.

To answer the question: Jabz uses non-prefixed method names, so currently it is mostly compatible with Ramda. However, being Ramda compatible is actually not something I care about. Because, to me, our current Fantasy Land integration brings very little value. Actually I consider it somewhat harmful as it sits in an uncanny valley where the little it does brings false expectations that it may do more. I'm sorry for these harsh statements. Allow me to explain myself.

Let's say I'm brand new user of Ramda. I've heard that it integrates with a specification for monads, traversables, etc. I've used these abstractions in some other language so I'm familiar with them.

The first thing I do is I start writing some code that uses maybe. I write something like `R.map(myFuntion, myMaybe)`. It works and I'm happy.

Pretty soon I end up with a maybe value inside a maybe value which I want to flatten. I know that maybe is a monad, I know that a core capability of monads is that they support flattening (the operation is often called `join`) and I know that Ramda works with monads. So I go looking for a function in Ramda to do this. I look at both `R.join` and `R.flatten` but both do something different. Now I'm confused. Flatten/join is a core operation on monads. Why can't I find it?

Later I do some work with a foldable. I know that all foldables support a size/length operation. Again I am disappointed to find out that `R.length` doesn't work on foldables. This disappointed increases when I realize that only `R.reduce` works on foldable. None of all the other function that _could_ work on foldables actually does.

My point with the above is this: The abstractions specified by fantasy land are a lot more than just the few methods that are explicitly part of the spec. The many operations that can be derived from these are equally important. To my knowledge `R.lift`, `R.pipeK`, `R.composeK`, `R.both` and `R.either` are the only derived operations that we provide. People who actually wants to get the full benefit of the abstractions will be disappointed. To put things into perspective, we have 1 function that works with foldables, Jabz has 15 functions (and counting) and Haskell's Data.Foldable has about 25. 

Supporting abstraction in this half-baked way is bad. But fixing this and going "all in" on the abstractions poses several problems as well:

* The size and complexity of the library would grow due to the added derived methods.
* We would have to include a lot of functions that are only useful with fantasy land abstractions. For instance having `toArray` for foldables is very useful. Another example, [`mapTo`](https://funkia.github.io/jabz/#mapTo) is useful for many functors but not with arrays. These functions are useless to people who just use Ramba with plain JS objects and arrays. This might alienate them.
* Performance for many operations would be bad. For instance `R.isEmpty` can work for any foldable but with fantasy land it would take `O(n)` time which is pretty tragic (the Jabz spec support it in constant time).

In my opinion we should tighten the focus of Ramda. It should be a set of utility functions for native JavaScript objects. All dispatching should be removed. People who wants to use fantasy land abstractions may use [sanctuary-type-classes](https://github.com/sanctuary-js/sanctuary-type-classes) which looks very nice.

SimonRichardson:
I don't normally chime in, but this is horseshit!

> Because, to me, our current Fantasy Land integration brings very little
value

So without FL you wouldn't have any standardisation for this, it's not
perfect and it's slow moving, but someone moaning about it because it
doesn't fit "their" needs is about as constructive as chocolate fire
guard!!!

> Jabz uses non-prefixed method names, so currently it is mostly compatible
with Ramda.

Do you even know why FL has them? Because a LOT of ramda developers wanted
it. They pushed hard for it!

paldepind:
> So without FL you wouldn't have any standardisation for this, it's not
perfect and it's slow moving, but someone moaning about it because it
doesn't fit "their" needs is about as constructive as chocolate fire
guard!!!

I think you're misunderstanding what I wrote. I'm not saying that Fantasy Land doesn't add any value. I'm saying that what Ramda is currently doing with the specification adds very little value. We just dispatch to the methods that Fantasy Land compatible structures already provide but forgoes any integration beyond that.

> Do you even know why FL has them? Because a LOT of ramda developers wanted
it. They pushed hard for it, in fact I was the one voice against it!

Yes I know. I was part of the discussion back then.

> We've asked you to open issues, you haven't and so you're now trying this method.

I don't know what you mean with the statement "you're now trying this method". This is a completely separate issue. I don't know what you think I'm trying. I'm making a case against dispatching in Ramda. I've been arguing for its removal several times in the past.

I will open issues I just haven't gotten around to do it yet. Ideally I want to get every idea from Jabz into Fantasy Land.

rickmed:
TL;DR: Any progress on this? Its a shame that is blocking other PRs from some time now.

I came here because I'm having problems with ap which is following FL's <v1.0 order. There is #2002 to upgrade to the current order but is blocked by this issue. I'm trying to write js in a functional way which I've really come to appreciate appreciate (using Tasks/Futures for async stuff which is js bread and butter), but now I'm limited to only using the data.task lib which uses the old ap order, the rest of the future libs uses the new ap order (including the new folktale AFAIK) so I need to decide between them or ramda. I know this issue is related to some decisions like the dispatching discussion (I've read a fair share of the issues), so I guess what I'm saying is that maybe no decision is worse than any?

Jacse:
Agree with @rickmed this should be resolved ASAP

CrossEye:
Honestly, there's simply not been a lot of work going into Ramda lately.  I *hope* that's about to change, but no guarantees.

kedashoe:
:+1: Can we merge this now @CrossEye ?

PR #1905, opened 2016-09-16
Update README with ng-ramda install instructions

JustinDFuller:
If you are interested, I've made a ng-ramda package that provides Ramda as an injectable Angular service.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,8 @@ These script tags add the variable `R` on the browser's global scope.
 
 Or you can inject ramda into virtually any unsuspecting website using [the bookmarklet](https://github.com/ramda/ramda/blob/master/BOOKMARKLET.md).
 
+Or you can install via ng-ramda npm package ```npm install ng-ramda```  then follow the directions at https://www.npmjs.com/package/ng-ramda
+
 ### Build
 
 * on Unix-based platforms, `npm run build` updates __dist/ramda.js__ and __dist/ramda.min.js__


CrossEye:
I'm not sure whether this belongs in the ReadMe.  We'll have to think about that a bit.  But I'm definitely interested in seeing your package.  There's a long-abandoned one at https://github.com/ramda/ramdangular.

pinging @prathamesh7pute

JustinDFuller:
@CrossEye Thanks for at least thinking about it. The reason I made this package is because ramdangular puts everything in the Angular root scope, which doesn't really fit in well with an Angular component architecture + dependency injections. It's essentially just crowding angular's "global" scope, which isn't very good to do.

CrossEye:
I'm not really much of an Angular user.  Do I understand correctly that all your module does is to make Ramda something that can be injected into an Angular module?  Is there any advantage to that over importing it in your file?

I can certainly understand your objection to what Ramdangular does.  I spoke to the original author, and he doesn't really use it much himself either.  Ramda dropped the equivalent ability from its core some time ago.  But I'm not sure I understand what the advantage is to have Ramda, which should need no access to anything internal to your application, to be injected into your Angular modules.  What am I missing?

buzzdecafe:
this would fit nicely in the [Related Projects](https://github.com/ramda/ramda/wiki) list, but I agree with @CrossEye probably not for the README

PR #1906, opened 2016-09-16
add function R.then

Bradcomp:
Per the discussion here: https://github.com/ramda/ramda/pull/1869#issuecomment-247388811

Hopefully this serves as a good com`promise`.

### DIFF
--- a/src/fail.js
+++ b/src/fail.js
@@ -1,0 +1,38 @@
+var _curry2 = require('./internal/_curry2');
+var _isFunction = require('./internal/_isFunction');
+var toString = require('./toString');
+
+
+/**
+ * Returns the result of applying the onFailure function to the value inside
+ * a failed promise. This is useful for handling rejected promises
+ * inside function compositions.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig (b -> c) -> (Promise a b) -> (Promise c d)
+ * @sig (b -> (Promise c d)) -> (Promise a b) -> (Promise c d)
+ * @param {Function} onFailure The function to apply.  Can return a value or a promise.
+ * @param {Promise} p
+ * @return {Promise} The result of calling `p.then(null, onFailure)`
+ * @see R.then
+ * @example
+ *
+ *      var failedFetch = (id) => Promise.reject('bad ID');
+ *      var useDefault = () => ({ firstName: 'Bob', lastName: 'Loblaw' })
+ *
+ *      //recoverFromFailure :: String -> Promise [{firstName, lastName}]
+ *      var recoverFromFailure = R.pipe(
+ *        failedFetch(12345),
+ *        R.fail(useDefault),
+ *        R.then(R.pick(['firstName', 'lastName']))
+ *      );
+ */
+module.exports = _curry2(function fail(f, p) {
+  if (p == null || !_isFunction(p.then)) {
+    throw new TypeError(toString(p) + ' does not have a method named "then"');
+  }
+
+  return p.then(null, f);
+});
--- a/src/then.js
+++ b/src/then.js
@@ -1,0 +1,37 @@
+var _curry2 = require('./internal/_curry2');
+var _isFunction = require('./internal/_isFunction');
+var toString = require('./toString');
+
+
+/**
+ * Returns the result of applying the onSuccess function to the value inside
+ * a successfully resolved promise. This is useful for working with promises
+ * inside function compositions.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig (a -> c) -> (Promise a b) -> (Promise c d)
+ * @sig (a -> (Promise c d)) -> (Promise a b) -> (Promise c d)
+ * @param {Function} onSuccess The function to apply.  Can return a value or a promise.
+ * @param {Promise} p
+ * @return {Promise} The result of calling `p.then(onSuccess)`
+ * @see R.fail
+ * @example
+ *
+ *      var makeQuery = (email) => ({ query: { email }});
+ *
+ *      //getMemberName :: String -> Promise [{firstName, lastName}]
+ *      var getMemberName = R.pipe(
+ *        makeQuery,
+ *        fetchMember,
+ *        R.then(R.pick(['firstName', 'lastName']))
+ *      );
+ */
+module.exports = _curry2(function then(f, p) {
+  if (p == null || !_isFunction(p.then)) {
+    throw new TypeError(toString(p) + ' does not have a method named "then"');
+  }
+
+  return p.then(f);
+});
--- a/test/then.js
+++ b/test/then.js
@@ -1,0 +1,55 @@
+var assert = require('assert');
+var Q = require('q');
+
+var R = require('..');
+var eq = require('./shared/eq');
+
+
+describe('then', function() {
+  it('invokes then on the promise with the function passed to it', function(done) {
+    var p = Q(1);
+
+    var f = function(n) {
+      eq(n, 1);
+      done();
+    };
+
+    R.then(f, p)
+  });
+
+  it('flattens promise returning functions', function(done) {
+    var incAndWrap = R.compose(Q, R.inc);
+    var asyncAddThree = R.pipe(incAndWrap, R.then(incAndWrap), R.then(incAndWrap));
+
+    R.then(function(result) {
+      eq(result, 4);
+      done();
+    })(asyncAddThree(1));
+  });
+
+  it('throws a typeError if the then method does not exist', function() {
+    assert.throws(
+      function() { R.then(R.inc, 1); },
+      function(err) {
+        return err.constructor === TypeError &&
+               err.message === '1 does not have a method named "then"';
+      }
+    )
+
+  });
+
+  it('is not dependent on a particular promise implementation', function(done) {
+    var thennable = {
+      then: function(f) {
+        return f(42);
+      }
+    };
+
+    var f = function(n) {
+      eq(n, 42);
+      done();
+    };
+
+    R.then(f, thennable);
+  });
+});
--- a/test/fail.js
+++ b/test/fail.js
@@ -1,0 +1,39 @@
+var assert = require('assert');
+var Q = require('q');
+
+var R = require('..');
+var eq = require('./shared/eq');
+
+
+describe('fail', function() {
+  it('catches failed promises', function(done) {
+    var p = Q.reject(1);
+
+    var f = function(n) {
+      eq(n, 1);
+      done();
+    };
+
+    R.fail(f, p);
+  });
+
+  it('does nothing to successfully resolved promises', function(done) {
+    var asyncAddTwo = R.pipe(Q, R.then(R.inc), R.fail(R.inc), R.then(R.inc));
+
+    R.then(function(result) {
+      eq(result, 3);
+      done();
+    })(asyncAddTwo(1));
+  });
+
+  it('throws a typeError if the then method does not exist', function() {
+    assert.throws(
+      function() { R.fail(R.inc, 1); },
+      function(err) {
+        return err.constructor === TypeError &&
+               err.message === '1 does not have a method named "then"';
+      }
+    )
+
+  });
+});


davidchambers:
I see what you did there, [__@Brad`com`p__][1]. :stuck_out_tongue_winking_eye:


[1]: https://github.com/Bradcomp

src/then.js
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Returns the result of calling `then` on the supplied promise with the
+ * supplied function. This is useful for working with promises inside
+ * function compositions.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig (a -> b) -> Promise a -> Promise b
```
davidchambers: I think we need a second signature:

```haskell
then :: (a -> Promise b) -> Promise a -> Promise b
```

Essentially, this function is both `map` and `chain`. :cry:

test/then.js
```diff
@@ -0,0 +1,44 @@
...
+        return f(42);
+      }
+    };
+
+    var f = function(n) {
+      eq(n, 42);
+    };
+
+    R.then(f, thennable);
+  });
```
davidchambers: Are we simply asserting that `R.then(f, thennable)` does not throw?

test/then.js
```diff
@@ -0,0 +1,44 @@
...
+
+  it('works with any object that has a then method', function() {
+    var thennable = {
+      then: function(f) {
+        return f(42);
+      }
+    };
+
+    var f = function(n) {
+      eq(n, 42);
```
davidchambers: I think we need `done()` here. Otherwise, the test would pass if `f` is never called.

kwijibo:
Should there also be a `R.catch` function to accompany this?

Bradcomp:
@kwijibo That sounds good to me.  I can update my PR tonight if there is interest.

`catch` has meaning outside of promises, but it would probably be obvious what the function is for.  

This led me to think about another issue, which is that `Promise.prototype.then` takes two parameters (_both_ being optional per the spec :disappointed: ).

My thought is that having explicit functions for the success and failure case is better practice anyway, so having something like `fork` for promises would be a step too far.

src/then.js
CrossEye: > Essentially, this function is both `map` and `chain`.

Yes, but please, we try not to say that in polite company.

asaf-romano:
unary then is sure going to confuse some. And since we don't support optional arguments, I'm not sure what's the alternative.

kwijibo:
Could there be any value in the abstraction implied by `pipeP` and `pipeK`: `pipeWith` ?

```
const pipeP = R.pipeWith(R.then)
const pipeK = R.pipeWith(R.chain)
```

buzzdecafe:
((:bell:)) let's not let this one die

Bradcomp:
Would it help things or harm things if I added an `R.catch`?  Either to this PR or its own?

buzzdecafe:
it would help things.

Bradcomp:
I'm working on the `catch` function, and it's making me think a little deeper about the type signature for both `then` and `catch`.  For the `chain`-like signatures I think they should be the following (taking into account the onSuccess, onReject argument order):

`then :: (a -> (Promise c d)) -> (Promise a b) -> (Promise c d)`

`catch:: (b -> (Promise c d)) -> (Promise a b) -> (Promise c d)`

Both can take a non-promise-returning function as well though, which is problematic.  This is because if a non-promise-returning function within a promise chain throws you end up with a rejected promise.  So you end up with something like

`then:: (a -> c (throws d)) -> (Promise a b) -> (Promise c d)`

`catch:: (b -> c (throws d)) -> (Promise a b) -> (Promise c d)`

Is it worth it to document this behavior, or do we ignore the case where a synchronous function could throw?  Is it better to explicitly document the error case instead, as we know the type of a function once it throws?

`then:: (a -> c) -> (Promise a b) -> (Promise c (Error d))`

The issue with Promises conflating `map` and `chain` becomes more prevalent the more I work on this. 😢

src/catchP.js
```diff
@@ -0,0 +1,38 @@
...
+ *      var useDefault = () => ({ firstName: 'Bob', lastName: 'Loblaw' })
+ *
+ *      //recoverFromFailure :: String -> Promise [{firstName, lastName}]
+ *      var recoverFromFailure = R.pipe(
+ *        failedFetch(12345),
+ *        R.catchP(useDefault),
+ *        R.then(R.pick(['firstName', 'lastName']))
+ *      );
+ */
+module.exports = _curry2(function catchP(f, p) {
```
Bradcomp: The name `catch` is already taken 😜 .  I am not sure on this name because it creates an asymmetry with `then`, but I am lacking a decent alternative.  I don't like using `R.onSuccess` and `R.onFailure` because it's less obvious the functions relate to Promises.

CrossEye: How about `then` / `fail` ?

iamstarkov: i would say as far as `pipe` for functions, `pipeP` for promises and `catch` for functions, then `catchP` is a good name

CrossEye: The point of `then` is to help us *eliminate* `pipeP`/`composeP`.  :smile:

slavaGanzin:
Please use ES6's ```catch``` and not Q's ```fail```. Standard is standard, if you like it or not

p.s. Worked in project with three different promise libs. It's frustrating when you try to guess what method should have "this promise"

Bradcomp:
@slavaGanzin Are you concerned over the name of the function, or the implementation?

Neither `fail` nor `catch` are part of the Promise A+ standard.  While the test uses Q as the promise library, the implementation of the function does not use a `catch` or a `fail` method on the promise that gets passed in, and should work with all A+ compliant promises.

Because `catch` is a JavaScript keyword, it can't be used as a function name.  `fail` was suggested by @CrossEye as an alternative name, which I think is preferable to `catchP`. Perhaps it would be good to mention that the function will work with all A+ compliant promises in the docs?

slavaGanzin:
> Because catch is a JavaScript keyword, it can't be used as a function name.
You can try this in your browser and node.js REPL:
```
Promise.reject(1).catch(x => console.log(x))
```


IMHO:
I'm talking about naming. And I don't care what was Promises A+, I see what was implemented in node and chrome, and that was ```.catch``` and not ```.fail```

And, besides, I don't know why you are using any promise library - promises are supported in all modern browser/node distribution. I think in current state only polyfills for older environments should be supported.

iamstarkov:
pls, just dont use `Q` its obsolete. try native `Promise` its supported everywhere, consumers with old environments can polyfill their environment easily

iamstarkov:
okay, if its just to make [node 0.10 happy in travis](https://github.com/ramda/ramda/blob/master/.travis.yml#L4), then i would suggest to use [pinkie-promise](https://github.com/floatdrop/pinkie-promise), which is [ponyfill](https://ponyfill.com) for promises. Otherwise, node 0.12 support native promises already, and [node 0.10's lts support ends in the end of this month](https://github.com/nodejs/lts#lts-schedule)

iamstarkov:
in other words, even ponyfill is needed only until the end of the month

Bradcomp:
> You can try this in your browser and node.js REPL:

> `Promise.reject(1).catch(x => console.log(x))`

Try defining:
`function catch(f, p) { return p.catch(f); }`

I get 
`Uncaught SyntaxError: Unexpected token catch`
in Chrome.
So while we could define a catch method, we can't define it as a stand alone function - which is a problem for including it in the library as the R object is created by the build process and not defined directly.

> And, besides, I don't know why you are using any promise library - promises are supported in all modern browser/node distribution. I think in current state only polyfills for older environments should be supported.

I am using a Promise library for testing because the tests include support for Node versions that don't have native Promises.  See comments by @iamstarkov above.  I would be happy to use a different library for those tests, I just used the one that was already included as a dev-dependency in the project.

The function itself has no reliance on _any_ promise implementation, only on the A+ standard being followed.

I'll take a look at swapping out Q, but that would involve updating the tests for `composeP` and `pipeP` as well, because they rely on Q.

Bradcomp:
The only thing Q is being used for is to create Promises to test in our Promise handling functions.  I don't really know if it's worth swapping it out.  None of the library code relies on it, and it's just there to make sure the test suite doesn't break when testing on legacy systems.

iamstarkov:
> Uncaught SyntaxError: Unexpected token catch

though, its not `Promise.catch`, its `try {} catch(e) {}`

iamstarkov:
> The only thing Q is being used for is to create Promises to test in our Promise handling functions. I don't really know if it's worth swapping it out. None of the library code relies on it, and it's just there to make sure the test suite doesn't break when testing on legacy systems.

are gonna drop support for NodeJS 0.10?

iamstarkov:
if so, lets use native promises and remove 0.10 from travis config

bergus:
Instead of adding a separate `catch`/`catchP`/`fail` function, it might be more wise to support the second argument to `then` directly. Maybe under a separate name (`then2`, `biThen`, `thenP`, `thenCatch`? I'm bad at naming), so that one can do
```js
const example = R.compose(R.thenX(logWith("Hooray"),
//                          ^^^^^
                                  logWith("Oops")),
                          R.then(mightThrowOrReject),
                          willReturnPromise)
```
Then everyone who needed a  `catch` can use `myCatch = R.thenX(null)` (equivalent to `thenX(id)`).

Bradcomp:
To sum up the current status of this PR:
* `then` has been created, and tests are written.  It would be nice if somebody could look at the docs for it.
* I have created a helper for dealing with rejected promises named `fail`.  There is dispute over the name, however `catch` is not a viable option due to a conflict with the native keyword.  
* There was discussion on deprecating Node 0.10 support and switching to native Promises for testing.  While I think this is a good idea, it seems like something that would be better as a stand alone PR where the virtues could be discussed separately from here.
* @bergus suggested a `thenX` (may I suggest `thenOr` or `thenOrElse` - param order may be an issue) that lets the user provide both a success and error handler.  Would this be a better option than `fail` or something in addition to `fail`?  

Is there anything I can do to help move this along?

slavaGanzin:
> I get 
Uncaught SyntaxError: Unexpected token catch

You can always use anonymous function and named export: 
```
module.exports.catch = _curry(function () {...
```

> @bergus suggested a thenX (may I suggest thenOr or thenOrElse - param order may be an issue) that lets the user provide both a success and error handler. Would this be a better option than fail or something in addition to fail?

Please stop inventing new names, remember Ocam's razor.

And. please, review:
https://pouchdb.com/2015/05/18/we-have-a-problem-with-promises.html
(especially "Advanced mistake #2: catch() isn't exactly like then(null, ...)" part)

Using .then(res, rej) will return you to callback hell again. It's inconsistency with modern promises and just plain madness.

bergus:
@slavaGanzin 
> Please stop inventing new names

Well, since Ramda does not like variadic functions we *do* have to invent a new name if we want to provide both unary and binary `then`.
And we'd have to invent a name anyway if we want to support error handling, because while using `R.catch` might work, `import {catch} from "ramda"` does not.

> And. please, review: https://pouchdb.com/2015/05/18/we-have-a-problem-with-promises.html
 (especially "Advanced mistake #2: catch() isn't exactly like then(null, ...)" part)

I know that article and the mistakes it contains. The specific one you're citing is total rubbish - `catch` *is exactly* like `then(null, …)` (by spec). What they *mean* is that [`.then(…).catch(…)` is not at all like `.then(…, …)`](http://stackoverflow.com/q/24662289/1048572) - a possibly not well-enough known fact, but with very different consequences.

> Using `.then(res, rej)` will return you to callback hell again. It's inconsistency with modern promises

Huh? `.then(onfulfill, onreject)` might be considered an advanced usage of promises (or a beginner mistake), but it has nothing to do with callback hell and there is no inconsistency. If you think otherwise, please elaborate.

Bradcomp:
I forgot one item on my list up above: 

* Type signatures: Do they seem correct and understandable?

bergus:
Ah, good call @Bradcomp. It looks like you could improve them, usually promises are typed like `Either` with the failure type first - `Promise error value`. So it should be
```
fail	 
@sig (e -> a) -> Promise e a -> Promise () a
@sig (e -> Promise f a) -> Promise e a -> Promise f a

then
@sig (a -> b) -> Promise e a -> Promise e b
@sig (a -> Promise e b) -> Promise e a -> Promise e b
```
Or if you want to allow union types and not force the callback to have the expected types, you could further complicate it to
```
fail	 
@sig (e -> a) -> Promise e a -> Promise () a
@sig (e -> Promise f b) -> Promise e a -> Promise f a|b

then
@sig (a -> b) -> Promise e a -> Promise e b
@sig (a -> Promise f b) -> Promise e a -> Promise e|f b
```
(I'm glad that I don't have to provide an accurate signature for `thenOr` yet :-))

slavaGanzin:
> And we'd have to invent a name anyway if we want to support error handling, because while using R.catch might work, import {catch} from "ramda" does not.

Well you catch me. Static imports. I agree. Plain ```then(success, reject)``` would be best decision (In my opinion).

buzzdecafe:
i like @bergus 's suggestion of `then2` (or whatever) that takes both success and fail handlers. I think it would be better to take the fail case first for currying, i expect that would typically be more general purpose than the success case.

Bradcomp:
> if so, lets use native promises and remove 0.10 from travis config

See https://github.com/ramda/ramda/pull/1956 for this change 😃 .  If that goes through I will update my tests for this PR.

>  It looks like you could improve them, usually promises are typed like `Either` with the failure type first.

Can do!  My only concern was that in some sense I view Promises as flipped due to the order of handlers being success, fail - but see below.

> i like @bergus 's suggestion of `then2`

Would this mean two functions, `then` and `then2`? Or all three? I'd prefer just two functions, as `fail` can be trivially derived as `then2(f, identity)`.  Also, I will go ahead and call it `then2` as a working title, but am open to suggestions.  

> I think it would be better to take the fail case first for currying

I agree, and this would also make me feel better about swapping the type signature.

buzzdecafe:
> Would this mean two functions, `then` and `then2`?

Good question; I prefer *one* function such as `fn_to_be_named_later :: (Any -> Any) -> (Any -> Any) -> Promise` with the failure case first.

Bradcomp:
I can go for that, with the (unfortunate) caveat that I wouldn't want it to be named `then` because there would be too much possibility for confusion.

Having a single function also moves back towards stepping away from Promises as a whole.

buzzdecafe:
i agree, should we go down the path i am suggesting, `then` would be a bad name

CrossEye:
This is reminding me a lot of [`bimap`][bi].  But as always with this stuff, my understanding is on fairly shaky grounds.  @scott-christopher, am I close to on-target here?


  [bi]: https://hackage.haskell.org/package/bifunctors-5/docs/Data-Bifunctor.html#v:bimap

scott-christopher:
> @CrossEye This is reminding me a lot of bimap.

Yep, this is a good candidate for `bimap`. With [Bifunctors](https://github.com/fantasyland/fantasy-land#bifunctor) now specified in FL, it might make sense to introduce `R.bimap` for dispatching to the FL method along with built-in support for `Promise e a` types.

Introducing `R.lmap :: Bimap f => (a -> c) -> f a b -> f c b` would then cover the `catch`/`fail` use-case too.

The caveat here being that we're still stuck with the overloaded `then` that collapses nested `Promise` instances which would affect the right-hand side of `bimap` too. If we're happy with that caveat (which I would be for the sake of practicality, providing it is clearly documented), then there's an argument that built-in support for `Promise` instances should be extended to `R.map` and `R.chain`. If we're not content with that caveat, then it's probably not best to make use of `bimap` here either.

bergus:
@CrossEye It's only superficially similar to `bimap`. `catchThen`/`thenOrElse`/`thenCatch` (thanks for filling in the `X`) is actually more like a `biChain` - it's signature is basically
```
then :: (e -> Promise f b) -> (a -> Promise f b) -> Promise e a -> Promise f b
```
(if we ignore plain values, exceptions, thenables, and the unions in the result type)

The problem with / limitation of `bimap` is that it does not allow to actually handle errors like `catch` does, it only allows to transform them to different errors. This is sometimes helpful as well, but not always generic enough:
```js
function bimap(onerror, onsuccess, promise) {
    return promise.then(R.compose(Promise.of, onsuccess), R.compose(Promise.reject, onerror);
}
```
But yes, I would like to see `R.bimap` in addition to `R.thenOrElse` - maybe briancavalier/creed#33 is also relevant here.

CrossEye:
Two of the smartest folks to ever post in these discussions are disagreeing.  Are you going to make me call on @joneshf?  :smile:

I will have to read up on `biChain`, which I haven't even seen.

CrossEye:
Two of the smartest folks to ever post in these discussions are disagreeing.  Are you going to make me call on @joneshf?  :smile:

I will have to read up on `biChain`, which I haven't even seen.

scott-christopher:
> The problem with / limitation of bimap is that it does not allow to actually handle errors like catch does, it only allows to transform them to different errors.

This is correct (and no disagreement here 😄 ). 

To throw a couple other names in the mix for the `catch` case: `handleP`/`recoverP`? These could be implemented via their generic alternatives `handlePWith`/`recoverPWith`.

```hs
handleP,     recoverP     :: (e -> a) -> Promise e a -> Promise Unit a
handlePWith, recoverPWith :: (e -> Promise e' a) -> Promise e a -> Promise e' a
```

And I'm not sure what `Bichain` would look like either, so I'd definitely be interested to see more info about it too.

bergus:
@CrossEye 
> I will have to read up on `biChain`, which I haven't even seen.

Which is probably because I made the name up, there is no authoritative source for it. The only other occurence I found is in [this `Either` library](https://github.com/jethrolarson/union-type-either#bichain--a---either-b---b---either-c---either-a-b---either-c). Also there is some talk about `Bimonad`s.
The idea is that it compares to `chain` like `bimap` compares to `map`. It's type is

    bichain :: (a -> m c d) -> (b -> m c d) -> m a b -> m c d

i.e. it is the cleaned-up monadic interface of `then`. Using `of` and `reject` as helpers, you can derive anything else from it (notably `chain`, `bimap` and `catch`). In contrast, it cannot be derived from any combination of those.

CrossEye:
@bergus:

I did figure out what it had to be, but I didn't think any further than that.  This makes sense.

CrossEye:
How did we end up dropping this?

I would like to merge this.  What do people think?

buzzdecafe:
so much api discussion left it in a state of limbo. i think the tests need to be tweaked to use native Promise, which also means dropping support for old node in travis.yml

although i am a bit leery of extending promise support further, it's not a strong objection

CrossEye:
@buzzdecafe: 
> i am a bit leery of extending promise support further, it's not a strong objection

I think of this as allowing us to *remove* specific support for Promises.  Yes, because Promises have become central to the language we have this wrapper function which is Promise-specific, but then we could remove all other Promise code, and never consider adding in any other.

MikaAK:
I like @CrossEye's thinking

Bradcomp:
> so much api discussion left it in a state of limbo.
This PR currently has `then` and `fail`. Is that where we want to go? I am fairly ambivalent...

> i think the tests need to be tweaked to use native Promise, which also means dropping support for old node in travis.yml

https://github.com/ramda/ramda/pull/1956

src/fail.js
```diff
@@ -0,0 +1,38 @@
...
+ * Returns the result of applying the onFailure function to the value inside
+ * a failed promise. This is useful for handling rejected promises
+ * inside function compositions.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig (b -> c) -> (Promise a b) -> (Promise c d)
+ * @sig (b -> (Promise c d)) -> (Promise a b) -> (Promise c d)
+ * @param {Function} onFailure The function to apply.  Can return a value or a promise.
```
kedashoe: > `apply.  Can`

extra space?

src/fail.js
```diff
@@ -0,0 +1,38 @@
...
+ * @param {Promise} p
+ * @return {Promise} The result of calling `p.then(null, onFailure)`
+ * @see R.then
+ * @example
+ *
+ *      var failedFetch = (id) => Promise.reject('bad ID');
+ *      var useDefault = () => ({ firstName: 'Bob', lastName: 'Loblaw' })
+ *
+ *      //recoverFromFailure :: String -> Promise [{firstName, lastName}]
```
kedashoe: Is something returning an array here? Or I'm just not understanding the syntax?

src/then.js
```diff
@@ -0,0 +1,37 @@
...
+ * @sig (a -> (Promise c d)) -> (Promise a b) -> (Promise c d)
+ * @param {Function} onSuccess The function to apply.  Can return a value or a promise.
+ * @param {Promise} p
+ * @return {Promise} The result of calling `p.then(onSuccess)`
+ * @see R.fail
+ * @example
+ *
+ *      var makeQuery = (email) => ({ query: { email }});
+ *
+ *      //getMemberName :: String -> Promise [{firstName, lastName}]
```
kedashoe: same as :arrow_up:

src/then.js
```diff
@@ -0,0 +1,37 @@
...
+ * Returns the result of applying the onSuccess function to the value inside
+ * a successfully resolved promise. This is useful for working with promises
+ * inside function compositions.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig (a -> c) -> (Promise a b) -> (Promise c d)
+ * @sig (a -> (Promise c d)) -> (Promise a b) -> (Promise c d)
+ * @param {Function} onSuccess The function to apply.  Can return a value or a promise.
```
kedashoe: :arrow_up:

src/then.js
```diff
@@ -0,0 +1,37 @@
...
+ *
+ *      //getMemberName :: String -> Promise [{firstName, lastName}]
+ *      var getMemberName = R.pipe(
+ *        makeQuery,
+ *        fetchMember,
+ *        R.then(R.pick(['firstName', 'lastName']))
+ *      );
+ */
+module.exports = _curry2(function then(f, p) {
+  if (p == null || !_isFunction(p.then)) {
```
kedashoe: should we define `_assertPromise` for these functions to use? Not really to stay DRY, just to be explicit. I'd prefer the error to yell at me about promises "`then` expected a promise, received ' + toString(p)" or what have you.

buzzdecafe: i like that idea

src/fail.js
Bradcomp: Heh. I've always put 2 spaces after a period... I'll fix it.

CrossEye: Yeah, me too.  I know it's not the most common choice nowadays, but it's so hard to break.

...

> `me too.  I know`
> `......   ^--- see?`

CrossEye:
OK, we dropped this again.  Anyone object to my merging it as is?

kedashoe:
> OK, we dropped this again. Anyone object to my merging it as is?

I think there are a couple comments in the source that haven't been addressed? Ping @Bradcomp

Bradcomp:
Hey, sorry! I think I started to work on this, but I might need to just create a new branch no that we've changed the structure and everything. Ok if I open a new PR which includes the changes?

kedashoe:
> Ok if I open a new PR which includes the changes?

:+1:

CrossEye:
Strange that GitHub is reporting that it can simply be merged.  I forgot just how old this is.

@Bradcomp: I would love to see a new version of this.

Bradcomp:
:sunglasses:  I will work on it, probably over the weekend!

CrossEye:
As @buzzdecafe said, above,

> ((:bell:)) let's not let this one die

Bradcomp:
Sorry! Life and holidays and work and all that... :(

CrossEye:
I am never one to talk about falling behind here!  I just didn't want to lose this.  It came up recently in StackOverflow comments.

MikaAK:
Any news on this one @CrossEye @davidchambers ?

Bradcomp:
:(

I'm really sorry I dropped the ball on this. OS work has taken a back seat for the past few months. 

I don't think I will be able to get around to this for a while, so feel free to take it.

CrossEye:
OK.  Maybe it can get into the long-delayed version I'm trying to get to.

Bradcomp:
@CrossEye I am at a point where my Open Source time has increased. Want me to take this back?

CrossEye:
@Bradcomp: sure.  I thought my time was also increasing, but my wife seems to be able to fill it.

PR #1908, opened 2016-09-17
Add chainRec function from Fantasy Land

### LINKED ISSUE #1902: Add chainRec function from Fantasy Land
As there is `chain` function in ramda I thought it would be nice also have `chainRec` from [ChainRec](https://github.com/fantasyland/fantasy-land#chainrec). Something like [this](https://gist.github.com/safareli/ea45b45bed99f32bae7002e19dcf13c2) should be default implementation for Array. If there is an interest in it, I would try to create a PR

safareli:
Implementation adopted from [sanctuary-type-classes/pull/2](https://github.com/sanctuary-js/sanctuary-type-classes/pull/2)

Fixes: #1902 
Blocker: #1900

### DIFF
--- a/src/chainRec.js
+++ b/src/chainRec.js
@@ -25,9 +25,9 @@ var _curry3 = require('./internal/_curry3');
  *
  */
 module.exports = _curry3(function _chainRec(chainRec, fn, i) {
-  if (chainRec != null && typeof chainRec.chainRec === 'function') {
+  if (typeof chainRec.chainRec === 'function') {
     return chainRec.chainRec(fn, i);
-  } else if (chainRec != null && chainRec.constructor != null && typeof chainRec.constructor.chainRec === 'function') {
+  } else if (chainRec.constructor != null && typeof chainRec.constructor.chainRec === 'function') {
     return chainRec.constructor.chainRec(fn, i);
   } else {
     return _arrayChainRec(fn, i);
--- a/src/chainRec.js
+++ b/src/chainRec.js
@@ -25,10 +25,10 @@ var _curry3 = require('./internal/_curry3');
  *
  */
 module.exports = _curry3(function _chainRec(chainRec, fn, i) {
-  if (typeof chainRec.chainRec === 'function') {
-    return chainRec.chainRec(fn, i);
-  } else if (chainRec.constructor != null && typeof chainRec.constructor.chainRec === 'function') {
-    return chainRec.constructor.chainRec(fn, i);
+  if (typeof chainRec['fantasy-land/chainRec'] === 'function') {
+    return chainRec['fantasy-land/chainRec'](fn, i);
+  } else if (chainRec.constructor != null && typeof chainRec.constructor['fantasy-land/chainRec'] === 'function') {
+    return chainRec.constructor['fantasy-land/chainRec'](fn, i);
   } else {
     return _arrayChainRec(fn, i);
   }
--- a/src/chainRec.js
+++ b/src/chainRec.js
@@ -1,0 +1,35 @@
+var _arrayChainRec = require('./internal/_arrayChainRec');
+var _curry3 = require('./internal/_curry3');
+
+
+/**
+ * `chainRec` does stack safe recurtion on List
+ *
+ * Dispatches to the `chainRec` method of first argument, if present,
+ * according to the [FantasyLand ChainRec spec](https://github.com/fantasyland/fantasy-land#chainrec).
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig ChainRec m => m -> ((a -> c, b -> c, a) -> m c) -> a -> m b
+ * @param {ChainRec} object or type which is ChainRec
+ * @param {Function} fn The function to compute next step of recursion
+ * @param {Any} initial value for computation
+ * @return {Array} The result of recursive computation
+ * @example
+ *
+ *      R.chainRec(Array, (next, done, a) => (
+ *        a.length == 2 ? [a + '!', a + '?'].map(done)
+ *                      : [a + 'a', a + 'b'].map(next)
+ *      ), ''); // [ 'aa!', 'aa?', 'ab!', 'ab?', 'ba!', 'ba?', 'bb!', 'bb?' ],
+ *
+ */
+module.exports = _curry3(function _chainRec(chainRec, fn, i) {
+  if (chainRec != null && typeof chainRec.chainRec === 'function') {
+    return chainRec.chainRec(fn, i);
+  } else if (chainRec != null && chainRec.constructor != null && typeof chainRec.constructor.chainRec === 'function') {
+    return chainRec.constructor.chainRec(fn, i);
+  } else {
+    return _arrayChainRec(fn, i);
+  }
+});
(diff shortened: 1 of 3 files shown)


safareli:
Wording and examples might not be the best, suggestions are welcome!

src/chainRec.js
```diff
@@ -0,0 +1,35 @@
...
+ * @example
+ *
+ *      R.chainRec(Array, (next, done, a) => (
+ *        a.length == 2 ? [a + '!', a + '?'].map(done)
+ *                      : [a + 'a', a + 'b'].map(next)
+ *      ), ''); // [ 'aa!', 'aa?', 'ab!', 'ab?', 'ba!', 'ba?', 'bb!', 'bb?' ],
+ *
+ */
+module.exports = _curry3(function _chainRec(chainRec, fn, i) {
+  if (chainRec != null && typeof chainRec.chainRec === 'function') {
```
davidchambers: Why `chainRec != null`? `chainRec` should never be `null` or `undefined`, surely? Having `R.chainRec(null)` behave like `R.chainRec([])` seems strange to me.

src/internal/_arrayChainRec.js
```diff
@@ -0,0 +1,19 @@
+module.exports = (function() {
+  function stepNext(x) { return {value: x, done: false }; }
+  function stepDone(x) { return {value: x, done: true }; }
+
+  return function _chainRec(f, i) {
```
davidchambers: Could we use `x` here rather than `i`? `i` makes me think "index" or "integer", neither of which applies here.

test/chainRec.js
```diff
@@ -0,0 +1,61 @@
...
+        return [done('DONE')]
+      } else {
+        return [next(n - 1)]
+      }
+    }, 100000), ['DONE'])
+  });
+
+  it('dispatches to objects that implement `chainRec`', function() {
+    eq(R.chainRec({ chainRec: function(f, i) { return true; }}, null, null), true);
+  });
```
davidchambers: We should dispatch to `fantasy-land/chainRec` rather than to `chainRec`. We need to wait until #1900 is merged before merging this pull request, I think.

test/chainRec.js
```diff
@@ -0,0 +1,61 @@
...
+    var f = R.chainRec(Array, function(next, done, i) {
+      return [done(i)];
+    });
+    eq(f(10), [10]);
+  });
+
+  it('correctly reports the arity of curried versions', function() {
+    var f = R.chainRec(Array);
+    eq(f.length, 2);
+  });
```
davidchambers: I suggest removing the two tests above. This behaviour is tested in [__invariants.js__][1].


[1]: https://github.com/ramda/ramda/blob/master/test/invariants.js

test/chainRec.js
```diff
@@ -0,0 +1,61 @@
...
+describe('chainRec', function() {
+
+  it('is equivalent to using chain', function() {
+    ArrayChainRecEquivalence(
+      ['aa!', 'aa?', 'ab!', 'ab?', 'ba!', 'ba?', 'bb!', 'bb?'],
+      function isDone(a) { return a.length == 2; },
+      function onDone(a) { return [a + '!', a + '?']; },
+      function onNext(a) { return [a + 'a', a + 'b']; },
+      ''
+    );
```
davidchambers: I :heart: the way you formatted this. :)

test/chainRec.js
```diff
@@ -0,0 +1,61 @@
...
+  });
+
+  it('is stacksafe', function() {
+    eq(R.chainRec(Array, function(next, done, n) {
+      if (n === 0) {
+        return [done('DONE')]
+      } else {
+        return [next(n - 1)]
+      }
```
davidchambers: How about `return n === 0 ? [done('DONE')] : [next(n - 1)];`?

test/chainRec.js
safareli: I was thinking to remove them too, but I have seen them in too many places (tests  of `map|chain|ap`). should they be removed too?

test/chainRec.js
safareli: As currently ramda is not supporting namespaced fantasy land methods I have not done that. Will update it and weight for that PR to be merged.

src/chainRec.js
safareli: agreed

src/internal/_arrayChainRec.js
safareli: `i` stands for `initial`, also it's used in fantasy land spec.
For that reason I think it's better to leave it as is.

davidchambers: Oh, I see. It's good to know what it stands for. `i` is fine, in that case. :)

test/chainRec.js
safareli: And if hay should be removed too, we should create issue/pr

davidchambers: I'd like to remove them, yes. Feel free to open an issue for this.

safareli: #1909

test/chainRec.js
safareli: Fixed

buzzdecafe:
i thought this was merged already? what is the status of this one?

safareli:
[blocked](https://github.com/ramda/ramda/pull/1908#discussion_r79291305) by #1900

buzzdecafe:
thanks

safareli:
I guess it's not blocked any more

src/chainRec.js
```diff
@@ -0,0 +1,35 @@
+var _arrayChainRec = require('./internal/_arrayChainRec');
+var _curry3 = require('./internal/_curry3');
+
+
+/**
+ * `chainRec` does stack safe recursion on List
```
kedashoe: It doesn't look like we use capital "List" in our descriptions, just for the category.. could you change this to "on lists."?

test/chainRec.js
```diff
@@ -0,0 +1,45 @@
...
+      function onDone(a) { return [a]; },
+      function onNext(a) { return a; },
+      [1, [[2, 3], 4], 5]
+    );
+  });
+
+  it('is stacksafe', function() {
+    eq(R.chainRec(Array, function(next, done, n) {
+      return n === 0 ? [done('DONE')] : [next(n - 1)];
+    }, 100000), ['DONE'])
```
kedashoe: I think CI is complaining about a missing semicolon here.

safareli:
Strange, locally tests are passing but on travis it fails

kedashoe:
> Strange, locally tests are passing but on travis it fails

I think https://github.com/ramda/ramda/pull/1928 is to blame, it looks like that was merged after you started this PR. Could you pull down the latest master and rebase? Then add a `chainRec` require to `/index.js`. You can run tests again at that point as well, once you are on latest master it should catch this locally.

CrossEye:
Part of the 2.1 discussion.

safareli:
> Part of the 2.1 discussion.

 
Any links to that?

CrossEye:
@safareli: Internal only, but in a few days I hope we can post something public,

In this case, I simply meant that it's excluded from our `v1.0` path, and `v2.0` will likely be only an implementation modernization release.  After that, we need to consider issues such as this one.

PR #1911, opened 2016-09-18
Fix spelling of JavaScript

ligaz:
The correct spelling of JavaScript is to capitalize both **J** and **S** letters. More info [here]( https://www.quora.com/Should-the-s-in-JavaScript-be-capitalized).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Ramda
 =============
 
-A practical functional library for Javascript programmers.
+A practical functional library for JavaScript programmers.
 
 [![Build Status](https://travis-ci.org/ramda/ramda.svg?branch=master)](https://travis-ci.org/ramda/ramda)
 [![npm module](https://badge.fury.io/js/ramda.svg)](https://www.npmjs.org/package/ramda)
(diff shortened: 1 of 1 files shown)


git-in-my-anus:
But Why?

buzzdecafe:
uh-oh you've stumbled upon one of @CrossEye 's pet peeves. 
🏃

ligaz:
I guess the reason is that I omitted the description. Fixed.

CrossEye:
Not exactly a pet peeve.  But I use "Javascript" for a generic word to represent the family of languages adhering to the various versions of the ECMA-262 specification.  I reserve the trademarked "JavaScript" for the Mozilla implementation, and similarly, although rarely, use "JScript" for the MS version.

So if everyone feels strongly about changing this, feel free.  I will probably continue to write in the same manner, though.

buzzdecafe:
LGTM :dromedary_camel:

anton-rudeshko:
#393 :)

CrossEye:
@anton-rudeshko:

Wow, I'd totally forgotten that we already did this!  :smile:

Issue #1413, opened 2015-09-28
When you pass a string to R.test instead of a regex, it always returns true


albrow:
The following code returns true:

```javascript
R.test('foo', 'bar')
```

[Here it is in the repl](http://bit.ly/1Jz66KK).

This seems like a pretty easy mistake to make, and I think the appropriate behavior is to throw an exception and point out that the first argument is not the correct type.

davidchambers:
Are you happy for us to check the type of the first argument, @CrossEye?

CrossEye:
Not really, but I can't think of anything better.

PR #1912, opened 2016-09-20
optional type checking

davidchambers:
From discussion in #1910 it seems there's at least some interest in using [sanctuary-def](https://github.com/sanctuary-js/sanctuary-def) in Ramda. I'm opening this pull request simply to provide a couple of concrete examples of how this would look.

If we decide to pursue this we'll need to provide a way for users to opt in to (or opt out of) type checking. Presumably this would work as in [Sanctuary](https://sanctuary.js.org/#type-checking), though we'd likely disable type checking by default.

This could continue to work exactly as it does now:

``` javascript
const R = require('ramda');
```

This would enable type checking to be turned on and off:

``` javascript
const {create, env} = require('ramda');

const R = create({checkTypes: true, env: env});
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,9 @@
     "browser_test": "testem ci",
     "postbrowser_test": "npm run posttest"
   },
-  "dependencies": {},
+  "dependencies": {
+    "sanctuary-def": "0.6.x"
+  },
   "devDependencies": {
     "acorn": "0.9.x",
     "benchmark": "~1.0.0",
--- a/scripts/build
+++ b/scripts/build
@@ -122,25 +122,14 @@ var abortIfExportNotLast = R.tap(R.cond([[
 ]]));
 
 //  abortIfDifferentRequireVar :: {*} -> {*}
-var abortIfDifferentRequireVar = R.tap(R.forEach(R.cond([[
-  R.converge(
-    R.complement(R.equals),
-    R.path(['id', 'name']),
-    R.pipe(
-      R.path(['init', 'arguments', '0', 'value']),
-      R.replace(/^[.][/]internal[/]/, './'),
-      R.replace(/^[.]{1,2}[/]/, ''),
-      R.replace(/[.]([a-z])/g, R.pipe(R.nthArg(1), R.toUpper))
-    )
-  ),
-  function(ast) {
+var abortIfDifferentRequireVar = R.tap(R.forEach(function(ast) {
+  var varName = ast.id.name;
+  var depName = ast.init.arguments[0].value;
+  if (/^[.]/.test(depName) && R.last(R.split('/', depName)) !== varName) {
     abort('Dependency declared with different variable name: `' +
-      ast.id.name + '` & `' +
-      ast.init.arguments[0].value +
-      '` in ' + ast.loc.source
-    );
+          varName + '` & `' + depName + '` in ' + ast.loc.source);
   }
-]])));
+}));
 
 //  abortIfEmptyBody :: {*} -> {*}
 var abortIfEmptyBody = R.tap(R.cond([[
@@ -178,8 +167,9 @@ R.pipe(identifierToFilename,
        R.takeWhile(R.both(R.propEq('type', 'VariableDeclaration'),
                           R.pipe(R.prop('declarations'),
                                  R.all(isRequireExpr)))),
-  R.pluck('declarations'),
+       R.pluck('declarations'),
        R.map(R.head),
+       R.reject(R.pathEq(['id', 'name'], '$')),
        abortIfNotSorted,
        abortIfDifferentRequireVar,
        R.map(R.path(['id', 'name'])));
--- a/scripts/template.js
+++ b/scripts/template.js
@@ -1,4 +1,16 @@
-;(function() {
+;(function(f) {
+
+  'use strict';
+
+  if (typeof module === 'object' && typeof module.exports === 'object') {
+    module.exports = f(require('sanctuary-def'));
+  } else if (typeof define === 'function' && define.amd != null) {
+    define(['sanctuary-def'], f);
+  } else {
+    self.R = f(self.sanctuaryDef);
+  }
+
+}(function($) {
 
   'use strict';
 
@@ -7,12 +19,6 @@
 
   /* TEST_ENTRY_POINT */
 
-  if (typeof exports === 'object') {
-    module.exports = R;
-  } else if (typeof define === 'function' && define.amd) {
-    define(function() { return R; });
-  } else {
-    this.R = R;
-  }
+  return R;
 
-}.call(this));
+}));
--- a/src/internal/_def.js
+++ b/src/internal/_def.js
@@ -1,0 +1,3 @@
+var $ = require('sanctuary-def');
+
+module.exports = $.create({checkTypes: true, env: $.env});
--- a/src/internal/_isRegExp.js
+++ b/src/internal/_isRegExp.js
@@ -1,3 +1,0 @@
-module.exports = function _isRegExp(x) {
-  return Object.prototype.toString.call(x) === '[object RegExp]';
-};
--- a/src/match.js
+++ b/src/match.js
@@ -1,4 +1,6 @@
-var _curry2 = require('./internal/_curry2');
+var $ = require('sanctuary-def');
+
+var _def = require('./internal/_def');
 
 
 /**
@@ -20,8 +22,9 @@ var _curry2 = require('./internal/_curry2');
  *
  *      R.match(/([a-z]a)/g, 'bananas'); //=> ['ba', 'na', 'na']
  *      R.match(/a/, 'b'); //=> []
- *      R.match(/a/, null); //=> TypeError: null does not have a method named "match"
  */
-module.exports = _curry2(function match(rx, str) {
-  return str.match(rx) || [];
-});
+module.exports =
+_def('match',
+     {},
+     [$.RegExp, $.String, $.Array($.Any)],
+     function(rx, str) { return str.match(rx) || []; });
--- a/src/test.js
+++ b/src/test.js
@@ -1,7 +1,7 @@
+var $ = require('sanctuary-def');
+
 var _cloneRegExp = require('./internal/_cloneRegExp');
-var _curry2 = require('./internal/_curry2');
-var _isRegExp = require('./internal/_isRegExp');
-var toString = require('./toString');
+var _def = require('./internal/_def');
 
 
 /**
@@ -21,9 +21,8 @@ var toString = require('./toString');
  *      R.test(/^x/, 'xyz'); //=> true
  *      R.test(/^y/, 'xyz'); //=> false
  */
-module.exports = _curry2(function test(pattern, str) {
-  if (!_isRegExp(pattern)) {
-    throw new TypeError('‘test’ requires a value of type RegExp as its first argument; received ' + toString(pattern));
-  }
-  return _cloneRegExp(pattern).test(str);
-});
+module.exports =
+_def('test',
+     {},
+     [$.RegExp, $.String, $.Boolean],
+     function test(pattern, str) { return _cloneRegExp(pattern).test(str); });
--- a/test/invariants.js
+++ b/test/invariants.js
@@ -4,28 +4,31 @@ var eq = require('./shared/eq');
 
 describe('invariants', function() {
 
+  //  toTest :: [String]
+  var toTest = R.without(['match', 'test'], R.keys(R.filter(R.is(Function), R)));
+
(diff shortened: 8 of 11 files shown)


buzzdecafe:
Here is a pros-and-cons table. I'll keep updating as discussion proceeds. We can then evaluate that against the project's goals (whatever those may be ...)

| PRO | CON |
|------|-------|
| Nice error messages | Heavier file size |
| Type-checking | Perf hit |
| More faithful to ML style and priorities | Steeper learning curve for JavaScript developers |

davidchambers:
> Steeper learning curve for JavaScript developers

Can you explain this one, @buzzdecafe? I don't see how this change would make Ramda less approachable for newcomers. Ramda users needn't know whether `R.test` is defined via `def` or `R.curry`.

buzzdecafe:
perhaps I am wrong, but doesn't this require the user to buy into sanctuary-def style type definitions in order to understand some of the error messages? Take your example for `concat`:

```
concat({}, {});
// ! TypeError: Type-class constraint violation
//
//   concat :: Semigroup a => a -> a -> a
//             ^^^^^^^^^^^    ^
```

"Semigroup? WTF is a semigroup?" quoth the learner at the bottom of the curve. The current ramda error in this case is '{} does not have a method named "concat"'

CrossEye:
> | PRO | CON |
> |------|-------|
> | Fails as soon as possible | Fails as soon as possible |

:smile:

Minor, but definite, con: adding first external dependency to Ramda.

And of course along with that is a pro: the ability to remove certain current internal type-checking.

Pro: shared maintenance of important facet of the library: not just Ramda team, but Sanctuary folks and others using it.  (Obviously there's a lot of overlap, but this is a powerful feature of right-sizing modules.)

davidchambers:
> Minor, but definite, con: adding first external dependency to Ramda.

I agree that this is minor, but I believe it will take significant effort to support this. Our build script, which has served us well for almost two years, assumes that all imports are internal (I had to work around this when creating this patch). If we're to introduce a dependency it would be a good opportunity to retire the build script and make the changes necessary to support tree shaking.

> Pro: shared maintenance of important facet of the library

This would be great.

It's worth noting that with type checking disabled `def` is currently equivalent to `R.curry`, but this will no longer be the case if #1843 is merged or if the change proposed in sanctuary-js/sanctuary-def#73 is made. So we'd either need to agree on placeholder behaviour, or have `$.create` take `respectPlaceholder :: Boolean` as part of the options record, so Ramda could disable this feature.

buzzdecafe:
> I agree that this is minor, but I believe it will take significant effort to support this. 

Rather than pour more energy into the custom build script, we would probably be better served to port to ES6 and use rollup (or something like it).

davidchambers:
> > I agree that this is minor, but I believe it will take significant effort to support this.
>
> Rather than pour more energy into the custom build script, we would probably be better served to port to ES6 and use rollup (or something like it).

Yes, that's what I'm suggesting. ;)

paldepind:
@buzzdecafe

> Working against the grain of the language

I don't see this. As evident by the type signatures in our documentation we are already restricting ourselves to an API expreseable by an ML-like type system. I.e.we are already refraining ourselves from JavaScripty things like ad hoc overloading and variadic functions.

The only thing this proposal would change is better error messages and that hardly goes against the grain of the language.

CrossEye:
@paldepind:
> The only thing this proposal would change is better error messages and that hardly goes against the grain of the language.

Well, even that might be up for debate! :smiling_imp: 

But unless I'm missing something important (always a possibility) this goes beyond error messages.  This introduces run-time type checking.  Ramda so far hews closer to the underlying language: it tries to do the best with what it's given.  This change would mean trying to catch discordant inputs earlier and throwing errors when they're detected.  It's the difference between:

```shell
> add(3, '5');
8
```

and
```shell
> add(3, '5');
Invalid value

add :: FiniteNumber -> FiniteNumber -> FiniteNumber
                       ^^^^^^^^^^^^
                            1

1)  "5" :: String

The value at position 1 is not a member of ‘FiniteNumber’.
```

Whether we want to do this or not is the purpose of this discussion, but let's not downplay how big a change this is to the core ideas of Ramda.  So far, Ramda has promoted itself more as a library for power users with little hand-holding.  This suggestion changes that notion significantly.

paldepind:
`add` seems to be a special case. I see disallowing coercion as a small added benefit. In my opinion coercion isn't part of the grain of the language. It's one of the mistakes we have to live with (along with workarounds such as `===`).

buzzdecafe:
> I don't see this. As evident by the type signatures in our documentation we are already restricting ourselves to an API expreseable by an ML-like type system.

I think it's fair to call bolting static(-ish) type checking onto a dynamic language "working against the grain."

davidchambers:
```javascript
> R.add([[['41']]], true)
42
```

Ah, yes, obviously. :japanese_ogre:

Seriously, though, we could define the type of `R.add` as `[$.Any, $.Any, $.Number]` if we wanted to. In fact, we could do this for all functions as a starting point. Then when something like #1413 is brought to our attention and we decide that type checking would be useful, we could simply replace the `$.Any` corresponding to the problematic argument with `$.RegExp` or whatever the more specific type is. Essentially we'd be doing the same type checks we already do, but in a more consistent manner. The price of consistency is verbosity, though, as defining functions with `def` requires a bit more boilerplate than defining functions with `R.curry`. Only a fraction of the boilerplate introduced by JSDoc, of course. ;)

CrossEye:
@paldepind: I wasn't thinking about coercion so much as the fact that part of the grain of the language, at least in the environments I've worked in, has to do with the user being able to work with odd and inconsistent data.  This is another, slightly more complex, example:

```js
const numericGrade = grade => ({A: 95, B: 85, C: 75, D: 65})[grade] || 0;

R.reduce(
  (total, score) => total + (is(String, score) ? numericGrade(score) : score), 
  0,
  [87, 93, 78, 'B', 82, 'A']
); //=> 520
```

If we switched to sanctuary-def, this would throw a type-variable constraint violation, since Sanctuary doesn't know of a common ancestor type for `87` and `'B'`.  If this is the kind of data you need to work with, it would be hard to find a sanctuary-hardened Ramda function that even *could* help convert it, I believe.

I deal with problems like this regularly.  My data is poorly scrubbed if at all.  One of the first things I need to do is to normalize it.  I would hate it if my toolkit refused to look at it because, well, it hasn't been normalized yet!  :wink:

davidchambers:
> I would hate it if my toolkit refused to look at it because, well, it hasn't been normalized yet!

Fair point. It turns out there's another knob we could turn to make sanctuary-def more permissive. Were we to add [`$.Any`][1] to the environment we could work with heterogeneous collections:

```javascript
const $ = require('sanctuary-def');

const a = $.TypeVariable('a');

$.create({checkTypes: true, env: $.env})('I', {}, [a, a], x => x)([87, 93, 78, 'B', 82, 'A']);
// ! TypeError: Type-variable constraint violation
//
//   I :: a -> a
//        ^
//        1
//
//   1)  [87, 93, 78, "B", 82, "A"] :: Array ???
//
//   Since there is no type of which all the above values are members, the type-variable constraint has been violated.

$.create({checkTypes: true, env: $.env.concat([$.Any])})('I', {}, [a, a], x => x)([87, 93, 78, 'B', 82, 'A']);
// => [87, 93, 78, 'B', 82, 'A']
```

This likely has unfortunate consequences, but I believe it is possible for Ramda and Sanctuary to share a dependency for currying and type checking even though the libraries have different ideas as to the appropriate level of strictness.


[1]: https://github.com/sanctuary-js/sanctuary-def#any

paldepind:
@buzzdecafe 

> I think it's fair to call bolting static(-ish) type checking onto a dynamic language "working against the grain."

Also when the library in question already is completely static(-ish) and "non-dymanic"? Anyway, I agree that viewed as a an approximation to static typing this is rather crude. But viewed as a systematic way to provide great error messages I think it fares better.

@CrossEye 

> I wasn't thinking about coercion so much as the fact that part of the grain of the language, at least in the environments I've worked in, has to do with the user being able to work with odd and inconsistent data.

Sorry, I misunderstood you. I find your new example much more convincing than the coercion one.

slavaGanzin:
Hello David and everyone

Just to point to alternatives: it would be great to have typing as annotations(which can be dropped in prod or so), like:
https://github.com/disnet/contracts.js

kedashoe:
I think it would be awesome to be able to plug sanctuary def into ramda to turn on run-time type checking, but I'm not sure about ramda depending on it.

For one, I consider this to have already been done: https://github.com/sanctuary-js/sanctuary. If ramda depends on sanctuary-def, what will be the difference between sanctuary and ramda? Transducers and variadics? If we decide *not* to depend on sanctuary-def but to create a new repo, ramda-sanctuary-def.. will we really maintain that and sanctuary? They are too similar, right? So if sanctuary and ramda-sanctuary-def are too similar, aren't sanctuary and ramda with sanctuary-def baked in? Even if sanctuary is better in many ways, it is more interesting to me to have libraries with different focuses that do some different things.

Another reason is library size. For example, @MarkusPfundstein recently wrote https://github.com/MarkusPfundstein/pmatch-js, which depends on Ramda and uses `all`, `keys`, and `zip`. Now, I think it is inevitable that ramda will support tree shaking, and we could also easily envision packages such as `ramda.all`, `ramda.zip` etc ala lodash. So in that scenario this library would be bringing in say 100 lines of code. If sanctuary-def is depended on, that becomes 1,000ish lines of code for just a couple functions.

Some type checking I would be more interested in exploring:

- officially supporting flow
- officially supporting typescript
- some form of light weight run time type checking (https://github.com/gcanti/babel-plugin-tcomb? ad-hoc?)
- all of the above

I love the stuff that is happening with sanctuary, but for me I would rather ramda not depend on sanctuary-def.

davidchambers:
Thanks for the thoughtful comment, Kevin. :)

CrossEye:
Thank you, @kedashoe, great stuff!

> Some type checking I would be more interested in exploring:
> 
> - officially supporting flow
> - officially supporting typescript

I would definitely be in favor of these... *if* 

  * there is someone who wants to take them on.  I haven't been convinced by either TS or Flow, and am not interested in diving in deeply enough to either of them to learn how to write and maintain their configuration.  I would also worry about whether those type systems are robust enough to capture everything the Ramda supports and how we might deal with cases where it's not.

  * we can do this without substantially slowing down our release process.  It's not that we're necessarily organized about it now.  But when we decide to cut a release, we can usually have it out in a matter of hours, certainly within a few days.  I would hate it if it became, "well, give me two more weeks to update the Flow config file."  I've only seen such things inside big companies, but in similar circumstances as this: one project trying to synchronize a public interface to another one on release schedules.

> - some form of light weight run time type checking (https://github.com/gcanti/babel-plugin-tcomb? ad-hoc?)

I'd have to be convinced about this, for many of the same reasons I'm not convinced that sanctuary-def fits with Ramda.

slavaGanzin:
Parsing doc string type annotations was implemented:
https://github.com/raine/ramda-t

Maybe it would be a good start?

tycho01:
> I would also worry about whether those [TS/Flow] type systems are robust enough to capture everything the Ramda supports

Maybe I could help answer that from experience working on the TS typings. The gist is:
- TS is still unable to properly infer return types in quite some cases, whether due to lack of features and/or presence of bugs.
- Notably placeholders can as of yet not be dealt with in a practical fashion in TS, though to my understanding these are not well liked here either.
From what I've heard Flow does better in some areas, worse in others.

> and how we might deal with cases where it's not.

The worst-case fallback is essentially the user needs to manually annotate return types. In these cases it fails to add great value, though at least it doesn't get in the way either.

You'd ask when it does get in the way, and I should admit quite some tests are still failing currently, mostly due to an issue with capturing generics (-> `pipe(identity)` loses its generics) that sort of breaks currying/composition as of now, fix WIP [here](https://github.com/Microsoft/TypeScript/issues/9949).

I respect the objections to a merge though, as per our previous discussion.

CrossEye:
@tycho01:

Thanks.  That's great feedback.  It's quite helpful to hear from someone who has actually tried this stuff!

>[P]laceholders can as of yet not be dealt with in a practical fashion in TS, though to my understanding these are not well liked here either.

They are controversial.  Some people seem to really like them, others, not so much.  Personally, I love placeholders in languages where they're built in for pattern matching.  But I find them sorely lacking here; I'd really like to have a better solution.

Rafi993:
If you really want type systems why not try with typescript. Using fixed types does improve performance in V8 (hidden classes)

Rafi993:
Plus you could avoid worry of getting over bloated

davidchambers:
Those who want a Ramda-like library with run-time type checking (optional or otherwise) can use [Sanctuary][1] itself. Ramda needn't have this capability too.

PR #1916, opened 2016-09-23
Document R.ap as S Combinator

MikaAK:
Ramda `ap` function can be used as S Combinator but I found no reference to it anywhere

``` javascript
R.ap(R.add, R.inc)(100) === R.add(100, R.inc(100))
```

### DIFF
--- a/src/ap.js
+++ b/src/ap.js
@@ -15,6 +15,7 @@ var map = require('./map');
  * @since v0.3.0
  * @category Function
  * @sig [a -> b] -> [a] -> [b]
+ * @sig (a -> b -> c) -> (a -> b) -> a -> c
  * @sig Apply f => f (a -> b) -> f a -> f b
  * @param {Array} fns An array of functions
  * @param {Array} vs An array of values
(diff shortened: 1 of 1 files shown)


davidchambers:
```haskell
ap :: Apply f => f (a -> b) -> f a -> f b
```

Replace `Apply f => f` with `Function x`:

```haskell
ap :: Function x (a -> b) -> Function x a -> Function x b
```

Use `* -> *` sugar for `Function * *`:

```haskell
ap :: (x -> a -> b) -> (x -> a) -> (x -> b)
```

Rename type variables:

```haskell
ap :: (a -> b -> c) -> (a -> b) -> (a -> c)
```

The Function-specific signature, like the Array-specific signature, is a clearer but less precise type signature. My preference is for each function to have one (canonical) signature, and for the usage examples to indicate how it applies to different types.

buzzdecafe:
@davidchambers so if @MikaAK removes the function-specific `@sig` from the pr, would you be in favor of merging? That is, just the example text? i would. :cow2:

davidchambers:
I would, but we can do better than `R.ap(R.add, R.inc)(5)`. The problem with this example is that `R.add` is commutative, so the order in which the arguments are provided is unclear. I'd prefer to see `R.ap(R.concat, R.toUpper)('Ramda')`.

buzzdecafe:
> `R.ap(R.concat, R.toUpper)('Ramda')`

Nice example! :wink: I'll buy that.  @MikaAK are you willing to make these changes, please?

MikaAK:
Sure will do when I get a quick sec!

MikaAK:
Moved to #2174 @davidchambers

Issue #1893, opened 2016-09-07
`forEach` to iterate over an object


centsent:
In my opinion, it will be better if `forEach` supports object iteration (like `lodash`'s forEach) .

BTW, I can open a PR to do this.

davidchambers:
We could do one or both of the following:

  - update `R.forEach` to iterate an object's values; and/or
  - define `R.forEachObjIndexed` which would provide access to keys as well as values.

Ramda avoids ad hoc polymorphism, so having `R.forEach` provide one argument to the iteratee in some cases and *two* in other cases is not an option.

Until one or both of the above changes is made, one can of course use [`R.map`][1] or [`R.mapObjIndexed`][2] and ignore the result.


[1]: http://ramdajs.com/0.22.1/docs/#map
[2]: http://ramdajs.com/0.22.1/docs/#mapObjIndexed

buzzdecafe:
> one can of course use `R.map` or `R.mapObjIndexed` and ignore the result.

:scream:

mjgpy3:
I think if this is really needed `R.toPairs(obj).forEach(...)` should suffice.

It's hard to reconcile `forEach` with functional programming values.

buzzdecafe:
> It's hard to reconcile `forEach` with functional programming values.

I agree; `forEach` is there for side effects. So it is a bit of a red-headed stepchild in the lib.

centsent:
> It's hard to reconcile `forEach` with functional programming values.

Agree, close this issue or I implement `R. forEachObjIndexed `?

davidchambers:
Since we have `R.forEach`, I think we should add `R.forEachObjIndexed`. Would you like to submit a pull request?

ricejasonf:
Both `R.reduce` and `R.forEach` should call the "iterator" function with a pair for each key/value when an object is provided.

Just to reference another functional programming library, Boost.Hana has a `Foldable` concept that, for "objects", it folds on pairs (key/value).

I could try a pull request for this suggestion if you find it acceptable.

buzzdecafe:
@ricejasonf  `reduce` is not well-defined for objects in javascript. This is a subject that has come up repeatedly around here. We'v come close to folding objects, but they've never made it in. The problem boils down to this: `reduce(f, acc, {a: 1, b: 2})` should be equal to `reduce(f, acc, {b: 2, a: 1})`. Easy to formulate the problem, but finding the correct answer -- if one exists -- is not so simple.

[...]

PR #1917, opened 2016-09-24
add R.forEachObjIndexed to iterate over an object

centsent:
Discussion here: #1893

### DIFF
--- a/src/forEachObjIndexed.js
+++ b/src/forEachObjIndexed.js
@@ -11,7 +11,7 @@ var keys = require('./keys');
  * @func
  * @memberOf R
  * @category Object
- * @sig f => (a -> b) -> f a -> f b
+ * @sig ((a, String, StrMap a) -> Any) -> StrMap a -> StrMap a
  * @param {Function} fn The function to invoke. Receives three argument, `value`, `key`, `obj`.
  * @param {Object} obj The object to iterate over.
  * @return {Object} The original object.
--- a/src/forEachObjIndexed.js
+++ b/src/forEachObjIndexed.js
@@ -1,0 +1,38 @@
+var _checkForMethod = require('./internal/_checkForMethod');
+var _curry2 = require('./internal/_curry2');
+var keys = require('./keys');
+
+
+/**
+ * Iterate over an input `object`, calling a provided function `fn` for each
+ * key and value in the object.
+ *
+ * `fn` receives three argument: *(value, key, obj)*.
+ *
+ * Dispatches to the `forEachObjIndexed` method of the second argument, if present.
+ *
+ * @func
+ * @memberOf R
+ * @category Object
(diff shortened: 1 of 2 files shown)


src/forEachObjIndexed.js
```diff
@@ -0,0 +1,38 @@
...
+ * @return {Object} The original object.
+ * @example
+ *
+ *      var printKeyConcatValue = (value, key) => console.log(key + ':' + value);
+ *      R.forEachObjIndexed(printKeyConcatValue, {x: 1, y: 2}); //=> {x: 1, y: 2}
+ *      // logs x:1
+ *      // logs y:2
+ * @symb R.forEachObjIndexed(f, {x: a, y: b}) = {x: a, y: b}
+ */
+module.exports = _curry2(_checkForMethod('forEachObjIndexed', function forEachObjIndexed(fn, obj) {
```
davidchambers: I don't think we should dispatch to a `forEachObjIndexed` method.

CrossEye: Agreed.

CrossEye:
I don't feel much of a need for this function, but I can see why someone might want it.

A tepid :herb: from me.

davidchambers:
I'd prefer to deprecate `R.forEach`, but so long as that function exists it seems completely reasonable to me to include the "indexed" equivalent.

src/forEachObjIndexed.js
centsent: Updated, review it again please.

src/forEachObjIndexed.js
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Iterate over an input `object`, calling a provided function `fn` for each
+ * key and value in the object.
+ *
+ * `fn` receives three argument: *(value, key, obj)*.
+ *
+ * @func
+ * @memberOf R
+ * @category Object
+ * @sig f => (a -> b) -> f a -> f b
```
davidchambers: I think the signature should be something like this:

```haskell
forEachObjIndexed :: ((a, String, StrMap a) -> Any) -> StrMap a -> StrMap a
```

test/forEachObjIndexed.js
```diff
@@ -0,0 +1,41 @@
...
+    eq(sideEffect, obj);
+  });
+
+  it('returns the original object', function() {
+    var s = '';
+    eq(R.forEachObjIndexed(function(value) { s += value; }, obj), obj);
+    eq('12123', s);
+  });
+
+  it('no dispatches to `forEachObjIndexed` method', function() {
```
davidchambers: I don't think it's necessary to assert this.

centsent: Removed, check it again :)

davidchambers:
Looks okay to me. :)

Issue #1904, opened 2016-09-15
R.construct example uses an arrow function as constructor 


iliocatallo:
Hi,

the [example](http://ramdajs.com/0.22.1/docs/#construct) for `R.construct` uses the `Widget` arrow function as a constructor function. However, my understanding is that arrow functions cannot be used as constructors because of lexical binding (see, e.g., [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) and [here](http://exploringjs.com/es6/ch_arrow-functions.html)). 

ciao!

PR #1921, opened 2016-09-29
Replace documentation Widget arrow with Constructor Function

knpwrs:
Arrow functions are not constructable.


CrossEye:
Absolutely.  This was clearly a bit of overeagerness when we converted our examples to arrows.

This was raised recently, and then immediately closed (#1904) and I'm not sure why.  But I missed it first time around. 

Thanks!

Issue #1922, opened 2016-09-29
over with lensPath turns arrays into objects


RedNifre:
Modifying nested objects/arrays with a lensPath seems to turn arrays into objects.

E.g. over(lensProp([0,0]), identity, [['x']]) returns {'0': {'0': 'x'}} instead of [['x']]

buzzdecafe:
I think you want `lensIndex` here. 

`over(lensIndex(0), identity, [['x']]) //=> [['x']]`

Also, `lensProp` expects a String, so the fact that it "works" at all when passed `[0, 0]` is ... disappointing

RedNifre:
No, I don't want to use lensIndex here because as you said it only goes down one level.

Consider `var thing = {foo: [{bar:1},{bar:9}]}`

Why should this not work?: `over(lensPath(["foo", 1, "bar"]), add(1), var) // => {foo: [{bar:1},{bar:10}]}`

I find that to be way more intuitive than `over(pipe(lensProp("foo"), lensIndex(1), lensProp("bar")), add(1), var)`

mjgpy3:
@RedNifre You should be able to compose the index lenses, e.g.

```
over(compose(lensIndex(0), lensIndex(0)), identity, [['x']]) 
```
returns `[['x']]`

buzzdecafe:
This is a related discussion: https://github.com/ramda/ramda/issues/1332

CrossEye:
> This is a related discussion: #1332

There was also discussion in #773, and #1839.  In the latter I proposed that we replace the current `lensPath` with something equivalent to 

```js
const lensPath = compose(
  apply(compose),
  map(ifElse(is(Number), lensIndex, lensProp))
);
```

And this would solve the problem expressed by @RedNifre.  But no one responded to that suggestion.  What do people think?  @buzzdecafe?  @davidchambers?  @asaf-romano?  @scott-christopher?  Others?

Before someone dings me on the signature, I would think it appropriate to create a union type such as 

```hs
PrimitiveKey = String | Int
```

and use it for this function:

```hs
lensPath :: [PrimitiveKey] -> Lens
```

scott-christopher:
>  In the latter I proposed that we replace the current lensPath with something equivalent to [...]

Seems sane to me. I've raised #1924 to capture and discuss this approach. Alternatively the implementation of `assocPath` could be updated to support numeric keys which would flow on down to address this issue too.

tolmasky:
I definitely want this behavior.

[...]

PR #1924, opened 2016-09-30
Add support for integer path segments in lensPath

scott-christopher:
This change adds support for using integer values as segments in the path given to `lensPath`. The previous version of this was coercing arrays back into objects when numeric keys were used due to the implementation of `assocPath`.

An alternative solution to this would be to add support numeric keys in `assocPath` and leave the current implementation for `lensPath` as it is.

Addresses #1922

### DIFF
--- a/src/lensPath.js
+++ b/src/lensPath.js
@@ -1,7 +1,12 @@
 var _curry1 = require('./internal/_curry1');
-var assocPath = require('./assocPath');
-var lens = require('./lens');
-var path = require('./path');
+var _isInteger = require('./internal/_isInteger');
+var _map = require('./internal/_map');
+var apply = require('./apply');
+var compose = require('./compose');
+var identity = require('./identity');
+var ifElse = require('./ifElse');
+var lensIndex = require('./lensIndex');
+var lensProp = require('./lensProp');
 
 
 /**
@@ -11,19 +16,27 @@ var path = require('./path');
  * @memberOf R
  * @since v0.19.0
  * @category Object
+ * @typedefn Idx = String | Int
  * @typedefn Lens s a = Functor f => (a -> f a) -> s -> f s
- * @sig [String] -> Lens s a
+ * @sig [Idx] -> Lens s a
  * @param {Array} path The path to use.
  * @return {Lens}
  * @see R.view, R.set, R.over
  * @example
  *
- *      var xyLens = R.lensPath(['x', 'y']);
+ *      var xHeadYLens = R.lensPath(['x', 0, 'y']);
  *
- *      R.view(xyLens, {x: {y: 2, z: 3}});            //=> 2
- *      R.set(xyLens, 4, {x: {y: 2, z: 3}});          //=> {x: {y: 4, z: 3}}
- *      R.over(xyLens, R.negate, {x: {y: 2, z: 3}});  //=> {x: {y: -2, z: 3}}
+ *      R.view(xHeadYLens, {x: [{y: 2, z: 3}, {y: 4, z: 5}]});
+ *      //=> 2
+ *      R.set(xHeadYLens, 1, {x: [{y: 2, z: 3}, {y: 4, z: 5}]});
+ *      //=> {x: [{y: 1, z: 3}, {y: 4, z: 5}]}
+ *      R.over(xHeadYLens, R.negate, {x: [{y: 2, z: 3}, {y: 4, z: 5}]});
+ *      //=> {x: [{y: -2, z: 3}, {y: 4, z: 5}]}
  */
-module.exports = _curry1(function lensPath(p) {
-  return lens(path(p), assocPath(p));
+module.exports = _curry1(function lensPath(path) {
+  if (path.length === 0) {
+    return identity;
+  } else {
+    return apply(compose, _map(ifElse(_isInteger, lensIndex, lensProp), path));
+  }
 });
--- a/src/path.js
+++ b/src/path.js
@@ -8,7 +8,8 @@ var _curry2 = require('./internal/_curry2');
  * @memberOf R
  * @since v0.2.0
  * @category Object
- * @sig [String] -> {k: v} -> v | Undefined
+ * @typedefn Idx = String | Int
+ * @sig [Idx] -> {a} -> a | Undefined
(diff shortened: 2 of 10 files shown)


buzzdecafe:
IMO if we're gonna do this, it makes sense to do it in `assocPath` and leave `lensPath` implementation as is.

src/lensPath.js
```diff
@@ -11,19 +12,23 @@ var path = require('./path');
...
+ *      //=> 2
+ *      R.set(xHeadYLens, 1, {x: [{y: 2, z: 3}, {y: 4, z: 5}]});
+ *      //=> {x: [{y: 1, z: 3}, {y: 4, z: 5}]}
+ *      R.over(xHeadYLens, R.negate, {x: [{y: 2, z: 3}, {y: 4, z: 5}]});
+ *      //=> {x: [{y: -2, z: 3}, {y: 4, z: 5}]}
  */
-module.exports = _curry1(function lensPath(p) {
-  return lens(path(p), assocPath(p));
-});
+module.exports = reduce(function(lens, path) {
```
davidchambers: Should we name the second parameter `idx` or `component` rather than `path`? Unless I'm missing something it's a path component rather than a path.

src/lensPath.js
```diff
@@ -11,19 +12,23 @@ var path = require('./path');
...
+  return compose(lens, (path | 0) === path ? lensIndex(path) : lensProp(path));
```
davidchambers: Does `(path | 0) === path` mean we'll treat `0` as an array index but treat `0.5` as a property name? This doesn't sit well with me. Also, `new Number(42)` will be treated as `'42'` rather than `42`. Could we use `_isNumber(path)` instead?

src/lensPath.js
CrossEye: ... or possibly `step`?

CrossEye:
Is this implementation chosen over the one I proposed simply due to distaste over variadic `compose`?

I would assume that there is some relative inefficiency in the output of this for `lensProp(['foo', 1, 'bar', 2, 'baz'])`:

```js
compose(lensProp('baz'), compose(lensIndex(2), compose(lensProp('bar'), compose(lensIndex(1), lensProp('foo')))))
```

compared to the other version's 

```js
compose(lensProp('baz'), lensIndex(2), lensProp('bar'), lensIndex(1), lensProp('foo'));
```

But perhaps not.  I haven't tested.

CrossEye:
@buzzdecafe:

Why the preference to do this on `assocPath` instead?  This seems to me useful for `over` and `view` as well as `set`.

buzzdecafe:
> Why the preference to do this on assocPath instead? This seems to me useful for over and view as well as set.

Totally agree -- if we are going down the variable-type path, we should do it everywhere. My only point is that since the current impl of `lensPath` is in terms of `assocPath`, then `lensPath` would just drop out.

RedNifre:
I'm all for it. Also, if paths can mix properties and indices, would it make sense to unify lensProp and lensIndex as well?

CrossEye:
> If we are going down the variable-type path, we should do it everywhere.

Yes, of course.  I had blanked out on the fact that we implemented it that way.

src/lensPath.js
scott-christopher: > Does (path | 0) === path mean we'll treat 0 as an array index but treat 0.5 as a property name?

It doesn't make sense to treat floats as array indices (what is the 0.5th element of an array?).

> Also, new Number(42) will be treated as '42' rather than 42. Could we use _isNumber(path) instead?

I'm not sure I follow, as it seems to behave as I'd expect:

```js
const x = new Number(1)
typeof (x | 0) //=> "number"

const xs = [10, 11, 12]
xs[x] = 42
xs //=> [10, 42, 12]
```

scott-christopher: Ah, I think I've caught up now :D
```js
const y = new Number(42)
(y | 0) === y //=> false
```

scott-christopher: If we do decide to go ahead with this kind of change, I'd still prefer to restrict the array branching behaviour to integers only, so the `new Number` issue could be avoided with a `valueOf` call.

```js
(y | 0) === y.valueOf() //=> true
```

scott-christopher:
> @CrossEye: Is this implementation chosen over the one I proposed simply due to distaste over variadic compose?

The only problem with the previous proposal was that the current tests suggest that we want to support passing an empty list for the path and compose doesn't like an empty argument list.

I guess we could just return `identity` if the list is empty and then carry on with your previous proposed implementation if desired. Otherwise I can update the current implementation to reduce over `_pipe` instead, which is effectively what `compose` does internally anyway.

CrossEye:
@scott-christopher: I don't see any great reason to change it.  I was was mostly curious about the motivation.  The empty case makes sense.  The more I've thought about it, the less I expect there to be significant performance differences.  (I miss JSPerf and have been lazy about settting up anything else.)

scott-christopher:
I've raised #1926 as an example of implementing this via `assocPath` instead.

src/lensPath.js
scott-christopher: I've updated it to `idx` to keep it inline with the typedef.

src/lensPath.js
scott-christopher: Just realised `_isInteger` is a thing... 🙃 (though it looks like the current implementation also doesn't support boxed `Number` instances).

I've replaced the inlined check with `_isInteger`.

scott-christopher:
@CrossEye: I've updated it to use your implementation because it also looks a lot cleaner.

src/lensPath.js
davidchambers: A path containing `0.5` should raise an exception. If non-integer values are treated as strings, our type definition is really `Idx = Integer | NonInteger`, more simply written `Idx = Any`. :stuck_out_tongue_winking_eye:

jigargosar:
I have suffered from this a lot. Arrays being converted to object. 

right now my alternative is to use:
const lensPath2 = unapply(compose(apply(compose), map(ifElse(isNumber, lensIndex, lensProp))))

scott-christopher:
Closed in favour of #1926

PR #1926, opened 2016-10-01
Add support for integer path segments in assocPath

scott-christopher:
This is an alternative proposal to #1924 for supporting integer path segments in `assocPath` (and `lensPath` by its current implementation).

### DIFF
--- a/src/assocPath.js
+++ b/src/assocPath.js
@@ -1,4 +1,7 @@
 var _curry3 = require('./internal/_curry3');
+var _has = require('./internal/_has');
+var _isArray = require('./internal/_isArray');
+var _isInteger = require('./internal/_isInteger');
 var assoc = require('./assoc');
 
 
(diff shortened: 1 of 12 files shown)


src/assocPath.js
```diff
@@ -27,12 +31,19 @@ var assoc = require('./assoc');
...
+  if (path.length === 0) {
+    return val;
+  }
+  var idx = path[0];
+  if (path.length > 1) {
+    var nextObj = _has(idx, obj) ? obj[idx] : _isInteger(path[1]) ? [] : {};
+    val = assocPath(_slice(path, 1), val, nextObj);
+  }
+  if (_isInteger(idx) && _isArray(obj)) {
+    var arr = [].concat(obj);
```
scott-christopher: Would be nice to use `update` here instead, though if we want to continue supporting the creation of intermediate paths when they don't exist in the target object then we need to allow for the creation of sparse arrays (which `update` prevents).

buzzdecafe: yes, sparse array creation is what i didn't like about this in prior discussions. Current lensIndex behavior does not update non-existent indices

CrossEye:
Since @buzzdecafe [set me straight][mh], I would definitely prefer this approach.

Any objections?


  [mh]: https://github.com/ramda/ramda/pull/1924#issuecomment-250798430

buzzdecafe:
not enough to block it. i fret about the sparse array thing. that is pretty gross. this impl is less surprising, and maybe that is enough to justify it.

jgoux:
Yes, yes and yes!
Can't wait to use it! :star2: 
Thanks @scott-christopher

buzzdecafe:
oops, this one slipped between the cracks. i think it's good to merge, but needs some conflict resolution

CrossEye:
Agreed.  @scott-christopher: do you have time to address the merge issue?

scott-christopher:
Sure, I should be able to take a look later this afternoon.

scott-christopher:
The conflict was just due to the change from `_slice(path, 1)` to `Array.prototype.slice.call(path, 1)` :zap:

buzzdecafe:
:cow2:

CrossEye:
:herb:

Very glad to have this change in!

jgoux:
Congrats! :sheep: 
Would it be possible to tag a release/publish it to npm?

buzzdecafe:
> Very glad to have this change in!

more equivocal. but maybe we can button up the loose ends later

diedsmiling:
👍 Thanks for implementing this feature. I need very much at my project. When would be a release to npm?

buzzdecafe:
still a few things to iron out before next release; gotta herd the cats

Issue #1913, opened 2016-09-20
R.either performance (and performance in general)


axyz:
Hi there, I was fiddling around with Ramda and I was trying to implement a very simple `R.path` with a default fallback value.

It seems to be a very simple use case for R.either, however the performance cost of R.either seems o be quite big compared to a simple native `or` after the `R.path`

here is a fiddle: https://goo.gl/F51qxU

This is of course just an example, and is probably related to some small implementation detail (like curring and lifting the or). However I'm quite sure there maybe other cases of similar performance bottleneck hard to be discovered.

This is totally fine (abstractions always have a cost and contributing may be fun), however it is hard to fully embrace Ramda functions and style if you have to test their performance before using them.

Wouldn't be great to have something like  a benchmark suite for every function, maybe comparing speed to native implementation and maybe major competitors like lodash and keeping a result page up to date? 

That would really help out both the decision to use or not use some specific functions in production, but above all to let possible contributors immediately find out which functions have to be improved.

P.S. I'm using some Ramda functions on high traffic real world applications, but if from one side I'd like to write more functional and composable code, from the other side I'm scared about nesting and composing Ramda functions too deeply (this applies to almost all utility libraries...), I know that zero cost abstractions in javascript are quite utopic, but I've also noticed that some functions from Ramda and other utility libraries sometimes are quite fast (also faster than native alternatives), to fill this gap between the reality and the dream about writing small and nicely composable functions everywhere it is really needed to address this issue, mainly because developers should trust the implementation and the reasonable overhead cost of any abstraction out there. Otherwise these great libraries will always have a very small space (using just a couple of functions that you know are fast and without daring too much with composition).

Sorry for the long issue (that may not really look like a proper issue...), but I think this is something interesting to be discussed, also to know what is your vision and plans about that.

buzzdecafe:
I'd be grateful for help with such a project. We have some benchmarks here:
https://github.com/ramda/ramda/tree/master/lib/bench

Any `*.bench.js` file dripped in there will get picked up by the `benchRunner` script. But we certainly haven't gone anywhere near as far as you recommend.

CrossEye:
I really would love to get some stronger benchmarks.

There are a few places where we know that Ramda is likely to be slower than native or than other libraries with overlapping functionality, mostly those places where we do some sort of value equality testing rather than reference comparisons.  Other than that, I would expect that if Ramda has performance issues, they're correctable... but only if we know about them.  So benchmarks would be great.

-----

[...]

PR #1929, opened 2016-10-06
Add native functions to benchmarks where they exist

bjdixon:
- Added benchmarks for either (addresses #1913)
 - Removed isSet benchmarks as isSet has been removed from ramda
 - Changed indexOf to includes in the contains benchmark as includes is
semantically closer to contains than indexOf


bjdixon:
I'm not sure that the additions I've made in (filter|find|findIndex)-where.bench.js are helpful as they mix native and ramda functions. In terms of benchmarking in general I would have thought it better to have a separate where.bench.js.

I'd like to create further PRs with benchmarks for the remaining unbenchmarked ramda functions (with native comparisons where appropriate) but will wait on feedback from this PR so I better understand the direction these should take.

lib/bench/either.bench.js
```diff
@@ -0,0 +1,34 @@
+var R = require('../..');
+
+var gt10 = x => x > 10;
+var even = x => x % 2 === 0;
+
+module.exports = {
+  name: 'either',
+  tests: {
+    'either(gt10, even, 101)': function() {
+      R.either(gt10, even, 101)
```
kedashoe: [Either](https://github.com/ramda/ramda/blob/v0.22.1/src/either.js) has an arity of 2 and returns a function so calling it like this doesn't do what it looks like.

lib/bench/contains.bench.js
```diff
@@ -16,8 +16,8 @@ module.exports = {
     'contains(c, cs)': function() {
       return contains(randomLowerAlpha(), alphabet);
     },
-    'native indexOf': function() {
-      return nativeIndexOf(randomLowerAlpha(), alphabet);
+    'native includes': function() {
```
kedashoe: I think it makes more sense to call this 'native contains' even if the native function we use is not called contains (either way definitely better than `indexOf` from before).

kedashoe:
Thanks for the PR, the benchmark stuff could definitely use some :heart:

> I'm not sure that the additions I've made in (filter|find|findIndex)-where.bench.js are helpful as they mix native and ramda functions. In terms of benchmarking in general I would have thought it better to have a separate where.bench.js.

Agreed, not sure we need native comparisons in benchmarks that test more than one ramda function.

In general I think it would be nice to have separate `Ramda vs Ramda` and `Ramda vs native vs etc` suites. Ramda vs others is interesting every now and then, but usually I just care if the changes I'm making make ramda itself faster or slower. Would also be nice to auto-compare a branch vs master.

lib/bench/either.bench.js
bjdixon: :flushed: oops

buzzdecafe:
> In general I think it would be nice to have separate `Ramda vs Ramda` and `Ramda vs native vs etc`

:+1: for this idea -- although it implies we would track benches over time?

CrossEye:
Wow, I missed this one entirely!  Very nice.

PR #1936, opened 2016-10-16
document behavior of functions in chain

buzzdecafe:


src/chain.js
```diff
@@ -11,19 +11,25 @@ var map = require('./map');
...
  * according to the [FantasyLand Chain spec](https://github.com/fantasyland/fantasy-land#chain).
+ * Treats the second argument as a FantasyLand Chain, when the the second argument is a function
+ * of the form `a -> x -> b`.
  *
  * @func
  * @memberOf R
  * @since v0.3.0
  * @category List
  * @sig Chain m => (a -> m b) -> m a -> m b
+ * @sig (x -> a) -> (a -> x -> b) -> (x -> b)
```
davidchambers: This isn't strictly necessary.

```haskell
chain :: Chain m => (a -> m b) -> m a -> m b
```

Replace `Chain m => m` with `Function x`:

```haskell
chain :: (a -> Function x b) -> Function x a -> Function x b
```

Use `x -> y` sugar for `Function x y`:

```haskell
chain :: (a -> x -> b) -> (x -> a) -> (x -> b)
```

I'd prefer to let the general signature tell the story.

src/chain.js
```diff
@@ -11,19 +11,25 @@ var map = require('./map');
  *
  * Dispatches to the `chain` method of the second argument, if present,
  * according to the [FantasyLand Chain spec](https://github.com/fantasyland/fantasy-land#chain).
+ * Treats the second argument as a FantasyLand Chain, when the the second argument is a function
+ * of the form `a -> x -> b`.
```
davidchambers: I think you mean the first argument. ;)

buzzdecafe: Not in this case. Did you see the example? `chain(head, append)`. Surely, the second argument is the `Chain` (`a -> x -> b`) here, and we are using dispatchiness to dictate how the arguments are consumed?

src/chain.js
buzzdecafe: > I'd prefer to let the general signature tell the story.

Does it tell the story? Or is it worth explicitly calling out the function case? At present the docs state:

```js
* Dispatches to the `chain` method of the second argument, if present,
* according to the [FantasyLand Chain spec]
```

Nobody reading the docs could tell that `Function` would work in this manner, since `Function` (typically) does not have a `chain` method.

src/chain.js
davidchambers: This program does not compile:

```haskell
module Main where

import Control.Monad ((>>=))

append :: a -> [a] -> [a]
append x xs = xs ++ [x]

chain :: Monad m => (a -> m b) -> m a -> m b
chain f m = m >>= f

f = chain head append

main = return ()
```

```console
$ ghc Chain.hs
[1 of 1] Compiling Main             ( Chain.hs, Chain.o )

Chain.hs:11:16:
    Couldn't match type ‘[a -> b]’ with ‘[a] -> [a]’
    Expected type: a -> [a -> b]
      Actual type: a -> [a] -> [a]
    Relevant bindings include f :: a -> b (bound at Chain.hs:11:1)
    Probable cause: ‘append’ is applied to too few arguments
    In the second argument of ‘chain’, namely ‘append’
    In the expression: chain head append
```

I believe there's a type error in `R.chain(R.head, R.append)` but it somehow "works" anyway. I don't see how to get from `Chain m => (a -> m b) -> m a -> m b` to the signature you're proposing. Do you?

This program compiles:

```haskell
module Main where

import Control.Monad ((>>=))

append :: a -> [a] -> [a]
append x xs = xs ++ [x]

chain :: Monad m => (a -> m b) -> m a -> m b
chain f m = m >>= f

f = chain append head

main = putStrLn $ show $ f [1,2,3]
```

It prints `[1,2,3,1]` when run. :)

It looks as though there's a bug in this code:

> ```javascript
> return function() {
>   return monad.call(this, fn.apply(this, arguments)).apply(this, arguments);
> };
> ```

buzzdecafe: Remember we are only doing *half* of Haskell. 

> I don't see how to get from Chain m => (a -> m b) -> m a -> m b to the signature you're proposing. Do you?

Yes, it's odd alright; clearly `head` is `x -> a` and `append` is `a -> x -> b`, and the other signature makes more sense. But I'm not proposing it. It's already a fact today, since https://github.com/ramda/ramda/commit/35acea0dd9a32f0327737689ac9cc7b279d8d058

I am simply recommending we document it. If you are suggesting we should treat this behavior as a defect, that is a very different conversation.

davidchambers: Did you notice that when I flipped the arguments I could compile the Haskell program and produce the desired value? I'm happy to document the behaviour, but we should fix the bug in our implementation first. ;)

Here's a valid implementation:

```javascript
//  Function$prototype$chain :: (a -> b) ~> (b -> a -> c) -> (a -> c)
var Function$prototype$chain = function(f) {
  var chain = this;
  return function(x) { return f(chain(x))(x); };
};
```

I'll submit a pull request shortly.

buzzdecafe:
see #1937; this does seem like weird behavior.

src/chain.js
CrossEye: No one seems to be mentioning that the [fantasy-land spec][fl] signature is different from ours.  I think that's much of the problem.  In FL, the signature is

```hs
chain :: Chain m => m a ~> (a -> m b) -> m b
```

and that's what our function-branch supports.  I haven't checked yet about our list-branch, as this stuff makes my head hurt.  All this started with an answer from @bergus in a [StackOverflow question][so].  But that answer seemed to be speculative, just assuming that Ramda would work as expected for functions, which, it turns out, it does.  I think the problem is not that the documentation is incomplete, but that it's entirely wrong, at least for functions.  And if it's wrong for functions, and right for other types, we've got significant problems.



  [fl]: https://github.com/fantasyland/fantasy-land#chain
  [so]: http://stackoverflow.com/questions/40026018

davidchambers: > In FL, the signature is
>
> ```haskell
> chain :: Chain m => m a ~> (a -> m b) -> m b
> ```
>
> and that's what our function-branch supports.

I believe you're mistaken. Here's the [implementation][1] in [sanctuary-type-classes][2]:

> ```javascript
> //  Function$prototype$chain :: (a -> b) ~> (b -> a -> c) -> (a -> c)
> var Function$prototype$chain = function(f) {
>   var chain = this;
>   return function(x) { return f(chain(x))(x); };
> };
> ```

[`Z.chain`][3] takes its arguments in the same order as `R.map`, yet the example works as expected:

```javascript
> Z.chain(R.append, R.head)([1, 2, 3])
[1, 2, 3, 1]
```

`R` and `Z` functions needn't take their arguments in the same order as the corresponding FL methods do.

> this stuff makes my head hurt

I believe it's because the various FL implementations are scattered throughout Ramda's (large) codebase, and they're not defined as methods (so one cannot easily verify that they match the type signatures in the spec).

I'd :heart: Ramda to depend on sanctuary-type-classes so all the FL implementations live in one place. It's a small library with no run-time dependencies. Depending on it would make the Ramda codebase easier to comprehend, and may even fix some bugs which have so far gone undetected.


[1]: https://github.com/sanctuary-js/sanctuary-type-classes/blob/master/index.js#L715-L719
[2]: https://github.com/sanctuary-js/sanctuary-type-classes
[3]: https://github.com/sanctuary-js/sanctuary-type-classes#chain

davidchambers: > It didn't help that the obviously bright @bergus seemed to pull this from nowhere and just expect it to work.

I imagine that @bergus is familiar with Haskell. It's difficult to guess the argument order of `R.chain` based on `(>>=)`, since the latter is usually used infix.

PR #1937, opened 2016-10-16
fix implementation of Function#chain

davidchambers:
See #1936

/cc @benperez

### DIFF
--- a/src/chain.js
+++ b/src/chain.js
@@ -24,12 +24,12 @@ var map = require('./map');
  *
  *      var duplicate = n => [n, n];
  *      R.chain(duplicate, [1, 2, 3]); //=> [1, 1, 2, 2, 3, 3]
+ *
+ *      R.chain(R.append, R.head)([1, 2, 3]); //=> [1, 2, 3, 1]
  */
 module.exports = _curry2(_dispatchable('chain', _xchain, function chain(fn, monad) {
   if (typeof monad === 'function') {
-    return function() {
-      return monad.call(this, fn.apply(this, arguments)).apply(this, arguments);
-    };
+    return function(x) { return fn(monad(x))(x); };
   }
   return _makeFlat(false)(map(fn, monad));
 }));
--- a/test/chain.js
+++ b/test/chain.js
@@ -33,10 +33,12 @@ describe('chain', function() {
         return r + a;
       };
     };
-    var bound = R.chain(h, f);
+    var bound = R.chain(f, h);
     // (>>=) :: (r -> a) -> (a -> r -> b) -> (r -> b)
     // h >>= f = \w -> f (h w) w
     eq(bound(10), (10 * 2) + 10);
+
+    eq(R.chain(R.append, R.head)([1, 2, 3]), [1, 2, 3, 1]);
   });
 
   it('dispatches to objects that implement `chain`', function() {


test/chain.js
```diff
@@ -33,7 +33,7 @@ describe('chain', function() {
         return r + a;
       };
     };
-    var bound = R.chain(h, f);
+    var bound = R.chain(f, h);
     // (>>=) :: (r -> a) -> (a -> r -> b) -> (r -> b)
     // h >>= f = \w -> f (h w) w
     eq(bound(10), (10 * 2) + 10);
```
buzzdecafe: i wouldn't mind seeing the `chain(append, head)([1, 2, 3]) //=> [1, 2, 3, 1]` example tested here. For my money that is a very clear illustration of the behavior

davidchambers: :zap:

buzzdecafe:
i'm glad i inadvertently stumbled upon the problem that resulted in this fix

:dromedary_camel:

src/chain.js
```diff
@@ -24,12 +24,12 @@ var map = require('./map');
...
  *      R.chain(duplicate, [1, 2, 3]); //=> [1, 1, 2, 2, 3, 3]
+ *
+ *      R.chain(R.append, R.head)([1, 2, 3]); //=> [1, 2, 3, 1]
  */
 module.exports = _curry2(_dispatchable('chain', _xchain, function chain(fn, monad) {
   if (typeof monad === 'function') {
-    return function() {
-      return monad.call(this, fn.apply(this, arguments)).apply(this, arguments);
-    };
+    return function(x) { return fn(monad(x))(x); };
```
buzzdecafe: of all the constraints we try to satisfy, supporting `this` context is the first one i would throw overboard.

davidchambers: Down with `this`!

benperez:
Nice catch folks, the implementation doesn't match the signature so that was definitely a bug. I think it probably went uncaught because we were trying to do something crazy like this 😛 :
`monad.call(this, fn.apply(this, arguments)).apply(this, arguments);`

How do we feel about the fact that our `Chain` method takes its args in the opposite order of `Fantasy-Land` (and also from Haskell's `(>>=)` FWIW). In the general case, I think taking the bound function first (`R.chain(x => fnThatReturnsAMaybe, Just(1))` makes more sense, since that's what we'd want to partially apply in a point-free style.

buzzdecafe:
> How do we feel about the fact that our Chain method takes its args in the opposite order of Fantasy-Land

Seems consistent with `map`, e.g.

FantasyLand `map`
`map :: Functor f => f a ~> (a -> b) -> f b`

Ramda `map`
`map :: Functor f =>  (a -> b) -> f a -> f b`

FantasyLand `chain`
`chain :: Chain m => m a ~> (a -> m b) -> m b`

Ramda `chain`
`chain :: Chain m => (a -> m b) -> m a -> m b`

davidchambers:
Adding `ap` to the picture makes the similarities between the functions even more apparent:

```haskell
map   :: Functor f => (a -> b) -> f a -> f b
ap    :: Apply f => f (a -> b) -> f a -> f b
chain :: Chain f => (a -> f b) -> f a -> f b
```

benperez:
Cool, 🐪 from me on this change

buzzdecafe:
me too
🐪

CrossEye:
:herb:

PR #1938, opened 2016-10-17
integrate sanctuary-type-classes

davidchambers:
This pull request provides an idea of how [sanctuary-type-classes](https://github.com/sanctuary-js/sanctuary-type-classes) could be used in Ramda. It would allow us to share the implementations of `toString`, `equals`, `map`, `ap`, `chain`, etc.

Ramda's Fantasy Land integration is currently outdated (`fantasy-land@1.x.x` is not yet supported), incomplete (some methods do not have corresponding `R` functions), and patchy (implementations for built-in types have been added ad hoc). I believe a central repository of FL implementations for built-in types is good for the JavaScript FP community. There's no reason for these functions to be defined in multiple libraries.

I'm opening this pull request for discussion. A lot of work will be required to get it ready to merge. I'll wait to learn what others think of the idea before spending more time on this patch.

Before:

``` javascript
var _map = require('./internal/_map');
var _reduce = require('./internal/_reduce');
var _xmap = require('./internal/_xmap');
var curryN = require('./curryN');
var keys = require('./keys');

module.exports = _curry2(_dispatchable('map', _xmap, function map(fn, functor) {
    case '[object Function]':
      return curryN(functor.length, function() {
        return fn.call(this, functor.apply(this, arguments));
      });
    case '[object Object]':
        acc[key] = fn(functor[key]);
        return acc;
      }, {}, keys(functor));
    default:
      return _map(fn, functor);
  }
}));
```

After:

``` javascript
var Z = require('sanctuary-type-classes');

var _xmap = require('./internal/_xmap');

module.exports = _curry2(_dispatchable(_xmap, Z.map));
```


Bradcomp:
Looking at your example, is the idea that Ramda would still dispatch to a `map` (or whatever) function if it exists, but by using `sanctuary-type-classes` you get _both_ the canonical implementations for the built in types _and_ the dispatching to the fantasy land methods if they exist? 

That seems like a really good idea to me.  

I have a theoretical concern that probably is of little practical relevance.  If a data type has both a `map` and `fantasyland/map` method, the regular `map` would be used over the FL implementation due to the `_dispatchable` wrapper.

davidchambers:
> Looking at your example, is the idea that Ramda would still dispatch to a `map` (or whatever) function if it exists, but by using `sanctuary-type-classes` you get _both_ the canonical implementations for the built in types _and_ the dispatching to the fantasy land methods if they exist?

That's right. Well put.

> If a data type has both a `map` and `fantasyland/map` method, the regular `map` would be used over the FL implementation due to the `_dispatchable` wrapper.

That's a very real concern. Removing these three lines might be part of the solution:

> ```javascript
> if (typeof obj[methodname] === 'function') {
>   return obj[methodname].apply(obj, args);
> }
> ```

buzzdecafe:
Should we go down this path, it looks as though we could get rid of `dispatchable` altogether. If that is so, I think that would be a very big win. 

We simply have to get agreement on what the scope of dispatching should be. Not an easy task! IMO, FL is sufficient to start. Should another specification arise then we should consider supporting that as well. But I am in favor of eliminating the current *ad hoc* dispatching.

davidchambers:
> Should we go down this path, it looks as though we could get rid of `dispatchable` altogether. If that is so, I think that would be a very big win.

Not quite. I assume we wish to continue to support transducers.

It's now apparent to me that there's a significant problem with our current approach to dispatching: it complects transducer support and method dispatching. As a result, we can't have functions which support transducers but do *not* dispatch, leading to silliness such as `R.dropRepeatsWith` dispatching to `dropRepeatsWith`. It's thus unclear whether we actually *intend* a certain function to dispatch.

I updated `_dispatchable` to only be concerned with transducers. It's now *dramatically* simpler. It's actually comprehensible! Previously I could not understand why the `_isArray` check was required; now there's an `_isTransformer` check but nothing else.

src/internal/_dispatchable.js
```diff
@@ -1,37 +1,22 @@
...
-    if (!_isArray(obj)) {
-      var args = _slice(arguments, 0, length - 1);
-      if (typeof obj[methodname] === 'function') {
-        return obj[methodname].apply(obj, args);
-      }
-      if (_isTransformer(obj)) {
-        var transducer = xf.apply(null, args);
-        return transducer(obj);
+    if (arguments.length > 0) {
+      var args = Array.prototype.slice.call(arguments, 0);
```
buzzdecafe: why not internal `_slice` here?

src/internal/_dispatchable.js
```diff
@@ -1,37 +1,22 @@
...
+ * a transformer; the result of applying `fn` to its arguments otherwise.
  *
  * @private
- * @param {String} methodname property to check for a custom implementation
  * @param {Function} xf transducer to initialize if object is transformer
  * @param {Function} fn default ramda implementation
  * @return {Function} A function that dispatches on object in list position
  */
-module.exports = function _dispatchable(methodname, xf, fn) {
+module.exports = function _dispatchable(xf, fn) {
```
buzzdecafe: since this is now just for supporting transducers, a name change may be in order, e.g. `_transducible` or `_xducible` or `_xduces` or ...

buzzdecafe:
> Not quite. I assume we wish to continue to support transducers.

Yes indeed; that occurred to me later. 

> I updated _dispatchable to only be concerned with transducers. It's now dramatically simpler. It's actually comprehensible!

:tada:

This feels pretty good to me, I'd be willing to give it a whirl. @CrossEye?

CrossEye:
I'm back to my old Internet connectivity problems, so I can't really look at code very well.

I absolutely agree with the idea of decomplecting Transducer support and dispatching.  I'm still in favor of much more dispatching than some others seem to be.  But as to the core question of this PR, I haven't been able to look very closely.  Perhaps I'll try tomorrow from work.

src/internal/_dispatchable.js
davidchambers: #1943

src/internal/_dispatchable.js
davidchambers: Good idea. I changed the function's name to `_transducible`. :)

buzzdecafe:
((:bell:)) where are we with this one?

CrossEye:
>  where are we with this one?

I think we have a bunch of PRs tied together: this one, #1900, #1949, #1950 (which has gotten absolutely no response.)

And it all comes down to the question of where our dispatching is headed.  While I like [your suggestion][co], that's still just a stopgap.

  [co]: https://github.com/ramda/ramda/pull/1900#issuecomment-257148243

davidchambers:
Here's a telling comparison:

```javascript
module.exports = _curry1(function empty(x) {
  return (
    (x != null && typeof x['fantasy-land/empty'] === 'function') ?
      x['fantasy-land/empty']() :
    (x != null && x.constructor != null && typeof x.constructor['fantasy-land/empty'] === 'function') ?
      x.constructor['fantasy-land/empty']() :
    (x != null && typeof x.empty === 'function') ?
      x.empty() :
    (x != null && x.constructor != null && typeof x.constructor.empty === 'function') ?
      x.constructor.empty() :
    _isArray(x) ?
      [] :
    _isString(x) ?
      '' :
    _isObject(x) ?
      {} :
    _isArguments(x) ?
      (function() { return arguments; }()) :
    // else
      void 0
  );
});
```

```javascript
var empty = function empty(typeRep) {
  return Monoid.methods.empty(typeRep)();
};
```

I think we should merge #1900, release v1.0.0, then drop support for old versions of the FL spec and release v2.0.0.

CrossEye:
@davidchambers:

What does "then drop support for old versions of the FL spec" mean to you?  Do you mean remove all dispatching not directly in support of the current version of FL?  Or something less significant?

And yes, that is a strong argument for the change (unless the same complexity is simply buried elsewhere) but I still entirely believe that ["we aim for an implementation both clean and elegant, but the API is king."][ph]  Changing the API in order to make for a nicer implementation will always seem backward to me.


  [ph]: https://github.com/ramda/ramda#philosophy

CrossEye:
>> What does "then drop support for old versions of the FL spec" mean to you? Do you mean remove all dispatching not directly in support of the current version of FL?

> I'd like to dispatch to `fantasy-land/map`, for example, but not to `map`. This was the whole point of the prefix, after all. :stuck_out_tongue_winking_eye:

Ah, but there I disagree.  To me, the point of the prefix is to identify those instances of such functions as `map` meant to fulfill a specific contract, in this case the FL `Functor` specification.  When I know that some input is at least claiming to act in accordance with those algebraic laws, I can then act on it as though those laws are in effect.  `lift`, which I actually use often, seems a good example<sup>1</sup>.

But when that is not true, I see no reason not to dispatch, and a very strong reason to do so: turning *methods* into composable function is to my mind extremely useful.

> Ramda has become a complex library with a complicated implementation. If we continue to add features without ever removing features or simplifying existing features the library will become more complex, and users will need to learn how all the various features interact with one another.

Quite possibly.  But to my mind, the features that I would prefer to remove are the placeholders and transducers.  Dispatching seems much too important to me.

----------
<sup>1</sup> What's the issue with `lift`?  I use it like this:
```js
R.lift(R.inc)(R.prop('age'))({age: 27}); //=> 27 + 1 = 28
R.lift(R.multiply)(R.inc, R.dec)(7); //=> 8 * 6 = 48
const f = (x) => `f(${x})`;
const g = (x) => `g(${x})`;
R.lift(f)(g)('x'); //=> "f(g(x))"
```

davidchambers:
> But when that is not true, I see no reason not to dispatch, and a very strong reason to do so: turning *methods* into composable function is to my mind extremely useful.

Useful, certainly, but unprincipled. `Functor f => (a -> b) -> f a -> f b` is not an accurate signature for `R.map` if it dispatches to a `map` method which does not even *claim* to satisfy the functor laws. We'd need something like `(a -> b) -> { map :: c ~> (a -> b) -> d } -> d`.

I agree that composing functions is more fun than composing methods. Were I asked to interact with a data type which provides a `map` method I might define a function to wrap it:

```javascript
//    mapFoo :: ??? -> ??? -> ???
const mapFoo = curry2((f, foo) => foo.map(f));
```

The name `mapFoo` suggests that this function is specialized to the Foo type. We should not pretend that this function is parametrically polymorphic. We can replace the question marks in the type signature with *concrete* types, but we have no idea how this function would operate on values of types other than Foo.

I was once in favour of having Ramda blindly dispatch, but I've completely changed my position. Lawless polymorphism is convenient but worthless as an abstraction. Lawful polymorphism, on the other hand, allows us to define functions such as [`filter`][1] and [`lift2`][2] once and forall.

It's perfectly reasonable to expect a user to define `mapFoo` herself, FFI-style. We support all sorts of things which could—and I think should—be left to users. Here are a few examples:

| Ramda                 | Sanctuary                     |
| --------------------- | ----------------------------- |
| `R.unary(f)`          | `x => f(x)`                   |
| `R.binary(f)`         | `(x, y) => f(x, y)`           |
| `R.invoker(1, 'foo')` | `(x, y) => y.foo(x)`          |
| `R.forEach(f, xs)`    | `xs.forEach(f)`               |
| `R.pipeP(f, g, h)`    | `x => f(x).then(g).then(h)`   |

> What's the issue with `lift`?

`R.lift` works *sometimes*, but the _let me guess what you meant based on arity_ approach is full of surprises. What happens if I want to use `R.add` as a *unary* function rather than a binary one? With `S.lift`, there's no problem:

```javascript
> R.ap(S.lift(R.add, S.Just(2)), S.Just(2))
Just(4)
```

I can't use `R.lift`, as it will not guess my intent correctly. I can of course use `R.liftN(1)`:

```javascript
> R.ap(R.liftN(1, R.add)(S.Just(2)), S.Just(2))
Just(4)
```

Let's compare the type signatures of the respective functions:

```
S.lift :: Functor f => (a -> b) -> f a -> f b
```

```
R.liftN(1) :: (*... -> *) -> ([*]... -> [*])
```

The latter is indecipherable. :confounded:


[1]: https://github.com/sanctuary-js/sanctuary-type-classes#filter
[2]: https://github.com/sanctuary-js/sanctuary-type-classes#lift2

buzzdecafe:
good points @davidchambers . 

1. On promiscuous dispatching, I agree. I don't see "general compatibility" with any `map` anywhere as the priority we should be pursuing. Less is most definitely more. For those who want to compose their object methods, see `invoker`. We shouldn't dispatch willy-nilly and hope for the best.
2. we should fix `lift`

CrossEye:
>> But when that is not true, I see no reason not to dispatch, and a very strong reason to do so: turning *methods* into composable function is to my mind extremely useful.

> Useful, certainly, but unprincipled. 

Come on man, this is Javascript!

> `Functor f => (a -> b) -> f a -> f b` is not an accurate signature for `R.map` if it dispatches to a `map` method which does not even *claim* to satisfy the functor laws.

No, but I always think of our type signatures as more aspirational than precise.  Documenting also that it will also dispatch to the user's method does not seem like much overhead.  Actually dispatching does not either.

> Lawless polymorphism is convenient but worthless as an abstraction.

For whom should this be worthwhile?  That it's not useful to the Ramda authors, or to someone trying to build generic abstractions on top of Ramda is certainly clear.  But that does not mean that it's useless to Ramda *users*.  If a user has a type with `filter`, `map`, and `take` methods, analogous to the Ramda functions, it's very convenient to her  to have Ramda support that sort of polymorphism.  Then she can simply `map` over her type, `filter` its values, `take` the first few elements.

My point is that I see only benefits; I don't see the harm.  If there's harm in Ramda doing this sort of dispatching, then we should [go all the way](https://github.com/ramda/ramda/pull/1950) and remove all dispatching.  Any argument I can see for dispatching to `fantasy-land/map`, which we can no more prove to be lawful than we can someone's random `map` function, seems to also support dispatching to `map`.

I guess I have a problem with the implications of the phrase "lawless polymorphism".  If there's a strong sense of lawfulness in Ramda, I picture it as the local constable keeping an eye out and not a hired gunslinger chasing down bandits.  I know that Sanctuary is striving for much more; while I applaud some of its efforts, some I think are simply in vain.  I don't think one user's `filter` method is inherently more problematic than another's `fantasy-land/map`.  While it's true that if the user's implementations follow the FantasyLand laws, there are further abstractions that can built on top of them, there is nothing to prevent

```js
MyType.prototpye['fantasy-land/map'] = (fn) => new MyType('foo'); // ignores `this` and fn
```

or worse.

----------

> We support all sorts of things which could—and I think should—be left to users. Here are a few examples:
> 
> | Ramda                 | Sanctuary                     |
> | --------------------- | ----------------------------- |
> | `R.unary(f)`          | `x => f(x)`                   |
> | `R.binary(f)`         | `(x, y) => f(x, y)`           |
> | `R.invoker(1, 'foo')` | `(x, y) => y.foo(x)`          |
> | `R.forEach(f, xs)`    | `xs.forEach(f)`               |
> | `R.pipeP(f, g, h)`    | `x => f(x).then(g).then(h)`   |

When we drop support for ES5 (will we ever finish 1.0 and start working on 2.0?!), I think dropping most of these would make sense.  ES2015's arrow syntax makes it much more palatable to do without them.  They still have a place, to my mind, until then.

----------

>> What's the issue with `lift`?

> `R.lift` works *sometimes*, but the _let me guess what you meant based on arity_ approach is full of surprises. What happens if I want to use `R.add` as a *unary* function rather than a binary one? With `S.lift`, there's no problem ...  I can't use `R.lift`, as it will not guess my intent correctly. I can of course use `R.liftN(1)`...

Of course `S.lift` fails for one of the cases I suggested:

```js
S.lift(R.multiply)(R.inc, R.dec)(7);
//=> Error: ‘lift’ requires two arguments; received three arguments
```

It would also fail for your suggestion in #1447 that we deprecate `converge` in favor of `lift`:

``` js
R.converge(R.subtract, [R.inc, Math.sqrt])(9); //=> 7
R.lift(R.subtract)(R.inc, Math.sqrt)(9); //=> 7
// but...
S.lift(R.subtract)(R.inc, Math.sqrt)(9); 
//=> Error: ‘lift’ requires two arguments; received three arguments
```

You need to use `S.lift2` here.  And then `S.lift3` for others.  Once you start needing all these different functions for the different arities, Ramda's `R.liftN(n)` starts to seem reasonable.  When `R.lift` can often simplify this further, all the better, but when it can't, we offer the lower level control.  I don't have an issue with that.  It's consistent with `curry` / `curryN` and `construct` / `constructN` too, of course.

All of this strikes me as the normal sort of work we tend to need to do to support user's potential variadic functions.  I think that attempts to force users to only work with fixed-arity functions are very much working against the grain of Javascript.

CrossEye:
> Let's compare the type signatures of the respective functions:
> 
> ```hs
> S.lift :: Functor f => (a -> b) -> f a -> f b
> ```
> 
> ```hs
> R.liftN(1) :: (*... -> *) -> ([*]... -> [*])
> ```
> 
> The latter is indecipherable. :confounded:

Agreed, 100%.  But I think we can fix that by moving the ellipsis from the internals to something between multiple signatures.  I'd like to do this for all such functions:

```hs
> R.lift :: (a -> b) -> f a -> f b
>           (a -> b -> c) -> f a -> f b -> f c
>           (a -> b -> c -> d) -> f a -> f b -> f c -> f d
>           ...
```

I think this makes these clear enough, even if we don't have a precise syntax for them.

davidchambers:
I've updated this pull request. If we adopt this approach we'll get to delete *lots* of code. :)

Is there support for removing transducers at the same time? This would allow the internals to be simplified further still.

buzzdecafe:
> While it's true that if the user's implementations follow the FantasyLand laws, there are further 
> abstractions that can built on top of them, there is nothing to prevent
>
> `MyType.prototpye['fantasy-land/map'] = (fn) => new MyType('foo'); // ignores `this` and fn`
> 
> or worse.

In that case, we are no worse off than we are today. But in some ways we will be better off: because if you follow the rules you get these powerful abstractions. And the cherry on top: We deep-six a bag of code we don't wanna maintain anyways.

My $0.02 is it makes sense to reuse Sanctuary's abstractions and take our chances with rogue `fantasy-land/map` implementers.

----
> Is there support for removing transducers at the same time? This would allow the internals to be simplified further still.

I am :-1: on that until we have some way to plug 'em in. Transducers are both pretty neat-o and our only nod to lazy-ish evaluation. Once we have a way to plug 'em in, I would be in favor of removal from the core lib. Although I do concede they are an odd fit.

davidchambers:
> I think @CrossEye 's sticking point will be promiscuous dispatch.

Scott, if you provide me with a comprehensive list of dispatching requirements or a rule I can apply to generate such a list, I will update this pull request. I'm envisaging something like this:

  - `R.clone` should dispatch to `clone` if present.
  - `R.map` should dispatch to `fantasy-land/map` if present, or `map` if present (in that order).
  - …

buzzdecafe:
((:bell:))

Bradcomp:
:heart: FL compatibility, and I think it makes sense to use sanctuary-type-classes for sure.

RE: Dispatching
I like the idea of having like a 'stack' of potential dispatches that Ramda would do, but I worry about the complexity when it comes to ensuring that we special case for the built in structures as well. Like, `any` could dispatch to `any` and `some`, but we would still want to special case for the standard array case...

While I may be in the minority here, I would even be on board with `map` and `chain` dispatching to `then` as well - `['fantasyland/map', 'map', 'then']`

davidchambers:
I'm going to close this pull request then open a new one. :)

PR #1943, opened 2016-10-18
simplify _slice

davidchambers:
I'd like to go a step further and use `Array.prototype.slice` directly at each call site. If that change would be accepted I'll close this pull request and open another.


CrossEye:
> I'd like to go a step further and use `Array.prototype.slice` directly at each call site. If that change would be accepted I'll close this pull request and open another.

:herb:

Issue #1920, opened 2016-09-28
leaking arguments object causes deopts


jaawerth:
I noticed Ramda methods like `curry`, as well as [many others throughout the codebase](https://github.com/ramda/ramda/search?utf8=%E2%9C%93&q=_slice%28arguments%29&type=Code) are passing the `arguments` object into functions (mostly `_slice` ~~or `Function#apply`~~ [_not apply - see comments_] quite frequently, including heavily-used internal utilities. In modern JS engines, and v8 in particular, this is a known [optimization killer](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#32-leaking-arguments).
## More background

Babel itself originally implemented rest args by compiling them into something along the same lines as `var _args = _slice(arguments)` but eventually updated to use the copy-loop approach as well, along with a number of other optimizations, and avoids leaking `arguments`.

https://github.com/babel/babel/pull/2833
## Solution

Removing this use of `_slice` ~~and `apply`~~ by either adding a build step that allows for args `rest` and `spread` (a SweetJS macro, babel, etc) or manually replacing it with a primitive copy loop as shown in the link above should improve performance across the board.

Personally I'd recommend adding a build step (it can be as simple as using babel with only a few specific plugins if you don't want unpredictable side effects) to avoid the maintainability/readability issues of making this change manually. Has anyone given this much thought? Adding a build step is a major change, so I wanted to bring it up here before I go and do it myself and submit a PR.

buzzdecafe:
@jaawerth thanks for the heads up. i think we want to get 1.0 out the door first. Then moving to ES6/rollup would make what you are proposing simpler. I hope we can get 1.0 out in the next few weeks?

bergus:
@jaawerth Notice that `fn.apply(this, arguments);` is explicitly exempt from being an optimisation killer

jaawerth:
@buzzdecafe Ah, good point regarding `apply` - I added that bit last-minute, that's what I get for over-editing ;-).

@buzzdecafe - That makes sense. While it isn't too hard, this wouldn't be a trivial change, particularly if you're in the midst of a major release!

CrossEye:
I agree that we need to think about this.  But I also believe we need to think very hard before we decide to pursue it.  We haven't had real-world complaints about the speed here, and I hate to optimize on spec.

[...]

PR #1944, opened 2016-10-19
remove _slice

davidchambers:
Supersedes #1943


buzzdecafe:
why? what is the benefit of this change?

davidchambers:
> why? what is the benefit of this change?

Another way to pose this is _What is the benefit of a custom `slice` implementation when `Array#slice` exists and may be applied to any array-like value?_

I looked through the commit history to see whether `_slice` was added to address a performance problem with `Array#slice`, but the function predates the build pipeline which we added about two years ago. `_slice` seems to me like a relic. Please correct me if I'm missing something.

CrossEye:
It seems a relic to me.  I don't think it's adding any value.  I can't recall if it ever did.

buzzdecafe:
> Another way to pose this is *What is the benefit of a custom slice implementation when Array#slice exists and may be applied to any array-like value?*

At the risk of infinite loop, another way to pose this is *What is the benefit of reverting to Array#slice when `_slice` already exists and may be applied to any array-like value?*

I honestly don't know if there is a reason to care about this or not. But I don't see how rephrasing the question is an argument for it?

buzzdecafe:
I suspect the internal `_slice` is an optimization along these lines:

> There are numerous ways to use arguments in a way that causes the function to be unoptimizable. 

https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#3-managing-arguments

If that is ture, then the existing implementation is worth it, especially since it is already there. I would need  a more persuasive argument (pardon the pun) than "it's a relic."

davidchambers:
It looks as though `_slice(arguments, 0)` and `Array.prototype.slice.call(arguments, 0)` make a function unoptimizable for the same reason: they "leak" `arguments`. This is the sort of change we would need to make in order to make these functions optimizable:

```diff
-  return _slice(list, idx);
+  var result = [];
+  while (idx < arguments.length) {
+    result.push(arguments[idx]);
+    idx += 1;
+  }
+  return result;
```

I'd prefer not to make such optimizations unless it becomes necessary.

I believe it's important not to favour the status quo when deciding how things should be done. This is a recipe for accumulating detritus. :stuck_out_tongue_winking_eye:

buzzdecafe:
If it ain't broke ...

CrossEye:
I was thinking of this as a _fix_ for that problem, but of course it isn't.  It's still passing the `arguments` object out.  I'm not particularly interested in the sorts of fix that would actually do something, some macro that expands to an inlined version of `_slice`, but if it turns out that we have serious performance problems without that, we might have to consider it.

Nonetheless, to me, removing an unnecessary internal function is a good thing on its own.

buzzdecafe:
i don't *know* that the "optimization killer" article was the motivation for `_slice`. That is pure speculation, so let's not elevate it to fact. But I am wiling to give the benefit of the doubt to whoever added it that there were good reasons for doing so. Before removing it we should understand those reasons and make an informed decision. In other words, we don't have enough information to decide whether it is "unnecessary" or not. The burden of proof is on those seeking to remove it to make that case.

CrossEye:
I thought that the reason for getting rid of it was an attempt to respond to #1920, but I could be way off base.

davidchambers:
The "optimization killer" article is not germane, since `_slice` currently prevents optimizations and this has always been the case.

The relevant commit is https://github.com/ramda/ramda/commit/87ccda0add7573f1f5c768c7a3b0168d746f2e09 from three and a half years ago. So the current implementation is supposedly faster than `Function.prototype.call.bind(Array.prototype.slice)`, which I do not find surprising. It appears that we have our current `_slice` implementation because it was (and probably still is) more efficient than the inefficient implementation it replaced.

So far we have not found any reason to favour `_slice` over the equivalent function provided for us by the language. On the other hand there's a compelling reason to remove `_slice`: it would reduce the number of places for bugs to hide.

buzzdecafe:
ah blast, it's my commit??? well i guess i can't defend it then. :frowning_face:

buzzdecafe:
... so if you want to revive this PR, i won't block

CrossEye:
:herb:

buzzdecafe:
yeah do it. the burden of proof was on you until it turned out to be my commit and i honestly have no idea why. and i definitely will not stand behind code i wrote three years ago

CrossEye:
> and i definitely will not stand behind code i wrote three years ago

Heck, for me, last month is pushing it!

Issue #994, opened 2015-04-01
Multi-field sort?


jbnv:
It's not readily apparent from the documentation how to structure a sort with secondary sort criteria. For example, I usually do a secondary sort by name. I can make it work by calling the secondary compare function inside the primary function, but that seems to defeat the Ramda philosophy.

buzzdecafe:
do you have some example code?

jbnv:
Yes. This is the result of refactoring some existing code to use Ramda. The recency sort puts items with the most recent errors first. Items with no recorded error go at the end of the list. For items with the same timestamp or no timestamp, the sort is by name.

```
var nameLowercase = R.pipe(R.prop('Name'), R.toLower);

var compareByName = function (a, b) {
	return nameLowercase(a) < nameLowercase(b);
}

var compareByRecency = R.comparator(function (a, b) {
	if (!a.LastErrorTimestamp && !b.LastErrorTimestamp) return compareByName(a, b);
	if (!a.LastErrorTimestamp) return false; // 'b' has an error and 'a' doesn't
	if (!b.LastErrorTimestamp) return true; // 'a' has an error and 'b' doesn't
	if (a.LastErrorTimestamp == b.LastErrorTimestamp) { return compareByName(a, b); }
	return a.LastErrorTimestamp > b.LastErrorTimestamp;
});

var sort = {
	"alpha": {
		label: "Alpha",
		fn: R.sortBy(nameLowercase)
	},
	"recency": {
		label: "Recent Errors First",
		fn: R.sort(compareByRecency)

	}
}
```

CrossEye:
Somehow I missed this issue when it came up. 

Do you have a suggested API, even one that breaks backward compatibility?

megawac:
Couple options here were discussed heavily for underscore and lodash

- function returning an array and comparing each item of the array https://github.com/jashkenas/underscore/pull/1751
- Array of strings or functions https://github.com/jashkenas/underscore/pull/1815
- Array of strings or objects https://github.com/lodash/lodash/issues/414

Also see https://lodash.com/docs#sortByAll

CrossEye:
One possibility I see is a three-function API:

    R.sortByAll([
        R.ascend(R.prop('name')),
        R.descend(R.prop('age'))
    ], people)

My idea was that `ascend` and `descend` returned comparators:

    ascend :: (Obj -> String) -> Obj -> Obj -> -1 | 0 | 1
[...]

PR #1946, opened 2016-10-19
sortWith, ascend and descend methods

davidgtonge:
See discussion in issue #994.
- `sortWith` sorts a list by an arbitrary number of comparators
- `ascend` and `descend` are helpful utility methods to create such comparators.

### DIFF
--- a/src/ascend.js
+++ b/src/ascend.js
@@ -1,0 +1,28 @@
+var _curry3 = require('./internal/_curry3');
+
+
+/**
+ * Makes an ascending comparator function out of a function that returns a value
+ * that can be compared with `<` and `>`.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig Ord b => (a -> b) -> a -> a -> Number
+ * @param {Function} fn A function of arity one that returns a value that can be compared
+ * @param {*} a The first item to be compared.
+ * @param {*} b The second item to be compared.
+ * @return {Number} `-1` if fn(a) < fn(b), `1` if fn(b) < fn(a), otherwise `0`
+ * @example
+ *
+ *      var byAge = R.ascend(R.prop('age'));
+ *      var people = [
+ *        // ...
+ *      ];
+ *      var peopleByYoungestFirst = R.sort(byAge, people);
+ */
+module.exports = _curry3(function ascend(fn, a, b) {
+  var aa = fn(a);
+  var bb = fn(b);
+  return aa < bb ? -1 : aa > bb ? 1 : 0;
+});
--- a/src/descend.js
+++ b/src/descend.js
@@ -1,0 +1,28 @@
+var _curry3 = require('./internal/_curry3');
+
+
+/**
+ * Makes a descending comparator function out of a function that returns a value
+ * that can be compared with `<` and `>`.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig Ord b => (a -> b) -> a -> a -> Number
+ * @param {Function} fn A function of arity one that returns a value that can be compared
+ * @param {*} a The first item to be compared.
+ * @param {*} b The second item to be compared.
+ * @return {Number} `-1` if fn(a) > fn(b), `1` if fn(b) > fn(a), otherwise `0`
+ * @example
+ *
+ *      var byAge = R.descend(R.prop('age'));
+ *      var people = [
+ *        // ...
+ *      ];
+ *      var peopleByOldestFirst = R.sort(byAge, people);
+ */
+module.exports = _curry3(function descend(fn, a, b) {
+  var aa = fn(a);
+  var bb = fn(b);
+  return aa > bb ? -1 : aa < bb ? 1 : 0;
+});
--- a/src/sortWith.js
+++ b/src/sortWith.js
@@ -1,0 +1,46 @@
+var _curry2 = require('./internal/_curry2');
+var _slice = require('./internal/_slice');
+
+
+/**
+ * Sorts a list according to a list of comparators.
+ *
+ * @func
+ * @memberOf R
+ * @category Relation
+ * @sig [a -> a -> Number] -> [a] -> [a]
+ * @param {Array} functions A list of comparator functions.
+ * @param {Array} list The list to sort.
+ * @return {Array} A new list sorted according to the comarator functions.
+ * @example
+ *
+ *      var alice = {
+ *        name: 'alice',
+ *        age: 40
+ *      };
+ *      var bob = {
+ *        name: 'bob',
+ *        age: 30
+ *      };
+ *      var clara = {
+ *        name: 'clara',
+ *        age: 40
+ *      };
+ *      var people = [clara, bob, alice];
+ *      var ageNameSort = R.sortWith([
+ *        R.descend(R.prop('age')),
+ *        R.ascend(R.prop('name'))
+ *      ]);
+ *      ageNameSort(people); //=> [alice, clara, bob]
+ */
+module.exports = _curry2(function sortWith(fns, list) {
+  return _slice(list).sort(function(a, b) {
+    var result = 0;
+    var i = 0;
+    while (result === 0 && i < fns.length) {
+      result = fns[i](a, b);
+      i += 1;
+    }
+    return result;
+  });
+});
--- a/src/sortWith.js
+++ b/src/sortWith.js
@@ -1,5 +1,4 @@
 var _curry2 = require('./internal/_curry2');
-var _slice = require('./internal/_slice');
 
 
 /**
@@ -34,7 +33,7 @@ var _slice = require('./internal/_slice');
  *      ageNameSort(people); //=> [alice, clara, bob]
  */
 module.exports = _curry2(function sortWith(fns, list) {
-  return _slice(list).sort(function(a, b) {
+  return Array.prototype.slice.call(list, 0).sort(function(a, b) {
     var result = 0;
     var i = 0;
     while (result === 0 && i < fns.length) {
--- a/test/sortWith.js
+++ b/test/sortWith.js
@@ -1,0 +1,89 @@
+var R = require('..');
+var eq = require('./shared/eq');
+
+
+var albums = [
+  {title: 'A Farewell to Kings', artist: 'Rush', genre: 'Rock', score: 3},
+  {title: 'Timeout', artist: 'Dave Brubeck Quartet', genre: 'Jazz', score: 3},
+  {title: 'Fly By Night', artist: 'Rush', genre: 'Rock', score: 5},
+  {title: 'Goldberg Variations', artist: 'Daniel Barenboim', genre: 'Baroque', score: 3},
+  {title: 'Art of the Fugue', artist: 'Glenn Gould', genre: 'Baroque', score: 3},
+  {title: 'New World Symphony', artist: 'Leonard Bernstein', genre: 'Romantic', score: 4},
+  {title: 'Romance with the Unseen', artist: 'Don Byron', genre: 'Jazz', score: 5},
+  {title: 'Somewhere In Time', artist: 'Iron Maiden', genre: 'Metal', score: 2},
+  {title: 'In Times of Desparation', artist: 'Danny Holt', genre: 'Modern', score: 1},
+  {title: 'Evita', artist: 'Various', genre: 'Broadway', score: 3},
+  {title: 'Five Leaves Left', artist: 'Nick Drake', genre: 'Folk', score: 1},
+  {title: 'The Magic Flute', artist: 'John Eliot Gardiner', genre: 'Classical', score: 4}
+];
+
+
+describe('sortWith', function() {
+  it('sorts by a simple property of the objects', function() {
+    var sortedAlbums = R.sortWith([
+      R.ascend(R.prop('title'))
+    ], albums);
+    eq(sortedAlbums.length, albums.length);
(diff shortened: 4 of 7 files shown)


src/ascend.js
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Makes an "ascending" comparator function out of a function that returns a value
+ * that can be compared
+ *
+ * @func
+ * @memberOf R
+ * @since v0.23.0
+ * @category Function
+ * @sig (a -> Number | String) -> (a, b -> Number)
```
kedashoe: How about

`@sig Ord b => (a -> b) -> (a, a -> Number)`

?

src/ascend.js
```diff
@@ -0,0 +1,29 @@
+var _curry1 = require('./internal/_curry1');
+
+
+/**
+ * Makes an "ascending" comparator function out of a function that returns a value
+ * that can be compared
+ *
+ * @func
+ * @memberOf R
+ * @since v0.23.0
```
kedashoe: You do not need to include these, `@since` annotations [are added automatically](https://github.com/ramda/ramda/blob/master/CONTRIBUTING.md)

src/ascend.js
```diff
@@ -0,0 +1,29 @@
+var _curry1 = require('./internal/_curry1');
+
+
+/**
+ * Makes an "ascending" comparator function out of a function that returns a value
+ * that can be compared
```
kedashoe: - We normally end these blurbs with periods.
- Does ascending need to be quoted?
- Maybe end with "that can be compared with `<` and `>`."

src/sortWith.js
```diff
@@ -0,0 +1,46 @@
...
+ *      var bob = {
+ *        name: 'bob',
+ *        age: 30
+ *      };
+ *      var clara = {
+ *        name: 'clara',
+ *        age: 40
+ *      };
+ *      var people = [clara, bob, alice];
+ *      R.sortWith([
```
kedashoe: Did you mean to set a variable to a function here? Something like

```js
var ageNameSort = R.sortWith([
        R.descend(R.prop('age')),
        R.ascend(R.prop('name'))
]);
ageNameSort(people); => //=> [alice, clara, bob]

File: src/ascend.js [hunk 5]
```diff
@@ -0,0 +1,29 @@
...
+ *
+ *      var byAge = R.ascend(R.prop('age'));
+ *      var people = [
+ *        // ...
+ *      ];
+ *      var peopleByYoungestFirst = R.sort(byAge, people);
+ */
+module.exports = _curry1(function ascend(fn) {
+  return function(a, b) {
+    var aa = fn(a)
```
kedashoe: I think we use semicolons to end lines of code? Did you see somewhere that does not use them?

test/sortWith.js
```diff
@@ -0,0 +1,89 @@
+var R = require('..');
+var eq = require('./shared/eq');
+
+
+var albums = [
```
kedashoe: Nice test data! I wouldn't mind something easier to look at and play around with though. Just something like

```js
[
  {a:'x', b:1, c:100},
  {a:'x', b:2, c:100},
  {a:'x', b:3, c:100},
  {a:'x', b:1, c:200},
  {a:'x', b:1, c:300},
  {a:'y', b:1, c:100},
...
]
```

or what have you. Not a big deal though if everyone is ok as it is.

CrossEye:
Thank you.  This is very nice.  :-)

I agree with all of @kedashoe's comments, except that I have mixed feelings about the test data.  He's probably right that a simpler set would be better, but since this set is borrowed from another version I wrote and come right off my music shelves, I still have some fondness for it.  :smile:

src/ascend.js
davidgtonge: Thanks, good catch - we should add an eslint rule for this. I tried adding it and got the following:
```
/ramda/src/applySpec.js
  37:76  error  Missing semicolon  semi

/ramda/src/dissoc.js
  25:22  error  Missing semicolon  semi

/ramda/test/complement.js
  1:29  error  Missing semicolon  semi

/ramda/test/indexBy.js
  33:46  error  Missing semicolon  semi
```

I'll create a separate pull request to fix those and add the rule in.

src/ascend.js
davidgtonge: Thanks - I'll add that in

src/sortWith.js
davidgtonge: My copy and paste from `sortBy` letting me down :-(. Thanks for the catch

src/ascend.js
davidgtonge: @kedashoe what do you suggest as the `@sig` for `sortWith`

src/sortWith.js
```diff
@@ -0,0 +1,45 @@
...
+ *      var bob = {
+ *        name: 'bob',
+ *        age: 30
+ *      };
+ *      var clara = {
+ *        name: 'clara',
+ *        age: 40
+ *      };
+ *      var people = [clara, bob, alice];
+ *      var ageNameSort - R.sortWith([
```
kedashoe: `var ageNameSort - R.sortWith([`

->

`var ageNameSort = R.sortWith([`

src/sortWith.js
```diff
@@ -0,0 +1,45 @@
+var _curry2 = require('./internal/_curry2');
+var _slice = require('./internal/_slice');
+
+/**
+ * Sorts a list according to a list of comparators.
+ *
+ * @func
+ * @memberOf R
+ * @category Relation
+ * @sig Ord b => (a -> b) -> [a] -> [a]
```
kedashoe: How about

@sig [(a, a -> Number)] -> [a] -> [a]

?

src/ascend.js
davidgtonge: I've created a separate PR: https://github.com/ramda/ramda/pull/1948

src/sortWith.js
```diff
@@ -0,0 +1,45 @@
...
+ *      };
+ *      var clara = {
+ *        name: 'clara',
+ *        age: 40
+ *      };
+ *      var people = [clara, bob, alice];
+ *      var ageNameSort - R.sortWith([
+ *        R.descend(R.prop('age')),
+ *        R.ascend(R.prop('name'))
+ *      ]); //=> []
```
kedashoe: I don't think we need the ` //=> []` here

src/ascend.js
```diff
@@ -2,14 +2,13 @@ var _curry1 = require('./internal/_curry1');
...
- * that can be compared
+ * Makes an ascending comparator function out of a function that returns a value
+ * that can be compared with `<` and `>`.
  *
  * @func
  * @memberOf R
- * @since v0.23.0
  * @category Function
- * @sig (a -> Number | String) -> (a, b -> Number)
+ * @sig Ord b => (a -> b) -> (a, a -> Number)
```
buzzdecafe: to reduce ambiguity in the sig, please group the arguments to the returned function, i.e.

`... ((a, a) -> Number)`

src/descend.js
```diff
@@ -2,14 +2,13 @@ var _curry1 = require('./internal/_curry1');
...
- * that can be compared
+ * Makes a descending comparator function out of a function that returns a value
+ * that can be compared with `<` and `>`.
  *
  * @func
  * @memberOf R
- * @since v0.23.0
  * @category Function
- * @sig (a -> Number | String) -> (a, b -> Number)
+ * @sig Ord b => (a -> b) -> (a, a -> Number)
```
buzzdecafe: as above

src/ascend.js
```diff
@@ -0,0 +1,28 @@
...
+ * @example
+ *
+ *      var byAge = R.ascend(R.prop('age'));
+ *      var people = [
+ *        // ...
+ *      ];
+ *      var peopleByYoungestFirst = R.sort(byAge, people);
+ */
+module.exports = _curry1(function ascend(fn) {
+  return function(a, b) {
```
buzzdecafe: should the returned function be curried?

src/descend.js
```diff
@@ -0,0 +1,28 @@
...
+ * @example
+ *
+ *      var byAge = R.descend(R.prop('age'));
+ *      var people = [
+ *        // ...
+ *      ];
+ *      var peopleByOldestFirst = R.sort(byAge, people);
+ */
+module.exports = _curry1(function descend(fn) {
+  return function(a, b) {
```
buzzdecafe: same question

src/sortWith.js
```diff
@@ -0,0 +1,45 @@
+var _curry2 = require('./internal/_curry2');
+var _slice = require('./internal/_slice');
+
+/**
+ * Sorts a list according to a list of comparators.
+ *
+ * @func
+ * @memberOf R
+ * @category Relation
+ * @sig [(a, a -> Number)] -> [a] -> [a]
```
buzzdecafe: same issue with ambiguous grouping here; `(a, a) -> Number` would be clearer. Or `a -> a -> Number` if curried

src/ascend.js
davidchambers: We should use `_curry3` here:

```haskell
ascend :: Ord b => (a -> b) -> a -> a -> Number
```

test/sortWith.js
buzzdecafe: i am a fan of test data with some character. adds a bit of pleasure to reading the tests that would otherwise be very dry.

Why not Gould's _Goldberg Variations_ (1955)? 😉

src/ascend.js
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Makes an ascending comparator function out of a function that returns a value
+ * that can be compared with `<` and `>`.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig Ord b => (a -> b) -> a -> a -> Number
+ * @param {Function} fn A function of arity one that returns a value that can be compared
+ * @return {Function} A Function :: a -> b -> Int that returns `-1` if a < b, `1` if b < a, otherwise `0`
```
davidchambers: This should be updated to reflect the fact that `ascend` is now ternary.

davidgtonge: Its also worth noting that `R.comparator` isn't ternary at the moment, perhaps we should change that also to make consistent?
```javascript
module.exports = _curry1(function comparator(pred) {
  return function(a, b) {
    return pred(a, b) ? -1 : pred(b, a) ? 1 : 0;
  };
});
```

src/sortWith.js
davidgtonge: Just checking before I commit, is this better?
`@sig [(a -> a -> Number)] -> [a] -> [a]`

davidchambers: If we're taking an array of curried binary functions:

```haskell
sortWith :: [a -> a -> Number] -> [a] -> [a]
```

Otherwise:

```haskell
sortWith :: [(a, a) -> Number] -> [a] -> [a]
```

src/ascend.js
CrossEye: Although I suppose we could make it ternary, that isn't the expected API.  The point is the signature as it stands, or better, as it should be cleaned up:

```hs
comparator:: ((a, a) -> Bool) -> ((a, a) -> Int)
-- or
comparator:: (a -> a -> Bool) -> (a -> a -> Int)
```

We accept a binary predicate (which really doesn't need to be curried, although it won't hurt if it is) and return a binary function that returns 0, -1, or +1, depending upon the values of `pred(first, second)` and `pred(second, first)`.  In other words, this function _creates_ a comparator suitable for use in `R.sort` or `Array.prototype.sort`.

Note that this is essentially using the predicated to extend the notion of `<` to additional types, although we don't have language level operator overloading.

CrossEye:
This looks good-to-go to me.   Any objections?

src/sortWith.js
```diff
@@ -0,0 +1,45 @@
...
+ *      };
+ *      var people = [clara, bob, alice];
+ *      var ageNameSort = R.sortWith([
+ *        R.descend(R.prop('age')),
+ *        R.ascend(R.prop('name'))
+ *      ]);
+ *      ageNameSort(people); //=> [alice, clara, bob]
+ */
+module.exports = _curry2(function sortWith(fns, list) {
+  return Array.prototype.slice.call(list, 0).sort(function(a, b) {
```
ascartabelli: Remember that, unless something changed, the ECMA specification doesn't require the sort algorithm to perform a stable sort.
It's also true that, as far as I know, every modern JS engine performs a stable sort, but still.

davidgtonge: true, but if I'm not wrong then `sortWith` would allow devs to ensure consistent sort order, for example by having the final `comparator` be something like `R.ascend(R.prop('id'))`. This is actually one of the primary use cases that I have for this function, its not just that stable sort isn't guaranteed, some API responses don't have guaranteed sort order either.

ascartabelli: I think I explained myself badly.
Was trying to point out that you were simply delegating the sorting to the native method without taking the "issue" of stable-sorting into account.
Now I notice that existing Ramda functions do the same, so they probably deemed this "issue" as a non-issue, which is fair enough.

CrossEye: >as far as I know, every modern JS engine performs a stable sort

Not last I knew [for Chrome][us].

We did consider adding a stable sort back in #1456, but that petered out.

  [us]: https://bugs.chromium.org/p/v8/issues/detail?id=90

ascartabelli: @CrossEye Thanks for pointing this out: I don't know why I was convinced that Chrome switched to stable sorting a while ago.

PR #1947, opened 2016-10-19
introduce odd & even funcs

webdude21:


davidchambers:
I'm opposed to these additions. The root problem is that Ramda does not restrict the domains of its functions: all functions are `Any -> ...` or `Any -> Any -> ...` or `Any -> Any -> Any -> ...`. As a result, a significant proportion of the discussion on the issue tracker is concerned with the behaviour of a Ramda function when applied to one of the infinite number of JavaScript values for which the mathematical function in question is undefined. If we're to add these functions, I imagine we'll see issues for questions such as these:

  - _Why does `R.even(2.1)` evaluate `true`?_
  - _Why does `R.even(Infinity)` evaluate `true`?_
  - _Why does `R.even(NaN)` evaluate `true`?_
  - _Why is `R.odd` not equivalent to `R.complement(R.even)`?_

In other words, because we've chosen to treat all JavaScript values as members of a universal type we can't reasonably support functions such as these which are defined only for the domain of integers. Sanctuary, on the other hand, does restrict function domains, so is able to provide [`S.even`][1] and [`S.odd`][2] without surprises.


[1]: https://sanctuary.js.org/#even
[2]: https://sanctuary.js.org/#odd

webdude21:
Hi David,

I see your point.
In this case you can go ahead and close the pull request.

Thank you for your time, and have a nice day :)

davidchambers:
@webdude21, I'm extremely grateful to you for opening this pull request. You have highlighted what I believe to be a serious problem with Ramda which I would like to discuss. There's absolutely nothing wrong with your pull request. If and when we add systematic type checking to Ramda I'll be in favour of merging it.

I'm just one of a handful of members of the [core team][1]. My thoughts on restricting function domains do not reflect the position of the team as a whole. It's quite possible we'll merge this pull request despite my protestations. :stuck_out_tongue_winking_eye:


[1]: https://github.com/ramda/ramda/wiki/Core-Team

webdude21:
@davidchambers, Thanks for the answer. Whatever you guys decide, I'll be happy either way. The library is awesome.

CrossEye:
@webdude21:

I think you've just gotten caught in some crossfire.  There is a large debate going on about whether Ramda wants to add such type-checking.

I also object, but for entirely different reasons.  I simply don't see a need.  Regardless of type-checking, I don't see `odd` and `even` as important enough to include.   Ones like `add`, which, like this one, are easy to build on ones own, are also widely-enough used to justify inclusion.  That doesn't seem to be the case for these.

PR #1949, opened 2016-10-21
remove unwanted dispatching

davidchambers:
To quote https://github.com/ramda/ramda/pull/1938#issuecomment-254425093:

> It's now apparent to me that there's a significant problem with our current approach to dispatching: it complects transducer support and method dispatching. As a result, we can't have functions which support transducers but do *not* dispatch, leading to silliness such as `R.dropRepeatsWith` dispatching to `dropRepeatsWith`. It's thus unclear whether we actually *intend* a certain function to dispatch.

The solution to this problem is to take an array of method names rather than a string. This will give us the freedom to have a function work with transducers without also dispatching. It will also give us the option of dispatching to one of *several* possible methods (e.g. `fantasy-land/map` or `map`).

At this stage the pull request does not actually remove dispatching. What I will do is comment on each of the functions in question and request that you vote :thumbsup: if you believe the function should continue to dispatch; :thumbsdown: otherwise. As we decide which functions should no longer dispatch I will update the pull request.

PR #1950, opened 2016-10-22
Remove dispatching

CrossEye:
This is an alternative to #1900 and #1949.

(**Edit**: I don't know why I originally used the words "delegation" and "dispatching" indiscriminately.  Fixed.) 

It seems to me that if we are going to get out of the business of dispatching, then we should do it wholeheartedly.  That means we should stop doing any dispatching whatsoever.
## What is the proposal?

Remove all dispatching.  The only types Ramda functions should operate upon are the ones it handles internally: lists, functions, objects, strings.  This means removing silly dispatching like `dropRepeatsWhile` that was only dispatching because of implementation details of our transducers, removing more useful dispatching like `filter`, which could easily work with user types such as trees as well as the built-in types, and removing quite common dispatching such as `map` that is part of the [FantasyLand Functor](https://github.com/fantasyland/fantasy-land#functor) specification.
## Does this make Ramda less powerful?

No, Ramda will still operate directly on the same types it always has, especially lists, often functions or plain objects, sometimes strings.  And a user [can still use Ramda](https://goo.gl/9ZglGJ) to call her custom type:

``` js
const fmap = R.invoker(1, 'map');
fmap(square, Just(5)); //=> Just(25)
fmap(square, Nothing()); //=> Nothing()
```
## Does this make Ramda less ergonomic?

It does, to an extent that depends on how much the user likes points-free style.  Perhaps a simple example is the change in the test for `composeK`:

**before**

``` js
R.compose(chain(h), chain(g), chain(f))
```

**after**

``` js
R.compose(x => x.chain(h), x => x.chain(g), x => x.chain(f))
```

While this is not horrible, it does make using Ramda less pleasant.
## Are you serious?

Partly.  It seems to me that we're headed very much in the wrong direction by removing dispatching.  (This is not exactly a reaction to #1949; some of the things we dispatch make little sense.)  I believe dispatching is one of the very good parts of Ramda.  But if we're going to remove it, I don't see any good reason to make an exception for the FantasyLand types.

My preference is still to stick with our dispatching.  But I'm not interested in a half-measure of removing the dispatching of everything but those functions currently supported by some other team.  I don't want our API to be subject to the whims of another project.

I'm sure this PR is incomplete.  I pulled it together in an hour, so it may be fairly low quality.  And since I last was online, other code has been merged that means that I'd have to rebase before we could even try to merge it.  That's ok.  I want to see where the conversation goes anyway.

### DIFF
--- a/.directory
+++ b/.directory
@@ -1,6 +1,0 @@
-[Dolphin]
-Timestamp=2016,10,22,11,5,58
-Version=3
-
-[Settings]
-HiddenFilesShown=true
--- a/.directory
+++ b/.directory
@@ -1,0 +1,6 @@
+[Dolphin]
+Timestamp=2016,10,22,11,5,58
+Version=3
+
+[Settings]
+HiddenFilesShown=true
--- a/.kateproject
+++ b/.kateproject
@@ -1,0 +1,4 @@
+{
+  "name": "CrossEye/ramda",
+  "files": [ { "git": 1 } ]
+}
--- a/.kateproject
+++ b/.kateproject
@@ -1,4 +1,0 @@
-{
-  "name": "CrossEye/ramda",
-  "files": [ { "git": 1 } ]
-}
--- a/src/ap.js
+++ b/src/ap.js
@@ -27,8 +27,6 @@ var map = require('./map');
  */
 module.exports = _curry2(function ap(applicative, fn) {
   return (
-    typeof applicative.ap === 'function' ?
-      applicative.ap(fn) :
     typeof applicative === 'function' ?
       function(x) { return applicative(x)(fn(x)); } :
     // else
--- a/src/composeK.js
+++ b/src/composeK.js
@@ -1,4 +1,3 @@
-var chain = require('./chain');
 var compose = require('./compose');
 var identity = require('./identity');
 var map = require('./map');
@@ -9,7 +8,7 @@ var prepend = require('./prepend');
  * Returns the right-to-left Kleisli composition of the provided functions,
  * each of which must return a value of a type supported by [`chain`](#chain).
  *
- * `R.composeK(h, g, f)` is equivalent to `R.compose(R.chain(h), R.chain(g), R.chain(f))`.
+ * `R.composeK(h, g, f)` is equivalent to `R.compose(x => x.chain(h), x => x.chain(g), x => x.chain(f))`.
  *
  * @func
  * @memberOf R
@@ -36,5 +35,5 @@ var prepend = require('./prepend');
  * @symb R.composeK(f, g, h)(a, b) = R.chain(f, R.chain(g, h(a, b)))
  */
 module.exports = function composeK() {
-  return compose.apply(this, prepend(identity, map(chain, arguments)));
+  return compose.apply(this, prepend(identity, map(function(arg) {return function(x) {return x.chain(arg);};}, arguments)));
 };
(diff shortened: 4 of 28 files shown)


Bradcomp:
> __before__
> `R.compose(chain(h), chain(g), chain(f))`
> __after__
> `R.compose(x => x.chain(h), x => x.chain(g), x => x.chain(f))`

I guess my concern is that the latter code won't actually work on arrays, while the former would work on both arrays and user defined types. So we would be losing some ability to abstract over both user defined types and built ins at the same time, without writing our own dispatching code (which I guess would be a live option).

> My preference is still to stick with our delegation. But I'm not interested in a half-measure of removing the delegation of everything but those functions currently supported by some other team. I don't want our API to be subject to the whims of another project.

Personally, I too like the dispatching.  An option that came up in another PR was to increase the dispatching - `all` could additionally dispatch to `every`, and `map` could dispatch to `fantasyland/map`, `map`, and even `then` if you wanted to go crazy with it 😛 .  This would obviously make the code more complex, but probably not _that_ more complex when stuff like transducers are taken into account.

CrossEye:
> I guess my concern is that the latter code won't actually work on arrays, while the former would work on both arrays and user defined types.

Yes, but if we restrict the dispatching to FantasyLand methods, this would work only on those user defined types that know the magical incantation of `fantasy-land/map`.

> An option that came up in another PR was to increase the dispatching - `all` could additionally dispatch to `every`, and `map` could dispatch to `fantasyland/map`, `map`, and even `then` if you wanted to go crazy with it.

I really like either doing either or supporting something like Bilby's environments][be], as discussed in #1037, to allow the user to add custom dispatching.


  [be]: http://bilby.brianmckenna.org/#environment

ntkoso:
For me the main difference between ramda and lodash/fp is dispatching. I think ramda needs more dispatching, not less. Ability to setup dispatching for different data structures, e.g. immutable-js integration.

davidchambers:
> > I guess my concern is that the latter code won't actually work on arrays, while the former would work on both arrays and user defined types.
>
> Yes, but if we restrict the dispatching to FantasyLand methods, this would work only on those user defined types that know the magical incantation of `fantasy-land/map`.

This is not true. `R.chain` currently supports arrays, even though `Array#chain` is not defined. `Array#map` *is* defined, but we take care to avoid using it in `R.map`!

There are two separate questions:

  - What do we hope to achieve by dispatching?
  - Should functions such as `R.chain` support built-in types where possible?

We're struggling to reach agreement on the first question, but I believe we're in agreement on the second question: yes, `R.chain` and friends should support built-in types where possible.

CrossEye:
>>> I guess my concern is that the latter code won't actually work on arrays, while the former would work on both arrays and user defined types.
>>
>> Yes, but if we restrict the dispatching to FantasyLand methods, this would work only on those user defined types that know the magical incantation of `fantasy-land/map`.

> This is not true. `R.chain` currently supports arrays, even though `Array#chain` is not defined. `Array#map` *is* defined, but we take care to avoid using it in `R.map`!

Oh, certainly.  I was just talking about the restriction among "user defined types".  Ramda would never lose the ability to work with lists and the like.

davidchambers:
Thanks for the clarification, Scott. I read your earlier comment as _this would only support (user-defined types which support `fantasy-land/map`)_ rather than _for user-defined types, this would only support those which support `fantasy-land/map`_. An important distinction!

CrossEye:
@davidchambers: Yes, I figured that's what it must be.  Here's to clarity of communications! :wine_glass:

PR #1952, opened 2016-10-25
`R.curry` unneeded by `R.lift`

kwijibo:
Remove unnecessary use of `R.curry` from `R.lift` examples; the function passed to `lift` doesn't need to be curried, so the example is clearer without it.


scott-christopher:
🍄  Thanks @kwijibo.

CrossEye:
:herb:

PR #1954, opened 2016-10-25
removed curry from the example for liftN

kwijibo:
you don't need to curry functions that you pass to `liftN`, it works just the same, so the example is clearer without the currying

### DIFF
--- a/src/liftN.js
+++ b/src/liftN.js
@@ -19,7 +19,7 @@ var map = require('./map');
  * @see R.lift, R.ap
  * @example
  *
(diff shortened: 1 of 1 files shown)


src/liftN.js
```diff
@@ -19,7 +19,7 @@ var map = require('./map');
  * @see R.lift, R.ap
  * @example
  *
- *      var madd3 = R.liftN(3, R.curryN(3, (...args) => R.sum(args)));
+ *      var madd3 = R.liftN(3,(...args) => R.sum(args));
```
davidchambers: Nit: Let's keep the space after the comma.

buzzdecafe:
:cow2:

PR #1955, opened 2016-10-26
Feature - OmitIf

zatziky:
`omitIf` omits properties of an object if they fulfill a test too.

```
var obj = {a: 1, b: 2, c: 3, d: 4, e: 5};

it('copies an object omitting the listed properties fulfilling a test', function() {
    eq(R.omitIf(R.lt(2), ['a', 'c', 'e'], obj), {a: 1, b: 2, d: 4});
});
```


CrossEye:
I'd like to know a little more about _why_ you see this as important enough to include.  I don't think I personally have ever reached for a function like this.

Also, the use of both the names and the predicate seems strange to me.  I would imagine that if I wanted this, I would want something similar to `filter`:

```js
const omitWith = useWith(filter, [complement, identity]);
omitWith(gt(__, 2), {a: 1, b: 2, c: 3, d: 1, e: 5}); //=> {"a": 1, "b": 2, "d": 1}
```

But I'd have to be convinced that it's a common enough use case to include in the library.

zatziky:
The `identity` is not necessary, right? `useWith(filter, [complement])`

I see all these methods as a utility belt where I reach whenever I need the code to be more descriptive, more declarative and shorter. 

I would also find handy from time to time `pickIf`. 

A use case can be:

```
R.map(item => {
    const result = {};
    
    if(!test(item.a)) result.a = item.a; 
    if(!test(item.b)) result.b = item.b;
    
    return item;
});

// OR

R.map(R.omitIf(test, ['a', 'b']))
```

CrossEye:
> The identity is not necessary, right? `useWith(filter, [complement])`

It will function, but it will report the wrong arity.  Whether that's important or not to you is your call.


> I would also find handy from time to time `pickIf`.

`filter` already plays that role, without the list of fields:

```js
filter(isOdd, {a: 8, b: 6, c: 7, d: 5, e: 3, f: 0, g: 9}); //=> {"c": 7, "d": 5, "e": 3, "g": 9}
```

Your sample code doesn't match the implementation, I believe.  But even if it did, that seems to just be restating what it does.  What's it for?  When would you use it?

zatziky:
Yes, I am especially interested in the named fields. Below is the real example of what I'd like to achieve. Maybe `omitIf` is really not necessary..? I need to get rid of fields `portsStorage` and `portsNetwork` if they're empty or nil. The field `locations` is an array too. I want to keep it.

```
const hosts = [
    {hostname: 'h1', portsNetwork: [], portsStorage: [], locations: []},
    {hostname: 'h2', portsNetwork: [], portsStorage: [1], locations: []},
    {hostname: 'h3', portsNetwork: [1], portsStorage: [], locations: []},
    {hostname: 'h4', portsNetwork: [1], portsStorage: [1], locations: []}
];

const hostsMapped = R.map(...)(hosts);

expect(hostsMapped).to.deep.eql([
    {hostname: 'h1', locations: []},
    {hostname: 'h2', portsStorage: [1], locations: []},
    {hostname: 'h3', portsNetwork: [1], locations: []},
    {hostname: 'h4', portsNetwork: [1], portsStorage: [1], locations: []}
]);
```

CrossEye:
I'm not convinced that there is a lot of use for this.  To my mind this would make more sense in the [Cookbook][rc].  Feel free to try to convince me, though.

Are there others who want to push for this?  @asaf-romano, @buzzdecafe, @davidchambers, @kedashoe, @paldepind, @raine, @scott-christopher, @TheLudd?



  [rc]: https://github.com/ramda/ramda/wiki/Cookbook

davidchambers:
> To my mind this would make more sense in the [Cookbook][1].

:+1:


scott-christopher:
I'm in two minds because I can't think of a simple way to achieve this using the current functions in Ramda, but I'm also wary of how often this would be used.

If we were to go ahead with this, I'd prefer to see it introduced with some symmetry to existing functions, or do so by the inclusion of some more generic functions to make this possible.

e.g.
```js
// pickWith   :: (a -> Boolean) -> [String] -> {a} -> {a}
pickWith = (f, ks, o) => // TODO

// unpickWith :: (a -> Boolean) -> [String] -> {a} -> {a}
unpickWith = (f, ks, o) =>
  pickWith(complement(f), ks, o)

pick   = pickWith(T)
unpick = unpickWith(T)
```

Alternatively, this could be simpler to implement if `R.unpick` existed:
```js
omitIf = (f, ks, o) =>
  merge(unpick(ks, o), reject(f, pick(ks, o)))
```

zatziky:
@scott-christopher, isn't better to use `omitWith` than `unpickWith`? Or is unpick somehow different from omit?

scott-christopher:
I didn't actually realise `R.omit` already existed, hahaha. Sorry for the confusion. 🙈 

In that case, I'd be inclined to keep this one for the cookbook as:

```js
omitIf = curry((f, ks, o) =>
  merge(omit(ks, o), reject(f, pick(ks, o))));
```

buzzdecafe:
:cookie: :book: 
 thanks

CrossEye:
> didn't actually realise `R.omit` already existed

Well, I did, and I still struggled to come up with a clean formulation!  Nice, @scott-christopher!

scott-christopher:
https://github.com/ramda/ramda/wiki/Cookbook#remove-a-subset-of-keys-from-an-object-whose-associated-values-satisfy-a-given-predicate

PR #1956, opened 2016-10-29
Removed node v0.10 support and Q devDependency

Bradcomp:
Node version 0.10 is EOL at the end of the month per the [LTS Schedule](https://github.com/nodejs/LTS#lts-schedule).

Doing this means no longer having to use an external dependency for testing promise handling functions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 sudo: false
 language: node_js
 node_js:
-- '0.10'
 - '0.12'
 - 4
 - 6
--- a/package.json
+++ b/package.json
@@ -63,7 +63,6 @@
     "js-yaml": "^3.2.5",
     "jsverify": "^0.7.3",
     "mocha": "2.x.x",
-    "q": "^1.1.1",
     "ramda": "0.17.x",
     "rimraf": "~2.3.2",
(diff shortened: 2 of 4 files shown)


CrossEye:
My only concern -- and while it's minor, it should be raised -- is that if we ever get back to browser testing, this would fail in IE * (probably important), Opera Mini (possibly important), and Android Browser 4.4 (probably unimportant):  http://caniuse.com/#feat=promises

Bradcomp:
@CrossEye I agree that it's a valid concern.  I don't know much about browser testing on CI services, but is it possible to use a polyfill when creating the test bundle?

The downside would be if somebody tried to sneak a Promise into the source code then it wouldn't be caught by CI, but that seems negligible.

CrossEye:
> is it possible to use a polyfill when creating the test bundle?

It probably is, and that might be our best bet, and we don't have to worry about it until we start browser testing again.  This should not be a production issue, since it will only arise in environments which actually have some sort of Promise implementations already.

> The downside would be if somebody tried to sneak a Promise into the source code then it wouldn't be caught by CI, but that seems negligible.

Agreed.

PR #1959, opened 2016-10-30
Add pre and post publish npm scripts

### LINKED ISSUE #1505: Make ramda more modular

xaviervia:
Motivation: fix #1505 .

The proposal is to simply copy the `src` folder into root when publishing to npm. The assumptions are:
1. Individual functions will not need UMD––the target is only CommonJS,
2. …so there is no need for building,
3. …and since the individual modules will be CommonJS, it makes sense to do this only for npm.

The scripts are set up so that any files created locally when building will be removed by the postpublish hook, but of course if you build in CI this is unnecessary.

Does it make sense?

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,9 @@
     "posttest": "git checkout -- dist",
     "prebrowser_test": "npm run pretest",
     "browser_test": "testem ci",
-    "postbrowser_test": "npm run posttest"
+    "postbrowser_test": "npm run posttest",
+    "prepublish": "cp -r src/* .",
+    "postpublish": "rm -rf internal && git clean -f && git checkout ."
   },
   "dependencies": {},
   "devDependencies": {


CrossEye:
I've wanted to do this for a long time, and I always thought it was relatively easy, but never quite as simple as this.

I very much like this idea, but cannot test from my mobile.  My only question is whether it still works alongside the recent changes in #1928.

xaviervia:
I tried this by running `npm run prepublish`, then npm linking and it works. I don't think it should conflict with #1928 =)

xaviervia:
(this is what I tried)

```javascript
const map = require('ramda/map')
const range = require('ramda/range')
const {reduce} = require('ramda')

console.log(map(
  (number) => `${number} lala`,
  range(0, 10)
))

console.log(reduce(
  (a, b) => a + b, 0,
  range(0, 10)
))
```

CrossEye:
I would love it if our NPM publish step created a directory to publish from, copied into it whatever artifacts it needed, performed the publication from there, and then deleted the directory.  That just feels safer to me.  Perhaps I'm showing my age.

But it seems that not what we have now, and I don't think I'd want to slow this down to fix that.  It can all be done later.

So, LGTM.  Others?  @buzzdecafe? @davidchambers?

If we do go this route, should we still include the `src` dir for a release with a deprecation note?

xaviervia:
> I would love it if our NPM publish step created a directory to publish from, copied into it whatever artifacts it needed, performed the publication from there, and then deleted the directory.

I fully agree, I think that is the robust way forward, there is no need to ship all the development artifacts to npm. This was a short-term, low-hanging-fruit change to have the simple `ramda/*` module paths =)

> If we do go this route, should we still include the `src` dir for a release with a deprecation note?

I don’t know how deprecation notes are done in npm, but I think it’s a good idea to keep the `src` around since people consuming ramda functions as modules are already using it and that would be a breaking change. I guess it could be updated later to:

- Ship only `ramda/*` modules and discontinue support for `ramda/src/*` and
- …do it with a `publish` folder where the assets for npm are pasted and that gets deleted after publishing.

Deschtex:
What's the status of this?

CrossEye:
Too many changes to incorporate this.  Still many good ideas.

PR #1961, opened 2016-11-01
replace R.lift and R.liftN with fixed-arity versions

davidchambers:
Before:

```purescript
lift :: (*... -> *) -> ([*]... -> [*])
liftN :: Number -> (*... -> *) -> ([*]... -> [*])
```

After:

```javascript
lift :: Functor f => (a -> b) -> f a -> f b
lift2 :: Apply f => (a -> b -> c) -> f a -> f b -> f c
lift3 :: Apply f => (a -> b -> c -> d) -> f a -> f b -> f c -> f d
```

The fixed-arity functions handle at least 99.99% of use cases and have meaningful type signatures! If someone happens to need `lift4` she can trivially define it in terms of `map` and `ap`.

@buzzdecafe expressed enthusiasm for this change in https://github.com/ramda/ramda/pull/1938#issuecomment-257418698.

### DIFF
--- a/src/lift.js
+++ b/src/lift.js
@@ -1,29 +1,21 @@
-var _curry1 = require('./internal/_curry1');
-var liftN = require('./liftN');
+var map = require('./map');
 
 
 /**
- * "lifts" a function of arity > 1 so that it may "map over" a list, Function or other
- * object that satisfies the [FantasyLand Apply spec](https://github.com/fantasyland/fantasy-land#apply).
+ * Alias of [`map`](#map).
  *
  * @func
  * @memberOf R
  * @since v0.7.0
  * @category Function
- * @sig (*... -> *) -> ([*]... -> [*])
- * @param {Function} fn The function to lift into higher context
- * @return {Function} The lifted function.
- * @see R.liftN
+ * @sig Functor f => (a -> b) -> f a -> f b
+ * @param {Function}
+ * @param {*}
+ * @return {*}
+ * @see R.lift2, R.lift3
  * @example
  *
- *      var madd3 = R.lift((a, b, c) => a + b + c);
- *
- *      madd3([1,2,3], [1,2,3], [1]); //=> [3, 4, 5, 4, 5, 6, 5, 6, 7]
- *
- *      var madd5 = R.lift((a, b, c, d, e) => a + b + c + d + e);
- *
- *      madd5([1,2], [3], [4, 5], [6], [7, 8]); //=> [21, 22, 22, 23, 22, 23, 23, 24]
+ *      R.lift(R.inc, Just(2)); //=> Just(3)
+ *      R.lift(R.inc, Nothing); //=> Nothing
  */
-module.exports = _curry1(function lift(fn) {
-  return liftN(fn.length, fn);
-});
+module.exports = map;
--- a/src/lift3.js
+++ b/src/lift3.js
@@ -1,0 +1,29 @@
+var ap = require('./ap');
+var curryN = require('./curryN');
+var map = require('./map');
+
+
+/**
+ * Promotes a curried ternary function to a function which operates on three
+ * [Apply](https://github.com/fantasyland/fantasy-land#apply)s.
+ *
+ * This function is derived from `map` and `ap`.
+ *
+ * @func
+ * @memberOf R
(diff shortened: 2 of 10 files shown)


src/lift3.js
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig Apply f => (a -> b -> c -> d) -> f a -> f b -> f c -> f d
+ * @param {Function}
+ * @param {*}
+ * @param {*}
+ * @param {*}
+ * @return {*}
```
davidchambers: `@param` and `@return` really are worthless for parametrically polymorphic functions. I'd love to get rid of this cruft. :smile:

src/lift.js
```diff
@@ -1,29 +1,21 @@
...
  * @category Function
- * @sig (*... -> *) -> ([*]... -> [*])
- * @param {Function} fn The function to lift into higher context
- * @return {Function} The lifted function.
- * @see R.liftN
+ * @sig Functor f => (a -> b) -> f a -> f b
+ * @param {Function}
+ * @param {*}
+ * @return {*}
+ * @see R.lift2, R.lift3
```
kedashoe: Maybe add `map` to @see, or provide a description beyond "Alias of `map`"?

davidchambers: I made "`map`" a hyperlink. Does that work for you?

scott-christopher:
I'd be in favour of dropping `R.lift` if it is just going to become an alias of `R.map`.

src/lift3.js
CrossEye: I'd love to get rid of JSDoc, but every time I've suggested that, all sorts of fans of it's integration with their tools get very upset!

CrossEye:
I didn't see this before I commented on a related issue here: https://github.com/ramda/ramda/pull/1938#issuecomment-257752500

I think working with user's variadic functions is part of the price of working in Javascript.  I think the only thing significantly wrong with `lift` / `liftN` right now is the signatures, and those could easily be fixed with something like:

```hs
> R.lift :: (a -> b) -> f a -> f b
>           (a -> b -> c) -> f a -> f b -> f c
>           (a -> b -> c -> d) -> f a -> f b -> f c -> f d
>           ...
```

buzzdecafe:
> I think working with user's variadic functions is part of the price of working in Javascript. 

data-free speculation: I would be very surprised if `map`, `lift2`, and `lift3` did not satisfy 99% of use cases. 

Nonetheless, I appreciate your thoughtful comment in https://github.com/ramda/ramda/pull/1938#issuecomment-257752500; perhaps `liftN` may be worth saving

davidchambers:
@CrossEye: :thumbsup: / :thumbsdown: ?

CrossEye:
Sorry, I thought I'd made it clear that I was :-1:.

We're incredibly generous with supporting arities up to 10 in some of these functions, but there are [still requests][ar] to allow more.  I don't want to go back to the `eval` necessary to support an unlimited number, but I'm loathe to restrict this further, or to start offering many similar functions with different arities/names simply because we can't document their signatures comfortably.


  [ar]: https://github.com/ramda/ramda/issues/2028

buzzdecafe:
we can have both: `map`, '`lift2`, 'lift3`, and `liftN`.

davidchambers:
We could do. Currently those who want the fixed-arity versions can use [`R.map`][1]/[`S.map`][2], [`S.lift2`][3], and [`S.lift3`][4]. It's okay for Ramda to be more Clojure-like and for Sanctuary to be more Haskell-like. I'll stop pressing for Ramda to become more Haskell-like.


[1]: http://ramdajs.com/0.23.0/docs/#map
[2]: https://sanctuary.js.org/#map
[3]: https://sanctuary.js.org/#lift2
[4]: https://sanctuary.js.org/#lift3

CrossEye:
> I'll stop pressing for Ramda to become more Haskell-like.

Honestly, I hope you don't.  I think Haskell should be one of the primary motivators for Ramda.  But I don't think of it as the only one.  And I do think that JS is closer in spirit to LISP than to ML, but clearly we're taking major ideas from ML-style languages as well.  I hope we continue to do so.

PR #1971, opened 2016-11-11
Update logical function documentation.

kid-icarus:
The existing documentation is confusing, as it seems to imply that params and return values are of Boolean type.

This is not the case, as both functions work in the same way that their operator counterparts do.

### DIFF
--- a/src/and.js
+++ b/src/and.js
@@ -8,10 +8,10 @@ var _curry2 = require('./internal/_curry2');
  * @memberOf R
  * @since v0.1.0
  * @category Logic
- * @sig * -> * -> *
- * @param {Boolean} a A boolean value
- * @param {Boolean} b A boolean value
- * @return {Boolean} `true` if both arguments are `true`, `false` otherwise
+ * @sig a -> b -> a | b
+ * @param {Any} a
+ * @param {Any} b
+ * @return {Any} the first argument if it is falsy, otherwise the second argument.
  * @see R.both
  * @example
  *
--- a/src/or.js
+++ b/src/or.js
@@ -9,10 +9,10 @@ var _curry2 = require('./internal/_curry2');
  * @memberOf R
  * @since v0.1.0
  * @category Logic
- * @sig * -> * -> *
- * @param {Boolean} a A boolean value
- * @param {Boolean} b A boolean value
- * @return {Boolean} `true` if one or both arguments are `true`, `false` otherwise
+ * @sig a -> b -> a | b
+ * @param {Any} a
+ * @param {Any} b
+ * @return {Any} the first argument if truthy, otherwise the second argument.
  * @see R.either
  * @example
  *


buzzdecafe:
Fair point. However, we are trying to move away from the "information-free" asterisk, to using the imaginary "Any" type instead. Maybe something like `Any a -> Any b -> Any a | Any b`? Gross, I know, but hey, it's JavaScript

kid-icarus:
No worries, updated accordingly. Let me know if I misunderstand your request :)

src/or.js
```diff
@@ -9,10 +9,10 @@ var _curry2 = require('./internal/_curry2');
...
  * @since v0.1.0
  * @category Logic
- * @sig * -> * -> *
- * @param {Boolean} a A boolean value
- * @param {Boolean} b A boolean value
- * @return {Boolean} `true` if one or both arguments are `true`, `false` otherwise
+ * @sig Any a -> Any b -> Any a | Any b
+ * @param {Any} a
+ * @param {Any} b
+ * @return {Any} the first argument if falsy otherwise the second argument.
```
buzzdecafe: the first argument if **truthy**

src/and.js
```diff
@@ -8,10 +8,10 @@ var _curry2 = require('./internal/_curry2');
...
  * @since v0.1.0
  * @category Logic
- * @sig * -> * -> *
- * @param {Boolean} a A boolean value
- * @param {Boolean} b A boolean value
- * @return {Boolean} `true` if both arguments are `true`, `false` otherwise
+ * @sig Any a -> Any b -> Any a | b
+ * @param {Any} a
+ * @param {Any} b
+ * @return {Any} the first truthy argument, otherwise the last argument.
```
buzzdecafe: the first argument if it is falsy, otherwise the second argument.

kid-icarus: My mistake, updated 👍

kid-icarus:
My mistake, updated 👍

buzzdecafe:
this appears to be a virtual reversion of https://github.com/ramda/ramda/pull/1335, so let's proceed with caution. @paldepind and @crosseye, any interest in re-litigating this?

CrossEye:
That was a long hard-fought battle that I seemed to have lost at the end.  My final word on this was in https://github.com/ramda/ramda/pull/1335#issuecomment-134471245, which hasn't changed since.  I would still/ very much be in favor of this change.

I think the signatures are still off though.  `Any` is not necessary.

```js
// :: a -> b -> a | b
```

looks reasonable to me for both of them.

TheLudd:
My thoughts on the matter is that we should not focus on what *may* be done with the functions but what we *intend* for the user to do with the functions. The way I think of them are:
```
Boolean -> Boolean -> Boolean
```

buzzdecafe:
@TheLudd there is a tension in the sigs between what they actually do and what they are documented as doing. I agree with you that the aspirational signature is `Bool -> Bool -> Bool`. But that is surprising when it _doesn't actually behave that way_. 

As I see it, our options are:
* Stick with the aspirational `Bool` signature and _caveat emptor_
* Document the actual behavior, as this PR does
* Cast the return value to `Bool` so you get `a -> b -> Bool`

TheLudd:
I prefer the first. I believe that if we go with option 2 then we are far from consistent when it comes to the documentation of this library. 

I also think that option 2 will create expectations on these functions that we may feel then need to fulfill in the future which in turn will make the library more complex. I can imagine issues like

 * "Since `and` supports any value I would expect it to..."
 * "Since `and` supports any value I would expect function X to..."

And I don't want to go down that tunnel =)

TheLudd:
But sure you can also do 
```
function and (a, b) {
  return a === true && b === true
}
```

CrossEye:
Although I have no problem with this PR as it stands, I could also live with this:

> * Cast the return value to `Bool` so you get `a -> b -> Bool`

amsross:
For what it's worth, I've stumbled over this myself, and witnessed two others others stumble over it in just the past week and try to rationalize it in Gitter.

If the resolution is "Stick with the aspirational `Bool` signature and *caveat emptor*," then having the docs reflect that decision would probably be a positive. As it is now, the docs reflect the intention but not the behavior, which sounds fine to me on a theoretical level, but in practice is causing problems for real users.

CrossEye:
I'd like to switch the signatures to 

```js
:: a -> b -> a | b
```

and be done with this.  What do others think?  @buzzdecafe, @paldepind, @TheLudd, @davidchambers, @scott-christopher?

TheLudd:
I think this is my least favorite. It can return type `a` or `b`?

If we're going that route I'd prefer
```
a -> a -> a 
```
Indicating that you should at leas be selecting between to values of the same type. But I would prefer the booleans.

CrossEye:
I could also live with updating the implementation to force the result to a boolean to give

```js
//:: a -> b -> Bool
```

But I really am not interested in forcing the user to pass only Booleans or even the same type.  Here perhaps the aspirational type could be 

```js
//:: a -> a -> Bool
```

as that will not cause many heartaches.  But I don't particularly like the way that this signature would work with, say

```js
and(1, 'foo'); //=> true
```

TheLudd:
What I don't get is what is this function used for? I mean, what does "and" mean in these contexts:
```
and(10, 15) //=> 15
and(0, 15) //=> 0
and(15, 0) //=> 0
```
I could not imagine using this function in my code if I did not give it Boolean input and expected Boolean output.

CrossEye:
That's a lot of what #1218 and #1335 were about.

In my mind, `and` simply converts the `&&` infix operator to a prefix function.  I've been in plenty of situations for instance where a function should return a zero in one circumstance, a non-zero number in others, but where it corner cases it might also return `null`, `undefined`, or possibly `NaN`, each of which should be treated similarly to zero.  In those cases, I might well combine this result with something else via `and`.

kid-icarus:
Yes, these functions simply offer prefix function version of infix operators. That's it. That's what this PR is trying to reflect.

TheLudd:
Well. I think I have made my position clear but I don't know if there is any official stance ramda has on matters like these and perhaps that is missing. If a function has a certain intent, I think it should document that because that is what it supports. If it turns out that the implementation supports some functionality that was not intended, I don't think that should be added to the documentation because then we have to support that as well and watch out for breaking changes.

Should we document this as well? It works, doesn't throw any errors.... https://goo.gl/IVSn8B

TheLudd:
Another thing. 
If you want the `or` behavior for non boolean values there is always `defaultTo`. Granted, it only checks for `null`, `undefined` and `NaN` but I find that good since I don't like the conceps of "truthy" and "falsey".

Here I think we have a clear case of 2 functions that are similar on the surface but are used for different purposes and thus named better.

kid-icarus:
Is the intention of these functions to do anything besides offer prefix function versions of infix operators?

I think I'm not understanding the intention of these functions.

buzzdecafe:
i agree with @TheLudd, it would be cleaner to *at least* enforce `Bool` return value. 

For JS coercion semantics, we'd be better with new functions `andy` and `ory`

TheLudd:
I agree that the intention is unclear. 

I think the intention should be to only work with Booleans. I would not mind enforcing it in code either. Descriptions will then be simple

 * `and`: returns `true` if both arguments are `true`, `false` otherwise
 * `or`: returns `true` if at least one argument is `true`, `false` otherwise

kid-icarus:
I suspect there may be some users using `and` and `or` relying on the fact that the return type isn't coerced.

Even though there is a `defaultTo` function, it's a common pattern in JS to use `||` to accomplish the same thing, in fact `defaultTo` uses `||`.

I'm in favor of returning a boolean, but keep in mind it will be a breaking change for people using and/or to assign non-boolean values.

CrossEye:
I suspect that coercing the return type would cause little issue.  But forcing the parameters to booleans would be a much more drastic, and more likely breaking, change.

But let's not lose sight of the fact that this is chiefly a documentation issue.  For those who want to use these as

```hs
:: Bool -> Bool -> Bool
```

these functions *already work* without change.

TheLudd:
Common patterns are not always good reasons. It is a common pattern to have the arguments in opposite order to ramda =)

Default to does not use `||` in the way we are talking about here. It actually checks two booleans in order to determine whether it should return the first or second argument to the function.

kid-icarus:
Can a separate issue be filed for arguing over how these functions operate? I don't really care to debate that here.

All I'm asking for is up to date documentation on how these functions _currently operate, regardless of intention._

These functions _do not necessarily_ return booleans, that is all I am trying to communicate. A coworker reviewing some code thought it incorrect because they looked at existing ramda documentation. The existing documentation was incorrect.

davidchambers:
> What do others think?

sanctuary-js/sanctuary#279

CrossEye:
@kid-icarus: would you be willing to update to remove the `Any`s, leaving just `a -> b -> a | b`?

Unless there's a really major objection, I think we can merge that, so at least people aren't getting the confusion seen here and on Gitter.  Then anyone who really wants to push for `Bool -> Bool -> Bool` can open a separate PR and we can re-fight that battle.  :cry:

TheLudd:
> All I'm asking for is up to date documentation on how these functions currently operate, regardless of intention

I was trying to argue against just that =)

CrossEye:
>> All I'm asking for is up to date documentation on how these functions currently operate, regardless of intention

> I was trying to argue against just that =) 

I'm often with you on this, @TheLudd.  

But I make the distinction here between signatures that report our aspirations for the function while ignoring accidental implementation details (`map(toUpper, 'foo'); //=> ['F', 'O', 'O']`) and ones which intentionally obscure what the functions do (turning `&&` and `||` into plain functions.)

The goal of these functions, ever since they swiped their names from `either` and `both` back in #843, was exactly to replicate the behavior of the native operators.  There was an attempt in #1218 to change this, which I fought.  But I didn't fight hard enough in the follow-up, #1335, to avoid changing the documentation to make them *look* like they acted differently than they were designed.  As this is now causing confusion  (several people on Gitter, here, #1979), I believe it is time to rectify it.

TheLudd:
Allright, well I have made my position clear. Enough said from me on this
topic

2016-11-17 16:30 GMT+01:00 Scott Sauyet <notifications@github.com>:

kid-icarus:
@CrossEye I have removed the Anys from the signature, did you want them removed from the `@param` and `@return` tags as well?

CrossEye:
@kid-icarus: no, that looks good to me.  We'll see if anyone wants to reopen the deeper issue about functionality.

CrossEye:
There was such a kerfuffle about this that I'm loathe to revisit it.  But I could live with changing to something like

```js
const and = (a, b) => Boolean(a && b)
```

It's not what I'd prefer, but I am tired of revisiting this.  Others?

Issue #1972, opened 2016-11-12
Should reduceRight's argument: fn flip its args?


adispring:
Haskell foldr transformer's arguments sequence is `\x acc -> ...`,  accumulator is the second argument, and `reduceRight`'s is `fn(acc, x)`, should we flip` fn(acc,x)` to `fn(x, acc)` ?

CrossEye:
That Haskell does it differently is a useful data point, but not entirely convincing on its own.  Care to try to make an argument for the change?

paldepind:
Very good question @adispring.

Since we order arguments to our functions so that they are most useful for currying the "structured" argument is last. However, if one is building a new structure with `reduceRight` then the function in `reduceRight` will get the "structured" argument _first_ which is less inconvenient.

Two examples where changing the order works better:

```js
concat = (list1, list2) => R.reduceRight(R.append, list2, list1);
R.reduceRight(R.flip(R.prop)(true), {}, ["foo", "bar", "foo"]); //=> {foo: true, bar: true}
```

Changing the order also increases the symmetry between `reduce` and `reduceRight`. Reduce can be understood as placing the function between the elements in the list. If our functions `reduce` and `reduceRight` were like Haskell's the difference between the two would only be a matter of how one places parentheses. I.e. the below would be accurate.

```js
R.reduce(R.add, 0, [1, 2, 3, 4]) = (((0 + 1) + 2) + 3) + 4
R.reduceRight(R.add, 0, [1, 2, 3, 4]) = 0 + (1 + (2 + (3 + 4)))
```

Changing whether the folding function associates to the left or to the right also changes which of its arguments will be the accumulator.

However, while one can say that symmetry is increased one could also say that from another perspective something is lost because the signatures of the functions will be more different.

CrossEye:
> If our functions `reduce` and `reduceRight` were like Haskell's the difference between the two would only be a matter of how one places parentheses. I.e. the below would be accurate.
> 
> ```js
> R.reduce(R.add, 0, [1, 2, 3, 4]) = (((0 + 1) + 2) + 3) + 4
> R.reduceRight(R.add, 0, [1, 2, 3, 4]) = 0 + (1 + (2 + (3 + 4)))
> ```

That to me is pretty convincing.  Even more interesting would be if the operation wasn't commutative, say `R.subtract`.

adispring:
Yes, the result will be changed if the computation depends on arguments order, like `R.subtract`
[...]

PR #1975, opened 2016-11-15
flip reduceRight reducer arguments order

adispring:


CrossEye:
See #1972 for discussion.

src/reduceRight.js
```diff
@@ -37,7 +37,7 @@ var _curry3 = require('./internal/_curry3');
 module.exports = _curry3(function reduceRight(fn, acc, list) {
   var idx = list.length - 1;
   while (idx >= 0) {
-    acc = fn(acc, list[idx]);
+    acc = fn(list[idx], acc);
     idx -= 1;
   }
   return acc;
```
CrossEye: I think the comments should also be updated, for this one and for `reduce`, probably with examples that between them demonstrate how `reduce` and `reduceRight` are fundamentally different.  Perhaps something like:

```js
R.reduce((a, b) => a - b, 0, [1, 2, 3, 4]); //=> (((0 - 1) - 2) - 3) - 4) => -10
R.reduceRight((a, b) => a - b, 0, [1, 2, 3, 4]); //=> 0 - (1 - (2 - (3 - 4))) => 2
```

test/reduceRight.js
```diff
@@ -21,7 +21,7 @@ describe('reduceRight', function() {
     var something = R.reduceRight(avg, 54);
     var rcat = R.reduceRight(R.concat, '');
     eq(something([12, 4, 10, 6]), 12);
-    eq(rcat(['1', '2', '3', '4']), '4321');
+    eq(rcat(['1', '2', '3', '4']), '1234');
   });
```
CrossEye: Similarly here.  I think the existing test cases no longer show anything of particular use.  But one that demonstrates the tree structure would be useful.

src/reduceRight.js
adispring: OK, I'll update the comment. The procedure of `reduceRight` should be as follows:
```js
R.reduce((a, b) => a - b, 0, [1, 2, 3, 4]); //=> (((0 - 1) - 2) - 3) - 4) => -10
R.reduceRight((a, b) => a - b, 0, [1, 2, 3, 4]); //=>  (1 - (2 - (3 - (4 - 0)))) => -2
```

test/reduceRight.js
adispring: I'll think about how to make an demonstrates.

paldepind:
I think we should also change `R.mapAccumRight` for consistency. They're closely related functions.

src/reduceRight.js
```diff
@@ -20,24 +20,42 @@ var _curry3 = require('./internal/_curry3');
...
+ *      R.reduce(R.subtract, 0, [1, 2, 3, 4]) // => ((((0 - 1) - 2) - 3) - 4) = -10
+ *                  -10
+ *                 /
+ *                -6 - 4
+ *               /
+ *              -3 - 3
+ *             /
+ *            -1 - 2
+ *           /
+ *          0 - 1
```
CrossEye: I find these diagrams a bit difficult to follow.  How about something like

```
  -              -2        
 / \             / \        
1   -           1   3       
   / \             / \      
  2   -    ==>    2  -1    
     / \             / \    
    3   -           3   4   
       / \             / \  
      4   0           4   0 
```
and
```
        -              -10
       / \             / \
      -   4          -4   4
     / \             / \
    -   3    ==>   -3   3
   / \             / \
  -   2          -1   2
 / \             / \
0   1           0   1
```
?  Also, I think the `reduce` example belongs with the `reduce` documentation.  Those who want to compare them can switch back and forth, especially since the two functions link to one another.

adispring: ok, I'll modify it, and add this comment to `reduce`.

I also changed the `mapAccumRight` iterator function arguments' order, can you and @paldepind review that?

src/reduceRight.js
```diff
@@ -20,24 +21,44 @@ var _curry3 = require('./internal/_curry3');
...
  * @param {*} acc The accumulator value.
  * @param {Array} list The list to iterate over.
  * @return {*} The final, accumulated value.
- * @see R.addIndex
+ * @see R.reduce, R.addIndex
  * @example
  *
- *      var pairs = [ ['a', 1], ['b', 2], ['c', 3] ];
- *      var flattenPairs = (acc, pair) => acc.concat(pair);
+ *      R.reduceRight(R.subtract, 0, [1, 2, 3, 4]) // => (1 - (2 - (3 - (4 - 0)))) = -2
```
adispring: I've modify the diagram which shows the difference of reduce and reduceRight.

src/reduce.js
```diff
@@ -27,13 +29,32 @@ var _reduce = require('./internal/_reduce');
...
+ *          -               -2
+ *         / \              / \
+ *        1   -            1   3
+ *           / \              / \
+ *          2   -     ==>    2  -1
+ *             / \              / \
+ *            3   -            3   4
+ *               / \              / \
+ *              4   0            4   0
  *
```
CrossEye: I don't think that the functions need to have both diagrams.  People interested in the difference can look back and forth between the `reduce` and `reduceRight` docs for this.

adispring: I've changed this.

CrossEye:
This looks great to me. I'd like to merge it.  Any objections?  @asaf-romano, @buzzdecafe, @davidchambers, @kedashoe, @paldepind, @raine, @scott-christopher, @TheLudd?

scott-christopher:
I do prefer the argument order as per this change, but we'll just have to be quite vocal about the breaking change if this lands.

I also tend to think `reduceRight` would be more useful if the the accumulator was wrapped in a nullary function to allow for lazy evaluation, but that's beyond the scope of this PR.

scott-christopher:
I also really like the inclusion of the tree diagrams 😀

TheLudd:
After this, would we also want `transduceRight`? 
That could mean no more flipping of functions, for instance 
```
// today
transduce(someMapFn, R.flip(R.append), [], someList)
// tomorrow
transduceRight(someMapFn, R.prepend, [], someList)
```

adispring:
Because `transduce`'s implementation depends on `_reduce`, if implement `transduceRight`, we should first enhance `reduceRight`. By enhancing `reduceRight`,  `reduceRight` can support `reduced` in passing.

paldepind:
LGTM :+1: :tada:

PR #1977, opened 2016-11-15
es6 modules

jonaskello:
This is a proof of concept for a re-write to es6 modules.

The files in `src` folder can now use es6 syntax for import/export. By then running `npm run babel` the files in the `src`folder will be compiled into the `cjs` folder but the only transform that will be applied is that es6 style import/export will be rewritten to commonjs style. By adding more options to `.babelrc` more transforms could be added. For example the files in the `src` folder could use es6 syntax if the appropriate options are added.

The main index.js file has been rewritten so it exports from the `cjs` folder. This makes running `npm test` work without modifing the tests.


jonaskello:
It might be hard to see the actual changed becuase of all the added files from the generated `cjs` folder. Maybe I should exclude the `cjs` folder?

jonaskello:
So far es6 style exports has only been added to `src/add.js`:

```
-module.exports = _curry2(function add(a, b) {
+export default _curry2(function add(a, b) {

```

CrossEye:
I'm curious about the [output format][of].  Is there some babel magic that would simplify it?

The current, hand-crafted CommonJS module looks like this:

> ```js
> var _curry2 = require('./internal/_curry2');
> 
> // [ ... API documentation comments ... ]
> 
> module.exports = _curry2(function add(a, b) {
>   return Number(a) + Number(b);
> });
> ```

The generated one looks like this:

> ```js
> 'use strict';
> 
> Object.defineProperty(exports, "__esModule", {
>   value: true
> });
> 
> // [ ... API documentation comments ... ]
> 
> var _curry2 = require('./internal/_curry2');
> 
> exports.default = _curry2(function add(a, b) {
>   return Number(a) + Number(b);
> });
> module.exports = exports['default'];
> ```

That's a fair bit of cruft.  Is it all necessary?


  [of]: https://github.com/jonaskello/ramda/blob/a4801687c01d1ce3533c36a3d030f9b2b9ec0ddd/cjs/add.js

lpan:
@CrossEye both Webpack 2 and Rollup support native es6 modules. That's why we need to [set `modules` to be `false`](https://babeljs.io/docs/plugins/preset-es2015/#options) when we want to take advantage of the tree shaking feature.

jonaskello:
Yes, the output is interesting. I added some options to remove some of the extra stuff:

```js
'use strict';
```
This was just to make all files have strict mode. I added a babel option to not have this.
```js
Object.defineProperty(exports, "__esModule", {
  value: true
});
```
This part is interesting. AFAIR this adds a flag property that is used by some module loaders (eg. systemjs). It says that this is a cjs module compiled down from a es6 module. It somehow helps the module loader when it loads the cjs module. If a module loader such as systemjs is not used I don't think it is relevant. I added a babel option that outputs a single line instead. I could not find any option to remove it completely but OTOH I did not spend much time trying.

Anyway with the latest commit the output is now a bit smaller but I'm not sure it is relevant. I'll try to explain why next.

jonaskello:
Regarding the outputted `cjs` folder. 

If/when all the internal files in ramda are updated to use es6 style modules the `cjs` folder will not be used internally anymore (right now it is used by the tests since I have not re-written them to es6 style imports). Then the `cjc` folder will simply be a build artifact that will only be used by external applications that want to import a single commonjs module with `const add = require('ramda/cjs/add')`.

So then we could think of the files in the `cjs` folder as a build artifact in the same sense as the files in the `dist` folder. Basically each file in the `cjs` folder is a mini-bundle of ramda with only the parts that are available in that file. For example when you do `const add = require('ramda/cjs/add')` you are requiring the add mini-bundle. I think lodash has the ability to create these kind of mini-bundles or special builds or whatever they call it. The point being that you want to import something smaller than the full bundle.

Anyway, if the files in `cjs` is only a build artifact I don't think it is necessary to have them optimized for human-readability for the same reason the files in `dist` are not optimized to be human-readable. In fact I think we could enable babel options to minify each file in the `cjs` folder.

iansinnott:
This is really interesting and could definitely provide some wins for anyone using tree shaking. If output format is a concern Babel's [loose option](http://www.2ality.com/2015/12/babel6-loose-mode.html) might help. 

The combination of loose mode and turning off ES module transformation is what the [babel-preset-env repo](https://github.com/babel/babel-preset-env) [currently recommends](https://github.com/babel/babel-preset-env#examples).

Example from the links above:

```json
{
  "presets": [
    ["env", {
      "targets": {
        "browsers": "last 2 versions"
      },
      "modules": false,
      "loose": true
    }]
  ]
}
```

jonaskello:
Currently this PR only demonstrates what is possible. So as demonstrated a `cjs` folder with files that look as close as possible to the original could be created.  Also this could be created at publish time and copied to the root so it would be possible to do `const map = require('ramda/map')`.

To do a full conversion I would need to invest some more work, going through all files and rewriting them. I'll be happy to do that but only if there is genuine interest to merge this eventually. So I'll leave this for now and see what the decision will become.

jpsc:
@jonaskello I could also help you want to move forward with this.
I think we can build those cjs files when you run the test only and add them to gitignore.

jonaskello:
@jpsc Did you comment on this? I got an email with your comment but I don't see the comment here. Anyway if you want to help I could make you a collaborator on my fork.

jpsc:
I did comment but then I retracted because I saw the discussion on the issues tab.
I didn't know about the current `ramda/src/map` option.

rickmed:
what happened to this? It looks super interesting.

alvaro450:
Hi guys, is there any update on this PR? Is this PR being considered to be merged?

buzzdecafe:
Hi, i ES6-izing the codebase is a post-1.0 priority. But we gotta get 1.0 out the door first. 
ping @kedashoe

kedashoe:
It is clear a lot of people would like to see this (myself included). I'm sure quite a bit has changed since this was initially submitted. @jonaskello , would you like to open a new PR based off current master with all the latest from rollup? I guess based off https://github.com/rollup/rollup-starter-lib and we can go from there with any recommendations from the ramda/rollup communities. I think we can leave out babel/buble for now and tackle that as a separate pull request if we decide to go in that direction. Sorry for the long delay and a big thanks to everyone for the continued input on this topic.

jonaskello:
Closing this in favor of #2254.

PR #1990, opened 2016-12-01
Add @see R.converge in useWith.js, and vice versa

adispring:


CrossEye:
I absolutely agree.  We should add these links.  I'm curious, though, as to why all these commits are showing up together above.  This should be the simplest of merges...

adispring:
Don't know the reason.
I pull force from ramda/ramda master branch, cherry-pick this commit, and create a new clean pr.

CrossEye:
ok, thanks.

PR #1991, opened 2016-12-01
add @see R.converge in useWith.js, and vice versa

adispring:


CrossEye:
Very good idea.  Thanks!

adispring:
😀

PR #1993, opened 2016-12-02
Add R.lensFind

bionicbrian:
The intention here is a mash up of `R.find` and `R.lens`.

Questions I asked myself before submitting this:

* Is the same effect straightforward by composing existing functions? _Not really_.
* Has someone already proposed a function like this? _Not that I can see_. However, I found out after fact that [this S.O. question](http://stackoverflow.com/questions/35538351/ramda-js-lens-for-deeply-nested-objects-with-nested-arrays-of-objects/35544228#35544228) is answered with a great lens written by [Scott Christopher](https://github.com/scott-christopher), which only tells me:
    - a). Other people probably need to do this sort of thing.
    - b). Scott already wrote a `lensMatching` function. Did he or someone else already consider adding it? Maybe I can close this and he can propose it? :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -106,6 +106,7 @@ module.exports = {
   length: require('./src/length'),
   lens: require('./src/lens'),
   lensIndex: require('./src/lensIndex'),
+  lensFind: require('./src/lensFind'),
   lensPath: require('./src/lensPath'),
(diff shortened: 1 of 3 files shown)


bionicbrian:
Ugh, spacing issue. Will resubmit.

PR #1994, opened 2016-12-02
Add R.lensFind

bionicbrian:
The intention here is a mash up of `R.find` and `R.lens`.

Questions I asked myself before submitting this:

* Is the same effect straightforward by composing existing functions? _Not really_.
* Has someone already proposed a function like this? _Not that I can see_. However, I found out after fact that [this S.O. question](http://stackoverflow.com/questions/35538351/ramda-js-lens-for-deeply-nested-objects-with-nested-arrays-of-objects/35544228#35544228) is answered with a great lens written by [Scott Christopher](https://github.com/scott-christopher), which only tells me:
    - a). Other people probably need to do this sort of thing.
    - b). Scott already wrote a `lensMatching` function. Did he or someone else already consider adding it? Maybe I can close this and he can propose it? :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -105,6 +105,7 @@ module.exports = {
   lastIndexOf: require('./src/lastIndexOf'),
   length: require('./src/length'),
   lens: require('./src/lens'),
+  lensFind: require('./src/lensFind'),
   lensIndex: require('./src/lensIndex'),
   lensPath: require('./src/lensPath'),
   lensProp: require('./src/lensProp'),
--- a/src/lensFind.js
+++ b/src/lensFind.js
@@ -1,0 +1,37 @@
+var __ = require('./__');
+var _curry3 = require('./internal/_curry3');
+var always = require('./always');
+var findIndex = require('./findIndex');
+var map = require('./map');
+var update = require('./update');
+
+
+/**
+ * Returns a lens whose focus is the first element that matches the predicate
+ *
+ * @func
+ * @memberOf R
+ * @category
+ * @typedefn Lens s a = Functor f => (a -> f a) -> s -> f s
+ * @sig (a -> Boolean) -> Lens s a
+ * @param {Function} fn The predicate function used to determine if the element is the
+ *        desired one.
+ * @return {Lens}
+ * @see R.view, R.set, R.over
+ * @example
+ *
+ *      var xs = [{a: 1}, {a: 2}, {a: 3}];
+ *      var xLens = R.lensFind(R.propEq('a', 1));
+ *
+ *      R.view(xLens, xs);                    //=> {a: 1}
+ *      R.set(xLens, {b: 4}, xs);             //=> [{b: 4}, {a: 2}, {a: 3}]
+ *      R.over(xLens, R.assoc('a', -1), xs);  //=> [{a: -1}, {a: 2}, {a: 3}]
+ */
+
+module.exports = _curry3(function lensFind(pred, toFunctorFn, list) {
+  var idx = findIndex(pred, list);
+  return map(
+    idx >= 0 ? update(idx, __, list) : always(list),
+    toFunctorFn(list[idx])
+  );
+});
--- a/test/lensFind.js
+++ b/test/lensFind.js
@@ -1,0 +1,91 @@
+var R = require('..');
+var eq = require('./shared/eq');
+
+
+describe('lensFind', function() {
+  var list1 = ['a', 'b', 'c'];
+  var obj1 = {a: 1};
+  var obj2 = {b: list1};
+  var testList = [10, obj1, obj2, 100];
+  var even = function(x) { return x % 2 === 0; };
+
+  describe('view', function() {
+    it('focuses the first list element found by the predicate', function() {
+      eq(R.view(R.lensFind(R.propEq('a', 1)), testList), obj1);
+      eq(R.view(R.lensFind(even), testList), 10);
+      eq(R.view(R.lensFind(R.lt(99)), testList), 100);
+    });
+    it('returns undefined if the element is not found', function() {
+      eq(R.view(R.lensFind(R.equals(1000)), testList), undefined);
+    });
+  });
+
+  describe('set', function() {
+    it('sets the list value found by the predicate and returns a new list', function() {
+      var newObj1 = {a:0};
+      var newList = R.update(R.findIndex(R.propEq('a', 1), testList), newObj1, testList);
+
+      eq(
+        R.set(
+          R.lensFind(R.propEq('a', 1)),
+          newObj1,
+          testList
+        ),
+        newList
+      );
+
+      eq(testList, [10, obj1, obj2, 100]);
+    });
+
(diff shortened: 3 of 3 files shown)


src/lensFind.js
```diff
@@ -0,0 +1,36 @@
...
+ *      R.over(xLens, R.assoc('a', -1), xs);  //=> [{a: -1}, {a: 2}, {a: 3}]
+ */
+
+module.exports = _curry3(function lensFind(pred, toFunctorFn, list) {
+  var i = findIndex(pred, list);
+  return map(
+    update(i, __, list),
+    toFunctorFn(list[i])
+  );
+});
```
CrossEye: One nit, and one more serious concern.

The nit is that we generally use `idx` not `i` for such variables.

The bigger concern has to do with when the value is *not* found.  I don't know enough of the theory behind lenses to understand this note from @scott-christopher: 

> (note: this will only be a valid lens if it is guaranteed to match at least one element of the list)

I'm not really sure what makes for a "valid lens".  But this implementation (and Scott's as well?) would have problems if there isn't a match:

```js
const xLens = lensFind(R.propEq('a', 5));
set(xLens, {b: 20}, [{a: 0}, {a: 1}, {a: 2}, {a: 3}])
//=> [{"a": 0}, {"a": 1}, {"a": 2}, {"b": 20}]
```

... since `update` uses negative indices as counting from the end, but `findIndex` returns `-1` as the result of not finding a value.  Perhaps something like this would be worth doing:

```js
  return map(
    idx > -1 ? update(idx, __, list) : always(list),
    toFunctorFn(list[idx])
  );
```

But I have no idea how that fits with the theory of what makes a valid lens.

CrossEye:
I like this idea, and mostly like the implementation.  I think this version is easier to understand than the one from the StackOverflow answer.  I made a suggestion inline for an alteration, but am not at all confident on whatever theory underlies lenses.  Any comments @davidchambers, @scott-christopher, @joneshf?

davidchambers:
I believe one of the lens laws requires that `R.over(lens, R.identity, x)` be equivalent to `x`. We currently make no attempt to ensure the lenses we create are lawful. For example:

```javascript
> R.over(R.lensProp('x'), R.inc, {})
{x: NaN}
```

Given the existence of `lensProp` it would be inconsistent to reject `lensFind` for not obeying the identity law. We should at least document the "invariant", though.

CrossEye:
> Given the existence of `lensProp` it would be inconsistent to reject `lensFind` for not obeying the identity law. We should at least document the "invariant", though.
 
I can buy that.

@bionicbrian: what do you think of those suggested changes?

bionicbrian:
Hi @CrossEye and @davidchambers 

Thanks for the quick feedback! I hope I understood it correctly.

I have made an update so that `set` and `over` return the original list when a.) the value is not found or b.) when the setter does not change the value. I have added a few tests covering `set` when a value is not found, "Putting back what you got doesn't change anything", and "Setting twice is the same as setting once". :)

Hopefully this makes sense. Thanks!

src/lensFind.js
```diff
@@ -0,0 +1,40 @@
...
+ *      R.view(xLens, xs);                    //=> {a: 1}
+ *      R.set(xLens, {b: 4}, xs);             //=> [{b: 4}, {a: 2}, {a: 3}]
+ *      R.over(xLens, R.assoc('a', -1), xs);  //=> [{a: -1}, {a: 2}, {a: 3}]
+ */
+
+module.exports = _curry3(function lensFind(pred, toFunctorFn, list) {
+  var idx = findIndex(pred, list);
+  return map(
+    function(x) {
+      if (idx > -1 && x !== list[idx]) {
```
davidchambers: Ramda style is `>= 0` rather than `> -1`.

`x !== list[idx]` strikes me as an unnecessary optimization. Furthermore, identity-based equality is a source of bugs, so let's avoid it. :)

bionicbrian: Sounds good. Will update.

bionicbrian:
Ok, @davidchambers updated with Ramda style index check.

I have removed the equality check. The intention was to leave the array alone if the new element is the same as the old one, saving the call to update but also trying to ensure the same array is returned. That said, the tests related to `R.identity` still pass with `eq`, which appears to do a value comparison, but the references are different.

Thanks again! :)

index.js
```diff
@@ -106,6 +106,7 @@ module.exports = {
   length: require('./src/length'),
   lens: require('./src/lens'),
   lensIndex: require('./src/lensIndex'),
+  lensFind: require('./src/lensFind'),
```
davidchambers: :abc:

bionicbrian: Coo. Sorry bout that. Changed.

CrossEye:
:herb:

CrossEye:
Any other feedback?  Anyone? Anyone? ~~Bueller?~~ @buzzdecafe?

polytypic:
`find` was one of the first lenses I wrote when we started using Ramda's lenses with reactive variables to deal with complex UI model objects and which lead to the development of my [partial.lenses](https://github.com/calmm-js/partial.lenses) library.  The [tutorial example](https://github.com/calmm-js/partial.lenses#tutorial), which is extracted from a real use case, also uses `find`.

Now, the most important thing to understand here is that `find` is not a lens aka total lens.  I briefly discuss partiality in the section [on partiality](https://github.com/calmm-js/partial.lenses#on-partiality) and also in several other sections of the documentation.  There are various ways to deal with partiality.  In partial.lenses the idea is to use that for something useful: using partiality to be able to *insert* and *remove* elements and to *simplify compositions*.  Partiality also allows a functor based implementation of lenses to support (partial) [isomorphisms](https://github.com/calmm-js/partial.lenses#isomorphisms), because it allows you to [use *partial* lenses to construct minimal data structures](https://github.com/calmm-js/partial.lenses#L-getInverse) out of `undefined`.

The choice taken in this PR, in case when no matching element is found, is to turn the lens into a kind of read-only lens.  In such a case the lens will not fulfill the [set-get (aka put-get) law](https://calmm-js.github.io/documentation/compositional-data-manipulation-using-lenses/#/2/4) (you may need to reload&mdash;CodeMirror does not just play nice with Reveal).  With the partial behaviour adopted in my library, the set-get law can be fulfilled in more cases.  There are also some read-only lenses in partial.lenses.  Most notable of those is [`zero`](https://github.com/calmm-js/partial.lenses#L-zero), which can act as the identity element when querying with lenses ([`choice`](https://github.com/calmm-js/partial.lenses/#L-choice)) and with conditional traversals ([`when`](https://github.com/calmm-js/partial.lenses/#L-when)).

Also, in partial.lenses the `find` lens combinator is implemented in terms of the [`choose`](https://github.com/calmm-js/partial.lenses#L-choose) lens combinator. `choose` is an important combinator as it supports creating lenses that can adapt to data-structures.

CrossEye:
I would love to revive this, or at least to consider it again.  But this PR is too old just to use.

PR #1997, opened 2016-12-06
Update path.js

FrancoisPala:
Hi, I've tried to do this function recursively and found out that the performances were better (unless for deep cases, like 5 or 6-deep keys in a object) using benchmark.js (https://github.com/bestiejs/benchmark.js/), is there any reason you guys did it iteratively instead?
Also I know it adds a new 'i' parameter which makes it less practical but I was just wondering... 
Thanks!

### DIFF
--- a/src/path.js
+++ b/src/path.js
@@ -19,15 +19,9 @@ var _curry2 = require('./internal/_curry2');
  *      R.path(['a', 'b'], {a: {b: 2}}); //=> 2
  *      R.path(['a', 'b'], {c: {b: 2}}); //=> undefined
  */
-module.exports = _curry2(function path(paths, obj) {
-  var val = obj;
-  var idx = 0;
-  while (idx < paths.length) {
-    if (val == null) {
-      return;
+module.exports = _curry2(function path(paths, obj, i) {
+    if (paths[i]) {
+        return path(paths, obj[paths[i]], i + 1);
     }
-    val = val[paths[idx]];
-    idx += 1;
-  }
-  return val;
+    return obj;
 });
--- a/src/path.js
+++ b/src/path.js
@@ -20,11 +20,11 @@ var _curry2 = require('./internal/_curry2');
  *      R.path(['a', 'b'], {c: {b: 2}}); //=> undefined
  */
 
-function recursiveFunc(paths, obj, i){
-    if (paths[i]) {
-        return func(paths, obj[paths[i]], i + 1);
-    }
-    return obj;
+function recursiveFunc(paths, obj, i) {
+  if (paths[i]) {
+    return recursiveFunc(paths, obj[paths[i]], i + 1);
+  }
+  return obj;
 }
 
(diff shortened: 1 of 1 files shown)


src/path.js
```diff
@@ -19,15 +19,9 @@ var _curry2 = require('./internal/_curry2');
  *      R.path(['a', 'b'], {a: {b: 2}}); //=> 2
  *      R.path(['a', 'b'], {c: {b: 2}}); //=> undefined
  */
-module.exports = _curry2(function path(paths, obj) {
-  var val = obj;
-  var idx = 0;
-  while (idx < paths.length) {
-    if (val == null) {
-      return;
+module.exports = _curry2(function path(paths, obj, i) {
```
FrancoisPala: Yeah definitely, one way to remove it would be to nest the function inside an other one calling for this one: you'd have the path(paths, obj) function calling inside it the recursiveFunc(paths, obj, 0).

CrossEye:
Ramda mostly avoids deep recursion for its functions.  It's predecessor, [Eweda][ew], was built with beautiful, elegant, recursive code, and it started falling down on the job as lists grew longer.  Ramda is an attempt to make things much more practical.

While I'd love to see your benchmarks, the notion of moving to deeply recursive implementations before tail-call optimizations are ubiquitous scares me.

Also note that your version seems to break a number of tests.  I haven't had time to dig into why.



FrancoisPala:
Hi, thanks for the input! So about the tests, some broke because the return values are 'null' instead of 'undefined' in case of empty objects or errors and they were waiting for the latter. Some other tests broke on other functions like lensPath which I suppose use path but I haven't had time to look into it yet.
Now for the benchmarks, I used the benchmark module and compared the result, also changed the path variable for different tests:
`
var obj = {
        a: "a",
        b: {
            z: "z",
            c: {
                e: "e",
                d: {
                    r: "r",
                    d: {
                        r: "r",
                        d: {
                            r: "r",
                            d: {
                                r: "r",
                                d: {
                                    r: "r",
                                }
                            }
                        }
                    }
                },
            },
        },
    };
`

    const path = ['b', 'c', 'd', 'r'];

    b.add('with ramda', function() {
        R.path(path, obj);
    })

    .add('with recursive', function() {
        path(path, obj);
    })

    .on('cycle', function(event) {
        console.log(String(event.target));
    })

    .on('complete', function() {
        console.log('Fastest is ' + this.filter('fastest').map('name'));
    })

    .run({ 'async': true });


Sorry for the bad formatting, I'm having trouble with the editor

CrossEye:
I really meant I wanted to see the *results* of your tests.  :smile:.

I didn't get that running in my 3-minute attempt.  Perhaps I'll try later.

But breaking tests, especially when it involves other functions is a very bright red light.  Do you think you could fix your function so that it doesn't break them and still has these performance gains?

FrancoisPala:
Hey, thanks for fixing the formatting :D I'll try to make it return approriate values when I get a moment

PR #2003, opened 2016-12-08
add isNotNil function to check value for not null or not undefined

anandsahil:
1. Allows checking for not null or not undefined , so that we do not need to always negate isNil .

### DIFF
--- a/index.js
+++ b/index.js
@@ -97,6 +97,7 @@ module.exports = {
   isArrayLike: require('./src/isArrayLike'),
   isEmpty: require('./src/isEmpty'),
   isNil: require('./src/isNil'),
+  isNotNil: require('./src/isNotNil'),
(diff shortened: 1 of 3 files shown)


davidchambers:
`x => x != null` seems quite reasonable to me. I'm not *strongly* opposed to adding this function, though. ;)

CrossEye:
I feel the same.  No major objection, but it's so easy to add yourself, and can be done as a simple arrow or as `complement(isNil)` that I see little need.

buzzdecafe:
i agree with @davidchambers . if you want to avoid the lambda for some reason, you could also `complement(isNil)`

anandsahil:
i agree but i feel that always , abstracting it in ramda would be good and to enjoy the functional and currying native features for free.

it would be really nice to have it

Issue #2006, opened 2016-12-11
allPass is returning a function unexpectedly


scull7:
I'm trying to use a `allPass` to determine if an integer matches a set of predicates.  However, when calling `allPass` with 2 parameters it always returns a function.  Is this the expected behavior?

https://goo.gl/Nxff5m

```javascript
R.allPass([ f1, f2, f3 ], x) // -> function
R.allPass([ f1, f2, f3 ])(x) // -> bool
```

davezuko:
Lurker jumping in: it looks like this is due to the fact that `allPass` [curries the generated function to the maximum arity among the supplied functions](https://github.com/ramda/ramda/blob/master/src/allPass.js#L34), so to handle this in a single invocation would in fact make `allPass` variadic.

The function's signature is documented as`[(*… → Boolean)] → (*… → Boolean)`, which demonstrates that this is in fact intentional. This says nothing of whether or not this is the ultimate desired behavior (though I think it is), just that it does act as documented.

scull7:
Ah, that makes sense now (probably should have in the first place); the `->` indicating the curry and the `()` indicating the grouping.  Just unexpected on my part as I was thinking that the `allPass` was a curried, fixed-arity function; similar to some of the other logic functions, but its behavior does fit with it's closer relatives. Thank you @davezuko for the explanation.

davidchambers:
[`S.allPass`][1] has the simpler type you expected, @scull7.


[1]: https://sanctuary.js.org/#allPass

scull7:
@davidchambers Thanks for the Sanctuary.js suggestion.  very helpful.

PR #2007, opened 2016-12-11
curry helper for functions that return functions

kedashoe:
Quick PoC. Addresses #2006. The idea is to be totally backwards compatible, functions that this is a fit for can be called as they always could, but can now also be invoked all in one go.

Before:

```js
allPass([gt(5), odd]); => Int -> Bool
allPass([gt(5), odd], 4); => Int -> Bool
allPass([gt(5), odd])(4); => false
```
After
```js
allPass([gt(5), odd]); => Int -> Bool
allPass([gt(5), odd], 4); => false
allPass([gt(5), odd])(4); => false
```

Some candidates: allPass anyPass bind cond constructN invoker liftN memoize nthArg once tryCatch

"curryX" could also be an internal function, I'm not sure it needs exporting?

### DIFF
--- a/src/allPass.js
+++ b/src/allPass.js
@@ -1,5 +1,5 @@
-var _curry1 = require('./internal/_curry1');
 var curryN = require('./curryN');
+var curryX = require('./curryX');
 var max = require('./max');
 var pluck = require('./pluck');
 var reduce = require('./reduce');
@@ -30,7 +30,8 @@ var reduce = require('./reduce');
  *      isQueenOfSpades({rank: 'Q', suit: '♣︎'}); //=> false
  *      isQueenOfSpades({rank: 'Q', suit: '♠︎'}); //=> true
  */
-module.exports = _curry1(function allPass(preds) {
+
+module.exports = curryX(function allPass(preds) {
   return curryN(reduce(max, 0, pluck('length', preds)), function() {
     var idx = 0;
     var len = preds.length;
--- a/src/curryX.js
+++ b/src/curryX.js
@@ -1,0 +1,33 @@
+var curryN = require('./curryN');
+
+
+/**
+ * todo
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig :'(
+ * @param {Function} A function that returns a function
+ * @return {Function} A curried function
+ * @see R.curry, R.curryN
+ * @example
+ *
+ *      todo
+ */
+module.exports = function curryX(f) {
+  return curryN(f.length, function() {
+    if (arguments.length > f.length) {
+      var fArgs = Array.prototype.slice.call(arguments, 0, f.length);
+      var gArgs = Array.prototype.slice.call(arguments, f.length);
+      var g = f.apply(null, fArgs);
+      if (gArgs.length >= g.length) {
+        return g.apply(null, gArgs);
+      } else {
+        return curryN(g.length, g).apply(null, gArgs);
+      }
+    } else {
+      return f.apply(null, arguments);
+    }
+  });
+}
--- a/src/either.js
+++ b/src/either.js
@@ -1,5 +1,7 @@
 var _curry2 = require('./internal/_curry2');
 var _isFunction = require('./internal/_isFunction');
+var curryN = require('./curryN');
+var curryX = require('./curryX');
 var lift = require('./lift');
 var or = require('./or');
 
@@ -31,10 +33,10 @@ var or = require('./or');
  *      f(101); //=> true
  *      f(8); //=> true
  */
-module.exports = _curry2(function either(f, g) {
+module.exports = curryX(function either(f, g) {
   return _isFunction(f) ?
-    function _either() {
+    curryN(f.length, function _either() {
       return f.apply(this, arguments) || g.apply(this, arguments);
-    } :
+    }) :
     lift(or)(f, g);
 });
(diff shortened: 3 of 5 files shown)


davidchambers:
:thumbsdown: from me. I believe the right approach is to fix the arity of every function—including those provided as arguments—rather than concoct ever more intricate :curry: functions.

I like this type:

```purescript
allPass :: Array (a -> Boolean) -> a -> Boolean
```

If there's demand, we could also provide `allPass2` and `allPass3`:

```purescript
allPass2 :: Array (a -> b -> Boolean) -> a -> b -> Boolean
allPass3 :: Array (a -> b -> c -> Boolean) -> a -> b -> c -> Boolean
```

We could also support uncurried predicates:

```purescript
allPass2_ :: Array ((a, b) -> Boolean) -> a -> b -> Boolean
allPass3_ :: Array ((a, b, c) -> Boolean) -> a -> b -> c -> Boolean
```

It's unlikely I'll get my way on this, but there's always Sanctuary. :stuck_out_tongue_winking_eye:

kedashoe:
> rather than concoct ever more intricate :curry: functions

I dunno, [I think more intricate currying can be ok if it gives the user a better experience](https://github.com/sanctuary-js/sanctuary-def/blob/master/index.js#L1019-L1176).

I wonder about `allPass2`, `allPass3`, etc.. this has always felt like a limitation in Hindley-Milner/Haskell. It cannot have felt good, for example, to write `unzip7` and then stop. And if we had a generic `unzip` such that the compiler could type check everything, and then someone came along and said "no no, this is all wrong, we need `unzip1-7`", would we then get rid of our generic version and use the numbered ones? I have not written enough Haskell to really know, but it feels awkward.

In any event, to the question at hand.. do you feel like this change would give the user a *worse* experience with ramda?

davidchambers:
I agree that numbered functions are not ideal, and indicate a shortcoming of Haskell.

In JavaScript, though, we have no compiler at all, so the burden of ensuring correctness lies entirely on the shoulders of the programmer. This being the case I consider it important to:

  - use simple constructs; and
  - favour constructs which we as people find easy to reason about.

Simple constructs are important because at some point someone will end up reading the source code to understand why `undefined is not a function`. Also, simple constructs are less likely to produce edge cases.

The tools at our disposal are reasoning and testing. Testing can tell us that something is wrong, but it can't tell us how to fix the problem. We must reason about our bugs. I'm fairly comfortable reasoning about applications of this function in my programs:

```purescript
lift2 :: Apply f => (a -> b -> c) -> f a -> f b -> f c
```

With this function, on the other hand, I'm forced to read the source code:

```purescript
lift :: (*... -> *) -> ([*]... -> [*])
```

> In any event, to the question at hand.. do you feel like this change would give the user a *worse* experience with ramda?

I like to opt out of much of JavaScript's dynamic features, in order to give myself a fighting chance of understanding my programs with my limited reasoning ability. Surely many people do not share this preference, though. I've no doubt that a significant proportion of Ramda users can reason about complex programs more effectively than I can. Also, I imagine that some Ramda users don't place much value on reasoning, preferring instead to arrive at working programs via trial and error.

scull7:
Once it was explained to me, I think that the current functionality makes the most sense.  I am of the same opinion ast @davidchambers.  I think that attempting to solve this problem within Ramda will only make it less approachable.  The only think that I think needs work is the documentation.  Need to mark functions which are not curried in a much clearer way.  The curried nature of the Ramda library as a whole makes a non-curried function behavior surprising.

CrossEye:
I like the idea.  I'll have to look more carefully at the implementation.  And I won't be able to tonight or tomorrow.

While I like David's suggestion that one should use simple constructs, I also think that Ramda should cater to the sort of code that Javascript developers use.  We're trying to make FP concepts more accessible to FP developers, but that doesn't mean that we're abandoning the norms of the language.

I really appreciate the consistency this could add to the library.

kedashoe:
I've updated the PR message, which was lazy and bad :disappointed:

I like simple constructs too! My whole idea here was to make things simpler and more consistent for the user.

Maybe this function would just be internal?

rjhilgefort:
I just hit this snag when trying to use `allPass` with `useWith`. Admittedly, the signature is clear and I should have read it more closely. I think for consistency's sake, this PR should be pushed along. I see that some of the closely related functions like `cond` and `ifElse` operate operate simlarly, but the line gets a little fuzzy for me when we consider things like `propSatisfies`, `pathSatisfies`.

My use case is hard to work around this limitation (if we stick with trying to use `useWith`.... this is a contrived example):

```javascript
const containsAll = useWith(allPass, [map(contains), identity])

containsAll(['A', 'B', 'C'], '_A_B__C__')
// -> function r(e){return 0===arguments.length||n(e)?r:t.apply(this,arguments)}
```

My workaround for the time being is to just make a new version of `allPass` and overwrite Ramda's:


```javascript
const allPass = curry((preds, data) => allPass(preds)(data));

const containsAll = useWith(allPass, [map(contains), identity])

containsAll(['A', 'B', 'C'], '_A_B__C__') // -> true
```

It seems this PR has lapsed and if I should create a new issue or something to re-spark the discussion, please let me know.

I like @davidchambers's proposed interfaces and would be happy to take a stab at taking this PR over in that spirit.

GingerPlusPlus:
`curryX` sounds like an idea similar to `uncurry`, maybe `uncurry` could be generalized to handle this case?

Issue #1998, opened 2016-12-06
Propose overPath to apply a function to a nested property.


MrRacoon:
Ramda has both `path` and `assocPath` for using the `[key]` strategy to get/set data respectively.

Curious if you guys would find `overPath` handy, where the signature could look something like:

```haskell
overPath :: [String] -> (a -> b) -> {k:v} -> {k:v}
```

and the implementation would be analogous to:

```javascript
const overPath = (pl, fn, obj) => {
    const was = path(pl, obj);
    return assocPath(pl, fn(was), obj);
}
```

The motivation being that, where lenses have `view`, `set`, and `over`. Path could be a lighter abstraction with `path`, `assocPath`, and `overPath` respectively.

If that sounds like something you guys dig, I wouldn't mind making a PR for it sometime this week.

CrossEye:
While not adamantly opposed, I'm more inclined in the opposite direction.  @davidchambers has made the case for removing some of the functions that overlap with lenses in favor of promoting a consistent lens-based interface.  While I might not be so happy with losing ones like `prop` or `path`, ISTM that `assoc` and its cronies could well go.

Part of the point is that while there is a slight advantage to the conceptual weight of `overPath`, the usage is not *much* simpler:

```js
overPath(['a', 'b', 'c'], myFunc, myObj);
```
is only slightly simpler than
```js
over(lensPath(['a', 'b', 'c']), myFunc, myObj);
```

And the latter comes along with all sorts of useful abstractions not available to the former.

MrRacoon:
I happen to totally agree with stripping out the redundancy. 

I would add though, IME "spreading the rhetoric" about how cool FP and Ramda are, I've noticed that the lens abstraction tends to shake new players. I find I've had to work up to it in most cases, especially in listeners who don't yet have that "FP itch".

Kind of what I planned to do with this change was to bait and switch listeners. Show them the various path functions and then put a lens in front of them at the last minute in an attempt to make the introduction less ground shattering.

---

Though, seeing how similar your example functions are, I'm starting to wonder if there would be benefit in merging the two styles?

```javascript
over(['a', 'b', 'c'], myFunc, myObj);
// could be equivalent to 
over(lensPath(['a', 'b', 'c']), myFunc, myObj);
```

If you're familiar with [partial.lenses](https://www.npmjs.com/package/partial.lenses), this is a similar approach they take. Using a `String` as a valid `lensProp`. Which is nice, I enjoy how minimal their lens compositions are.

[...]

PR #2010, opened 2016-12-13
Add `assocWith` and `assocPathWith` functions

MrRacoon:
References #1998 

Similar to `assoc` and `assocPath`. These functions are used to apply a
function to a value at a given property/path. If there is no previous
value, then the function will be given `undefined` as its argument.


MrRacoon:
Need to revise the type signatures

PR #2011, opened 2016-12-14
Add `assocWith` and `assocPathWith` functions

MrRacoon:
Similar to `assoc` and `assocPath`. These functions are used to apply a
function to a value at a given property/path. If there is no previous
value, then the function will be given `undefined` as its argument.

references #1998


CrossEye:
Sorry, missed this first time around.  Thanks for the ping.

I'm still where I was in the discussions in #1998:  I have no major objection, but I don't feel much of a need.  Are there core contributors who really like this idea?  I know that @kedashoe has already spoken up in favor.  Any others?  @asaf-romano? @buzzdecafe? @davidchambers? @paldepind? @raine? @scott-christopher? @TheLudd? @Bradcomp?  Anyone else?

scott-christopher:
I'm personally happy with the use of lenses for this functionality, however I can see the appeal and it would provide some parity with `R.update` and `R.adjust`, so I have no objections to this landing.

It looks like the implementation of `assocPathWith` will need to be updated to support numeric indexes, as per the recent changes to `path` and `assocPath`.

Bradcomp:
> I have no major objection, but I don't feel much of a need. 

Same. 

I actually perform operations like these fairly regularly, but I haven't found it to be any trouble to use existing functions.  

If I need to modify a nested structure I'll use `over(lensPath(p), f, obj)` and for the `prop` case I just use `evolve`.

kedashoe:
Ya, pretty much agree with the sentiments so far. I just felt like these are so common they deserve a place in the lib. More general in a sense as well, `assoc` and `assocPath` can use the `*With` version as `update` uses `adjust`.

> It looks like the implementation of assocPathWith will need to be updated to support numeric indexes, as per the recent changes to path and assocPath.

@MrRacoon , can you copy of the implementations of `assoc` and `assocPath` to their respective `*With` functions and then have those use `always` similar to `update` and `adjust`?

MrRacoon:
@kedashoe I'm sure I can.

paldepind:
```js
R.assocWith(inc, 'b', {a: 1, b: 2});
```

is identical to

```js
R.evolve({b: inc}, {a: 1, b: 2});
```

The only advantage of the former is that it can be curried on the first argument. To me that does reduce the appeal of adding a new function somewhat.

Both `R.assocWith` and `R.assocPathWith` are implemented so that they handle arrays as well as objects? This makes them strictly more powerful than `R.adjust` and `R.update`. If that is the case then why not just replace `R.adjust` and `R.update` with these two new functions? I think we should either do that or remove the (currently undocumented) array handling.

`R.assocPathWith` is implemented so that `R.assocPathWith(fn, [], o)` is equal to `fn(o)`. I like the behavior but I do think it should be documented.

MrRacoon:
@paldepind Okie dokie, So action Items I got out of this were:

* Document edge cases more clearly
* Replace the implementations of `adjust` and `update` to use `assoc*With`

Is that correct? You didn't necessarily mean to remove `adjust` and `update` from the library, no?

paldepind:
@MrRacoon 

Yes. I like the way you handle the edge case. But I think it needs documentation so that people who wonder if it handles or what it does with an empty array can see it in the documentation.

I actually meant implement `R.assocWith` with the name `R.adjust` and remove the old `R.adjust` implementation.

Currently `R.assocWith` can take both a number and a string as it's second argument and an object or array as its third argument. That is, it can do everything that `R.adjust` can do. I don't think it would be a good idea to have one function whose functionality is a superset of another one. So either the functionality should be removed from `R.assocWith` or `R.adjust` should be able to handle objects (i.e. implemented like `assocWith`).

I'm leaning towards the later but I think that it is somewhat controversial as `assocWith` is currently overloaded. It's signature is more like this:

```
(v -> w) -> (string | number) -> ({k: v} | [v]) -> ({k: v} | [v]})
```

And overloading is something that Ramda often goes against. In this case however, I do think it is justified.

MrRacoon:
@paldepind sorry for the radio silence.

The only thing that scares me about this is that it doesn't totally make sense to index into a list with a string, or a map with a number (in most cases, I'm aware that the latter is legal).

It strikes me as slightly misleading. But if you think that's fine, I can move forward.

MrRacoon:
Tweaked the docs a little bit, I was hoping to hear maybe one more person chime in about cutting out `adjust` before I rolled through and did so.

JoshWhite:
Hi, I was gonna make a pull request for assocWith, but found this thread. 

Is there a reason this hasn't been pulled in yet?

CrossEye:
> Is there a reason this hasn't been pulled in yet?

Probably not.  I'll take a look.

ashnur:
This would help people whose job is to write paths all day. Probably not every javascript developer needs it, probably not all those who use Ramda will need it, but for some of us, it would be very useful. 

IMHO lenses are just a bit too verbose for this situation.

MrRacoon:
No time

PR #2013, opened 2016-12-20
TypeError: Cannot read property 'length' of undefined

waitingcheung:
I am trying to split a string into groups and I found the ``groupWith`` function helpful. However, my string is not always available. I am getting a TypeError in such case.

```
TypeError: Cannot read property 'length' of undefined
    at /path/ramda.js:1910:23
    at /path/ramda.js:466:28
    at f1 (/ramda.js:446:27)
```

Here is a simple test case to reproduce that. The function ``readStr`` first reads a string and sends it to ``groupWith``. If the string is not available, it returns ``undefined``. If ``undefined`` is passed to ``groupWith``, I am getting the above error.

```js
var str = readStr(params); // return undefined if the string is not available
console.log(R.groupWith(R.equals)(str));
```

I think if there is a default value of zero for ``len`` on [line 1910](https://github.com/ramda/ramda/blob/master/dist/ramda.js#L1910) then I can still get the empty array from ``res`` on [line 1908](https://github.com/ramda/ramda/blob/master/dist/ramda.js#L1908) without crashing.

### DIFF
--- a/src/groupWith.js
+++ b/src/groupWith.js
@@ -29,7 +29,7 @@ var _curry2 = require('./internal/_curry2');
 module.exports = _curry2(function(fn, list) {
   var res = [];
   var idx = 0;
-  var len = list.length;
+  var len = Array.isArray(list) || typeof list === 'string' ? list.length : 0;
   while (idx < len) {
     var nextidx = idx + 1;
     while (nextidx < len && fn(list[idx], list[nextidx])) {
--- a/test/groupWith.js
+++ b/test/groupWith.js
@@ -25,4 +25,7 @@ describe('groupWith', function() {
     eq(R.groupWith(R.equals)('Mississippi'), ['M','i','ss','i','ss','i','pp','i']);
   });
 
+  it('returns an empty array for non-supported values', function() {
+    eq(R.groupWith(R.equals, undefined), []);
+  });
 });


waitingcheung:
@CrossEye I notice from the API documentation that ``groupWith`` supports only arrays and strings. In my pull request, I check if ``list`` is either type and set it to zero if it is not. Please let me know how it looks to you and if you need further changes. Thank you.

davidchambers:
I'm opposed to this change. If a function expects a value of a certain type it should not accept values of *other* types. `Any -> Any` functions are difficult to reason about.

buzzdecafe:
I agree with @davidchambers ; i'd rather see it blow up than try to be "helpful" by providing a default.

waitingcheung:
I think the implementation should be consistent throughout the whole library. For example in [line 2426](https://github.com/ramda/ramda/blob/master/dist/ramda.js#L2426) of the ``length`` function, instead of simply returning ``list.length`` and let it blow up, it checks whether the list is ``null`` and whether its length is a number. The same pattern should also apply here.

CrossEye:
My answer is slightly different.  I don't care what type the input is.  It's something with a `length` property that responds to indexing; our implementation also requires it to respond to `slice`, but that's a detail.  So I don't want to test for `Array`/`String`.

But, I would have no problem with something that extended this as the original idea suggested, so that `undefined` simply returned an empty list.  I know this is something on which @davidchambers  and I have fundamental disagreements, and I think @buzzdecafe and I are also at least somewhat at odds.  To me, throwing exceptions should be reserved for the most dire of circumstances, when there is simply nothing left to do.  Here there is a reasonable option.  I'd prefer to go for that.

PR #2016, opened 2016-12-20
Fixed groupWith to pass adjacent elements to testing fn

### LINKED ISSUE #2012: Add groupByStreak - group values in a list by testing consecutive/adjasent values if they keep the steak going
I couldn't find anything like this in the docs, and implementation using simple re-combination of existing functions was not apparent either. 
As a side note, `R.groupWith` is *almost* that, except it does not pass adjacent items in the list, but rather tests against first item in the group against each next item. "fixing" `groupWith` changes semantic, so I propose a new function: `groupByStreak`

Especially useful to group records with dates into ranges.

Here is my implementation: 
```
const streak = (sameStreak) => {
  let s = 0
  let valueSet = false
  let lastVal = null
  return (v) => {
    if(!valueSet) {
      valueSet = true
      lastVal = v
      s = 0
    } else if(!sameStreak(lastVal,v)) {
      lastVal = v
      s += 1
    } else {
      lastVal = v
    }
    return s
  }
}
const groupByStreak = fn => R.compose(
  R.map(R.map(R.nth(1))),
  R.values,
  R.groupBy(R.nth(0)),
  R.converge(R.zip,[R.map(streak(fn)), R.identity])
)

const isConsecutive = (a, b) => a + 1 == b
groupByStreak(isConsecutive)([0,1,2,3,1,2,0,4,5,6,9]) // [[0, 1, 2, 3], [1, 2], [0], [4, 5, 6], [9]]
```

Would be happy to submit PR if the owners approve.

kmamykin:
Very minimal fix to the src code. Added docs and tests.

Fixes #2012

### DIFF
--- a/src/groupWith.js
+++ b/src/groupWith.js
@@ -2,7 +2,8 @@ var _curry2 = require('./internal/_curry2');
 
 /**
  * Takes a list and returns a list of lists where each sublist's elements are
- * all "equal" according to the provided equality function.
+ * all satisfied pairwise comparison according to the provided function.
+ * Only adjacent elements are passed to the comparison function.
  *
  * @func
  * @memberOf R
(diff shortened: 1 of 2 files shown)


CrossEye:
:herb:

kedashoe:
:mans_shoe:

CrossEye:
Any other feedback?

kedashoe:
Anyone else? Are we good to merge this?

Issue #1968, opened 2016-11-10
Migrate to ES6 modules


lpan:
It will be great if we can migrate to ES6 modules like [three.js did](https://github.com/mrdoob/three.js/pull/9310) so the end user can take advantage of Webpack2/Rollup's `tree shaking` feature.

buzzdecafe:
whom do you mean by "we"?

lpan:
@buzzdecafe I am just suggesting here xD feel free to close this issue if you don't like it

CrossEye:
The problem for now is that Ramda is also available as individual modules.  You can simply do 

```js
const map = require('ramda/src/map');
```

, soon to be replaced with just:

```js
const map = require('ramda/map');
```

I don't see any way to maintain that -- which is important to a number of people, still, and switch to a Rollup-built ES6 module format.  If this actually is possible, I'd really like to know!

jonaskello:
Out of curiosity, what is the benefit of doing 

```js
const map = require('ramda/map');
```

vs 

```js
const map = require('ramda').map
```

I can only think of reducing bundle size but I don't think that would be relevant when using it on nodejs like this?

jonaskello:
Regarding having the original code as ES6 modules and still being able to have separate commonjs modules I think there are several ways to do that. One way would be to use [this function](https://babeljs.io/docs/plugins/transform-es2015-modules-commonjs/) in babel. The idea is to write the modules in ES6 and have babel transform them to commonjs. If the transform is applied on a per file basis I think it should be possible to require the resulting commonjs files separately like today.

CrossEye:
I would not at all mind seeing a demonstration of these ideas if you're interested in creating a PR.

I don't know what the benefits are, except that I know a number of people do seem to want this.

jonaskello:
I did a small POC in #1977. Only `src/add.js`has been updated to use es6 style export. If the concept is accepted I could rewrite all to this style.

faergeek:
@CrossEye About the benefits of proper build for rollup/webpack 2. You can continue to import `ramda` like this:
```
import {map} from 'ramda'
```
And with rollup/webpack 2 it will work like this (roughly), because if tree-shaking:
```
import map from 'ramda/src/map'
```
So you actually bundle only the things that you import (and use).

jonaskello:
Here follows some speculation on the need to have commonjs modules that are importable as separate files. If you are doing this in your application:

```js
const map = require('ramda/map');
```

[...]

PR #2026, opened 2016-12-27
search script added

Rafi993:
search script has been added that will search all .js files for finding list of ramda functions used and make them into custom build using partial build script.
**example**
`node search -f C:\Users\human\Desktop\projectDir > custombuild.js`


MattMS:
This is a nice idea, but will it actually work the way you have specified?
Wouldn't the `search` file need to be in the folder you are running `node search` from?

If this is to be continued, I would suggest using the [commander][1] package for command-line parsing, as it is used in [other scripts][2].

[1]: https://www.npmjs.com/package/commander
[2]: https://github.com/ramda/ramda/blob/master/scripts/build#L257

Rafi993:
Yes you need to be in the scripts folder. when absolute path of the project folder is given to it as argument it does work.

MattMS:
You may want to change this to a [npm bin script][1] by adding it to `package.json`.

I would also suggest renaming the script to something like `ramda-usage`.

Having said all that, I'm wondering if this would be better suited as a separate package, given that it is a fairly specific use case.

Rafi993:
@MattMS may be you are right. May be the partial builds ,search and similarl things should be moved to a separate repo as a CLI tool similar to lodash cli

Rafi993:
That will keep the main repo clean and these additional tools independent of main repo.

Rafi993:
That said since complicated build scripts are not here. Till then a separate CLI tool will be an overkill.

davidchambers:
I wonder whether we'd be better served by addressing #1968.

Rafi993:
@davidchambers  rather than porting to just es6 wouldn't it be better if it was typescript

MattMS:
I agree with @davidchambers on this, which is why I suggested it may be suited to a separate package.

Rafi993:
Closing this since it is not relevant anymore.

PR #2027, opened 2016-12-28
Explain filterable

jethrolarson:


src/filter.js
```diff
@@ -22,8 +22,8 @@ var keys = require('./keys');
  * @category List
  * @sig Filterable f => (a -> Boolean) -> f a -> f a
  * @param {Function} pred
- * @param {Array} filterable
- * @return {Array}
+ * @param {Array} Filterable
+ * @return {Array} Filterable
```
davidchambers: Isn't the thing after the `{…}` meant to be the parameter name rather than the type?

jethrolarson:
Oh yeah. I guess it should be `xs` or something

kedashoe:
Thanks for the PR @jethrolarson, this is definitely needed :)

What I would love to see here long-ish term is for ramda to use santuary-type-classes, at which point we could point to those docs (hopefully we'll have a table over there of which types implement which classes.. there has been discussion and I think everyone is in favor but there were some issues with it.. I can't find the thread at the moment).

In the meantime, some thoughts on this PR:

- we also filter objects, which do not have a `filter` method
- `partition` and `reject` also mention "Filterable"

Can you think of somewhere we can put a definition that these can link to? Or should we just repeat it in each place?

CrossEye:
> Can you think of somewhere we can put a definition that these can link to? Or should we just repeat it in each place?

I'd really like to have a common place in the docs for these to point to.  And I keep not getting to it.

jethrolarson:
I'd like to see some sections for the type classes that ramda supports. They could just be instances in the list of functions. However they're not functions so I don't know how the documentation generation would work with it. Beyond that they could be put at the top or bottom of the docs.

I'd also like to see the docs search work for more tags than the current "List", "Object", etc.
e.g. `R.map` is under "List" but should also apply to `Object`.

kedashoe:
Updates look good to me.

kedashoe:
> While we're at it we might as well include search by signature. :)

I've done some work on this. I recently updated [the parser](https://github.com/kedashoe/hindley-milner-parser-js), I reckon it's in decent shape now. Issues/pull requests welcome :) I have no time over these last few days of the holidays, but I'll be raising some issues myself once I have some free time.

CrossEye:
@jethrolarson: I agree, we should list the Typeclasses used, `Ord`, `Filterable`, `Lens`, etc., with useful descriptions, and the function descriptions that use them should link to these descriptions.  Also, I've always meant for functions to be able to be listed under multiple categories.  We should -- at a minimum -- be able to support that in searches.  But I would love it if the search were much stronger, looking through the entire description, and, as suggested, the type signatures.

@kedashoe: That's great!  I hope I get a chance to play with this over the next few days.  Do you see a chance of this leading to something like our own mini Hoogle-search?  Or is that too far away still?  I love being able to do this: https://www.haskell.org/hoogle/?hoogle=%28a+-%3E+b%29+-%3E+[a]+-%3E+[b]

kedashoe:
> Do you see a chance of this leading to something like our own mini Hoogle-search?

Yep! I put up a POC for sanctuary a while ago. Mostly all I've done since then is refactor the parser. Still a bit of work to do on that side of things. I plan to create a repo to search for these things, so you can just pass it a list of function signatures to build the database and then a string to search them.

jethrolarson:
In the PR I considered switching the Type to Filerable from Array but I
worried that'd mess with tooling like typescript

CrossEye:
(Thinking of how we might structure a one-page document with all the relevant type-classes, so that the API docs could simply link to it.  This is simply a working sample.  If people like the idea, I'll move it to the wiki for polishing and expansion.  We would still have to figure out how to load it to the documentation site, I think.  I really believe all the docs should live there long-term.)

---------

Functor
======

Describes an object that can be mapped over: a function is applied to the contents and a new structure of the same sort is returned holding the results.

Behavior
---------

  - `map`: A functor must have a `map` method.  Ramda supplies this for the following types:
    - Lists: returns a new list of the same length whose elements are the results of applying the supplied function to each element of the original list
    - Objects: returns a new object whose values for each key are the results of applying the supplied function to the value at that key in the original object.
    - Function: returns the composition of the supplied function with this one.

    All other objects which want to take advantage of this behavior must supply a `map` method.


Laws
----

A functor, `fnctr`, is subject to the following two laws (which means that implementors of `map` methods must ensure that their implementations follow them):

**Identity**
```js
map(x => x, fnctr) ≍ fnctr
```

**Composabiliy**
```js
map(f, map(g, fnctr)) ≍ map(x => f(g(x)), fnctr)
```

Examples:
---------

**List of Numbers** 
```js
map(n => n * n, [1, 2, 3, 4]); //=> [1, 4, 9, 16]
```

**List of Strings**
```js
map(toUpper, ['foo', 'bar', 'baz']); //=> ['FOO', 'BAR', 'BAZ']
```

Note that a functor will always turn a list into a list, the contents *inside* the list do not have to stay the same type:

**List of Strings -> List of Numbers**
```js
map(length, ['The', 'quick', 'brown', 'fox', 'jumped']); //=> [3, 5, 5, 3, 6]
```

But it's not just lists.

**Object containing Number**
```js
map(n => n * n, {a: 2, b: 3, c: 5}); //=> {a: 4, b: 9, c: 25}
```

For functions, it's equivalent to functional composition:

**Function**
```js
const f = map(length, vowelsOnly);
// f('testing') ≍ length(vowelsOnly('testing')) ≍length('ei') ≍ 2
```

But this works for a fairly standard `Maybe` type as well:

**Maybes**
```js
map(inc, Just(6)); //=> Just(7);
map(inc, Nothing()); //=> Nothing)()
```

Specifications
--------------

  - [FantasyLand][fl]
  - [StaticLand][sl]


Further Reading
---------------

  - [Functional Programming Jargon][ja]
  - TODO: more here?

  [fl]: https://github.com/fantasyland/fantasy-land#functor
  [ja]: https://github.com/hemanth/functional-programming-jargon#functor
  [sl]: https://github.com/rpominov/static-land/blob/master/docs/spec.md#functor

davidchambers:
I'd love to see convergence between Ramda and [sanctuary-type-classes][1]. The approach you've outlined above seems very nice, @CrossEye, and it would be wonderful for this to live in one place rather than having Folktale, Ramda, Sanctuary, etc. each explain these concepts and duplicate the implementations for built-in types.

On a related note, "any object with a `filter` method" is not the definition I favour. The types of [`Z.filter`][2] and [`Z.filterM`][3] do not mention "Filterable":

```purescript
filter :: (Applicative f, Foldable f, Monoid (f a)) => (a -> Boolean, f a) -> f a
filterM :: (Monad m, Monoid (m a)) => (a -> Boolean, m a) -> m a
```


[2]: https://github.com/sanctuary-js/sanctuary-type-classes#filter
[3]: https://github.com/sanctuary-js/sanctuary-type-classes#filterM

jethrolarson:
Yeah. We can just close this and open a real issue for what we want

jethrolarson:
I only found out today that Filterable is a proper thing from
magicreadalong.com

kedashoe:
> Yeah. We can just close this and open a real issue for what we want

I'd rather copy paste what you have to `partition` and `reject` and merge it. Who knows when we'll have a better place to put it.

jethrolarson:
I've added the descriptions for reject and partition in case you want to merge this in the interim while the better type definition page comes together.

CrossEye:
I'm in favor of merging this for now, and coming back to consider the outstanding questions.  @davidchambers, you had some objections to mentioning the dispatching to a `filter` method.  But that's how it works at the moment.  Do you object to making this documentation change now and considering a more disciplined approach later?

davidchambers:
> Do you object to making this documentation change now and considering a more disciplined approach later?

Not at all. Documenting the current behaviour is a good first step. :)

PR #2036, opened 2017-01-02
Adds integer path support to dissocPath

scott-christopher:
This change addresses that `R.dissocPath` was missed when support for integer path segments was added to the other path functions.


CrossEye:
:herb:

kedashoe:
:+1:

PR #2039, opened 2017-01-03
Replaces function arg with clone to avoid mutation

dmitriz:
The argument `acc` passed to the function `reduceRight` was mutated by the function.
This change avoids mutation and makes the function pure.

### DIFF
--- a/src/reduceRight.js
+++ b/src/reduceRight.js
@@ -44,10 +44,11 @@ var _curry3 = require('./internal/_curry3');
  * @symb R.reduceRight(f, a, [b, c, d]) = f(b, f(c, f(d, a)))
  */
 module.exports = _curry3(function reduceRight(fn, acc, list) {
+	var accClone = acc;
   var idx = list.length - 1;
   while (idx >= 0) {
-    acc = fn(list[idx], acc);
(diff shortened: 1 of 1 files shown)


fix-fix:
Assignment operator doesn't mutate the variable on the left hand side, so your propsed code is exactly the same.

dmitriz:
Ok, good point, but I still think it is less safe.
E.g. it is easier to introduce mutation by switching to properties:

```js
app.prop = ...
```

CrossEye:
The philosophy of Ramda is pretty simple here: we do not mutate user data.

But we do not prevent *you* from doing so.  So if you want to mutate the accumulator in a fold, that's your lookout.  While there's an argument to be made for making it impossible, we do try to work with the grain of the language.  If a user really wants to update in place some value using `reduce`, I see no reason that Ramda should prevent that.  Of course she will have to do it herself in her iteration function, since Ramda won't do so.  But Ramda will cheerfully ignore it too.

I don't see any real reason to change this.

----------

On a side note, please see our [contribution guide][cg].  This request didn't pass our linter and had no tests for the expected new behavior.


  [cg]: https://github.com/ramda/ramda/blob/master/CONTRIBUTING.md

PR #2041, opened 2017-01-05
Fixing minor regression in assocPath - fixes #2040

### LINKED ISSUE #2040: assocPath seems to be broken in 0.23.0
This is my code:
```
const scope = {
  baseValues: undefined,
}

assocPath(
  ["baseValues", "income", "categoryGroups"],
  { '9': { categories: { '32': 4 } } },
  scope,
)
```

Which works in 0.22.1 but is broken in 0.23.0. The error is: "Cannot convert undefined or null to object". There is nothing mentioned in the upgrade guide, so I assume this is not expected behaviour.

CrossEye:
Minor fix for issue raised in #2040


CrossEye:
@scott-christopher: comments?

PR #2043, opened 2017-01-06
Simplified `reduced` example

jethrolarson:


kedashoe:
`it` adds some noise for me.. just a convention I'm not aware of? Is `it` short for "item" or is it the pronoun? Maybe just use `x`?

I'm a fan of simple examples though, :+1:

CrossEye:
Same here.  `item` would be helpful.  Otherwise, :+1:

fix-fix:
@CrossEye the example now contains typo inside condition:
`(acc, item) => it > 3 ? R.reduced(acc) : acc.concat(item)`

It refers to **`it`** variable, which is not defined, instead of `item` argument.

jethrolarson:
Oops.

PR #2045, opened 2017-01-09
Correct sortwith sortby docs

burt202:
Just noticed the example result for both `sortWith` and `sortBy` is incorrect. This PR fixes that

### DIFF
--- a/src/sortBy.js
+++ b/src/sortBy.js
@@ -31,7 +31,7 @@ var _curry2 = require('./internal/_curry2');
  *        age: 314.159
  *      };
  *      var people = [clara, bob, alice];
- *      sortByNameCaseInsensitive(people); //=> [alice, bob, clara]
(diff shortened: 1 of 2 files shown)


CrossEye:
I'm mixed about this.  While I have no major objection, the original tried to point out that it returned only a new list, not new copies of the elements of that list.  Ramda is almost entirely concerned with value equality, but this is one of those places where people might conceivably think that the result starts from a deeper copy of the data than it truly is, and think they might be able to alter their new copies without harm.

The trouble is that

```js
sortByNameCaseInsensitive(people)[0] === alice; //=> true
```

So it was documented to show that it returned the original data, just ordered differently.  But that certainly could be confusing, especially when testing in the REPL.  Should this simply be noted in text or with a more explicit assertion like the line above?

burt202:
Fair point, well made. Happy to close

CrossEye:
All right.  But if you think it's confusing as it stands, perhaps that there should be such an assertion in the example or a note in the comments, please say so.

Issue #1125, opened 2015-05-28
Human readable type errors


raine:
Late last night I was coding and got a few TypeErrors here and there, for example from accidentally applying `toUpper` to a list.

```
node_modules/ramda/dist/ramda.js:5220
            return target[method].apply(target, args);
                                 ^
TypeError: Cannot read property 'apply' of undefined
```

So I wondered aloud on gitter room if it was in any way feasible for Ramda to provide a bit of hand holding in this respect and show a more friendly error message in case some thing like that happens.

Thoughts?

davidchambers:
I am 100% in favour of Ramda functions checking the types of their arguments. Particularly when using `R.compose` or `R.pipe` it can be difficult to track down the location of a type mismatch.

TGOlson:
Maybe contracts? That way even higher-order functions (like `compose`) can be contracted. [Rho-contracts](https://github.com/sefaira/rho-contracts.js) seems pretty legit, if not a little verbose, but hey, it's JavaScript.

buzzdecafe:
this may be a good use case for metadata on the function. see @megawac 's initial work in https://github.com/ramda/ramda/pull/907

buzzdecafe:
and this: https://github.com/sebinsua/ramda-debug

paldepind:
Ok. What about an internal (maybe even exposed) function that does type checking.

Can be used like this:

```javascript
var nth = _curry2(typecheck([isNumber], _nth));
var add = _curry2(typecheck([isNumber, isNumber],_add));
var filter = _curry2(typecheck([isFunction, isArray], _dispatchable('filter', _xfilter, _filter)));
```

It takes an array of predicates and a function. It then decorates the function and validates passed arguments against the predicates. Validator functions should have a `name` property. Then really nice error messages could be thrown:

```javascript
filter(12, [1,2,3]); // Throws: TypeError: filter expected a function as it's first argument, received `12`
```

This would allow to very easily and concisely add type checking to functions. It would also keep the main functions free from `throw` statements which in most current JS engines harms the performance of the entire containing function.

buzzdecafe:
i don't like the idea of folding this into the main build, but only for a debug build.

CrossEye:
I'm not convinced that we want to do this.  This sort of thing scares me:

> ```js
> var filter = _curry2(typecheck([isFunction, isArray],
>                                _dispatchable('filter', _xfilter, _filter)));
> ```

[...]

PR #2047, opened 2017-01-11
Added more helpful error when _reduce is passed null or undefined

DavidBrear:
`R.reduce(R.subtract, 0, null)` throws `Cannot read property 'reduce' of null` which is misleading. This commit checks for null or undefined and throws a more helpful message.

### DIFF
--- a/src/internal/_reduce.js
+++ b/src/internal/_reduce.js
@@ -37,6 +37,9 @@ module.exports = (function() {
 
   var symIterator = (typeof Symbol !== 'undefined') ? Symbol.iterator : '@@iterator';
   return function _reduce(fn, acc, list) {
+    if (list === null || typeof list === 'undefined') {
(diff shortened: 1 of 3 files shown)


davidchambers:
See #1125. If you're interested in helpful error messages for type errors, [Sanctuary][1] may be a better option than Ramda. Here's what you would see with the upcoming v0.12.0 release:

```javascript
S.reduce_(S.sub, 0, null);
// ! TypeError: Type-class constraint violation
//
//   reduce_ :: Foldable f => ((a, b) -> a) -> a -> f b -> a
//              ^^^^^^^^^^                          ^^^
//                                                   1
//
//   1)  null :: Null
//
//   ‘reduce_’ requires ‘f’ to satisfy the Foldable type-class constraint; the value at position 1 does not.
```

PR #2052, opened 2017-01-17
add "," before otherwise

adispring:


kedashoe:
How would you feel about a semicolon instead?

CrossEye:
:herb:

PR #2053, opened 2017-01-17
optimize R.call comment

adispring:


src/call.js
```diff
@@ -4,8 +4,8 @@ var curry = require('./curry');
 /**
  * Returns the result of calling its first argument with the remaining
  * arguments. This is occasionally useful as a converging function for
- * `R.converge`: the left branch can produce a function while the right branch
- * produces a value to be passed to that function as an argument.
+ * `R.converge`: the first branch can produce a function while the rest
+ * branches produce values to be passed to that function as its arguments.
```
kedashoe: > while the rest branches

how about:

while the remaining branches...

or

while the rest of the branches...

adispring: I have changed "rest" to "remaining"

PR #2054, opened 2017-01-17
Fix indexOf and lastIndexOf on undefined objects

waitingcheung:
The functions ``indexOf`` and ``lastIndexOf`` can work on objects with the prototype ``indexOf`` and ``lastIndexOf``. If the objects are ``undefined`` the following hard crash can occur. I have added the corresponding test cases and fixes.

```
        if (typeof xs.lastIndexOf === 'function' && !_isArray(xs)) {
                     ^

TypeError: Cannot read property 'lastIndexOf' of undefined
    at lastIndexOf (/path/ramda.js:6069:22)
    at Object.f2 [as lastIndexOf] (/path/ramda.js:442:22)
```

### DIFF
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -1,6 +1,6 @@
 //  Ramda v0.23.0
 //  https://github.com/ramda/ramda
-//  (c) 2013-2016 Scott Sauyet, Michael Hurley, and David Chambers
+//  (c) 2013-2017 Scott Sauyet, Michael Hurley, and David Chambers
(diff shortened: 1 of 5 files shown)


davidchambers:
:thumbsdown: from me. `R.indexOf('x', null)` is a type error and should be treated as such. We do have the option of explaining *why* it is an error if we so choose (#1125).

CrossEye:
:-1: from me as well, but more from a no-hand-holding point of view.

PR #2071, opened 2017-02-01
Add yarn.lock for faster and safer dependency installing

MarcoWorms:
gotta go fast
image


davidchambers:
How will this file be kept up to date? Manually? Having to remember to run commands manually is not satisfactory. The file will inevitably get out of sync with __package.json__.

TheLudd:
This looks like something that can be added to the automated test. (Note, I have never used it myself so I cannot vouch for it)

https://yarnpkg.com/docs/cli/check/

Issue #2074, opened 2017-02-03
tap with promise chains


bijoythomas:
I'd like to add tap variant to a promise chain (from composeP or pipeP) with the idea that the tap is dealing with a promise. Does a tapP function have any use? Or do we just do

```
var tapP = function (fn) {
     return function (promise) {
         return promise
             .then(function (x) {
                  fn(x);
                  return promise;
             });
     };
};
```

CrossEye:
My feeling is that this would be best placed in the [Cookbook][cb].

Promises are not well-loved around here, and we're much more likely to be removing Promise-related functionality than to be adding it.


  [cb]: https://github.com/ramda/ramda/wiki/Cookbook

Bradcomp:
Seems like `tap` works just fine inside a promise chain https://goo.gl/uOL9cG

bijoythomas:
Indeed it does. Not sure what I was thinking when I cooked up the idea for tapP. Thanks for pointing it out.

CrossEye:
Even better!

stevenyap:
@bijoythomas @Bradcomp 
`tap` only works if wrapped function does not return a promise. In other words, if `tap` wraps a promise, `pipeP` will not wait not for the promise to return before passing down the chain.

Please see my example here: http://bit.ly/2lClYM6
The first promise ('first') should resolve before the second promise ('second') but it did not.

We need a `tapP` which is a promise-aware `tap` that will resolve the promise before passing down the chain. (I need it... @bijoythomas needs it... I hope more people need it so that it becomes part of the module)

Please advise if this makes sense or if there are alternatives.
If this is useful to the community, I can help to contribute this feature.

A quick idea of `tapP` is like below:
```
var tapP = (fn) => (promise) => {
  return promise.then((result) => {
    return fn(result).then(() => result)
  })
}
```

bijoythomas:
@stevenyap Just to clarify, I was under the wrong impression that the function wrapped with`tap` in a `pipeP` or `composeP` chain would be called the promise itself instead of what it resolves to.

The function wrapped with `tap` gets called with the result of the previous member of the chain in a non-promise chain, and the actual resolution of the promise in a promise chain. This helps so that I just need one function that I can wrap with `tap` regardless of the chain I use.

The other behavior of tap is that its return value is never used. It just calls the wrapped function with the argument and whatever it returns is sort of discarded in the chain. 

See example here: [https://goo.gl/pHoQQ0](url)

[...]

PR #2075, opened 2017-02-03
Adding promise aware tap function

bijoythomas:


CrossEye:
Note:

```shell
> eslint ...
/home/travis/build/ramda/ramda/src/tapP.js
  24:19  error  Unexpected space before function parentheses            space-before-function-paren
  27:3   error  Expected indentation of 4 space characters but found 2  indent
  28:2   error  Newline required at end of file but not found           eol-last
```

But see https://github.com/ramda/ramda/issues/2074#issuecomment-277152148 before bothering to fix it.

bijoythomas:
Thanks for the review @CrossEye 

I'm sort of lost between Promises and Ramda.

I published the request for the 'Q' library here:  https://github.com/kriskowal/q/pull/773
but got the response that it doesn't belong there.

I like ramda's composeP/pipeP, but without a tapP (promise aware), can't debug much and I guess vanilla ramda maybe does not want to work with promises anymore?

So, is this something that belongs in ramda fantasy given that promises are like tasks?

bijoythomas:
tap works fine with composeP and pipeP as pointed out here: [https://github.com/ramda/ramda/issues/2074](url)

CrossEye:
That's the reason people have been down on Promises: they're similar to Tasks/Futures, but not lawful.  One can't reason about them or use them in ways consistent with other types.

But I'm glad the current `tap` already works for you.  That makes things much easier.

Issue #2077, opened 2017-02-06
A "problem"(?) with repeat


notarseniy:
Hello there! Thank you for this very useful library.

Today I ran into some interesting issue, that caused 1 hour of debugging 😅:
```javascript
import { repeat } from 'ramda';

// here i define two dimensional array
let coolArray = repeat([
  repeat(null, 20)
], 20);

// then I change [0][0] element of this array
coolArray[0][0] = true;

// and suddenly i wanna check [1][0] element of this array
console.log(coolArray[1][0]); // excepting 'null' but got 'true' — sadface.rar
```

This happens because 'repeat' function works with reference of value. 

So, there's question: is that behaviour normal in terms of library? Should this will be fixed?
<details>
  <summary>P.S. Fixed code</summary>
  <code>
  let coolArray = map(() => repeat(null, 20), new Array(20)); 
  </code>
</details>

notarseniy:
oh, i have reread docs of repeat and finaly understood what repeat does.
This behaviour is normal, since repeated elements should be equal.

CrossEye:
You might look at [`times`][ti].  I'm not sure if it will do what you want, but it's more flexible than `repeat`.


  [ti]: http://ramdajs.com/docs/#times

notarseniy:
@CrossEye, yeah, exactly this function. thank you!

```javascript
// returns array of 20 arrays of 'null'
let arr = times(() => times(() => null, 20), 20);
```

notarseniy:
I have an idea, @CrossEye. Can we add 'see also times' into docs? I can PR into https://github.com/ramda/ramda.github.io
![image](https://cloud.githubusercontent.com/assets/1084409/22654929/6a06ba38-ec9f-11e6-9bac-d646097b18e7.png)

CrossEye:
absolutely.  A good idea.

PR #2080, opened 2017-02-06
R.repeat: Add to jsdoc @see R.times

notarseniy:
Adding reference into documentation of R.repeat to R.times, because they do somewhat similar things and R.repeat is based on R.times

see also: #2077

### DIFF
--- a/src/repeat.js
+++ b/src/repeat.js
@@ -14,6 +14,7 @@ var times = require('./times');
  * @param {*} value The value to repeat.
  * @param {Number} n The desired size of the output list.
  * @return {Array} A new array containing `n` `value`s.
+ * @see R.times
  * @example
  *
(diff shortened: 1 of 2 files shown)


kedashoe:
:+1: 

Could you please also add a `@see` from `times` to `repeat` as part of this pr?

notarseniy:
Done. @kedashoe

notarseniy:
thanks, @kedashoe!

kedashoe:
thank you for the PR @notarseniy :)

notarseniy:
oh, i forgot! probably we should regenerate https://github.com/ramda/ramda.github.io? it is should be another PR or what?

kedashoe:
No need for another PR, it will be included automatically in the next release when the docs are generated

notarseniy:
Got it.

CrossEye:
Thanks, @notarseniy!

Issue #1958, opened 2016-10-30
poll: transducer usage


davidchambers:
**Q:** How frequently do you use Ramda's transducer functionality?

**A:** Often (:heart:), occasionally (:thumbsup:), or never (:thumbsdown:).

kedashoe:
I do occasionally use it but I would prefer we had multiple builds and that it were not in the default.

jigargosar:
I would love to use it, but this noob needs some examples/tutorials. Once I understand, I be willing to contribute more examples.

MattMS:
Same here. I never get very far each time I try to figure out how to get started with it.

jigargosar:
@kedashoe @davidchambers I rather have it in core library, unless there is a significant benefit of removing it. 

My reasons being, my functional programming improved because every time I go through Ramda docs, the functions that I haven't used reinforce the idea that I need to dig-up/learn more. 

I would like Ramda library to have even more functions. So that noobs like me, get some exposure, and start googling :)

I am not sure what the aim of the library designer is. Perhaps to keep it compact, etc. But this world needs real-functional-programming, not underscore.js , and ramda is the easiest to get started with. So, I hope that more focus is given on introducing/exposing more concepts for newly functional programmer (pun intended). 

I try to contribute as much as I can through [my publication](https://medium.com/the-functional-programmer), but I am still a noob.

kedashoe:
> But this world needs real-functional-programming, not underscore.js , and ramda is the easiest to get started with. So, I hope that more focus is given on introducing/exposing more concepts for newly functional programmer (pun intended).

I am with you a million percent. Ramda was the first thing that finally got me to take a serious look at functional programming. This project is a great way to learn about many functional programming concepts. I do not think though that we need all these things in `ramda.js`. For example, [lenses](https://github.com/ramda/ramda-lens) and [ADTs](https://github.com/ramda/ramda-fantasy) are in separate projects. I think transducers also either belong in their own project (they might actually get even more love and attention this way!) or included in a non-default build.

jigargosar:
@kedashoe I hope you are right. I don't mind separate/projects builds. But I still hope, there is a universal umbrella, that refers to all of them, ADT, lenses, transducers etc. 

A single opinionated package of functional goodies, that noobs can get started with, later they can pick and choose. 

Right now, Ramda is my only hope. Perhaps, I should be the one creating a bundle as separate project. But I don't know how much attentions will it get, but it is better than nothing.

[...]

Issue #1000, opened 2015-04-02
Time to think about what needs to be done for 1.0


CrossEye:
Brainstorming time.

What would we really like accomplished before we publish version 1.0?

Remember that at 1.0, we start playing right with full Semver compatibility, which means, among other things, that we don't break backward compatibility except on major versions.  This probably means slower iteration time, and a slightly more contemplative approach to API changes.

Thanks to all the hard work of so many contributors, we've gotten quite a useful library built, and thanks to some hard work especially by @davidchambers, both the API and the implementation are relatively consistent.  But there's probably still work to do in both API and implementation.

Moreover there are several other places that need work.  There is infrastructure work to be done.  @buzzdecafe has just started switching over to using NPM instead of grunt for many build tasks.  I'd really like to get to the point where everything we need can be built through NPM.  And there are some other issues regarding the running of Sauce, reporting of our Benchmark tests, the possibility of more sophisticated custom builds.

Our documentation has improved remarkably fairly recently.  But I think there is still work to do.  It really should have a more responsive design so that it works well on mobile.  It would be nice to include source code links.  Although we're publishing multiple versions, there's no indication that the docs for older versions are online; we should have a link to them.  Dreaming a little further, it would be nice if a hover of an example offered you the ability to launch it in the REPL.

Speaking of the REPL, it's a very nice piece of work, but it could also use some love.  It would be great to skin it to look like any other of our pages, for one.  But there could be other improvements, like capturing `console.log` statements and displaying their output.  Perhaps we could hover help to Ramda functions used, with the content of the documentation serving as a guide.

So, while the most important issues will always be the core code itself, it's obvious that there are many other things that need to be done.

I'm not sure of the best way to start addressing these points.  Should we just leave open an issue like this?  Should be keep track of this on our little-used wiki?  Something else?

I'm not going to struggle too hard to make this happen, but if it's possible, I would love to see 1.0 come out in May or June.  Does that sound reasonable?

Are there people out there willing to help with some of these tasks?  I don't think Michael, David, and I could possibly do all this in two months.

buzzdecafe:
my $0.02:

# Must:

- [ ] compile a list of functions that should be transducer-comaptible
- [ ] make the list transducer-compatibile
- [ ] decide where exactly we stand on variadic functions in the lib
- [ ] decide on returning a curried function from `curry` and `compose`
- [ ] consider curried, binary `compose` along with another compose function that takes a list

# Would be nice:

- [ ] docs category dropdown
[...]

PR #2082, opened 2017-02-08
proof of concept: decoupling of dispatching and transducers

rickmed:
This attempts to provide a solution to the situation described in https://github.com/ramda/ramda/issues/2083.

The idea is to have a ramda-core in which each function provides a core set of features that can be imported and composed to build additional features from there. This way, there can be how many different flavors of ramda-like libraries you would want.

The current PR takes the map function as an example. Like so, each function takes a (dispatcher, config, dispatchable, however you want to call it) which takes the core features which the core function provides and builds from there. The new map core function showed would only map over native arrays and objects. Then there are two different dispatchers: one adds all the rest of the features that the current ramda/map has and the other one does not (it does not support transducers, currying with placeholders or function composition via map) while providing additional functionality (dispatches to fantasy-land first and then to the map method).

This way you would import (from npm or repo) all the functions in ramda-core, partially apply them with the functionality you want and publish it. In the example provided I would publish a "ramda-fantasy-light" package or something like that. You could publish a flavor which only dispatches to fantasy-land methods (call it "ramda-dictatorship") or another one that dispatches to the method as is (call it "ramda-anarchy"). You could, as well, take the opportunity provide maybe/either wrapped functions and/or add some monads implementations or whatnot for a comprehensive fp toolbox.
Also, it would make more sense to publish individual lightweight functions a la lodash (eg: "ramda.filter").

All in all, I hope, would make ramda more focused, flexible, faster and shareable.

PS:
1) Although I've not read all of the source code to guarantee that all the transducer and default currying can be decoupled without major issues, I'm optimistic.
2) Some test of external functions may fail probably because I hardcoded (temporarily as a proof of concept) a piece of transducer dispatch (commented in code).

### DIFF
--- a/src/dispatchers/current_dispatch.js
+++ b/src/dispatchers/current_dispatch.js
@@ -1,0 +1,67 @@
+var _isArray = require('../internal/_isArray');
+var _isTransformer = require('../internal/_isTransformer');
+var _curry2 = require('../internal/_curry2');
+var _xmap = require('../internal/_xmap');
+var curryN = require('../curryN');
+var _reduce = require('../internal/_reduce');
+var keys = require('../keys');
+
+
+
+/**
+ * Returns a function that dispatches with different strategies based on the
+o * Otherwise, if it has a function with one of the given method names, it will
+ * execute that function (functor case). Otherwise, if it is a transformer,
+ * uses transducer [xf] to return a new transformer (transducer case).
+ * Otherwise, it will default to executing [fn].
+ *
+ * @private
+ * @param {Array} methodNames properties to check for a custom implementation
+ * @param {Function} xf transducer to initialize if object is transformer
+ * @param {Function} fn default ramda implementation
+ * @return {Function} A function that dispatches on object in list position
+ */
+module.exports = function dispatcher(coreImp, fnName) {
+
+  function implemented(fn, functor) {
+    if (arguments.length === 0) {
+      return fn();
+    }
+    var args = Array.prototype.slice.call(arguments, 0);
+    var obj = args.pop();
+    if (!_isArray(obj)) {
+      var fl = 'fantasy-land/' + fnName;
+      var methodNames = [fl, fnName];
+      var idx = 0;
+      while (idx < methodNames.length) {
+        if (typeof obj[methodNames[idx]] === 'function') {
+          return obj[methodNames[idx]].apply(obj, args);
+        }
+        idx += 1;
+      }
+      if (_isTransformer(obj)) {
+    // xmap is hardoded for map but you would look appropriately
+        var transducer = _xmap.apply(null, args);
+        return transducer(obj);
+      }
+    }
+    var func = Object.prototype.toString.call(functor);
+    if (func === '[object Function]') {
+      return curryN(functor.length, function() {
+        return fn.call(this, functor.apply(this, arguments));
+      });
+    }
+    if (func === '[object Object]') {
+      return _reduce(function(acc, key) {
+        acc[key] = fn(functor[key]);
+        return acc;
+      }, {}, keys(functor));
+    }
+
+    return coreImp.apply(this, arguments);
+  }
+
+  return _curry2(implemented);
+
+};
+
--- a/src/dispatchers/noPHnoTrans.js
+++ b/src/dispatchers/noPHnoTrans.js
@@ -1,0 +1,43 @@
+var _isArray = require('../internal/_isArray');
+var curry = require('../simple_curry');
+
+/**
+ * Returns a function that dispatches with different strategies based on the
+o * Otherwise, if it has a function with one of the given method names, it will
+ * execute that function (functor case). Otherwise, if it is a transformer,
+ * uses transducer [xf] to return a new transformer (transducer case).
+ * Otherwise, it will default to executing [fn].
+ *
+ * @private
+ * @param {Array} methodNames properties to check for a custom implementation
+ * @param {Function} xf transducer to initialize if object is transformer
+ * @param {Function} fn default ramda implementation
+ * @return {Function} A function that dispatches on object in list position
+ */
+module.exports = function dispatcher(coreImp, fnName) {
+
+  function implemented(fn, functor) {
+    if (arguments.length === 0) {
+      return fn();
+    }
+    var args = Array.prototype.slice.call(arguments, 0);
+    var obj = args.pop();
+    if (!_isArray(obj)) {
+      var fl = 'fantasy-land/' + fnName;
+      var methodNames = [fl, fnName];
+      var idx = 0;
+      while (idx < methodNames.length) {
+        if (typeof obj[methodNames[idx]] === 'function') {
+          return obj[methodNames[idx]].apply(obj, args);
+        }
+        idx += 1;
+      }
+    }
+
+    return coreImp.apply(this, arguments);
+  }
+
+  return curry(implemented);
+
+};
+
--- a/src/new_map.js
+++ b/src/new_map.js
@@ -1,5 +1,4 @@
 var _map = require('./internal/_map');
-var currentDispatcher = require('./dispatchers/current_dispatch');
 
 /**
  * Takes a function and
@@ -55,4 +54,4 @@ function map(dispatcher) {
 }
 
 
-module.exports = map(currentDispatcher);
+module.exports = map;
--- a/src/simple_curry.js
+++ b/src/simple_curry.js
@@ -1,0 +1,16 @@
+module.exports = function curry(fx) {
+  var arity = fx.length;
+
+  return function f1() {
+    var args = Array.prototype.slice.call(arguments, 0);
+    if (args.length >= arity) {
+      return fx.apply(null, args);
+    }
+    else {
+      return function f2() {
+        var args2 = Array.prototype.slice.call(arguments, 0);
+        return f1.apply(null, args.concat(args2));
+      };
+    }
+  };
+};
--- a/test/new_map.js
+++ b/test/new_map.js
@@ -1,10 +1,15 @@
 var listXf = require('./helpers/listXf');
-
-var R = require('..');
 var eq = require('./shared/eq');
 
+var R = require('..');
+var map = require('../src/new_map');
+var dispatcher1 = require('../src/dispatchers/noPHnoTrans');
 
-describe('map', function() {
+
+describe('map 1: no placeholders|transducer|composition, dispatch fl then .map', function() {
+
+  R.map = map(dispatcher1);
+
   var times2 = function(x) {return x * 2;};
   var add1 = function(x) {return x + 1;};
   var dec = function(x) { return x - 1; };
@@ -14,12 +19,12 @@ describe('map', function() {
     eq(R.map(times2, [1, 2, 3, 4]), [2, 4, 6, 8]);
   });
 
-  it('supports placeholders', function() {
-    eq(R.map(R.__, [1, 2, 3, 4])(times2), [2, 4, 6, 8]);
-  });
-
   it('maps simple functions into arrays', function() {
     eq(intoArray(R.map(times2), [1, 2, 3, 4]), [2, 4, 6, 8]);
+  });
+
+  it('supports placeholders', function() {
+    eq(R.map(R.__, [1, 2, 3, 4])(times2), [2, 4, 6, 8]);
   });
 
   it('maps over objects', function() {
@@ -36,6 +41,11 @@ describe('map', function() {
 
   it('dispatches to objects that implement `map`', function() {
     var obj = {x: 100, map: function(f) { return f(this.x); }};
+    eq(R.map(add1, obj), 101);
+  });
+
+  it('dispatches to objects that implement `fantasy-land/map`', function() {
+    var obj = {x: 100, ['fantasy-land/map']: function(f) { return f(this.x); }};
     eq(R.map(add1, obj), 101);
   });
 


davidchambers:
_Disclaimer: I'm a Sanctuary contributor._

@rickmed, have you considered using [sanctuary-type-classes][1] as the "core"? It provides functions which dispatch to `fantasy-land/`-prefixed methods, and lawful implementations of methods such as `Array#fantasy-land/map`, `Object#fantasy-land/map`, and `Function#fantasy-land/map`.

Adding this dependency would allow us to remove hundreds of lines of code from Ramda, and would allow Ramda and Sanctuary to share a well-designed and well-tested core.


rickmed:
@davidchambers I have. I really like sanctuary and thought hard about how ramda and sanctuary could share a code base. In my ideal world they would. But considering the recurring theme in this issue (don't enforce features to users, but provide a frictionless as possible environment to provide them, which I believe it goes in line with [ramda's philosophy](http://ramdajs.com/), there would be these issues (please correct me where I'm wrong about sanctuary-type-classes):

1. **Dependencies:** sanctuary-type-classes functions depend on sanctuary-type-identifiers and other internal pieces of code. I understand they build up to the type checking system in sanctuary but that is a feature people may not want/need. Also, would make less sense to publish them individually (more below).

2. **Dispatch:** again, dispatching to fantasy-land is a feature people may not want/need.

Regarding 1 and 2. Would send us back to the eternal discussions about type checking at runtime at dispatch mechanisms.

3.**Performance**: consider Arrayfilter and ArrayMap. AFAIK sanctuary-type-classes dispatches to the  array's method. Ramda implements [this filter](https://github.com/ramda/ramda/blob/v0.23.0/src/internal/_filter.js) and [this map](https://github.com/ramda/ramda/blob/v0.23.0/src/internal/_map.js). I assume the latter are more performant.

More importantly, if you ask me, that is exactly the ramda's position in the js world. Those two algorithms are good, performant, independent and lightweight pieces of code. As a made the PR, I took a look at lodash's fp version. It made me curious why they it doesn't have individual functions published as for the OO versions. They problem is that they weren't coded for immutability from the get go, so they need to pass to a converter, which make them impractical to publish them.

The curried versions of this PR's filter and map (for arrays and objects) would be around 15-20 loc. Compose, pipe, curry, lift, etc much less. For a user just using promises, she can download a few of these to compose them and build super powerful pipelines -enabling her into fp.

Back to ramda-sanctuary-type-classes, there would be other issues like the monolithic index.js that I think would be secondary. So why not the other way around? sanctuary-type-classes to use ramda-core and build functionality from there as desired.

CrossEye:
I will have to look at this a lot more closely later.

As to the concepts, I really like the idea of separating out the transducers code; it adds a great deal of weight to the codebase and although there is a lot of potential there, I don't think it's lived up to its promise so far.

I'm far less convinced about the dispatching and about separating the currying with placeholders.  I'm also concerned by the possible combinatorial explosion of Ramda versions.

But I recently returned to an earlier experiment that might address certain implementation and performance concerns about many of these things -- everything, in fact, except the transducers.  I've been much slower than I'd like with this metadata approach.  (My main computer has developed some issues at the same time I've had to replace three major household appliances in a few weeks.  So I haven't fixed it, and finally resorted to using an ancient machine.)  But I'm relatively happy with these basic ideas, and might like to combine some of them with what you're suggesting.  I won't be ready for at least a week, though, and I don't want to slow down what you're doing.  This is very interesting.

davidchambers:
> sanctuary-type-classes functions depend on sanctuary-type-identifiers and other internal pieces of code.

True, but [sanctuary-type-identifiers][1] is tiny. This is all there is to it:

```javascript
//  type :: Any -> String
function type(x) {
  return x != null &&
         x.constructor != null &&
         x.constructor.prototype !== x &&
         typeof x.constructor['@@type'] === 'string' ?
    x.constructor['@@type'] :
    Object.prototype.toString.call(x).slice('[object '.length, -']'.length);
}
```

As for "internal pieces of code", at least 90% is necessary to support the exported functions. Less than 10% (I estimate) is dedicated to the `TypeClass` values such as [`Z.Functor`][2] exported for use with sanctuary-def.

> dispatching to fantasy-land is a feature people may not want/need

In that case sharing code is more challenging. We could at least share the implementations of `Array.map`, `Object.map`, `Function.map`, etc.

> AFAIK sanctuary-type-classes dispatches to the array's [`map`] method.

It does. Here's the current implementation:

```javascript
//  Array$prototype$map :: Array a ~> (a -> b) -> Array b
function Array$prototype$map(f) {
  return this.map(function(x) { return f(x); });
}
```

If there's a performance benefit to an imperative implementation, we can easily update this. I'd love to see a pull request for this. :)

> So why not the other way around? sanctuary-type-classes to use ramda-core and build functionality from there as desired.

sanctuary-type-classes is far more modular than Ramda. Compare the `map` implementations:

```javascript
function map(f, functor) {
  return Functor.methods.map(functor)(f);
}
```

```javascript
var _map = require('./internal/_map');
var _reduce = require('./internal/_reduce');
var _xmap = require('./internal/_xmap');
var curryN = require('./curryN');
var keys = require('./keys');

module.exports = _curry2(_dispatchable(['map'], _xmap, function map(fn, functor) {
    case '[object Function]':
      return curryN(functor.length, function() {
        return fn.call(this, functor.apply(this, arguments));
      });
    case '[object Object]':
        acc[key] = fn(functor[key]);
        return acc;
      }, {}, keys(functor));
    default:
      return _map(fn, functor);
  }
}));
```

Note that Ramda's implementation switches based on type and provides inline implementations for `Function#map` and `Object#map` (the implementation for `Array#map` lives in `_map`, presumably).

The `map` function in sanctuary-type-classes dispatches to the appropriate method, which may be an implementation (such as `Array$prototype$map` above) provided by the library itself. As a result the `map` function itself is trivial, and each of the `fantasy-land/map` methods are defined separately.

In summary, Ramda and Sanctuary should share the implementations of `Array#fantasy-land/map` and friends at the very least. I'd been hoping that sanctuary-type-classes already allows us to do exactly this.


[1]: https://github.com/sanctuary-js/sanctuary-type-identifiers
[2]: https://github.com/sanctuary-js/sanctuary-type-classes#Functor

rickmed:
@CrossEye 
#### Dispatcher
(regarding the name see below), I inverted the control to it precisely to support a simpler currying version. In its current implementation it has placeholder support built-in. I don't want to be "forced" to have code in my code base that I don't use. Specially, in a "toolbox" kind of library. But also I don't want to "force" people otherwise. So hopefully, the "dispatcher" makes frictionless to support whichever version of currying you would like. To me that is what fp is all about (part of :)); small functions that do 1 thing and compose them to create additional features. 

#### Combinatorial Ramdas
It brings me back to the original post. IE: there needs to be better educational resources, not a more complex library as a way to educate. If you look at http://ramdajs.com/docs/#map, it takes 5 long lines to explain what it does (not including the fantasy-land's functor explanation). This confuses fp newcomers and disappoints (eg: dispatch discussion) experienced users looking to use the library in a different way (ie: reuse the code). I imagine the docs in home of map with just "Ramda provides suitable map implementations for Array and Object, so this function may be applied to [1, 2, 3] or {x: 1, y: 2, z: 3}." and a link to a list of ramda-flavors (officially or user maintained) to point experienced users that understand what the differences entail (blind map dispatch, FL, currying versions etc.). The list can be arbitrarily limited. For a "toolbox" kind of library, massively reusing its flexible source code is exactly what I would expect.
Things like the cookbook should be directly linked in the home page to teach newcomers how to easily compose functions (Re: placeholder support).  

#### Forward
I would also like to move promptly. It would be great if you could advance a sketch or something :)


PS: About "dispatching" -just to make sure I get right that it is about the name, or at least to document. I kept the name "dispatching" to keep a reference to the current implementation. But it has nothing to do with it (in the object oriented and current implementation sense). Meaning, it is not calling any object's method -as the current "_dispatchable" is. We would change the name to reference the "function passing"/"inversion of control".

rickmed:
> sanctuary-type-classes functions depend on sanctuary-type-identifiers and other internal pieces of code.
> True, but sanctuary-type-identifiers is tiny. This is all there is to it:
> 
> //  type :: Any -> String
> function type(x) {
>   return x != null &&
>          x.constructor != null &&
>          x.constructor.prototype !== x &&
>          typeof x.constructor['@@type'] === 'string' ?
>     x.constructor['@@type'] :
>     Object.prototype.toString.call(x).slice('[object '.length, -']'.length);
> }
> As for "internal pieces of code", at least 90% is necessary to support the exported functions. Less than 10% (I estimate) is dedicated to the TypeClass values such as Z.Functor exported for use with sanctuary-def.


In any case, again, why "force" users to have code they don't want?



> In that case sharing code is more challenging.

how so?



> We could at least share the implementations of Array.map, Object.map, Function.map, etc.

I agree, that is what the current PR is all about.

> If there's a performance benefit to an imperative implementation, we can easily update this. I'd love to see a pull request for this. :)

what's wrong with the version I linked to?

> sanctuary-type-classes is far more modular than Ramda. Compare the map implementations:

Currently, yes. That's the whole point of the (PR)[https://github.com/rickmed/ramda/blob/262415315fc30531f61f10800f454a5f49d083d1/src/new_map.js]. Which can be simplified.

> In summary, Ramda and Sanctuary should share the implementations of Array#fantasy-land/map and friends at the very least. I'd been hoping that sanctuary-type-classes already allows us to do exactly this.

Again, why not build sanctuary on top of ramda-core?

davidchambers:
> > In that case sharing code is more challenging.
>
> how so?

Because Ramda can't simply depend on sanctuary-type-classes.

> Again, why not build sanctuary on top of ramda-core?

Because it doesn't exist. :stuck_out_tongue_winking_eye:

If and when ramda-core has been published to npm, we'll certainly consider having sanctuary-type-classes depend on it.

rickmed:
> In that case sharing code is more challenging.
> how so?
> Because Ramda can't simply depend on sanctuary-type-classes.

oh, right. Exactly, my point is to share core code. The rest of the features can be built on top.

> Again, why not build sanctuary on top of ramda-core?
> Because it doesn't exist. 😜

Well let's go for it! :)

rickmed:
Just to be clear, I don't mean that npm/ramda would necessarily mean ramda-core. Personally, I prefer a code base with no transducers, no placeholders and dispatch to FL first, then object's method, but I don't pretend for that to be npm/ramda. If decided, npm/ramda can keep exactly the same features it currently has -I would be happy to publish npm/my-lib (it doesn't even has to have the ramda prefix).

buzzdecafe:
Just checking in. FWIW, I support this line of inquiry. To my mind it is worth asking how we got here and re-evaluating if we've made the right decisions.

And FWIW, my _opinions_ on what I'd like to see stay (👍 ) or go (👎 )

* Fantasy-Land 👍 
* Transducers 😑 
* Placeholders 👎

rickmed:
https://github.com/ramda/ramda/issues/1000

rickmed:
Hopefully I made a mistake, but I actually "went through ramda's code and couldn't believe what I found out". R exports 238 functions.

CrossEye:
Same here:

* Fantasy-Land 👍 
* Transducers 😑 
* Placeholders 👎

davidchambers:
Since @buzzdecafe, @CrossEye, and I are all in favour of dispatching to `fantasy-land/`-prefixed methods, I'll revive #1938!

CrossEye:
@rickmed:

>It brings me back to my original post. IE: there needs to be better educational resources, not a more complex library as a way to educate. [ ... ] I imagine the docs in home of map with just
>> Ramda provides suitable map implementations for Array and Object, so this function may be applied to [1, 2, 3] or {x: 1, y: 2, z: 3}.

> and a link to a list of ramda-flavors (official and/or user maintained) to point experienced users to who understand functional concepts and what their differences entail [ ... ]

I don't see Ramda as being primarily an educational tool.  Or rather, it was a library that @buzzdecafe and I started as a way to educate ourselves, but it has been the *building* of it that is educational, not the library itself.  We did it as a way to understand better functional concepts that made sense but were not part of our regular routines.  Many of these ideas weren't particularly practical to use in JS.  Ramda has been an attempt to make them more practical.

When it started to catch on, we were flabbergasted.  This was a little toy library for our own uses that only a few friends knew about.  It has shifted, and with the addition of @davidchambers and other contributors, it has grown into a more mature library, but it still maintains the focus of being an attempt to make practical in JS ideas from the FP community.

Thus, while transducers are not essential (they were a Clojure experiment ported with good reason to JS, but we might have been too quick to adopt them) and the placeholders might not be worth their implementation costs now that ES6+ is so prevalent, having `map` operate on arbitrary functors is quite core to what I think of Ramda as doing.  So dispatching to `map` methods, and supplying implementations for built-in functors -- Arrays, Objects, Functions -- are the baseline requirement.  Details of how we dispatch are certainly up for debate, and there's a serious debate about how much dispatching we should be doing.  But losing this dispatch would be for a different library, not for Ramda.  At least that's how I see it.


> I would like to move promptly as well. It would be great if you could advance a sketch or something :)

Here's the basic idea:

There is one internal ramda function (I'm calling it `ramda` for now) that is responsible for calling functions that have been configured with such helpers as `curry`, `flip`, `partial`, etc.  It stores a configuration object with a reference to the original function, with all parameters that have been supplied so far, a mapping of where subsequent parameters need to go to be fed to the original function.  A call to, say, `map`, calls this function, which checks where to place the incoming parameters and whether all of the required parameters are supplied (the way `curry` does, but somewhat more sophisticated), and either calls the original function or returns a new one with an updated config.  I keep vacillating between making the config a property of the function and storing it in a closure; I do expect something will eventually prevent the closure, but so far it continues to work.  This would also give me an easy place to use Dr. Boolean's [Debugging Functional][df] techniques, which would be useful.  It also would give us the place to do a lodash-style `rearg` if we like.

The point is that there wouldn't be nested calls to `curry`; placeholder checks would be centralized (if we decided to keep them; as `rearg` would be a real alternative.)  Each call to a Ramda-wrapped function should involve one call to this function, which would either call the original function or return a new further-configured function.  It should drop our stacks tremendously.  I don't have my head wrapped firmly enough around transducers to know how to integrate them with this vision or even if its possible. 

Anyway, that's the idea.  All my experiments are on a machine with no network connections.  I will try to fix that this weekend ... assuming I have no more appliance problems!  :frowning: 

I don't know how much or how little that will conflict with what you're doing.

  [df]: https://medium.com/@drboolean/debugging-functional-7deb4688a08c

rickmed:
@CrossEye 

> When it started to catch on, we were flabbergasted. This was a little toy library for our own uses that only a few friends knew about. It has shifted, and with the addition of @davidchambers and other contributors, it has grown into a more mature library, but it still maintains the focus of being an attempt to make practical in JS ideas from the FP community.

Not sure if it has not been apparent from my other posts but I really appreciate the work that you have done with ramda. You have made a direct and positive impact in my life by helping me write (and learn along the way) fp in js which I greatly enjoy. **So thank you**. Also, that same impact has been made to a lot of people and growing by the day. So much, that many of them use ramda in serious projects (their jobs or otherwise). Myself included. I believe that, whether you expected/wanted it or not, ramda is currently in an important position. If you want to write "js fp style" there is no way around using ramda. This is quite a combination; javascript which is the most popular language in the world and fp -from which people choose whole career paths based on it. 

My goal with these contributions is to help accelerate moving the library forward however I can and best serve the above mentioned users -I don't really care which contributions it involves as long as it achieves it. I felt sad when reading the issue from almost 2 years ago for ramda 1.0 and it died. I made this PR to provide a solution to the never ending discussion about the configuration (dispatching, placeholders, transducers...). The dispatching discussion alone have stalled many PRs at least for 5 months -I understand also it's an unfunded OS project as many others.

I apologize if I'm intruding, being insensitive or whatnot. Or all of them :) 

> I don't see Ramda as being primarily an educational tool. 

I'm not saying that either. However, given the its position, it makes a big impact in that regard. (imo, ramdajs.com is super important). My argument is that its base code should not be used to educate.

> Here's the basic idea:
> 
> There is one internal ramda function (I'm calling it ramda for now) that is responsible for calling functions that have been configured with such helpers as curry, flip, partial, etc. It stores a configuration object with a reference to the original function, with all parameters that have been supplied so far, a mapping of where subsequent parameters need to go to be fed to the original function. A call to, say, map, calls this function, which checks where to place the incoming parameters and whether all of the required parameters are supplied (the way curry does, but somewhat more sophisticated), and either calls the original function or returns a new one with an updated config. I keep vacillating between making the config a property of the function and storing it in a closure; I do expect something will eventually prevent the closure, but so far it continues to work. This would also give me an easy place to use Dr. Boolean's Debugging Functional techniques, which would be useful. It also would give us the place to do a lodash-style rearg if we like.
> 
> The point is that there wouldn't be nested calls to curry; placeholder checks would be centralized (if we decided to keep them; as rearg would be a real alternative.) Each call to a Ramda-wrapped function should involve one call to this function, which would either call the original function or return a new further-configured function. It should drop our stacks tremendously. I don't have my head wrapped firmly enough around transducers to know how to integrate them with this vision or even if its possible.

Thank you. I'm having trouble imagining it in hard code, though. The first paragraph makes me think it's a much more convoluted solution to this PR's and the second one makes me think they are not all that different (specially if using closures). Without seeing some code, I'll risk it with the following questions:

> is responsible for calling functions that have been configured with such helpers as curry, flip, partial, etc.

This part is what I think is confusing me the most. Not clear _how_ the configuring would work. Who configures who?
Also, would the function be lightweight as to publish them individually?

> It stores a configuration object with a reference to the original function, with all parameters that have been supplied so far, a mapping of where subsequent parameters need to go to be fed to the original function. A call to, say, map, calls this function, which checks where to place the incoming parameters and whether all of the required parameters are supplied (the way curry does, but somewhat more sophisticated), and either calls the original function or returns a new one with an updated config.

Is it for placeholders support? I wouldn't get why a simple curry wouldn't suffice.

> The point is that there wouldn't be nested calls to curry; 

Yes. It's currently a problem since all functions export curried functions which some of them are used by other ones and so on. ie: currying is too coupled.

> It should drop our stacks tremendously.

This PR solves that two. Both proposed refactors would take a fair amount of work though. Because of the above.

> This would also give me an easy place to use Dr. Boolean's Debugging Functional techniques, which would be useful. It also would give us the place to do a lodash-style rearg if we like.

+1! Imo, should be an important part of what ramda provides. I've been doing some research about it but wanted to open issues one by one.

> Thus, while transducers are not essential (they were a Clojure experiment ported with good reason to JS, but we might have been too quick to adopt them) and the placeholders might not be worth their implementation costs now that ES6+ is so prevalent, having map operate on arbitrary functors is quite core to what I think of Ramda as doing. So dispatching to map methods, and supplying implementations for built-in functors -- Arrays, Objects, Functions -- are the baseline requirement. Details of how we dispatch are certainly up for debate, and there's a serious debate about how much dispatching we should be doing. But losing this dispatch would be for a different library, not for Ramda. At least that's how I see it.

This PR's refactor would comply with all of those requirements and more. Additionally, what if the requirement of dispatching to functors' methods changes? (however unlikely as I assume the placeholders support was in the past). What if the fantasy-land api changes? What if you want to go back to transducers support?, etc.

All of these refactors would take zero changes in core code. Fantasy-land changed? change a few _characters_ . New functionality, build from there however you want.

> But losing this dispatch would be for a different library, not for Ramda.

This really confuses me. This does not suggest anything of that sort.

> Details of how we dispatch are certainly up for debate, and there's a serious debate about how much dispatching we should be doing.

This PR aims exactly at that.

CrossEye:
@rickmed:

> I apologize if I'm intruding, being insensitive or whatnot. Or all of them :)

You are not.  Not in the least. :smile:

> My argument is that its base code should not be used to educate.

It can't be right now, and I do think that this is a shame.

> The first paragraph makes me think it's a much more convoluted solution to this PR's and the second one makes me think they are not all that different (specially if using closures).

Note that the goal of my work is entirely different from yours, even if they might have some overlap in some of their effects.  

Yours, as I understand it,, is about simplifying the source code and allowing us to build versions of the library with different subsets of our current feature-list -- not the functions, things like default currying, FantasyLand and other dispatching, placeholders, transducers, etc.

Mine is more about call stacks, debuggability, and -- most likely -- performance.  My goal is that if you step into a Ramda-wrapped function, you walk through what will quickly become a very recognizable process of parameter juggling that leads you down one of only two paths: returning a further-configured function, or calling the original function with a full set of parameters.  If you call a wrapping function -- `curry`, `partial`, `flip`, `arity`, `binary`, etc -- it returns a function that will call that one function with the configuration object in scope.

I will either try to sneaker-net the code from that machine, get that machine working, or recreate it.  It's not that much code.  But it's not entirely working yet, and I still need some time to complete it.

>> The point is that there wouldn't be nested calls to curry;

> Yes. It's currently a problem since all functions are being exported curried by default, which some of them are used by other ones and so on. ie: currying is too coupled.

Well the auto-curried function is one of the things which most users first find helpful about Ramda.  It was probably the single most important initial driving factor for the library.

>> It should drop our stacks tremendously.

> This PR solves that too. Both proposed refactors would take a fair amount of work though. Because of the above.

Yes, and I can't tell if yours can be layered on top of mine... I'm pretty sure the other way wouldn't work.


> Additionally, what if the requirement of dispatching to functors' methods changes? 

I don't mean to suggest this is a feature that `map` should have.  I mean that this is the actual definition of `map`.  

>> But losing this dispatch would be for a different library, not for Ramda.
>I'm confused. This does not suggest anything of that sort. https://github.com/ramda/ramda/pull/2082#issuecomment-278403235

I meant to suggest that the other features discussed, transducers, placeholders, seem less central to Ramda, and might well be relegated to opt-in separate builds, but that I can't really imagine a Ramda build that doesn't include dispatching `map`.

rickmed:
@CrossEye 

Thank you for the quick reply -in the context of your technical issues.

> My argument is that its base code should not be used to educate.
> It can't be right now, and I do think that this is a shame.

I was referring to the transducers support (among others) which many people expressed were interested in for educational [purposes](https://github.com/ramda/ramda/issues/1958). But I should open a separate issue about that.

> Yours, as I understand it,, is about simplifying the source code and allowing us to build versions of the library with different subsets of our current feature-list -- not the functions, things like default currying, FantasyLand and other dispatching, placeholders, transducers, etc.

My motivation was the one stated above. It just so happens, logically, that the solution was to decouple code which makes the core extendable -there was demand for this in the past btw- and future-proof.

> Note that the goal of my work is entirely different from yours, even if they might have some overlap in some of their effects.

We would overlap in less call stacks and performance. Mine would not change debuggability -apart of a consequence of having less stacks. I would propose the application of the techniques in drboolean's article in a different PR.

> Yes, and I can't tell if yours can be layered on top of mine...

I only speculate from your description, but I think it may actually work.

> Well the auto-curried function is one of the things which most users first find helpful about Ramda. It was probably the single most important initial driving factor for the library.

> but that I can't really imagine a Ramda build that doesn't include dispatching map.

Again, this is not what this PR is about. No one would use ramda-core directly. It would make no sense at all a ramda without curried functions. I refactored it that way to settle the discussion about placeholders. Then I realized it would also solve the nested currying problem. It would also support users which, for any obscure reason, can't use arrow functions as placeholders replacement.

idem; map. It addresses [this](https://github.com/ramda/ramda/pull/1900#issuecomment-249395452). I share your view in that comment. I also share Fantasy-Land 👍, Placeholders 👎. But I would want to be able to not choose between ramda and a library that, for example, complies with fl but needs to maintain a legacy map -and current users in the same position. Additionally, I speculate, it would be able to integrate with libraries like Immutable.

customcommander:
@CrossEye What do we do with this one?

CrossEye:
This is one of those things that keeps coming back to haunt us.  I'm adding the `Maturation` tag and closing it.  All the `1.0 Discussion` and `Maturation` tags to me are ideas that we want to revisit post v1.0 (or more likely post v2.0).  This to me is at the top of that list.  It might make 2.0 easier to write.

Issue #2081, opened 2017-02-07
promise based reduceWhile


Cinderella-Man:
I was thinking would there be an interest in something like reduce while but working with promises? It looks like chain of responsibility - here's a draft that I needed to do yesterday:

https://gist.github.com/Cinderella-Man/2c9d6b89c40a62901e9cbbd6dab50778

Also would there be a chance to add simple methods like: isNumber, isString - I know that it can be achieved by combining methods but that's a little bit too verbose and I was thinking that this is what we would like to avoid?

CrossEye:
We've mostly been avoiding Promise-related functionality.

Most regulars around here prefer Futures/Tasks and find Promises too difficult to think about.

I do like the idea of this functionality.  I would love to see an equivalent proposal for RamdaFantasy/Future.

Cinderella-Man:
Ok. I'm new to Futures/Taks so would need to give it a look but still I'm happy to give it a try and contribute some PR if you are interested?

Any comments regards `isString` for example?

CrossEye:
@Cinderella-Man:

> Ok. I'm new to Futures/Tasks so would need to give it a look but still I'm happy to give it a try and contribute some PR if you are interested?

Very much so.

> Any comments regards `isString` for example?

Sorry, missed that first time.

We made a decision a while ago to forgo those.  `R.is` covers these cases well enough, and makes it easy enough to create such helpers if the user wants without adding cruft to the library.

```js
R.is(String, 'foo'); //=> true
R.is(Array, [1, 2, 3]); //=> true
R.is(Array, 'foo'); //=> false

class Rectangle { /* ... */ }
R.is(Rectangle, new Rectangle(5, 3)); //=> true
R.is(Rectangle, {height: 3, width: 5}); //=> false

const isString = R.is(String);
isString('bar'); //=> true
isString([1, 2, 3]); //=> false
```

etc.

PR #2086, opened 2017-02-09
Added anyPassT - Task based chain of responsibility like function

Cinderella-Man:
Hi,

As per my recent conversation(promise based reduceWhile #2081), I wrote a snippet of code that would work as a chain-of-responsibility functionality. I was thinking for a long time how to name it, it's not like reduce as you can't collect data on the way (it would make it more complex but it could be useful to add it as well - the same flow but passing starting value). The only thing it's similar to is `anyPass` but that's not perfect either. `T` suffix is a guess - big time.

I know that implementation is probably not utilising half of the methods that you are using in your files but tried my best and made it browser friend (apply instead of spread etc) - I need to start somewhere.

I'm keen to get your opinion.

PS. 5 starts for contributors guide - just easy to do it!

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,6 +11,7 @@ module.exports = {
   and: require('./src/and'),
   any: require('./src/any'),
   anyPass: require('./src/anyPass'),
+  anyPassT: require('./src/anyPassT'),
   ap: require('./src/ap'),
   aperture: require('./src/aperture'),
   append: require('./src/append'),
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "browserify": "10.x.x",
     "cli-table": "0.3.x",
     "commander": "2.5.x",
+    "data.task": "^3.1.1",
     "dox": "latest",
     "envvar": "1.x.x",
     "escodegen": "1.4.x",
--- a/src/anyPassT.js
+++ b/src/anyPassT.js
@@ -1,0 +1,48 @@
+var _curry3 = require('./internal/_curry3');
+var head = require('./head');
+var isEmpty = require('./isEmpty');
+var slice = require('./slice');
+var tail = require('./tail');
+
+/**
+ * Takes a list of functions producing Tasks, a predicate and a value. Each function is
+ * called with a value to produce Task which is chained to retrieve it's result.
+ * Predicate is applied to that result and if results in true, anyPassT won't call next
+ * Task producer function.
+ *
+ * @func
+ * @memberOf R
+ * @since v0.24.0
+ * @category Logic
+ * @sig [(*... -> Task)] -> (* -> Boolean) -> [*] -> *
+ * @param {Array} functionsList An array of Tasks producting functions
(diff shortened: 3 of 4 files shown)


kedashoe:
Thank you for the PR @Cinderella-Man :) This feels a bit narrow to include in ramda to me. Have you seen https://github.com/fantasyland/fantasy-land#chainrec? Implemented for example here: https://github.com/fluture-js/Fluture#chainrec. Does it look like you could achieve what you are aiming for with that?

rickmed:
@Cinderella-Man good stuff! But imo, specially regarding fp in js, it is better to teach a man to fish instead of giving him the fish (from ramda's perspective, not mine). Let me explain:

One of the characteristics of fp is that things are (aim to be) composable. This means that with a few fundamental concepts and tiny little functions, you can compose them to solve wildly different problems. Thus, you don't need to remember the names of a bunch of individual functions that solves specific problems.

Regarding what anyPassT is solving. What if I wanted the same functionality but slightly different? Fire up all tasks in parallel and return the first one that passes the predicate? Or to concat successful results? Or return a custom thing when none passed the predicate (more on that below)? What about a combination of some of the those? You see, if ramda were to provide all this functions, it would not only grow infinitely, but encourage users to look for the one that solves her specific problem at hand, while not profiting from one of the big benefits, instead of applying the fundamental concepts. 

Therefore, it makes sense for ramda to be an enabler (encourage fishing), instead of a provider (give the fish). This does not work in "imperative style" and that is part of the reason lodash/underscore need to be providers (with so many utilities).

Your example showcases how to combine good stuff such as async operations with tasks, recursion, lifting values... So I believe it should be consired for the [cookbook](https://github.com/ramda/ramda/wiki/Cookbook) or similar.

I'm far from an expert, but if you ask me, here is how I would write anyPassT:

```javascript
const anyPassT = (functionsList, predicate, arguments) =>
  pipe(
    head,
    apply(arguments),
    chain(x =>
      predicate(x) ?
        task.constructor.of(x) :
        nth(1, functionsList) ?
          task.constructor.reject('unsuccessful') :
          anyPassT(tail(functionsList), predicate, arguments)
    )
  )(functionsList)
```

Setting apart cosmetics and ecma versions, the things to note are the rejection of the task instead of task.of(false) and the little trick of calling pipe at the end so that you can use it and the rest of the arguments within the closure. Also, this version would reject the whole task if any of them fail, without trying the next (not sure if that is what you want).  

PS: I'm not 100% sure but the test are passing bc they are not being ran at all. Being async so you would need to use the completion callback in the tests.

CrossEye:
I was suggesting in #2081 that this might be a good fit for [Ramda-Fantasy][rf].  I have not been very active in that repo, but if it fits anywhere, I would think it would fit there, for the [Future][fu] monad.

I agree with @kedashoe that this is too specific for Ramda proper.

  [fu]: https://github.com/ramda/ramda-fantasy/blob/master/src/Future.js

PR #2088, opened 2017-02-12
Update Chain Documentation

keithbro:
Clarifies that functions can be composed and treated as chains.


davidchambers:
Please don't update __dist/ramda.js__ (see the [guidelines for contributing][1]).


[1]: https://github.com/ramda/ramda/blob/master/CONTRIBUTING.md

keithbro:
Apologies, commit removed.

acconrad:
[Documentation has already been updated](https://github.com/ramda/ramda/blob/master/source/chain.js#L15) so I think this PR can be closed.

Issue #2089, opened 2017-02-13
R.rename?


tycho01:
Every now and then I wonder "should this kind of function be in Ramda?".
I just went over the docs for Python's `Pandas` library (R-like `DataFrame`s), trying to get a sense of parts it currently made more convenient. Among these was an object to rename columns in a dataframe. I couldn't directly see a Ramda equivalent easier than some combination of `pipe`, `toPairs`, `map`, and `fromPairs`, so given that I'd propose a `R.rename({ a: 'b', foo: 'bar' }, { a: 1, foo: 2 }) -> { b: 1, bar: 2 }`. Pretty random, I know, especially since Pandas is by no means a direct competitor to Ramda users. Otherwise maybe fun for cookbook.
Now to get corrected on a super-straightforward equivalent after all!

On a side-note, other bits that stood out to me from its [cheat sheet](https://github.com/pandas-dev/pandas/blob/master/doc/cheatsheet/Pandas_Cheat_Sheet.pdf) as not having evident Ramda equivalents were NA handling (we'd probably end up deferring to Maybes?) / sparse data, stat-like functions e.g. binning into quartiles/percentiles, as well as maybe pivoting/unpivoting (which is kinda interesting too). Not a common comparison, I know, but that's why I felt it'd be interesting to check. :)

Bradcomp:
I don't have a super simple equivalent, but this is in the cookbook: https://github.com/ramda/ramda/wiki/Cookbook#rename-keys-of-an-object

For what it's worth, I do use this function on occasion.

CrossEye:
@tycho01:
Does @Bradcomp's suggestion fit the bill for you?

As to the Panda functions, there are some nice things in there, including some that I might personally use, but I see little that seems to make sense in Ramda.  Do you?

mattgrande:
I used the `rename` function as provided in the cookbook more than once, so I opened a PR for this (#2094).

If you want to merge it, great. If not, no big deal. I can see why it's not strictly "ramda-y," but I, personally, think it's helpful.

adispring:
I also use `rename` sometimes.

tycho01:
I tried to add a few ideas from that Pandas to the cookbook (binning, pivoting, joins).

I'd be inclined to defer judgement on `rename`, so I'll mark this thread as resolved; I'm probably not in a great position to judge the identity of Ramda or what it means to people.

[...]

PR #2094, opened 2017-02-21
Add a rename function.

mattgrande:
As suggested in #2089; Code modified from the [cookbook](https://github.com/ramda/ramda/wiki/Cookbook#rename-keys-of-an-object).

**Edit:** I missed that there was a linter (should've checked, my apologies). Consider adding that step to `CONTRIBUTING.md`?


adispring:
I just modified the implement of `rename` more neat. https://github.com/ramda/ramda/wiki/Cookbook#rename-keys-of-an-object

davidchambers:
@adispring, while you're at it you should remove the incorrect use of `||` in `keysMap[key] || key`. This will not work correctly if the left-hand operand evaluates to `''`.

src/rename.js
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Creates a new object with the own properties of the provided object, but the
+ * keys renamed according to the keysMap object as `{oldKey: newKey}`.
+ * When some key is not found in the keysMap, then it's passed as-is.
+ *
+ * @sig {a: b} -> {a: *} -> {b: *}
+ */
+module.exports = _curry2(function rename(keysMap, obj) {
+  return reduce(function(acc, key) {
+    acc[keysMap[key] || key] = obj[key];
```
davidchambers: We cannot use `||` here, as the left-hand side may evaluate to `''`.

mattgrande: Huh, I didn't realise that `''` would be a valid key for an object... Although I guess it makes sense.

I'll replace this with a null/undefined check.

davidchambers: > I'll replace this with a null/undefined check.

That would be problematic too, due to inherited properties:

```javascript
> ({})['toString'] || 'toString'
function toString() { [native code] }
```

Using [`R.has`][1] might be a good idea.


[1]: http://ramdajs.com/0.23.0/docs/#has

CrossEye:
> Sorry, null and undefined can also be valid Object keys
> (edit: ignore that, they just coerce to the strings 'null' and 'undefined')

It's an interesting philosophical point though.  :smile:

Because they coerce also on a get, it's only because of things like `Object.keys` that you can even tell that they're not really allowed.  Thus:

```js
const obj = {};
obj[null]; //=> undefined
null in obj; //=> false

obj[null] = 42;
obj[null]; //=> 42
null in obj; //=> true
```

You can tell that it's not actually a property by noting that 

```js
Object.keys(obj); //=> ["null"]
```

But that's fairly subtle.

mattgrande:
Updated; Using `has` instead of null check. Sorry for the delay.

adrian-gierakowski:
I have used exactly this kind of function quite a few times, however I think it would be more beneficial to add more general `mapKeys` function ( like https://lodash.com/docs/4.17.4#mapKeys ). `mapKeys` could work with functions which accept only one arg ( key ) and `mapKeysWithValue` could take a function which accepts both `key` and `value` ( o `value` and `key` ). `rename` could then be expressed in therms of one of those

CrossEye:
@adrian-gierakowski:

Although a `mapKeys` function (and/or a `mapKeysWithValue` one) might be useful, I don't think they'd take the place of this one.  While such *could* be used to create `renameKeys` behavior, it would not be particularly comfortable for a user to do so, and it seems a reasonably common request.

So if we do `mapKeys`, let's let it stand separate from this one.

mattgrande:
Trying to clear out some long-open PRs. Do people still have interest in this one, or should I close it?

CrossEye:
I have no strong feelings either way.  Others in @ramda/core ?

mattgrande:
Seems like there was never much demand for this; Closing.

rdsedmundo:
I wish we had this. :)

PR #2098, opened 2017-02-23
add R.addKey

KiaraGrouwstra:
tests WIP


CrossEye:
Don't know how this PR got missed entirely, but somehow it did, and it's been sitting here ever since.  Sorry.

I understand the parallel with `addIndex`, but I think the need is much less common.  I would expect those who need to do something like this to build it on top of `toPairs` / `fromPairs` or their native counterparts.

So I'm closing this now, but feel free to comment if you still want to argue for it and try to open it back up.

PR #2100, opened 2017-02-25
fix insert.js typo

adispring:


CrossEye:
I wonder if this should read "at position `index`".

adispring:
at the/a specified/given index ?

CrossEye:
"at the specified index" sounds fine to me.  The original used the name of a parameter, which we do occasionally, but I don't much like.  This is better.

CrossEye:
Thank you.

adispring:
😃

PR #2105, opened 2017-03-01
optimize unless type

adispring:


CrossEye:
I don't think I agree here.  We try to document what we've taken to calling *aspirational signatures*: signatures that describe the intent of the function and ignore implementation details.

For instance, `map` is documented as

```js
map :: Functor f => (a -> b) -> f a -> f b 
```

But note that as well as this:

```js
R.map(R.length, ['foo', 'blah']); //=> [3, 4]
```

We also get this:

```js
R.map(R.length, 'foo'); //=> [1, 1, 1] 
```

even though `String` is not a Functor.  This function has additional ways of operating.  They are not intentional, simply falling out of our implementation and the fact that we don't do significant type-checking.

For `unless` (and `where`, which is very similar) we would expect users to pass a transformation of `a -> a`, because otherwise the resulting function *would* have the unfortunate `a -> a | b`.  While the latter is possible, and users can take advantage of it, we don't promote it, and if our implementation were to change, we wouldn't feel too bad about breaking such user code.

PR #2112, opened 2017-03-04
Add through function

agentultra:
This function takes a list of getter functions and an object. It returns
the first non-nil value from the list of getters. It returns null if
nothing is found.


buzzdecafe:
i am :-1: on this for a couple of reasons.

* returning `null` is not idiomatic for ramda, but that is the lesser problem;
* the bigger problem is relying on the truthiness of the evaluation to branch: `if (r) { return r; }`

I think the second point is why this function feels to me like it is trying to do too much. On the one hand, it appears the intent is to evaluate a list of predicates against a value to see if any satisfy. On the other, it wants to overload those predicates to return something other than a Boolean value.

That overloading also a problem because it is impossible to ever return a "falsy" value (viz. `""`, `0`, or `false` itself.) Given that the included examples involved picking elements inside an object, one would be unable to get at properties that were bound to `false`, as they would get forced to `null` by the branching.

agentultra:
You're right. It was late at night. 😜

I think it needs to be more general really. I can take another stab at it later.

CrossEye:
I think the function that you're trying to create would be easy enough to patch up to get beyond the objections made by @buzzdecafe.  But it seems too specific to be included in Ramda.  A more general version with a slightly different API, might be written for the [Cookbook][cb] using `cond`, [like this][rr]:

```js
const through = compose(cond, map(fn => [compose(not, isNil, fn), fn]));
```

The API difference, dictated by `cond`, would be that you would have to call with your 'getters' (although they could be anything, really) first and get back a function which you could then call with your object:

```js
through(
  [prop('foo'), prop('bar'), prop('baz')]
)({bar: 1, baz: 2}); //=> 1
```

I might name it something related to `cond` and `notNil` if I could find a good combination, because that's what it seems closest to.


  [cb]: http://github.com/ramda/ramda/wiki/Cookbook
  [rr]: https://goo.gl/p4rEHN

agentultra:
I updated the patch according to the review @buzzdecafe 

Thanks everyone for the feedback. If this function is still not hitting the mark for Ramda I'd be happy to change to a recipe. It's a function I've used a fair bit to support scenarios where I need to map over mixed-structure data structures.

PR #2116, opened 2017-03-06
Make the invoker documentation a little more understandable to newbies like myself

krainboltgreene:


src/invoker.js
```diff
@@ -22,11 +23,28 @@ var toString = require('./toString');
...
  * @see R.construct
  * @example
+ *      // A function with no arguments
+ *      const asJson = invoker(0, "json")
+ *      // Just like calling .then((response) => response.json())
+ *      fetch("http://example.com/index.json").then(asJson)
+ *        
  *
- *      var sliceFrom = R.invoker(1, 'slice');
+ *      // A function with one arguments
```
buzzdecafe: s/arguments/argument/

src/invoker.js
```diff
@@ -22,11 +23,28 @@ var toString = require('./toString');
...
+ *      
+ *      // A function with two arguments
+ *      var sliceFrom6 = invoker(2, 'slice')(6);
  *      sliceFrom6(8, 'abcdefghijklm'); //=> 'gh'
+ *
+ *      // NOTE: You can't simply pass some of the arguments to the initial invoker function.
+ *      const firstCreditCardSection = invoker(2, "slice", 0, 4)
+ *      firstCreditCardSection("4242 4242 4242 4242") // => Function<...>
+ *
+ *      // Instead you must immediately invoke the invoker (better wording tbd)
```
buzzdecafe: How about:

"Since `invoker` returns a curried function, you may partially apply it to create the function you need."

buzzdecafe (dismissed):

CrossEye:
Although these changes (with the suggestions from @buzzdecafe) are a definite improvement, I wonder if we shouldn't go further and stop lying with "Turns a named method ..."  We do not start with a method/function, only with a name.  And the same `invoker`-derived function can easily be called on objects with different types, and with different functions associated with the given name:


```js
const indexOf = invoker(1, 'indexOf');
indexOf('bar', ['foo', 'bar', 'baz']); //=>1, using Array.prototype.indexOf
indexOf('x', 'Supercalifragilisticexpialidocious'); //=> 21, using String,prototype.indexOf
```

I'm still not sure what the right phrasing is, but even with the proposed changes, this is still somewhat misleading.

krainboltgreene:
Changes made @buzzdecafe, but also feel free to make changes.

buzzdecafe:
@CrossEye is right, the first sentence of the existing documentation is misleading at best. How about:

"Given a Number arity and a String method name, `invoker` returns a function that takes `arity` arguments and a context object, and invokes the method with the arguments in the context specified object."

Is that better? 🤔

krainboltgreene:
> Given an `arity` (Number) and a `name` (String) the `invoker()` function returns a curried function that takes `arity` arguments and a `context` object. It will "invoke" the `name`'d function (a method) on the `context` object.

buzzdecafe:
@krainboltgreene i think yours is definitely an improvement on mine

krainboltgreene:
@buzzdecafe Committed!

src/invoker.js
```diff
@@ -5,11 +5,10 @@ var toString = require('./toString');
 
 
 /**
- * Turns a named method with a specified arity into a function that can be
- * called directly supplied with arguments and a target object.
- *
- * The returned function is curried and accepts `arity + 1` parameters where
- * the final parameter is the target object.
+ * Given an `arity` (Number) and a `name` (String) the `invoker()` function
```
buzzdecafe: i think you can omit the parens after `invoker`. Otherwise, this looks good to me.

buzzdecafe:
There are some linting errors, but otherwise, this LGTM. 

:cow2:

customcommander:
Summary of changes:

1. Rebased against ramda's master branch
2. Fixed merge conflict
3. Fixed linter errors
4. Address @buzzdecafe latest comment
5. Used `const` instead of `var` in example (a pull request predating this one did something similar; so just being consistent here)

@ramda/ramda-docs Good to go?

CrossEye (approved):

PR #2117, opened 2017-03-07
Add debounce function to Ramda

Jkaveri:
Create a debounced that delays a pass in function util wait milliseconds. It is a helpful util when we need to create an event handler which handles a big rate of executions.

See: https://css-tricks.com/debouncing-throttling-explained-examples/

### DIFF
--- a/index.js
+++ b/index.js
@@ -40,6 +40,7 @@ module.exports = {
   countBy: require('./src/countBy'),
   curry: require('./src/curry'),
   curryN: require('./src/curryN'),
+  debounce: require('./src/debounce'),
   dec: require('./src/dec'),
   defaultTo: require('./src/defaultTo'),
   descend: require('./src/descend'),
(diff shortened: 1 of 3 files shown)


buzzdecafe:
while i can see the usefulness of this procedure, i don't think it is a good fit for ramda. `debounce` is completely about side-effects. It does not even return a value.

Jkaveri:
Thank you @buzzdecafe. I agreed it doesn't fit with ramda, maybe. However, It should useful in my project. I submit this request because I just want to contribute to ramda because I love it :).

Jkaveri:
If It looks like a useless to you guys :). we can close this request :).
Thank you for your comment :)

davidchambers:
It's certainly not useless, @Jkaveri. Everything in its right place, as the Radiohead song goes. Ramda just isn't the right place for this function.

I appreciate your enthusiasm for Ramda. :)

buzzdecafe:
there are some contrib libraries cropping for extending ramda in different directions (e.g. i think there are some on the [list of related projects](https://github.com/ramda/ramda/wiki); not a complete list!)

Thanks for the kind words

PR #2119, opened 2017-03-07
Fixed `xtake` where n < 0

davidnormo:
I noticed that the behaviour of `take` was inconsistent when used normally vs as a transducer. Specifically, `take(-1)` would return a transducer that would only process 1 element. I was expecting it to process every element, effectively doing nothing.

I'm happy for this to be closed if this is the intended behaviour :)

### DIFF
--- a/src/internal/_xtake.js
+++ b/src/internal/_xtake.js
@@ -13,7 +13,7 @@ module.exports = (function() {
   XTake.prototype['@@transducer/step'] = function(result, input) {
     this.i += 1;
     var ret = this.n === 0 ? result : this.xf['@@transducer/step'](result, input);
-    return this.i >= this.n ? _reduced(ret) : ret;
+    return this.n >= 0 && this.i >= this.n ? _reduced(ret) : ret;
   };
 
   return _curry2(function _xtake(n, xf) { return new XTake(n, xf); });
--- a/test/take.js
+++ b/test/take.js
@@ -46,4 +46,10 @@ describe('take', function() {
     sinon.assert.calledTwice(spy);
   });
 
+  it('transducer called for every member of list if `n` is < 0', function() {
+    var spy = sinon.spy();
+    R.into([], R.compose(R.map(spy), R.take(-1)), [1, 2, 3]);
+    sinon.assert.calledThrice(spy);
+  });
+
 });


CrossEye:
I definitely don't think that this behavior is intended.

I don't recall why we chose this:

```js
    (n, xs) => slice(0, n < 0 ? Infinity : n, xs);
```
rather than this:

```js
    (n, xs) => slice(0, n < 0 ? 0 : n, xs);
```

for our implementation on lists, though.

So, while I'm :+1: on this change, I think that sometimes soon we should also talk about whether our handling of negative numbers makes sense, for `take/drop(/Last)(/While)`

davidchambers:
> `take/drop(/Last)(/While)`

This is the notation you seek, @CrossEye:

```console
$ echo {take,drop}{,Last,While}
take takeLast takeWhile drop dropLast dropWhile
```

CrossEye:
Thank you.  I `{stand,sit}{Corrected,Uninterested}`.

davidnormo:
@CrossEye IMO, I like how the negative numbers are handled, it makes `take` more powerful. What are the pros/cons you can see?

Regarding the current `Infinity` implementation, I poked around the git history and found #1159, seems like it has something to do with how some functions could also handle strings as they use `slice` under the hood.

CrossEye:
@davidnomo:
> I like how the negative numbers are handled, it makes take more powerful. What are the pros/cons you can see?

Just that this looks odd:

```js
take(3, [1, 2, 3, 4, 5]); //=> [1, 2, 3]
take(2, [1, 2, 3, 4, 5]); //=> [1, 2]
take(1, [1, 2, 3, 4, 5]); //=> [1]
take(0, [1, 2, 3, 4, 5]); //=> []
take(-1, [1, 2, 3, 4, 5]); //=> [1, 2, 3, 4, 5]
```

It simply makes little sense to me.  Returning `[]` seems more logical.

There was a suggestion at one point that `take(-3, [1, 2, 3, 4, 5]; //=> [3, 4, 5]`, but we chose instead to add `takeLast`.  I won't rehash the discussion for that.

davidnormo:
For clarities sake, the following is the current behaviour:

```
// as normal
take(2, [1,2,3]) // => [1,2]
take(1, [1,2,3]) // => [1]
take(0, [1,2,3]) // => []
take(-1, [1,2,3]) // => [1,2,3] - ?
take(-2, [1,2,3]) // => [1,2,3] - ??

// as transducer
into([], take(2), [1,2,3]) // => [1,2]
into([], take(1), [1,2,3]) // => [1]
into([], take(0), [1,2,3]) // => [0]
into([], take(-1), [1,2,3]) // => [1] - ??
into([], take(-2), [1,2,3]) // => [1] - ???
```

After some though, I think I disagree with my previous point. It seems simpler to say that `n` is the number of elements returned up to the length of the list. Then if we want to take everything in a list we can either use `Infinity` or `list.length`.

If you agree, then I propose we change `take` to match the statement above, both for its normal use and as a transducer. Which would result in this:

```
// function interface
take(2, [1,2,3]) // => [1,2]
take(1, [1,2,3]) // => [1]
take(0, [1,2,3]) // => []
take(-1, [1,2,3]) // => []
take(-2, [1,2,3]) // => []

// transducer interface
into([], take(2), [1,2,3]) // => [1,2]
into([], take(1), [1,2,3]) // => [1]
into([], take(0), [1,2,3]) // => []
into([], take(-1), [1,2,3]) // => []
into([], take(-2), [1,2,3]) // => []
```

davidnormo:
If not, no worries, the breaking change might not be worth it at this point. This PR just fixes the transducer version to match the normal function behaviour.

CrossEye:
Sorry, I lost the thread here for a bit.

I agree that this would be a useful change.  Consistency is mostly good, even if Emerson has some objections.

kedashoe:
> So, while I'm :+1: on this change, I think that sometimes soon we should also talk about whether our handling of negative numbers makes sense

agreed on both points

buzzdecafe:
count me :-1: on negative indices generally

Issue #1874, opened 2016-08-21
Some more string functions


ccorcos:
I know we generally do not want to bloat Ramda with functions that are easily created, but I find myself creating these functions when manipulating strings all the time. They're so simple and help with point free programming so I still think we should consider adding them.

```js
const prefix = R.add
const suffix = R.flip(R.add)
const startsWith = R.curry((q, str) => str.startsWith(q))
const endsWith = R.curry((q, str) => str.endsWith(q))
const stripl = R.curry((q, str) => startsWith(q, str) ? str.slice(q.length) : str)
const stripr = R.curry((q, str) => endsWith(q, str) ? str.slice(0, str.length - q.length) : str)
const strip = R.curry((q, str) => stripl(q, stripr(q, str)))
```

What do you think?

olsonpm:
I'm pretty sure they're trying to minimize aliases (R.add -> prefix and its fipped counterpart).

I'm on board with starts/endsWith, except keep in mind your implementation is not ie friendly

Not on board with your implementations of stripl/r simply because it's not in lodash.  My only reasoning for adhering to lodash in this case is that it's likely the most popular utility lib next to underscore.  If lodash hasn't found a need to implement those functions, then, considering ramda's smaller string category, I would assume it shouldn't be implemented here.

fix-fix:
You can use [underscore.string.fp](https://www.npmjs.com/package/underscore.string.fp) or other string manipulation library.

olsonpm:
Nice library - thanks for mentioning it.

Notice their aliases conflict with OP's strip implementation

> strip     = trim
> lstrip    = ltrim
> rstrip    = rtrim

ccorcos:
Hmm. I just find it so awkward that I have to keep writing these functions. Seems like JS should really just have a standard lib...

davidchambers:
I'd like to leave this open for a few days to give others a chance to comment.

ccorcos:
That works for me :)

CrossEye:
>  If lodash hasn't found a need to implement those functions, then, considering ramda's smaller string category, I would assume it shouldn't be included here.

I wouldn't buy that at all.

You might notice, in fact, that lodash/fp is essentially an API clone of Ramda, with an entirely different implementation (and, of course, assorted incompatibilities.)  

[...]

Issue #1842, opened 2016-07-29
Add the method "startsWith" from Strings


zatziky:
Hello, can you please add `R.startsWith()` just like it is in lodash - https://lodash.com/docs#startsWith ?

CrossEye:
If you want to create a PR, I'm sure we'd consider it.  We would not want the lodash API.  Ramda avoids optional arguments.

But I'm not sure we'd accept it, as this seems so easy to create with existing tools:

```js
    const startsWith = R.invoker(1, 'startsWith');
    startsWith('a', 'abc'); //=> true
    startsWith('b', 'abc'); //=> false
```

asaf-romano:
FWIW, my personal startsWith supports both strings and arrays:
``const startsWith = curry((prefix, xs) => equals(take(xs.length, xs), xs));``

asaf-romano:
What's nice about that is that it plays really nice with value equality semantics..

asaf-romano:
With those semantics, I'd happily accept both startsWith and endsWith.

Tapsi:
@asaf-romano I tried your snippet and it behaves different than expected. Am I correct that `const startsWith = curry((prefix, xs) => equals(take(prefix.length, xs), prefix));` is the correct implementation? I think prefix.length is correct because you want to check prefix against xs, right?

asaf-romano:
:)

asaf-romano:
endsWith would be pretty much the same, but with takeLast.

zatziky:
I let the PR to someone else. I just would like to see a list of functions that is, say, longer. As I see functional methods as operators and e.g. in rx you have got lots of them. Then you are just composing them. And that's fine. ;-) Goodjob you do otherwise!

jethrolarson:
This wouldn't just be `invoker(1, 'startsWith')` as the browser support of the native is poor: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith

CrossEye:
You're right.  Those pesky browsers.

I've used this one a lot in Node.  I'd be interested in something like @asaf-romano's formulation, then.  Anyone interested in creating a PR?

mattgrande:
The only issue I have with @asaf-romano's answer is it has the parameters in the opposite order of lodash.

I'll try whipping up a PR for this tonight (about 8-10 hours from posting this), but if someone else gets to it first, feel free.

jethrolarson:
[...]

PR #2120, opened 2017-03-09
Add startsWith and endsWith functions

mattgrande:
Add `R.startsWith` and `R.endsWith` as suggested in #1842 and (to a lesser extent) #1874.

### DIFF
--- a/test/endsWith.js
+++ b/test/endsWith.js
@@ -1,0 +1,20 @@
+var R = require('..');
+var eq = require('./shared/eq');
+
+describe('startsWith', function() {
+  it('should return true when a string ends with the provided value', function() {
+    eq(R.endsWith('c', 'abc'), true);
+  });
+
+  it('should return false when a string does not end with the provided value', function() {
+    eq(R.endsWith('b', 'abc'), false);
+  });
+
+  it('should return true when an array ends with the provided value', function() {
+    eq(R.endsWith(['c'], ['a', 'b', 'c']), true);
+  });
+
+  it('should return false when an array does not end with the provided value', function() {
+    eq(R.endsWith(['b'], ['a', 'b', 'c']), false);
+  });
+});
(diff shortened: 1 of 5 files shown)


agentultra:
Looks straight forward to me.

test/endsWith.js
```diff
@@ -0,0 +1,20 @@
...
+  });
+
+  it('should return true when an array ends with the provided value', function() {
+    eq(R.endsWith(['c'], ['a', 'b', 'c']), true);
+  });
+
+  it('should return false when an array does not end with the provided value', function() {
+    eq(R.endsWith(['b'], ['a', 'b', 'c']), false);
+  });
+});
```
CrossEye: I would prefer it if at least one test demonstrated that the input does not have to be unary:  `endsWith(
'ology', 'physics'); //=> false`, or `endsWith([3, 0, 9], [8, 6, 7, 5, 3, 0, 0])' //=> true`,  This could be an additional test, or a change to one that's there.

test/startsWith.js
```diff
@@ -0,0 +1,20 @@
...
+  });
+
+  it('should return true when an array starts with the provided value', function() {
+    eq(R.startsWith(['a'], ['a', 'b', 'c']), true);
+  });
+
+  it('should return false when an array does not start with the provided value', function() {
+    eq(R.startsWith(['b'], ['a', 'b', 'c']), false);
+  });
+});
```
CrossEye: Ditto.

CrossEye:
Made a minor comment in-line requesting a slight change to the tests, but this looks good.

mattgrande:
Good call, @CrossEye. I've extended the number of tests.

CrossEye:
This LGTM.  Anyone else want to chime in?

asaf-romano:
🌷

PR #2127, opened 2017-03-13
binary, curried composition

buzzdecafe:
anyone interested in a stricter `compose` function?

### DIFF
--- a/src/o.js
+++ b/src/o.js
@@ -1,0 +1,29 @@
+var _curry3 = require('./internal/_curry3');
+
+
+/**
+ * `o` is a curried composition function that returns a unary function.
+ * Like `compose`, `o` performs right-to-left function composition. Unlike `compose`,
+ * the rightmost function passed to `o` will be invoked with only one argument.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig (b -> c) -> (a -> b) -> a -> c
+ * @param {Function} f
+ * @param {Function} g
+ * @return {Function}
+ * @see R.compose, R.pipe
+ * @example
+ *
+ *      var classyGreeting = name => "The name's " + name.last + ", " + name.first + " " + lastName
+ *      var yellGreeting = R.o(R.toUpper, classyGreeting);
+ *      yellGreeting({first: 'James', last: 'Bond'}); //=> "THE NAME'S BOND, JAMES BOND"
+ *
+ *      R.o(R.multiply(10), R.add(10))(-4) //=> 60
+ *
+ * @symb R.o(f, g, x) = f(g(x))
+ */
+module.exports = _curry3(function o(f, g, x) {
+  return f(g(x));
+});
--- a/index.js
+++ b/index.js
@@ -138,6 +138,7 @@ module.exports = {
   not: require('./src/not'),
   nth: require('./src/nth'),
   nthArg: require('./src/nthArg'),
+  o: require('./src/o'),
   objOf: require('./src/objOf'),
   of: require('./src/of'),
   omit: require('./src/omit'),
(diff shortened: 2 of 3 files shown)


src/o.js
```diff
@@ -0,0 +1,31 @@
...
+ *
+ *      var classyGreeting = name => "The name's " + name.last + ", " + name.first + " " + lastName
+ *      var yellGreeting = R.o(R.toUpper, classyGreeting);
+ *      yellGreeting({first: 'James', last: 'Bond'}); //=> "THE NAME'S BOND, JAMES BOND"
+ *
+ *      R.o(Math.abs, R.multiply(2))(-4) //=> 8
+ *
+ * @symb R.o(f, g)(x) = f(g(x))
+ */
+module.exports = _curry2(function o(f, g) {
```
davidchambers: Why not `_curry3(function o(f, g, x) { ... })`?

buzzdecafe: no good reason

src/o.js
```diff
@@ -0,0 +1,31 @@
...
+ * @param {Function} g
+ * @return {Function}
+ * @see R.compose, R.pipe
+ * @example
+ *
+ *      var classyGreeting = name => "The name's " + name.last + ", " + name.first + " " + lastName
+ *      var yellGreeting = R.o(R.toUpper, classyGreeting);
+ *      yellGreeting({first: 'James', last: 'Bond'}); //=> "THE NAME'S BOND, JAMES BOND"
+ *
+ *      R.o(Math.abs, R.multiply(2))(-4) //=> 8
```
davidchambers: You should choose an example which demonstrates the order in which the functions are applied.

CrossEye:
:herb:, except that, even though I understand the reason for `'o'`, I'm not a real fan of Ramda using such one-character names.  Even `R.T` and `R.F` bother me.

davidchambers:
The right name for this function is `compose`. ;)

src/o.js
```diff
@@ -0,0 +1,29 @@
...
+ * @see R.compose, R.pipe
+ * @example
+ *
+ *      var classyGreeting = name => "The name's " + name.last + ", " + name.first + " " + lastName
+ *      var yellGreeting = R.o(R.toUpper, classyGreeting);
+ *      yellGreeting({first: 'James', last: 'Bond'}); //=> "THE NAME'S BOND, JAMES BOND"
+ *
+ *      R.o(R.multiply(10), R.add(10))(-4) //=> 60
+ *
+ * @symb R.o(f, g)(x) = f(g(x))
```
davidchambers: Should we use `R.o(f, g, x)` here?

CrossEye:
Or `B`.  

Or `bluebird`.  

Perhaps we could call it `supercalifragilisticexpialidocious`.

src/o.js
buzzdecafe: yes, well spotted

buzzdecafe:
> The right name for this function is `compose`. ;)

:fire:

ok i think this one is good to go; please give it a final look & merge.

CrossEye:
@buzzdecafe:
> ok i think this one is good to go; please give it a final look & merge.

You really want to go with the name `o`?

buzzdecafe:
> You really want to go with the name o?

i agree with @davidchambers that i'd rather see this called `compose` and the other get renamed to `comp` [a la clojure](https://clojuredocs.org/clojure.core/comp), which it resembles; or something else more descriptive, e.g. `variadicCompose`, `composeVar` etc.

Given that the "right" name is squatted on atm, naming the function for the composition operator seems good to me. It also has the advantage of concision for what ought to be one of the primary functions of the library.

CrossEye:
Merging under protest.  :smile:

PR #2130, opened 2017-03-16
Add memoizeWith

### LINKED ISSUE #2125: Mention toString implementation of R.memoize
Hi,

Looking at the source code, I've noticed that `R.memoize` uses `toString` to memoize the result.
See related issue - https://github.com/ramda/ramda/issues/1384

Would it be an idea to mention this in a documentation? I assume that programmers are using this function in a hope it will increase performance, but in a lots of scenarios, it can actually decrease perf. because `toString` is rather expensive operation. Since people don't check the source code every time, it would be useful to have it stated in a docs I think

I'll be happy to create a PR if the issue is accepted.

svozza:
Closes #2125. This isn't as straightforward as being able to do `R.memoizeWith(R.prop('a'))` because we have to support multiple arity functions. This means that the customising function is applied to `arguments` so the correct way to use the function is `R.memoizeWith(R.compose(R.prop('a'), R.head))`. I think the examples in the documentation  I've provided aren't good enough and I was hoping people had some better suggestions.

### DIFF
--- a/src/memoizeWith.js
+++ b/src/memoizeWith.js
@@ -1,0 +1,43 @@
+var _arity = require('./internal/_arity');
+var _curry2 = require('./internal/_curry2');
+var _has = require('./internal/_has');
+
+
+/**
+ * A customisable version of R.memoize. memoizeWith takes an additional
+ * function that will be applied to a given argument set and used to create
+ * the cache key under which the results of the function to be memoized will
+ * be stored. Care must be taken when implementing key generation to avoid
+ * clashes that may overwrite previous entries erroneously.
+ *
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig (*... -> String) -> (*... -> a) -> (*... -> a)
+ * @param {Function} fn The function to generate the cache key.
+ * @param {Function} fn The function to memoize.
+ * @return {Function} Memoized version of `fn`.
+ * @see R.memoize
+ * @example
+ *
+ *      let count = 0;
+ *      const factorial = R.memoizeWith(R.identity, n => {
+ *        count += 1;
+ *        return R.product(R.range(1, n + 1));
+ *      });
+ *      factorial(5); //=> 120
+ *      factorial(5); //=> 120
+ *      factorial(5); //=> 120
+ *      count; //=> 1
+ */
+module.exports = _curry2(function memoizeWith(mFn, fn) {
+  var cache = {};
+  return _arity(fn.length, function() {
+    var key = mFn.apply(this, arguments);
+    if (!_has(key, cache)) {
+      cache[key] = fn.apply(this, arguments);
+    }
+    return cache[key];
+  });
+});
--- a/index.js
+++ b/index.js
@@ -124,6 +124,7 @@ module.exports = {
   mean: require('./src/mean'),
   median: require('./src/median'),
   memoize: require('./src/memoize'),
+  memoizeWith: require('./src/memoizeWith'),
   merge: require('./src/merge'),
   mergeAll: require('./src/mergeAll'),
   mergeWith: require('./src/mergeWith'),
--- a/src/memoize.js
+++ b/src/memoize.js
@@ -1,6 +1,4 @@
-var _arity = require('./internal/_arity');
-var _curry1 = require('./internal/_curry1');
-var _has = require('./internal/_has');
+var memoizeWith = require('./memoizeWith');
 var toString = require('./toString');
 
 
@@ -20,8 +18,8 @@ var toString = require('./toString');
  * @return {Function} Memoized version of `fn`.
  * @example
  *
- *      var count = 0;
- *      var factorial = R.memoize(n => {
+ *      let count = 0;
+ *      const factorial = R.memoize(n => {
  *        count += 1;
  *        return R.product(R.range(1, n + 1));
  *      });
@@ -30,13 +28,4 @@ var toString = require('./toString');
  *      factorial(5); //=> 120
  *      count; //=> 1
  */
-module.exports = _curry1(function memoize(fn) {
-  var cache = {};
-  return _arity(fn.length, function() {
-    var key = toString(arguments);
-    if (!_has(key, cache)) {
-      cache[key] = fn.apply(this, arguments);
-    }
-    return cache[key];
-  });
-});
+module.exports = memoizeWith(toString);
--- a/test/memoize.js
+++ b/test/memoize.js
@@ -3,28 +3,6 @@ var eq = require('./shared/eq');
 
 
 describe('memoize', function() {
-  it('calculates the value for a given input only once', function() {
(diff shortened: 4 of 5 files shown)


src/memoizeWith.js
```diff
@@ -0,0 +1,42 @@
...
+ * which the results of the function to be memoized will be stored. Care must
+ * be taken when implementing key generation to avoid clashes that may
+ * overwrite previous entries erroneously.
+ *
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig (*... -> a) -> (*... -> a)
+ * @param {Function} fn The function to memoize.
```
kedashoe: Another @param here for mFn?

src/memoizeWith.js
```diff
@@ -0,0 +1,42 @@
...
+ *      });
+ *      factorial(5); //=> 120
+ *      factorial(5); //=> 120
+ *      factorial(5); //=> 120
+ *      count; //=> 1
+ */
+module.exports = _curry2(function memoizeWith(mFn, fn) {
+  var cache = {};
+  return _arity(fn.length, function() {
+    var key = mFn(arguments);
```
kedashoe: Should this be like the line below? ie

`var key = mFn.apply(this, arguments);`

?

src/memoizeWith.js
svozza: :zap: Update the type sig too.

src/memoizeWith.js
svozza: Are we supporting `this`?

kedashoe: I don't think so, I would rather not as well, I just copied what you had below. The point was more that using `apply` makes the key function more friendly, eg would simply be `id` for your example function, and you can use `prop` rather than `compose(prop, head)`.

svozza: Ah very good. I like that.

svozza: :zap:

CrossEye: No more `this`!!  :smile:

CrossEye:
I don't believe you can capture enough detail in the sort of short docs we try to write.  But the unit tests cover the gaps quite well.

This looks very good to me.

:herb:

src/memoizeWith.js
```diff
@@ -0,0 +1,43 @@
+var _arity = require('./internal/_arity');
+var _curry2 = require('./internal/_curry2');
+var _has = require('./internal/_has');
+
+
+/**
+ * A customisable version of R.memoize: takes an additional function that will
```
buzzdecafe: "A customisable version of R.memoize _that_ takes ..."

buzzdecafe: "A customisable version of `R.memoize`.  `memoizeWith` takes ..."

buzzdecafe:
me too
:cow2:

src/memoizeWith.js
svozza: I purposely chose the colon so I wouldn't use the word 'that' twice in such close proximity.

svozza: Yeah, that works. Not near a computer so can't change it right now.

svozza: :zap: I also changed the example code to use `let` and `const`.

ln613:
With memoizeWith, the key is always a string. Could you shed some light on how to provide a key if I want to cache the result when none of the references of all args is changed (prev arg === current arg)? Thanks!

buzzdecafe:
If you can get different results for the same arguments, then that method is not suitable for memoization.

ln613:
thanks for the reply. it is a pure function, it will get the same result for the same arguments. but the purpose here is to avoid re-calculation by comparing the references of the arguments, If an arg's reference is not changed, but the content is changed, theoretically we should recalculate, but we won't re-calculate to gain performance by assuming all arguments are immutable , just like what react or reselect is doing. So for that purpose, is memoizeWith suitbale? if not, is there any other functions in ramda that can achieve this? Thanks

buzzdecafe:
if i understand what you are saying, you just want to test reference equality? If so then

`const memo = memoizeWith(identical)`

ought to do the trick

ln613:
do you mean identity? identical will return a boolean, there is no point to have a key with only 2 values true and false. identity will also not work, because it only takes 1 argument. but the way the key function is called inside memoizeWith is

`var key = mFn.apply(this, arguments);`

arguments is the list of arguments to the underlying function. so if you use identity as the key function, all arguments other than the first one will be ignored. also, the generated key will be used as an object key, so if you use identity on an object argument, it will return an object as the key, which will be converted to '[object Object]' when it's used as an object key. so using identity as a key function only works on simple types like number or string.

to use memoizeWith, my understanding is, you need to generate a unique string key based on all the arguments provided to the underlying function. but what i couldn't figure out is, how to generate a unique string based on the references of those arguments

buzzdecafe:
yeah apologies i should've looked at the docs myself.

ln613:
np. I'm just trying to implement what [reselect](https://github.com/reactjs/reselect/blob/master/src/index.js) is doing.

the main difference is that ramda's memoizeWith function caches a list of results based on a key, whereas reselect's implementation caches only the previous result, so it doesn't need a key, it only remembers the last argument list, then it can compare between the last argument list and the current argument list based on a specified equality function (in this case, reference equality).

it would be nice if ramda can provide a different version of memoizeWith, which caches only the last result, but is able to compare the argument list based on a user provided equality function

CrossEye:
@ln613

Sorry, I've been off for a bit, having my head examined<sup>*</sup> and I missed this.

While I wouldn't object to seeing a PR for this, I'd have to think about it a lot before I decide that I was interested in including it.  In any of its forms, `memoize` has always been an odd fit for Ramda.  Because it's only about avoiding re-performing complex calculations, it's technically not about side-effects, but it really feels as though it is.  (I have the same feeling about `tap` -- it's useful, but an odd fit for the library.)  For this reason, I have a difficult time wanting to include more `memoize`-related functions here.

If you feel strongly about it, don't hesitate to open a PR; it will certainly get a fair hearing.

----------
<sup>*</sup> So, it was just my right eye and one side of my forehead, covered in shingles bumps -- that's still part of my head!

ln613:
Thanks for the reply @CrossEye. I totally agree, not a good fit for Ramda. I implemented it myself for my project.

Although knowing it's a FP library, when I use Ramda more and more, I tend to use it as a one-stop shop, trying to find a solution for everything with Ramda. For example, defining a 'tap(console.log)' function would be the first thing when I create a new project.

Thanks for providing Ramda, making javascript programming more functional.

CrossEye:
Thank you for the kind words.

In any project I do, I have my own utility library, these days generally layered atop Ramda, for all those functions that I need but which don't exist in Ramda.  If I reuse them in many projects, and if they otherwise seem to fit, I might propose them for Ramda.  But often they end up similar but slightly different across projects.

Issue #2128, opened 2017-03-15
Lazy but finite reduce for recursive data structures and an unbounded number of reducers  (see comments)


sadasant:
Good day everyone!

Over the past weeks we've been working with arbitrarily nested objects and arrays (mixed altogether), up until we've spotted the necessity of a recursive reduce function that could be used to transform, but at least to obtain, certain if not every property of this nested structure, which could also be stopped at any given point if a condition is sufficient. Essentially, we've been looking at what in haskell would be a `fold takeWhile`, but for deep/recursive objects/arrays.

After several iterations, the function valid for these use cases that we haven't been able to simplify anymore is the following:

```js
// Recursive fold takeWhile (like in haskell)
// fn is the iterator
// obj is the iteratee
// r is the accumulator
const foldWhile = (fn, obj, r = []) => {
  let innerFold = o => !(typeof o === 'object') ? o
    : Object.keys(o).every(k => fn(r, o[k], k) && innerFold(o[k]))
  innerFold(obj)
  return r
}
```

_**Note:** this is implemented with lodash in our code, but since it doesn't need lodash, I'm using generic es6 to show what it does._  
_**Note2:** Besides using functions from the Ramda ecosystem, this would benefit from a trampoline._

The main point of the function is combining the use of `every` recursively to stop at any point of the (_lazy_) evaluation of the nested object/array, with the use of an accumulator, which is handled to the user.

To see use cases and tests, see this gist: https://gist.github.com/sadasant/de106b879c61f21797dffdaa577f7868

We have been looking for opensource alternatives, but we haven't been successful.

I know that there have been several issues in this repo about lazy evaluation and nested data structures, however there doesn't seem to be an agreement on how to combine both the nested structures and the stopping of the fold at any given point. I know that Ramda has reduceWhile and until, but they don't work over nested structures, right?.

Having that said,  
I wonder if this is interesting to anybody. We would definitely prefer using opensource instead of maintaining our own code.

Thank you for the time.

sadasant:
( I also asked this on lodash https://github.com/lodash/lodash/issues/3058 )

sadasant:
Here's an alternative that works for infinitely many reducers (pardon my lodash):

```
// Recursive reduce nested objects of any kind that
// stops at any point if a falsy value is returned.
export const flowReduce = (...fns) => (obj, acc = []) => {
    if (typeof obj !== 'object') return acc
    let _acc = acc
    let type = typeof acc
    _.every(k => {
[...]

PR #2133, opened 2017-03-21
depthReduce, a stoppable depth-first reduce for deep iterables

sadasant:
Returns a single item by running a depth-first iteration through a given multi-dimensional iterable (list, object, nested objects and arrays or cyclic object), successively calling the iterator function and passing it an accumulator value, the current value from the iterable, the key where the current value is stored in the iterable, a list with the keys of the parent iterables of the current position, and then passing the result to the next call.

The iterator function receives four values: *(acc, value, key, parentsKeys)*.  It's expected that the iterator returns it's results wrapped in an array, of which only the first position is considered. If the iterator returns an empty array, depthReduce immediately stops, returning the last non-empty result.

See: https://github.com/ramda/ramda/issues/2128#issuecomment-287766254

Please review 🙏🙇

### DIFF
--- a/src/depthReduce.js
+++ b/src/depthReduce.js
@@ -1,0 +1,85 @@
+var _curry3 = require('./internal/_curry3');
+var equals = require('./equals');
+var keys = require('./keys');
+
+
+/**
+ * Returns a single item by running a depth-first iteration through a given multi-dimensional
+ * iterable (list, object, nested objects and arrays or cyclic object), successively calling
+ * the iterator function and passing it an accumulator value, the current
+ * value from the iterable, the key where the current value is stored in the iterable,
+ * a list with the keys of the parent iterables of the current position,
+ * and then passing the result to the next call.
+ *
+ * The iterator function receives four values: *(acc, value, key, parentsKeys)*.
+ * It's expected that the iterator returns it's results wrapped in an array, of which only
+ * the first position is considered. If the iterator returns an empty array, depthReduce
+ * immediately stops, returning the last non-empty result.
+ *
+ * Note: `R.depthReduce` does not skip deleted or unassigned indices (sparse
+ * arrays), unlike the native `Array.prototype.reduce` method. For more details
+ * on this behavior, see:
+ * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce#Description
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig (a,v,k,parentKeys -> [a]) -> a -> [{...}] -> a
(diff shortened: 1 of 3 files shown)


src/depthReduce.js
```diff
@@ -0,0 +1,85 @@
...
+    }
+    var key = env.keys.shift();
+    var value = env.target[key];
+    var result = reducer(accepted, value, key, env.path);
+    if (equals(result, empty)) {
+      return accepted;
+    }
+    accepted = (result && result.length) ? result[0] : result;
+    if (typeof value === 'object') {
+      levels.push({
```
sadasant: 👀 _sublevels_ 🤘 https://open.spotify.com/track/0VwQ3VnCAVT9dqTQa4FYiK 😂

sadasant:
@davidchambers thoughts on this?
[](https://media.giphy.com/media/3oz8xWrNol65TMrc1a/source.gif)

Also, anything else you'd like me to tackle? I see about 175~ issues, I could take some.

src/depthReduce.js
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * Note: `R.depthReduce` does not skip deleted or unassigned indices (sparse
+ * arrays), unlike the native `Array.prototype.reduce` method. For more details
+ * on this behavior, see:
+ * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce#Description
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig (a,v,k,parentKeys -> [a]) -> a -> [{...}] -> a
```
sadasant: I'm guessing this can be better as `(a,b,c,d -> [a]) -> a -> b -> a`

timini:
??

CrossEye:
@timini: Good `??`  I don't know how this got dropped.  I don't remember ever even seeing it.

@sadasant: I'm very sorry.  I  will try to take a look tomorrow.

CrossEye:
This is a fascinating function that -- like too many old issues/prs was completely dropped.

@sadasant : Do you want to make a new pitch for this?

sadasant:
@CrossEye oh not at all! Sorry for the trouble. Thank you for keeping track of things!

CrossEye:
`Well, the problem is that I wasn't keeping track of things, and neither was anyone else.  @customcommander has joined the team and is helping us get back on track.

I'll close this one now.  If you want to revive it, either comment here or raise a new PR.  Thank you.

And again, sorry.

sadasant:
@CrossEye that sounds like good news! Congrats on onboarding @customcommander ! Ramda is awesome ❤️

Issue #2138, opened 2017-03-22
Use Array (Pair a b) in unfold


sadasant:
Quoting @davidchambers:
> Since we don't have the option of using Maybe (Pair a b), we'd need to use Array (Pair a b) and assume that the array's length will either be zero or one ([] or [['foo', 42]], say).

Perhaps in this form?

```
module.exports = _curry2(function unfold(fn, seed) {
  var pair = fn(seed);
  var result = [];
  while (pair && pair[0] && pair[0].length === 2) {
    result[result.length] = pair[0][0];
    pair = fn(pair[0][1]);
  }
  return result;
});
```

davidchambers:
Rather than `pair && pair[0] && pair[0].length === 2` the check should be `pair.length === 1`.

sadasant:
What if the user returns `['oops']` ?

davidchambers:
> What if the user returns `['oops']` ?

We likely fail with an unhelpful type error. That's the Ramda way. *Suppressing* the error would be far worse, though.

Also, `pair` is a misleading name. I suggest `maybePair` instead.

sadasant:
Aye captain!

CrossEye:
After weeks away, I'm back.  And this is fascinating.  I finally understand the original posts on this.  And I really like updating `unfold`.

PR #2139, opened 2017-03-22
Unfold with Maybe Pairs

sadasant:
Or more precisely, with Array (Pairs a b)

Fixes: #2138

Review wanted :)

### DIFF
--- a/src/unfold.js
+++ b/src/unfold.js
@@ -3,9 +3,9 @@ var _curry2 = require('./internal/_curry2');
 
 /**
  * Builds a list from a seed value. Accepts an iterator function, which returns
- * either false to stop iteration or an array of length 2 containing the value
- * to add to the resulting list and the seed to be used in the next call to the
- * iterator function.
+ * either an empty array to stop iteration, or a singleton array containing an
+ * array of length 2, with the value that's added to the resulting list and the
+ * seed to be used in the next call of the iterator function.
  *
  * The iterator function receives one argument: *(seed)*.
  *
@@ -13,25 +13,27 @@ var _curry2 = require('./internal/_curry2');
  * @memberOf R
  * @since v0.10.0
  * @category List
- * @sig (a -> [b]) -> * -> [b]
+ * @sig (b -> [(a, b)]) -> b -> [a]
  * @param {Function} fn The iterator function. receives one argument, `seed`, and returns
- *        either false to quit iteration or an array of length two to proceed. The element
- *        at index 0 of this array will be added to the resulting array, and the element
- *        at index 1 will be passed to the next call to `fn`.
+ *        either an empty array to quit iteration or an array with an array of length two
+ *        to proceed. The element at index 0 of the array of length two will be added
+ *        to the resulting array, and the element at index 1 will be passed
+ *        to the next call to `fn`.
  * @param {*} seed The seed value.
  * @return {Array} The final list.
  * @example
  *
- *      var f = n => n > 50 ? false : [-n, n + 10];
+ *      var f = n => n > 50 ? [] : [[-n, n + 10]];
  *      R.unfold(f, 10); //=> [-10, -20, -30, -40, -50]
- * @symb R.unfold(f, x) = [f(x)[0], f(f(x)[1])[0], f(f(f(x)[1])[1])[0], ...]
+ * @symb R.unfold(f, x) = [f(x)[0][0], f(f(x)[0][1])[0][0], f(f(f(x)[0][1])[0][1])[0][0], ...]
  */
 module.exports = _curry2(function unfold(fn, seed) {
-  var pair = fn(seed);
+  var maybePair = fn(seed);
   var result = [];
-  while (pair && pair.length) {
+  while (maybePair.length === 1) {
+    var pair = maybePair[0];
     result[result.length] = pair[0];
-    pair = fn(pair[1]);
+    maybePair = fn(pair[1]);
   }
   return result;
 });
--- a/test/unfold.js
+++ b/test/unfold.js
@@ -4,7 +4,7 @@ var eq = require('./shared/eq');
 
 describe('unfold', function() {
   it('unfolds simple functions with a starting point to create a list', function() {
-    eq(R.unfold(function(n) {if (n > 0) {return [n, n - 1];}}, 10), [10, 9, 8, 7, 6, 5, 4, 3, 2, 1]);
+    eq(R.unfold(function(n) { return n > 0 ? [[n, n - 1]] : []; }, 5), [5, 4, 3, 2, 1]);
   });
 
   it('is cool!', function() {
@@ -13,7 +13,11 @@ describe('unfold', function() {
       return function(pair) {
         count += 1;
         if (count <= n) {
-          return [pair[0], [pair[1], pair[0] + pair[1]]];
+          var head = pair[0];
+          var tail = [pair[1], pair[0] + pair[1]];
+          return [[head, tail]];
+        } else {
+          return [];
         }
       };
     };


src/unfold.js
```diff
@@ -3,9 +3,9 @@ var _curry2 = require('./internal/_curry2');
 
 /**
  * Builds a list from a seed value. Accepts an iterator function, which returns
- * either false to stop iteration or an array of length 2 containing the value
- * to add to the resulting list and the seed to be used in the next call to the
- * iterator function.
+ * either an empty array to stop iteration or an array with an array of length 2,
```
davidchambers: Perhaps _a singleton array containing an array_ in place of _an array with an array_?

src/unfold.js
```diff
@@ -15,23 +15,25 @@ var _curry2 = require('./internal/_curry2');
  * @category List
  * @sig (a -> [b]) -> * -> [b]
```
davidchambers: Please update the type signature.

test/unfold.js
```diff
@@ -4,17 +4,15 @@ var eq = require('./shared/eq');
...
 
   it('is cool!', function() {
     var fib = function(n) {
       var count = 0;
       return function(pair) {
         count += 1;
-        if (count <= n) {
-          return [pair[0], [pair[1], pair[0] + pair[1]]];
-        }
+        return count <= n ? [[pair[0], [pair[1], pair[0] + pair[1]]]] : [];
```
davidchambers: `]]]]` – yikes!

src/unfold.js
sadasant: Got it! omw

src/unfold.js
sadasant: Would it be good as `(a -> [[b]]) -> * -> [b]` ?

davidchambers: It should look similar to the signature of [`S.unfoldr`][1], though in Ramda the type of `['foo', 'bar', 'baz']` is written `[String]` rather than `Array String`.


[1]: https://sanctuary.js.org/#unfoldr

test/unfold.js
```diff
@@ -4,7 +4,7 @@ var eq = require('./shared/eq');
 
 describe('unfold', function() {
   it('unfolds simple functions with a starting point to create a list', function() {
-    eq(R.unfold(function(n) {if (n > 0) {return [n, n - 1];}}, 10), [10, 9, 8, 7, 6, 5, 4, 3, 2, 1]);
+    eq(R.unfold(function(n) { return n > 0 ? [[n, n - 1]] : []; }, 5), [5, 4, 3, 2, 1]);
```
sadasant: Made these 5 to shorten the line length.

test/unfold.js
```diff
@@ -13,7 +13,11 @@ describe('unfold', function() {
       return function(pair) {
         count += 1;
         if (count <= n) {
-          return [pair[0], [pair[1], pair[0] + pair[1]]];
+          return [[
+            pair[0], [pair[1], pair[0] + pair[1]]
+          ]];
```
sadasant: Is this better?

src/unfold.js
sadasant: So, `(a -> Maybe [b]) -> * -> [b]` ? 👀

src/unfold.js
```diff
@@ -3,35 +3,37 @@ var _curry2 = require('./internal/_curry2');
...
+ * seed to be used in the next call of the iterator function.
  *
  * The iterator function receives one argument: *(seed)*.
  *
  * @func
  * @memberOf R
  * @since v0.10.0
  * @category List
- * @sig (a -> [b]) -> * -> [b]
+ * @sig (a -> Maybe [b]) -> * -> [b]
```
sadasant: Hey, sorry for pushing this. I actually don't like having `Maybe` here.

`(a -> Array (Array b)) -> * -> Array b` ?

sadasant: `(a -> Array Array b) -> * -> Array b`  ?

davidchambers:
```javascript
R.replace('Maybe', 'Array', R.nth(1, R.split(' :: ', String(S.unfoldr))))
```

sadasant: Pushed `(b -> Array (Pair a b)) -> b -> Array a` \o/
Yep, code is a lot better than words 👍 thanks for the patience.

davidchambers: The final step is to replace occurrences of `Array _` with `[_]`, to match Ramda style.

sadasant: Hey! Thank you very much, I'm learning a couple of things on these signatures 👍 

So, it should be `(b -> [Pair a b]) -> b -> [a]` ? I'm worried about `Pair` not being a thing in Ramda. I know there is `R.pair` though.

davidchambers: We could use `(a, b)` in place of `Pair a b`. The Ramda team hasn't yet decided how pairs should be represented (#1554).

sadasant: It has been taking me a while to understand that you were referring to the usage order of `a` and `b` within my previous proposals 😅 now I see it, and it makes sense! (since what's being pass to the function is the second value of the previous iteration).

sadasant: I'm guessing now that `(b -> [[a, b]]) -> b -> [a]` is the right approach since it doesn't add anything new to the existing signatures. (?)

sadasant: Oh, I just got your message. Ok making the changes.

src/unfold.js
```diff
@@ -3,35 +3,37 @@ var _curry2 = require('./internal/_curry2');
...
+ * seed to be used in the next call of the iterator function.
  *
  * The iterator function receives one argument: *(seed)*.
  *
  * @func
  * @memberOf R
  * @since v0.10.0
  * @category List
- * @sig (a -> [b]) -> * -> [b]
+ * @sig (b -> [(a, b)]) -> b -> [a]
```
sadasant: ⬆️ 👀 🙇  @davidchambers

davidchambers: LGTM

buzzdecafe: since it has to be implemented as an Array for it to work, i am dubious about the aspirational signature here. Makes me wanna revive pr to add Maybe to API and have the function be `(b -> Maybe [a, b])`

sadasant: I wouldn't mind any solution, but since we're asking ourselves this question, I would propose to rely on a boolean value for these checks. It will make the code cleaner, and it will be easier to understand for newcomers. Otherwise I'm up for the use of a generic Maybe exposed through Ramda's API. We could even use a [Symbol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) for this.

sadasant: We could also use throw and try-catch it. Is that a bad idea?

sadasant: On favor of arrays, nothing will be shorter than just wrapping the result in `[]`.

sadasant: There's another option, which is using functions (?) `()=>myValue` (?) or `x=>MyValue`

sadasant: Actually, if we define the stopping value to, let's say, `false`, people would still be able to aggregate boolean values with an array `[true || false]`. Isn't it better?

buzzdecafe: > Actually, if we define the stopping value to, let's say, false, people would still be able to aggregate boolean values with an array [true || false]. Isn't it better?

nope, not in my opinion

sadasant: Valid! Let's define a Ramda way, I can help migrate the other functions ❤

CrossEye:
This must have come in one of those periods where I wasn't active.  I don't remember it at all.

The current `unfold` is awful, and I would love to replace it.  But I think I would rather something like this:

```js
const unfold = (fn, init) => 
  fn (
    init,
    (v, s) => [v, ... unfold (fn, s)],
    () => []
  )

const fibo = (n) => 
  unfold (
    ([a, b], next, stop) => b > n ? stop () : next (b, [b, a + b]),
    [0, 1]
  )


fibo (20) //=> [1, 1, 2, 3, 5, 8, 13]
```

which because of recursion-depth limitations and our ES3-style, would probably end up looking more like this:

```js
const unfold = function (fn, init)  {
  var done = false, 
      res = [],
      seed = init;
  while (!done) {
    fn (
      seed,
      function (v, s) {res .push (v); seed = s;},
      function () {done = true;}
    )
  }
  return res
}
```

But as we're trying not to introduce breaking changes on our way to `v1.0`, I think this will have to be put aside until a later version.  Tagging it to look at later and closing for now.

PR #2141, opened 2017-03-23
mapObject

cglane:
Allow for dictionary creation from two lists, the first list being a list of keys, and the second list being the list of values.


davezuko:
I think you may be looking for http://ramdajs.com/docs/#zipObj, in which case you're in luck :).

cglane:
Well there you go.
-Charles

buzzdecafe:
yup, looks like `zipObj`

PR #2142, opened 2017-03-23
Rephrase docs for 'all'

lottec:
Rephrased description for 'all' to make it clearer that it should return 'true' for an empty list

### DIFF
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -990,8 +990,7 @@
     });
 
     /**
-     * Returns `true` if all elements of the list match the predicate, `false` if
-     * there are any that don't.
+     * Returns `false` if any element does not satisfy the predicate, `true` otherwise.
      *
      * Dispatches to the `all` method of the second argument, if present.
      *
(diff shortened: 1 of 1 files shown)


dist/ramda.js
```diff
@@ -990,8 +990,7 @@
     });
 
     /**
-     * Returns `true` if all elements of the list match the predicate, `false` if
-     * there are any that don't.
+     * Returns `false` if any element does not satisfy the predicate, `true` otherwise.
```
davidchambers: That *is* the definition of all. :trollface:

buzzdecafe:
please update src/all.js instead of the dist/ramda.js file, which is generated from the sources.

dist/ramda.js
CrossEye: @davidchambers: Are you saying that the original *is* the definition of `all` or that the suggested version is?

CrossEye:
I know that @buzzdecafe commented in favor of this change in https://github.com/ramda/ramda.github.io/pull/172#issuecomment-288777477, but I'm not so sure.  

I think a focus on the positive side is clearer.  If it's necessary to explain, I would rather add a second sentence with a link to the [Wikipedia article on `Vacuous truth`][vt] than to try to explain `all(predicate)` as `not(any(not(predicate)))`.


:-1:

  [vt]: https://en.wikipedia.org/wiki/Vacuous_truth

dist/ramda.js
davidchambers: I believe I was trying to say that the two sentences (red and green) are semantically equivalent.

kedashoe:
> I would rather add a second sentence with a link to the Wikipedia article on Vacuous truth

I like this suggestion. It comes up often enough. Another sentence simply stating that an empty list will return `true` with a link to the wiki article would be nice. Would you like to update your PR, @lottec ?

Issue #489, opened 2014-11-02
InvokerN should take a method name – not a function


paldepind:
According to the documentation `R.invokerN` it is a function that

> Turns a named method of an object (or object prototype) into a function that can be called directly.

It appears that the description for `R.invokerN` was not updated after [this commit](https://github.com/ramda/ramda/pull/391) and thus isn't accurate anymore. What `invokerN` currently does is turn a function into a function taking one additional argument and calling the original function bound to the additional argument. Basically the function is just curry and `Function.prototype.call()` with a different argument order. With that definition this piece of code

```
var charAt = R.invoker(String.prototype.charAt);
charAt(6, 'abcdefghijklm'); //=> 'g'
```

does the same thing as

```
charAt = Function.prototype.call.bind(String.prototype.charAt);
charAt('abcdefghijklm', 6); //=> 'g'
```

But the above two code examples does something quite different from this:

   'abcdefghijklm'.charAt(6) //=> 'g'

This last piece of code actually utilizes the prototype chain. The current definition of `invokerN` does not however and thus isn't an adequate replacement for the typical invocation of a function on an object in JavaScript.

In my opinion `invokerN` would be a much more useful function if its first argument was a named method given as a string instead of a function. It would then invoke the named method on it's last argument.

Here is an example implementation:

```
var invokerN2 = R.invokerN2 = function invokerN2(len, method) {
  return R.curryN(len + 1, function() {
    var target = arguments[len];
    return target[method].apply(target, Array.prototype.slice.call(arguments, 0, len));
  });
};
```

To show an example where the difference actually matter consider the case where one would like to create an invoker for finding child elements with `querySelector`:

```
var querySelector1 = R.invokerN(1, Element.prototype.querySelector);
var querySelector2 = R.invokerN2(1, 'querySelector'); // <- Uses the proposed definiton of invokerN
```

Now one can do both `querySelector1('h1', document.body)` and `querySelector2('h1', document.body)`. However while `querySelector2('h1', document)` works fine `querySelector1('h1', document)` does _not_. Instead an error is thrown because `Element.prototype.querySelector` cant' be called bound to `document` since `document` is an instance of `Document` and not `Element`. The former works because `Document` has a [`querySelector` function on its prototype](https://developer.mozilla.org/en-US/docs/Web/API/document.querySelector) and the proposed invoker definition walks the objects actual prototype chain to find the name method.

[...]

PR #2143, opened 2017-03-23
Add boundProp function

mearns:
Add `boundProp` function to return the specified property from an object, and bind it to that object.

### DIFF
--- a/index.js
+++ b/index.js
@@ -22,6 +22,7 @@ module.exports = {
   binary: require('./src/binary'),
   bind: require('./src/bind'),
   both: require('./src/both'),
+  boundProp: require('./src/boundProp'),
   call: require('./src/call'),
   chain: require('./src/chain'),
   clamp: require('./src/clamp'),
--- a/src/boundProp.js
+++ b/src/boundProp.js
@@ -1,6 +1,6 @@
 var _curry2 = require('./internal/_curry2');
-var hasIn = require('./hasIn');
 var bind = require('./bind');
+var hasIn = require('./hasIn');
 
 /**
  * Returns a function that when supplied an object returns the indicated
@@ -16,8 +16,8 @@ var bind = require('./bind');
  * @param {String} p The property name
  * @param {Object} obj The object to query
  * @return {*} The value at `obj.p`, bound to `obj` as with `bind`.
- * @see R.prop
  * @see R.bind
+ * @see R.prop
  * @example
  *
  *      var getX = R.boundProp('getX');
--- a/src/boundProp.js
+++ b/src/boundProp.js
@@ -1,0 +1,51 @@
+var _curry2 = require('./internal/_curry2');
+var hasIn = require('./hasIn');
+var bind = require('./bind');
+
+/**
+ * Returns a function that when supplied an object returns the indicated
+ * property of that object, bound to that object, if it exists. Note that
+ * if the property exists but is not a function, this function will _not_
+ * fail, but you will get an error when you try to invoke the returned
+ * function.
+ *
+ * @func
+ * @memberOf R
+ * @category Object
+ * @sig s -> {s: a} -> a | Undefined
+ * @param {String} p The property name
+ * @param {Object} obj The object to query
+ * @return {*} The value at `obj.p`, bound to `obj` as with `bind`.
+ * @see R.prop
+ * @see R.bind
+ * @example
+ *
+ *      var getX = R.boundProp('getX');
+ *
+ *      getX({
+ *        x: 'X',
+ *        getX: function () {return this.x}
+ *      })();   //=> 'X'
+ *
+ *      getX({});   //=> undefined
+ *
+ *
+ *      // A more practical use case:
+ *
+ *      function Square(length) {
+ *        this._length = length;
+ *      }
+ *      Square.prototype.getArea = function() {
+ *        return this._length * this._length;
+ *      };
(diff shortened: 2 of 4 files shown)


buzzdecafe:
i am sorry to say i am a bit cold on this PR.  Use of `this` is not something I think we should encourage. I am just one voice; but I would need to hear a compelling reason to include such a function.

mearns:
@buzzdecafe I can certainly appreciate your reservations, given that this is a functional library. I guess my argument would be that this is a functional library for a language that supports objects and binding. To that point, ramda already provides a `bind` function to do the "objectionable" part of the added `boundProp` function, this just makes it easier to apply in some circumstances.

I don't think this function really encourages the use of `this`, per se, but it permits you to use ramda and functional techniques to more easily deal with existing uses of `this`.

The use case I encountered that prompted this PR was fairly general _if_ you're using objects with methods. I had a collection of objects that all had a particular named method on them. I wanted to map the collection of objects to a collection of the results of calling that method on each object. In plain old JS, it would simply be:

```javascript
const results = myObjects.map((obj) => obj.myMethod());
```

With out the added function, I would use ramda as follows:

```javascript
const results = myObjects.map(R.converge(R.bind, [R.prop('myMethod'), R.identity]).map(R.call);
```

With the added `boundProp` function, this can be reduced to:

```javascript
const results = myObjects.map(R.boundProp('myMethod')).map(R.call);
```

svozza:
I agree, `R.bind` should be deprecated. :wink: Seriously though, if one needs the functionality, it is possible to just use an arrow function as that binds `this` in the appropriate way.

mearns:
@svozza (meh). You can use an arrow function. You can use arrow functions to replace all of ramda's functions. The point is to make it more easier and more functional, right?

I don't feel strongly about this feature if there isn't general support for it. But I think if it's discarded specifically because it promotes the idea of object binding, then `bind` should deprecated as suggested above, for the same reason.

buzzdecafe:
it's an open question where the "boundaries" of ramda are. It occupies an idiosyncratic space; not as strict as sanctuary, but more opinionated than underscore. `R.bind` is not something I find myself using, and it certainly is a concession to the realities of JS. 

If we agree that `bind` is a bit ... gross, then arguing that since `bind` is included, therefore sugar for `bind` should also be included, is not persuasive to me.

Your examples are useful:

```js
// 1
const results = myObjects.map((obj) => obj.myMethod());

// 2
const results = R.map(R.call, R.map(R.converge(R.bind, [R.prop('myMethod'), R.identity]), myObjects));

// 3
const results = R.map(R.call, R.map(R.boundProp('myMethod'), myObjects));
```

To my mind, example `1` is by far the best of the bunch for clarity. If your goal is to make a composable function, then you could use `invoker` as well, another of ramda's red-headed stepchildren: 

```js
// 4
// f :: [Obj] -> [???]
const f = R.map(R.invoker(arityOfMethod, 'myMethod')); 
```

buzzdecafe:
> I thought R.bind, R.invoker ("ramda's red-headed stepchildren") are provided to enable using JS class methods in the same functional manner as "free-standing" JS functions? ES 6 class and HTML5 Api methods need to be bound to a suitable this when invoked. Why would Ramda deny them syntactic sugar? To encourage OO-style programming with those APIs?

It depends on what the goals of the library are. This area is really @CrossEye 's domain I think. If I may be so bold as to play his spokesman for a moment, I think he sees the lib as a gateway to a different way of thinking, and wants to smooth over the bits that don't fit well (e.g. `this` etc.)  I think that's where the red-headed stepchildren came from: @CrossEye attempting to bridge the gap from native JavaScript to a more "functional" idiom, or as you put it, "to enable using JS class methods in the same functional manner as 'free-standing' JS functions." 

That's not a priority for me at all; but that's why I said I am just one voice. 

> Btw. I do not see so many differences between `R.boundProp` and `R.invoker`.

I noticed the same thing; I agree.

mearns:
Yup, `invoker` is very similar. I did a scan for functions that already solved my problem and did not find this one =\. The only thing I'll say about it is that with the proposed `boundProp`, you don't need to know the arity of the method, and specifically, you can invoke the resulting function with variable number of arguments.

CrossEye:
I've been out of the loop for a while now.

@buzzdecafe has it right.  I originally imagined including pure versions of all the native methods, for things like `Date`, `Regexp`, `Array`, etc.  But early on we realized that his was unwieldy.  `Invoker` was a way to allow the users to do this for whatever objects they like.

But it is absolutely a concession to the way JS was and is being written rather than a principled addition to the library.  I can easily imagine a later version of Ramda removing these.  I never use `bind` myself, although I do use `invoker`.

`invoker` went through some serious changes in its early days (see #489 for some details).  We did consider variants where one didn't have to supply the arity, and ones with better handling of variadic functions.  The current version simply seemed to strike the best balance.  So I don't see much call for `boundProp`.

-----

@buzzdecafe: In the era of Trump and Spicer, "spokesman" takes on some pretty awful connotations!  :smile:

buzzdecafe:
> @buzzdecafe: In the era of Trump and Spicer, "spokesman" takes on some pretty awful connotations! 😄

FWIW I think you come out slightly worse in that comparison

mearns:
Based on this discussion, I'm comfortable closing this PR. If there is any different decision, feel free to re-open it =)

PR #2144, opened 2017-03-24
Fix example for transpose documentation

simplysh:
This fixes example for `transpose` failing in REPL because of invalid comment.
http://ramdajs.com/docs/#transpose


kedashoe:
ty!

PR #2146, opened 2017-03-24
Added `repeatWith` function

### LINKED ISSUE #2145: Repeat like replicateM in Haskell?
Is there a function like `replicateM` from Haskell in ramda?

What I mean by this is a function, lets call it `repeatFun` for now, that behaves like this:
```
repeatFun(Math.random, 3) //=> [0.156, 0.789, 0.456]
```

If there is not such a function that is a feature I would gladly try to write one.

TheGrandmother:
Added a `repeatWith` function that behaves like `replicateM` in Haskell.

The function takes a nullary function `f` and creates a list of `n` containing the return values of `f` for `n` consecutive calls.

Example:
```
R.repeatWith(Math.random, 3) 
//=> [ 0.44077919192880755, 0.09582463920980899, 0.13967090898447432 ]
```

The side effects produced by `f` will occur in the order from "left to right":
```
var i = 0;
R.repeatWith(() => i = i+1, 5) //=> [1,2,3,4,5]
```

Closes #2145

### DIFF
--- a/index.js
+++ b/index.js
@@ -183,6 +183,7 @@ module.exports = {
   reject: require('./src/reject'),
   remove: require('./src/remove'),
   repeat: require('./src/repeat'),
+  repeatWith: require('./src/repeatWith'),
   replace: require('./src/replace'),
   reverse: require('./src/reverse'),
   scan: require('./src/scan'),
--- a/src/repeatWith.js
+++ b/src/repeatWith.js
@@ -1,0 +1,27 @@
+var _curry2 = require('./internal/_curry2');
+var map = require('./map');
+var repeat = require('./repeat');
+
+
+/**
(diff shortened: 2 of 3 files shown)


CrossEye:
I'm trying to catch up on older discussions that I missed.

I'm afraid I'm :-1: on this.

Ramda is about making it easier to do functional programming in JS.

The only feature that this function offers not already available with either `repeat` or `times` is the ability to use an impure unary function to get differing values for each repeat.  That seems very much contrary to Ramda's focus on FP.

paldepind:
I definitely agree with @CrossEye. This function is only useful as an imperative function. Furthermore, the goal can easily be accomplished by using existing functions.

PR #2147, opened 2017-03-25
add Chinese translation link to README.md (home page)

adispring:
We've just finished Chinese translation of ramda, and build a website: http://ramda.cn/.

The work is done by five of us(4 other colleagues). we then checked the translation twice carefully to assure quality.

We currently replace the English comment with Chinese in-place on the api file, in the forked [repo](https://github.com/ramdacn/ramda). I'm writing some scripts to move the translation to a single directory, which will make the future translation more easily: 

1. update the forked ramda repo with original repo;
2. show the change parts;
3. only translate the change apis;
4. update the forked ramda.github.io with new translation;
5. update Chinese website.

Should the translations separate from original ramda(only add a link), or be part of it?

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,10 @@ If you have _PhantomJS_ installed, you can run `testem -l phantomjs` to run the
 tests completely headlessly.
 
 
+Translations
+-----------------
+
+[Chinese(中文)](http://ramda.cn/)
 
 
 Acknowledgements


kedashoe:
:+1:

CrossEye:
Just getting back to Ramda after some time away.  This is great.  Thanks for all the hard work.

One other option for making it easier to add translations might be to move the documentation comments out of the source code files, and store them in a parallel directory.  It shouldn't be difficult to marry them together in some pre-docs step.  And then we could bring this back into Ramda proper.  I like it for another reason too: I find the verbose comments needed for JSDoc to interfere with reading the source.

adispring:
@CrossEye Yes, I have just write two shell scripts to separate and translate apis automatically.

1. Pick all comments to separate dir: https://github.com/ramdacn/ramda/blob/coding-pages/scripts/pickComment.sh
2. translation script: https://github.com/ramdacn/ramda/blob/coding-pages/scripts/translate.sh
3. separated i18n dir: https://github.com/ramdacn/ramda/tree/coding-pages/i18n

CrossEye:
That's great!

We'd have to see if the overall appetite is there from the team, but if you're interested, I'd love to fold this back into Ramda itself.

Again, thanks for the great work!

adispring:
Fold this back to Ramda is a good idea.

svozza:
Nice work!

PR #2156, opened 2017-04-13
Updated type for pluck

Bradcomp:
Pluck is probably one of our more well used Ramda functions. If we see `map(prop(...))` during code review we'll typically change it to a `pluck`. I recently wrote some code that was using `map(prop())` to pull some values off a nested object, and of course it was pointed out that `pluck` could be used there as well.

While obvious in retrospect, it isn't obvious from the documentation that `pluck` can be used on things that aren't arrays. To me, this behavior seems useful, and isn't ad-hoc due to the `Functor` instance for objects. 

I understand the desire for the docs to be easily understandable, and I know that `Functor` is less well known in the JS community than `Array` is. I don't want to remove the idea that this function operates on lists, simply add in the idea that it will work with other Functors as well.

I wasn't really sure what to do with the jsdoc annotations.  Looking at `map` it still has Array as the type for the second parameter and return value, so I left that here as well.

### DIFF
--- a/src/pluck.js
+++ b/src/pluck.js
@@ -7,19 +7,24 @@ var prop = require('./prop');
  * Returns a new list by plucking the same named property off all objects in
  * the list supplied.
  *
+ * `pluck` will work on
+ * any [functor](https://github.com/fantasyland/fantasy-land#functor) in
+ * addition to arrays, as it is equivalent to `R.map(R.prop(k), f)`.
+ *
  * @func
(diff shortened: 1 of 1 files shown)


kedashoe:
:+1:

davidchambers:
I think it would be useful to mention the equivalence between `pluck` and `map` + `prop`.

buzzdecafe:
:cow2:

PR #2164, opened 2017-04-22
Refactored values

idcmardelplata:
Hi, I have refactored the function values

### DIFF
--- a/src/values.js
+++ b/src/values.js
@@ -18,6 +18,9 @@ var keys = require('./keys');
  *
  *      R.values({a: 1, b: 2, c: 3}); //=> [1, 2, 3]
  */
-module.exports = _curry1(function values(obj){
-  return keys(obj).map(function(key) { return obj [key]; });
+module.exports = _curry1(function values(obj) {
+  return keys(obj)
+    .map(function(key) {
+      return obj[key];
+    });
 });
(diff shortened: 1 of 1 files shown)


davidchambers:
What's the motivation for this change, @idcmardelplata?

idcmardelplata:
Less lines, more functional and ¿readability? :)

fix-fix:
It doesn't really matter if the innards are written in "functional" style or not, as long as they convey the same semantics and don't break the rules.

I think the sole reason behind for/while loops was performance. More precisely `Array#map` is more than order of magnitude slower in V8, you can run the [benchmark here](https://jsperf.com/native-map-versus-array-looping/85) and see the difference.

With this and the [project's philosophy ](https://github.com/ramda/ramda#philosophy) in mind I don't think this change is worth it.

svozza:
It's also ES5 only, although I'm not sure if that really matters now.

CrossEye:
https://jsperf.com/values-direct-vs-map/1

Ramda ideally does not worry about implementation elegance.  We've [been down that road before][ew] and it wasn't pretty!  :wink:

idcmardelplata:
Ha ok, excuse me I did not know about the performance, I just thought I could make that contribution, but thanks for clarifying it so I know for the next.

CrossEye:
I meant to add to that post before I sent it.  I apologize for not doing so.

I also wanted to thank you for your contribution.  There are probably many ways our implementation can be improved.  I hope you'll continue to look around.  More eyes are always a good thing, and there is nothing fundamentally wrong with your suggestion.  We tend to keep performance in mind here after having been burned badly by it before.  But we do also value elegant code, so please keep on investigating!

PR #2168, opened 2017-04-26
Fix reduced value handling when dispatching to objects implementing the iterable/iterator protocols

ognen:
According the documentation of `reduce`:

    The iterator function receives two values: *(acc, value)*. It may use 
   `R.reduced` to shortcut the iteration.

and

    Dispatches to the `reduce` method of the third argument, if present.

This pull request fixes the situation where an object with a `reduce` method is passed in list position and a transducer which uses `reduced` (such as `take`) is used as the reducing fn.

### DIFF
--- a/test/reduce.js
+++ b/test/reduce.js
@@ -35,4 +35,51 @@ describe('reduce', function() {
     eq(sum.length, 1);
   });
 
+  it('Prefers the use of the iterator of an object over reduce (and handles short-circuits)', function() {
+    var symIterator = (typeof Symbol !== 'undefined') ? Symbol.iterator : '@@iterator';
+
+    function Reducible(arr) {
+      this.arr = arr;
+    }
+
+    Reducible.prototype.reduce = function(f, init) {
+      var acc = init;
+      for (var i = 0; i < this.arr.length; i += 1) {
+        acc = f(acc, this.arr[i]);
+      }
+      return acc;
+    };
+
+    Reducible.prototype[symIterator] = function() {
+      var a = this.arr;
+      return {
+        _pos: 0,
+
+        next: function() {
+          if (this._pos < a.length) {
+            var v = a[this._pos];
+            this._pos += 1;
+            return {
+              value: v,
+              done: false
+            };
+          } else {
+            return {
+              done: true
+            };
+          }
+        }
+      };
+    };
+
+    var xf = R.take(2);
+    var apendingT = { };
+    apendingT['@@transducer/result'] = R.identity;
+    apendingT['@@transducer/step'] = R.flip(R.append);
+
+    var rfn = xf(apendingT);
+    var list = new Reducible([1, 2, 3, 4, 5, 6]);
+
+    eq(R.reduce(rfn, [], list), [1, 2]);
+  });
 });
--- a/src/internal/_reduce.js
+++ b/src/internal/_reduce.js
@@ -46,15 +46,16 @@ module.exports = (function() {
     if (typeof list['fantasy-land/reduce'] === 'function') {
       return _methodReduce(fn, acc, list, 'fantasy-land/reduce');
     }
-    if (typeof list.reduce === 'function') {
-      return _methodReduce(fn, acc, list, 'reduce');
-    }
     if (list[symIterator] != null) {
       return _iterableReduce(fn, acc, list[symIterator]());
     }
     if (typeof list.next === 'function') {
       return _iterableReduce(fn, acc, list);
     }
+    if (typeof list.reduce === 'function') {
+      return _methodReduce(fn, acc, list, 'reduce');
+    }
+
     throw new TypeError('reduce: list must be array or iterable');
   };
 }());
--- a/src/reduce.js
+++ b/src/reduce.js
@@ -17,7 +17,9 @@ var _reduce = require('./internal/_reduce');
  * on this behavior, see:
  * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce#Description
  *
- * Dispatches to the `reduce` method of the third argument, if present.
+ * Dispatches to the `reduce` method of the third argument, if present. When
+ * doing so, it is up to the user to handle the `R.reduced` shortcuting, as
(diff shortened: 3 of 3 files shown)


kedashoe:
Good catch, thanks for the PR @ognen . This is a tricky one. I don't think it is possible to fix this in ramda. What this PR does is see that we have received `reduced` and continue returning that value rather than calling step. This is kind of ok. One problem is that if you `R.take(2)` from a structure with a million elements, the user's `reduce` method will still get called a million times. A bigger problem is that if a user has an infinite iterator in their reduce, it will go on forever even if we get `reduced`..

The best solution I can think of is to tell the user they have to handle `reduced` in their `reduce` method. Would you like to update the docs to say this? I actually wouldn't mind if you kept the test as well with a description like 'allows the user to handle reduced values when dispatching to the reduce method' that shows how one can do this.

ognen:
@kedashoe Yes, the fix won't perform well (only correct). I've considered the issue at length and the conclusion was:
* the `reduce` method, as is defined in Array is not possible to short-cut, this is the definition we are working with;
 * If the user owns the "custom" class providing the `reduce` method then yes, the user would be able to shortcut the reduction within it, but 
* what if the user is using ramda on a 3rd party provided library and has no control over the reduce method?

Context: I'm writing a [wrapper][ramda-immutable] around ramda that enables it to be used with immutable.js. Immutable's collections respond to `reduce`; are not aware of the shortcutting; and one cannot change their reduce method. 

How do I proceed?

[ramda-immutable]: https://github.com/netceteragroup/girders-elements/tree/features/ramda-package/packages/ramda

-

CrossEye:
I don't think I'm understanding this correctly.

Thank you @kedashoe for your explanation.  I was having a difficult time trying to remember what I had once known about transducing.

But I'm still missing how this fix, @ognen, helps you solve your problem.  If your underlying function does not return `reduced`, how can a wrapper around it return `reduced`?

At first take, I like @kedshoe's suggestion.  But I think I'm still confused.

test/reduce.js
```diff
@@ -1,5 +1,6 @@
 var R = require('..');
 var eq = require('./shared/eq');
+var _xwrap = require('../src/internal/_xwrap');
```
CrossEye: Uggh, I don't know if we do this anywhere else, but I really hate the idea of a test depending directly on our internals.

kedashoe:
> Context: I'm writing a wrapper around ramda that enables it to be used with immutable.js. Immutable's collections respond to reduce; are not aware of the shortcutting; and one cannot change their reduce method.

Maybe write a wrapper around immutable objects to intercept reduce?

Another option.. Immutable collections implement the iterator protocol. That means we could move [these lines](https://github.com/ramda/ramda/blob/master/src/internal/_reduce.js#L52-L57) above the ones that check for a reduce method and your immutable stuff will go through that path and short circuit. Would that work for you?

CrossEye:
> Another option.. Immutable collections implement the iterator protocol. That means we could move [these lines](https://github.com/ramda/ramda/blob/master/src/internal/_reduce.js#L52-L57) above the ones that check for a reduce method and your immutable stuff will go through that path and short circuit. Would that work for you?

That sounds like better behavior in any case, no?  Dispatching should only happen once we have exhausted the known types.

kedashoe:
> That sounds like better behavior in any case, no? Dispatching should only happen once we have exhausted the known types.

Works for me. Thoughts @ognen ?

ognen:
@kedashoe @CrossEye Yes, this is the better option. The interable protocol is well defined and can be short-cut by ramda.  Giving it priority over `reduce` makes total sense! 

Will update the pull request.

test/reduce.js
ognen: Ugly, yes, but I wasn't able to find a public fn that "lifts" a normal reducing function `a -> b -> a` into a "transformer" (an object that contains the step, result and init cases of a). 

Transducers (e.g. `take(2)`) expect to be applied to a transformer.

test/reduce.js
```diff
@@ -35,4 +35,51 @@ describe('reduce', function() {
...
+      }
+      return acc;
+    };
+
+    Reducible.prototype[symIterator] = function() {
+      var a = this.arr;
+      return {
+        _pos: 0,
+
+        next() {
```
kedashoe: It looks like this syntax isn't working for some older envs in our CI.

CrossEye: Yes, for version 1 -- if we ever get there, we're sticking with ES5.  Actually, we're very close to ES3, in fact.

ognen: Yes, sorry about that... ES6 syntax does get under one's skin after a little while.

kedashoe:
lgtm, could you squash the commits @ognen ?

buzzdecafe:
@kedashoe you can squash the commits with the big green button

kedashoe:
> @kedashoe you can squash the commits with the big green button

True, but someone around here mentioned they don't like messing with other's commits and it makes sense to me.

I have done it before for documentation only PRs.. but even then the UI stabbed me in the back, resulting in a commit message I was not expecting :open_mouth:

buzzdecafe:
ah, my apologies. i didn't mean to aggravate your PTSD. 😉 
Carry on then

CrossEye:
>> @kedashoe you can squash the commits with the big green button

> True, but someone around here mentioned they don't like messing with other's commits and it makes sense to me.

I think we need to make a decision on this.  I for one was very happy with the big green button solution when it was introduced.  I find it painful to squash, and waiting for contributors to do so gains little, so this seemed ideal.  But I've also now occasionally been bitten by it.  I'm still in favor, but less strongly.

What do people think?  Should we keep the Squash Merge / Rebase Merge options or always choose between merging a collection of commits and requesting the user to squash?

mattgrande:
I like to Squash Merge with the big green button on my own projects, personally. I haven't tried the Big Green Rebase yet, though.

ognen:
Squashed.

kedashoe:
@ognen :+1:

CrossEye:
Thanks, @ognen!

PR #2170, opened 2017-05-04
Add @see tags linking binary, nAry and unary

Delapouite:
Hi, this PR adds a few more links between arity functions.


buzzdecafe:
Thank you

:cow2:

PR #2171, opened 2017-05-04
Add @see tags between ascend and descend

Delapouite:
Hi. These two functions offer similar abilities.


buzzdecafe:
Thanks again!

:cow2:

PR #2174, opened 2017-05-10
Document R.ap as S Combinator

MikaAK:
#1916 Redo since I lost access to original repo. Not sure if you want signatures just let me know which one of these to use

```
ap :: Function x (a -> b) -> Function x a -> Function x b
```
```
ap :: (x -> a -> b) -> (x -> a) -> (x -> b)
```
```
ap :: (a -> b -> c) -> (a -> b) -> (a -> c)
```

### DIFF
--- a/src/ap.js
+++ b/src/ap.js
@@ -16,6 +16,7 @@ var map = require('./map');
  * @category Function
  * @sig [a -> b] -> [a] -> [b]
  * @sig Apply f => f (a -> b) -> f a -> f b
+ * @sig (a -> b -> c) -> (a -> b) -> (a -> c)
  * @param {*} applyF
  * @param {*} applyX
  * @return {*}
--- a/src/ap.js
+++ b/src/ap.js
@@ -23,6 +23,10 @@ var map = require('./map');
  *
  *      R.ap([R.multiply(2), R.add(3)], [1,2,3]); //=> [2, 4, 6, 4, 5, 6]
  *      R.ap([R.concat('tasty '), R.toUpper], ['pizza', 'salad']); //=> ["tasty pizza", "tasty salad", "PIZZA", "SALAD"]
+ *
+ *      // R.ap can also be used as S combinator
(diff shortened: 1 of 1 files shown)


src/ap.js
```diff
@@ -23,6 +23,10 @@ var map = require('./map');
  *
  *      R.ap([R.multiply(2), R.add(3)], [1,2,3]); //=> [2, 4, 6, 4, 5, 6]
  *      R.ap([R.concat('tasty '), R.toUpper], ['pizza', 'salad']); //=> ["tasty pizza", "tasty salad", "PIZZA", "SALAD"]
+ *
+ *      // R.ap can also be used as S combinator
+ *      // when only two functions are passed
+ *      R.ap(R.concat, R.toUpper)('Ramda') //=> 'RamdaRAMDA'
```
davidchambers: I'd prefer `, ` in place of `)(`.

CrossEye:
I think I'd prefer this one:

```hs
ap :: (a -> b -> c) -> (a -> b) -> (a -> c)
```

src/ap.js
kedashoe: Like `R.ap(R.concat, R.toUpper, 'Ramda')`? I don't think we can call it like that

davidchambers: You're quite right, Kevin! Ignore my comment, @MikaAK. :)

kedashoe:
@MikaAK would you like to add the signature specialized for functions? Please squash as well :)

MikaAK:
@kedashoe  that's `ap :: (a -> b -> c) -> (a -> b) -> (a -> c)`? sorry I'm not very good with type sigs yet :(

kedashoe:
> that's ap :: (a -> b -> c) -> (a -> b) -> (a -> c)

yep :)

MikaAK:
Hopefully I got that one right! You should be able to squash on merge using the github ui 😄  but I can totally do it manually too if required!

PR #2176, opened 2017-05-12
docs: add missing intra links between Ramda's functions

Delapouite:
Hello.

Quite often while reading the online docs I feel frustrated because in function descriptions, sometimes function names are clickable, and sometimes they are not.
To ease the navigation and discoverability of related functions, I added the missing links between functions, following what had already been done for a few of them.

So, in this PR:
- no wordings was changed, just added links
- I built the docs on my local machine to try to click on each of them and trap eventual typos.

Potential fixes for a future PR: light inconsistency between Ramda's function starting with a `R.`prefix  and sometimes not.

### DIFF
--- a/src/addIndex.js
+++ b/src/addIndex.js
@@ -7,7 +7,7 @@ var curryN = require('./curryN');
  * Creates a new list iteration function from an existing one by adding two new
  * parameters to its callback function: the current index, and the entire list.
  *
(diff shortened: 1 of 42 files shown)


CrossEye:
This is a fantastic idea!

kedashoe:
Great work, thanks @Delapouite !

PR #2179, opened 2017-05-19
Add lazy evaluation example for R.transduce

zhangchiqing:
transducer is made for lazy evaluation. This is a power feature. I'd like to highlight it in the document. 

In order to do lazy evaluation with `R.transcue`, we could compose with functions that return `R.reduced` in the transducer for termination, such as `R.take`. But what's important is that the execution order of the composed functions is from left to right, which is the opposite order when calling the transducer directly. 

For instance
```
var numbers = [1, 2, 3, 4];
var transducer = R.compose(R.map(R.add(1)), R.take(2));
```
If I call the `transducer` directly like this `transducer(numbers)`, then `R.take(2)` will be executed first, then `R.map(R.add(1))` , which is "right to left" order
If I call the `transducer` with `R.transduce` like this:
```
R.transduce(transducer, R.flip(R.append), [], numbers); //=> [2, 3]
```
then `R.map(R.add(1))` will be executed first, `R.take(2)` next, which is "left to right" order.

However since `map f . take x == take x . map f`, regardless "left to right" or "right to left" order, they both yield to the same result. 

The current document uses this as the example, I think it's not clear enough for reader to notice the execution order of the composed transducers, as well as the lazy evaluation feature.

In this PR, I'm trying to make it more clear by adding one more example so that the reader will see transducers are exeucted from left to right. That is, `R.filter(isOdd)` will be executed first, then `R.take(1)` will terminate before traversing to the end of the array. The other way round won't make sense.

Please review. Thanks!

### DIFF
--- a/src/transduce.js
+++ b/src/transduce.js
@@ -44,8 +44,11 @@ var curryN = require('./curryN');
  *
  *      var numbers = [1, 2, 3, 4];
  *      var transducer = R.compose(R.map(R.add(1)), R.take(2));
- *
  *      R.transduce(transducer, R.flip(R.append), [], numbers); //=> [2, 3]
+ *
+ *      var isOdd = (x) => x % 2 === 1;
+ *      var firstOddTransducer = R.compose(R.filter(isOdd), R.take(1));
+ *      R.transduce(firstOddTransducer, R.flip(R.append), [], R.range(0, 100)); //=> [1]
  */
 module.exports = curryN(4, function transduce(xf, fn, acc, list) {
   return _reduce(xf(typeof fn === 'function' ? _xwrap(fn) : fn), acc, list);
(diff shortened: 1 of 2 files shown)


buzzdecafe:
This example looks good to me, although I wonder if there is a way to make the "laziness" more explicit.

zhangchiqing:
How about:

```js
var firstOddTransducer = R.compose(R.filter(isOdd), R.take(1));

R.transduce(firstOddTransducer, R.flip(R.append), [], R.range(0, 100)); //[1]
```

zhangchiqing:
Hi @buzzdecafe ,

I've updated the PR with the above example. Can I get your thoughts on it?

buzzdecafe:
LGTM 🐄 

i am still pondering how to make the pseudo-laziness more explicit in a test, but no reason to delay this.

PR #2181, opened 2017-05-23
Test ap with fantasy-land and use consistent ordering of combination

LiamGoodacre:
This PR is assuming that fantasy-land's `ap` is supposed to be like Haskell's `(<**>) :: Applicative f => f a -> f (a -> b) -> f b`, and not like `flip (<*>) :: Applicative f => f a -> f (a -> b) -> f b` (which is how Ramda currently treats it).

My test case attempts to show the difference by using `Either`.  The important case is the first one:

```javascript
eq(R.ap(leftFn, leftVal).simple(), leftFn.simple());

// i.e: forall e x. R.ap(Left(e), x) = Left(e)
```

### DIFF
--- a/src/ap.js
+++ b/src/ap.js
@@ -27,8 +27,10 @@ var map = require('./map');
  */
 module.exports = _curry2(function ap(applyF, applyX) {
   return (
-    typeof applyX['fantasy-land/ap'] === 'function' ?
-      applyX['fantasy-land/ap'](applyF) :
+    typeof applyF['fantasy-land/ap'] === 'function' ?
+      applyF['fantasy-land/ap'](map(
+        function (x) { return function (f) { return f(x); }; }, applyX)
+      ) :
     typeof applyF.ap === 'function' ?
       applyF.ap(applyX) :
     typeof applyF === 'function' ?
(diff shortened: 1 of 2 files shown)


kedashoe:
Thanks for the PR @LiamGoodacre . To comply with FL, we just pass along the first argument to the second if the latter has fantasy-land/ap. So for example, if sanctuary's Either were passed in to us, we would get here:

https://github.com/sanctuary-js/sanctuary/blob/master/index.js#L2519

. In the instance you highlighted, `other` is not a `Right` and we would get back `other`, ie `Left(e)` in your example. Some quick tests seem to confirm this.

Ramda doesn't really have much to say about it, it is up to the instance to implement `ap` correctly.

LiamGoodacre:
So you're saying that for FL, `ma['fantasy-land/ap'](mf)` is supposed to be implemented like `mf <*> ma` and not `ma <**> mf`?

I haven't been able to see on the FL spec that it is supposed to be one or the other.  Given the type signature I was expecting `<**>`.  Conversations in the Gitter channel also suggest it was supposed to be `<**>`.

kedashoe:
You probably know more about this than me :) There have been many discussions about `ap` in FL, see https://github.com/fantasyland/fantasy-land/pull/145 as a starting point. FL also has a gitter that is quite active, so you can ask there as well.

LiamGoodacre:
Closing because: lack of interest; I no-longer have the need to use Ramda; I'm not particularly invested in the outcome.

PR #2182, opened 2017-05-24
move deprecated isArrayLike to internal

kedashoe:
https://github.com/ramda/ramda/pull/1877

This turned out to be a bit more involved than expected. If someone could take a quick glance would be appreciated!


src/unless.js
```diff
@@ -21,10 +21,9 @@ var _curry3 = require('./internal/_curry3');
  * @see R.ifElse, R.when
  * @example
  *
- *      // coerceArray :: (a|[a]) -> [a]
- *      var coerceArray = R.unless(R.isArrayLike, R.of);
- *      coerceArray([1, 2, 3]); //=> [1, 2, 3]
- *      coerceArray(1);         //=> [1]
+ *      let longUpper = R.unless(x => x.length < 3, R.toUpper);
```
davidchambers: This isn't a great example for `R.unless`, as it could easily be expressed in terms of `R.when`.

kedashoe: how about something with `R.unless(R.isNil)`? do you have something in mind?

davidchambers: `R.unless(R.isNil)` sounds good to me!

buzzdecafe:
FWIW `_isArrayLike` appears in docs for 0.24.0.

Issue #2183, opened 2017-05-27
0.24.0 upgrade guide


kedashoe:
Commits: [v0.23.0...v0.24.0](https://github.com/ramda/ramda/compare/v0.23.0...v0.24.0)

Added:

- #1867 :new: `mergeDeepLeft`, `mergeDeepRight`, `mergeDeepWith`, `mergeDeepWithKey`
- #2120 :new: `startsWith`, `endsWith` 
- #2127 :new: `o`, binary compose 
- #2130 :new: `memoizeWith` 
- #1868 :new: `innerJoin` 

Removed:

- #2182 :x: `isArrayLike` (#1877)

Deprecated:

- #1868 :floppy_disk: `intersectionWith` 

Changes:

- #1900 [FantasyLand](https://github.com/fantasyland/fantasy-land) 1.0 :tada:
- #2036 `dissocPath` support integers
- #2041 `assocPath` handle nil
- #2119 `xtake` when n < 0 
- #2016 `groupWith` predicate receives adjacent elements
- #2168 `reduce` move dispatch to `reduce` function to least priority

Let me know if I missed anything!

Many thanks to all the contributors supporting this release, especially to those helping with numerous improvements to tests and documentation :tada:

angeloocana:
Hi!

When I upgrade to 0.24.0 my tests that uses memoize fail.

https://travis-ci.org/angeloocana/joj-core/builds/236872907

I don't know why, if i use 0.23.0 it works fine.

MarcoWorms:
The fantasy-land link after #1900 is broken > https://github.com/ramda/ramda/issues/fantasy-land/fantasy-land

kedashoe:
> The fantasy-land link after #1900 is broken > https://github.com/ramda/ramda/issues/fantasy-land/fantasy-land

ty

kedashoe:
@angeloocana 

The only test I see that mentions memoize looks like it is passing. Can you re-create the issue in the [repl](ramdajs.com/repl/)?

angeloocana:
@kedashoe  thanks for the answer!

I found the issue, in 0.24.0 memoize only checks the first param and ignores other params.

My example: https://goo.gl/kHxxAl

Simple example: https://goo.gl/ilvs7b

Is it right to memoize a function with more than 1 param?

I can try to fix ramda if it is really wrong behavior.

angeloocana:
The error happens because **memoize** is using **memoizeWith(toString)**

```js
[...]

PR #2185, opened 2017-05-30
Add toStringN and fix memoize(...n)

angeloocana:
After version 0.24.0 memoize with more than 1 argument stoped working.

issue: https://github.com/ramda/ramda/issues/2183

To fix it, I created toStringN() examples:
```js
	R.toStringN(0, -0); // '0,-0'
	R.toStringN(true, false); // 'true,false'
```


davidchambers:
This gets a :thumbsdown: from me, I'm afraid. If one wishes to apply `R.toString` to multiple values one already has two ways to do so:

```javascript
const x = 'foo';
const y = true;
const z = 42;

R.map(R.toString, [x, y, z]);
// => ['"foo"', 'true', '42']

R.toString([x, y, z]);
// => '["foo", true, 42]'
```

kedashoe:
Agreed w/ @davidchambers , I don't think we need `toStringN`. How about something like this for `memoize`, @angeloocana 

```js
module.exports = R.memoizeWith(function() {
  return R.toString(arguments);
});
```

angeloocana:
@davidchambers thanks! you are right! 
@kedashoe thanks! it worked like a charm! New PR created.

PR #2186, opened 2017-05-30
Fix memoize with more than 1 arguments

angeloocana:
After 0.24.0 calling memoize with 2 or more arguments stoped working.

issue: #2183

To fix it, we call **toString** with **arguments**.


davidchambers:
Please remove __yarn.lock__ and revert all the formatting changes.

angeloocana:
@davidchambers  vs code automatically added the spaces, now its fixed.

what about yarn.lock should I add it to .gitignore?

I saw a lot of projects with yarn.lock, this keeps everybody with same versions.

davidchambers:
Please remove __yarn.lock__. It's useful for applications but doesn't make much sense for libraries.

angeloocana:
ok

PR #2187, opened 2017-05-30
Fix memoize 2 arguments

angeloocana:
After 0.24.0 calling memoize with 2 or more arguments stopped working.

issue: #2183

To fix it, we call **toString** with **arguments**:

Before:
```js
memoizeWith(toString);
```

After:
```js
memoizeWith(function() {
  return toString(arguments);
});
```


davidchambers:
It should be possible to update an existing pull request rather than creating a new one, @angeloocana. ;)

angeloocana:
@davidchambers it would be awesome. But it's great to keep history, immutable rocks =D
And I've learned a lot, I will not repeat the same mistakes anymore. 

Thanks!

kedashoe:
Looks good to me. Anyone else? @angeloocana , could you squash and update the commit message?

angeloocana:
Sorry @kedashoe , but I didn't understand what do you want me to do.
I'm still learning English.
Could you be more specific?

buzzdecafe:
FWIW the tests show that the wrapped function works as expected, but do not show that the evaluation is only performed once.

angeloocana:
You are right @buzzdecafe !

I couldn't think of a simple example, so I got my real code that was broken after 0.24.0.
If is it fine, I'll create a new PR.

Feel free to judge my code!
Jump over jump is the hardest project of my life, rewriting it with functional programming is helping me a lot.

Thanks!

buzzdecafe:
i think that is too much to demonstrate the effect. Maybe something like:

```js

var called = 0;

const f = (x, y) => {
  called += 1;
  return x + y;
};

const mf = memoize(f);

var _ = mf(10, 20);
eq(called, 1);
var _ = mf(10, 20);
eq(called, 1);
var _ = mf(20, 10)
eq(called, 2);
```


```

**Comment by angeloocana** [comment]
Done!

Anything else?

Should I create a new PR?

**Comment by buzzdecafe** [comment]
Now you should prove that the 2 argument case produces the correct output on each call. that oughta be all, though

**Comment by angeloocana** [comment]
@buzzdecafe I don't know if I fully understood your last comment.

Am I supposed to test it with strings, arrays, objs...?
Or Am I supposed to test the return values? If so, I already did it.

On your sample I noticed that the return values was going to ***var _ =*** and it was not been used anywhere.
And I had to rewrite it a little bit, because arrow functions are not allowed.

Code commited:
```js
it('works with 2 arguments', function() {
    var called = 0;

    var add = R.memoize(function(x, y) {
      called += 1;
      return x + y;
    });

    eq(add(1, 2), 3);
    eq(called, 1);

    eq(add(1, 2), 3);
    eq(called, 1);

    eq(add(1, 3), 4);
    eq(called, 2);
  });
```

buzzdecafe:
LGTM
:cow2:

kedashoe:
> LGTM

:+1: , @angeloocana , would you like to squash your commits and update the commit message, or is it ok if one of us does it?

angeloocana:
@kedashoe  please help me! I'm not good at English and this is my first PR.

kedashoe:
Sure!

> squash your commits

Right now for this PR we have 9 commits but it should really be just 1. In git, to go from many commits (9 in our case) to fewer commits (1 in our case) is commonly known as "sqaushing". You can do this in git using rebase. See https://www.google.com/search?q=git+squash+commits

> update the commit message

As part of this squashing process, instead of the 9 commit message we currently have, we would write a single commit message.

The github UI also allows us to do this, so if you would rather not do it, just let us know and we would be happy to :)

angeloocana:
I did the squash and It changed my local history, but when I pushed it git not allowed without a merge.

Should I Fork again and start it from scratch?

Fell free to update the commit message and any thing else needed.

kedashoe:
@angeloocana Thanks for reporting the issue and the PR

buzzdecafe:
... and congratulations on your first PR

angeloocana:
Thank you all!!!

I learned a lot, and I'm rethinking a lot of things to be more minimalist.
Thank you!

PR #2188, opened 2017-06-01
Transducer doc was fixed

owanturist:
Hello! Thanks for great library.

When I read the documentation I have found one mistake which confused me. It relative with a transducer signature. Documentation says that signature of the [transducer](http://ramdajs.com/docs/#transduce) looks like `(c → c) → (a,b → a) → a → [b] → a` but below the doc says
> The iteration is performed with R.reduce after initializing the transducer.

I think that signature of the transducer should be looks like `(c → c) → ((a, b) → a) → a → [b] → a`.

### DIFF
--- a/src/transduce.js
+++ b/src/transduce.js
@@ -31,7 +31,7 @@ var curryN = require('./curryN');
  * @memberOf R
  * @since v0.12.0
  * @category List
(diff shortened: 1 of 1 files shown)


buzzdecafe:
Yes! much better, thank you

:cow2:

Issue #2093, opened 2017-02-20
Inconsistency between lensIndex/lensPath


joaomilho:
Just found this issue:
```js
  R.view(R.lensIndex(-1), ['a','b','c']) // => 'c'
  R.view(R.lensPath([-1]), ['a','b','c']) // => undefined 
```

CrossEye:
This looks like it would be fixed by a suggestion made in https://github.com/ramda/ramda/issues/1988#issuecomment-264376144, but not followed up.

@scott-christopher, any objections to this?

scott-christopher:
The only concern I have is if we change the implementation of `lensPath` to match `lensIndex`, then it will become inconsistent with the behaviour of `path` and `assocPath`.

Could we instead update the implementation of `path` and `assocPath` to support negative indexes?

CrossEye:
Yes, I think that would make sense.  I'll try to look into it soon.

keithbro:
Hi, hope you don't mind me jumping on to help on this one. I was looking for an issue to help with and this one seemed straight forward. I've raised a PR, but I'm not sure that it solves the original issue.

CrossEye:
Very sorry.  At some point I'll find the time to make a new release.

PR #2189, opened 2017-06-03
`path` and `assocPath` to support negative indexes

keithbro:
As required for #2093.


src/assocPath.js
```diff
@@ -41,7 +43,8 @@ module.exports = _curry3(function assocPath(path, val, obj) {
   }
   if (_isInteger(idx) && _isArray(obj)) {
     var arr = [].concat(obj);
-    arr[idx] = val;
+    var arrIdx = idx < 0 ? arr.length + idx : idx;
```
keithbro: I'm wondering if it would be nicer if this logic was handled by `assoc`?

test/assocPath.js
```diff
@@ -49,4 +49,9 @@ describe('assocPath', function() {
     eq(assocPath(['foo', 'bar', 'baz'], 42, {foo: null}), {foo: {bar: {baz: 42}}});
   });
 
+  it('supports array indices', function() {
+    eq(assocPath([0], 42, [1, 2, 3]), [42, 2, 3]);
+    eq(assocPath([-1], 42, [1, 2, 3]), [1, 2, 42]);
```
keithbro: it's not clear to me what the expected result of 

    assocPath([99], 42, [1, 2, 3]);

should be?

src/assocPath.js
kedashoe: I think it is the other way around? `assoc` can be easily defined in terms of `assocPath`. They have historically had divergent behavior with certain input, which is why this hasn't happened yet. Maybe it is time we do that, as this PR as it stands would be yet another difference.

src/assocPath.js
```diff
@@ -29,6 +29,8 @@ var isNil = require('./isNil');
  *
  *      // Any missing or non-object keys in path will be overridden
  *      R.assocPath(['a', 'b', 'c'], 42, {a: 5}); //=> {a: {b: {c: 42}}}
+ *
+ *      R.assocPath([-1], 42, [1,2,3]); //=> [1,2,42]
  */
 module.exports = _curry3(function assocPath(path, val, obj) {
   if (path.length === 0) {
```
kedashoe: github won't let me comment on a line that isn't part of the diff, but I think we need some logic here to handle multiple negative indexes (eg `assocPath([-1, -1], 10, [[1, 2, 3], [4, 5, 6]])`):

https://github.com/keithbro/ramda/blob/d8bb8c71e70ca26d74aec1ffeafd4e6eabc80b17/src/assocPath.js#L40-L43

let's add a test for this as well

test/assocPath.js
kedashoe: I assume whatever js would normally do here is fine.. create a sparse array?

src/assocPath.js
keithbro: I'm not sure. I feel like the logic of associating a value to an index of a flat array or a property given an flat object is sufficiently complex to warrant it's own function (`assoc`). `assocPath` could use that function when `path.length === 1`. If we put all the logic in `assocPath` then I think we are going to end up with quite a complex function. But perhaps I'm missing a simple solution. 

I'll push up a solution I think is good and you can let me know what you think.

test/assocPath.js
buzzdecafe: > I assume whatever js would normally do here is fine.. create a sparse array?

😱

keithbro: For positive indices, sparse makes sense.. but for negative it gets confusing. In Perl it throws and error which seems sane:

    $ perl -e '[]->[-1] = 1'
    Modification of non-creatable array value attempted, subscript -1 at -e line 1.

src/assocPath.js
kedashoe: Sorry for the delay on this one @keithbro . I'd prefer `assoc` to be `function assoc(prop, val, obj) { return assocPath([prop], val, obj); }` and for all the logic to be in `assocPath`. But let's see what others think.. anyone?

src/assoc.js
```diff
@@ -22,6 +23,16 @@ var _curry3 = require('./internal/_curry3');
  *      R.assoc('c', 3, {a: 1, b: 2}); //=> {a: 1, b: 2, c: 3}
  */
 module.exports = _curry3(function assoc(prop, val, obj) {
+  if (_isArray(obj)) {
+    var newArr = [].concat(obj);
+    if (prop < 0 && Math.abs(prop) > newArr.length) {
+      throw new Error('Invalid index');
```
PopeFelix: You might change this to "index out of range"

davidchambers: Also, this should be a [RangeError][1].


[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RangeError

CrossEye:
While there is definitely some work to do in figuring out the relationships between our `prop` and `path` functions and in our handling of negative indices, this has fallen behind some other work.  We'll address the underlying issue post `v1.0`.

PR #2190, opened 2017-06-05
Update examples of lift and liftN

luislee818:
* Use samples better reflect the calculation of Cartesian product

This PR comes from a suggestion from Scott Sauyet in [one discussion on StackOverflow](https://stackoverflow.com/questions/36558598/cant-wrap-my-head-around-lift-in-ramda-js/44306719?noredirect=1#comment75702132_44306719).


src/lift.js
```diff
@@ -18,11 +18,11 @@ var liftN = require('./liftN');
  *
  *      var madd3 = R.lift((a, b, c) => a + b + c);
  *
- *      madd3([1,2,3], [1,2,3], [1]); //=> [3, 4, 5, 4, 5, 6, 5, 6, 7]
+ *      madd3([100, 200], [30, 40, 50], [6, 7]); //=> [136, 137, 146, 147, 156, 157, 236, 237, 246, 247, 256, 257]
```
kedashoe: Maybe a comment that `lift` computes the cartesian product for arrays in ramda?

src/lift.js
```diff
@@ -18,11 +18,11 @@ var liftN = require('./liftN');
...
  *
  *      var madd5 = R.lift((a, b, c, d, e) => a + b + c + d + e);
  *
- *      madd5([1,2], [3], [4, 5], [6], [7, 8]); //=> [21, 22, 22, 23, 22, 23, 23, 24]
+ *      madd5([10000, 20000], [3000], [400, 500], [60], [7, 8]); //=> [13467, 13468, 13567, 13568, 23467, 23468, 23567, 23568]
```
kedashoe: We have `sanctuary` in the repl, how about something like

`madd3(S.Just(1), S.Just(2), Just(3)); //=> Just(6)` as a second example rather than `madd5`

luislee818: Good idea! I updated PR for this.

src/lift.js
```diff
@@ -18,11 +18,11 @@ var liftN = require('./liftN');
  *
  *      var madd3 = R.lift((a, b, c) => a + b + c);
  *
- *      madd3([1,2,3], [1,2,3], [1]); //=> [3, 4, 5, 4, 5, 6, 5, 6, 7]
+ *      // computes Cartesian product for Arrays
+ *      madd3([100, 200], [30, 40, 50], [6, 7]); //=> [136, 137, 146, 147, 156, 157, 236, 237, 246, 247, 256, 257]
```
buzzdecafe: sorry, but  i disagree with @kedashoe here. The Cartesian product would produce a set of ordered tuples (triples in this example): `(x <- X, y <- Y, z <- Z)`.  `lift` is not doing that.

luislee818: Thanks for the comment @buzzdecafe, I think I put _Cartesian product_ in the comment in the first place. Do you have any suggestions for the explanatory comment?

buzzdecafe: maybe something like :

"In Array context, the `lift`-ed function applies over the Cartesian product of the argument Arrays."

?

CrossEye:
Eventually handled in #3109

PR #2191, opened 2017-06-05
optimize mergeDeepWith comment

adispring:


kedashoe:
I think I prefer the current wording. The sentence does though use the term `value` a lot.. maybe

>otherwise the provided function is applied to associated values using the ~~resulting value~~ result as the new value associated with the key

PR #2194, opened 2017-06-06
optimize innerJoin & intersection time complexity with memoize

adispring:


kedashoe:
I think this would actually slow things down in many cases. `memoize` uses `toString` which can be very slow. If a user finds that one of these functions is slowing them down, they could always use `memoizeWith` themselves.

adispring:
ok, I'll close this pr.

Issue #2161, opened 2017-04-21
Dependencies


CrossEye:
Does anyone else find this a bit disconcerting?

https://goo.gl/coH47H

kedashoe:
That `toString` is blowing up the dependency graph for `concat`? Do you think we should get rid of it and display a less helpful error message? I don't currently have a strong opinion one way or the other.

CrossEye:
LOL.  I hadn't chased it down at all, just saw how large and convoluted the whole graph was.

This is **much** better:

    https://goo.gl/rMed9O

I do think we need to consider removing `toString` here.  Using individual functions or custom builds becomes very bloated.  But I'm glad to realize that the complexity is all around that and not spread throughout.

mattgrande:
Hey @CrossEye - How did you generate the mermaid syntax?

CrossEye:
Manually.  I thought it was going to be just a few nodes... and then I got stubborn!  ;-)

I'm sure it would be easy to hook together a dependency tree walker, but this was a one-off.

davidchambers:
#1938 is the solution, I think. :)

By adding one simple, well-tested dependency we can remove vast swathes of internal Ramda code, including `_equals` and `_toString`.

CrossEye:
> By adding one simple, well-tested dependency we can remove vast swathes of internal Ramda code, including `_equals` and `_toString`.

That only helps if the code in sanctuary-type-classes is significantly simpler than that in Ramda.  My concern was about the function-level dependencies.  I don't care for these purposes what library contains those functions.

While I'm quite happy with Ramda's lack of external dependencies, I'm by no means wedded to it.  But if we're going to add one, we must be keeping in mind the convenience of the *users* of the library.  One of the issues that comes up regularly is the ability to import only what you want from Ramda, either through `const concat = require('ramda/src/concat)` (or es6 import) or through a custom build.  If building any Ramda function then required importing an entire second library, I think we may be doing a disservice to our users.

davidchambers:
Ramda is not currently suitable for users who wish to import one or two functions without bringing in many bytes. If a user wishes to use `R.map`, but no other Ramda functions, she must still bring in:

  - the various `fantasy-land/map` implementations for built-in types;
  - the machinery for dispatching to `fantasy-land/map` and `map`;
  - the machinery for transducers; and
  - the machinery for currying (including support for placeholders).

[...]

PR #2195, opened 2017-06-06
Improve performance of omit

tomhicks:
Rather than iterating the names to omit for every single property name,
which is potentially O(n*m) complexity, create a map prior to iterating
the source object, which we can then look up in O(1) time for each
property name.

This makes the omit function O(n+m) rather than O(n*m).

---

Resubmitted from #2192

### DIFF
--- a/src/omit.js
+++ b/src/omit.js
@@ -1,6 +1,4 @@
-var _contains = require('./internal/_contains');
 var _curry2 = require('./internal/_curry2');
-
 
 /**
  * Returns a partial copy of an object omitting the keys specified.
@@ -20,8 +18,17 @@ var _curry2 = require('./internal/_curry2');
  */
 module.exports = _curry2(function omit(names, obj) {
   var result = {};
+  var index = {};
+  var idx = 0;
+  var len = names.length;
+
+  while (idx < len) {
+    index[names[idx]] = 1;
+    idx += 1;
+  }
+
   for (var prop in obj) {
-    if (!_contains(prop, names)) {
+    if (!index.hasOwnProperty(prop)) {
       result[prop] = obj[prop];
     }
   }
--- a/test/indexBy.js
+++ b/test/indexBy.js
@@ -28,7 +28,7 @@ describe('indexBy', function() {
       R.indexBy(R.prop('id')),
       R.map(R.pipe(
         R.adjust(R.toUpper, 0),
-        R.adjust(R.omit('id'), 1)
+        R.adjust(R.omit(['id']), 1)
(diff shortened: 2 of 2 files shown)


kedashoe:
:+1:

buzzdecafe:
LGTM
:cow2:

svozza:
Rather than bringing in internal dependencies like `indexBy` and `identity`, would it not be easier just to create the lookup object with a simple loop?

tomhicks-bsf:
a) depends what you mean by "easier". The previous implementation used `_contains` when it could have done its own `indexOf` logic. I presumed that there is a desire to use internal implementations rather than rolling your own?
b) I don't think I'm using anything undocumented. `indexBy(R.prop("length"), ["a", "ab"])` is still turning "a list of objects into an object indexing the objects by the given key". The key is not explicitly written out longhand, but it's still there.

But whatever. I just want to get rid of the O(n*m) complexity.

svozza:
a) No, not necessarily because it increases bundle sizes of [partial builds](https://github.com/ramda/ramda#partial-builds). There's a few issues and PRs around this topic: #2161 https://github.com/ramda/ramda/issues/1709#issuecomment-205089684 #879
b) When I say undocumented, I mean that the use violates the type signature in documentation:

`(a → String) → [{k: v}] → {k: {k: v}}`

tomhicks-bsf:
a) OK, fair enough.
b) Then I'd suggest that the documentation or the explanation or the implementation is all a bit mixed up. If it literally meant that it indexes by an actual property, then instead of a function as the first argument, it should take a property name.

kedashoe:
I don't mind either way. Anyone have a stance on requiring indexBy vs creating a lookup map?

tomhicks:
There we go. No dependencies other that `_curry`

src/omit.js
```diff
@@ -20,8 +18,17 @@ var _curry2 = require('./internal/_curry2');
  */
 module.exports = _curry2(function omit(names, obj) {
   var result = {};
+  var index = {};
+  var i = 0;
```
kedashoe: Could you please change `i` to `idx` (ramda convention)? Besides that, lgtm

tomhicks:
changed `i` to `idx` as requested.

kedashoe:
:+1:

buzzdecafe:
thanks

asaf-romano:
For the average case, wouldn't it be better to shallow-clone the original object (which uses native Object.assign) and then ``delete`` the given properties?

PR #2197, opened 2017-06-07
fix: o.js example lastName -> name.last

adispring:


buzzdecafe:
whoops, my bad. Thanks for fixing it

:cow2:

Issue #2111, opened 2017-03-03
nthOr headOr


jethrolarson:
```
nthOr :: a -> Int -> [a] -> a
headOr :: a -> [a] -> a
```
Gives safe way of extracting a value from a list that can be empty, similar to `propOr`

CrossEye:
An interesting alternative to introducing `Maybe`  into Ramda proper.

I'd have to think about it a bit.  I would worry that such things would proliferate and obviously they are easy to compose on ones own.  But they are convenient.

jethrolarson:
Yeah, as much as I like sanctuary, this tactic that ramda has adopted of requiring defaults for safe functions is a decent idea that I think is more approachable than Maybes.

Obviously it has it's own pitfalls:
```js
const getStuff = propOr([], 'stuff')
// surprise impurity due to the mutability of arrays
```

jethrolarson:
For the record, I'm still in support of ramda getting `Maybe`

adispring:
We can compose some ramda 'get' API with `R.defaultTo`:

```js
const nthOr = R.curry((defaultVal, list) => R.compose(
  R.defaultTo(defaultVal),
  R.nth
)(list));
```

It's convenient, but too many api will need to write xxOr version. So I also support import `data.Maybe`, or compose APIs to do that by ourselves.

Jacse:
@adispring alternatively we should create a function called R.or with a type signature of a -> (b -> a) -> b -> a

That way we could do: R.or([], R.prop('children')) and thus the user could wrap all existing functions easily without having to compose. Not sure if regular composition is preferable though.

On mobile so sorry for the poor formatting, and apologies for the probably wrong type signature, I hope you catch my drift.

jethrolarson:
There's already R.or and R.either with different signatures

buzzdecafe:
> For the record, I'm still in support of ramda getting `Maybe`

... bumping this ...

adispring:
@Jacse ,
```
const default = R.curry((defaultVal, func) => R.compose(
  R.defaultTo(defaultVal),
  func,
));
```

CrossEye:
@adispring: how about `addDefault` in parallel with `addIndex`?

adispring:
ok, I'll make a pr with api named `addDefault`.

davidchambers:
@adispring, you seem to be ignoring the existence of `undefined`.

```javascript
> S.maybe('EMPTY', S.toString, S.head([]))
'EMPTY'

[...]

PR #2198, opened 2017-06-07
Add one api: R.addDefault

adispring:
`addDefault` is used to add one default returned value to function. If function's result is `null`, `undefined` or `NaN`, the given default value will be returned.

Discuss link: https://github.com/ramda/ramda/issues/2111

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,6 +4,7 @@ module.exports = {
   __: require('./src/__'),
   add: require('./src/add'),
   addIndex: require('./src/addIndex'),
+  addDefault: require('./src/addDefault'),
   adjust: require('./src/adjust'),
   all: require('./src/all'),
   allPass: require('./src/allPass'),
--- a/src/addDefault.js
+++ b/src/addDefault.js
@@ -4,9 +4,9 @@ var defaultTo = require('./defaultTo');
 
 
 /**
- * Creates a new function(`'fn`) from an existing one(`fn`) by adding a default returned
+ * Creates a new function(`fn'`) from an existing one(`fn`) by adding a default returned
  * value(`defaultVal`). If the result of `fn` is not `null`, `undefined` or `NaN`, it
- * will be returned by `'fn`; otherwise, `defaultVal` will be returned.
+ * will be returned by `fn'`; otherwise, `defaultVal` will be returned.
  *
  * @func
  * @memberOf R
--- a/src/addDefault.js
+++ b/src/addDefault.js
@@ -1,0 +1,32 @@
+var _curry2 = require('./internal/_curry2');
+var compose = require('./compose');
+var defaultTo = require('./defaultTo');
+
+
+/**
+ * Creates a new function(`'fn`) from an existing one(`fn`) by adding a default returned
+ * value(`defaultVal`). If the result of `fn` is not `null`, `undefined` or `NaN`, it
+ * will be returned by `'fn`; otherwise, `defaultVal` will be returned.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @category List
+ * @sig b -> (a -> b) -> (a -> b)
+ * @param {b} defaultVal `defaultVal` will be returned ultimately if `fn`'s result is not `null`, `undefined` or `NaN`.
+ * @param {Function} fn original function
+ * @return {Function} function with a default returned value
+ * @example
+ *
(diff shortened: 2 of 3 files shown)


CrossEye:
Sorry, I don't remember ever seeing this, although I participated in the discussion leading to it.  I'm reopening, because it seems like a reasonable idea.

adispring:
Ok，If the core team support this pr，I'll resolve the conflict, and put the `addDefault` module to proper directory.

CrossEye:
I think this would be a good addition?  Anyone else?  @ramda/core?

semmel:
**Pro:**
- Currying might come handy:
```javascript
var returnsDefaultPhone = R.addDefault(getReceptionPhone());
var findPhoneByName = returnsDefaultPhone(getPhoneFromName));
var findPhoneByRoomNr = returnsDefaultPhone(getPhoneFromRoomNr));
var call = R.compose(callNumber, findPhoneByName);
call('Max');
```
**Cons:**
- API bloat. (It's just`R.compose(R.defaultTo(value), fn)`. Which should be simple for Ramda users to write`(*)`
- For me the name suggests some kind of addition (numeric or size) which I find misleading. What about naming it `returnDefault` ?

`(*)` - provided they know about `R.defaultTo` which I haven't for all the years using Ramda 😳

Bradcomp:
> provided they know about R.defaultTo which I haven't for all the years using Ramda

😆 

I agree that it's questionable whether this is necessary. I don't know how often I would reach for it. If we move forward, what about something like `withDefault` for the name?

CrossEye:
The only advantage of `addDefeult` as a name is the parallel with `addIndex`.  It's not at all necessary, either the name or the function.  I'm mixed about it.  There is a reasonable argument to be made, but it doesn't seem essential.

adispring:
> **Cons:**
> 
> * API bloat. (It's just`R.compose(R.defaultTo(value), fn)`. Which should be simple for Ramda users to write`(*)`

I have encountered some situations that let me write this:
```js
R.compose(R.defaultTo(value), fn)
// or
const maybeNil = fn(...args)
const result = R.defaultTo(maybeNil)
``` 
again and again，if the `fn` is other functions instead of `path` or `prop`, both of which have `***Or` version. 

`addDefault`/`withDefault` maybe alleviate this situation.

CrossEye:
We never did finish this conversation.  What do you think, @adispring, @semmel, @Bradcomp, @customcommander?

customcommander:
I like decorators too. Not strong feeling for or against this new function. I tend to use the `with` prefix for decorators, e.g. `withDefault(42)(fn)`. Do we have any established naming convention for those?

semmel:
> We never did finish this conversation. What do you think, @adispring, @semmel, @Bradcomp, @customcommander?

It *fits well* with Ramda's notion of an "absent/missing" value. Many Ramda functions (`propOr`, `defaultTo`, ...) interpret `null`, `undefined` and `NaN` as such an "absent/missing" value. This is *consistent* with many JavaScript APIs which return one of those to indicate something is missing. `document.querySelector('impossible-element') // -> null` comes into my mind first.

So I consider this beneficial for wrapping third-party APIs.

However,  @davidchambers pointed out in [his comment](https://github.com/ramda/ramda/issues/2111#issuecomment-294781435), that an "absent/missing" value is much better represented as a `Nothing`. 
Unfortunately the discussion in #2111 got quickly bikeshedded into a discussion about sparse arrays. 

If in *your own code*, you sometimes convey a missing item with `null`, `undefined`,  and sometimes with a `Nothing` you'll make reasoning about your code more difficult. Although using `Maybe`s is not everyones cup of tea I cherish *ramda/ramda-fantasy's* [`Maybe.maybe`](https://github.com/ramda/ramda-fantasy/blob/master/docs/Maybe.md#maybemaybe) which does about the same thing as this PR.

So on one hand Ramda features functions (e.g. `ap`, `promap`, ...) which make few sense without *FantasyLand types*, while on the other hand Ramda encourages working with `null`, `undefined`. Covering such a very wide area of application, it could get hard for Ramda to keep the balancing act.

TLDR: Right now I would not use this function.

CrossEye:
I think there isn't enough strong support to include this.  Closing for now.

adispring:
OK, I think composing `R.defaultTo` and other api that need default value is enough.

PR #2200, opened 2017-06-09
Improve performance of clone

hsalkaline:
Instead of creating copy function on each _clone call, create it once.

### DIFF
--- a/src/internal/_clone.js
+++ b/src/internal/_clone.js
@@ -1,19 +1,8 @@
 var _cloneRegExp = require('./_cloneRegExp');
 var type = require('../type');
 
-
-/**
- * Copies an object.
- *
- * @private
- * @param {*} value The value to be copied
- * @param {Array} refFrom Array containing the source references
- * @param {Array} refTo Array containing the copied source references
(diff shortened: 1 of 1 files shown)


hsalkaline:
I have unintentionally discovered that ramda's clone noticeably slower than the one from lodash. https://jsperf.com/region-clone-original/

One simple thing to improve in current clone implementation is to create `copy` function once, not every time `_clone` is called.
Нere is the same benchmark as above, but with ramda built from code of this PR: https://jsperf.com/regions-clone/

buzzdecafe:
i see how your implementation should be an improvement, but I don't see any actual difference when I run the benches on the two versions. Testing in Chrome 58.0.3029 / Mac OS X 10.11.6

hsalkaline:
I ran benchmark in chrome-based browser and got theese numbers:
https://yadi.sk/i/yl_ZfXp23JzY7Q and https://yadi.sk/i/wVmsvJg13JzY8q
However, in Chrome 58.0.3029 I also can't see any difference.
Maybe it's a bug in Yandex.Browser or maybe newer version of js engine became smart enough to optimize this case.

PR #2203, opened 2017-06-12
Reduce method example

ivanzusko:
While opening REPL from the documentation page we are getting this:

image

I offer to add extra `//` on each line of the explanation so that will not break REPL example:

image


kedashoe:
:+1:

PR #2205, opened 2017-06-12
There was a problem with the ```code tag```

archae0pteryx:
I think i corrected the issue. There was a break cutting off half of the code because of a newline somewhere in the middle of the snippet. I deleted the newline and adjusted the ``` at the end. I still can't get your bookmarklet to work correctly however... That'll be another issue for another time however.


kedashoe:
Anyone who knows anything about bookmarklets can confirm?

buzzdecafe:
makes sense to me; somehow a `\n` got introduced, and that broke the script.

buzzdecafe:
dammit i merged the duplicate by mistake

PR #2212, opened 2017-06-16
Add example of how R.type treats `undefined`

chriddyp:
I find myself looking at the source for this one (e.g. "is `undefined` an "Object" or "Null" or something else?"). Perhaps controversial since we never really do assignments of `undefined`


buzzdecafe:
:cow2:
thanks

PR #2224, opened 2017-07-01
add links between propEq and whereEq, also between propSatisfies and …

adispring:
…where


src/where.js
```diff
@@ -20,6 +20,7 @@ var _has = require('./internal/_has');
  * @param {Object} spec
  * @param {Object} testObj
  * @return {Boolean}
+ * @see R.propSatisfies
```
buzzdecafe: should probably point to whereEq here as well

buzzdecafe:
Thank you!
:cow2:

PR #2229, opened 2017-07-06
Add transduce support for tap

wennergr:
Tap will act as tranducer if given a transformer in the second parameter

### DIFF
--- a/src/internal/_xtap.js
+++ b/src/internal/_xtap.js
@@ -1,0 +1,18 @@
+var _curry2 = require('./_curry2');
+var _xfBase = require('./_xfBase');
+
+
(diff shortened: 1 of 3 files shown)


buzzdecafe:
I can see how this would be very useful

buzzdecafe:
:cow2:

lax4mike:
@wennergr This change breaks the following code:
```
R.compose(
  R.tap(console.log)
)(null)

// "Cannot read property '@@transducer/step' of null"
```

is this intended?

CrossEye:
We really need to get a new release out.  This was fixed in 37da986.

Issue #2227, opened 2017-07-04
Bug: Equals acts wrong on ES6 Sets


Jabher:
Set is an abstract structure which should not rely upon elements order. However, Ramda do not follow this spec.
E.g.
```javascript
R.equals(
  new Set([2,1]),
  new Set([1,2])
) //=> false
```
This is misbehavior; this two sets are actually equal. I clearly understand why this is happening; but, however, I'm definitely sure this should be fixed.

kedashoe:
Yep. Using `entries` also doesn't make any sense. Would you like to submit a PR?

Jabher:
My pleasure!

jadengore:
@Jabher This should resolve issues with calling `R.isEmpty` on a set as well correct? I believe it uses the `equals` function.

kedashoe:
closed #2232

PR #2232, opened 2017-07-10
fixing Set inequality - https://github.com/ramda/ramda/issues/2227

Jabher:
I do not know how to implement it properly in ramda-way - I'm unsure about this `stackA, stackB` thing.

But it works, and I'm ready to re-write if you can provide any good advice about how to do it better.

### DIFF
--- a/src/internal/_equals.js
+++ b/src/internal/_equals.js
@@ -1,114 +1,161 @@
 var _arrayFromIterator = require('./_arrayFromIterator');
+var _containsWith = require('./_containsWith');
 var _functionName = require('./_functionName');
 var _has = require('./_has');
 var identical = require('../identical');
 var keys = require('../keys');
 var type = require('../type');
 
+module.exports = (function() {
+  /**
+   * private _uniqContentEquals function.
+   * That function is checking equality of 2 iterator contents with 2 assumptions
+   * - iterators lengths are the same
+   * - iterators values are unique
+   *
+   * false-positive result will be returned for comparision of, e.g.
+   * - [1,2,3] and [1,2,3,4]
+   * - [1,1,1] and [1,2,3]
+   * */
 
-module.exports = function _equals(a, b, stackA, stackB) {
-  if (identical(a, b)) {
-    return true;
+  function _uniqContentEquals(aIterator, bIterator, stackA, stackB) {
+    var a = _arrayFromIterator(aIterator);
+    var b = _arrayFromIterator(bIterator);
+
+    function eq(_a, _b) {
+      return _equals(_a, _b, stackA.slice(), stackB.slice());
+    }
+
+    // if *a* array contains any element that is not included in *b*
+    return !_containsWith(function(b, aItem) {
+      return !_containsWith(eq, aItem, b);
+    }, b, a);
   }
 
-  if (type(a) !== type(b)) {
-    return false;
-  }
+  function _equals(a, b, stackA, stackB) {
+    if (identical(a, b)) {
+      return true;
+    }
 
-  if (a == null || b == null) {
-    return false;
-  }
+    var typeA = type(a);
 
-  if (typeof a['fantasy-land/equals'] === 'function' || typeof b['fantasy-land/equals'] === 'function') {
-    return typeof a['fantasy-land/equals'] === 'function' && a['fantasy-land/equals'](b) &&
-           typeof b['fantasy-land/equals'] === 'function' && b['fantasy-land/equals'](a);
-  }
+    if (typeA !== type(b)) {
+      return false;
+    }
 
-  if (typeof a.equals === 'function' || typeof b.equals === 'function') {
-    return typeof a.equals === 'function' && a.equals(b) &&
-           typeof b.equals === 'function' && b.equals(a);
-  }
+    if (a == null || b == null) {
+      return false;
+    }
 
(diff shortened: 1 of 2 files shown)


Bradcomp:
I think it would be worthwhile to handle the `Map` case at the same time. Right now it suffers from the same problem `Set` does... https://goo.gl/8yb45f

Jabher:
done. However, I'm pretty sure something had to be done with this stacks, I'm really pretty sure I did something wrong with them

kedashoe:
> I'm pretty sure something had to be done with this stacks

The stacks help detect/test recursive data structures. If we want to do this we sets we need a slightly different algorithm as order matters for the current one. eg

```js
let a = [];
let b = [];
let c = [a];
let d = [b];
c.push(c);
d.unshift(d);
a.push(c);
b.push(d);
console.log(R.equals(a, b));
```
logs "false", whereas if the arrays were Sets I believe this should be true?

Also I think once we have the array from the Map/Set iterator, the algorithm is the same, can we create a local `unorderedEquals` function for this?

Jabher:
Guys, I've implemented stack support - please have a look. It's working, but I really do not like the fact that stack equality check is called 3 times here - code can be possibly decomposed, but I did not plan to fully re-write the function :)

src/internal/_equals.js
```diff
@@ -1,114 +1,196 @@
...
 var _has = require('./_has');
 var identical = require('../identical');
 var keys = require('../keys');
 var type = require('../type');
 
+module.exports = (function() {
+  /**
+   * private _uniqContentEquals function
+   * Notes:
+   * 1. this function will return true on [1,2,3] === [1,1,1]
```
kedashoe: Do you mean something like `true on [1,2,3] === [1,1,1,2,3]`?

src/internal/_equals.js
```diff
@@ -1,114 +1,196 @@
...
+   * 2. this function do not expect iterables of different size
+   * it works as a shared function for ES6 comparators
```
kedashoe: not sure what this line means? "it works as a shared function for ES6 comparators"?

src/internal/_equals.js
```diff
@@ -1,114 +1,196 @@
...
-  }
+        var stackEquals = _stackEquals(a, b, stackA, stackB);
+        if (stackEquals !== undefined) {
+          return stackEquals;
+        }
+
+        stackA.push(a);
+        stackB.push(b);
+
+        // this optimization can be not needed, depends on use case.
```
kedashoe: `entries()` will contain the keys, so I think we are fine without this block, right?

src/internal/_equals.js
```diff
@@ -1,114 +1,196 @@
...
+
+        stackA.pop();
+        stackB.pop();
+        return true;
+      case 'Set':
+        if (a.size !== b.size) {
+          return false;
+        }
+
+        var idx = stackA.length - 1;
```
kedashoe: We don't need this here anymore since you are calling your new `_stackEquals` function below.

src/internal/_equals.js
Jabher: my bad - wrong order taken. 
```javascript
_uniqContentEquals([1,1,1], [1,2,3], [], []) // => true
```
I've changed the comment to make it more explanatory

src/internal/_equals.js
Jabher: I'm not sure. It is performance-related question.
This is the thing to discuss: on one hand, in case we're checking keys first, we're spending **n** time on equality check first, and then, if all keys are same, we're spending around **2n** time to validate KVs. So, if keys are same, we're spending **n** time on it, and **3n** otherwise. On other hand, if we're simply checking the KVs, we're spending around **2n** time anyway.

On other hand - if we're talking about the small Maps, we can consider that wrapping code working time is nearly same as inside code, and for big maps - the bigger the maps are, the less is the chance we're talking about 2 maps of same length and different keys (general collision case is "let's have 2 different maps for same values to store different whatever, cache, for example")

Jabher: So, yeah, this is much probably useless code, causing computation time to increase on big maps, and for small maps it is not influencing a lot. I'll remove it, thanks

Jabher:
done :)

src/internal/_equals.js
kedashoe: I see, thanks for the update, I see what is going on now.

src/internal/_equals.js
```diff
@@ -1,114 +1,161 @@
+var _containsWith = require('./_containsWith');
```
kedashoe: Can you move this below the next line? The requires need to be in alphabetical order..

buzzdecafe:
:cow2:

Jabher:
Guys, this is becoming weird :)

kedashoe:
> Guys, this is becoming weird :)

haha the :cow2: is a :+1: . Thanks for your work on this, could you please squash the commits and we'll merge it :tada:

buzzdecafe:
@kedashoe, you want the honors? you guided this one through

kedashoe:
thanks @Jabher !

Jabher:
🍻 
hope I'll find some time to fix other es6 collections issues here

CrossEye:
:herb:

PR #2233, opened 2017-07-11
updated type sigs for function parameters

Bradcomp:
I've updated the type signatures so as not to imply that certain functions take curried functions when they don't. I've also tried to normalize the spacing and paren placement on functions with 2+ arity functions parameters.


CrossEye:
Very nice!

PR #2236, opened 2017-07-11
Fix traverse doc

1ven:


src/traverse.js
```diff
@@ -9,7 +9,7 @@ var sequence = require('./sequence');
  * then uses [`sequence`](#sequence) to transform the resulting Traversable of Applicative
  * into an Applicative of Traversable.
  *
- * Dispatches to the `sequence` method of the third argument, if present.
+ * Dispatches to the `traverse` method of the third argument, if present.
```
davidchambers: Should this be `fantasy-land/traverse` rather than `traverse`?

1ven: @davidchambers I think no, as [map](https://github.com/ramda/ramda/blob/v0.24.1/src/map.js#L45) is dispatching both, `map` and `fantasy-land/map` but docs are mentioning just `map` method.
As I correctly understand, we need to dispatch `traverse` method along with `fantasy-land/traverse` like it's done in `map`

paldepind: The docs state that the function supports Fantasy Land traversables. I don't think it's necessary to mention that it dispatches to `fantasy-land/traverse`. Can't we consider that to be more of an implementation detail? Users should just have to think about the abstractions.

paldepind:
LGTM 👍

PR #2237, opened 2017-07-13
Remove Object.prototype.toString.call from map

rtivital:
`R.type` provides same functionality

### DIFF
--- a/src/map.js
+++ b/src/map.js
@@ -5,6 +5,7 @@ var _reduce = require('./internal/_reduce');
 var _xmap = require('./internal/_xmap');
 var curryN = require('./curryN');
 var keys = require('./keys');
+var type = require('./type');
 
 
 /**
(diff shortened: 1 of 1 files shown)


davidchambers:
I'd prefer to limit the number of internal dependencies. This will be useful if and when "tree shaking" becomes possible.

CrossEye:
> I'd prefer to limit the number of internal dependencies.

Agreed.

davidchambers:
Thanks for the pull request nonetheless, @rtivital. :)

buzzdecafe:
not reopening, but doesn't the dependency/tree-shaking argument cut both ways? i.e., if i include two functions and they both implement the same dependecnies themselves , haven't I duplicated the payload for no reason?

davidchambers:
> haven't I duplicated the payload for no reason?

I'd expect no duplication after gzipping.

buzzdecafe:
> I'd expect no duplication after gzipping.

doesn't that assume all the independent implementations are identical?

davidchambers:
What is duplicated?

  - `Object.prototype.toString.call`
  - `[object `
  - `]`

That's all we'd avoid by using `R.type` here. These character sequences are bound to occur elsewhere, so will gzip very well.

davidchambers:
I'm not sure we can generalize. For a non-trivial function I would favour reusing the Ramda function.

CrossEye:
> I'm not sure we can generalize. For a non-trivial function I would favour reusing the Ramda function.

Although it's hard to quantify, I agree that this is the right standard.

buzzdecafe:
I'll buy that

Issue #2243, opened 2017-07-29
Not working example in the doc of "sequence" method


barbiturat:
This example from here http://ramdajs.com/docs/#sequence is not working: 
`R.sequence(R.of, Maybe.Just([1, 2, 3]));`
In reality it returns [[]]

dustinws:
This is curious. All objects besides `Array` _have to implement sequence_ for that example to work, including the `Maybe` in question for this issue. Otherwise ramda treats the traversable as an array, which is why the output is `[[]]`.

I'm wondering if this breaks the `(Applicative f, Traversable t)` type constraint. Even if some custom type did implement `Traversable`, it's still going to get treated like an array unless it also has `sequence`.

Since traverse is implemented with sequence, this could easily be fixed by also checking for `fantasy-land/traverse` inside of `sequence` and just calling `traversable.traverse(identity, of)`.

I'm happy to submit a PR with a fix / tests if anyone is interested 😉

buzzdecafe:
I'd love to see a fix for this, thanks

dustinws:
Hey @buzzdecafe, I have a PR for this over [here](https://github.com/ramda/ramda/pull/2248). If you get the time to check it out, I'd really appreciate your feedback!

PR #2248, opened 2017-08-08
Make "Maybe" Traversable and fix sequence

dustinws:
This PR fixes the broken type constraint of `R.sequence`.

**Currently Broken**
R.sequence lists the type constraint `(Applicative f, Traversable t)` in it's type signature, but this breaks for custom data types.

*Example*
Assuming that `Id` implements `Traversable`, the following *is not true*
`R.sequence(R.of, Id([1, 2, 3])) === [Id(1), Id(2), Id(3)]`

This is because `R.sequence` only checks for `sequence` before deferring to `reduceRight` as shown [here](https://github.com/ramda/ramda/blob/master/src/sequence.js#L35).

**The Fix**
Perform an additional check for `fantasy-land/traverse` *after* checking for `sequence`. If the type *does implement fantasy-land/traverse*, sequence will defer to it.

To write the tests for this, I had to implement `Traversable` on something. I chose `Maybe` since it is what is used in the original issue that inspired me to create this PR, and `Id` already has `sequence`, so I wouldn't have been able to write *failing tests first*. To implement Traversable, I also added Foldable, but only for "correctness" since `reduce` isn't used in any tests.

This would fix the bug in #2243

### DIFF
--- a/src/sequence.js
+++ b/src/sequence.js
@@ -1,5 +1,6 @@
 var _curry2 = require('./internal/_curry2');
 var ap = require('./ap');
+var identity = require('./identity');
 var map = require('./map');
 var prepend = require('./prepend');
 var reduceRight = require('./reduceRight');
@@ -30,8 +31,12 @@ var reduceRight = require('./reduceRight');
  *      R.sequence(R.of, Nothing());       //=> [Nothing()]
  */
 module.exports = _curry2(function sequence(of, traversable) {
-  return typeof traversable.sequence === 'function' ?
-    traversable.sequence(of) :
+  if (typeof traversable.sequence === 'function') {
+    return traversable.sequence(of);
+  }
+
+  return typeof traversable['fantasy-land/traverse'] === 'function' ?
+    traversable['fantasy-land/traverse'](of, identity) :
     reduceRight(function(x, acc) { return ap(map(prepend, x), acc); },
                 of([]),
                 traversable);
--- a/test/shared/Maybe.js
+++ b/test/shared/Maybe.js
@@ -49,12 +49,12 @@ Maybe.prototype['fantasy-land/chain'] = function(f) {
   return this.isJust ? f(this.value) : Nothing;
 };
 
-// Maybe#reduce :: Maybe f => f a ~> ((b, a) -> b, b) -> b
+// Maybe#reduce :: Foldable f => f a ~> ((b, a) -> b, b) -> b
 Maybe.prototype['fantasy-land/reduce'] = function(f, acc) {
   return this.isJust ? f(acc, this.value) : acc;
 };
 
-// Maybe#traverse :: Applicative f, Maybe m => m (f a) ~> (f, a -> f b) -> f (m b)
+// Maybe#traverse :: Applicative f, Traversable t => t (f a) ~> (f, a -> f b) -> f (t b)
 Maybe.prototype['fantasy-land/traverse']  = function(of, f) {
   return this.isJust ? map(Maybe.Just, f(this.value)) : of(Nothing);
 };
--- a/test/shared/Maybe.js
+++ b/test/shared/Maybe.js
@@ -51,7 +51,7 @@ Maybe.prototype['fantasy-land/chain'] = function(f) {
 
 // Maybe#reduce :: Maybe f => f a ~> ((b, a) -> b, b) -> b
 Maybe.prototype['fantasy-land/reduce'] = function(f, acc) {
-  return this.isJust ? f(acc, this.value) : Nothing;
+  return this.isJust ? f(acc, this.value) : acc;
 };
 
 // Maybe#traverse :: Applicative f, Maybe m => m (f a) ~> (f, a -> f b) -> f (m b)
--- a/test/shared/Maybe.js
+++ b/test/shared/Maybe.js
@@ -1,6 +1,6 @@
 var equals = require('../../src/equals');
 var toString = require('../../src/toString');
-
+var map = require('../../src/map');
 
 var sentinel = {};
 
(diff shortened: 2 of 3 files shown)


test/shared/Maybe.js
```diff
@@ -49,6 +49,16 @@ Maybe.prototype['fantasy-land/chain'] = function(f) {
   return this.isJust ? f(this.value) : Nothing;
 };
 
+// Maybe#reduce :: Maybe f => f a ~> ((b, a) -> b, b) -> b
```
kedashoe: `:: Maybe f =>` Maybe is not a type class, you can just use `Foldable` here.

test/shared/Maybe.js
```diff
@@ -49,6 +49,16 @@ Maybe.prototype['fantasy-land/chain'] = function(f) {
...
+Maybe.prototype['fantasy-land/reduce'] = function(f, acc) {
+  return this.isJust ? f(acc, this.value) : acc;
+};
+
+// Maybe#traverse :: Applicative f, Maybe m => m (f a) ~> (f, a -> f b) -> f (m b)
```
kedashoe: Same as above with `Maybe` -> `Traversable`

dustinws:
@kedashoe Thanks for pointing out those type signatures! I'm still learning :)

I think dispatching to traverse makes a lot more sense, and stays true with the type signature. _Technically_ if everyone is following the constraints, it _shouldn't_ be a breaking change, but of course that isn't the reality.

I'm curious to see if anyone else has thoughts on it.

omugeri:
👍

src/sequence.js
```diff
@@ -30,8 +31,12 @@ var reduceRight = require('./reduceRight');
  *      R.sequence(R.of, Nothing());       //=> [Nothing()]
  */
 module.exports = _curry2(function sequence(of, traversable) {
-  return typeof traversable.sequence === 'function' ?
-    traversable.sequence(of) :
+  if (typeof traversable.sequence === 'function') {
+    return traversable.sequence(of);
+  }
```
davidchambers: I'm strongly in favour of not dispatching to `sequence`, which is vestigial.

:thumbsup: / :thumbsdown: ?

davidchambers:
@dustinws, do you happen to know which (old) version of the Fantasy Land spec Ramda supports?

dustinws:
@davidchambers I remember seeing that ramda only supports the original spec before FL introduced versioning. However, I can't find the source and I'm not positive that it's still the case.

kedashoe:
@dustinws since we're all in agreement and nobody else has chimed in, could you update the PR to

1. not check for `sequence`
2. check for `traverse` in addition to `fantasy-land/traverse`

dustinws:
@kedashoe Will do!

I do have a question for you or @davidchambers about something, though.
I noticed that in ramda, traverse takes it's arguments in the order of `of -> mappingFn -> traversable`

Notice in my pull request, I'm calling traverse like this, `traversable.traverse(of, identity)` to match up with the order that ramda uses. It looks like this is also how Fantasy Land treats `traverse`, where the `TypeRef` is the first parameter.

However, when `R.traverse` dispatches to fantasy-land/traverse, it switches the order of the arguments, which doesn't seem consistent with the FL spec, and is also counter intuitive for those expecting the signature to stay the same for custom data types (except the data argument, of course, ie. the traversable).

When I dispatch to traverse in `sequence`, (and also in the new traverse method on `Maybe`), I'm not switching the arguments.
Should I be? Or should `R.traverse` also be updated?

davidchambers:
Before we can answer these questions, @dustinws, we need to determine the versions of the Fantasy Land spec with which Ramda is intended to be compatible.

dustinws:
My thought was that newcomers would not be able to implement their own Traversables without looking at the source code for `R.traverse`.

This is because the current version of `Traversable` in FL and the type signature of `R.traverse` line up perfectly, but the actual implementation uses the _original_ traversable spec (which doesn't stay true to R.traverse's signature).

I'm happy to make the changes so that we stick with the old version of traverse `traverse(f, of)` instead of `traverse(of, f)`, I just think it will be confusing unless we point to a specific version of Traversable (perhaps a link to that part of the spec at the last version that had that signature).

If there isn't an indication of a specific version, then it makes sense to support whatever is current in FL. Is there any plan to officially support a specific version?

davidchambers:
> I just think it will be confusing unless we point to a specific version of Traversable (perhaps a link to that part of the spec at the last version that had that signature).

This is what I had in mind.

> If there isn't an indication of a specific version, then it makes sense to support whatever is current in FL. Is there any plan to officially support a specific version?

I would like to see Ramda adopt [sanctuary-type-classes][1]. We would then be able to delete a great deal of source code, and could incorporate additions to the Fantasy Land spec simply by upgrading our sanctuary-type-classes dependency.

I'll be happy to revive #1938 if and when @CrossEye indicates how he'd like to handle dispatching.


buzzdecafe:
I think @CrossEye may be on vacation atm? I'm just back myself; it's that time of year

dustinws:
That PR looks great, I think it would be a big step, but definitely in the right direction.

It looks like the scope of that PR would also trickle back down to this implementation of `R.traverse`. In the meantime, we could do the following -

- Revert my changes to the original spec for Traverse
- Add a link in the description of R.traverse that points to FL spec. We would link directly to the `Traversable` section, using the README from _the last version of FL to support that signature_.

I think that's fine, as long as the docs are explicit about _which_ Traversable is being used. This part isn't actually necessary, but I think it could save a decent amount of confusion.

I don't think linking to FL on a "per function" basis like this is a long term solution like 1938, but I do think it's acceptable until those bigger decisions are made about dispatching, etc...

Any thoughts on this?

CrossEye:
The first two weeks of August, I simply don't participate; my vacation time is sacrosanct.  I missed this one entirely.  It makes sense to me on a quick read.  @dustinws: do you want to update this PR to remove the conflicts?

PR #2253, opened 2017-08-12
Unwrapped exported modules from unnecessary IIFEs - which improves in…

Andarist:
…itialization time

I might be wrong, but I have not found in my head any reason why those (and many others) are wrapped in IIFEs. If this PR makes sense for you, I'll amend it later with the same change for all of the other modules.

### DIFF
--- a/src/internal/_xtakeWhile.js
+++ b/src/internal/_xtakeWhile.js
@@ -3,16 +3,14 @@ var _reduced = require('./_reduced');
 var _xfBase = require('./_xfBase');
 
 
-module.exports = (function() {
-  function XTakeWhile(f, xf) {
-    this.xf = xf;
-    this.f = f;
-  }
-  XTakeWhile.prototype['@@transducer/init'] = _xfBase.init;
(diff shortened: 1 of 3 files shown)


davidchambers:
The build script essentially concatenates all the source files, so it's not a good idea to define top-level variables. They would end up being shared which could cause problems.

Andarist:
@davidchambers Would u be interested in replacing this build script with some light-weight bundler like rollup?

kedashoe:
> @davidchambers Would u be interested in replacing this build script with some light-weight bundler like rollup?

We have an open PR for it and a couple issues discussing it. I'd like to get to it pretty soon here.

Andarist:
I have no found a PR about changing build of `dist/ramda.min.js`.

There is a POC about refactoring to es6 modules though, which i would love to see happen too. [I've commented](https://github.com/ramda/ramda/issues/1968#issuecomment-322033556) a little bit in the main (i think) thread about it.

So my question kinda still stands, as way of building UMD build and refactoring to es6 modules are imho separate topics.

Andarist:
closed in favor of #2254

PR #2254, opened 2017-08-15
Refactored the library to es6 modules, added babel build step for out…

### LINKED ISSUE #1968: Migrate to ES6 modules

Andarist:
…putting commonjs files, using rollup to make UMD builds

fixes #1968

I'm a little hot headed and prepared a refactor based on [this request](https://github.com/ramda/ramda/pull/1977#issuecomment-322070124).

This is a breaking change for cjs users, so migration/transition paths need to be established.

cc @jonaskello @kedashoe


jonaskello:
Nice work :-)

Andarist:
@jonaskello if you have any suggestions what could be changed to make it better, I would highly appreciate any input :) especially in regards of transitioning between versions

Andarist:
> do we need to break things for cjs?

I think we can preserve old behaviour by using custom cjs transform.

With the current setup such code:
```js
import _curry2 from './internal/_curry2';

var add = _curry2(function add(a, b) {
  return Number(a) + Number(b);
});
export default add;
```
produces this for the commonjs format:
```js
'use strict';
Object.defineProperty(exports, "__esModule", {
  value: true
});
var _curry = require('./internal/_curry2');
var _curry3 = _interopRequireDefault(_curry);
function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }

var add = (0, _curry3.default)(function add(a, b) {
  return Number(a) + Number(b);
});
exports.default = add;
```
Therefore such module have to be consumed like this when using `require` - `var add = require('ramda/cjs/add').default`. There is no problem when requiring whole `ramda`, nothing changes there - `var add = require('ramda').add`.

Attempt to change those exports to named ones yield such result:
```js
// same 'banner'
var add = exports.add = (0, _curry3.default)(function add(a, b) {
  return Number(a) + Number(b);
});
```
Even with other options passed to the transform like `{ loose: true, noInterop: true, strict: false }` we cannot completely preserve old structure, because it assumes we want new structure (for better interoperability) by using es6 modules syntax.

Therefore I believe that if we want to support the old way just the way it is we need to:
- write a custom es modules transform (I can do it, should be fairly easy)
- move current `/src` to `/es` and transpile `/es` to `/src`

> do we need babel?

We dont need it *per se* but id argue we want it, because working with code transformations on ASTs is way easier/cleaner than transforming bare files and its contents as string. It also opens immediately its whole ecosystem and can allow further refactors to es6 - most notably arrow functions which would be a perfect match for ramda's style :)



@kedashoe sorry for a wall of text, probably could have slim it down, but I wanted to share the exact reasoning. Please decide how I should proceed with this further

jonaskello:
I recall using the [add-module-exports](https://github.com/59naga/babel-plugin-add-module-exports) plugin to fix the .default require issue.

Andarist:
@jonaskello nice! I had a feeling this must be a solved problem, but couldnt find a plugin for this :)

with this setting:
```js
// .babelrc
{
  "plugins": [
  	"add-module-exports",
  	["transform-es2015-modules-commonjs", {"loose": true, "noInterop": true, "strict": false}]
  ]
}
```
we can go down to this
```js
exports.__esModule = true;
var _curry = require('./internal/_curry2');

var add = (0, _curry.default)(function add(a, b) {
  return Number(a) + Number(b);
});
exports.default = add;
module.exports = exports['default'];
```

Its still slighly bigger than original though:
```js

module.exports = _curry2(function add(a, b) {
  return Number(a) + Number(b);
});
```
and you need to decide if thats acceptable or not :)

kedashoe:
Thanks for the rec @jonaskello , sounds like it's worth a shot. Can you update the PR to use https://github.com/59naga/babel-plugin-add-module-exports, @Andarist ?

Also, what do you think about https://github.com/morlay/babel-plugin-annotate-pure-call-in-variable-declarator for automatically adding the PURE annotations?

Can we remove the `acorn` dependency now?

Is there a reason not to use buble? Seems to be the rollup recommended way to integrate babel?

Makefile
```diff
@@ -1,23 +1,16 @@
-UGLIFY = node_modules/.bin/uglifyjs
```
Andarist: there is still release script left there, personally im not really fond of Makefiles, but i assumed its preferred by the Ramda's core team, so I have left it in place

Andarist:
> Thanks for the rec @jonaskello , sounds like it's worth a shot. Can you update the PR to use https://github.com/59naga/babel-plugin-add-module-exports, @Andarist ?

added

> Also, what do you think about https://github.com/morlay/babel-plugin-annotate-pure-call-in-variable-declarator for automatically adding the PURE annotations?

oh, thats a newie! OTOH at the moment I transpile only once to cjs, adding this would require transpiling twice - to cjs target and es target (because source code, while written with es module, wouldnt have those comments)

the only problem I have with this is that I dont know how to call the new directory :trollface:
`lib` - taken, some scripts in there
`src` - target for cjs, .gitignored, but published to npm
`es` - source code, in repo, and published to npm
`dist` - umd builds

Also at the moment I think this plugin is overdoing the transformation, because it annotates every variable declarators and I think (if im wrong, please correct me) its only needed for the top level calls (for children of the body - ast speaking), so it produces:
```js
export const A = /*#__PURE__*/(() => {
  var B = /*#__PURE__*/call();
})();
```
That would increase quite significantly file sizes on the npm, however its easily fixable. Gonna raise an issue there.

> Can we remove the acorn dependency now?

dropped

> Is there a reason not to use buble? Seems to be the rollup recommended way to integrate babel?

At the moment building with rollup doesnt need any babel integration, because we do not transform anything else than es modules, and rollup already knows how to do that. It will be needed though if you decide to add any other transformations like arrow functions or others. Ive always done it with adding `rollup-plugin-babel` though. I believe buble is an alternative for babel which only supports es2015 transformations.

Andarist:
Also while adding `https://github.com/59naga/babel-plugin-add-module-exports` helped with keeping the old style of requires, it interferes with `transform-es2015-modules-commonjs` because of that I couldnt use `"noInterop": true` for the latter - this results in
```js
function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }
```
being included in each file that imports other module. That is because used imports are written like this:
```js
(0, _imported.default)
```
Without interop `_imported` became module itself (cause of `module.exports = exports['default'];`) and had no `.default` property on it.

Makefile
davidchambers: I love makefiles, and use them in all my projects. I don't contribute to Ramda much any longer, though, so I'm happy for you people to decide what is best for the project.

I urge you to keep in mind, @kedashoe, that as you currently seem to be the primary maintainer of this project you will be most affected by the consequences—positive and negative—of any changes to the way Ramda is built and distributed.

kedashoe:
Sorry I'm not understanding where we are with babel. Up above I asked if we needed it and you said

> We dont need it per se but id argue we want it, because working with code transformations on ASTs is way easier/cleaner than transforming bare files and its contents as string.

but asking about buble you say

> > At the moment building with rollup doesnt need any babel integration, because we do not transform anything else than es modules, and rollup already knows how to do that.

As to https://github.com/59naga/babel-plugin-add-module-exports, if that interferes with the exports, how about writing our own transform?

Andarist:
@kedashoe 
I'll try to clarify what I've meant :)

From what I understand buble is an alternative to babel. I thought u were suggesting using buble in combination with rollup, thinking it is a plugin for babel integration. Actually for a rollup (which creates UMD builds for us) at the moment any additional AST transformations are not needed, because the only thing we want to transpile  **for now** are es modules which rollup handles on its own - without any additional tools.

In theory we could transform source files (containing es modules) 'by hand' to the commonjs format, because it should be quite trivial, in similar way like [you build](https://github.com/ramda/ramda/blob/fc95d85b0ee385060b1f15a0610dc35e59363f05/scripts/build) ur UMD files today.

However transformations on AST (what babel does) is a more high level concept than parsing source code as a string and less brittle, so I advise to use this full-blown solution (babel) even for this single (for now) transform we need to do.

> As to https://github.com/59naga/babel-plugin-add-module-exports, if that interferes with the exports, how about writing our own transform?

Sure thing, I've proposed this as viable option before. Gonna make this in following days.

Andarist:
@kedashoe I've written custom transformer quickly, which is heavily based on the original babel's transform. It supports ramda's needs nicely - doesnt cover each form of the import/export syntax, where I could Ive added checks which should throw against wrong usage (i.e. using named exports). 

I've omitted some cases in the implementation not because they are somehow invalid, but not used by the project (like exporting ClassDeclaration) and I throw then too - those can be implemented when needed.

I dont feel like this transform needs any additional tests, if it mess up with the code regular tests should catch it, so in a sense they are testing this transform too, because they test transformed code.

kedashoe:
This is great @Andarist , thanks!

I'd like to see a more minimal implementation to start out. What are your thoughts on:

- a file `scripts/babel/esm-to-cjs.js`
- let's get rid of the error checks for a minute
- rather than looping and searching in `Program / exit`, having visitors for `ExportDefaultDeclaration` and `ImpotDeclaration` (let's move these to top level functions as well, always bothers me all the indentation you end up with with babel)

eg

```js
function exportDefaultDeclaration(path) {
  var declaration = path.get('declaration');

  if (declaration.isFunctionDeclaration()) {
    var id = declaration.node.id;
    if (id) {
      path.replaceWithMultiple([
        declaration.node,
        buildExportsAssignment(
          t.identifier(id.name)
        )
      ]);
    } else {
      path.replaceWith(
        buildExportsAssignment(declaration)
      );
    }
  } else {
    path.replaceWith(
      buildExportsAssignment(declaration.node)
    );
  }
}

function importDeclaration(path) {
  // etc
}

module.exports = function() {
  return {
    visitor: {
      ExportDefaultDeclaration: exportDefaultDeclaration,
      ImportDeclaration: importDeclaration,
    }
  };
};
```

Could you leave the current file, `scripts/transform-es2015-modules-bare-cjs.js`, around for now as well to make it easier to see what we've done.

Andarist:
@kedashoe I've prepared a simpler transform [here](https://github.com/Andarist/ramda/blob/01a0f234c1db580257687b4263dace5071b3cc95/scripts/transform-esm-to-cjs.js) (its ofc part of the PR now, just pointing to it directly as navigating through this this 374 file PR is not handy 😄 )

I have left some (only like 3 now) mentioned checks in place, as I was not so sure that those were safe to remove - transform code make some assumptions about the structure of the project and is not a universal transform. We can ofc remove those too, ur call here :)

kedashoe:
> I've prepared a simpler transform here (its ofc part of the PR now, just pointing to it directly as navigating through this this 374 file PR is not handy :smile: )

Thanks looks good!

> I have left some (only like 3 now) mentioned checks in place, as I was not so sure that those were safe to remove - transform code make some assumptions about the structure of the project and is not a universal transform.

Sounds good.

Can we add another `transform` file to automatically insert the `PURE` annotations? Annoying seeing them in the source code imho. Let me know your thoughts on that.

Andarist:
> Can we add another transform file to automatically insert the PURE annotations? Annoying seeing them in the source code imho. Let me know your thoughts on that.

Sure thing, although I would write my own for now. Gonna work with [`annotate-pure-call-in-variable-declarator's`](https://github.com/morlay/babel-plugin-annotate-pure-call-in-variable-declarator) author simultaneously though so in the future we could drop custom transform and replace it with it, because at the moment we cannot use it as is.

Andarist:
@kedashoe ok, so I've written a super simple plugin for annotating those top-level calls automatically and it does a great job for ramda's use case

Travid build is failing because `cross-env` package uses `Object.assign` which isnt available in old versions of Node. Would u consider dropping support for those no longer maintained node versions or some other solution needs to be found? Let me know :)

kedashoe:
@Andarist @jonaskello @olsonpm / whoever else would like to test it out, I've publish this WIP to npm

```bash
npm install ramda@es-rc
```

Any feedback greatly appreciated!

olsonpm:
will do tomorrow.  i have a perfect small ephemeral work project to test it out with.

glad to see you found the time for open source!  once my new job winds down a bit i'm hoping to hop back in.

Andarist:
@kedashoe 

Followuping here to the https://github.com/ramda/ramda/issues/1968#issuecomment-327151903

Several files (the ones using 'constructor pattern') needs to wrapped in IIFEs for better tree-shakeability. There are actually 3 ways we could approach this:
- wrapping in IIFEs by hand
- write a custom babel transform for wrapping
- use es6 classes and transpile them down to es5 with standard [`babel-plugin-transform-es2015-classes`](https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-es2015-classes)

kedashoe:
I'd say lets avoid classes for this PR. I like the babel transform since it makes it obvious why we are doing it. How will you determine when to trigger it? Detect prototype assignment?

Andarist:
Yeah, prototype assignment + traversal upwards, gathering whole 'class' together and wrapping it in IIFE. Gonna cook this up when I have time :)

olsonpm:
Well considering you guys went with the pure annotation approach, I'm less interested in this branch.  Comparing the bundled output of ramda@es-rc vs [mine](https://github.com/olsonpm/ramda-with-rollup-example) (where I manually added the pure annotations), yours produces a lot of unused code - implying the [annotation plugin](https://github.com/Andarist/ramda/blob/75c0fbc9ae2926a1e3b2c71c142276c9789b6404/.babelrc#L4) being used needs some work.

Andarist:
> Well considering you guys went with the pure annotation approach, I'm less interested in this branch. 

Is there any other approach we should consider?

> yours produces a lot of unused code - implying the annotation plugin being used needs some work.

True that, I will soon add next custom transform to support better tree-shakeability like mentioned above. The reason why we end up with a lot on unused code is that I have unwrapped constructor patterns like [this one](https://github.com/olsonpm/ramda/blob/4a59eca30a076b8f221805dd650e7303e55ed432/src/internal/_xmap.js#L5) and because of the prototype assignments being in the top level scope those 'classes' cannot be tree-shaken properly. Im gonna add a transform which wraps those in the IIFE and then the already used [annotation plugin](https://github.com/Andarist/ramda/blob/75c0fbc9ae2926a1e3b2c71c142276c9789b6404/scripts/annotate-pure-calls.js) will kick in and annotate those IIFEs. Ive checked that this approach will yield the expected results. More can be read [here](https://github.com/ramda/ramda/issues/1968#issuecomment-327151903). So stay tuned! :)

Andarist:
@kedashoe babel plugin for IIFE wrapping added :) consumers' bundles should get much smaller now

kedashoe:
> babel plugin for IIFE wrapping added :) consumers' bundles should get much smaller now

Published!

For those who want to test, same command to install:

`npm install ramda@es-rc`

My test bundle went from 30k to 4k :)

Andarist:
> My test bundle went from 30k to 4k :)

🎉 🎉 🎉 🎉 🎉 🎉 

Gonna rebase this PR some time today, to keep it merge-able ;)

Andarist:
Done. No conflicts and tests are passing :)

vakhtang:
Will these changes work with webpack's tree-shaking via UglifyJS?

kedashoe:
> Should these changes work with webpack's tree-shaking via UglifyJS?

I've not used webpack before. I gave it a try and couldn't figure it out. Help anyone?

Andarist:
@vakhtang 
Those changes are bundler agnostic, should work just fine with webpack. Because of heavy usage of higher order functions rollup alone is not able to tree-shake ramda properly (at least not for now). Every call to the function which appears to be on the top level of the file is treated as possibly impure. That's why we are adding (as a transpilation step) a `/*#__PURE__*/` comment to every such call so it can be later minified properly by UglifyJS. 

Webpack is not doing much in regards of dead code elimination anyway - it builds a dependency graph, detects those which are unused and remove their assignments to the `exports`. So basically what webpack's tree shaking does is this:

**INPUT**
```
export function used(value) {}
export function unused(value) {}
```

**OUTPUT**
```
function(module, exports, __webpack_require__) {
    "use strict";
    Object.defineProperty(exports, "__esModule", {
        value: true
    });

    function used(value) {}
    function unused(value) {}
    exports.used = used;
    /* unused harmony export */ unused;
}
```

This is not exact output code, but it roughly looks something like this. 

What we can learn from this snippet is that actually Webpack doesnt do Dead Code Elimination to the full extent. It just removes `exports` assignments of the unused code and only latter uglifyJS removes this dead code from the final bundle when it detects its safe to do.

So with libraries like `ramda` for now there is no bundler (to the best of my knowledge) which tree-shakes it properly and we need to rely on minifying process for this (aided with pure annotations) to make this.

vakhtang:
@Andarist 

Thanks for the explanation. I guess tree-shaking is kind of a misnomer with respect to webpack, huh? I am seeing the `/*#__PURE__*/` annotations in the (unminified) bundle, but when using `UglifyJsPlugin` it doesn't seem to be dropping them. Is there a specific configuration UglifyJS requires?

Andarist:
Nope it doesnt any additional configuration (at least not any I know of). Do u run ur tests with latest UglifyJS' version? Could u share your repository or the repository with the issue reproduced in it?

vakhtang:
@Andarist Give this a shot: https://github.com/vakhtang/tree-shaking-demo. I'm using [UglifyJsPlugin](https://github.com/webpack-contrib/uglifyjs-webpack-plugin), which seems to be using a recent version of UglifyJS.

Andarist:
Ive inspected the final bundle and indeed some things arent tree-shaken. I need to investigate this further.

However - 162 modules were detected as unused and marked with `/* unused harmony default export */` annotation by webpack.
While code itself of those unused exports got tree shaken, unfortunately webpack modules containing those tree shaken modules were not.

**Webpack's module example output**
```js
/* 319 */
/***/ (function(module, __webpack_exports__, __webpack_require__) {

"use strict";
/* harmony import */ var __WEBPACK_IMPORTED_MODULE_0__internal_curry3__ = __webpack_require__(2);

var zipWith = /*#__PURE__*/Object(__WEBPACK_IMPORTED_MODULE_0__internal_curry3__["a" /* default */])(function zipWith(fn, a, b) {
  var rv = [];
  var idx = 0;
  var len = Math.min(a.length, b.length);
  while (idx < len) {
    rv[idx] = fn(a[idx], b[idx]);
    idx += 1;
  }
  return rv;
});
/* unused harmony default export */ var _unused_webpack_default_export = (zipWith);

/***/ })
```

***after uglifyjs pass***
```js
function(module, __webpack_exports__, __webpack_require__) {
    "use strict";
    __webpack_require__(2);
}
```

So as you can see the code of `zipWith` is no longer in the bundle, but the webpack's module code is still there because it still contains `__webpack_require__(2)` (`_curry3`) call. And probably because of such 'partial' tree shaking some modules, such as `_curry3` in this example, stay in the bundle.

I'll reach out to the webpack community if there is any solution for this.

Thanks for the helpful repository!

Andarist:
@vakhtang ive added a PS to the last comment, while u have already liked the reply, please see it ;)

vakhtang:
@Andarist Now that's an improvement, thanks! I'm not sure I understand why that plugin helps, but that's interesting.

Andarist:
> I'm not sure I fully understand why that plugin helps, but that's interesting.

With standard webpack configuration each module is contained in a function, all modules are gathered in a gigantic array. We could imagine it like this:
```js
var modules = [
  function(module, __webpack_exports__, __webpack_require__) { /* module A */ },
  function(module, __webpack_exports__, __webpack_require__) {
     /* module B */
     var a = __webpack_require__(0);
     /* ... */
   },
  /* ... */
]
```

And the webpack's runtime references other modules by accessing them through this array and its indexes. This brings some runtime costs, because each module needs to be initialized in the runtime.

`ModuleConcatenationPlugin`  simply tries to concatenate all modules (let assume we do not do any code-splitting) into a single function - this is called a scope hoisting (because all modules start to share the same scope - in case of naming conflicts hoisted things get renamed) and its an idea popularized by rollup.

So with regular webpack config even if we tree shake unused exports the imported modules (those calls to the `__webpack_require__`) stays there (which is a huge bummer! ive just realized this). What is even worse - if we leave call to `__webpack_require__` which requires other unused exports - the unused export cannot be tree shaken, because its referenced. Im gonna give a simple example of this at the end of this reply.

Going back to the `ModuleConcatenationPlugin` - since it hoists everything to the single closure there are no calls to `__webpack_require__` there and UglifyJS recognizes all unused stuff properly and dead code eliminate them.

What's funny that even tree shaken modules leave in the bundle 'empty modules':
```js
function(module, __webpack_exports__, __webpack_require__) {
    "use strict";
}
```

Webpack's broken tree shaking example:
<details>
 <summary>entry point - main.js</summary>
```js
import { add } from './maths'
console.log(add(1, 2));
```
<details>

<details>
 <summary>maths.js</summary>
```js
export { default as sub } from './sub'
export { default as add } from './add'
export { default as mult } from './mult'
export { default as pow } from './pow'
```
<details>

<details>
 <summary>add.js</summary>
```js
export default function add(a, b) {
	return a + b;
}
```
<details>

<details>
 <summary>sub.js</summary>
```js
export default function sub(a, b) {
	return a - b;
}
```
<details>

<details>
 <summary>mult.js</summary>
```js
export default function mult(a, b) {
	return a * b;
}
```
<details>

<details>
 <summary>pow.js</summary>
```js
import mult from './mult'

export default function pow(a, b) {
	let res = a;
	while (--b) {
		res = mult(res, a);
	}
	return res;
}
```
<details>

<details>
 <summary>output bundle</summary>
```js
!function(modules) {
    function __webpack_require__(moduleId) {
        if (installedModules[moduleId]) return installedModules[moduleId].exports;
        var module = installedModules[moduleId] = {
            i: moduleId,
            l: !1,
            exports: {}
        };
        return modules[moduleId].call(module.exports, module, module.exports, __webpack_require__), 
        module.l = !0, module.exports;
    }
    var installedModules = {};
    __webpack_require__.m = modules, __webpack_require__.c = installedModules, __webpack_require__.d = function(exports, name, getter) {
        __webpack_require__.o(exports, name) || Object.defineProperty(exports, name, {
            configurable: !1,
            enumerable: !0,
            get: getter
        });
    }, __webpack_require__.n = function(module) {
        var getter = module && module.__esModule ? function() {
            return module.default;
        } : function() {
            return module;
        };
        return __webpack_require__.d(getter, "a", getter), getter;
    }, __webpack_require__.o = function(object, property) {
        return Object.prototype.hasOwnProperty.call(object, property);
    }, __webpack_require__.p = "", __webpack_require__(__webpack_require__.s = 1);
}([ function(module, __webpack_exports__, __webpack_require__) {
    "use strict";
    function mult(a, b) {
        return a * b;
    }
    __webpack_exports__.a = mult;
}, function(module, __webpack_exports__, __webpack_require__) {
    "use strict";
    Object.defineProperty(__webpack_exports__, "__esModule", {
        value: !0
    });
    var __WEBPACK_IMPORTED_MODULE_0__maths__ = __webpack_require__(2);
    console.log(Object(__WEBPACK_IMPORTED_MODULE_0__maths__.a)(1, 2));
}, function(module, __webpack_exports__, __webpack_require__) {
    "use strict";
    /* this is maths.js */
    var __WEBPACK_IMPORTED_MODULE_1__add__ = (__webpack_require__(3), __webpack_require__(4));
    __webpack_require__.d(__webpack_exports__, "a", function() {
        return __WEBPACK_IMPORTED_MODULE_1__add__.a;
    });
    __webpack_require__(0), __webpack_require__(5);
}, function(module, __webpack_exports__, __webpack_require__) {
    "use strict";
    /* this is sub.js */ 
}, function(module, __webpack_exports__, __webpack_require__) {
    "use strict";
    function add(a, b) {
        return a + b;
    }
    __webpack_exports__.a = add;
}, function(module, __webpack_exports__, __webpack_require__) {
    "use strict";
    /* this is tree-shaken pow.js, still containing a require call to mult.js */ 
    __webpack_require__(0);
} ]);
```
<details>

We can easily inspect the output bundle to observe the outcome, but to sum it up:
- `add` is ofc used and stays in the bundle
- `sub` got tree-shaken easily and is not in the bundle
- `pow`'s implementation is gone (tree-shaken) but the implementation referenced `mult` module ad the require call to it stays in the bundle
- because "`require('./mult')`" stays in the bundle it cannot be tree shaken although its unused

Andarist:
I've tested ramda with webpack and its latest addition (not yet released! you can try it out with `npm install webpack/webpack#next`) - reading `side-effects` entry from the `package.json`.

I've [added it](https://github.com/ramda/ramda/pull/2254/commits/42eaa74b3711b344fdf11fff60d78e9ed384598d) in this PR, even though its not a released feature of webpack yet - I feel its good to be ready for it, especially that its just a 1 line change.

It ensures webpack that a contained modules of a package (or your app if you put this in yours app's `package.json`) are side-effect free and allows for an easier to apply tree-shaking algorithm. After my simple test with ramda:
```js
import { add } from 'ramda'
console.log(add(1, 2));
```
I've confirmed that it now includes only relevant modules, even without applying uglify.
<details>
  <summary>Output bundle</summary>

```js
/******/ (function(modules) { // webpackBootstrap
/******/ 	// The module cache
/******/ 	var installedModules = {};
/******/
/******/ 	// The require function
/******/ 	function __webpack_require__(moduleId) {
/******/
/******/ 		// Check if module is in cache
/******/ 		if(installedModules[moduleId]) {
/******/ 			return installedModules[moduleId].exports;
/******/ 		}
/******/ 		// Create a new module (and put it into the cache)
/******/ 		var module = installedModules[moduleId] = {
/******/ 			i: moduleId,
/******/ 			l: false,
/******/ 			exports: {}
/******/ 		};
/******/
/******/ 		// Execute the module function
/******/ 		modules[moduleId].call(module.exports, module, module.exports, __webpack_require__);
/******/
/******/ 		// Flag the module as loaded
/******/ 		module.l = true;
/******/
/******/ 		// Return the exports of the module
/******/ 		return module.exports;
/******/ 	}
/******/
/******/
/******/ 	// expose the modules object (__webpack_modules__)
/******/ 	__webpack_require__.m = modules;
/******/
/******/ 	// expose the module cache
/******/ 	__webpack_require__.c = installedModules;
/******/
/******/ 	// define getter function for harmony exports
/******/ 	__webpack_require__.d = function(exports, name, getter) {
/******/ 		if(!__webpack_require__.o(exports, name)) {
/******/ 			Object.defineProperty(exports, name, {
/******/ 				configurable: false,
/******/ 				enumerable: true,
/******/ 				get: getter
/******/ 			});
/******/ 		}
/******/ 	};
/******/
/******/ 	// getDefaultExport function for compatibility with non-harmony modules
/******/ 	__webpack_require__.n = function(module) {
/******/ 		var getter = module && module.__esModule ?
/******/ 			function getDefault() { return module['default']; } :
/******/ 			function getModuleExports() { return module; };
/******/ 		__webpack_require__.d(getter, 'a', getter);
/******/ 		return getter;
/******/ 	};
/******/
/******/ 	// Object.prototype.hasOwnProperty.call
/******/ 	__webpack_require__.o = function(object, property) { return Object.prototype.hasOwnProperty.call(object, property); };
/******/
/******/ 	// __webpack_public_path__
/******/ 	__webpack_require__.p = "";
/******/
/******/ 	// Load entry module and return exports
/******/ 	return __webpack_require__(__webpack_require__.s = 2);
/******/ })
/************************************************************************/
/******/ ([
/* 0 */
/***/ (function(module, __webpack_exports__, __webpack_require__) {

"use strict";
/* harmony export (binding) */ __webpack_require__.d(__webpack_exports__, "a", function() { return _isPlaceholder; });
function _isPlaceholder(a) {
       return a != null && typeof a === 'object' && a['@@functional/placeholder'] === true;
}

/***/ }),
/* 1 */
/***/ (function(module, __webpack_exports__, __webpack_require__) {

"use strict";
/* harmony export (binding) */ __webpack_require__.d(__webpack_exports__, "a", function() { return _curry1; });
/* harmony import */ var _isPlaceholder__WEBPACK_IMPORTED_MODULE_0__ = __webpack_require__(0);


/**
 * Optimized internal one-arity curry function.
 *
 * @private
 * @category Function
 * @param {Function} fn The function to curry.
 * @return {Function} The curried function.
 */
function _curry1(fn) {
  return function f1(a) {
    if (arguments.length === 0 || Object(_isPlaceholder__WEBPACK_IMPORTED_MODULE_0__["a" /* default */])(a)) {
      return f1;
    } else {
      return fn.apply(this, arguments);
    }
  };
}

/***/ }),
/* 2 */
/***/ (function(module, __webpack_exports__, __webpack_require__) {

"use strict";
Object.defineProperty(__webpack_exports__, "__esModule", { value: true });
/* harmony import */ var ramda__WEBPACK_IMPORTED_MODULE_0__ = __webpack_require__(3);


console.log(Object(ramda__WEBPACK_IMPORTED_MODULE_0__["a" /* default */])(1, 2));


/***/ }),
/* 3 */
/***/ (function(module, __webpack_exports__, __webpack_require__) {

"use strict";
/* harmony import */ var _internal_curry2__WEBPACK_IMPORTED_MODULE_0__ = __webpack_require__(4);


/**
 * Adds two values.
 *
 * @func
 * @memberOf R
 * @since v0.1.0
 * @category Math
 * @sig Number -> Number -> Number
 * @param {Number} a
 * @param {Number} b
 * @return {Number}
 * @see R.subtract
 * @example
 *
 *      R.add(2, 3);       //=>  5
 *      R.add(7)(10);      //=> 17
 */
var add = /*#__PURE__*/Object(_internal_curry2__WEBPACK_IMPORTED_MODULE_0__["a" /* default */])(function add(a, b) {
  return Number(a) + Number(b);
});
/* harmony default export */ __webpack_exports__["a"] = (add);

/***/ }),
/* 4 */
/***/ (function(module, __webpack_exports__, __webpack_require__) {

"use strict";
/* harmony export (binding) */ __webpack_require__.d(__webpack_exports__, "a", function() { return _curry2; });
/* harmony import */ var _curry1__WEBPACK_IMPORTED_MODULE_0__ = __webpack_require__(1);
/* harmony import */ var _isPlaceholder__WEBPACK_IMPORTED_MODULE_1__ = __webpack_require__(0);



/**
 * Optimized internal two-arity curry function.
 *
 * @private
 * @category Function
 * @param {Function} fn The function to curry.
 * @return {Function} The curried function.
 */
function _curry2(fn) {
  return function f2(a, b) {
    switch (arguments.length) {
      case 0:
        return f2;
      case 1:
        return Object(_isPlaceholder__WEBPACK_IMPORTED_MODULE_1__["a" /* default */])(a) ? f2 : Object(_curry1__WEBPACK_IMPORTED_MODULE_0__["a" /* default */])(function (_b) {
          return fn(a, _b);
        });
      default:
        return Object(_isPlaceholder__WEBPACK_IMPORTED_MODULE_1__["a" /* default */])(a) && Object(_isPlaceholder__WEBPACK_IMPORTED_MODULE_1__["a" /* default */])(b) ? f2 : Object(_isPlaceholder__WEBPACK_IMPORTED_MODULE_1__["a" /* default */])(a) ? Object(_curry1__WEBPACK_IMPORTED_MODULE_0__["a" /* default */])(function (_a) {
          return fn(_a, b);
        }) : Object(_isPlaceholder__WEBPACK_IMPORTED_MODULE_1__["a" /* default */])(b) ? Object(_curry1__WEBPACK_IMPORTED_MODULE_0__["a" /* default */])(function (_b) {
          return fn(a, _b);
        }) : fn(a, b);
    }
  };
}

/***/ })
/******/ ]);
```
</details>

**Note**: it only works for modules written with es6 syntax, I've tried it against latest ramda (`0.24.1`) and still the whole library got pulled into the bundle.

CrossEye:
After some serious prodding from @kedashoe, I've finally spent some real time with this and believe I understand pretty well what's going on.  Sorry it's taken me so long.

:herb:

I'd be very happy to see this move forward.

Andarist:
I have just spotted that I probably need to restore old [build script](https://github.com/ramda/ramda/blob/e76b9a75cca909c1c1559f62d58f96b2be7af1b4/scripts/build) (probably tweaked or something). I haven't noticed that it was used also for [partial builds](https://github.com/ramda/ramda#partial-builds).

kedashoe:
:100: thank you for all your hard work on this @Andarist and everyone else who helped with this one!

PR #2256, opened 2017-08-19
add String Type to drop/takeWhile API

adispring:


src/dropWhile.js
```diff
@@ -1,6 +1,7 @@
 var _curry2 = require('./internal/_curry2');
 var _dispatchable = require('./internal/_dispatchable');
 var _xdropWhile = require('./internal/_xdropWhile');
+var slice = require('./slice');
```
buzzdecafe: i believe the internal references were deliberately removed to simplify the dependency tree?

adispring: But if using Array.prototype.slice, the returned value will always be an array. Then how to let dropWhile compatible with String?

buzzdecafe: Yes you are correct of course, my apologies for being dumb. I blame the eclipse

CrossEye: In fact, it's the year of the eclipse.  We can use that excuse until January, I believe. :smile:

CrossEye:
:+1:

buzzdecafe:
:cow2:

src/dropWhile.js
adispring: A big and long eclipse that hide something 😁, thanks for merging.

PR #2261, opened 2017-08-23
add an alias for R.identity: R.I

dronsfield:
`R.identity` is a great placeholder or default function. I use it a lot when a function may not exist.

But because of the nature of how `R.identity` is used, increased brevity would be great.

```
// this is annoying
if (someFunction) {
  someFunction();
}

// this is better
(someFunction || R.identity)();

// this would be even better
(someFunction || R.I)();
```

I think `R.I` would be a perfect alias. `I` is not just the first letter of "identity", it is also commonly used to denote the identity matrix in mathematics.

I don't know how the maintainers feel about aliases, but I think this would be cool.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,7 @@
 module.exports = {
   F: require('./src/F'),
   T: require('./src/T'),
+  I: require('./src/I'),
   __: require('./src/__'),
   add: require('./src/add'),
   addIndex: require('./src/addIndex'),
--- a/src/I.js
+++ b/src/I.js
@@ -1,0 +1,24 @@
+var identity = require('./identity');
+
+
+/**
+ * An alias for R.identity (a function that does nothing but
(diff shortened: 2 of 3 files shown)


Bradcomp:
Hi @dronsfield,
Thank you for the pull request! 

Ramda generally doesn't go for aliases, and in fact this exact alias was removed in a past PR. You can see that PR [here](https://github.com/ramda/ramda/issues/859), as well as the justification for the stance [here](https://github.com/ramda/ramda/pull/574#issuecomment-64124850).

CrossEye:
`R.I` is the only alias (except perhaps for `R,foldr`/`R.foldl` ones) that I actually miss.  But I definitely have internalized [David's parable][pa].

  [pa]: https://github.com/ramda/ramda/pull/574#issuecomment-64124850

PR #2267, opened 2017-08-30
Proposing a new function: chainPath

halis:
- While `path` allows you to navigate an object
- `chainPath` allows you to navigate through a `List` of
deeply nested objects and flattens the result
- Filters out undefined values in the resulting array
- See tests for examples

### DIFF
--- a/src/chainPath.js
+++ b/src/chainPath.js
@@ -1,0 +1,85 @@
+var _curry2 = require('./internal/_curry2');
+var chain = require('./chain');
+var prop = require('./prop');
+
+function chainProp(list, propName) {
+  return chain(prop(propName), list);
+}
+
+function notUndefined(x) {
+  return x !== undefined;
+}
+
+
+/**
+ * Retrieve the flat array of values at a given path over an array.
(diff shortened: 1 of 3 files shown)


src/chainPath.js
```diff
@@ -0,0 +1,85 @@
...
+ *              name: { first: 'Jimmy' }
+ *            },
+ *            {
+ *              hobbies: [{ name: 'Covfefe' }],
+ *              name: { first: 'Timmy' }
+ *            }
+ *          ]
+ *        }
+ *      ]
+ *        {
```
CrossEye: Is this section spurious?  It doesn't seem to match output below.

CrossEye:
At an initial look, this seems to do too much.

What if a user *doesn't* want the `undefined` values filtered?  What if she would prefer that the result *not* be flattened?  When we have to ask questions like this, I tend to think the function is doing more than necessary.

kedashoe:
Agreed with @CrossEye , gonna close. Thanks though for the pr @halis

PR #2284, opened 2017-09-12
fix #2279: zipObj doc

### LINKED ISSUE #2279: zipObj equivalence sentence
Hello, there is a line of documentation that sais:

> Note: zipObj is equivalent to pipe(zipWith(pair), fromPairs).

https://github.com/ramda/ramda/blob/7df44747aab2987e6b671b5658ab74abaebdbbdc/src/zipObj.js#L7

I don't see the need to use zipWith when you can use plain zip:

```js
pipe(zip,fromPairs)(['a','b','c'],[1,2,3]) //=> {a: 1, b: 2, c: 3}
```

The output is exactly the same.

Regards

buzzdecafe:


Bradcomp:
:bowtie:

CrossEye:
What, accuracy is not enough?  Now we have to have elegance?!

:stuck_out_tongue_winking_eye:

PR #2286, opened 2017-09-13
docs(concat): add note about Semigroup fantasy-land spec

char0n:


CrossEye:
I agree with the idea, but the wording is not quite correct:

> Can also concatenate two fantasy-land compatible semigroups

A semigroup is a *type*, a collection of values.  The collection of all Strings, along with String concatenation is a Semigroup.  The positive integers along with addition is a Semigroup.  But `R.concat` will not concatenate this collection of Strings with the positive integers.

Perhaps:

> Can also concatenate two members of a fantasy-land compatible semigroup.

?

char0n:
Rewording complete.

buzzdecafe:
:cow2:

Thanks

CrossEye:
Thanks @char0n !

char0n:
Np my pleasure ;]

PR #2288, opened 2017-09-15
add recursive predicating to R.where

adispring:
Add recursive predicating to R.where. If this pr can be accepted, I'll add recursive predicating to R.whereEq.

### DIFF
--- a/src/where.js
+++ b/src/where.js
@@ -4,10 +4,9 @@ var _has = require('./internal/_has');
 
 /**
  * Takes a spec object and a test object; returns true if the test satisfies
- * the spec. Each of the spec's own properties must be a predicate function.
- * Each predicate is applied to the value of the corresponding property of the
- * test object. `where` returns true if all the predicates return true, false
- * otherwise.
+ * the spec. Each predicate can be applied recursively to the value of the
+ * corresponding property of the test object. `where` returns true if all the
+ * predicates return true, false otherwise.
  *
  * `where` is well suited to declaratively expressing constraints for other
  * functions such as [`filter`](#filter) and [`find`](#find).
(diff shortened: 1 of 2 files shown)


buzzdecafe:
I'm afraid I am a "no" vote on this. This makes the function do too much. I would rather keep separate the generation of the predicate from the descent into the tree. 

There are also numerous hairy edge-cases this raises that are not dealt with here; e.g. what if the spec has a circular reference?

CrossEye:
Hmm, I thought I responded earlier.  I'm with @buzzdecafe: this seems a step too far.

But I would be interested in fixing `where` so that this would work:

```js
const predRcrsv = R.where({
  a: R.where({
    b: R.equals('bar'),
  }),
  x: R.gt(R.__, 10),
})
```js

It passes the others but fails at the moment on this test case: 

```js
predRcrsv({b: 'bar', x: 11, y: 19}) 
```

That's because the `a` property we pass into the internal `where` is undefined.  I could see fixing this up so that in that case, the predicate automatically returns false.

I haven't dug into the code yet, but will try to do so soon.

Anyone have thoughts on this?

asaf-romano:
This came up before:
https://github.com/ramda/ramda/pull/1618

Back then, Scott was kinda-sorta on my side ;)

Given that ``evolve`` and ``applySpec`` are recursive, I still think this makes sense. The only difference is that ``where`` can also be compared with ``whereEq``, which cannot be recursive.

Anyway, here's a function from our code-base, which would used in order to make our ``where`` recursive:

``js

const isPlainObject = pipe(type, equals("Object"));

const toRecursiveSpecFn = specFn => curryN(
    specFn.length,
    function(spec, ...args)
    {
        let recursiveSpec = map(when(either(Array.isArray, isPlainObject),
                                     subSpec => toRecursiveSpecFn(specFn)(subSpec)),
                                spec);
        return specFn.apply(this, [recursiveSpec, ...args]);
    }
);

```

Demo: https://goo.gl/roXfAc

**Comment by adispring** [comment]
This "recursive" R.where can work well if only bind predicate function to the most inner properties, 
Such as:
```js
const predRcrsv = R.where({
  a: {
    b: R.equals('bar'),
    c: { d: R.equals('foo'), },
  }
  x: R.gt(R.__, 10),
})
```

```js
predRcrsv({b: 'bar', c: 1, x: 11, y: 19}) 
```

PR #2292, opened 2017-09-20
Add R.first function

myrubapa:
Add R.first function

### DIFF
--- a/test/last.js
+++ b/test/last.js
@@ -6,7 +6,7 @@ var eq = require('./shared/eq');
 
 describe('last', function() {
 
-  it('returns the first element of an ordered collection', function() {
+  it('returns the last element of an ordered collection', function() {
     eq(R.last([1, 2, 3]), 3);
     eq(R.last([1, 2]), 2);
     eq(R.last([1]), 1);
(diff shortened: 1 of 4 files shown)


mattgrande:
I think this is a duplicate of `R.head`

http://ramdajs.com/docs/#head

test/last.js
```diff
@@ -6,7 +6,7 @@ var eq = require('./shared/eq');
 
 describe('last', function() {
 
-  it('returns the first element of an ordered collection', function() {
+  it('returns the last element of an ordered collection', function() {
```
CrossEye: Oops!  This should get fixed, regardless of the current PR.

CrossEye:
Thanks for the suggestion.

As @mattgrande said, this is a duplicate of the functionality of `head`.  (At some point, when we'd had only `head` and `tail`, we had requests for one of the the others and we added both `init` and `last` for symmetry.)  So there's no reason for `first`, and we have a no-aliases policy.

Nonetheless, thank you for your interest and for a nice straightforward PR.

I would love a tiny PR focused on fixing the typo in `test/last.js` though, if you're interested.

tanrax:
Wouldn't it be interesting to rename _head_ a _first_?

CrossEye:
@tanrax: No, not to me.  Why do you think it would be an improvement?  `head`/`tail` is a very old convention.

PR #2293, opened 2017-09-20
Add unpkg and jsdelivr field to package.json

porsager:
It would be great to be able to use ramda directly through https://unpkg.com/ramda

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
   },
   "main": "index.js",
   "unpkg": "dist/ramda.min.js",
+  "jsdelivr": "dist/ramda.min.js",
   "files": [
     "index.js",
     "src",
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "url": "git://github.com/ramda/ramda.git"
   },
   "main": "index.js",
+  "browser": "dist/ramda.min.js",
   "files": [
     "index.js",
     "src",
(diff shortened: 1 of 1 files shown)


git-in-my-anus:
Maybe the more commonly used field `"browser"` should be added instead?

porsager:
I did that at first, but I didn't want to be the one who broke peoples code if they had their bundler set up to use the browser field if available (I know rollup has an option to do that at least) 😊

But yes, generally it would be best with browser, but I somehow think that could be considered a breaking change. 
I can see the issue targeting unpkg directly, because then jsdelivr should probably be added as well.

MartinKolarik:
To make it even more complicated, unpkg is going to remove support for `browser` 😄 The best option right now is setting both `jsdelivr` and `unpkg` which doesn't risk breaking anything and will work with both services.

porsager:
Ah, thanks @MartinKolarik for that info.. Just read through the thread at https://github.com/unpkg/unpkg-website/issues/63

I'll add jsdelivr to this pull request as well, and cross my fingers a common solution like a `cdn` field will win over in the future.

MartinKolarik:
>cross my fingers a common solution like a cdn field will win over in the future.

jsDelivr already supports `cdn` but Michael wasn't interested in adding support to unpkg so there isn't much we can do unless his users start asking for it.

CrossEye:
I see no reason not to do this.  :herb:

Any objections?

buzzdecafe:
:cow2:

PR #2298, opened 2017-09-22
Remove precoverage script

mattgrande:
The `precoverage` script just called `pretest`, however `pretest` was removed in PR #1928.

So, because of `precoverage` failing, we've been unable to run `coverage` since.


buzzdecafe:
LGTM, nice catch

:cow2:

CrossEye:
Thank you!

PR #2300, opened 2017-09-23
transduce: fix docs example

MarcoWorms:
Fix the first transduce example that had an incorrect result on it

### DIFF
--- a/src/transduce.js
+++ b/src/transduce.js
@@ -44,7 +44,7 @@ var curryN = require('./curryN');
  *
  *      var numbers = [1, 2, 3, 4];
  *      var transducer = R.compose(R.map(R.add(1)), R.take(2));
- *      R.transduce(transducer, R.flip(R.append), [], numbers); //=> [2, 3]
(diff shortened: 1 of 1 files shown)


adispring:
Why do you think the result is wrong?

MarcoWorms:
I've opened the example on the REPL and i ended up with this [REPL example](http://ramdajs.com/repl/?v=0.24.1#?var%20numbers%20%3D%20%5B1%2C%202%2C%203%2C%204%5D%3B%0Avar%20transducer%20%3D%20R.compose%28R.map%28R.add%282%29%29%2C%20R.take%282%29%29%3B%0AR.transduce%28transducer%2C%20R.flip%28R.append%29%2C%20%5B%5D%2C%20numbers%29%3B%20%2F%2F%3D%3E%20%5B2%2C%203%5D)

I tried to reproduce but in fact it's correct, It was my mistake I probably changed that `add` to 2. Sorry! :P

PR #2305, opened 2017-09-27
Reimplement prop via path

### LINKED ISSUE #1941: Path and prop handle undefined inconsistently

datenreisender:
As discussed in #2059 we are unsatisfied that `prop` and `path` have
different behaviour when passing `undefined` as the object.

We could not yet agree which behaviour we want to have but at least
we want to have the same behaviour. To achieve that, we now
reimplement `prop` with `path`.

So if we ever define the behaviour of `path` then `prop`
automatically also has the same behaviour.

This also closes #1941.

### DIFF
--- a/src/prop.js
+++ b/src/prop.js
@@ -1,4 +1,5 @@
 var _curry2 = require('./internal/_curry2');
+var path = require('./path');
 
 
 /**
@@ -19,4 +20,4 @@ var _curry2 = require('./internal/_curry2');
  *      R.prop('x', {x: 100}); //=> 100
  *      R.prop('x', {}); //=> undefined
  */
-module.exports = _curry2(function prop(p, obj) { return obj[p]; });
+module.exports = _curry2(function prop(p, obj) { return path([p], obj); });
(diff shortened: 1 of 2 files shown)


datenreisender:
Just realised that while I reimplemented prop via path, I did not write a test to ensure that somebody else does break that consistency again unintentionally. 😬  So I added a test for that.

Bradcomp:
While I don't think it should hold up the PR, I think this is a great opportunity to add some property based tests, if only because we have a very clear property we can test for: 

`forall string s, a -> prop(s, a) is equivalent to path([s], a)`

CrossEye:
@Bradcomp, I agree both that we should start looking at where to use property-based testing, and that this shouldn't hold up the PR.

CrossEye:
Thank you very much, @datenreisender!

Issue #2304, opened 2017-09-27
Can someone accept this pr for @types/ramda


BjornMelgaard:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/18763

authors are not answering

buzzdecafe:
nobody here has any control over what goes on over there.

BjornMelgaard:
@buzzdecafe do you mind moving typings from `@types/ramda` to ramda (like in [ramda-adjunct](https://github.com/char0n/ramda-adjunct/blob/master/src/index.d.ts))

CrossEye:
@BjornMelgaard: Few of the regulars here have the expertise to maintain it.  Perhaps, ... @donnut are you interested?

BjornMelgaard:
Closing in favor of [npm-ramda](https://github.com/types/npm-ramda/)

Issue #2303, opened 2017-09-27
No typescript type for pipeP 


BjornMelgaard:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/ramda/index.d.ts

BjornMelgaard:
https://github.com/DefinitelyTyped/DefinitelyTyped/issues/17796

BjornMelgaard:
Closing in favor of [npm-ramda](https://github.com/types/npm-ramda/)

PR #2306, opened 2017-09-29
Feature/integrate ts typings

BjornMelgaard:
@donnut is not answering, so I decided to try myself

I have tried to use [dtslint](https://github.com/Microsoft/dtslint) instead of tslint, but ran into number of errors
Because `tests.ts` still doesn't use `$ExpectType` feature of dtslint, I've decided to do without it

In future I want to integrate number of [open pr's from DefenitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/pulls?utf8=%E2%9C%93&q=is%3Apr%20is%3Aopen%20ramda) and add missing types


BjornMelgaard:
https://github.com/ramda/ramda/issues/2304
https://github.com/ramda/ramda/issues/2303

CrossEye:
While I do think this is a good idea to have, I believe we really should get buy-in from the team that is maintaining this, and find out of they're interested in hosting this here.  

So what do you think, @mdekrey, @LiamGoodacre, @mrdziuban, @sbking, @afharo, @teves-castro, @1M0reBug, @hojberg, @charlespwd, and especially @donnut?  Should these definitions be folded into Ramda? 


I don't know the TS environment and exactly what it would mean to have this hosted here, but my biggest concern is that this becomes one more thing to update at publish time.  Do you have any suggestions  about how to make that work?  Are the tests enough?  Or are they just a start?

So I would also like to hear back from some of the core team on this.  What do you think, @raine, @davidchambers, @paldepind, @TheLudd, @buzzdecafe, @kedashoe, @Bradcomp, @scott-christopher, @asaf-romano?

BjornMelgaard:
IMHO maintaining typings from DefinitelyTyped is hard

CrossEye:
You see, although I've seen DefinitelyTyped, usually in response to issues here, I know nothing about it or how it fits in the TS environment.  They don't look pleasant, but I don't know how much is manual in the process.

charlespwd:
It's pretty common for typescript users to simply add `@types/${package-name}` as a dependency in their project. 

To be honest, if Ramda isn't written in typescript, I don't see why the typings should live in here. Especially if the maintainers are not the consumers of those types.

BjornMelgaard:
@charlespwd because as you see typings become stale and we need to do something with this

Consider `ramda-adjunct` - everyone who wants to add new feature have to change typings

BjornMelgaard:
@CrossEye @charlespwd so what do you think?

kedashoe:
I agree with @CrossEye , it makes more sense to host typescript stuff elsewhere. Otherwise you'll just run into the same trouble with things becoming stale, as none of the active maintainers know typescript well.

I don't know enough about DefinitelyTyped to guess why PRs there might be held up, but it does seem like the best place for it.

BjornMelgaard:
@CrossEye @kedashoe I think you underestimate your knowledge of typescript.  

I've created fork of this [pr](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/18763), and now @charlespwd is in author list and able to accept pr

I've made typings generator and hope you'll like it, and it works, but DefinitelyTyped swears at me because of unused files, maybe you'll reconsider your decision about integration typings to ramda?

BjornMelgaard:
I would say that pr https://github.com/DefinitelyTyped/DefinitelyTyped/pull/20177/ is ready for review if not `.gitignore` in generator dir.

If you still dont want to see typescript in this repository - ok, Its possible to remove `.gitignore` and `package.json` and left only generation script ([like it done in lodash typings](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/lodash/scripts/generate-modules.ts)), but I would rather not do this, because of [dependencies](https://github.com/BjornMelgaard/DefinitelyTyped/blob/2090f51a4367543e0498ecf86ac435fbfaebb0fa/types/ramda/generator/package.json) (and hooks to run generator before publishing would be very handy)

And I would rather not do this because I still think that integration is the right thing. 

> I don't know enough about DefinitelyTyped to guess why PRs there might be held up

Because authors often abandon their projects and no one wants to hold responsibility of broken things. Maintainers of DefenitelyTyped never volunteer to maintain other people typings. I know plenty of awesome prs hanging unaccepted there for eternity and will hang more because of this.

> Otherwise you'll just run into the same trouble with things becoming stale, as none of the active maintainers know typescript well

No, there are plenty of people who know typescript and want to help you, they just need help too.

I understand that you dont want an extra load, but you will kill typings for ramda.

BjornMelgaard:
I did it all in vain. [npm-ramda](https://github.com/types/npm-ramda/) has much better typings. I wont close pr to DefenitelyTyped in case someone want to make use of it.

PR #2309, opened 2017-10-04
add thrush combinator as applyTo

### LINKED ISSUE #1846: add the T combinator : R.thrush ?

kedashoe:
Closes #1846, #1847

### DIFF
--- a/source/index.js
+++ b/source/index.js
@@ -15,6 +15,7 @@ export { default as aperture } from './aperture';
 export { default as append } from './append';
 export { default as apply } from './apply';
 export { default as applySpec } from './applySpec';
+export { default as applyTo } from './applyTo';
 export { default as ascend } from './ascend';
 export { default as assoc } from './assoc';
 export { default as assocPath } from './assocPath';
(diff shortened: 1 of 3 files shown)


source/index.js
```diff
@@ -15,6 +15,7 @@ export { default as aperture } from './aperture';
 export { default as append } from './append';
 export { default as apply } from './apply';
 export { default as applySpec } from './applySpec';
+export { default as applyTo } from './applyTo';
```
Bradcomp: Is this something we will need to add into our PRs in the future? Can we get a build script to do this?

buzzdecafe:
:cow2:

source/index.js
kedashoe: I was in favor of doing this automatically via a script when we first added an index.js file but we decided against it.. I think it is fine to do it manually as I believe in all cases (adding, moving, deleting functions) either tests or building will fail if index.js is not updated.

Bradcomp: It may be worth it at this point to update the instructions in CONTRIBUTING.md

PR #2311, opened 2017-10-04
fix flip to return function with the same arity as the original

### LINKED ISSUE #1714: R.flip always curries two arguments only
``` js
R.flip((a, b, c) => { return `${a} ${b} ${c}`})(1, 2);
// => "2 1 undefined" (should be function)
```

http://goo.gl/dkpLXU

kedashoe:
Closes #1714


buzzdecafe:
:cow2:

PR #2320, opened 2017-10-09
Correct source folder in contributing.md

codeclown:
Seems like instructions in `CONTRIBUTING.md` referred to wrong folder.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,15 +6,15 @@
 
         $ git checkout -b <branch>
 
-3.  If adding a function `R.foo`, define it in __src/foo.js__, require it in
-    __index.js__, and include unit tests in __test/foo.js__. It is not necessary
+3.  If adding a function `R.foo`, define it in `source/foo.js`, export it from
+    `source/index.js`, and include unit tests in `test/foo.js`. It is not necessary
     to include the `@since` annotation for new functions as it will be set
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -6,11 +6,11 @@
 
         $ git checkout -b <branch>
 
-3.  If adding a function `R.foo`, define it in __src/foo.js__, require it in
+3.  If adding a function `R.foo`, define it in __source/foo.js__, require it in
     __index.js__, and include unit tests in __test/foo.js__. It is not necessary
```
kedashoe: `index.js` has also moved, to `source/index.js`, and we now use es6 modules, so it should be `export`ed rather than `require`d

CONTRIBUTING.md
```diff
@@ -6,11 +6,11 @@
...
     to include the `@since` annotation for new functions as it will be set
     during the release process of the following release. If adding an internal
-    function `_foo`, define it in __src/internal/_foo.js__.
+    function `_foo`, define it in __source/internal/_foo.js__.
```
kedashoe: this markdown isn't rendering correctly, let's just go with backticks, ``` `source/internal/_foo.js` ```

codeclown:
@kedashoe I've updated the PR.

CrossEye:
LGTM

PR #2321, opened 2017-10-09
Add examples of negative indexes to R.update and R.adjust

codeclown:
It was not clear that this functionality was supported until I looked at the tests.

### DIFF
--- a/source/adjust.js
+++ b/source/adjust.js
@@ -23,7 +23,6 @@ import _curry3 from './internal/_curry3';
  * @example
  *
  *      R.adjust(R.toUpper, 1, ['a', 'b', 'c', 'd']);      //=> ['a', 'B', 'c', 'd']
- *      R.adjust(R.toUpper)(1)(['a', 'b', 'c', 'd']);      //=> ['a', 'B', 'c', 'd']
  *      R.adjust(R.toUpper, -1, ['a', 'b', 'c', 'd']);     //=> ['a', 'b', 'c', 'D']
  * @symb R.adjust(f, -1, [a, b]) = [a, f(b)]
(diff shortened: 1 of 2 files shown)


CrossEye:
I agree that this is worth doing, but while we're in here, I think that there are too many numbers in the examples.  It's harder to tell what the important index parameter means.

Would anyone object to something like this instead?:

```js
R.adjust(R.toUpper, 1, ['a', 'b', 'c', 'd']);      //=> ["a", "B", "c", "d"]
R.adjust(R.toUpper)(1)(['a', 'b', 'c', 'd']);      //=> ["a", "B", "c", "d"]
R.adjust(R.toUpper, -1, ['a', 'b', 'c', 'd']);     //=> ["a", "b", "c", "D"]
```

codeclown:
I think that's a great idea. Any particular reason why you mixed single and double quotes?

I'll gladly update this PR after clarification on the quotes!

CrossEye:
The double quotes were just pasted from the console.  I usually clean them up when I'm posting such a response, but I simply forgot this time.

codeclown:
Updated to use characters/strings instead of numbers in examples.

kedashoe:
> R.adjust(R.toUpper)(1)(['a', 'b', 'c', 'd']);
> R.update(1)('_')(['a', 'b', 'c']);

Do we need these examples? Just adds noise for me..

codeclown:
I agree, they don't seem necessary nor consistent with other documentation (which generally lacks such currying examples). Shall remove while we're at it?

buzzdecafe:
LGTM! :cow2:

PR #2323, opened 2017-10-09
fix version updates missed by prepublish and patch prepublish script

kedashoe:


scripts/prepublish
```diff
@@ -18,6 +20,9 @@ done
 npm run clean
 npm run build
 
+# build reads version from package.json, which isn't updated until after this script is run by xyz
+sed -i 's#//  Ramda v'${PREVIOUS_VERSION}'#//  Ramda v'${VERSION}'#' -- dist/ramda.js
```
davidchambers: On Darwin this needs to be `-i ''` rather than `-i`. Could we simply remove the version number from __dist/ramda.js__ to avoid this complexity?

scripts/prepublish
```diff
@@ -5,6 +5,8 @@ slash=/
 README="$(cat README.md)"
 README="${README//${PREVIOUS_VERSION}${slash}ramda.min.js/${VERSION}${slash}ramda.min.js}"
 README="${README//${PREVIOUS_VERSION%.*}${slash}ramda.min.js/${VERSION%.*}${slash}ramda.min.js}"
+README="${README//${PREVIOUS_VERSION}${slash}dist${slash}ramda.min.js/${VERSION}${slash}dist${slash}ramda.min.js}"
+README="${README//${PREVIOUS_VERSION%.*}${slash}dist${slash}ramda.min.js/${VERSION%.*}${slash}dist${slash}ramda.min.js}"
```
davidchambers: Should these lines replace the preceding two lines, or do we really need to perform four substitutions?

scripts/prepublish
kedashoe: Hopefully no tooling depends on it, so I guess that means we could? Seems kind of nice though?

I could also just use the same bash replacement technique we use for the README.

scripts/prepublish
kedashoe: Currently we need three, the two new lines but only one of the old lines.

davidchambers: Okay. Let's remove whichever substitution is no longer necessary.

scripts/prepublish
kedashoe: Updated to use bash replacement instead of sed. cc @buzzdecafe @CrossEye do you want to keep this line stating the version in `dist/ramda.js` or get rid of it?

CrossEye: I'd rather keep it.

kedashoe: > I'd rather keep it.

:+1: ok

PR #2221, opened 2017-06-29
fix bookmarklet code block + raw

aretecode:

PR #2326, opened 2017-10-11
Put the whole bookmarklet code inside markdown snippet

feimosi:
I don't know if this was made on purpose to break the code in half, but looks like GitHub handles such a long line without a problem.

I've also seen it was fixed in the past in #2221.


buzzdecafe:
def not on purpose, not sure how that is happening. 😕 

thanks for the fix
:cow2:

feimosi:
You're welcome! Hopefully it won't happen again 🙂

PR #2327, opened 2017-10-11
Fix trim and keys which cause isses when generating document

cht8687:
In current 0.25.0 document, there are extra _ applied to [trim](http://ramdajs.com/docs/#_trim) and [keys](http://ramdajs.com/docs/#_keys).

This is caused by the variable name which jsdocs used to generate.

keysandtrim


buzzdecafe:
:cow2:

Thank you

CrossEye:
Thanks!  :herb:

Issue #2328, opened 2017-10-11
`toString` & `contains` dist API has something wrong


adispring:
`toString` becomes `toString$1`
`contains` becomes `contains$1`

<img width="453" alt="screen shot 2017-10-11 at 11 21 37 pm" src="https://user-images.githubusercontent.com/9584123/31450040-0b11e71e-ae6e-11e7-94fd-0ddce52528a0.png">

buzzdecafe:
i wonder if this is related to #2327

adispring:
In https://github.com/ramda/ramda/pull/2327, trim & keys spell typos are in `source` directory, but these two typos are in `dist` directory. This seems due to `build:umd` scripts or `rollup` which produces dist/ramda.js

<img width="1319" alt="screen shot 2017-10-12 at 12 01 33 am" src="https://user-images.githubusercontent.com/9584123/31452293-24a3a23e-ae74-11e7-9c38-2fa5e454fabe.png">

CrossEye:
ping @Andarist: Any ideas?

Andarist:
Yeah, rollup sometimes renames identifiers like that to avoid conflicts between concatenated (scope hoisted) modules.

I had no idea that documentation is built directly from the umd build (which I guess is the case here). Gonna look into fixing this soon!

Andarist:
So like I've stated above - rollup has renamed them due to name conflicts caused by [this](https://github.com/ramda/ramda/blob/ad09c10ebef5c6d20203ed44d2bce5a86599de36/source/keys.js#L15) and [this](https://github.com/ramda/ramda/blob/ad09c10ebef5c6d20203ed44d2bce5a86599de36/source/internal/_isArguments.js#L4).

Imho trimming those suffixes should be done in the docs script to avoid this also in the future (alternative would be to rename those 2 mentioned functions, but I dont feel like its right way to solve this). 

Need some time to figure out how this things is generated and then gonna send a PR with a fix.

cht8687:
Hi guys, I've found the root cause for this and commited a fix for it.

https://github.com/ramda/ramda/pull/2334

CrossEye:
Thanks, @cht8687, @adispring, and @Andarist!

PR #2334, opened 2017-10-12
Fix contains and toString when generate docs

cht8687:
This is related to https://github.com/ramda/ramda/issues/2328.

Reason behind this is that, some internal functions used the contains and toString namespaces, so that rollup can only add $1 to the real functions' names.

Below is the doc which I used this fix to generate:

contains
tostring

### DIFF
--- a/source/internal/_isArguments.js
+++ b/source/internal/_isArguments.js
@@ -1,10 +1,10 @@
 import _has from './_has';
 
 
-var toString = Object.prototype.toString;
+var toStringFunc = Object.prototype.toString;
 var _isArguments = function() {
-  return toString.call(arguments) === '[object Arguments]' ?
-    function _isArguments(x) { return toString.call(x) === '[object Arguments]'; } :
+  return toStringFunc.call(arguments) === '[object Arguments]' ?
+    function _isArguments(x) { return toStringFunc.call(x) === '[object Arguments]'; } :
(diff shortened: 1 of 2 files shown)


Andarist:
Dont like this change as it treats fixes the symptom rather than root cause. The more appropriate fix is already pending in the PRs to the docs repo

kedashoe:
> Dont like this change as it treats fixes the symptom rather than root cause. The more appropriate fix is already pending in the PRs to the docs repo

:+1: We could add something to our test suite that checks for name clashes, but baring that, changing the way the docs are built is a better solution imho

CrossEye:
Oh, now that I've gotten through the Ramda core issues and started looking at the docs ones, I agree completely.  I had assumed that we had hit a fundamental problem with JSDoc.  (I'll always believe the worst of that tool!) . This is entirely unnecessary. 

That's not to say that there was anything wrong with the code in the PR, @cht8687.  There's simply a better way to solve this issue.  Thank you for this contribution.

cht8687:
thank you @CrossEye for your comments. I totally agree with what you said. Only one issue is that, I realised that if we declare the two - toStrings and contains locally, the generated build will append $1 to all contains and toString. There are 45+ appearance of this kind of changes in generated build file. But again, thanks @Andarist for fixing this.

PR #2338, opened 2017-10-13
Use Object.is as the implementation of R.identical if available

asaf-romano:
Object.is is identical to R.identical. Similar to what we did with
Object.assign few versions ago, this PR sets R.identical to
~~curry(Object.is) when possible, and polyfills it otherwise.

Also in this PR:
1. Simply “polyfilling” conventions (which I am to blame for, btw)
2) In R.equals use _objectIs rather than identical, which is what we
should do at least until curry performance issues are addressed. This
code path is simply evaluated too much for any abstraction to survive.

### DIFF
--- a/source/identical.js
+++ b/source/identical.js
@@ -1,3 +1,4 @@
+import _objectIs from './internal/_objectIs';
 import _curry2 from './internal/_curry2';
 
 
@@ -5,6 +6,8 @@ import _curry2 from './internal/_curry2';
  * Returns true if its arguments are identical, false otherwise. Values are
  * identical if they reference the same memory. `NaN` is identical to `NaN`;
  * `0` and `-0` are not identical.
+ *
+ * Note this is merely a curried version of ES6 `Object.is`.
  *
  * @func
  * @memberOf R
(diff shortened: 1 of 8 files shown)


source/identical.js
```diff
@@ -6,6 +7,8 @@ import _curry2 from './internal/_curry2';
  * identical if they reference the same memory. `NaN` is identical to `NaN`;
  * `0` and `-0` are not identical.
  *
+ * Note this is merely a curried version of ES6 `Object.is`.
```
polytypic: 👍

CrossEye:
I assume the main rationale is performance.  If so, do you have any before-and-after numbers for this?

I always hesitate to assume that the built-in functions are faster than custom ones; too often that's not the case.

asaf-romano:
Here it's the same algorithm, with all same edge cases (unlike the infamous sparse array case). I doubt there's any chance a native implementation can be slower, but I'll try to come up with numbers later today.

And it's not just performance. Once we ditch ES3 support, I see no reason to polyfill this at al. It's the spec-algorithm as-is.

CrossEye:
Thanks, @asaf-romano!

asaf-romano:
Kinda-sorta: https://goo.gl/6NkpcT

CrossEye:
Well, in my few tests, it seems that this is 10 - 50% faster than the current version, although the numbers vary quite a bit between FF and Chrome (both on Mac.)

But this means that at the very least we're not moving backward on performance.  (It does happen; some years ago. `R.map` was tremendously faster than `Array.prototype.map`... mostly because it was simpler and didn't have to handle too many odd cases.  It would have been surprising if the same algorithm was faster in our implementation than in the native ones.  Surprising, but not shocking.

In any case, :+1: from me.

buzzdecafe:
:cow2:

Issue #2264, opened 2017-08-26
R.into does not work properly with R.slice


xgbuils:
Hi,

I picked up the example of `R.into` defined in [documentation](https://goo.gl/QtcCgq) and I changed `R.take(2)` per `R.slice(1, 3)` and it throws [this error](https://goo.gl/JSzWKC):

``` javascript
t.@@transducer/step is not a function
```

I expected:
``` javascript
[3, 4]
```

However `R.compose` works well:

``` javascript
R.compose(R.map(R.add(1)), R.slice(1, 3))([1, 2, 3, 4]) // [3, 4]
```

Bradcomp:
Hi @xgbuils,

Right now, `R.slice` is not a function that works with transducers. Looking at you issue, I suspect it probably should be though. 

`take` and `drop` both support transducers, so you should be able to wire something up with those two functions like [this](https://goo.gl/XzXnqS).

Might you be interested in looking into adding transducer support to `slice`?

asaf-romano:
Let's keep this open - no reason not to support this behavior.

PR #2342, opened 2017-10-15
Added possibility to use transducers with slice

Deadid:
Refers to #2264.
Added posibility to use transucers with slice. 
Test made from issue scenario.


Bradcomp:
Hi @Deadid, thanks for the PR!  This looks pretty good, but can you fix the CI error you have? Also, can you add a note into the docs for `slice` along the lines of 
> Acts as a transducer if a transformer is given in list position.

Deadid:
@Bradcomp  Done. Build was red because of lack of spaces :) Also I've added doc.

kedashoe:
Does this support negative indexes like current `slice` does?

buzzdecafe:
I wish we did not support negative indices.

Deadid:
I assume no, but if it needed to merge this PR I could take a look at them later.

Issue #2333, opened 2017-10-12
Add rangeWith?


adispring:
rangeWith :: number -> number -> stepFn -> [number],

```js
rangeWith(1, 7, multiply(2)) => [1, 2, 4]
rangeWith(1, 7, inc) => [1, 2, ..., 6]
rangeWith(7, 1, dec) => [7, 6, ..., 2]
rangeWith(2, 9, square) => [2, 4, 8]
...
```

CrossEye:
If nothing else, http://fr.umio.us/ranging-near-and-far/

But I don't thoroughly understand your examples.  Should the last one be something like this?:

```js
rangeWith(2, 20, square) //=> [2, 4, 16]
```

And how would you  distinguish 

```js
rangeWith(7, 1, dec); //=> [7, 6, ..., 2]
```

where the iteration stops when we hit something  *less than or equal to*  the stop value of `1` from the others in which it stops when we hit something no *greater than or equal to* the stop value ?

Do you have a suggested implementation for this?

Or maybe I should just go back to my original response: http://fr.umio.us/ranging-near-and-far/.

adispring:
Yes, my last example is wrong.

> rangeWith(7, 1, dec); //=> [7, 6, ..., 2]
We can compare the first two arguments to decide the list is increasing or decreasing. I'll see the article.

CrossEye:
> We can compare the first two arguments to decide the list is increasing or decreasing. I'll see the article.

But what if the function is not monotonic?  If it increases then decreases or vice versa?

buzzdecafe:
i'd rather see the "stepping" function passed as the first param. then you can generate the `range` function you need.

I see @crosseye's point: how do we know when to stop generating the range?. Would this really require a terminating predicate? Er, that is getting a bit complicated

Qquanwei:
I like `step` function as other programming language
```
step :: Number -> Number -> Number -> [Number]
```
step(begin, width, end)
step(0, 10, 30) => [0,10,20,30]
step(30, -10, 0) => [30, 20, 10, 0]

scott-christopher:
I could see me using reaching for something like `step :: Number -> Number -> Number -> [Number]` more often than `rangeWith`, partly because we already have `R.unfold` for generating the sequences with a provided step function.

CrossEye:
[...]

PR #2343, opened 2017-10-16
R.range improvement

ripichipi:
Added Increasing and Decreasing capabilities.
Added Inclusive `to` parameter.

### DIFF
--- a/source/range.js
+++ b/source/range.js
@@ -3,8 +3,8 @@ import _isNumber from './internal/_isNumber';
 
 
 /**
- * Returns a list of numbers from `from` (inclusive) to `to` (exclusive).
- *
+ * Returns a list of numbers from `from` (inclusive) to `to` (inclusive).
+ * The range can be either increasing or decreasing. 
  * @func
  * @memberOf R
  * @since v0.1.0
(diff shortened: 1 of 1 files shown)


buzzdecafe:
I am :-1: on the inclusive `to`. I like that `to - from` gives the length of the output list. 

I am also squeamish about testing `from > to` and behaving differently. I would rather have something like `rangeWith(f, from, to)` and then define `range` e.g. as  `range = rangeWith(inc)`

ripichipi:
Hi @buzzdecafe  thanks for your comment. What the "f" parameter stand for?
`rangeWith(?, from ,to)`

buzzdecafe:
i imagine `f` in that signature would be a function `Number -> Number`, so increasing range could be `rangeWith(x => x + 1)`; decreasing range: `rangeWith(x => x - 1)`, etc. This also makes other `range` functions possible, e.g. with different `step` sizes, etc.

CrossEye:
I'm also :-1: on this, for the reasons I wrote in that article.  Even `rangeWith` bothers me, as discussed recently in #2333, although I imagine we could get that to work if we included both a `step` function and an `end` predicate.  But that seems far too complex, when any reasonable range could likely be created from a linear transformation of `[1 ... n]`.

@buzzdecafe: who's ripichipi?

buzzdecafe:
whoops good question. edited.

PR #2344, opened 2017-10-16
Guard null and undefined before accessing prop in _IsTransformer

kurtmilam:
_IsTransformer was naively checking for an object property on an
argument without first ensuring that the argument was neither null nor
undefined.

Example:
```js
tap( console.log, null ) //-> Cannot read property '@@transducer/step' of null
tap( console.log, undefined ) //-> Cannot read property '@@transducer/step' of undefined
```

[REPL](https://goo.gl/JT7LE2)

### DIFF
--- a/source/internal/_isTransformer.js
+++ b/source/internal/_isTransformer.js
@@ -1,3 +1,3 @@
 export default function _isTransformer(obj) {
-  return typeof obj['@@transducer/step'] === 'function';
+  return obj != null && typeof obj['@@transducer/step'] === 'function';
 }
--- a/test/map.js
+++ b/test/map.js
@@ -1,6 +1,7 @@
 var listXf = require('./helpers/listXf');
 
 var R = require('..');
+var assert = require('assert');
 var eq = require('./shared/eq');
 var Id = require('./shared/Id');
 
@@ -40,6 +41,11 @@ describe('map', function() {
       f: add1,
       xf: listXf
     });
+  });
+
+  it('throws a TypeError on null and undefined', function() {
+    assert.throws(function() { return R.map(times2, null); }, TypeError);
+    assert.throws(function() { return R.map(times2, undefined); }, TypeError);
   });
 
   it('composes', function() {
--- a/test/tap.js
+++ b/test/tap.js
@@ -11,6 +11,8 @@ describe('tap', function() {
     var f = R.tap(R.identity);
     eq(typeof f, 'function');
     eq(f(100), 100);
+    eq(f(undefined), undefined);
+    eq(f(null), null);
   });
 
   it("may take a function as the first argument that executes with tap's argument", function() {


source/internal/_isTransformer.js
```diff
@@ -1,3 +1,5 @@
 export default function _isTransformer(obj) {
-  return typeof obj['@@transducer/step'] === 'function';
+  return obj !== null &&
+         typeof obj !== 'undefined' &&
```
buzzdecafe: could use shorthand `obj != null` here

kurtmilam: I had that first, then switched to the more explicit checks. I'll go with whatever you prefer.

kurtmilam: My reasoning: I try to avoid depending on magic coercion, but I'm happy to stick to house style here, whatever that may be.

buzzdecafe: i believe that `==` is house style in the `null || undefined` case. I am sympathetic with you wanting to be explicit.

kedashoe: :+1: for `!= null`

kedashoe:
What happens now with `tap`? What about `map` or anything else that does this check? Should we add a test for these?

source/internal/_isTransformer.js
CrossEye: I agree that this is consistent.  I'm not sure how or why we settled on this style as I would also prefer to be explicit.   But I don't find it a big deal.

kurtmilam:
@kedashoe 
Currently:
```js
map(identity, null) // -> Cannot read property 'fantasy-land/map' of null
```
Looks like we're naively checking for a `fantasy-land/map` method on the object before we run the `_isTransformer` check in `map`, so I don't think there will be any change. But we should probably consider adding a similar check in `internal/_dispatchable.js` before we check for the existence of `fantasy-land`-prefixed methods.

After the fix:
```js
tap(identity)(undefined) // -> undefined
tap(identity)(null) // -> null
```

CrossEye:
@kedashoe: Agreed.  We should probably add a check for one or two public functions here.  I don't feel any need to do this for all such functions, as our change is to remove an undocumented exception.  But one or two would be good.

I don't think the test of the internal functions should be included though.

kurtmilam:
Apologies, I got my IDE files mixed in with the amended commit. I'll clean that up.

@CrossEye I have added tests for `tap`. Do you (or @kedashoe or @buzzdecafe ) have a preference for a second public function to test?

CrossEye:
No strong preference, but as `map` has already been brought up...

kedashoe:
> map(identity, null) // -> Cannot read property 'fantasy-land/map' of null

Seems like an improvement, I'm fine with this, will be easier to track down errors now. Adding a test to `map` is fine with me.

.idea/terminal.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kedashoe: local dev files snuck into PR?

CrossEye: https://github.com/ramda/ramda/pull/2344#issuecomment-337034984

kurtmilam:
@kedashoe @CrossEye I want to be sure that I've communicated clearly how this change affects `map`.

Namely, it has no effect at all on `map` or any other functions wrapped in `isDispatchable` as long as the `methodNames` argument is not empty.

Since `isDispatchable` already naively checks for the existence of the method properties signified in the `methodNames` argument before it calls `isTransformer`, it already throws a TypeError when the list/last argument is `null` or `undefined`. It will continue to throw the exact same TypeError after this change is merged.

kurtmilam:
To continue, the interesting functions whose behavior will be affected by this change are those wrapped in `isDispatchable` with an empty `methodNames` argument.

I took a look at `dropLast`, which matches this pattern.

Current behavior ([REPL](https://goo.gl/7jok5g)):
```js
dropLast( 1, null ) //-> TypeError: Cannot read property '@@transducer/step' of null
dropLast( 1, undefined ) //-> TypeError: Cannot read property '@@transducer/step' of undefined
```
New behavior:
```js
dropLast( 1, null ) //-> TypeError: Cannot read property 'length' of null
dropLast( 1, undefined ) //-> TypeError: Cannot read property 'length' of null
```

Do you think we need to look at other functions that fit this pattern? Do you think we need to add test for functions other than `map` and `tap`? Or are we satisfied by the fact that the handful of new tests plus the large set of existing tests passes after the change?

kurtmilam:
I believe this is a comprehensive list of functions that will be affected by this change:
`aperture`, `dropLast`, `dropLastWhile`, `dropRepeats`, `dropRepeatsWith`, `findIndex`, `findLast`, `findLastIndex`, into, `reduceBy`, `tap`.

kurtmilam:
Out of an abundance of caution, I compared the current behavior of each of the aforementioned functions (https://github.com/ramda/ramda/pull/2344#issuecomment-337170421).

`tap` is the only one that changes materially. The other all already throw TypeErrors before the change and will continue to throw TypeErrors after the change.

The only difference (and I didn't keep good enough track to say for sure) is that some of the functions may throw a different TypeError after the change than they threw before it.

With that, I'm satisfied that this change is safe and ready to be merged.

kurtmilam:
Last note (I think): Since `tap` is the only function materially affected by this change, another option would be to move the guard into `tap` to avoid adding the performance overhead of the check to the other functions that depend on `_isTransform`.

Bradcomp:
> Since tap is the only function materially affected by this change, another option would be to move the guard into tap to avoid adding the performance overhead of the check to the other functions that depend on _isTransform.

I think the change to `_isTransformer` is beneficial, even though it only materially changes `tap`. In _most_ cases, I don't think a boolean check should throw.

CrossEye:
I think it's better to keep this check in `_isTransformer`, or some later code that depends on it could end up breaking the same way.

kurtmilam:
I'm totally fine with leaving the check where it is. Wanted to throw the option out, just in case, based on the results of my testing :)

@CrossEye I think I addressed all open concerns with the latest commit. The IDE files are gone, the internal test is gone, and I added tests for `map`. If there are any remaining issues to be fixed, let me know!

buzzdecafe:
:cow2:

dashmug:
Do we have timeline when this can be included in a release?

Latest version is 0.25.0 (Oct. 10, 2017) and this fix was merged to master last Oct. 19, 2017.

dashmug:
@CrossEye Any plans to release this?

wolfflow:
Coming up with the same question

PR #2347, opened 2017-10-17
integrate sanctuary-type-classes

davidchambers:
Supersedes #1938 (exactly one year later).

Unfortunately the Ramda codebase is littered with special cases, many of which are covered by the test suite. This makes it impossible to adopt principled support for Fantasy Land without making breaking changes. I've used my judgement, but it's possible this pull request should introduce fewer or even more breaking changes. @CrossEye, we *really* need you to tell us how you want dispatching to work. My vote is to remove dispatching from the codebase entirely (removing support for transducers and ad hoc polymorphism), leaving us with only indirect dispatching to `fantasy-land/`-prefixed methods (provided by sanctuary-type-classes).


source/ap.js
```diff
@@ -30,16 +29,5 @@ import map from './map';
...
-      applyX['fantasy-land/ap'](applyF) :
-    typeof applyF.ap === 'function' ?
-      applyF.ap(applyX) :
-    typeof applyF === 'function' ?
-      function(x) { return applyF(x)(applyX(x)); } :
-    // else
-      _reduce(function(acc, f) { return _concat(acc, map(f, applyX)); }, [], applyF)
-  );
-});
+var ap = _curry2(Z.ap);
```
davidchambers: Changes such as this one please me greatly. :smile:

source/internal/_isArguments.js
```diff
@@ -2,10 +2,10 @@ import _has from './_has';
 
 
 var toString = Object.prototype.toString;
-var _isArguments = function() {
+var _isArguments = (function() {
   return toString.call(arguments) === '[object Arguments]' ?
     function _isArguments(x) { return toString.call(x) === '[object Arguments]'; } :
     function _isArguments(x) { return _has('callee', x); };
-};
+}());
```
davidchambers: I should open a separate pull request to fix this bug.

davidchambers: #2348

CrossEye:
I thought I'd made it clear how I'd prefer to do dispatching.  But I do not feel this is my decision alone.  And if there's a near-consensus on a different approach, I won't stand in its way.

My take is that functions that might reasonably dispatch do so to all objects presenting the requisite method.  Moreover, I would rather expand on that than reduce it.  I mean this in two ways: by expanding the functions we actually dispatch, and by dispatching to more names (`fmap`, `foldl`, etc.)

Without suggesting that this is the exact syntax here, I mean 

```js
var map = _curry2(_dispatchable(['fantasy-land/map', 'map', 'fmap'] ...
```

```js
var reduce = _curry3(_dispatchable(['fantasy-land/reduce', 'reduce', 'foldl']...
```

```js
var filter = _curry2(_dispatchable(['filter']...
```

I feel that FantasyLand should be in an only slightly privileged place for us.  It is privileged because it's at the moment the only significant spec for important types.  But it's only slightly privileged because other types are quite reasonable.  I can see performing many of our list functions on objects that present the iteration protocol.  I can see performing `filter` on a Tree type that doesn't present FL methods, but does have its own `filter.  I can see performing `add` on someone's Complex Number type.  I can see clearly now, the rain is gone... 

Something like the above does that reasonably well.

davidchambers:
This tells me roughly what you have in mind, Scott. Thank you. It's far from a specification, though. I have many questions:

  - Which functions should support dispatching?
  - To which methods should they dispatch?
  - For functions which can dispatch to more than one method, what is the precedence?
  - Should we make any exceptions? For example, if we are to dispatch to `map` should we make an exception for `Array#map`?
  - Do we attempt to support old versions of Fantasy Land? For example, if we dispatch to `ap` should we use the old argument order?
  - Would derived functions ever dispatch? For example, `sequence` is trivially derived from `traverse` which should dispatch to `fantasy-land/traverse`. Should we also dispatch to `sequence`?

I'm strongly opposed to unprincipled dispatching. Ramda has become—in my view—a hairball of complexity, despite our best intentions. Special cases mean that many Ramda functions—like many Clojure functions—are not *abstractions*. Dispatching to more methods will only worsen this problem.

CrossEye:
> It's far from a specification, though.

Certainly, and it was not meant to be one.

> I'm strongly opposed to unprincipled dispatching. 

I don't know what would constitute principled dispatching.  The language does not allow us to know if some object's `'fantasy-land/map'` method takes a function which converts its value type into another value type.  And if it does, we don't know that it returns another container of the same type.  And if it does that, how do we know that it satisfies the composition law?  (That last affects even Haskell, but the others are more specific to a JS library.)

----------


>  - Which functions should support dispatching?
>  - To which methods should they dispatch?
>  - For functions which can dispatch to more than one method, what is the precedence?

Although all of these would be handled in an _ad hoc_ manner, I think that we should dispatch as many functions as we reasonably can, that they should dispatch to the fantasy-land method, to the method with the same name as the function, to other common names for that function, in that order.  I don't have any suggestions for how to order that last group, but also don't think it's likely to be common or that it would be problematic to solve when we need to.

>  - Should we make any exceptions? For example, if we are to dispatch to `map` should we make an exception for `Array#map`?

Yes, I think that's fairly core to Ramda.  We would supply the implementations for built-in types.  While that might sometimes involve calling a built-in method, that's our choice.

I actually want to go further with this in terms of our organization.  I would like to separately define an internal type for List/Array, for Object, for Function, for Number, etc, and for each that has a `map` implementation, I would define it there, and then in some way to combine them along with the dispatch names into a single public `map` function.

This clearly is a post-1.0 idea, though.

>  - Do we attempt to support old versions of Fantasy Land? For example, if we dispatch to `ap` should we use the old argument order?

I'd say no.  Each version of Ramda should explicitly support a fixed version of FantasyLand.


>  - Would derived functions ever dispatch? For example, `sequence` is trivially derived from `traverse` which should dispatch to `fantasy-land/traverse`. Should we also dispatch to `sequence`?

I would very much like to do this as much as possible.  So if the type has `ap` and `map`, then we could supply a `lift` if the type doesn't have one of its own.  But I'm not clear yet on the details.

davidchambers:
> I would like to separately define an internal type for List/Array, for Object, for Function, for Number, etc, and for each that has a `map` implementation, I would define it there, and then in some way to combine them along with the dispatch names into a single public `map` function.

This is exactly how sanctuary-type-classes is structured. `map` is defined like so:

> ```javascript
> function map(f, functor) {
>   return Functor.methods.map(functor)(f);
> }
> ```

`Array#fantasy-land/map` is defined like so:

> ```javascript
> //  Array$prototype$map :: Array a ~> (a -> b) -> Array b
> function Array$prototype$map(f) {
>   return this.map(function(x) { return f(x); });
> }
> ```

`Object#fantasy-land/map` is defined like so:

> ```javascript
> //  Object$prototype$map :: StrMap a ~> (a -> b) -> StrMap b
> function Object$prototype$map(f) {
>   var result = {};
>   forEachKey(this, function(k) { result[k] = f(this[k]); });
>   return result;
> }
> ```

`Function#fantasy-land/map` is defined like so:

> ```javascript
> //  Function$prototype$map :: (a -> b) ~> (b -> c) -> (a -> c)
> function Function$prototype$map(f) {
>   var functor = this;
>   return function(x) { return f(functor(x)); };
> }
> ```

I've been advocating a sanctuary-type-classes dependency for the past year as I see no reason to recreate the library. Ramda, Sanctuary, and other projects all benefit from sharing a well-tested standard library for Fantasy Land.

source/map.js
```diff
@@ -42,19 +43,5 @@ import keys from './keys';
...
-    case '[object Object]':
-      return _reduce(function(acc, key) {
-        acc[key] = fn(functor[key]);
-        return acc;
-      }, {}, keys(functor));
-    default:
-      return _map(fn, functor);
-  }
-}));
+var map = _curry2(_dispatchable(['map', 'fmap', '(<$>)', '<$>'], _xmap, Z.map));
```
davidchambers: `(<$>)` and `<$>` are tongue-in-cheek. I assume we should drop these, Scott?

source/filter.js
```diff
@@ -21,7 +21,10 @@ import keys from './keys';
  * @memberOf R
  * @since v0.1.0
  * @category List
- * @sig Filterable f => (a -> Boolean) -> f a -> f a
+ * @sig (a -> Boolean) -> [a] -> [a]
+ * @sig (a -> Boolean) -> StrMap a -> StrMap a
+ * @sig (Applicative f, Foldable f, Monoid (f a)) => (a -> Boolean) -> f a -> f a
+ * @sig (Alternative m, Monad m) => (a -> Boolean) -> m a -> m a
```
CrossEye: Although I'm greatly in favor of expanding our type signatures in this way, and I'm not particularly worried about scaring people with all the types (except maybe `StrMap`: "What happened to my `Object`?!" :smile:), I am worried about losing the description of "unprincipled dispatching" in `Filterable f => (a -> Boolean) -> f a -> f a`. where `Filterable` would point to a description that specifies only having for Type T, `filter:: t a ~> (a -> Boolean) -> t a`.

source/internal/_createPartialApplicator.js
```diff
@@ -1,11 +0,0 @@
...
-import _curry2 from './_curry2';
-
-
-export default function _createPartialApplicator(concat) {
-  return _curry2(function(fn, args) {
-    return _arity(Math.max(0, fn.length - args.length), function() {
-      return fn.apply(this, concat(args, arguments));
-    });
-  });
-}
```
CrossEye: I do love getting rid of many `_internal` functions.  They've always rubbed me the wrong way.

test/sequence.js
```diff
@@ -8,29 +8,29 @@ var eq = require('./shared/eq');
...
-    eq(R.sequence(R.of, [[1], [2, 3, 4]]), [[1, 2], [1, 3], [1, 4]]);
-    eq(R.sequence(R.of, [[1, 2], [3, 4]]), [[1, 3], [1, 4], [2, 3], [2, 4]]);
-    eq(R.sequence(R.of, [[1, 2, 3], [4]]), [[1, 4], [2, 4], [3, 4]]);
-    eq(R.sequence(R.of, [[1, 2, 3, 4], []]), []);
+    eq(R.sequence(Array, []), [[]]);
+    eq(R.sequence(Array, [[], [1, 2, 3, 4]]), []);
+    eq(R.sequence(Array, [[1], [2, 3, 4]]), [[1, 2], [1, 3], [1, 4]]);
+    eq(R.sequence(Array, [[1, 2], [3, 4]]), [[1, 3], [1, 4], [2, 3], [2, 4]]);
+    eq(R.sequence(Array, [[1, 2, 3], [4]]), [[1, 4], [2, 4], [3, 4]]);
+    eq(R.sequence(Array, [[1, 2, 3, 4], []]), []);
```
CrossEye: Damn, this was *always* better!

test/unnest.js
```diff
@@ -1,28 +1,13 @@
...
 
 
 describe('unnest', function() {
 
   it('only flattens one layer deep of a nested list', function() {
-    var nest = [1, [2], [3, [4, 5], 6, [[[7], 8]]], 9, 10];
-    eq(R.unnest(nest), [1, 2, 3, [4, 5], 6, [[[7], 8]], 9, 10]);
-    nest = [[[[3]], 2, 1], 0, [[-1, -2], -3]];
-    eq(R.unnest(nest), [[[3]], 2, 1, 0, [-1, -2], -3]);
-    eq(R.unnest([1, 2, 3, 4, 5]), [1, 2, 3, 4, 5]);
```
CrossEye: I agree that the rest of these can go.  But I think this edge-case provides a useful description of the operation.  What happens if there are *no* sublists?

test/without.js
```diff
@@ -17,15 +19,10 @@ describe('without', function() {
   });
 
   it('has R.equals semantics', function() {
-    function Just(x) { this.value = x; }
-    Just.prototype.equals = function(x) {
-      return x instanceof Just && R.equals(x.value, this.value);
-    };
-
```
CrossEye: It's definitely a nice thing to remove these types from the individual test.  But the win is not that big, as we could simply place them in `test/shared` (and probably already should have.)

CrossEye:
> Ramda, Sanctuary, and other projects all benefit from sharing a well-tested standard library for Fantasy Land.

While I agree in part, there is another part of me that feels that having distinct implementations is decidedly useful.  When some on the FantasyLand team started trying to push people to use the implementations on their site as the defaults, I had no interest, because I felt that having groups trying different approaches improved the overall environment.

Here I'm a little more mixed, partly because the main goal is to provide exactly the FL specification for built-in types (well, and "several functions derived from these functions.")  While there are many ways to organize such code, the specification will likely not provide much wiggle-room in the core code.  It's not clear just how much benefit competing implementations would offer.

But there is one obvious benefit: the ability to go ones own way.  I am not wedded to the fact that Ramda has no dependencies.  That mostly fell out of the way it was built, and was not a real design goal.  But it has been very nice.  If we need a change in Ramda, we can make it.  With this, if there was a disagreement with how sanctuary-type-classes was handling something, and if discussion didn't reach agreement, Ramda would be left with the option of accepting a decision it didn't like or forking.  That's never a pleasant choice.

Could this happen?  I think so.

Ramda has always stuck a little more closely to Javascript's warts.

```js
R.lt('5', 7) //=> true
R.lt(7, '5') //=> false
```

Presumably this would generate a type error in the updated code.  I don't want to defend the Ramda version in particular, but Ramda's no-handholding philosophy and its closer ties to the way the language does things could easily lead to actual conflicts with sanctuary-type-classes stricter approach.

That does not mean that I'm opposed.  In fact, I'm slightly leaning toward accepting this.  But I am feeling fairly cautious about this.

David, do you want to try to get this in for 1.0, or would it be all right with you to get 1.0 out the door and then include this among the likely more drastic changes in a 2.0 version?

source/map.js
davidchambers: I'll remove the symbolic names. :)

source/filter.js
davidchambers: I don't think we should use type classes for unprincipled dispatching. It's okay† to decide to dispatch to a `quux` method, and we should document this if we do so, but I'd oppose inventing a lawless type class named Quuxable and using `Quuxable q =>` in type signatures.

If you'd like a lawful Filterable type class so we could remove the long-winded signatures of the derivations, please support fantasyland/fantasy-land#274. :)

---

† _As I've stated, I'm opposed to dispatching to methods which don't claim to conform to a specification, but I'll set aside this opposition for this discussion._

source/internal/_isPlainObject.js
```diff
@@ -0,0 +1,4 @@
+export default function _isPlainObject(x) {
+  return Object.prototype.toString.call(x) === '[object Object]' &&
+         x.constructor.name === 'Object';
+}
```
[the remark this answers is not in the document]
davidchambers: Are you suggesting we replace `x.constructor.name` with `_functionName(x.constructor)`? If so, consider it done. :)

source/internal/_xdropLastWhile.js
```diff
@@ -18,7 +19,7 @@ XDropLastWhile.prototype['@@transducer/step'] = function(result, input) {
                        : this.flush(result, input);
 };
 XDropLastWhile.prototype.flush = function(result, input) {
-  result = _reduce(
+  result = Z.reduce(
     this.xf['@@transducer/step'],
     result,
     this.retained
```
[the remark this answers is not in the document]
davidchambers: I used `Z` functions wherever possible. They're simpler than the Ramda equivalents. They do not support transducing, they dispatch only to `fantasy-land/`-prefixed methods, and they don't contain special cases (vestigial or otherwise).

The sites at which I could *not* replace the application of a Ramda function with the application of its `Z` equivalent are where the dragons live. :dragon:

test/sequence.js
davidchambers: I agree. Type representatives provide a tidy home for ancillary functions such as `Array#fantasy-land/of`.

test/unnest.js
davidchambers: > What happens if there are *no* sublists?

Let's ask Haskell:

```
> import Control.Monad

> :t join
join :: Monad m => m (m a) -> m a

> join [[[1], [2]], [[3], [4]]]
[[1], [2], [3], [4]]

> join [[1, 2], [3, 4]]
[1, 2, 3, 4]

> join [1, 2, 3, 4]
    • Non type-variable argument in the constraint: Num [a]
      (Use FlexibleContexts to permit this)
    • When checking the inferred type
        it :: forall a. Num [a] => [a]
```

Let's ask Sanctuary:

```
> S.join
join :: Chain m => m (m a) -> m a

> S.join([[[1], [2]], [[3], [4]]])
[[1], [2], [3], [4]]

> S.join([[1, 2], [3, 4]])
[1, 2, 3, 4]

> S.join([1, 2, 3, 4])
TypeError: Invalid value

join :: Chain m => m (m a) -> m a
                     ^^^^^
                       1

1)  1 :: Number, FiniteNumber, NonZeroFiniteNumber, Integer, ValidNumber

The value at position 1 is not a member of ‘m a’.
```

My understanding is that we do not specify the behaviour of a function when applied to invalid arguments. In this case, `[1, 2, 3, 4, 5]` is an invalid argument.

test/without.js
davidchambers: I'm not sure what you're suggesting here. Would you prefer to duplicate the code for Sanctuary's Maybe type in the Ramda test suite? I worry that the Ramda version wouldn't be kept up to date as additions are made to the Fantasy Land spec.

davidchambers:
> While there are many ways to organize [function wrappers for Fantasy Land methods], the specification will likely not provide much wiggle-room in the core code. It's not clear just how much benefit competing implementations would offer.

I agree wholeheartedly. There's no value in a second semantically equivalent implementation. It could be more efficient, of course, but we're better off working together as a community on one efficient, well-tested, well-documented standard library for Fantasy Land. Many libraries with differing flavours can be built on this foundation, but one can't even define `fantasy-land/equals` for a custom data type without `Z.equals`.

> Presumably this would generate a type error in the updated code.

Not so. We built sanctuary-type-classes with Ramda and other libraries in mind. We made an unopinionated foundation upon which opinionated user-facing libraries such as Sanctuary can be built. `Z` functions do not perform type checking. `Z.lt` returns `false` given values of different types:

```javascript
> Z.lt('5', 7)
false

> Z.lt(7, '5')
false
```

If you'd prefer to preserve the coercive semantics of JavaScript's `<` operator I suggest we make it clear that `R.lt(x, y)` is just sugar for `x < y` by removing references to Ord in our type signatures and function descriptions.

test/without.js
CrossEye: Actually, it turns out that there's already a `Maybe` there.  These tests should have referred to it.  But I don't at all mind switching to using Sanctuary's `Maybe` instead... except that if we try to track latest, we could be burned by tests breaking without an (internal) change.

test/unnest.js
CrossEye: I think I can buy that.  I may give it a little more thought, though.

source/internal/_isPlainObject.js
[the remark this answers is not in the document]
CrossEye: Yes, if we're targeting this at 1.0.  Not that I like it, but I'd rather not add any incompatibilities.

source/filter.js
CrossEye: > I don't think we should use type classes for unprincipled dispatching.

Yes, I think you're right.  We do need to document the names of methods to which we might dispatch.

Although I have not followed FL all that much lately, I am watching that issue; I just haven't had anything to contribute yet.

CrossEye:
> We built sanctuary-type-classes with Ramda and other libraries in mind. We made an unopinionated foundation upon which opinionated user-facing libraries such as Sanctuary can be built. 

I do tend to conflate santuary-type-classes too much with Sanctuary itself, obviously.  But this makes me feel better about it.

It still doesn't address the issue of the advantages of competition.  But I think the team will simply have to make a call on that.

Could you expand on this?:

> Many libraries with differing flavours can be built on this foundation, but one can't even define `fantasy-land/equals` for a custom data type without `Z.equals`.

kedashoe:
> But there is one obvious benefit: the ability to go ones own way.

I agree with this and would like to maintain that at least for now, give FL support a shot within ramda and see where we go from there.

We just got tree shaking support, it would be a shame to immediately add 1,000-ish lines of code plus a sanctuary-type-identifiers dependency to every build! Even with some sort of streamlined s-t-c build there would always be machinery that ramda's approach does not need.

FWIW currently we have I think at least one clear ramda-ism that sanctuary would not go for in the [function instance of functor](https://github.com/ramda/ramda/blob/c58dd1bbc3f525a92991afe813c86c7676fad2d5/source/map.js#L48-L50).

Both sanctuary and ramda "depend" on FL. The two libraries approaching support from their own angles only helps FL. If it is clear down the line we are converging on a common approach, we can merge everything that we share in fantasy-land/what-have-you, both depend on that, and reap the sweet sweet rewards.

I threw together a little WIP today, I'll put the branch up tomorrow for something concrete to look at.

davidchambers:
> I do tend to conflate santuary-type-classes too much with Sanctuary itself, obviously. But this makes me feel better about it.

I'm pleased that you're feeling better. :)

Perhaps the JVM is a good analogy. It's the underlying foundation shared by Java, Scala, and Clojure. Just as Java is not the only language which can run on a Java virtual machine, Sanctuary is not the only library which can be built atop sanctuary-type-classes.

> It still doesn't address the issue of the advantages of competition.

Competition would not be reduced by sharing a foundation. To use the JVM analogy again, Scala and Clojure are very different from Java (and from one another) despite their shared foundation. Similarly, Ramda and Sanctuary would continue to be very different libraries, but the maintenance burden on the Ramda team would be reduced and Ramda users would get new Fantasy Land functions much sooner.

> Could you expand on this?:
>
> > Many libraries with differing flavours can be built on this foundation, but one can't even define `fantasy-land/equals` for a custom data type without `Z.equals`.

Say we wish to implement the Identity type. One of the methods we will define is `Identity#fantasy-land/equals`:

```javascript
//  Identity#fantasy-land/equals :: Identity a ~> Identity a -> Boolean
Identity.prototype['fantasy-land/equals'] = function(other) {
  //  Return `true` iff `this.value` and `other.value` are equal.
};
```

In order to implement this method we need `Z.equals` or an equivalent function. `R.equals` is roughly equivalent and could be used here. My point, though, was that authors of algebraic data types must depend on *some* `equals` function in order to define their own `fantasy-land/equals` methods.

We can't pretend that Fantasy Land is an island: built-in types do not provide `fantasy-land/`-prefixed methods, so they must be provided by a library.

davidchambers:
> FWIW currently we have I think at least one clear ramda-ism that sanctuary would not go for in the [function instance of functor][1].

[`Function#fantasy-land/map`][2] is defined in sanctuary-type-classes:

```javascript
//  Function$prototype$map :: (a -> b) ~> (b -> c) -> (a -> c)
function Function$prototype$map(f) {
  var functor = this;
  return function(x) { return f(functor(x)); };
}
```

The [`S.map`][3] documentation even includes a step-by-step derivation of the Function-specific type from the generalized type.

It's worth keeping in mind, Kevin, that sanctuary-type-classes was created by many of the same people who added Fantasy Land support to Ramda. As is often the case with software, we were able to create something simpler and better organized the second time around.

> Both sanctuary and ramda "depend" on FL. The two libraries approaching support from their own angles only helps FL.

Could you list the benefits you have in mind? The downsides which occur to me are:

  - increased maintenance burden for the Ramda team; and
  - increased likelihood of buggy implementations.

This discussion reminds me of several Fantasy Land discussions in which @SimonRichardson and others urged people not to create yet another implementation of the Maybe type. I'm pushing for a reduction in duplicated effort.


[1]: https://github.com/ramda/ramda/blob/c58dd1bbc3f525a92991afe813c86c7676fad2d5/source/map.js#L48-L50
[2]: https://github.com/sanctuary-js/sanctuary-type-classes/blob/v7.0.0/index.js#L954-L958
[3]: https://sanctuary.js.org/#map

CrossEye:
> It's worth keeping in mind, Kevin, that sanctuary-type-classes was created by many of the same people who added Fantasy Land support to Ramda. As is often the case with software, we were able to create something simpler and better organized the second time around.

But who's to say that a third try would not be still simpler and better organized?

This is the point I was making about competition.  It is important.  I'm wavering here, because I don't really want to go through the effort if it's not likely to lead to something different enough to be interesting.  But I'm almost entirely on the other side of what @SimonRichardson said in those discussions.  [Fluture][fl] would not have been created if everyone remained satisfied with [fantasy-promises][fp] or [data.task][dt].

>> Both sanctuary and ramda "depend" on FL. The two libraries approaching support from their own angles only helps FL.

> Could you list the benefits you have in mind?

I can't speak for Kevin, but in my mind the main point is one of innovation.  If there were twenty competing libraries, then having some of them share a core would make perfect sense.  When there are fewer, it feels more like the (abandoned?) [Underscore/lodash merger][ul]: does it really offer benefits for either team or for the community?

Ramda is made better because there is such a library as Sanctuary.  I was sorry to see [FKit][fk] wither, and am glad to see the emergence of [fp-ts][ts] and [funfix][fu].

And of course @buzzdecafe and I are in a somewhat different position regarding Ramda than most users: as we've often pointed out, this has been a playground for us, a learning experience.  That it works well for others has been very gratifying, but was not the original goal.  While we make some concessions to it now being a full-fledged library, it's difficult to accept the notion that I don't need to tinker with something because it's already been done.

In other words, I'm still very mixed about this.


  [dt]: https://github.com/folktale/data.task
  [fk]: https://github.com/nullobject/fkit
  [fl]: https://github.com/fluture-js/Fluture
  [fp]: https://github.com/fantasyland/fantasy-promises
  [fu]: https://github.com/funfix/funfix
  [ts]: https://github.com/gcanti/fp-ts
  [ul]: https://github.com/jashkenas/underscore/issues/2182

davidchambers:
> But who's to say that a third try would not be still simpler and better organized?

This may well be the case, and I'd love to hear ideas. @Avaq opened sanctuary-js/sanctuary-type-classes#56 to share several good ideas (the pull request stalled as it changed so many things at once, but it would be great to see the improvements discussed in separate pull requests).

> [Fluture][1] would not have been created if everyone remained satisfied with [fantasy-promises][2] or [data.task][3].

I took Simon's point to be about collaboration. There's a spectrum. If we all collaborate on an existing solution to a problem we'll create a well-tested and well-understood library, but it might be a local maximum. If on the other hand we create hundreds of competing solutions we'll have more innovation but also more fragmentation. Fragmentation limits interoperability and the lifespans of blog posts and books focused on particular libraries (rather than spend time creating new JSON parsers, the Haskell community has written some terrific tutorials on aeson).

Everyone lies somewhere on this collaboration ↔︎ innovation spectrum. There's no correct position. You're a bit closer to the innovation end of the spectrum than I am, Scott. I respect that.

As for Folktale and Fluture, now that we have two well-documented and well-tested Future/Task implementations I would strongly discourage others from creating another implementation without a very clear reason for doing so and without first discussing with @robotlolita and @Avaq the possibility of Folktale and/or Fluture adopting the wonderful new approach.


[1]: https://github.com/fluture-js/Fluture
[2]: https://github.com/fantasyland/fantasy-promises
[3]: https://github.com/folktale/data.task

source/internal/_isPlainObject.js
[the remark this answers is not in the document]
davidchambers: :zap:

kedashoe:
> Function#fantasy-land/map is defined in sanctuary-type-classes:

I know, I'm saying the implementation is different.

After sleeping on this I'm more convinced it is the way forward. A project in the FL namespace that does nothing but implement the class instances and is built with tree shaking can grow alongside. Both projects can then depend on it. s-t-c could then add/subtract/replace whatever it wants, eg sanctuary-type-identifiers. Ramda could do the same, eg use our weird `Function#fantasy-land/map` implementation.

davidchambers:
> I'm saying the implementation is different.

Are you referring to the hoops `R.map` jumps through to supports functions of arity greater than one? If so, I'm sorry for misunderstanding your earlier comment. :dove:

At least in the common case the functions behave the same way:

```javascript
> R.map(Math.sqrt, R.add(1))(99)
10

> S.map(Math.sqrt, S.add(1))(99)
10
```

kedashoe:
> Are you referring to the hoops R.map jumps through to supports functions of arity greater than one? If so, I'm sorry for misunderstanding your earlier comment. :dove:

Ya that's all I was saying there :peace_symbol:

buzzdecafe:
> it's difficult to accept the notion that I don't need to tinker with something because it's already been done.

💯

CrossEye:
> Everyone lies somewhere on this collaboration ↔︎ innovation spectrum. There's no correct position. You're a bit closer to the innovation end of the spectrum than I am, Scott. I respect that.

Thanks.  I and hope its clear that I hold a great deal of respect for your POV as well.

I guess I am a little closer to the innovation end, as made clear by my reaction to this:

> As for Folktale and Fluture, now that we have two well-documented and well-tested Future/Task implementations I would strongly discourage others from creating another implementation without a very clear reason for doing so [ ... ]

To me, two is still clearly too few.  Somewhere around three or four is where I'm more likely to be happy; by five or six I might start thinking about which one of them can go.  But two is not enough, and one is scary.  I was there when jQuery came along to disrupt Prototype/Dojo; for all jQuery's warts, it was a very good thing.  And it became a very bad thing when it achieved its eventual dominance.  I'm not happy now with the perceived dichotomy of Angular/React, although this isn't as bad.  I saw the same thing during my Java years as well.

In any case, I have started playing with my own version of this idea, one that is based on some old work I'd done around an alternative to the FL *methods*-based typeclasses.  I don't really expect it to grow into anything useful, but I like playing around with it to see if those ideas might end up providing some compelling alternative.

davidchambers:
Here's an approximation of the Fantasy Land functions Ramda is currently missing:

```console
$ node --print 'const R = require("ramda"), Z = require("sanctuary-type-classes"); R.join("\n", Z.sort(R.difference(R.reject(R.test(/^[A-Z]/), R.keys(Z)), R.keys(R))))'
alt
apFirst
apSecond
bimap
chainRec
contramap
duplicate
elem
extend
extract
id
lift2
lift3
mapLeft
promap
size
zero
```

source/dropLastWhile.js
```diff
@@ -31,5 +31,11 @@ import _xdropLastWhile from './internal/_xdropLastWhile';
  *
  *      R.dropLastWhile(x => x !== 'd' , 'Ramda'); //=> 'Ramd'
  */
-var dropLastWhile = _curry2(_dispatchable([], _xdropLastWhile, _dropLastWhile));
+var dropLastWhile = _curry2(_dispatchable(['dropLastWhile'], _xdropLastWhile, function dropLastWhile(pred, xs) {
+  var idx = xs.length - 1;
+  while (idx >= 0 && pred(xs[idx])) {
+    idx -= 1;
+  }
+  return slice(0, idx + 1, xs);
```
MadDeveloper: What about a more functional way, without mutation, and less code?
```
dropLastWhile(pred, xs) {
  var lastIdx = Z.reduce(idx => idx - Number(idx >= 0 && pred(xs[idx])), xs.length - 1, Array(xs.length))

  return slice(0, lastIdx + 1, xs);
```

MadDeveloper: What about a more functional way? Without mutation?
```js
dropLastWhile2(pred, xs) {
  var length = xs.length
  var lastIdx = Z.reduce(idx => idx - Number(idx >= 0 && pred(xs[idx])), length - 1, Array(length))

  return slice(0, lastIdx + 1, xs);
}
```

davidchambers:
I don't think this pull request would be merged were it to remain open. ;)

PR #2348, opened 2017-10-17
fix _isArguments

davidchambers:
We need to invoke the anonymous function to receive `_isArguments`.


davidchambers:
GitHub is currently having problems. Hopefully Travis will be notified of this pull request eventually.

Issue #2317, opened 2017-10-09
"adjust"-like for properties proposal


haskellcamargo:
This issue proposes the addition of a new function to the default Ramda library.

## Introduction

The function `adjust` is made to apply a function to an specific index of an array, but when you want to work on a specific property of an object, you need to have a manual and verbose combination that could be implemented as default, as long as there are lots of functions to work with properties.

## Example

- Using lists
```js
const incrementAt2 = adjust(2, inc);
console.log(incrementAt2([1, 2, 3, 4])); // [1, 2, 4, 4]
```

- Using objects
```js
const uppercaseName = over(lensProp('name'), toUpper);
// Or
const uppercaseName = evolve({ name: toUpper });
```

## Use case

Having a function able to be an alias to `evolve({ [prop]: fn })` would help in several cases, such as while using `React`'s `setState` with functions instead of values:

```
handleClickPlus() {
  this.setState(adjust('counter', inc));
}
```

## How to

This can be solved by:

- Adding a new function to do this (I have no idea about names)
- Making `adjust` work with objects

CrossEye:
Thank you for the extremely well-written proposal.

While I don't have a substantial objection to this, I also don't see it as anywhere near essential.  Can you convince me?

It seem to me that there is little overhead in writing this:

```js
handleClickPlus() {
  this.setState(evolve({counter: inc}));
}
```

instead of

```js
handleClickPlus() {
  this.setState(ourNewFn('counter', inc));
}
```

Am I missing something essential?

haskellcamargo:
It is purely abstraction and a common necessity on the projects I've been working, where I'm creating a `util` function to reuse that everytime. Some real-world use cases:

- You want to use _functions_ in React `setState`

```js
class Counter extends Component {
  constructor(props) {
    super(props);
    this.state = { counter: 0 };
  }

  handleClickDec() {
    this.setState(in('counter', dec));
  }

  handleClickInc() {
    this.setState(in('counter', inc));
  }

  render() {
    return (
      <div>
        <button onClick={ this.handleClickDec.bind(this) }>-</button>
        <div>{ this.state.counter }</div>
        <button onClick={ this.handleClickInc.bind(this) }>+</button>
      </div>
    );
  }
}
```

- **Consistency** / It would have direct map for functions that work on lists

| List     | Object            |
|----------|-------------------|
| `adjust` | `over . lensProp` | 
| `update` | `assoc`           |
| `remove` | `dissoc`          |
[...]

PR #2353, opened 2017-10-19
Change argument order of adjust

### LINKED ISSUE #1716: update and adjust arguments order
`update` and `adjust` is very similar method. I frequently change between them in code. But it seems inconsistent that `adjust` have index as second argument, but `update` have index as first argument. This often leads to errors.

paldepind:
This PR changes the argument order of `adjust` as discussed in #1716. As to why that is beneficial, I think @iofjuupasli expressed it very well in the original PR:

> `update` and `adjust` is very similar method. I frequently change between them in code. But it seems inconsistent that `adjust` have index as second argument, but `update` have index as first argument. This often leads to errors.

I recently stumbled across this myself when I was implementing `adjust` and `update` for [list](https://github.com/funkia/list).

This is a breaking change. But everybody in the original discussion seemed to be in favor of changing it. Letting `adjust` take the index first not only makes it consistent with `update`, it also makes it more consistent with `assoc`, `dissoc`, `objOf`, `set`, and possibly other functions that all take a "path" before its associated value.

While at it I also removed a curry test from `adjust` in the spirit of @davidchamber's 5e234c9.

This PR closes #1716.

### DIFF
--- a/test/adjust.js
+++ b/test/adjust.js
@@ -3,34 +3,30 @@ var eq = require('./shared/eq');
 
 describe('adjust', function() {
   it('applies the given function to the value at the given index of the supplied array', function() {
-    eq(R.adjust(R.add(1), 2, [0, 1, 2, 3]), [0, 1, 3, 3]);
+    eq(R.adjust(2, R.add(1), [0, 1, 2, 3]), [0, 1, 3, 3]);
   });
 
   it('offsets negative indexes from the end of the array', function() {
-    eq(R.adjust(R.add(1), -3, [0, 1, 2, 3]), [0, 2, 2, 3]);
+    eq(R.adjust(-3, R.add(1), [0, 1, 2, 3]), [0, 2, 2, 3]);
   });
 
   it('returns the original array if the supplied index is out of bounds', function() {
     var list = [0, 1, 2, 3];
-    eq(R.adjust(R.add(1), 4, list), list);
-    eq(R.adjust(R.add(1), -5, list), list);
+    eq(R.adjust(4, R.add(1), list), list);
+    eq(R.adjust(-5, R.add(1), list), list);
   });
 
   it('does not mutate the original array', function() {
     var list = [0, 1, 2, 3];
-    eq(R.adjust(R.add(1), 2, list), [0, 1, 3, 3]);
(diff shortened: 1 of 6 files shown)


test/adjust.js
```diff
@@ -3,34 +3,30 @@ var eq = require('./shared/eq');
...
     var list = [0, 1, 2, 3];
-    eq(R.adjust(R.add(1), 2, list), [0, 1, 3, 3]);
+    eq(R.adjust(2, R.add(1), list), [0, 1, 3, 3]);
     eq(list, [0, 1, 2, 3]);
   });
 
-  it('curries the arguments', function() {
-    eq(R.adjust(R.add(1))(2)([0, 1, 2, 3]), [0, 1, 3, 3]);
-  });
-
```
davidchambers: :thumbsup:

I missed this one.

kedashoe:
I agree update/adjust should be in same order, it seems better to me though to switch the order of `update`. I like `adjust` better with the function first.

test/adjust.js
paldepind: There was a lot of them 😸

paldepind:
@kedashoe I can definitely see the benefit of having the function first. It does seem more useful for partial application. I'm guessing that wanting to apply the same function to multiple indexes happens more often than wanting to apply several functions to the same index.

On the other hand, I think the argument for changing `adjust` is that it's more consistent with other Ramda functions. The general pattern seems to be that a "path" comes before what should be done to the path. For instance `set` takes the thing to set first and then the new value. `assoc` takes a path which is a property and then the value to insert.

It's a bit of a dilemma but as long as `update` and `adjust` becomes the same I think we'd have a huge improvement 😄

kedashoe:
Sums up my thoughts on it pretty well. Can let it marinate for a day or two, if nobody has any strong opinion on it, moving function to 2nd param is ok with me :+1:

CrossEye:
It feels as though we're switching from the correct order to the incorrect one for this function, but I think intra-library consistency should rule here.

:herb:

CrossEye:
I'm going to merge this so #2317 can proceed.

If not for that, I would have just added this to the ongoing discussion about 1.0.

PR #2354, opened 2017-10-19
structured internal fantasy land dispatching

kedashoe:
Not much to it, tried not to do anything fancy.

Implementations are in `source/instances/{type}_{method}.js`, dispatching done in `source/dispatch/{class}_{method}.js`. I believe the latter could be created quite easily with code gen if we wanted to.

See https://github.com/ramda/ramda/pull/2347

### DIFF
--- a/source/concat.js
+++ b/source/concat.js
@@ -1,8 +1,6 @@
 import _curry2 from './internal/_curry2';
-import _isArray from './internal/_isArray';
 import _isFunction from './internal/_isFunction';
-import _isString from './internal/_isString';
-import toString from './toString';
+import semigroup_concat from './dispatch/semigroup_concat';
 
 
 /**
@@ -34,24 +32,12 @@ import toString from './toString';
  *      R.concat([], []); //=> []
  */
 var concat = _curry2(function concat(a, b) {
-  if (_isArray(a)) {
-    if (_isArray(b)) {
-      return a.concat(b);
-    }
-    throw new TypeError(toString(b) + ' is not an array');
-  }
-  if (_isString(a)) {
-    if (_isString(b)) {
-      return a + b;
-    }
-    throw new TypeError(toString(b) + ' is not a string');
-  }
   if (a != null && _isFunction(a['fantasy-land/concat'])) {
     return a['fantasy-land/concat'](b);
   }
   if (a != null && _isFunction(a.concat)) {
     return a.concat(b);
   }
-  throw new TypeError(toString(a) + ' does not have a method named "concat" or "fantasy-land/concat"');
+  return semigroup_concat(a, b);
 });
 export default concat;
--- a/source/dispatch/functor_map.js
+++ b/source/dispatch/functor_map.js
@@ -1,0 +1,13 @@
+import _isArrayLike from '../internal/_isArrayLike';
+import _isFunction from '../internal/_isFunction';
+import _isObject from '../internal/_isObject';
+import array_map from '../instances/array_map';
+import function_map from '../instances/function_map';
+import object_map from '../instances/object_map';
+
+export default function functor_map(fn, functor) {
+  if (_isArrayLike(functor)) return array_map(fn, functor);
+  if (_isFunction(functor)) return function_map(fn, functor);
+  if (_isObject(functor)) return object_map(fn, functor);
+  throw new TypeError('No functor instance for ' + functor);
+};
--- a/source/dispatch/semigroup_concat.js
+++ b/source/dispatch/semigroup_concat.js
@@ -1,0 +1,13 @@
+import _isArray from '../internal/_isArray';
+import _isObject from '../internal/_isObject';
+import _isString from '../internal/_isString';
+import array_concat from '../instances/array_concat';
+import object_concat from '../instances/object_concat';
+import string_concat from '../instances/string_concat';
+
+export default function semigroup_concat(a, b) {
+  if (_isArray(a)) return array_concat(a, b);
+  if (_isObject(a)) return object_concat(a, b);
+  if (_isString(a)) return string_concat(a, b);
+  throw new TypeError('No semigroup instance for ' + a);
+};
--- a/source/instances/array_concat.js
+++ b/source/instances/array_concat.js
@@ -1,0 +1,3 @@
+export default function array_concat(a, b) {
+  return a.concat(b);
+};
--- a/source/instances/function_map.js
+++ b/source/instances/function_map.js
@@ -1,0 +1,7 @@
+import curryN from '../curryN';
+
+export default function function_map(f, g) {
+  return curryN(g.length, function() {
+    return f.call(this, g.apply(this, arguments));
+  });
+};
--- a/source/instances/object_concat.js
+++ b/source/instances/object_concat.js
@@ -1,0 +1,10 @@
+export default function object_concat(a, b) {
+  var c = {};
+  for (var k in a) {
+    c[k] = a[k];
+  }
+  for (var k in b) {
+    c[k] = b[k];
+  }
+  return c;
+};
--- a/source/instances/object_map.js
+++ b/source/instances/object_map.js
@@ -1,0 +1,9 @@
+import _reduce from '../internal/_reduce';
+import keys from '../keys';
+
+export default function object_map(f, obj) {
+  return _reduce(function(acc, key) {
+    acc[key] = f(obj[key]);
+    return acc;
+  }, {}, keys(obj));
+};
--- a/source/instances/array_map.js
+++ b/source/instances/array_map.js
@@ -1,0 +1,7 @@
+export default function array_map(f, a) {
+  var b = [];
+  for (var i = 0; i < a.length; ++i) {
+    b.push(f(a[i]));
+  }
+  return b;
+};
--- a/source/instances/string_concat.js
+++ b/source/instances/string_concat.js
@@ -1,0 +1,3 @@
+export default function string_concat(a, b) {
+  return a + b;
+};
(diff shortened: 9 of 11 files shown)


source/instances/function_map.js
```diff
@@ -0,0 +1,7 @@
+import curryN from '../curryN';
+
+export default function function_map(f, g) {
+  return curryN(g.length, function() {
```
kedashoe: Note [this test](https://github.com/ramda/ramda/blob/1a61378c1811479d9820f13581345a9baacf4129/test/complement.js#L18) fails if we implement this as `f(g(x))`.

source/instances/object_map.js
```diff
@@ -0,0 +1,9 @@
+import _reduce from '../internal/_reduce';
+import keys from '../keys';
+
+export default function object_map(f, obj) {
+  return _reduce(function(acc, key) {
```
kedashoe: [This test](https://github.com/ramda/ramda/blob/1a61378c1811479d9820f13581345a9baacf4129/test/whereEq.js#L63) fails with simple `for (k in obj)` loop, due to the keys iterated over vs what we get with `keys`.

source/dispatch/functor_map.js
```diff
@@ -0,0 +1,13 @@
+import _isArrayLike from '../internal/_isArrayLike';
+import _isFunction from '../internal/_isFunction';
+import _isObject from '../internal/_isObject';
+import array_map from '../instances/array_map';
+import function_map from '../instances/function_map';
+import object_map from '../instances/object_map';
+
+export default function functor_map(fn, functor) {
+  if (_isArrayLike(functor)) return array_map(fn, functor);
```
kedashoe: I have just `_isArray` below for `concat`. Both should be `isArray` probably, but we used to just fall through to array implementation for `map`, so this is more like the old behaviour?

source/dispatch/semigroup_concat.js
```diff
@@ -0,0 +1,13 @@
...
+import array_concat from '../instances/array_concat';
+import object_concat from '../instances/object_concat';
+import string_concat from '../instances/string_concat';
+
+export default function semigroup_concat(a, b) {
+  if (_isArray(a)) return array_concat(a, b);
+  if (_isObject(a)) return object_concat(a, b);
+  if (_isString(a)) return string_concat(a, b);
+  throw new TypeError('No semigroup instance for ' + a);
+};
```
davidchambers: This seems messy to me compared with the way mappings are defined in sanctuary-type-classes:

```javascript
var implementations = {
  Null: {
    prototype: {
      toString:                   Null$prototype$toString,
      'fantasy-land/equals':      Null$prototype$equals,
      'fantasy-land/lte':         Null$prototype$lte
    }
  },
  ...
};
```

Did you consider using a data structure rather than lots of branching logic?

source/instances/array_concat.js
```diff
@@ -0,0 +1,3 @@
+export default function array_concat(a, b) {
+  return a.concat(b);
```
davidchambers: The argument should be `[b]` rather than `b`.

source/instances/object_concat.js
```diff
@@ -0,0 +1,10 @@
+export default function object_concat(a, b) {
+  var c = {};
+  for (var k in a) {
+    c[k] = a[k];
+  }
+  for (var k in b) {
+    c[k] = b[k];
+  }
+  return c;
+};
```
davidchambers: This is semantically different from the Sanctuary implementation. See sanctuary-js/sanctuary-type-classes#70.

source/instances/array_concat.js
kedashoe: > The argument should be [b] rather than b.

Not sure I understand? `b` is an array, right?
```js
let a = [1, 2];
let b = [3, 4];
a.concat([b]); //=> [1, 2, [3, 4]]
```

davidchambers: You're quite right. My mistake. :blush:

source/instances/object_concat.js
kedashoe: Probably should use internal `keys`

source/dispatch/semigroup_concat.js
kedashoe: What do you mean by messy? Looks pretty clean to me? Obviously it is a bit boilerplate-y. I did think about a data structure but I think this is simple and naturally supports tree shaking with no overhead.

It would be neat if fantasy-land had a data file of some sort we could add as a dev dependency and use to auto-generate these files.

source/instances/array_concat.js
kedashoe: Note [we never get here anyway](https://github.com/kedashoe/ramda/blob/8b38cc5ab35ca3eb96e3397845e4a73846da5019/source/concat.js#L38-L40) haha :confounded:

source/dispatch/semigroup_concat.js
davidchambers: > I did think about a data structure but I think this is simple and naturally supports tree shaking with no overhead.

Ah, I see. Your approach seems good, in that case. :thumbsup:

source/instances/array_concat.js
davidchambers: It seems we need to make an exception to the exception. :stuck_out_tongue_winking_eye:

source/dispatch/functor_map.js
CrossEye: Does that apply to `Array` too?  How about `Function` or `Object`?  The point here is to decide where we want to apply it *beyond* any dispatching.

source/dispatch/semigroup_concat.js
CrossEye: I have a similar start at this, but also using a data structure.  This approach would definitely seem more modular.

source/concat.js
```diff
@@ -34,24 +32,12 @@ import toString from './toString';
...
-    throw new TypeError(toString(b) + ' is not a string');
-  }
   if (a != null && _isFunction(a['fantasy-land/concat'])) {
     return a['fantasy-land/concat'](b);
   }
   if (a != null && _isFunction(a.concat)) {
     return a.concat(b);
   }
-  throw new TypeError(toString(a) + ' does not have a method named "concat" or "fantasy-land/concat"');
+  return semigroup_concat(a, b);
```
CrossEye: Two things: is `semigroup_concat` used elsewhere?  If not, shouldn't we internalize it here?

Second, and this will probably only work well if the answer to that it yes, we still have some expectations that while all sorts of things can be `concat`ed, don't we want to order our tests from most likely to least, in which case, I would prefer to check Arrays, Strings, and Objects before the fantasy-land check.

source/dispatch/functor_map.js
davidchambers: Taking an Arguments object and returning an array is not compliant with the spec:

> `map` must return a value of the same Functor

If we want to be able to map over Arguments objects we'd need to support this:

```haskell
map :: (a -> b) -> Arguments a -> Arguments b
```

This seems silly to me, but I wouldn't oppose it.

I think we should provide the same implementations for built-in types `Z.map` does.

CrossEye: This comes back to our fundamental disagreement on Ramda's relationship with FantasyLand.

I would like Ramda to be able to ergonomically interoperate with many tools: things like Immutable.js, streaming libraries, reactive toolkits, and types derived from specifications such as FantasyLand.

I don't think of any of those specifying how Ramda should work.

So I don't think of FantasyLand as a specification for Ramda, only as a way to specify certain types.  I would like us to be able to interoperate with those types, but that does not imply that our functions can't do things incompatible with FL laws for other types.

I don't have any strong desire to work with `Arguments` or `NodeList` objects in `map`, but I don't think of FantasyLand as specifying that we cannot do this if we choose:

```js
const squareArgs = function() {return map(n => n * n, arguments)}
squareArgs(1, 2, 3, 4) //=> [1, 4, 9, 16]
```

even though that invocation of `map` does not conform to the FL spec.

davidchambers: I think I now see your point, Scott. Are you suggesting that we behave lawfully for values which do satisfy a particular Fantasy Land type class, but that we're free to define ad hoc behaviour—rather than throwing—for values which do not?

CrossEye: I think those types that can satisfy normal algebraic laws -- and whose functions we supply -- should by all means supply those laws.  But I think that we should not ever try to enforce those laws on our own, and I believe that there are common usages around some of these function names, regardless of whether the types obey the laws as specified by FantasyLand.  I would like to be able to interoperate with this types nonetheless.

For the same reason I would like to be able to do `map(square, Just(5))`, even though `Maybe` supplies a method and I can so `Just(5).map(square)` instead, I would also like to be able do `map(square, myStream)` even when `myStream.map(square)` is available.  And this is true, even if someone hasn't proven the composability of `map` on these streams.  It's the same abstraction and should be offered in the same way.

So yes, I would say behaving lawfully where we can is important, but R.map should not be the function that tries to serve all lawful functors and keep out anything that isn't.

davidchambers: > It's the same abstraction and should be offered in the same way.

Ah, I see the distinction you're making. You're happy with `map(square, myStream)` but you'd consider `empty($divs)` an abuse? (The jQuery expression `$divs.empty()` has nothing to do with the Monoid type class.)

CrossEye: Yes, and that's a _caveat emptor_, of course.  If you use this with types that don't conform, bad things could happen.  But from Ramda's perspective that's no different from

```js
 map(square, {
  'fantasy-land/map': launchTheMissiles
});
```

We don't get to test that the functions specified by the types conform to any particular rules, so let's not bother restricting the dispatch to those claiming to.

PR #2357, opened 2017-10-23
add @see to into

adispring:


CrossEye:
I definitely understand why `@see transduce`, but why `reduce` as well?

adispring:
I copied from `transduce`. Have removed `reduce`.

CrossEye:
Thanks!

PR #2358, opened 2017-10-23
link reduceBy to groupBy

adispring:


CrossEye:
Why do you want this link?  Just because `groupBy` is implemented with `reduceBy`?

adispring:
I've seen the `reduceBy` example and its description:

```js
 * This function is basically a more general [`groupBy`](#groupBy) function.
```

CrossEye:
Good enough for me.

PR #2361, opened 2017-10-23
optimize and.js && or.js description

adispring:


adispring:
So many times of discussions. 🤔

PR #2362, opened 2017-10-23
optimize both & either description

adispring:


buzzdecafe:
:cow2:

CrossEye:
@buzzdecafe:

Are you sure we should merge this?  I just shut down #2361, which was very similar, because it was reverting a change that had been a hard-fought battle.

buzzdecafe:
ah apologies, seemed straightforward. feel free to revert

CrossEye:
Reverted for the reasons described here: https://github.com/ramda/ramda/pull/2361#issuecomment-338697472.  @adispring: If you want to argue for this, be my guest.  But I don't think we should change it back without such a discussion.

adispring:
As my understanding, `both` and `either` is just two special case of `allPass` and `anyPass`, can we adding links between them?

CrossEye:
Yes, that's not how they arose historically, but it represents the API as it stands.  Links would be good, and in both directions.  That is, 

```js
// both
// * @see R.either, R.allPass

// allPass
// * @see R.anyPass, R.both

// etc.
```

PR #2365, opened 2017-10-23
add mergeLeft and mergeRight

### LINKED ISSUE #2225: Add mergeLeft and mergeRight
Pretty often I find myself using `R.merge(R.__, object)`, it would be nice to have R.mergeLeft and R.mergeRight by analogy with R.mergeDeepLeft and R.mergeDeepRight.

hrldcpr:
This adds `mergeLeft` and `mergeRight` methods, consistent with the existing `mergeDeepLeft` and `mergeDeepRight` methods. Closes #2225.

If `merge` is going to remain as an alias of `mergeRight` then we should probably also introduce a `mergeDeep` alias of `mergeDeepRight` to be consistent. Alternatively we should deprecate `merge`.

### DIFF
--- a/source/mergeLeft.js
+++ b/source/mergeLeft.js
@@ -5,28 +5,26 @@ import _curry2 from './internal/_curry2';
 /**
  * Create a new object with the own properties of the first object merged with
  * the own properties of the second object. If a key exists in both objects,
- * the value from the second object will be used.
+ * the value from the first object will be used.
  *
  * @func
  * @memberOf R
- * @since v0.1.0
  * @category Object
  * @sig {k: v} -> {k: v} -> {k: v}
  * @param {Object} l
  * @param {Object} r
  * @return {Object}
- * @see R.mergeRight, R.mergeDeepRight, R.mergeWith, R.mergeWithKey
- * @deprecated
+ * @see R.mergeRight, R.mergeDeepLeft, R.mergeWith, R.mergeWithKey
  * @example
  *
- *      R.merge({ 'name': 'fred', 'age': 10 }, { 'age': 40 });
+ *      R.mergeLeft({ 'age': 40 }, { 'name': 'fred', 'age': 10 });
  *      //=> { 'name': 'fred', 'age': 40 }
  *
- *      var withDefaults = R.merge({x: 0, y: 0});
- *      withDefaults({y: 2}); //=> {x: 0, y: 2}
- * @symb R.merge(a, b) = {...a, ...b}
+ *      var resetToDefault = R.mergeLeft({x: 0});
+ *      resetToDefault({x: 5, y: 2}); //=> {x: 0, y: 2}
+ * @symb R.mergeLeft(a, b) = {...b, ...a}
  */
-var merge = _curry2(function merge(l, r) {
-  return _objectAssign({}, l, r);
+var mergeLeft = _curry2(function mergeLeft(l, r) {
+  return _objectAssign({}, r, l);
 });
-export default merge;
+export default mergeLeft;
--- a/test/mergeLeft.js
+++ b/test/mergeLeft.js
@@ -1,0 +1,43 @@
+var assert = require('assert');
+
+var R = require('..');
+var eq = require('./shared/eq');
+
+
+describe('mergeLeft', function() {
+  it('takes two objects, merges their own properties and returns a new object', function() {
+    var a = {w: 1, x: 2};
+    var b = {y: 3, z: 4};
+    eq(R.mergeLeft(a, b), {w: 1, x: 2, y: 3, z: 4});
+  });
+
+  it('overrides properties in the second object with properties in the first object', function() {
+    var a = {w: 1, x: 2};
+    var b = {w: 100, y: 3, z: 4};
+    eq(R.mergeLeft(a, b), {w: 1, x: 2, y: 3, z: 4});
+  });
+
+  it('is not destructive', function() {
+    var a = {w: 1, x: 2};
+    var res = R.mergeLeft(a, {x: 3, y: 4});
+    assert.notStrictEqual(a, res);
+    eq(res, {w: 1, x: 2, y: 4});
+  });
+
+  it('reports only own properties', function() {
+    var a = {w: 1, x: 2};
+    function Cla() {}
+    Cla.prototype.x = 5;
+    eq(R.mergeLeft(new Cla(), a), {w: 1, x: 2});
+    eq(R.mergeLeft(a, new Cla()), {w: 1, x: 2});
+  });
+
+  it('is shallow', function() {
+    var a = { x: { u: 1, v: 2 }, y: 0 };
+    var b = { x: { u: 3, w: 4 }, z: 0 };
+    var res = R.mergeLeft(a, b);
+    assert.strictEqual(a.x, res.x);
+    eq(res, { x: { u: 1, v: 2 }, y: 0, z: 0 });
+  });
+
+});
--- a/test/mergeRight.js
+++ b/test/mergeRight.js
@@ -1,0 +1,43 @@
+var assert = require('assert');
+
+var R = require('..');
+var eq = require('./shared/eq');
+
+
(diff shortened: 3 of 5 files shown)


source/merge.js
```diff
@@ -15,17 +14,15 @@ import _curry2 from './internal/_curry2';
...
- *      resetToDefault({x: 5, y: 2}); //=> {x: 0, y: 2}
+ *      var withDefaults = R.merge({x: 0, y: 0});
+ *      withDefaults({y: 2}); //=> {x: 0, y: 2}
  * @symb R.merge({ x: 1, y: 2 }, { y: 5, z: 3 }) = { x: 1, y: 5, z: 3 }
  */
-var merge = _curry2(function merge(l, r) {
-  return _objectAssign({}, l, r);
-});
+var merge = mergeRight;
 export default merge;
```
CrossEye: How about if we leave this one alone except to add "deprecated"/"see" tags to it?

It feels odd to be changing the implementation just in order to deprecate it.  I don't think the small amount of repeated code is an issue.

test/mergeLeft.js
```diff
@@ -0,0 +1,41 @@
...
+    Cla.prototype.x = 5;
+    eq(R.mergeLeft(new Cla(), a), {w: 1, x: 2});
+    eq(R.mergeLeft(a, new Cla()), {w: 1, x: 2});
+  });
+
+  it('is shallow', function() {
+    var a = { a: { b: 1, c: 2 }, y: 0 };
+    var b = { a: { b: 3, d: 4 }, z: 0 };
+    eq(R.mergeLeft(a, b), { a: { b: 1, c: 2 }, y: 0, z: 0 });
+  });
```
CrossEye: This doesn't show that it's shallow.  You need to do a strict equality check on `b.a` and the 'a' property of the merged object.

test/mergeRight.js
```diff
@@ -0,0 +1,41 @@
...
+    Cla.prototype.x = 5;
+    eq(R.mergeRight(new Cla(), a), {w: 1, x: 2});
+    eq(R.mergeRight(a, new Cla()), {w: 1, x: 2});
+  });
+
+  it('is shallow', function() {
+    var a = { a: { b: 1, c: 2 }, y: 0 };
+    var b = { a: { b: 3, d: 4 }, z: 0 };
+    eq(R.mergeRight(a, b), { a: { b: 3, d: 4 }, y: 0, z: 0 });
+  });
```
CrossEye: Ditto

CrossEye:
This looks mostly good.  I left a few comments in the review.  Do others agree with me that we can just leave `merge` alone except for documenting that it's deprecated?

source/merge.js
hrldcpr: Yeah I'm fine with that, updated 👍

test/mergeLeft.js
hrldcpr: Good point, added an assertion 👍

hrldcpr:
@CrossEye thanks for the review! Comments addressed 👍

Bradcomp:
What about `mergeWith` and `mergeWithKey`? Are we going to deprecate those and add two new versions as well? What to do with `mergeAll`?

I definitely agree that if we add these functions we should deprecate `merge`, but I would argue that the symmetric merging functions aren't really necessary. Between a simple lambda, the placeholder, and `flip` there are 3 ways to accomplish `mergeRight`, and they will work for all the other `merge*` functions too.  

I probably should have brought this up during the `mergeDeep` discussions, this seems like inflating the size of the API without a clear benefit.

hrldcpr:
@Bradcomp at least for `mergeWith` and `mergeWithKey` it doesn't matter, since they are not biased left or right. `mergeAll` is a good point though.

I've frequently wanted `mergeLeft` behavior when currying, though you're right that `flip(merge)` isn't too terrible.

Anyway I think we can all agree it's weird to have `mergeDeepLeft` but no `mergeLeft`, so yeah maybe getting rid of that is the easier / slimmer solution?

source/mergeLeft.js
```diff
@@ -0,0 +1,30 @@
...
+ * @return {Object}
+ * @see R.mergeRight, R.mergeDeepLeft, R.mergeWith, R.mergeWithKey
+ * @example
+ *
+ *      R.mergeLeft({ 'age': 40 }, { 'name': 'fred', 'age': 10 });
+ *      //=> { 'name': 'fred', 'age': 40 }
+ *
+ *      var resetToDefault = R.mergeLeft({x: 0});
+ *      resetToDefault({x: 5, y: 2}); //=> {x: 0, y: 2}
+ * @symb R.mergeLeft({ x: 1, y: 2 }, { y: 5, z: 3 }) = { x: 1, y: 2, z: 3 }
```
kedashoe: hmm, I see some of the other `merge` functions have `@symb` definitions like this.. this doesn't really feel like it is in the spirit of `@symb` to me, can we just leave these off? Anyone else thoughts?

hrldcpr: Yeah I just copied that stuff from `merge`. In fact I don't even know what `@symb` does, it doesn't seem to show up in the docs… what is it for?

Meanwhile I'll go ahead and remove it 👍

hrldcpr: Looking at other `@symb` examples, it looks like it's more about giving generic (symbolic?) definitions, so maybe showing the ES6 equivalent or something would be useful? I.e.:

```js
@symb R.mergeLeft(a, b) = { ...b, ...a }
```

…or maybe that's just confusing. Anyway I've completely removed the `@symb` stuff for now.

kedashoe: :+1:

hrldcpr:
I wanted to check in on this pull request; any thoughts or things I should change?

CrossEye:
@Bradcomp: are you ok with this, or should we revisit?

Bradcomp:
I'm OK with this.

nickmccurdy:
Is merge still going to be deprecated? Personally I'd rather deprecate `mergeLeft` and `mergeDeepLeft` (renaming it to `mergeDeep`) to be more consistent with methods that start from the left by default like reduce.

hrldcpr:
@nickmccurdy yeah I agree `deepMergeLeft` was an unusual addition in the first place, given that there are no other "left" methods but instead: `mapAccum`, `mapAccumRight`, `partial`, `partialRight`, `reduce`, `reduceRight`

hrldcpr:
@nickmccurdy yeah I agree `deepMergeLeft` (and now `mergeLeft`) seem a bit out of place, given that there are no other "left" methods but instead: `mapAccum`, `mapAccumRight`, `partial`, `partialRight`, `reduce`, `reduceRight`

**But** there's the additional inconsistency that `merge` is equivalent to `mergeRight`, *not* `mergeLeft` (to be consistent with `deepMergeRight` and `deepMergeLeft`) so it's not quite so easy to remedy the situation, and it's probably too late to reverse the meaning of `deepMergeRight`

lax4mike:
I feel like these functions are not necessary and add bloat to library and the docs.  

I reached for `mergeDeep` the other day and found `mergeDeepRight` and `mergeDeepLeft`.  I was confused about what they did, _"it can't just be the ordering, can it?"_.  I feel like it would be equally as confusing encountering these functions reading someone else's code.

I'm going to reiterate what @Bradcomp said above:

> Between a simple lambda, the placeholder, and flip there are 3 ways to accomplish `merge[Left]`, and they will work for all the other merge* functions too.

I prefer the shorter, clearer function names (`merge`, `mergeDeep`) and I can handle `mergeLeft` with any of the 3 ways above, which are simple and ubiquitous.

It seems the reason folks want `mergeLeft` is so it works better when partially applied. But didn't we decided against that for `subtract`, `divide`, `lt`, etc?

CrossEye:
@lax4mike: You may well be right.  We seem to be adding functions that don't have a really strong purpose.  I'm going to contemplate this for a day or two.  I'd love to hear what the @ramda/core team thinks about it.

PR #2368, opened 2017-10-25
Optimze both either description

adispring:


kedashoe (dismissed):

customcommander:
@adispring You have a merge conflict ;)

CrossEye:
Definitely a good idea.  I'm still triaging issues, but if no one has gotten around to it before I get a chance, I'll clean up the conflicts.

adispring:
The conflict has been fixed.

customcommander:
@adispring would you mind fixing the linter errors as well?

adispring:
Ok, I'll see what happened.

adispring:
There is a eslint error in test/test.js, so I have to add a babel-eslint parser to fix this problem: https://stackoverflow.com/questions/51936079/eslint-how-to-ignore-parsing-error-invalid-regular-expression-flag

.eslintrc
```diff
@@ -7,6 +7,7 @@
     "mocha": true,
     "es6": true
   },
+  "parser": "babel-eslint",
```
customcommander, line 10: Are you sure this is necessary? I didn't see any linter errors in your last push; only a failing test. So I am bit confused here.

adispring: As this [stackoverflow](https://stackoverflow.com/questions/51936079/eslint-how-to-ignore-parsing-error-invalid-regular-expression-flag) said, regex symbol `s` is a ecmascript 2018 feature, eslint's default parser can not parse `/***/s` to a valid AST, and can not lint it.

There's 2 ways to fix this:

1. Add `"ecmaVersion": 6` to .eslintrc, but this will cause other eslint problems related to ecam version.
2. add babel-eslint parser to parse `/***/s` .

adispring: https://github.com/ramda/ramda/pull/3158

customcommander: Sorry I must be missing something here.

I know that #3158 introduced the `s` flag but last time I checked your pull request failed for exactly the same reason master is failing right now:

This is the only test that is failing on master:

Screenshot 2022-01-23 at 09 01 02

I'm failing to see how this relates to the regex flag.

That same test used to fail on your pull request however you have now chosen to disable that test.

test/swap.js
```diff
@@ -45,7 +45,7 @@ describe('swap', function() {
     eq(R.swap(0, 2, 'foo'), 'oof');
   });
 
-  it('swaps object indexes in map from one index to another', function() {
-    eq(R.swap(o, o2, map), new Map([[o, 2], [o2, 1]]));
-  });
+  // it('swaps object indexes in map from one index to another', function() {
+  // eq(R.swap(o, o2, map), new Map([[o, 2], [o2, 1]]));
+  // });
```
customcommander, line 50: This is the only unit test that used to fail before you disabled it:

Screenshot 2022-01-23 at 09 01 02

customcommander (changes requested):

.eslintrc
customcommander: What I am trying to say is that your pr didn't fail because of that regex flag you mention but because of that unit test failure.

package.json
```diff
@@ -86,14 +86,14 @@
...
   },
-  "dependencies": {},
   "devDependencies": {
     "@babel/cli": "^7.4.4",
     "@babel/core": "^7.4.5",
     "@babel/preset-env": "^7.7.4",
     "@babel/register": "^7.4.4",
     "@babel/types": "^7.4.4",
     "@rollup/plugin-babel": "^5.0.4",
+    "babel-eslint": "^10.1.0",
```
customcommander, line 96: I played with the `.eslintrc` config and found that we don't need a plugin to fix the `s` regex flag issue. This seems to work fine too:

```none
  "parserOptions": {
    "sourceType": "module",
    "ecmaVersion": 2018 <-- HERE
  },
```

.eslintrc
customcommander: Ok ignore this last comment. I just figured out that adding that `babel-eslint` was intended to fix a lint issue. But it does seem that we can fix that lint error without depending on another third-party plugin, we can just set the parser options in the eslint config file. See my other comment.

test/swap.js
customcommander: I think we should try to fix the test or the implementation (or both).

.eslintrc
customcommander: So I spent some time trying to figure out the issue with that test and I _think_ that the test simply can't work at all considering the implementation of `swap`. I opened a PR to fix the linter and tests failures so that this PR can focus on just the documentation. See #3224

customcommander:
@adispring: With #3224 merged only this commit of yours should be necessary now: https://github.com/ramda/ramda/pull/2368/commits/99cf77c21035fe74c8b2ccb44e64096b31318af9

I can help you rebasing this branch (and others) if you don't have the time right now.

adispring:
Just reset to the first PR commit. @customcommander

customcommander (approved):

Issue #2244, opened 2017-07-29
Misleading documentation of "either" and "both" methods. They can return not only functions


barbiturat:
Documentation said about the "either" method: 
"**Returns function** a function that applies its arguments to `f` and `g` and `||`s their outputs together."
But the method returns not only a function, but any applicative functor of corresponded type. Example:
`either(Maybe.Just(false), Maybe.Just(55)); // => Maybe.Just(55)`
`either([false, false, 'a'], [11]) // => [11, 11, "a"]`
This description may mislead.
Same about the "both" method

buzzdecafe:
agreed, would you be willing to submit a PR to update the docs?

karatsuba:
hey, here is a fix for that: https://github.com/ramda/ramda/pull/2373

karatsuba:
Hello @buzzdecafe!
I guess this issue can be closed now.

PR #2373, opened 2017-10-31
Misleading documentation of "either" and "both" methods fix

karatsuba:
Hello! Thanks a lot for such a cool library! I really like it!

Here is a fix for https://github.com/ramda/ramda/issues/2244

### DIFF
--- a/source/both.js
+++ b/source/both.js
@@ -31,6 +31,9 @@ import lift from './lift';
  *      var f = R.both(gt10, lt20);
  *      f(15); //=> true
  *      f(30); //=> false
+ *
+ *      R.both(Maybe.Just(false), Maybe.Just(55)); // => Maybe.Just(false)
+ *      R.both([false, false, 'a'], [11]); //=> [false, false, 11]
  */
 var both = _curry2(function both(f, g) {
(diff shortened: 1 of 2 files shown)


source/both.js
```diff
@@ -31,6 +33,9 @@ import lift from './lift';
  *      var f = R.both(gt10, lt20);
  *      f(15); //=> true
  *      f(30); //=> false
+ *
+ *      both(Maybe.Just(false), Maybe.Just(55)); // => Maybe.Just(false)
```
kedashoe: `both` -> `R.both` (it is standard for the doc examples to use the `R.` prefix)

source/both.js
```diff
@@ -12,6 +12,8 @@ import lift from './lift';
  * meaning that the second function will not be invoked if the first returns a
  * false-y value.
  *
+ * Also the method returns not only a function, but any applicative functor of corresponded type.
```
kedashoe: I don't think we need this, there is already a note about these functions working with any applicative

kedashoe:
ping @karatsuba

karatsuba:
hi @kedashoe! I pushed my changes 😄

PR #2379, opened 2017-11-09
Update tryCatch example to show correct behaviour

lolcookie:
Before the docs showed

```js
R.tryCatch(R.prop('x'), R.F)(null);      //=> false
```

which actually evaluates as
```js
R.tryCatch(R.prop('x'), R.F)(null);      //=> undefined
```

Updated to use JSON.parse as it actually throws exceptions
```js
R.tryCatch(JSON.parse, R.F)({});      //=> false
```


buzzdecafe:
looks good, thanks
:cow2:

PR #2385, opened 2017-11-14
Preserve property descriptors in assoc and dissoc

### LINKED ISSUE #2383: Computed property getters not preserved
When operating on objects, it appears that property getters are not preserved. Instead, they're resolved and replaced with their value. Here's a minimal example:

```js
let rectangle = {
  width: 2,
  height: 3,
  get area () {
    return this.width * this.height
  }
}

console.log(rectangle) // => { width: 2, height: 3, area: [Getter] }

rectangle = R.over(R.lensProp('width'), R.multiply(2), rectangle)

console.log(rectangle) // => { width: 4, height: 3, area: 6 }
```

Is this something worth fixing? Or is the solution simply to not use property getters and instead define a function for doing so (like `area :: Rectangle -> Number`, in this case)? Thanks!

jtormey:
Fixes #2383 

This change adds an internal function `_cloneWithDescriptors` which creates a shallow object clone, preserving any existing property descriptors. It is used in `assoc` and `dissoc` so that property getters and setters are copied to the returned object.

Something that @CrossEye mentioned is that there may be a performance concern. I tested this in a JSPerf [here](https://jsperf.com/ramda-assoc-perf), and found that this performs ~10x worse than before (~1M ops/sec as opposed to ~10M), which could be a reason to reject this change.

Also... I changed the test script to run tests in `test/internal` 😄

### DIFF
--- a/source/internal/_cloneWithDescriptors.js
+++ b/source/internal/_cloneWithDescriptors.js
@@ -1,0 +1,22 @@
+
+
+/**
+ * Performs a shallow clone of an object while preserving property descriptors.
+ *
+ * @private
+ * @param {Object} obj The object to clone
+ * @return {Object} The cloned object
+ */
+export default function _cloneWithDescriptors(obj) {
+  var result = {};
+  var desc;
+  for (var p in obj) {
+    desc = Object.getOwnPropertyDescriptor(obj, p);
+    if (desc === undefined) {
+      result[p] = obj[p];
+    } else {
+      Object.defineProperty(result, p, desc);
+    }
+  }
+  return result;
+}
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "postcoverage": "npm run posttest",
     "lint": "eslint scripts/bookmarklet scripts/*.js source/*.js source/internal/*.js test/*.js test/**/*.js lib/sauce/*.js lib/bench/*.js",
     "pretest": "npm run build:cjs",
-    "test": "mocha --reporter spec",
+    "test": "mocha --reporter spec -- test test/internal",
     "posttest": "git checkout -- dist",
     "prebrowser_test": "npm run pretest",
     "browser_test": "testem ci",
--- a/source/assoc.js
+++ b/source/assoc.js
@@ -1,4 +1,5 @@
 import _curry3 from './internal/_curry3';
+import _cloneWithDescriptors from './internal/_cloneWithDescriptors';
 
 
 /**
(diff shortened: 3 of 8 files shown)


source/internal/_cloneWithDescriptors.js
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Performs a shallow clone of an object while preserving property descriptors.
+ *
+ * @private
+ * @param {Object} obj The object to clone
+ * @return {Object} The cloned object
+ */
+export default function _cloneWithDescriptors(obj) {
+  var result = {};
```
CrossEye: If we're going to do this, would it make sense to start with the source object's prototype and get rid of our annoying "flattens prototype properties onto the new object" while we're at it?

CrossEye:
Obviously the performance changes are significant.  I see a 9X difference in Chrome and Brave and a 6X difference in Firefox with that test, all on Ubuntu 17.04

JSPerf was fighting me, and I never got to try my other version, which was to inline the _cloneWithDescriptors.  I doubt the function call contributed that much, but I wanted to see.  Perhaps I'll try later.

I think this would be great to have, but those perf changes scare me.  I'm hoping others will chime in here with their thoughts.

CrossEye:
Anyone have thoughts on this?  @buzzdecafe, @kedashoe, @davidchambers, @paledpind, @TheLudd, @scott-christopher, @asaf-romano, @raine, @Bradcomp?

davidchambers:
As far as I'm concerned these are the signatures:

```haskell
assoc :: String -> a -> StrMap a -> StrMap a
dissoc :: String -> StrMap a -> StrMap a
```

String maps do not have getters and setters, so I think we should ignore these.

Historically Ramda has accommodated much more of JavaScript's complexity than I would like, so it may be that preserving getters and setters is more consistent with decisions we've made in the past.

CrossEye:
> String maps do not have getters and setters

By some definition/convention?  I think I'm going to come down against this based on the performance and on our disinterest in working with OO.  But I don't see that how the String property is created changes whether this is a StrMap.

davidchambers:
String maps are just data. Getters and setters are for triggering actions (possibly side effects).

CrossEye:
But how does the implementation of properties (even if they also add side effects) change the API of a `StrMap`?  For instance, if a `StrMap` has a commonly accessed, but time-consuming to calculate, derived property it's easy enough to imagine using setters and getters to control the properties upon which it depends.  That doesn't change the use of that object.

Maybe I should just quit here.  I agree with your conclusion.  Perhaps it doesn't matter that I disagree with your reasons.  :smile:

CrossEye:
Closing as stale.  But leaving a label to remind us to come back to this later.  If we can solve the performance problems, something here or in #2383 might be worth pursuing.

Issue #2386, opened 2017-11-15
mapAccumRight argument order?


tacomanator:
Is the argument order for `mapAccumRight` correct? Docs say 

> iterator function receives two arguments, value and acc, and should return a tuple [value, acc]

But using as documented:

```javascript
var iterator = (value, acc) => [value, acc];
R.mapAccumRight(iterator, "acc", ["a", "b"]); //=> [["b", "acc"], "a"]
```

Reversing order of tuple returned by iterator works as expected:
```javascript
var iterator = (value, acc) => [acc, value];
R.mapAccumRight(iterator, "acc", ["a", "b"]); //=> [["a", "b"], "acc"]

**Comment by kedashoe** [comment]
Can't say I understand this function well, but it seems the parameters are being passed along as advertised: https://goo.gl/XgrztS

```js
var iterator = (value, acc) => (console.log(`val=${value}, acc=${acc}`), [value, acc]);
R.mapAccumRight(iterator, "acc", ["a", "b", "c"]);

//=> val=c, acc=acc
//=> val=b, acc=c
//=> val=a, acc=b
```

tacomanator:
@kedashoe I could be misunderstanding but I don't think that it is. Given that your iterator does not change acc, just passes it along, shouldn't the output instead be as follows?
```javascript
//=> val=c, acc=acc
//=> val=b, acc=acc
//=> val=a, acc=acc
```

That's how `mapAccum` works, just in the opposite orderin the opposite order (argument order for iterator and tuple reversed per docs for `mapAccum`):

```javascript
var iterator = (acc, value) => (console.log(`val=${value}, acc=${acc}`),[acc, value]);
R.mapAccum(iterator, "acc", ["a", "b", "c"]);

val=a, acc=acc
val=b, acc=acc
val=c, acc=acc
```

kedashoe:
Looks to me like there is an issue. We brought this over from [Haskell](http://hackage.haskell.org/package/base-4.7.0.1/docs/Data-List.html#v:mapAccumR). We changed what this function does in https://github.com/ramda/ramda/pull/1975. Haskell's version has the same parameter order for the Left and Right versions, and iterates the list is the same order, but applies the accumulator differently.

```haskell
-- haskell
Prelude Data.List> mapAccumR (\acc val -> (acc ++ "x", val ++ acc)) "x" ["a", "b", "c"]
("xxxx",["axxx","bxx","cx"])
[...]

PR #2390, opened 2017-11-21
Undo change to mapAccumRight accumulator argument order

tacomanator:
The order of arguments expected by mapAccumRight's accumulator function
was reversed in #1975 (c0292498fa798118b70c8d6298f40c74211e1225), to be
consistent with reduceRight. There was a bug in the implementation and
it was decided to undo the change and let the argument order be
consistent with Haskell's mapAccumR in #2386.

### DIFF
--- a/source/mapAccumRight.js
+++ b/source/mapAccumRight.js
@@ -10,14 +10,14 @@ import _curry3 from './internal/_curry3';
  * Similar to [`mapAccum`](#mapAccum), except moves through the input list from
  * the right to the left.
  *
- * The iterator function receives two arguments, *value* and *acc*, and should
- * return a tuple *[value, acc]*.
+ * The iterator function receives two arguments, *acc* and *value*, and should
+ * return a tuple *[acc, value]*.
  *
  * @func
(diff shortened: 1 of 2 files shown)


tacomanator:
Will be a breaking change for anyone that relies on current (incorrect) behavior.

source/mapAccumRight.js
```diff
@@ -26,27 +26,27 @@ import _curry3 from './internal/_curry3';
  * @example
  *
  *      var digits = ['1', '2', '3', '4'];
- *      var append = (a, b) => [a + b, a + b];
+ *      var append = (a, b) => [b + a, b + a];
```
buzzdecafe: this won't work in the repl because `append` is already defined. please give `append` a different name. i realize you didn't introduce this bug, but as long as you are in there ...

tacomanator: Ah yes I noticed that bug but forgot to fix, thank you for reminding me.

kedashoe:
:+1: ty!

PR #2394, opened 2017-11-22
reimplementing propOr atop pathOr, fixes #2391

### LINKED ISSUE #2391: Discrepancy between propOr and pathOr
See REPL link here: https://goo.gl/7UGda4

It seems to me that `propOr` and `pathOr` behave differently.

```
const getProp1 = R.propOr([], 'x');
const getProp2 = R.pathOr([], ['x']);

getProp1({x: null});    // returns null
getProp2({x: null});    // returns []
```

Without checking the code, I'm guessing `propOr` uses `has` internally, whereas `pathOr` is just checking for a falsey result.

Is this expected, or is this a bug?

CrossEye:
We've done this elsewhere recently, implementing the `prop` version in terms of the `path` one for consistency.

This meant removing a test; but I don't think that test was really testing the API we *meant* to expose, only our implementation.  If you disagree, let me know, and we can solve this inconsistency in another manner.

### DIFF
--- a/source/propOr.js
+++ b/source/propOr.js
@@ -1,5 +1,5 @@
 import _curry3 from './internal/_curry3';
-import _has from './internal/_has';
+import pathOr from './pathOr';
 
 
 /**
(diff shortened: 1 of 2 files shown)


buzzdecafe:
:cow2:

Bradcomp:
:bowtie:

Issue #2381, opened 2017-11-09
Proposal: Add Clojure's thread-left (->>) as function


yosbelms:
What I propose is a deviation from Clojure's macro. The function should pass the first value to the rest of the functions as it where composed from left to right; functions must be unary.

Example:
```js
// thread :: (x, (a -> b), (b -> c), ... (o -> p)) -> p
let thread = (v, ...fns) => {
  return fns.reduce((acc, f) => f(acc), v)
}

thread(4, inc, add(2)) // 7
```

It looks very close to `pipe`, so, anyone can argue that similar result can be achieved with `pipe` or `compose`. But the problem it addresses (regarding all the functions that Ramda offers) is not about function nor features but code structure. I find it more readable in cases when `pipe` is needed in a not poin-free-style code, in other words, when using `pipe` or `compose` and call the resulting function immediately after the function composition.

A simple example of above could be `pipe(f, g)(a)` it could read as `compose this functions, execute passing 'a'`. The previous example could look trivial and easy to deal with, but the the things starts to get harder when dealing with large compositions. The same code could be written in terms of `thread` as `thread(a, f, g)` which I read like `pipe 'a' though the rest of functions` or `execute this pipeline starting with 'a'`. Noticeably, the second way looks more declarative than the former.

Furthermore, this `thread` function is a good complement for `pipe`, `ifElse`, `cond` or any other high order function. Lets take a look at the following two examples:
```js
// with pipe
thread(4, pipe(inc, add(2)))

// with cond
thread(n, cond([
    [gt(__, 0), always(1)],
    [lt(__, 0), always(-1)],
    [T, always(0)]
]))
```

CrossEye:
While I think I understand your rationale, my first instinct is :-1: on this. 

Ramda is designed around functional composition with parameter order chosen to make such composition easier.

This proposal is aimed at a very different style of coding than Ramda is built around.  Do note that there are proposals to add this to the language as an operator.  (I'm on my mobile so won't look them up but just look for the proposal for the pipeline operator, `¦>`.)

To me

```js
thread(x, f, g)
```

is less readable than

```js
g(f(x)) 
```

so I'm afraid I don't see the point.

yosbelms:
Function composition is easy with Ramda but  when calling the result of that composition inline we have to use `()()` syntax, example:
```js
pipe(
   someFunc1,
   someFunc2,
   someFunc3(),
   someFunc4(a, b),
   someFunc5(c),
   someFunc6)(x)
```
[...]

PR #2398, opened 2017-11-27
add "drive" function ref ramda/ramda#2381

yosbelms:
Notice the docs added must be enhanced since my english is not good.

### DIFF
--- a/source/drive.js
+++ b/source/drive.js
@@ -1,5 +1,5 @@
 import _curry2 from './internal/_curry2';
-import reduce from './reduce';
+import _reduce from './internal/_reduce';
 
 
 /**
@@ -15,14 +15,15 @@ import reduce from './reduce';
  * @param {*} x The value
  * @param {Array} functions A list of functions.
  * @return {*} The result of sequentially applying the list of function to appliying to `x`
+ * @see R.applyTo, R.pipe, R.compose
  * @example
  *
  *      var add2MultBy3 = R.drive(R.__, [R.add(2), R.mult(3)]);
  *      add2MultBy3(2); //=> 12
  */
 var drive = _curry2(function drive(x, fns) {
-  return reduce(function(acc, f) {
+  return _reduce(function(acc, f) {
     return f(acc);
   }, x, fns);
 });
-module.exports = drive;
+export default drive;
--- a/source/drive.js
+++ b/source/drive.js
@@ -1,0 +1,28 @@
+var _curry2 = /*#__PURE__*/require('./internal/_curry2');
+var reduce = /*#__PURE__*/require('./reduce');
+
+/**
+ * Accepts a value and a list of functions and returns the result of
+ * passing the value to the leftmost function which its return value will
+ * be passed to the next functions in the list.
+ *
+ * @fun
+ * @memberOf R
+ * @since v0.25.0
+ * @category Function
+ * @sig a -> [(a -> b), (b -> c) ,... (d -> e)] -> e
+ * @param {*} x The value
+ * @param {Array} functions A list of functions.
+ * @return {*} The result of sequentially applying the list of function to appliying to `x`
+ * @example
+ *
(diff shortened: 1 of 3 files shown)


source/drive.js
```diff
@@ -0,0 +1,28 @@
+import _curry2 from './internal/_curry2';
+import reduce from './reduce';
```
Bradcomp: Given that we aren't relying on the currying provided by our public reduce function, we should use `./internal/_reduce` here instead of the public version.

source/drive.js
```diff
@@ -0,0 +1,28 @@
...
+ * be passed to the next functions in the list.
+ *
+ * @fun
+ * @memberOf R
+ * @since v0.25.0
+ * @category Function
+ * @sig a -> [(a -> b), (b -> c) ,... (d -> e)] -> e
+ * @param {*} x The value
+ * @param {Array} functions A list of functions.
+ * @return {*} The result of sequentially applying the list of function to appliying to `x`
```
Bradcomp: We should add `@see R.pipe, R.compose`  here.

source/drive.js
```diff
@@ -0,0 +1,28 @@
...
+ *
+ *      var add2MultBy3 = R.drive(R.__, [R.add(2), R.mult(3)]);
+ *      add2MultBy3(2); //=> 12
+ */
+var drive = _curry2(function drive(x, fns) {
+  return reduce(function(acc, f) {
+    return f(acc);
+  }, x, fns);
+});
+module.exports = drive;
```
Bradcomp: This should be `export default drive`

test/drive.js
```diff
@@ -0,0 +1,22 @@
...
+
+  it('is a curried function', function() {
+    var x = 2;
+    var g = R.drive(R.__, [R.add(3)]);
+    eq(g(x), 5);
+  });
+
+  it('pass the value through a list of functions', function() {
+    eq(R.drive(2, [R.add(3), R.multiply(3)]), 15);
+  });
```
Bradcomp: Should there be a test (unit or property based) that ensures this function performs the same as `pipe`?

for all `a`, `...fs`: `drive(a, [fs]) == pipe(...fs)(a)`

test/drive.js
```diff
@@ -0,0 +1,22 @@
...
+
+  it('is not a variadic function', function() {
+    eq(typeof R.drive, 'function');
+    eq(R.drive.length, 2);
+  });
+
+  it('is a curried function', function() {
+    var x = 2;
+    var g = R.drive(R.__, [R.add(3)]);
+    eq(g(x), 5);
```
Bradcomp: This tests that it works with the placeholder, but it doesn't check the other way.

leothorp:
Maybe I'm missing something- but is this not the same functionality that `pipe` already offers?

Bradcomp:
@leothorp See here: https://github.com/ramda/ramda/issues/2381 for discussion.

test/drive.js
```diff
@@ -0,0 +1,22 @@
...
+
+
+describe('drive', function() {
+
+  it('is not a variadic function', function() {
+    eq(typeof R.drive, 'function');
+    eq(R.drive.length, 2);
+  });
+
+  it('is a curried function', function() {
```
CrossEye: I don't think this test is necessary.  We've been moving away from explicitly noting the currying.

test/drive.js
```diff
@@ -0,0 +1,22 @@
...
+    var x = 2;
+    var g = R.drive(R.__, [R.add(3)]);
+    eq(g(x), 5);
+  });
+
+  it('pass the value through a list of functions', function() {
+    eq(R.drive(2, [R.add(3), R.multiply(3)]), 15);
+  });
+
+});
```
CrossEye: I think we should have a test to show what happens if the list of functions is empty.

CrossEye:
@yosbelms: Note the linting error: https://travis-ci.org/ramda/ramda/jobs/308977297

Issue #2396, opened 2017-11-26
Documentation - not clear that arrays are allowed in the first argument of applySpec


JLRishe:
I spent about an hour and a half looking for a method that could take an array of functions, apply them to an arbitrary value, and produce an array of the results. 

After some experimentation, I found out that `applySpec` is able to do precisely that:

```js
R.applySpec([n => n + 1, n => n - 1])(5);    //=> [6, 4]
```

and that it even works on arrays within objects and objects within arrays, ad infinitum. But the documentation doesn't say anything about this.

I think it would be helpful if the docs mentioned that `applySpec` is able to do this, or is this deliberately undocumented?

davidchambers:
This is not an answer to your question, but this is how I would express your algorithm in code:

```javascript
//    T :: a -> (a -> b) -> b
const T = x => f => f(x);

R.map(T(5), [R.add(1), R.add(-1)]);  // => [6, 4]
```

The T combinator was added in #2309 and will be available—as `R.applyTo`—in the next release.

JLRishe:
@davidchambers Thanks for your comment. In my case, `applySpec` is perfect for my needs because the functions I'm applying are constant and it's the value that changes, so it's helpful to have the arguments in the order `[functions]`, `value` rather than `value`, `[functions]`.

This is how I'm actually using it:

```js
// DOMNode -> [DOMNode]
const allDescendants = 
    node => compose(flatten, map(applySpec([identity, allDescendants])), allChildNodes)(node);

// doing this instead of filtering allDescendants in order to avoid potentially 
// building a huge array and filtering it
// (String, String) -> DOMNode -> [DOMNode]
const descendants = 
    (localName, uri) => node => compose(
            flatten, 
            applySpec([filter(hasName(localName, uri)), map(descendants(localName, uri))]), 
            allChildNodes
    )(node);
```

I'm very new to functional programming, so maybe there's a better way to go about this, but it produces the desired output.

CrossEye:
A PR to update the the documentation would be apt.  This works only by accident of the implementation, but it's behavior worth keeping if we ever change the function.  So a new test case would be a good idea too.

JLRishe:
@CrossEye Sure, I'll prepare a PR with some documentation adds and unit tests. Thanks.

PR #2401, opened 2017-11-29
Proposed documentation and unit tests for #2396

JLRishe:
#2396

### DIFF
--- a/source/applySpec.js
+++ b/source/applySpec.js
@@ -9,19 +9,20 @@ import values from './values';
 
 
 /**
- * Given a spec object recursively mapping properties to functions, creates a
- * function producing an object of the same structure, by mapping each property
- * to the result of calling its associated function with the supplied arguments.
+ * Given a spec hierarchy of objects and/or arrays recursively mapping properties/elements
+ * to functions, creates a value of the same structure, by mapping each 
+ * property/element to the result of calling its associated function with 
+ * the supplied arguments.
  *
  * @func
  * @memberOf R
  * @since v0.20.0
  * @category Function
- * @sig {k: ((a, b, ..., m) -> v)} -> ((a, b, ..., m) -> {k: v})
- * @param {Object} spec an object recursively mapping properties to functions for
- *        producing the values for these properties.
- * @return {Function} A function that returns an object of the same structure
- * as `spec', with each property set to the value returned by calling its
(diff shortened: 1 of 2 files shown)


source/applySpec.js
```diff
@@ -9,19 +9,20 @@ import values from './values';
...
  * @func
  * @memberOf R
  * @since v0.20.0
  * @category Function
- * @sig {k: ((a, b, ..., m) -> v)} -> ((a, b, ..., m) -> {k: v})
- * @param {Object} spec an object recursively mapping properties to functions for
- *        producing the values for these properties.
- * @return {Function} A function that returns an object of the same structure
- * as `spec', with each property set to the value returned by calling its
+ * @sig {k: ((a, b, ..., m) -> v)} | [((a, b, ..., m) -> v)] -> ((a, b, ..., m) -> {k: v} | [v])
```
CrossEye: I'm not really sure how to write a good signature for this with its recursive nature as well as the variadic functions.  But this one simply confuses me.

I don't know if we're better with a list of related signatures, but they would at least individually be much simpler:

```hs
applySpec :: {k: a -> v} -> (a -> {k: v}
          :: {k: (a, b) -> v} -> ((a, b) -> {k, v})
          :: {k: (a, b, c) -> v} -> ((a, b, c) -> {k, v})
          :: ...
          :: {[(a -> v)] -> (a -> [v]) 
          :: {[((a, b) -> v)] -> ((a, b) -> [v]) 
          :: {[((a, b, c) -> v)] -> ((a, b, c) -> [v]) 
          :: ...
```

But that still does not capture the recursive nature of this.  I now very much regret that this is recursive; we seem to have abandoned our focus on simplicity here.

customcommander:
@CrossEye Shall we close this for now?

customcommander:
@CrossEye I'm tempted to reopen as we have docs and test cases. Perhaps we can simply drop the type signature for now?

CrossEye:
Reopened, but with the focus on the signature, I don't remember ever really looking over those examples and docs.  I like the `[dec, identity, inc]` one, but I'm less thrilled with the `always('add')` one.  In all of these cases, I'd rather use simple arrow functions than require someone to understand other Ramda functions to follow the docs for this one.

I won't have time this evening, but I'll try soon to make suggested changes.

PR #2408, opened 2017-12-07
Simple fix to the traverse example so that it works in the REPL

arizonatribe:


JasonSooter:
@arizonatribe Should the comments referring to the return values be changed as well adding ```Maybe.``` to the front of each of them. Maybe just a detail for a newbie learning about Maybe/Just

PR #2409, opened 2017-12-08
Added documentation related to changes in default export

chiraggshah:
Added note for developers using ramda version > 0.25

Ramda versions > 0.25 don't have a default export.
So instead of `import R from 'ramda';`, one has to use `import * as R from 'ramda';`
Or better yet, import only the required functions via `import { functionName } from 'ramda';`


CrossEye:
Thank you very much.  This is a good idea.

PR #2412, opened 2017-12-09
Fix docs for startsWith and endsWith

plumdog:
- Fix the signatures
- Rewrite the docstring - hopefully to make it a bit clearer, and to mention that they works for strings too.

### DIFF
--- a/source/endsWith.js
+++ b/source/endsWith.js
@@ -3,14 +3,16 @@ import equals from './equals';
 import takeLast from './takeLast';
 
 /**
- * Checks if a list ends with the provided values
+ * Checks if a list ends with the provided sublist.
+ *
+ * Similarly, checks if a string ends with the provided substring.
  *
  * @func
  * @memberOf R
  * @since v0.24.0
  * @category List
- * @sig [a] -> Boolean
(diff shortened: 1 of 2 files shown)


CrossEye:
I don't have time tonight to dig into this.  I can't recall if this behavior on strings was intentional or simply fell out of the implementation.

If it's the latter, then we need to consider this a little more carefully.  My take is always that we document only intentional behavior.  If it's something that we didn't intend, we don't write tests for it, we don't document it, and we make no promises that it will survive in subsequent versions.

I'll try to look into it soon.

asaf-romano:
Yes, it intentionally works for strings too.

asaf-romano:
They were in fact the original use case, see https://github.com/ramda/ramda/issues/1842

asaf-romano:
🌷

mattgrande:
Strings was the original intent, but I included unit tests for both string and arrays, since all of the underlying functions that are used by this (`take` and `takeLast`) function on both.

See also: #2120

CrossEye:
Thanks.  I thought I remembered something like that.

PR #2417, opened 2017-12-21
Allow value-returning functions in pipeP (and composeP)

antogyn:
The content of this PR makes the behaviour of `R.pipeP` and `R.composeP` more consistent by allowing value-returning functions, even for the first function. It works by wraping in a promise the return value of the first fonction in `R.pipeP`. 

Right now, this kind of composition does not work :
```js
R.pipeP(R.inc, R.inc)(1); // Uncaught TypeError: f.apply(...).then is not a function
```
however, since the implementation uses `.then`, it works if the first function returns a Promise : 
```js
R.pipeP(Promise.resolve.bind(Promise), R.inc, R.inc)(1); // Promise {<resolved>: 3}
```

With this change the first example would work, and swapping functions around in the second example wouldn't break it.

shortcomings : 
- it requires the global `Promise` object since it uses `Promise.resolve`, so in its current form it is probably not mergeable
- it doesn't stick to the documentation

Is it something that can be considered? Maybe in a new function?

### DIFF
--- a/source/pipeP.js
+++ b/source/pipeP.js
@@ -23,9 +23,10 @@ import tail from './tail';
  *      var followersForUser = R.pipeP(db.getUserById, db.getFollowers);
  */
 export default function pipeP() {
-  if (arguments.length === 0) {
+  var args = arguments;
+  if (args.length === 0) {
     throw new Error('pipeP requires at least one argument');
   }
-  return _arity(arguments[0].length,
-                reduce(_pipeP, arguments[0], tail(arguments)));
+  return _arity(args[0].length,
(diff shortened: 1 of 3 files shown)


CrossEye:
We would actually like to remove these functions entirely.  The Ramda team seems to almost universally prefer `Future`s or `Task`s over `Promise`s.  So I feel little need to enhance them.

See #1869 and #1906 for the discussions.

antogyn:
Thank you for your answer. I also prefer Futures, it's too bad they aren't as idiomatic as Promises but it definitely makes sense in the case of Ramda to remove these functions. `R.then` seems like a good compromise.

I'm closing this PR

Issue #2418, opened 2017-12-24
R.prepend and R.append should work for strings too (String → String → String)


GingerPlusPlus:
[`R.concat`](http://ramdajs.com/docs/#concat) does work for strings.

GingerPlusPlus:
`R.concat(s)` or `R.concat(R.__, s)` could be used instead, but I think `R.prepend(s)` and `R.append(s)` just sound better.

GingerPlusPlus:
At least for me, they map more naturally to what I wanna do. For me, curried `R.concat` is less readable than equivalent arrow function.

Bestulo:
This, a few hours ago I spent some minutes looking for the right function for my module and the first choice in my head was to look for `R.append`, and then I saw that there's no such thing for strings and had to spend a few minutes remembering that it's actually called `R.concat` here and that it's actually inverted, so I had to invert the logic and direction of my pipes.

CrossEye:
I make no guarantees that it would be approved, but I would be happy to see a PR for this.

paldepind:
IMHO using `concat` for strings seems nicer and we already support that.

GingerPlusPlus:
Just saying, changing `append` definition to `list.concat([el])`, ie. using `.concat` instead of `_concat` does enable it to concat strings due to JS's weak typing (I'm not saying that it's a good idea, I don't know). Found no similar trick for `prepend`.

mulholio:
R.flip(R.concat) is a nice way of giving you both append and prepend equivalents

PR #2422, opened 2017-12-27
R.append and R.prepend work for strings

GingerPlusPlus:
I haven't changed `@param` and `@returns` in JSDoc.

#2418.

### DIFF
--- a/source/append.js
+++ b/source/append.js
@@ -1,19 +1,23 @@
 import _concat from './internal/_concat';
 import _curry2 from './internal/_curry2';
+import _isString from './internal/_isString';
 
 
 /**
  * Returns a new list containing the contents of the given list, followed by
  * the given element.
  *
+ * When string is passed as second arguments,
+ * appends first argument to it instead.
+ *
  * @func
  * @memberOf R
  * @since v0.1.0
  * @category List
  * @sig a -> [a] -> [a]
+ * @sig String -> String -> String
  * @param {*} el The element to add to the end of the new list.
  * @param {Array} list The list of elements to add a new item to.
- *        list.
  * @return {Array} A new list containing the elements of the old list followed by `el`.
  * @see R.prepend
  * @example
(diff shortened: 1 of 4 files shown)


CrossEye:
I'd like to see what other members of the @ramda/core team think of this first.  Does this stray too far from our philosophy?  In other words, `concat` works on both Strings and lists (as well as other Semigroups) because we delegate it.

Is there a good reason to extend `prepend`/`append` to Strings?  Are there good reasons not to (besides the obvious API bloat, which might be enough on its own)?

CrossEye:
((:bell:)) @ramda/core

Bradcomp:
I have a somewhat odd middle viewpoint here. I wouldn't mind if `append` and `prepend` _worked_ with strings, but I wouldn't like it if it was documented.

Strings are kinda weird in Javascript because there's no such thing as a `char` type, and because they don't actually work like Arrays either. You can't nest them, for instance.

When I append or prepend something, given the signature, and what the words mean, I know what to expect. 
```
prepend (1, [2, 3, 4])  // => [1, 2, 3, 4] 

prepend ([1], [2, 3, 4]) // => [[1], 2, 3, 4]

prepend ([1, 2], [3, 4]) // => [[1, 2], 3, 4]
```

with that in mind, what should prepend ("ab", "cd") return? Well, the only sensible answers are either "abcd" or ["ab", "cd"], neither of which really really capture the signature of append. The first takes String to be both the unit and the container, while the second assumes you just forgot the Array.

This is, to me, because while String is a Monoid, it isn't a List (while JS Arrays are both a Monoid and a List). `concat` is the function we use to join two monoids, and semantically I think it makes more sense to use that then it does to use these two list-specific functions.

That being said, I think that treating the functions as `concat` and `flip(concat)` for Strings would be a sane default to have, but I certainly wouldn't want to encourage that use.

CrossEye:
@Bradcomp: I'm usually fine with having undocumented, but somewhat useful behavior.  But for me, that is because we don't do much type-checking, and I'm perfectly willing to take whatever behavior falls out of our implementation for those cases we don't explicitly support.

Intentionally adding behavior without documenting is uncomfortable to me.

CrossEye:
@Bradcomp: I agree, but would still love to hear from more of the @ramda/core team.

paldepind:
I don't think this is necessary. `R.prepend("a", "bcd")` is equivalent to `R.concat("a", "bcd")` and `R.append("d", "abc")` is just `R.flip(R.concat)("d", "abc")`.

If JavaScript had had a character type I think supporting that in `R.prepend` and `R.append` would have been nice. But since it doesn't I think using `R.concat` is more appropriate.

CrossEye:
@paldepind: I agree.  I've been on the fence about this one.  But there is no common abstraction between lists and strings that we're trying to support for all our list functions.  Since this is easily achievable with other of our functions, I don't see the need.

GingerPlusPlus:
I agree with you after all. Special casing it is not that good idea.

PR #2433, opened 2018-01-09
Prevent TypeError in case of undefined `obj`

nisargthakkar:


source/pickAll.js
```diff
@@ -21,6 +21,7 @@ import _curry2 from './internal/_curry2';
  */
 var pickAll = _curry2(function pickAll(names, obj) {
   var result = {};
+  var obj = obj || {};
```
CrossEye: While Ramda does not object to internal mutations or reassignments, I really don't like the idea of reassigning a parameter.  Could you alter the parameter name?

buzzdecafe:
@CrossEye I disagree, I don't think the library should paper over the fact that it was passed something nonsensical. Let it fail

CrossEye:
I don't have a strong feeling in favor of this.  I absolutely don't want to go out of our way with hand-holding, but I also hate exceptions.  This seemed a simple solution that doesn't go out of our way.

davidchambers:
Ramda feels lost. We've been having this discussion for years now. :\

nisargthakkar:
@CrossEye I have made the change in the variable. However, I noticed that passing a `null`/`undefined` names array would also lead to a crash due to the `names.length`.

That can be fixed by changing the line to:
```var len = names && names.length || 0;```

I can put both the changes in the same PR and squash commits if it makes sense for both to go together.

nisargthakkar:
@buzzdecafe I think both sides of the argument have their advantages. By not handling in the library, users can catch hold of missed scenarios. But by handling it, we can prevent code from crashing in production where data is probably richer than dev/qa/test environments.

CrossEye:
@nisargthakkar: Those changes are fine.  I don't know that the other change could go along with this; that one starts to feel a bit more odd to me.

I imagine part of the objection from @buzzdecafe has to do with the slippery slope: if we do this little bit of type-checking here, why not that next bit... and then that other one?  Where do we quit doing type checking in a language hostile to it?

I'm not sure the line is crystal-clear to me, but I feel it's something like this: a function answers a certain question regarding its arguments.  If we can supply a definitive answer for that question in the face of unexpected argument values, then we should do so.  If we can't, we let the chips fall where they may.

Of course sometimes whether we can do so depends upon how we phrase the question.  For `pickAll(names, obj)`, one version might be, "What is the minimal object with `names` properties whose values match `obj`?"  A good answer to this when `obj` is `nil` is supplied with your current changes.  But I'm not sure that we have a good answer when `names` is not as expected.  What if someone uses as `names` the value `new Rectangle(3, 5)`?  What if she supplies `{length: 2}`?  These sorts of things make me suspect that such a change is too far down that slippery slope.

CrossEye:
@buzzdecafe:

I was suggesting that we don't have to think of this as a type error.  Something like what we discussed in #848.

buzzdecafe:
Yes, I understand that. Feels more ad hoc/hopeful to me than designed/deliberate. So my objection stands.

PR #2436, opened 2018-01-13
R.contains using string

pito-svk:


pito-svk:
I think I've done it.

CrossEye:
Yes, looks as though you have.  Thanks!

Issue #2380, opened 2017-11-09
composeK errors out if passed only one function as argument


niksosf:
Hi there, I don't know if this is by design (if so, perhaps we could add some explanation to the doc? ) but composeK with just one function passed in throws that compose (not composeK) `compose requires at least one argument` 

The source has this line where it `pop()` the last of the arguments of `composeK` and sending the rest to `compose`, which it clearly will complain if only one was passed to `composeK` to begin with. 
here's an example:
https://goo.gl/uf1sx4

Thanks for reading  this.

davidchambers:
This sounds like a bug to me. Thanks for raising the issue.

niksosf:
Thanks for looking into it. let me know if there's anything I could help.

Thank you * million for this wonderful lib.

kedashoe:
Would you like to submit a PR @niksosf ?

niksosf:
@kedashoe would love to contribute. Is there any code standard that the ramda team follows one should read up first?

buzzdecafe:
@niksosf please start here: https://github.com/ramda/ramda/blob/master/CONTRIBUTING.md

CrossEye:
`composeK` has been removed.  No longer relevant.

PR #2438, opened 2018-01-17
If only one function is passed to composeK, composeK will return that…

niksosf:
Related to issue [#2380 ](https://github.com/ramda/ramda/issues/2380)
where if only one function is passed to `composeK`, `composeK` will try to pass an empty array to `compose` which will throw as it requires at least one argument.

### DIFF
--- a/source/composeK.js
+++ b/source/composeK.js
@@ -39,5 +39,7 @@ export default function composeK() {
   }
   var init = Array.prototype.slice.call(arguments);
   var last = init.pop();
-  return compose(compose.apply(this, map(chain, init)), last);
+  return init.length === 0
+    ? last
+    : compose(compose.apply(this, map(chain, init)), last);
 }


source/composeK.js
```diff
@@ -39,5 +38,9 @@ export default function composeK() {
   }
   var init = Array.prototype.slice.call(arguments);
   var last = init.pop();
-  return compose(compose.apply(this, map(chain, init)), last);
+  if (init.length === 0) {
+    last;
```
skatcat31: Should this have a return in front of it?

source/composeK.js
```diff
@@ -39,5 +38,9 @@ export default function composeK() {
...
+  } else {
```
skatcat31: Assuming previous comment is a return, isn't the else here unneeded?

source/composeK.js
niksosf: Definitely! Should I make a change to re-make a pull request?

source/composeK.js
niksosf: Fair point. we could do something like this
```js
if (init.length === 0) {
    return last;
  }
  return compose(compose.apply(this, map(chain, init)), last);
```

niksosf:
@CrossEye The previous tests must not have covered the case of only one function being passed to composeK, thus never reaching the body of the first if condition. maybe we should add a test for it. If agreed, I am up for it.

source/composeK.js
```diff
@@ -2,7 +2,6 @@ import chain from './chain';
 import compose from './compose';
 import map from './map';
 
-
```
davidchambers: Please revert this change.

yarn.lock
```diff
@@ -0,0 +1,3784 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
davidchambers: Please remove this file from the pull request.

source/composeK.js
```diff
@@ -39,5 +38,8 @@ export default function composeK() {
   }
   var init = Array.prototype.slice.call(arguments);
   var last = init.pop();
+  if (init.length === 0) {
+    return last;
+  }
```
davidchambers: Even nicer, I think:

```javascript
return init.length === 0 ?
  last :
  compose(compose.apply(this, map(chain, init)), last);
```

Expressions are just about always preferable to statements. :)

source/composeK.js
niksosf: Hi there, you mean the @@ -2,7 +2,6 @@ import chain from './chain'; part correct? I am looking at the source now and it is importing chain. If this is a git history thing, how does one remove this bit of change from the commit?

source/composeK.js
niksosf: I was totally going to do that;) then I saw the if statement up top and resorted to conform to my imaginary standard that Ramda team might have had...

source/composeK.js
niksosf: Ahh, it was prettier auto-formatting then eslint restoring the single-quotes that triggered that change. I am going to reset everything to master and reapply all the changes then .

CrossEye:
Yes, I would very much like a test case that shows what this change supports that wasn't supported before.

niksosf:
Sorry took so long, will write extra test cases for it now.

Issue #2355, opened 2017-10-20
Profit from Tree Shaking in Ramda in Webpack


scabbiaza:
Hello, guys!

During my experiments with Tree Shaking in Ramda I found that the maximum profit I can get from it is the reducing a bundle on about 7Kb.

I'm wondering, is this because of the dependencies in Ramda or because of Webpack, or something else?

Here is how I tested it.
I created a file with the next code
```js
import * as R from "ramda"

R.identity()
```

And compiled it two times:<br/>with `ramda@0.24` the size of the bundle was  59.2 kB <br/>and with `ramda@0.25` - 51.2 kB.

In my real project, where I did the same experiment, four bundles that use ramda decreased their size on about 7Kb.

On the other hand, when I imported `identity` function directly from the source file, like this:
```js
import identity from "ramda/src/identity"

identity()
```

size of the bundle was only 916 bytes. That kinda result I was expecting from Tree Shaking.

Code and Webpack configurations I uploaded to this repo:
https://github.com/scabbiaza/ramda-webpack-tree-shaking-examples

benji6:
I've recently spent a lot of time debugging a very similar problem. The way Webpack tree-shaking works is it will still output the unused exports but leave it to the minifier to eliminate them. In the case I was looking at it wasn't eliminating any of the unused exports because the exported modules were accessing or setting properties which the minifier realised could potentially cause side-effects.

Without looking into it properly I would suspect that the unused exports aren't being cleaned up because they are often being wrapped by another function (often some sort of curry function) and as far as the minifier is aware those functions could have side-effects so it just leaves them.

CrossEye:
I think this is worth discussing in terms of better modularization of Ramda.  I'm adding it to the list of items under discussion by the core team.

buzzdecafe:
thanks for the report @scabbiaza  that is good info.

Andarist:
Love the detailed report! I've created rollup example - https://github.com/scabbiaza/ramda-webpack-tree-shaking-examples/pull/1

Resulting bundle - 513 bytes (that is umd wrapper included).

Ramda at the moment is suited for tree shaking as far as it can. Unfortunately webpack with its bug and its poor algorithm (not covering many cases) is not doing well here. 

I've described some findings in comments under the refactoring PR, which I've discovered while working on it. Please read my comments there, starting from [this one](https://github.com/ramda/ramda/pull/2254#issuecomment-328136280). However I see you probably already know most of the things Im describing there.

[...]

PR #2441, opened 2018-01-20
Add docs on reducing bundle size for Ramda.

AndrewSouthpaw:
Based on discussion in several threads, including:

- #2355 
- #1505 
- this [repo](https://github.com/scabbiaza/ramda-webpack-tree-shaking-examples) comparing tree shaking

I figured it'd be a good idea to actually add some of this collective knowledge to the docs, since most people probably don't realize their bundle size may be bigger than they'd expect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,6 +194,31 @@ If you have _PhantomJS_ installed, you can run `testem -l phantomjs` to run the
 tests completely headlessly.
 
 
+Usage
+-----------------
+
+For `v0.25` and up, import the whole library or pick ES modules directly from the library:
+
+```js
+import R, { identity } from 'ramda'
+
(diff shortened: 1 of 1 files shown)


CrossEye:
This looks great to me.  Thanks.

Anyone else have suggestions?  @ramda/core ?

JasonSooter:
Does a distinction between using src and es directories need to be made?

CrossEye:
@JasonSooter: that's probably a good idea.  Any specific suggestions?

CrossEye:
I'm going to merge as it is.  If someone has a good suggestion for changes to document the use of `es` vs `src`, please open another PR.

PR #2450, opened 2018-01-26
Update #length docs to demonstrate function as parameter.

AndrewSouthpaw:
`R.length` can accept a function as a parameter and provide a meaningful result: the number of arguments expected for the function.

[Demo](http://ramdajs.com/repl/#?const%20foo%20%3D%20%28a%2C%20b%2C%20c%29%20%3D%3E%20%7B%7D%0A%0AR.length%28foo%29)

### DIFF
--- a/source/length.js
+++ b/source/length.js
@@ -11,7 +11,7 @@ import _isNumber from './internal/_isNumber';
  * @category List
  * @sig [a] | (*… → *) -> Number
  * @param {Array} item The array or function to inspect.
- * @return {Number} The length of the array or the expected number of parameters for the function.
+ * @return {Number} The length of the array or the expected number of arguments for the function.
  * @example
  *
  *      R.length([]); //=> 0
--- a/source/length.js
+++ b/source/length.js
@@ -9,13 +9,14 @@ import _isNumber from './internal/_isNumber';
  * @memberOf R
  * @since v0.3.0
  * @category List
- * @sig [a] -> Number
- * @param {Array} list The array to inspect.
- * @return {Number} The length of the array.
+ * @sig [a] | (*… → *) -> Number
+ * @param {Array} item The array or function to inspect.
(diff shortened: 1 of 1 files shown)


CrossEye:
We should probably document that it delegates on any object with a number-returning 'length` method.  

```js
R.length(myTrip) // => 743
R.length(rectangle(5, 3)) //=> 5
R.length({length: 1.618034}) //=> 1.618034
``

But I don't think we should call out `function` specifically.

**Comment by davidchambers** [comment]
I suggest removing this function. `R.prop('length')` is clearer and requires no documentation at all. ;)

**Comment by CrossEye** [comment]
> `R.prop('length')` is clearer and requires no documentation at all.

Perhaps.  On the down side, though, it has a promiscuous return type.

**Comment by davidchambers** [comment]
> On the down side, though, it has a promiscuous return type.

The return type is consistent if one treats the argument as a record:

```javascript
R.prop('length') :: { length :: a } -> a
```

CrossEye:
I don't really have any issue removing this.  I don't find it very common or very useful, and there is an easy alternative.  What do others think?  @ramda/core

kurtmilam:
I'm not core, but I think `R.length` is a good candidate for removal.

functionalStoic:
I just used it on Friday. But ```prop('length')``` is simple as well I suppose

Bradcomp:
I use `length` sometimes (I _also_ happened to use it on Friday). I don't necessarily have a problem with `prop('length')` other than that I like to use non-stringy parameters whenever I can to prevent typos. 

`R.lenght` would be caught by my IDE set up, whereas `R.prop('lenght')` would most likely slip through. While this isn't really a big deal, it can be a pain in the butt sometimes.

Overall I don't have strong feelings about this one.

AndrewSouthpaw:
> whereas R.prop('lenght') would most likely slip through

Unless you have spellcheck enabled, in which case it would probably warn out. 😃

PR #2451, opened 2018-01-28
add example to defaultTo

dimitropoulos:
This is to clarify that some falsey values are allowed through.  I realize that the reader can deduce that, but this simple example might be a flag to the uninitiated that this isn't a falsey check.


CrossEye:
LGTM.  Any objections?

paldepind:
Thanks for the contribution @dimitropoulos 👍

PR #2452, opened 2018-01-30
Rename contains to includes

### LINKED ISSUE #2447: Rename contains to includes?
Since JavaScript has both [Array#includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) and [String#includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) I was wondering if we should rename `R.contains` to `R.includes`?

Lodash renamed their `_.contains` to `_.includes` a while ago.

It seems that going forward "includes" is going to be what most JavaScript programmers expect this function to be called.

I can do a PR with the renaming and depreciation in case we decide on the renaming.

paldepind:
Rename `R.contains` to `R.includes` and deprecate `R.contains`.

The internal `_containsWith` is also renamed for consistency.

Closes #2447.


source/contains.js
```diff
@@ -1,4 +1,4 @@
-import _contains from './internal/_contains';
+import _includes from "./internal/_includes";
```
CrossEye: `'` -> `"`

source/includes.js
```diff
@@ -0,0 +1,27 @@
+import _includes from "./internal/_includes";
+import _curry2 from "./internal/_curry2";
+
```
CrossEye: `'` -> `"`

paldepind:
@CrossEye Thank you. I've fixed the quotes 😸

semmel:
As I understand the motivation for this change is three-fold (see #2447 ):

> Since JavaScript has both Array#includes and String#includes I was wondering if we should rename R.contains to R.includes?

On the other hand there are the ES6 collections with `Map.prototype.has`, `Set.prototype.has`, ... So the ES committee itself seems to be undecided naming this  *check for element* thing. 
So instead could we not merge `R.has` and `R.contains` to a new `R.has` which accepts both lists and objects?

>Lodash renamed their _.contains to _.includes a while ago.

Did they have any other reasons to do so? IMO one uses either lodash/fp or Ramda, not both.

>It seems that going forward "includes" is going to be what most JavaScript programmers expect this function to be called.

Maybe, yes. It's a good thing if Ramda functions are named according to the expectations of the average JS programmer. 

It is this last point which could win me over while with disgruntlement converting my source code...

paldepind:
@semmel 

> On the other hand there are the ES6 collections with `Map.prototype.has`, `Set.prototype.has`, ... So the ES committee itself seems to be undecided naming this  *check for element* thing. 
So instead could we not merge `R.has` and `R.contains` to a new `R.has` which accepts both lists and objects?

I think you have a point. But I also think it's possible to "explain" the names that the committee has chosen: A `Map` has both keys and values. `has` tells you if a map has a given _key_. `include` on arrays tells you if the array has a given _value_.

> Maybe, yes. It's a good thing if Ramda functions are named according to the expectations of the average JS programmer.

Yes 😄 That is the goal. The fact that both JavaScript and Lodash uses `includes` mean that `contains` will increasingly be forgotten?

source/includes.js
```diff
@@ -0,0 +1,27 @@
...
+import _curry2 from './internal/_curry2';
+
+/**
+ * Returns `true` if the specified value is equal, in [`R.equals`](#equals)
+ * terms, to at least one element of the given list; `false` otherwise.
+ * Works also with strings.
+ *
+ * @func
+ * @memberOf R
+ * @since v0.1.0
```
evan-king: `contains` has been around since `0.1.0`, while `includes` has only been around since this PR went into a release.

And at least 3 times now I've countered "includes is not a function" on a project at work, double-checked the Ramda documentation,  and concluded "nope, this is correct and has been since 0.1.0 so it's definitely not because we're not using latest Ramda version."

I don't need to tell you that fixing a mistake is very difficult when you're confidently ruled out the actual solution.

Bradcomp: This has been fixed on master: https://github.com/ramda/ramda/blob/master/source/includes.js

It will be included in the next documentation release.

PR #2454, opened 2018-01-31
Removed references to deprecated function in docs

Xorhail:
I'm assuming R.mergeDeep was deprecated leaving some dead links in the docs.


paldepind:
Thanks a lot @Xorhail :+1:  Leaving dead-links to removed functions certainly isn't a good idea.

PR #2460, opened 2018-02-03
Arity length explicitly defined

### LINKED ISSUE #2459: dead code elimination cause problems with _arity
Some of the dead code eliminators (like babel-plugin-minify-dead-code-elimination) remove arguments from function's definition. Which cause problem with `_arity.js`

from:
`case 2: return function(a0, a1) { return fn.apply(this, arguments); };`
will become:
`case 2: return function() { return fn.apply(this, arguments); };`

and it will cause problems in combinations like `uncurryN and compose` because they are depending on the `function.length` returned from `_arity`.

I would like to open discussion if `_arity` can be more defensive. Because the function really depends on the dead code.

One of the possible solutions can be

```javascript
const arity2 = function() { return fn.apply(this, arguments);
arity2.arity = 2;
return arity2;
```

And then it is possible to check a function length by specified arity (if not specified than fallback to length).

What do you think about that?

problematic example:
```javascript
const lastFn = curry((firstArg, secondArg) => {
  console.log("lastFn-firstArg", firstArg);
  console.log("lastFn-secondArg", secondArg);
  return "finish";
});

const firstFn = (firstArg) => {
  console.log("firstFn-firstArg", firstArg); // there will be undefined when minified
  return firstArg;
};
const incorrectBehaviour = uncurryN(2, compose(lastFn, firstFn));


you can play with this example at
https://github.com/aizerin/dead-code-issue-example
just type yarn and yarn start
```
I have encountered the same problem with less complicated and usual code, but I cannot remember now.

note: args elimination can be disabled in the plugin mentioned above.

aizerin:
PR for the issue #2459 
I tried to revert `07ecdac`, but it was also removed by dead-code elimination.
So the best solution I found is to use `Object.defineProperty` as @CrossEye suggested. And It works fine. (I kept property attributes same as the specs.)

I discovered that the mocha will not run tests in subdirectories, so the internal tests were not loaded. I've added `--recursive` flag to the runner. Then there was an old test `test.examplesRunner.js` with outdated functions, so I just skip it.

closes #2459

### DIFF
--- a/source/internal/_arity.js
+++ b/source/internal/_arity.js
@@ -1,17 +1,15 @@
+import _isInteger from './_isInteger';
+
 export default function _arity(n, fn) {
-  /* eslint-disable no-unused-vars */
-  switch (n) {
-    case 0: return function() { return fn.apply(this, arguments); };
-    case 1: return function(a0) { return fn.apply(this, arguments); };
-    case 2: return function(a0, a1) { return fn.apply(this, arguments); };
-    case 3: return function(a0, a1, a2) { return fn.apply(this, arguments); };
-    case 4: return function(a0, a1, a2, a3) { return fn.apply(this, arguments); };
-    case 5: return function(a0, a1, a2, a3, a4) { return fn.apply(this, arguments); };
-    case 6: return function(a0, a1, a2, a3, a4, a5) { return fn.apply(this, arguments); };
-    case 7: return function(a0, a1, a2, a3, a4, a5, a6) { return fn.apply(this, arguments); };
-    case 8: return function(a0, a1, a2, a3, a4, a5, a6, a7) { return fn.apply(this, arguments); };
-    case 9: return function(a0, a1, a2, a3, a4, a5, a6, a7, a8) { return fn.apply(this, arguments); };
-    case 10: return function(a0, a1, a2, a3, a4, a5, a6, a7, a8, a9) { return fn.apply(this, arguments); };
-    default: throw new Error('First argument to _arity must be a non-negative integer no greater than ten');
+  if (!_isInteger(n) || n > 10 || n < 0) {
+    throw new Error('First argument to _arity must be a non-negative integer no greater than ten');
   }
+  var arityFn = function() {
+    return fn.apply(this, arguments);
+  };
+  Object.defineProperty(arityFn, 'length', {
+    value: n,
+    configurable: true
+  });
+  return arityFn;
 }


aizerin:
Build failed on node.js 0.12. This version of node has no support for configuring the property `length`.

http://node.green/#ES2015-misc-miscellaneous-function--length--is-configurable

aizerin:
> First of all, if this is the only technique that will work, we should probably mark this for a post-v1.0 release, as until then, we expect to be able to run in older environments. Is there an annotation to allow us to skip dead-code elimination in a block or a file?

Unfortunately I didn't find any annotation or the way how to skip the file. I will try to find some more compatible solution.

> I thought we'd gotten rid of these. But if not, we probably should. Ramda has taken the stand that only the public API should have unit tests. Internal is just that, internal. We offer no contracts on it and can replace or eliminate any part of that whenever we choose.

Ok, no problem I dropped the commits

source/internal/_arity.js
```diff
@@ -1,17 +1,15 @@
...
-    case 4: return function(a0, a1, a2, a3) { return fn.apply(this, arguments); };
-    case 5: return function(a0, a1, a2, a3, a4) { return fn.apply(this, arguments); };
-    case 6: return function(a0, a1, a2, a3, a4, a5) { return fn.apply(this, arguments); };
-    case 7: return function(a0, a1, a2, a3, a4, a5, a6) { return fn.apply(this, arguments); };
-    case 8: return function(a0, a1, a2, a3, a4, a5, a6, a7) { return fn.apply(this, arguments); };
-    case 9: return function(a0, a1, a2, a3, a4, a5, a6, a7, a8) { return fn.apply(this, arguments); };
-    case 10: return function(a0, a1, a2, a3, a4, a5, a6, a7, a8, a9) { return fn.apply(this, arguments); };
-    default: throw new Error('First argument to _arity must be a non-negative integer no greater than ten');
+  if (!_isInteger(n) || n > 10 || n < 0) {
+    throw new Error('First argument to _arity must be a non-negative integer no greater than ten');
```
GingerPlusPlus: There's no point to limiting arguments to 10 with this implementation.

source/internal/_arity.js
```diff
@@ -1,17 +1,15 @@
...
   }
+  var arityFn = function() {
+    return fn.apply(this, arguments);
```
GingerPlusPlus: Old implementation of `_arity`, in addition to setting length, ignores further arguments, so you can do things like 
```js
['1', '1', '1'].map(_arity(1, parseInt))
```

source/internal/_arity.js
```diff
@@ -1,17 +1,15 @@
...
-    case 6: return function(a0, a1, a2, a3, a4, a5) { return fn.apply(this, arguments); };
-    case 7: return function(a0, a1, a2, a3, a4, a5, a6) { return fn.apply(this, arguments); };
-    case 8: return function(a0, a1, a2, a3, a4, a5, a6, a7) { return fn.apply(this, arguments); };
-    case 9: return function(a0, a1, a2, a3, a4, a5, a6, a7, a8) { return fn.apply(this, arguments); };
-    case 10: return function(a0, a1, a2, a3, a4, a5, a6, a7, a8, a9) { return fn.apply(this, arguments); };
-    default: throw new Error('First argument to _arity must be a non-negative integer no greater than ten');
+  if (!_isInteger(n) || n > 10 || n < 0) {
+    throw new Error('First argument to _arity must be a non-negative integer no greater than ten');
   }
+  var arityFn = function() {
```
GingerPlusPlus: It could be changed to `arityFn = fn` to avoid creating extra function. `fn` is always a function created by Ramda that is not reused for anything else, so it should be safe to mutate it.

CrossEye:
To revist after `v1.0`

PR #2462, opened 2018-02-06
Fix order of examples for sortBy.

chrokh:
The two independent examples of `sortBy` seems to be accidentally "overlapped".

Removes overlapping and adds blank line between the two independent examples.


CrossEye:
Thank you.  Definitely an improvement.

PR #2465, opened 2018-02-07
add @see for startsWith and endsWith

leo19920823:


source/endsWith.js
```diff
@@ -16,6 +16,7 @@ import takeLast from './takeLast';
  * @param {*} suffix
  * @param {*} list
  * @return {Boolean}
+ *  * @see R.startsWith
```
Bradcomp: Can you fix the extra asterisk here?

CrossEye:
Thank you!

PR #2467, opened 2018-02-08
update sig and example for lift/liftN function

running-grass:


source/liftN.js
```diff
@@ -13,14 +13,21 @@ import map from './map';
  * @memberOf R
  * @since v0.7.0
  * @category Function
- * @sig Number -> (*... -> *) -> ([*]... -> [*])
+ * @sig Number -> (a, b, c, ... -> z) -> ([a], [b], [c], ... -> [z])
```
CrossEye: Shouldn't this  one end at `n`, not `z`?

leo19920823:
@CrossEye  I have fixed it

CrossEye:
I'm sorry, I may have been too terse.  I meant that for `liftN`, perhaps instead of `a, b, c, ..., z`, we should always have `a, b, c, ..., n`.  It should be clear that the `n` here is related to the one in `liftN`, the first parameter.

leo19920823:
@CrossEye  But , If n is less than the number of parameters of the function, What will happen? crash?

CrossEye:
Eventually handled in #3109

PR #2471, opened 2018-02-16
Readme: Import changed in v0.25

andys8:
The readme explains that a default import is not existing anymore. The usage example was not adapted and shouldn't be working. I changed it.


andys8:
`* as` can not be combined with a named import. I changed it to destructuring. Not sure if it's the preferred way, but the current example of the readme will not work and we should change it.

CrossEye:
Thank you.  Good catch!

PR #2472, opened 2018-02-17
Changed var to const in examples

### LINKED ISSUE #2470: Examples could use const/let instead of var
The documentation uses `var`, e.g. for [cond](http://ramdajs.com/docs/#cond).

```js
var fn = R.cond([
  [R.equals(0),   R.always('water freezes at 0°C')],
  [R.equals(100), R.always('water boils at 100°C')],
  [R.T,           temp => 'nothing special happens at ' + temp + '°C']
]);
```

I'd prefer using `const` and it should be better practice. Is there a cause why `var` is used instead? Or an opinion against?

andys8:
This PR changes all examples used in the documentation to use `const` instead of `var`. I used a regex to replace all occurrences in the sources. I guess the updated version and the deployment will happen with the next release.

Fixes #2470


CrossEye:
Thank you very much.  This is a simple, but very nice, fix.

PR #2473, opened 2018-02-17
Travis using node in version 8

andys8:
I'd suggest since node 8 is becoming the new default lts version, it should be added to the travis build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,5 +4,6 @@ node_js:
 - '0.12'
 - 4
 - 6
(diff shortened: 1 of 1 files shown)


andys8:
Here is the log of the build using node 8.9.4 and working.
https://travis-ci.org/ramda/ramda/jobs/342738409#L438-L439

CrossEye:
Absolutely.  I wonder if we should also remove 4...

Issue #1734, opened 2016-04-15
R.all(R.always(false),[]) === true?


afharo:
Hi, just found out that R.all, no matter the predicate, it always returns `true` for an empty array `[]`.
Is this intended or a bug?

I saw in `test/all.js`
```Javascript
it('returns true for an empty list', function() {
  eq(R.all(T, []), true);
});
```
so I think it is intended but I don't understand why.

Documentation says __`true` if the predicate is satisfied by every element, `false` otherwise__ so I expect that if predicate cannot be satisfied because there are no elements in the array, then predicate isn't satisfied at all => return `false`.

TheLudd:
This is the correct behavior. You can think of it this way: `all` will return `false` if at least one element yields the result `false`. There is no such element in an empty array.

The opposite is true for `any` which will return `true` if at leas one element yields the result `true`, which means that `any`will return `false` for an empty array.

#1380

afharo:
Thanks @TheLudd for your reply!
I think documentation should say it clearer, because it could lead to be misunderstood as _if predicate isn’t evaluated as true by all elements, return false. So, in empty arrays, as predicate isn't evaluated at all, it leads to a false_.
It's difficult to find a proper phrase to be as clearer as possible with this one. But I understand the reason. Closing my MR :smiley:

benperez:
> I think documentation should say it clearer, because it could lead to be misunderstood

I think if anything, the "false otherwise", should be removed.

```diff
- true if the predicate is satisfied by every element, false otherwise
+ true if the predicate is satisfied by every element.
```

CrossEye:
> I think if anything, the "false otherwise", should be removed.

Possibly, but I don't think this would be enough.  This comes up frequently enough that we should probably be more explicit, and even offer a link to https://en.wikipedia.org/wiki/Vacuous_truth for those really curious.

benperez:
OK, I can open a PR to update the docs for `all` and `any` to be more explicit. Are there any other functions that should be included in this list? Here are the functions I'd consider related:

1. `R.any`
2. `R.all`
3. `R.none`

CrossEye:
That's all I can think of.

Issue #1380, opened 2015-09-12
R.all always returns true for empty lists


TGOlson:
`R.all` returns `true` for any predicate when passed an empty list. Is this the intended behavior? 

```js
R.all(R.equals('a'), []) // => true
```

FWIW, `R.any` returns `false` when passed an empty list.

davidchambers:
All zero of the values in `[]` are equal to `'a'`. ;)

TGOlson:
Indeed, and it looks like the native `array.every` seems to concur, thanks for the response.

CrossEye:
@TGOlson 

Another way to think about this is how `all` and `any` interact:

    all(pred, list) ≍  ¬any(¬pred, list))
    any(pred, list) ≍  ¬all(¬pred, list))

So

    R.all(R.equals('a'), []) ≍ R.complement(R.any(R.complement(R.equals('a')), []))

The right-hand-side will be `false` only if this is `true`:

    R.any(R.complement(R.equals('a')), [])

which is true only if there is some element of `[]` which is not equal to 'a'.  But there is no element of `[]`, so this is `false` and the original statement is `true`.

This is what's known in mathematics as a [vacuous truth][vt].

  [vt]: https://en.wikipedia.org/wiki/Vacuous_truth

PR #2475, opened 2018-02-20
Return false from where and whereEq when testObj is null or undefined

wojpawlik:
Why?

1. `null` and `undefined` are the only things that throw on property access and cause the functions to throw.

2.  Better composition with `pathSatisifes` (example from [my real code](https://github.com/GingerPlusPlus/Rextester-bot-v3/blob/ca3d993b26e6e1429643b53db929da35604ecca7/utils/telegram.js#L5-L8)): 

    ```js
    const isCommand = R.pathSatisfies(
        R.whereEq({ offset: 0, type: 'bot_command' }),
        [ 'entities', 0 ]
    );

    isCommand({ entities: [] })
    // Before: throws TypeError: Cannot read property 'offset' of undefined
    // After: returns false
    ```

3. `where` within `where`: https://github.com/ramda/ramda/pull/2288#issuecomment-329931297

### DIFF
--- a/source/where.js
+++ b/source/where.js
@@ -1,6 +1,7 @@
 import _curry2 from './internal/_curry2';
 import _has from './internal/_has';
 
+var nullObj = Object.create(null);
 
 /**
  * Takes a spec object and a test object; returns true if the test satisfies
@@ -38,8 +39,9 @@ import _has from './internal/_has';
  *      pred({a: 'foo', b: 'xxx', x: 11, y: 20}); //=> false
  */
 var where = _curry2(function where(spec, testObj) {
+  var _testObj = (testObj == null ? nullObj : testObj);
   for (var prop in spec) {
-    if (_has(prop, spec) && !spec[prop](testObj[prop])) {
+    if (_has(prop, spec) && !spec[prop](_testObj[prop])) {
       return false;
     }
   }
--- a/test/where.js
+++ b/test/where.js
@@ -41,6 +41,12 @@ describe('where', function() {
     eq(R.where({}, {a: 1}), true);
   });
 
+  it('returns false if testObj is null or undefined', function() {
+    var spec = { x: R.equals(1), y: R.equals(2) };
+    eq(R.where(spec, null), false);
+    eq(R.where(spec, undefined), false);
+  });
+
   it('matches inherited properties', function() {
     var spec = {
       toString: R.equals(Object.prototype.toString),
--- a/test/whereEq.js
+++ b/test/whereEq.js
@@ -41,6 +41,12 @@ describe('whereEq', function() {
     eq(R.whereEq({}, {a: 1}), true);
   });
 
+  it('returns false if testObj is null or undefined', function() {
+    var spec = { x: 1, y: 2 };
+    eq(R.whereEq(spec, null), false);
+    eq(R.whereEq(spec, undefined), false);
+  });
+
   it('reports true when the object equals the spec', function() {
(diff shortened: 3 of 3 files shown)


CrossEye:
See discussion in #2433 

Also, in the case `where({}, null)`, why would this return `false`?  For every other value, it will (vacuously) return true.

GingerPlusPlus:
> Also, in the case `where({}, null)`, why would this return `false`?

`R.where({})` IMO says "return true for any _object_-like", and `null` and `undefined` are definitely not object-like, they're "less than" empty object, they can't have properties. 

If you feel like it should return `true`, that's doable too:

```js
var nullObj = Object.create(null);

var where = _curry2(function where(spec, testObj) {
  var _testObj = (testObj == null ? nullObj : testObj);
  for (var prop in spec) {
    if (_has(prop, spec) && !spec[prop](_testObj[prop])) {
      return false;
    }
  }
  return true;
});
```

If you feel user should be able to decide what's the right thing to do for their case, `if (testObj == null) return testObj;` would allow them to use `R.defaultTo` for that.


> See discussion in #2433 

> I absolutely don't want to go out of our way with hand-holding. (https://github.com/ramda/ramda/pull/2433#issuecomment-356463134)

This case is not about hand-holding, it's about making the library more composible with itself. Some functions return `undefined`, so some functions should be able to cope with it, when the right answer is clear.

I agree with this reasoning: https://github.com/ramda/ramda/issues/848#issuecomment-76588968

You mentioned #2433, do you have any specific problem or comment you'd like to point me to?

CrossEye:
> You mentioned #2433, do you have any specific problem or comment you'd like to point me to?

Just that this was discussed recently and the two founders of the library were in disagreement.  These are tricky and controversial questions.

GingerPlusPlus:
I searched for "null" in issues and PRs, and I now understand the concern of nullchecking being a slippery slope (https://github.com/ramda/ramda/pull/2433#issuecomment-357556043). People will compose everything with everything, and will ask for nullchecking.

Still, this is one of two ways to enable easy recursion in `R.where`. And it's more explicit than the alternative.

CrossEye:
Adding recursion to `where` is probably less of a motivating factor than would be removing it from `evolve` *et al*.  See #1618 and #2288.

I personally would be happy fixing this so that `null` and `undefined` do not cause this function to throw.  But you might here a different response from @buzzdecafe.

I definitely would prefer to return `true` if the set of conditions are empty for the reasons described in #1380, #1734, and [elsewhere][va].


  [va]: https://github.com/ramda/ramda/search?q=vacuous&type=Issues

GingerPlusPlus:
> I definitely would prefer to return `true` if the set of conditions are empty

I see. It's also more consistent with how `R.prop` and `R.path` treat `null` and `undefined`. I changed it.

Should the names of `testObj` and `_testObj` be swapped? I'm probably overthinking.

CrossEye:
There is a lot of merit here, and I'd like to think about it.  I'm spending some more time on Ramda in recent weeks and expect to spend still more over the next few months.  I'd like to look at this one again.

GingerPlusPlus:
Nice to hear that. I closed it because I realized that this can't make `whereEq` recursive.

I think optional chaining allows to avoid throwing on nil way more cleanly: https://github.com/GingerPlusPlus/ramda/commit/648461e29f6f9eb6e2ee48f5299107b7798c0a96.

wojpawlik:
Feel free to open a new issue then; I'm no longer interested in getting this merged.

Issue #2480, opened 2018-02-24
Proposal: assocBy ?


vitalets:
I've noticed that often I need a variant of [R.assoc](http://ramdajs.com/docs/#assoc) that instead of explicit value can receive function that transforms object and sets the result to the prop.
For example I want to join `x` and `y` props into `z`:
```js
const joinXY = R.pipe(R.props(['x', 'y']), R.join(''));
const f = R.converge(R.assoc('z'), [joinXY, R.identity]);
f({x: 'foo', y: 'bar'}); // => {"x": "foo", "y": "bar", "z": "foobar"}
```
What do you think about `R.assocBy` that allows to do it more straight way:
```js
const f = R.assocBy('z', joinXY);
f({x: 'foo', y: 'bar'}); // => {"x": "foo", "y": "bar", "z": "foobar"}
```
I've tried to create such `assocBy` function: 
```js
const assocBy = R.compose(R.over, R.lens(R.identity), R.assoc);
```
But I only works with separate arguments and can't receive all arguments at once:
```js
// Pass arguments separatly
const f = R.assocBy('z')(joinXY);
f({x: 'foo', y: 'bar'}); // => {"x": "foo", "y": "bar", "z": "foobar"}

// Pass 2 arguments at once
const f = R.assocBy('z', joinXY);
f({x: 'foo', y: 'bar'}); // => function r(e){return...
```

CrossEye:
I think the simplest version with current tools would be

```js
const f = chain(assoc('z'), joinXY)
```

A generic version could be

```js
const assocBy = curry((name, fn, obj) => assoc(name, fn(obj), obj))
```

I'm sure we could make this point-free if we tried hard enough, but it would probably be much less readable.

I keep wanting to find a way to make this work with a `lens`-style getter and setter, but anything that did what you want would likely fail one of the lens laws.  Lenses with `over` would be the obvious solution if your modification function just needed to use the value at the focus.  This however is different.

This doesn't sound like something common enough to put in the library core, but it certainly is worth considering for the [Cookbook][cb].

  [cb]: https://github.com/ramda/ramda/wiki/Cookbook

vitalets:
@CrossEye thanks for the explanation!
I can try to prepare an example for Cookbook. 
But yet I'm impressed with your solution via `chain` and still can't realize how it works 🤔
[...]

PR #2484, opened 2018-03-08
Add docs for `chain` about function as second argument

vitalets:
It was discussed in https://github.com/ramda/ramda/issues/2480#issuecomment-368917060.

### DIFF
--- a/source/chain.js
+++ b/source/chain.js
@@ -12,7 +12,9 @@ import map from './map';
  * Dispatches to the `chain` method of the second argument, if present,
  * according to the [FantasyLand Chain spec](https://github.com/fantasyland/fantasy-land#chain).
  *
- * Acts as a transducer if a transformer is given in list position. `chain(f, g)(x)` is equivalent to `f(g(x), x)`.
+ * If second argument is a function, `chain(f, g)(x)` is equivalent to `f(g(x), x)`.
+ *
+ * Acts as a transducer if a transformer is given in list position.
  *
  * @func
  * @memberOf R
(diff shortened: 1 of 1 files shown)


source/chain.js
```diff
@@ -7,12 +7,12 @@ import map from './map';
...
 /**
  * `chain` maps a function over a list and concatenates the results. `chain`
- * is also known as `flatMap` in some libraries
+ * is also known as `flatMap` in some libraries.
  *
  * Dispatches to the `chain` method of the second argument, if present,
  * according to the [FantasyLand Chain spec](https://github.com/fantasyland/fantasy-land#chain).
  *
- * Acts as a transducer if a transformer is given in list position.
+ * Acts as a transducer if a transformer is given in list position. `chain(f, g)(x)` is equivalent to `f(g(x), x)`.
```
CrossEye: I think this needs something about "When operating on functions".  We don't want to give the impression that chain only works on functions.

Also, I would like to move this before the transducers line.  I'm pretty sure that wherever we use that line, it's the last one of the documentation.

vitalets: Ок. 
As the first argument is always a function maybe write "If second argument is a function" instead of "When operating on functions"?

CrossEye: Yes, anything like that would do.  The point is that we don't want to make people think that it *only* works on functions.

vitalets: Done.

CrossEye:
Thank you very much!  Looks great!

PR #2489, opened 2018-03-21
Update docs and tests for `and` and `or`

johnvh:
This clarifies the doc descriptions for `and` and `or`. Since these use logical operators, they will return one of the operands, not necessarily a boolean. I also added tests so that this behavior is covered.

### DIFF
--- a/source/and.js
+++ b/source/and.js
@@ -2,7 +2,7 @@ import _curry2 from './internal/_curry2';
 
 
 /**
- * Returns `true` if both arguments are `true`; `false` otherwise.
+ * Returns the first argument if it is falsy, otherwise the second argument.
  *
  * @func
  * @memberOf R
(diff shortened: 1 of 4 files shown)


johnvh:
Previously proposed and discussed... good enough for me! I should've dug a little deeper.

functionalStoic:
@CrossEye I've been following the Ramda Github for a bit and I've learned a bunch. I can only imagine you literally laughing out loud when you posted your comment above. Ha! Such a contentious topic

CrossEye:
@johnvh: No reason for you to know.  I just didn't feel like discussing it unless you had a new spin on it.

@functionalStoic: Actually, I just went digging for the most recent version of this, which I knew wasn't too long ago.  No laughing involved, I'm afraid. :wink:

functionalStoic:
@CrossEye Ok, well, maybe not. I have a hard time not laughing a few times through #1218 [().](https://github.com/ramda/ramda/pull/1218#issuecomment-114901352) Such good debate about a tough subject with strong opinions.  Back to work... ;-)

PR #2496, opened 2018-04-02
introduces a mergeSpec() function

arizonatribe:
Created `mergeSpec()` function to handle cases where you may want `applySpec()` to merge with the original object.

An example of its proposed use:

```
const formatUser = mergeSpec({
  fullName: R.compose(R.join(' '), R.values, R.pick(['firstName', 'lastName'])),
  address: R.pipe(
    R.prop('address'),
    R.evolve({
      street: R.trim,
      city: compose(str => str.replace(/(?:^|\s)\S/g, R.toUpper), R.trim),
      state: R.toUpper,
      zip: compose(R.trim, R.when(R.is(Number), R.toString))
    })
  )
});

formatUser({
  firstName: 'Montgomery',
  lastName: 'Burns',
  address: {
    street: '1000 Mammon Lane, ',
    city: 'springfield',
    state: 'or',
    zip: 97403
  }
}); // => { firsName: 'Montgomery', lastName: 'Burns', fullName: 'Montgomery Burns', address: { street: '1000 Mammon Lane,', city: 'Springfield', state: 'OR', zip: '97403' } }
```

I've tried to observe the coding style for the Ramda code-base, however I may not quite be up to its standards. I'll welcome any critiques or requests for more unit tests.

Thanks


test/mergeSpec.js
```diff
@@ -0,0 +1,68 @@
...
+  });
+
+  it('merges new props onto the original object', function() {
+    eq(R.mergeSpec({
+      fullName: R.compose(R.join(' '), R.values, R.pick(['firstName', 'lastName'])),
+      address: R.pipe(
+        R.prop('address'),
+        R.evolve({
+          street: R.trim,
+          city: R.compose(str => str.replace(/(?:^|\s)\S/g, R.toUpper), R.trim),
```
CrossEye: Ramda (for the moment) supports old environments that don't have arrow functions.

Can you switch this to `function(str) {/* ... */}`?

We will be dropping this restriction post 1.0, but we're not there yet.

source/mergeSpec.js
```diff
@@ -0,0 +1,68 @@
...
+             spec);
+  return curryN(reduce(max, 0, pluck('length', values(spec))),
+                function() {
+                  var args = arguments;
+                  if (args.length === 1 && _isObject(args[0])) {
+                    return merge(args[0], map(function(f) { return apply(f, args); }, spec));
+                  }
+                  return map(function(f) { return apply(f, args); }, spec);
+                });
+});
```
CrossEye: Something is wrong with the indentation here.

test/mergeSpec.js
arizonatribe: Sure, no problem

source/mergeSpec.js
arizonatribe: I just tried to figure out what it might be (there weren't any eslint alarms in the final version of the file, but I agree it looks a little weird). I had the `source/applySpec.js` open when creating this file, to use as a guide, so perhaps I shouldn't have been following that one? The formatting `source/anyPass.js` looks a little more like what I should probably follow.

Also, I just realized that a lot of the source code in Ramda line breaks before reaching 80 character on a single line. I can change line 63 to observe that convention too.

arizonatribe: Maybe more like this?

```
var mergeSpec = _curry1(function mergeSpec(spec) {
  spec = map(function(v) { return typeof v == 'function' ? v : mergeSpec(v); },
             spec);
  return curryN(reduce(max, 0, pluck('length', values(spec))), function() {
    var args = arguments;
    if (args.length === 1 && _isObject(args[0])) {
      return merge(args[0], map(function(f) { return apply(f, args); }, spec));
    }
    return map(function(f) { return apply(f, args); }, spec);
  });
});
```

PR #2497, opened 2018-04-02
inRange

cglane:
Shortcut to find if number is within a range of numbers.


CrossEye:
I can see the uses of this, but I think this is a more likely fit for the [Cookbook][co] and/or [ramda-adjunct][ra].

I am curious about the API.  Why would you choose an array, which is quite clearly ordered, for the endpoints, and then not use that order?  That is why does `R.inRange(2, [4, 1])` return `true` rather than either return `false` or throw an exception, either of which would seem more logical to me.

But more basically, why use the array at all?  Why not `inRange(1, 4, 2) //=> true`?

Also, is there any strong reason for being inclusive on both ends of the range?  For instance in `R.range`, we're inclusive on the left, exclusive on the right, which matches some prior art from a few languages, and makes sense if you're partitioning a larger range.  Is there any particularly reason that you include both endpoints?

Finally, I think `between` is a much nicer name for this. 

  [co]: https://github.com/ramda/ramda/wiki/Cookbook

cglane:
Hey Scott,
Thanks for the quick reply.
My inspiration for ‘inRange’ was coming from a python project that iI have been working on where I was passing a python list of line coordinates and finding if a given coordinate was on the line. Therefore >= coordinate would still be on the line.

The reason that [4,1] returns true, is because I am sorting the array— in the event that it is passed out of order. I do recognize your point that is more straight-forwards passing it as two separate inputs.

You are right about the name, it would most likely be more readily ‘searchable’ for someone looking for the aforementioned functionality looking for a ‘between or inBetween’

-Charles

tommmyy:
@cglane Hi, if you can, we would happily accept PR to https://github.com/tommmyy/ramda-extension

see https://github.com/tommmyy/ramda-extension/issues/72

CrossEye:
> The reason that [4,1] returns true, is because I am sorting the array

Yes, I understood the implementation.  The question is *why* you would want it to do that.  Essentially you're creating a function that says

    (val, [start, end]) => (start <= val && val <= end) || (end <= val && val <= start)

What is the advantage of that over

    (val, [start, end]) => start <= val && val <= end) 

??

> in the event that it is passed out of order

Shouldn't that either be an error or simply the equivalent of `val => false`?  Doesn't this obscure real programming errors or data problems?

Undistraction:
Fwiw `inRange` is included in the next release of Ramda Adjunct which should be out in the next few days. See: https://github.com/char0n/ramda-adjunct/pull/436

tommmyy:
Ramda extension will include function between - https://github.com/tommmyy/ramda-extension/pull/73

PR #2498, opened 2018-04-03
add "successor" property test for R.add

iamstarkov:
See more https://en.wikipedia.org/wiki/Addition#Successor

### DIFF
--- a/test/add.js
+++ b/test/add.js
@@ -31,4 +31,8 @@ describe('add properties', function() {
   jsv.property('identity', jsv.number, function(a) {
     return R.add(a, 0) === a && R.add(0, a) === a;
   });
+
+  jsv.property('successor', jsv.number, function(a) {
+    return R.add(a, 1) > a;
+  });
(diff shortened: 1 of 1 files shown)


davidchambers:
This property doesn't hold for `Infinity`, `-Infinity`, or `NaN`.

If you're in the mood you could submit a similar pull request for [`S.add`][1]: its stricter input type means the property should hold for all values in its domain.


[1]: https://sanctuary.js.org/#add

iamstarkov:
`jsverify` disagrees with you, but okay

CrossEye:
This will also fail with anything greater than `Number.MAX_SAFE_INTEGER`

That is, 

```js
Number.MAX_SAFE_INTEGER + 2 > Number.MAX_SAFE_INTEGER + 1 //=> false
```

davidchambers:
Good point, Scott. This means that the property does not hold for `S.add` either.

PR #2500, opened 2018-04-06
Add additional example to prop.js

craigayre:
This PR is to add an additional example when partially applying `prop`.

I use this form the most often and thought the docs could benefit from the addition.

### DIFF
--- a/source/prop.js
+++ b/source/prop.js
@@ -19,6 +19,7 @@ import path from './path';
  *
  *      R.prop('x', {x: 100}); //=> 100
  *      R.prop('x', {}); //=> undefined
+ *      R.map(R.prop('length'), ['Elephant', 'Penguin']) //=> [8, 7]
  */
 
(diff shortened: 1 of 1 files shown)


Bradcomp:
I like the idea of using a partially applied example, but I would prefer a different example. 

`R.map(R.prop(x))` is equivalent to `R.pluck(x)`, and `R.prop('length')` is equivalent to `R.length`, so I wouldn't recommend using `prop` in a case like the above.

Do you think you could use a different example, perhaps with `compose` or something?

`R.compose(R.inc, R.prop('x'))({x: 3})   // => 4` or something to that effect?

craigayre:
Absolutely - the example provided was a bit contrived for brevity. I'll update

CrossEye:
With that change, this looks good to me... but not to our linter.

It's picky about trailing spaces.  Could you update one more time to fix that?

CrossEye:
Thank you!

Issue #2499, opened 2018-04-05
R.over for lens path that does not exist


vitalets:
Assume a task: if object has some path - uppercase the value, otherwise - do nothing.
The first approach I made is following:
```js
const f = R.over(R.lensPath(['x', 'y']), R.toUpper);

f({x: {y: 'foo'}}); // => {"x": {"y": "FOO"}}
```
but if path does not exist I get an error:
```js
f({x: {z: 'foo'}}); // => undefined does not have a method named "toUpperCase"
```
I've constructed the function that works:
```js
const f = R.ifElse(
  R.pipe(R.path(['x', 'y']), R.isNil), 
  R.identity, 
  R.over(R.lensPath(['x', 'y']), R.toUpper)
);

f({x: {y: 'foo'}}); // => {"x": {"y": "FOO"}}
f({x: {z: 'foo'}}); // => {"x": {"z": "foo"}}
```
but it seems to me rather verbose and un-optimal: path `['x', 'y']` is repeated twice.

My questions/suggestions are following:
1. maybe I miss something and it can be achieved in more elegant way without path duplication?
2. is there a standard solution for dealing with non-existent lens? I looks to me a common case.
3. I've noticed the [R.viewOr](https://github.com/ramda/ramda/wiki/Cookbook#viewor) in Cookbook but it does not help. I need to modify path with `R.over` not `R.view`. R.overOr?
4. what about adding `R.hasPath()` function similar to current `R.has()`? It will partially help.

Thanks!

tommmyy:
This is more compat way:

```js
const myLens = R.lensPath(['x', 'y'])
                   
const f = unless(o(isNil, view(myLens)), over(myLens, toUpper));
```
And you can make your more general function like that:

```
const overUnlessIsNil = (lens, fn) => unless(o(isNil, view(lens)), over(lens, fn))

const f = overUnlessIsNil(lensPath(['x', 'y']), toUpper)
```

CrossEye:
I don't think there will be any reasonable way to do this with our lenses.  You might be interested to check out @polytypic's  [partial-lenses library][pl] for a tool that might help.

It's not too hard to write something that works in Ramda without using lenses:

```js
const hasPath = curry((pth, obj) => {
  const parent = path(init(pth), obj)
  return is(Object, parent) && last(pth) in parent
})

const updatePath = curry((pth, fn, obj) => {
  return hasPath(pth, obj) ? assocPath(pth, fn(path(pth, obj)), obj) : obj
[...]

PR #2501, opened 2018-04-07
add hasPath

vitalets:
As discussed in #2499 here is a proposal for `hasPath` function. 

Notes:
1. only own properties are checked
2. imho `hasPathIn` should not be implemented right now (until somebody will come with the request)

### DIFF
--- a/source/has.js
+++ b/source/has.js
@@ -1,5 +1,5 @@
 import _curry2 from './internal/_curry2';
-import _has from './internal/_has';
+import hasPath from './hasPath';
 
 
 /**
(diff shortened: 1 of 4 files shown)


CrossEye:
This looks great.  And I agree about `hasPathIn`.  Does anyone disagree?  @ramda/core?

One thing I would like to add, though, is to reimplement `has` in terms of this.  (`curry2(function has(prop, obj) {return hasPath([prop], obj);}`) That way these two functions cannot get out of sync.

vitalets:
> is to reimplement `has`

Great, will do.

One more thing I'd like to point attention: currently `hasPath()` with zero-length path returns false. Similar to `has()` with empty string. But `path()` with empty path argument returns object itself. Is it ok?
```js
hasPath([], {x: 'foo'}); // => false
path([], {x: 'foo'});    // => {x: 'foo'}
```

CrossEye:
This looks very good to me.  I think this is quite a useful little function.

vitalets:
Thank you! 👍

PR #2505, opened 2018-04-10
flipAll

yazeedb:
I initially thought [R.flip](http://ramdajs.com/docs/#flip) flipped _all_ arguments. Does that use case warrant official support?

### DIFF
--- a/source/flipAll.js
+++ b/source/flipAll.js
@@ -4,10 +4,8 @@ import curryN from './curryN';
 var flipAll = _curry1(function flipAll(fn) {
   return curryN(fn.length, function() {
     var args = Array.prototype.slice.call(arguments, 0);
-    var first = args[0];
-    var rest = args.slice(1);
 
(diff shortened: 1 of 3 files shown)


CrossEye:
Do you want to make a case for this?  I've never needed it that I can remember, and I'm not sure that I see any need for this that wouldn't be better served with a lodash-style `rearg`.

yazeedb:
@CrossEye Hello! No, just raising the question. Didn't know about [lodash's rearg function](https://lodash.com/docs#rearg). Is there interest in adding it to Ramda?

CrossEye:
I have rarely had the need for it.  On those few occasions, it was just as easy for me to write

```js
const g = (b, c, a) => f(a, b, c)
```

perhaps with a `curry` thrown in.  So I again haven't found much of a reason for it.  But I'm willing to listen if someone wants to suggest it.

PR #2506, opened 2018-04-10
Add size snapshot rollup plugin

TrySound:
This diff adds size snapshot plugin for rollup which on every build
writes snapshot of umd with actual, minified and gzipped sizes. This
allows to track changes in size on every commit.

Also this diff combines development and production rollup configs in one
multiconfig to run node with rollup only once.


.size-snapshot.json
```diff
@@ -0,0 +1,7 @@
+{
```
Bradcomp: This file should be added to `.gitignore` as it is generated, so it should not be committed.

TrySound: The idea is to commit size snapshot so on every new commit we could see impact on final size.

TrySound:
Removed lockfile

CrossEye:
While this doesn't hurt, I also don't have strong feeling in favor of it.

Anyone want to make a case for or against this?

TrySound:
There is a case I didn't mentioned. size-snapshot is able to track treeshakability of rollup and webpack for esm bundles. I think this is useful since there are two ways of using this library, with named exports (`import { map } from 'ramda'`) and path imports (`import map from 'ramda/map'`). Sometimes this both ways are used in the same project and this affect bundle size. So I'm gonna propose the solution with size snapshot plugin.

Issue #2289, opened 2017-09-16
Array-friendly evolve


asaf-romano:
Anybody against re-implementing evolve on top of addIndex(map)/mapObjIndex so that it is array-friendly?

Demo: https://goo.gl/rq7DhT

Keep in mind ``evolve`` is already recursive.

CrossEye:
I missed this last month.  I can't see any objection.

PR #2511, opened 2018-04-19
Be able to evolve with an array of transformations

bobwei:
update evolve functions to support evolving with an array of transformations.

An example of use case:

```
var transf   = [R.add(1), R.add(-1)];
var object   = [100, 1400];
var expected = [101, 1399];
eq(R.evolve(transf, object), expected);
```

reference issue
https://github.com/ramda/ramda/issues/2289

### DIFF
--- a/source/evolve.js
+++ b/source/evolve.js
@@ -29,7 +29,7 @@ import _curry2 from './internal/_curry2';
  *      R.evolve(transformations, tomato); //=> {firstName: 'Tomato', data: {elapsed: 101, remaining: 1399}, id:123}
  */
 var evolve = _curry2(function evolve(transformations, object) {
-  var result = {};
+  var result = object instanceof Array ? [] : {};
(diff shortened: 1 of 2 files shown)


CrossEye:
I have no objections to this.  I regret the `evolve` API as it is.  This makes it no worse, and probably makes it better.  Others?  @ramda/core ?

ben-eb:
I'm in favour of this proposal as I'm often doing some mapping on array elements with `converge`, for instance:

```js
converge(pair, [
  compose(inc, head),
  compose(dec, last),
])([1, 3])
```

That could be refactored to:

```js
evolve([inc, dec])([1, 3])
```

CrossEye:
@bobwei: Thank you for this PR.  Sorry it fell off the radar.

@ben-eb: Thank you for bringing this back to our attention.

PR #2512, opened 2018-04-19
Added pipeSome and composeSome functions

skallet:
I've added pipeSome and composeSome function which works similarly like threading functions in Clojure some->>. This is good when you use some expensive operation in composition and you don't want to compute for invalid values. You can, of course, wrap those functions inside `when` function, but I think this approach is much more readable.


CrossEye:
I'm conflicted about this.

I can see the utility, but we have too many `compose` functions as it is.

And although I can't immediately think of better ones, I don't particularly like the names "`pipeSome`" and "`composeSome`".

Anyone else have feedback?

Bradcomp:
Personally, I don't like treating `null` and `undefined` as sentinel values in this way.  

A `composeWhile` function that takes a predicate and an array of functions would be more interesting and potentially useful.

I think this behavior is well captured with existing techniques. One could use `reduce` with `reduced`, or `reduceWhile`, or `when`, or `defaultTo`, or preferably `Maybe`, of which this provides the perfect use case. 

Most interesting to me would be to deprecate `compose/pipeP` and `compose/pipeK` in preference for a `composeWith` function. `composeP` becomes `composeWith(then)` and `composeK` becomes composeWith(chain)`. The functions above could be constructed as 

```
composeWith((f, val) => isNil(val) ? val : f(val));
```

CrossEye:
@Bradcomp:  That makes a lot of sense.

CrossEye:
:+1:

Except that I think we need to resolve the question @Bradcomp raised about `Proxy`  (no, they're not always callable), we should definitely do this.

skallet:
I agree with composeWith function as the more general tool. But I don't think that compose/pipe should be deprecated. In every project you use ramda there will be line with something like this:
`const compose = R.composeWith(R.T);`

So I will rewrite PR to composeWith/pipeWith instead of composeSome/pipeSome, agree?

CrossEye:
@skallet: absolutely.  We do not try to create a bare-bones library.  While we mostly don't include functions that could be trivially implemented with existing ones, if they're generally useful enough, we certainly will do so.  I would not remove `compose`/`pipe`, but I would at least *consider* re-implementing them in terms of `composeWith` and/or `pipeWith`.  I would in the end probably not do that, as `compose`/`pipe` are as central to the library as I can imagine, and I would want to be able to squeeze out whatever performance I could.  But I would definitely consider it.

skallet:
Ok, closing this PR, make another one with composeWith / pipeWith.

PR #2513, opened 2018-04-20
Enhance `_isFunction` To Handle `AsyncFunction` (+ more)

rjhilgefort:
👋 Hello! First time contributor here so let me know what idiomatic Ramda repo things I'm missing from these changes and I'll be more than happy to make updates ASAP.

I was using `R.invoker` on a `node@8.11.1` project and discovered that it was incorrectly throwing when I would try to invoke an `async` function on an object. Here's a basic example of what I was doing:

```js
const person = {
  speak: async () => 'Hi!'
};
const speak = R.invoker(0, 'speak');
speak(person).then(R.identity) //=> 'Hi!'
```

This [works fine in the Ramda Repl](https://goo.gl/g7WxPS) because `Object.prototype.toString` seems to always give back `'[object Function]'`. In `node@8.11.1` however, `toString` gives back different strings for different types of functions and this was producing false negatives when I tried to use `R.invoker` on my `async` methods.

This PR adds support for `AsyncFunction`, as well as `GeneratorFunction` ~and `Proxy`~ functions. I took a page out of `lodash`'s [`isFunction` implementation](https://github.com/lodash/lodash/blob/4.17.5/lodash.js#L11638).

I didn't add any tests because I didn't see any specific tests for the internal `_isFunction` helper, nor could I easily add tests for `R.invoker` that handles `async` functions because of the `ESLint@2.11.0` dev dependency. I see in the `travis.yml` file that you're targeting `node@8.x.x` so I think the changes are valid given that support.

### DIFF
--- a/source/internal/_isFunction.js
+++ b/source/internal/_isFunction.js
@@ -2,6 +2,5 @@ export default function _isFunction(x) {
   var type = Object.prototype.toString.call(x);
   return type  === '[object Function]' ||
     type === '[object AsyncFunction]' ||
-    type === '[object GeneratorFunction]' ||
-    type === '[object Proxy]';
+    type === '[object GeneratorFunction]';
 }
--- a/source/internal/_isFunction.js
+++ b/source/internal/_isFunction.js
@@ -1,3 +1,7 @@
 export default function _isFunction(x) {
-  return Object.prototype.toString.call(x) === '[object Function]';
+  var type = Object.prototype.toString.call(x);
+  return type  === '[object Function]' ||
+    type === '[object AsyncFunction]' ||
+    type === '[object GeneratorFunction]' ||
+    type === '[object Proxy]';
 }
--- a/source/internal/_isFunction.js
+++ b/source/internal/_isFunction.js
@@ -2,5 +2,6 @@ export default function _isFunction(x) {
   var type = Object.prototype.toString.call(x);
   return type  === '[object Function]' ||
     type === '[object AsyncFunction]' ||
-    type === '[object GeneratorFunction]';
+    type === '[object GeneratorFunction]' ||
+    type === '[object AsyncGeneratorFunction]';
 }
--- a/source/invoker.js
+++ b/source/invoker.js
@@ -27,6 +27,16 @@ import toString from './toString';
  *      sliceFrom(6, 'abcdefghijklm'); //=> 'ghijklm'
  *      const sliceFrom6 = R.invoker(2, 'slice')(6);
  *      sliceFrom6(8, 'abcdefghijklm'); //=> 'gh'
+ *
+ *      const dog = {
+ *        speak: async () => 'Woof!'
+ *      };
+ *      const person = {
+ *        speak: async () => 'Hi!'
+ *      };
+ *      const speak = R.invoker(0, 'speak');
+ *      speak(dog).then(R.identity) //=> 'Woof!'
+ *      speak(person).then(R.identity) //=> 'Hi!'
  * @symb R.invoker(0, 'method')(o) = o['method']()
  * @symb R.invoker(1, 'method')(a, o) = o['method'](a)
  * @symb R.invoker(2, 'method')(a, b, o) = o['method'](a, b)
--- a/source/invoker.js
+++ b/source/invoker.js
@@ -31,12 +31,9 @@ import toString from './toString';
  *      const dog = {
  *        speak: async () => 'Woof!'
  *      };
- *      const person = {
- *        speak: async () => 'Hi!'
- *      };
  *      const speak = R.invoker(0, 'speak');
- *      speak(dog).then(R.identity) //=> 'Woof!'
- *      speak(person).then(R.identity) //=> 'Hi!'
+ *      speak(dog).then(console.log) //~> 'Woof!'
+ *
  * @symb R.invoker(0, 'method')(o) = o['method']()
  * @symb R.invoker(1, 'method')(a, o) = o['method'](a)
  * @symb R.invoker(2, 'method')(a, b, o) = o['method'](a, b)


source/internal/_isFunction.js
```diff
@@ -1,3 +1,7 @@
 export default function _isFunction(x) {
-  return Object.prototype.toString.call(x) === '[object Function]';
+  var type = Object.prototype.toString.call(x);
+  return type  === '[object Function]' ||
+    type === '[object AsyncFunction]' ||
+    type === '[object GeneratorFunction]' ||
+    type === '[object Proxy]';
```
Bradcomp: Is a proxy always callable?

rjhilgefort: I'm actually not sure. I played around on the console in a `node` repl and I couldn't come to a conclusion. I personally don't use a lot of proxies and am not that familiar with edge cases. I would be fine removing the check for `'[object Proxy]'` from the list.

If someone feels strongly about keeping it, then perhaps the real answer is to put some guards in `R.invoker`?

CrossEye:
I don't know what happened to my reply.  I *thought* I said that this was a good idea, and that outside the `Proxy` issue, we should just do this.

I think we should drop Proxy, as they are not generally callable.

rjhilgefort:
@CrossEye 
> I think we should drop Proxy, as they are not generally callable.

Updated.

kurtmilam:
@CrossEye your reply ended up over here https://github.com/ramda/ramda/pull/2512#issuecomment-383183459

CrossEye:
Whoops.  Thanks!

rjhilgefort:
> I have to ask, why not just typeof x === 'function'?

```js
class Foo {
}
typeof Foo
// -> 'function'
```

rjhilgefort:
> I have to ask, why not just typeof x === 'function'?

@GingerPlusPlus I did some quick testing and it would seem to handle the cases I tried so ¯\\_(ツ)_/¯. That discussion might be a bit out of the scope of this PR perhaps? I'm merely trying to add support given the current implementation strategy. Thanks for the input!

rjhilgefort:
@Bradcomp or @CrossEye is there anything else you would like to see with this or are we just holding until a release or something to merge a bunch of PRs.

willheslam:
I've found another case where the Function/AsyncFunction (+ GeneratorFunction) difference could potentially have an impact - `map`.

@rjhilgefort's PR is, as it is, perfectly consistent and improves ramda without my changes, so it can be merged without them but I'm personally keen to see `map` extended to support async functions too!

I've created a second PR for my changes here: https://github.com/ramda/ramda/pull/2675

As an aside, in theory ramda's reliance on a stable `Object.prototype.toString` could have broken things for anyone moving from Babel to delivering native ES6, so IMO that's another point towards using `typeof` over `toString`... on the other hand, it's known Babel isn't perfect!

Specifically, Babel doesn't seem to match v8's behaviour (and the latest ecma spec says that AsyncFunctions' @@toStringTag property should be "AsyncFunction", so it looks like v8 is correct here:
https://www.ecma-international.org/ecma-262/9.0/index.html#sec-async-function-prototype-properties-toStringTag
).

```
async function foo(){
}

console.log('typeof', typeof foo)
console.log('toString', Object.prototype.toString.call(foo))
```
transpiles to
```
let foo = (() => {
  var _ref = _asyncToGenerator(function* () {});

  return function foo() {
    return _ref.apply(this, arguments);
  };
})();

function _asyncToGenerator(fn) { return function () { var gen = fn.apply(this, arguments); return new Promise(function (resolve, reject) { function step(key, arg) { try { var info = gen[key](arg); var value = info.value; } catch (error) { reject(error); return; } if (info.done) { resolve(value); } else { return Promise.resolve(value).then(function (value) { step("next", value); }, function (err) { step("throw", err); }); } } return step("next"); }); }; }

console.log('typeof', typeof foo);
console.log('toString', Object.prototype.toString.call(foo));
```
and outputs
```
typeof function
toString [object Function]
```
not `[object AsyncFunction]` as v8 does for:
```
Object.prototype.toString.call(async function(){ })
```

GingerPlusPlus:
Another, probably more future-proof possible implementation:

```js
/^\[object \w*Function\]$/.test(Object.prototype.toString.call(x))
```

CrossEye:
Yes, until TC39 in its infinite wisdom decides to add some type where `Object.prototype.toString` returns `"[object NotAFunction]"`.  :smile:

rjhilgefort:
Anything needed from me to push this along?

GingerPlusPlus:
`AsyncGeneratorFunction`?

rjhilgefort:
Sounded like your PR could happen independent of my change. Is that what's holding this PR up?

GingerPlusPlus:
Nope. My PR conflicts with yours.

What's holding this PR? No idea. @CrossEye has a tendency to not merge for no reason.

CrossEye:
> What's holding this PR? No idea. @CrossEye has a tendency to not merge for no reason.

Yes, not only is my free time limited, but my interest waxes and wanes.  I've never felt myself to be the one person to merge or analyze PRs.  I think it might be time we update the members of our core team, as only a few of them are not actively involved.

As to this request, I see no reason not to merge it, except for a minor issue I'll put in a review.

source/invoker.js
```diff
@@ -27,6 +27,16 @@ import toString from './toString';
  *      sliceFrom(6, 'abcdefghijklm'); //=> 'ghijklm'
  *      const sliceFrom6 = R.invoker(2, 'slice')(6);
  *      sliceFrom6(8, 'abcdefghijklm'); //=> 'gh'
+ *
+ *      const dog = {
+ *        speak: async () => 'Woof!'
+ *      };
+ *      const person = {
+ *        speak: async () => 'Hi!'
+ *      };
```
CrossEye: Is there a strong reason for giving two separate examples of the same behavior here?  I think either `dog` or `person` would be fine.

source/invoker.js
```diff
@@ -27,6 +27,16 @@ import toString from './toString';
...
+ *      const speak = R.invoker(0, 'speak');
+ *      speak(dog).then(R.identity) //=> 'Woof!'
+ *      speak(person).then(R.identity) //=> 'Hi!'
```
CrossEye: I like to reserve `//=>` for `returns`.  I think perhaps `//~>` would work here to imply that it logs that result.

source/invoker.js
rjhilgefort: The reasoning was to show that `invoker` can be used on either piece of data. I agree it's not necessary, will remove.

source/invoker.js
rjhilgefort: Fair enough will update!

GingerPlusPlus:
@CrossEye have you seen #2745?

willheslam:
@GingerPlusPlus 
I think taking AsyncGenerator into account is a great idea, however as @CrossEye pointed out here:
https://github.com/ramda/ramda/pull/2513#issuecomment-447983093
`/^\[object \w*?Function\]$/` is potentially dangerous.
It may seem unlikely, but IMO it has similar dangers of mootools' flatten
(https://github.com/mootools/mootools-core/issues/2797 #smooshgate)

I personally feel it's better to be explicit and enumerate types one by one - that way you are future proofing your codebase against unexpected browser features added later.
e.g. if in the future TC39 add `"[object SpecialCaseFunction]"` and I want to use ramda with it, we could submit a PR then I could upgrade to the latest version of ramda without risking a potential
`"[object NotAFunction]"` scenario.
It's less immediately convenient than a generalised regexp, but much safer in the long run.

Perhaps you could adjust your PR to be a pure additive to this one, adding 'AsyncGenerator' as a new case for `_isFunction`, in the same way I have with https://github.com/ramda/ramda/pull/2675?

GingerPlusPlus:
I'd argue otherwise. It's not unlikely that new kinds of functions will be added, but I cannot imagine any scenario in which Ramda wouldn't want to treat them as functions. Sure, they may need to be special-cased in some places, but would probably be treated like a standard function everywhere else.

And thus a regex is more future-proof, not less.

But that's just my opinion, I can't predict the future.
I am not complaining that this PR doesn't regex, I am complaining it doesn't handle `AsyncGeneratorFunction`.

> Perhaps you could adjust your PR to be a pure additive to this one

Only if this PR gets merged. My PR was meant as an alternative to this and yours.

source/invoker.js
GingerPlusPlus: `.then(R.identity)` does literally nothing. I'd personally use `.then(console.log)` in the examples.

source/internal/_isFunction.js
```diff
@@ -1,3 +1,6 @@
 export default function _isFunction(x) {
-  return Object.prototype.toString.call(x) === '[object Function]';
+  var type = Object.prototype.toString.call(x);
+  return type  === '[object Function]' ||
+    type === '[object AsyncFunction]' ||
+    type === '[object GeneratorFunction]';
```
GingerPlusPlus: Does not support `AsyncGeneratorFunction`.

source/invoker.js
rjhilgefort: Having console logs in examples is not very common in Ramda and that’s why I went with identity. If you look above to the slice example, the value returned is what’s in the comment- same applies here.

rjhilgefort: Having console logs in examples is not very common in Ramda and that’s why I went with identity. If you look above to the slice example, the value returned is what’s in the comment- same applies here.

source/internal/_isFunction.js
rjhilgefort: I don’t mind adding that if you’re certain that’s the exact string of the extra case I need to add.

rjhilgefort: I don’t mind adding that if you’re certain that’s the exact string of the extra case I need to add.

source/invoker.js
rjhilgefort: It actually doesn’t log that result and in fact would be the evaluates value. Looking at this again, I think its consistent with the above example and would argue it should remain as is. I don’t feel strongly though.

source/internal/_isFunction.js
GingerPlusPlus: `"[object AsyncGeneratorFunction]"`

rjhilgefort: Will update! @CrossEye, please speak up if this change would prevent having this merged.

source/invoker.js
GingerPlusPlus: Hm, how about
```js
speak(dog) //=> Promise.resolve('Woof!')
speak(person) //=> Promise.resolve('Hi!')
```

rjhilgefort: IMO that’s not as clear as just doing the dance of “unwrapping” with identity. Perhaps I could open up a closure and put the comment in there?

GingerPlusPlus: Invoker should obviously work with async functions,
perhaps the additional examples aren't necessary?

CrossEye: I have no problem with `console.log` in an example.  It is rare in Ramda because our pure functions should return something that w)

source/invoker.js
rjhilgefort: updated

source/invoker.js
rjhilgefort: Updated to use `console.log` and `~>`.

source/internal/_isFunction.js
rjhilgefort: updated

rjhilgefort:
@CrossEye I believe I've addressed the recent feedback. Ready for a final look (I hope).

PR #2514, opened 2018-04-23
Update match's type signature

brookemitchell:
From the example it looks like match always returns a (possibly empty) array of strings - no undefined values. 

Have updated the signature to reflect this after I saw a flowtype error.

### DIFF
--- a/source/match.js
+++ b/source/match.js
@@ -11,7 +11,7 @@ import _curry2 from './internal/_curry2';
  * @memberOf R
  * @since v0.1.0
  * @category String
- * @sig RegExp -> String -> [String | Undefined]
+ * @sig RegExp -> String -> [String]
  * @param {RegExp} rx A regular expression.
  * @param {String} str The string to match against
  * @return {Array} The list of matches or empty array.


CrossEye:
It makes sense to me.  Can anyone think of a case where an element of the return array could be `undefined`?

GingerPlusPlus:
Yes, optional capturing group not present in string:
```js
> /^a(b)?$/.exec('a')
[ 'a', undefined, index: 0, input: 'a', groups: undefined ]
> 'a'.match(/^a(b)?$/)
[ 'a', undefined, index: 0, input: 'a', groups: undefined ]
```

CrossEye:
Of course.  Should we then accept this PR rather than close it?

GingerPlusPlus:
Just realized that this function, like `String.prototype.match` that it uses, has 2 distinct behaviors:
- if regex has `g` flag, it'll return array of match strings (which can't contain `undefined`),
- if it does not have `g`, it'll return a single match object (which can contain `undefined`).

2nd behavior is IMO not documented enough, and if this was merged, it'd become totally undocumented.

CrossEye:
I never realized that about `String.prototype.match`.  Interesting!

PR #2515, opened 2018-04-23
Added pipeWith/composeWith functions

skallet:
Implemented pipeWith and composeWith function for better-controlled threadings. See #2512.


skallet:
@CrossEye are there any other test cases you would add?

source/internal/_pipeWith.js
```diff
@@ -0,0 +1,27 @@
...
+      arguments[0].length,
+      function() {
+        var inputArgs = arguments;
+
+        return reduce(
+          function(result, f) {
+            return transformFn.call(this, f, result);
+          },
+          pipeArgs[0].apply(this, inputArgs),
+          tail(pipeArgs)
```
polytypic: This is a known [optimization killer](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#32-leaking-arguments).  The `arguments` object is "leaked" (it [escapes](https://en.wikipedia.org/wiki/Escape_analysis)) and slower code is generated.

polytypic: Unfortunately no. It is necessary to copy the arguments in a loop.

skallet: I rewrite it to copy

CrossEye:
I don't know if we want to re-implement `pipeK`/`composeK` with this, or if we should simply deprecate them.  I know we want to deprecate `pipeP`/`composeP`, in any case.  I will work on adding `then` tomorrow and then we can point to `composeWith(then)` for those who want this behavior.  I think we should do the same with `pipeK`/`composeK`, pointing to `composeWith(chain)`, but I'd like to hear from others what they think.

skallet:
Ok, let me know if you need some help or I should rewrite code in some way.

source/internal/_pipeWith.js
skallet: Due to _arity fn there is always maximum of 10 arguments, so I think the current way should be fine.

source/composeWith.js
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Performs right-to-left function composition using transforming function. The rightmost function may have
+ * any arity; the remaining functions must be unary.
+ *
+ * **Note:** The result of compose is not automatically curried.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig ((y -> z), (x -> y), ..., (o -> p), ((a, b, ..., n) -> o)) -> ((a, b, ..., n) -> z)
```
Bradcomp: This signature looks like it might have been copied over from `compose`. I am somewhat at a loss for what the signature for this function _should_ be though...

source/pipeWith.js
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Performs left-to-right function composition using transforming function. The leftmost function may have
+ * any arity; the remaining functions must be unary.
+ *
+ * **Note:** The result of pipeWith is not automatically curried.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig (((a, b, ..., n) -> o), (o -> p), ..., (x -> y), (y -> z)) -> ((a, b, ..., n) -> z)
```
Bradcomp: This signature also must be updated

Bradcomp:
I agree we should deprecate `[(pipe)|(compose)][k|p]` with the introduction of these functions.

I have one concern about the way this is implemented, and I'd like to open up the conversation.

The way it is set up now, it's a unary function that returns a variadic function that returns a variadic function.

For me, I would prefer a curried binary function that takes the transformer and an array containing the composition, and returns the uncurried variadic composed function.

As for the signature, I am not 100% on what the signature of the transforming function should be.

skallet:
@CrossEye I would not make both approaches, I think better is to stick to one general approach. I prefer passing list as well, but due to current compose function, I think that is the good argument to stick to it. Also, I do think that change of current compose functionality to the passing list is just unnecessary BC.

buzzdecafe:
@CrossEye 

> If we were starting from scratch, our main compose function would not be ...
>
> `compose(baz, bar, foo)`
>
> but rather 

`compose(f)(g)`

FTFY :wink:

source/composeWith.js
skallet: @Bradcomp can you help me with this?

Bradcomp:
Re: signature

I can help, but I currently don't have an answer. 

`chain` has the signature of `Monad m:: (a -> m b) -> m a -> m b`

`then` has the signature of `(a -> Promise b) -> Promise a -> Promise b`

`applyIfExists` has a signature like ` (a -> b) -> a? -> b?`

if we apply `identity` as the transformer we get `(a -> b) -> a -> b`

All of those have the same general shape, but there might be other examples that follow a subtly different shape. For instance, `pipeWith((f, res) => f(res + 1))` doesn't really fall in line, but I think we could design a monad that captured the side effect of incrementing and applying, so long as we restrict the domain to numbers. 

In any case the examples we have so far assume there is some way to treat the results of the composed functions as `Monads` while giving the user the ability to define exactly what the treatment of these 'pseudo-monads' should be.

That gives us something like `Something x:: (a -> x b) -> [(m -> x n), (l -> x m), ... ((a, ...) -> x b)] -> (a, ...) -> x n` and we need to solve for x.

That's as far as I have so far.

CrossEye:
@Bradcomp:

I had gone down the same path, and ended up talking about a lawless `Wrapper` type.  I'd like something better than that, but we may not be able to find anything.  I think we could treat `applyIfExists` as `(a -> Nullable b) -> Nullable a -> Nullable b` or some such.  What I don't think we can do is assume we actually have Monads and could use `chain`.  That's unfortunate, but I think we're stuck with that.

skallet:
So what is the next step?

CrossEye:
@skallet: I would really like to see two changes.

One: remove the de-optimizing `tail` call as @polytypic suggested.

Two: replace the variadic call passing the functions with one that takes a list of them.  We'd like to move Ramda in the direction of *removing* variadic function, not adding them.  (I don't mean anything here about supporting users' variadic function; that's a very different discussion.  But I would like Ramda to export only fixed-arity functions.)  This of course would also allow the main functions to be curried, which is a minor win.  I know this makes them less like `compose` and `pipe`, but then they are more like `just`, `converge`, and `useWith`, and aligned with the direction we'd like to see the library move.

skallet:
I'll try to complete it this week. I have been little busy lately.

CrossEye:
> I'll try to complete it this week

Thank you.


> I have been little busy lately.

Understood.  No rush.

Bradcomp:
This looks really good. I have a couple of comments, but nothing that would hold up moving forward with this.

1. It would be nice to see at least one more example other than the notNil one. Maybe reproducing `pipeP`.
2. Would it be prudent to deprecate `[(pipe)|(compose)][PK]` at the same time?
3. I can see at least a couple property based tests to use here. One would be that `pipeWith((f, a) => f(a))` === `pipe` for all functions and inputs.

skallet:
@Bradcomp I agree with 1 and 3, so I try to figure out more examples/tests. 
> Would it be prudent to deprecate [(pipe)|(compose)][PK] at the same time?

There is currently no support for promises in pipeWith/composeWith, should I add it?

Bradcomp:
@skallet The beauty is that we don't _need_ Promise support due to the nature of `pipeWith`. 

```
const then = (f, p) => p.then(f);

const pipeP = pipeWith(then);
```

With that, we have reproduced `pipeP`

CrossEye:
> There is currently no support for promises in pipeWith/composeWith, should I add it?

No, I dropped the ball, but will push a `then` function soon.  Together with this users can replace `pipeP` and `composeP` and together with `chain` they can replace `pipeK/composeK`.

CrossEye:
Also note that deprecation is handled in one version by adding a JSDoc tag, for instance:

https://github.com/ramda/ramda/commit/d3766c873dceb57cbf5d8244681e6270d93bbcb2#diff-ea5373a922fd05320c39d0d733bfd3caR20

skallet:
@Bradcomp you were right! I did not realize it can work like this. I've added promise example into tests.

skallet:
@CrossEye is there anything I should add?

CrossEye:
No, you're fine.  There's another change I want to commit around the same time.  I will try to work on that tomorrow.

skallet:
@CrossEye any update?

source/pipeWith.js
```diff
@@ -0,0 +1,43 @@
...
+ *
+ *      const pipeWhileNotNil = R.pipeWith((f, res) => R.isNil(res) ? res : f(res));
+ *      const f = pipeWhileNotNil([Math.pow, R.negate, R.inc])
+ *
+ *      f(3, 4); // -(3^4) + 1
+ * @symb R.pipeWith(f)([g, h, i])(...args) = f(i, f(h, f(g, ...args)))
+ */
+var pipeWith = _curry2(function pipeWith(xf, list) {
+  if (list.length <= 0) {
+    throw new Error('composition must contains at least one function');
```
GingerPlusPlus: `return identity`?

source/pipeWith.js
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  return function() {
+    return reduce(
```
GingerPlusPlus: Could use internal, non-curried `_reduce` here.

source/pipeWith.js
```diff
@@ -0,0 +1,43 @@
...
+ *
+ *      f(3, 4); // -(3^4) + 1
+ * @symb R.pipeWith(f)([g, h, i])(...args) = f(i, f(h, f(g, ...args)))
+ */
+var pipeWith = _curry2(function pipeWith(xf, list) {
+  if (list.length <= 0) {
+    throw new Error('composition must contains at least one function');
+  }
+
+  return function() {
```
GingerPlusPlus: Wrap the returned function in `_arity(head(list).length)`.

source/pipeWith.js
```diff
@@ -0,0 +1,43 @@
...
+    return reduce(
+      function(result, f) {
+        return xf.call(this, f, result);
+      },
+      head(list).apply(this, arguments),
+      tail(list)
```
GingerPlusPlus: Move the `tail(list)` call outside of the returned function, so that it only runs once, when creating piped function, instead of on every call.

test/composeWith.js
```diff
@@ -0,0 +1,47 @@
...
+      return f(res);
+    })([R.map, R.multiply, parseInt]);
+
+    eq(f.length, 0);
+    eq(f('10')([1, 2, 3]), [10, 20, 30]);
+    eq(f('10', 2)([1, 2, 3]), [2, 4, 6]);
+  });
+
+  it('performs right-to-left function while not nil result', function() {
+    var isOdd = R.modulo(R.__, 2);
```
GingerPlusPlus: Please use `R.flip(R.modulo)(2)`; I'm right now working on removing the placeholder.

GingerPlusPlus:
@CrossEye: no need to wait for #2565, users can easily implement `then` and `fail` / `otherwise` with `invoker`.

skallet:
@GingerPlusPlus I've updated code. Please check it.

Bradcomp:
This looks good to me, can you squash all the commits?

skallet:
@Bradcomp how do I do that? I thought that you can do merge squash.

skallet:
See #2630.

source/pipeWith.js
GingerPlusPlus: This was just a suggestion (question mark), perhaps I didn't make it clear enough.

Now there is an inconsistency: when passed no functions, `R.compose`, `R.pipe` and Scheme's `compose` throw, while `R.composeWith`, `R.pipeWith`, and Clojure's `comp` return identity function.

PR #2516, opened 2018-04-23
Remove unnecessary currying from pluck example

Kiran-Rao:
I've seen several new developers on my project use currying unnecessarily with `R.pluck` due to the documentation.

### DIFF
--- a/source/pluck.js
+++ b/source/pluck.js
@@ -22,8 +22,8 @@ import prop from './prop';
  * @see R.props
  * @example
  *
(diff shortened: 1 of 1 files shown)


CrossEye:
I have no problem with this change.  But I wonder if we should include one example with currying, perhaps something like

    const getAges = pluck('age');
    getAges([{name: 'fred', age:: 29}, {name: 'wilma', age: 27}]) //=> [29, 27]

as this is the way I think of Ramda as most often being used.

CrossEye:
Thank you very much!

Issue #2519, opened 2018-04-29
tryCatch documentation error


edjroz:
Straight from the example of the tryCatch method:
```
const R = require('ramda');
console.log(R.tryCatch(R.prop('x'), R.F)(null));
```
Should return false, however it's returning undefined.

Passing a function and trying to console log anything from within doesn't seem to work either

```
console.log(R.tryCatch(R.prop('x'), () => {
  console.log(1)
  return false;
})(null));
```

edjroz:
This issue actually doesn't happen because of the `tryCatch()` but actually because of `R.prop()` not throwing an error in regards the value receiving.

```Javascript
console.log(R.tryCatch(R.concat('a', R.__), R.F)('a'));  //=> false
```

CrossEye:
Yes.  A PR for this would be most welcome.

PR #2520, opened 2018-04-29
fix documentation on tryCatch method

edjroz:
In regards #2519. 

However I can see that the it was fixed on the source with another example 
`R.tryCatch(JSON.parse, R.F)({});      //=> false`

 But hasn't been built to reflect it on the documentation

### DIFF
--- a/source/tryCatch.js
+++ b/source/tryCatch.js
@@ -2,7 +2,6 @@ import _arity from './internal/_arity';
 import _concat from './internal/_concat';
 import _curry2 from './internal/_curry2';
 
-
 /**
  * `tryCatch` takes two functions, a `tryer` and a `catcher`. The returned
  * function evaluates the `tryer`; if it does not throw, it simply returns the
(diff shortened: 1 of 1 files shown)


source/tryCatch.js
```diff
@@ -22,8 +21,9 @@ import _curry2 from './internal/_curry2';
  * @example
  *
  *      R.tryCatch(R.prop('x'), R.F)({x: true}); //=> true
- *      R.tryCatch(JSON.parse, R.F)({});      //=> false
- */
+ *      R.tryCatch(R.test(/^a/, R.__), R.F)('b'); // => false
+ *      R.tryCatch(R.times(R.identity), R.always([]))('s') // => []
```
CrossEye: I think perhaps we should add a much more explicit example:

   R.tryCatch(() => {throw 'oops'}, () => 'foo')('bar')  //=> 'foo'

edjroz:
Changed the examples (I squashed the commits to avoid pollution) 

for a simple scenario:
```Javascript
R.tryCatch(() => { throw 'foo'}, R.always('catched'))('bar') // => 'catched'
```
and a little more advanced one:
```Javascript
R.tryCatch(R.times(R.identity), R.always([]))('s') // => []
```

CrossEye:
Thank you!

JulianoGTZ:
Is this issue still open?

CrossEye:
This was merged some time ago; it is closed.

Issue #2527, opened 2018-05-14
Ramda website served on non secure website.


LoganElliott:
http://ramdajs.com points to a non-secure site and when you navigate to https://ramdajs.com it shows up with an error.
![httpsramdasite](https://user-images.githubusercontent.com/8405563/39977118-c97b94ae-578b-11e8-92c7-ed8da1636a60.png)

Do you have any plans to move the Ramda site to be served on https?

CrossEye:
Thanks for bringing this up.  I saw the recent announcement, but forgot to follow up.

I started the process.  We're waiting on this step:

> Not yet available for your site because the certificate has not finished being issued 

I'm not quite sure how long that takes, but then I'll choose the "Require SSL" option too.

tbsvttr:
This should go rather fast, but it does not work now. In most cases that means something does not work correctly. Maybe retry or remove and add the custom domain in the process.

CrossEye:
I did try that earlier, and again just now.  I will go through the instructions again to make sure I didn't miss a step.  If that doesn't work, I will see if I can find some GH support.

functionalStoic:
I recently tried it out on a site. Got nowhere after trying lots of changes. Based on googling I found a quick email to GH support will `nudge` it along (That's the word they use in their default response).

<img width="888" alt="inbox_ _jason_sooter_gmail_com" src="https://user-images.githubusercontent.com/7215306/40131931-f5b68ecc-5900-11e8-826a-d5b62ff8f8d4.png">

CrossEye:
Thanks.  It's still not there, so I've sent a support request.

CrossEye:
I never heard back on my support request, but it's working now.  The main docs page still has a small problem, in that we serve the logo only from its artist's insecure page.

@buzzdecafe: could you check with JCPhillips to see whether it's ok to self-host this, or if there are plans in the works to move her site to SSL?

buzzdecafe:
no problem at all serving that img statically from ramdajs.com

CrossEye:
Ok, it's up now.  Thanks to @LoganElliott for bringing this up and to @functionalStoic for the help.

PR #2541, opened 2018-05-24
Change http references to https

LoganElliott:
Fix a couple of http references since #2527 has been solved.

The reference on the Ramda GitHub homepage also needs to be changed by a maintainer.


CrossEye:
Thank you.

Issue #2543, opened 2018-05-25
`dissocPath` converts object to Array when number in path


vclav:
In `dissocPath` targeting an *Object* which have numeric keys, with path having segment for the object `typeof === 'number'` causes the *Object* being converted to an empty *Array*.

Here is proof from debugger ran in _Node 9.3.0_ where input is `definition`, output is `definitionWithoutTile` and path is `['content', 3]`. You can see, that in output is na empty *Array*.

https://www.evernote.com/l/AkMyGtHNf19EZJlDZw1U8Cq5e_yfDNXb08k 

And here is the same code, where number `3` in path is converted to `String` and output is dissociated *Object* as expected.

https://www.evernote.com/l/AkMMFnPjf3RAGoKACpSZpDzOFlmTNGuhEC4

CrossEye:
It's possible that there could be a fix for this for `dissocPath`.  I'm not quite sure what that would be, but a PR would be welcome.

However, for `assoc`, this is clearly working as expected.

Given 

```js
const tree = {2: {v: 1}, a: [{b: 'hey'}]}
```

You point out the difference between these:

```js
assocPath(['a', 1], 'hi', tree)
//=> {2: {v: 1}, a: [{b: 'hey'}, 'hi']}

assocPath(['a', '1'], 'hi', tree)
//=> {2: {v: 1}, a: {0: {b: 'hey'}, 1: 'hi'}}
```

That is by design.  When originally written, `assoc` and `assocPath` worked only with objects, and had no idea about arrays.  We changed it so that when passing integers as keys, they would act as array indices.  This is a much more useful  API, allowing us to work on exactly the kinds of mixed data presented in `tree`.

That allows for these sorts of behavior:

```js
assocPath(['a', 1], {b: 'howdy'}, tree)
//=> {2: {v: 1}, a: [{b: 'hey'}, {b: 'howdy'}]}

assocPath(['c', 0], 'howdy', tree)
//=> {2: {v: 1}, a: [{b: 'hey'}], c: ['howdy']}

assocPath(['a', 1], 'aloha', {a: ['hey', 'howdy', 'hi'], 2: {v: 1}})
//=> {2: {v: 1}, a: ['hey', 'aloha', 'hi']}
```

[...]

PR #2545, opened 2018-05-26
fix dissocPath for non-string types in path #2543

MilanLempera:
This change unifies behavior of `dissocPath` and `dissoc`. I only add tests and behavior by  https://github.com/ramda/ramda/blob/master/test/dissoc.js#L26

### DIFF
--- a/test/dissocPath.js
+++ b/test/dissocPath.js
@@ -60,4 +60,10 @@ describe('dissocPath', function() {
     eq(R.dissocPath([], {a: 1, b: 2}), {a: 1, b: 2});
   });
 
+  it('coerces non-string types', function() {
+    eq(R.dissocPath([42], {a: 1, b: 2, 42: 3}), {a: 1, b: 2});
+    eq(R.dissocPath([null], {a: 1, b: 2, 'null': 3}), {a: 1, b: 2});
+    eq(R.dissocPath([undefined], {a: 1, b: 2, undefined: 3}), {a: 1, b: 2});
+  });
+
 });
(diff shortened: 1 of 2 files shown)


test/dissocPath.js
```diff
@@ -60,4 +60,10 @@ describe('dissocPath', function() {
     eq(R.dissocPath([], {a: 1, b: 2}), {a: 1, b: 2});
   });
 
+  it('coerces non-string types', function() {
+    eq(R.dissocPath([42], {a: 1, b: 2, 42: 3}), {a: 1, b: 2});
+    eq(R.dissocPath([null], {a: 1, b: 2, 'null': 3}), {a: 1, b: 2});
+    eq(R.dissocPath([undefined], {a: 1, b: 2, undefined: 3}), {a: 1, b: 2});
```
CrossEye: I don't think we want to commit to this behavior on `null` and `undefined`.  We might one day want to insist that the user supplies an array of index types (strings, numbers, perhaps Symbols), so I would rather leave this behavior unspecified for the moment.

Can we remove those last two assertions?

CrossEye:
This looks good, except that I think two of the tests are committing us to behavior better left unspecified for the moment.

MilanLempera:
@CrossEye I have worked out the required changes, but I don't know why build job failed on node v0.12.18. 
I think fail is not caused by my change. Is there any way to run it again?

CrossEye:
Sorry, I missed the previous message.  This looks good to me.  Anyone object?  @ramda/core?

CrossEye:
Sorry, I have been little-focused on Ramda lately.  I will try to get back to it, release a version and work on the 1.0 and 2.0 roadmap.

colinkahn:
@CrossEye no need to apologize! I'd be glad to help in some way if that's possible.

PR #2551, opened 2018-06-02
Add thunkify function

nfantone:
This PR adds a new, simple `thunkify` function. It takes an `fn` function and returns a new function that asks for arguments before returning a third function that, when called, invokes `fn` with the provided arguments.

```js
const thunkify = fn => (...args) => () => fn(...args);
```

Original PR idea came from @CrossEye at a [Stackoverflow answer](https://stackoverflow.com/questions/50495874/how-to-create-a-function-that-delays-call-to-given-function?noredirect=1#comment88255490_50495874).

### Motivation
- While very simple, wrapping functions to delay their invocations is a really common pattern.
- Ramda itself does not provide any out-of-the-box tool to help with this in a straight forward manner.
- It is specially useful when arguments to the _thunkified_ function are **not known until a later execution stage**.
- It helps with creating composable functions that get arguments from composed chains.
- The `fn` function could be variadic and can be called by the thunk with any number of arguments.

### Example

This has been adapted from a real world application I'm currently working on.

```js
function throwErrorWith(message, data) {
  throw Object.assign(new Error(message), data);
}

const thenThrowErrorWith = R.thunkify(throwErrorWith);

// ...later

const throwIfBadFormat = unless(
  validFormat,
  thenThrowErrorWith(
    'Invalid response format: expected one of <result></result> or <error></error>'
  )
);
```

---

Some considerations:

- Not very keen on the `thunkify` name, to be honest.
- This is my first PR on the project. My knowledge of the internal functions is rather limited, so I'm far from sure if this is the current "Ramda" approach to writing the solution. 
- There might be better ways and/or other internals I should be using. Looking forward to your input!
- I'm not very familiar with Haskell-like signatures so wasn't quite sure how to describe `thunkify`. Help appreciated.

### DIFF
--- a/source/thunkify.js
+++ b/source/thunkify.js
@@ -1,0 +1,30 @@
+import _curry1 from './internal/_curry1';
+import _arity from './internal/_arity';
+
+/**
+ * Creates a thunk out of a function. A thunk delays a calculation until
+ * its result is needed, providing lazy evaluation of arguments.
+ *
+ * @func
+ * @memberOf R
+ * @since v0.25.0
+ * @category Function
+ * @sig (a -> b) -> (a -> b) // FIXME: write appropiate signature
+ * @param {Function} fn A function to wrap in a thunk
+ * @return {Function} Expects arguments for `fn` and returns a new function
+ *  that, when called, applies those arguments to `fn`.
+ * @see R.partial, R.partialRight
+ * @example
+ *
+ *      R.thunkify(R.identity)(42)(); //=> 42
+ */
+var thunkify = _curry1(function thunkify(fn) {
+  return _arity(fn.length, function createThunk() {
+    const fnArgs = arguments;
+    return function invokeThunk() {
+      return fn.apply(this, fnArgs);
+    };
+  });
+});
+
+export default thunkify;
--- a/source/thunkify.js
+++ b/source/thunkify.js
@@ -1,4 +1,4 @@
-import _arity from './internal/_arity';
+import _curryN from './internal/_curryN';
 import _curry1 from './internal/_curry1';
 
 /**
@@ -19,7 +19,7 @@ import _curry1 from './internal/_curry1';
  *      R.thunkify((a, b) => a + b)(25, 17)(); //=> 42
  */
 var thunkify = _curry1(function thunkify(fn) {
-  return _arity(fn.length, function createThunk() {
+  return _curryN(fn.length, [], function createThunk() {
     const fnArgs = arguments;
     return function invokeThunk() {
       return fn.apply(this, fnArgs);
--- a/source/thunkify.js
+++ b/source/thunkify.js
@@ -7,7 +7,6 @@ import _arity from './internal/_arity';
  *
  * @func
  * @memberOf R
- * @since v0.25.0
  * @category Function
  * @sig (a -> b) -> (a -> b) // FIXME: write appropiate signature
  * @param {Function} fn A function to wrap in a thunk
--- a/source/thunkify.js
+++ b/source/thunkify.js
@@ -8,7 +8,7 @@ import _arity from './internal/_arity';
  * @func
  * @memberOf R
  * @category Function
- * @sig (a -> b) -> (a -> b) // FIXME: write appropiate signature
+ * @sig ((a, b, ..., j) -> k) -> (a, b, ..., j) -> (() -> k)
  * @param {Function} fn A function to wrap in a thunk
  * @return {Function} Expects arguments for `fn` and returns a new function
  *  that, when called, applies those arguments to `fn`.
--- a/source/thunkify.js
+++ b/source/thunkify.js
@@ -1,4 +1,4 @@
-import _curryN from './internal/_curryN';
+import curryN from './curryN';
 import _curry1 from './internal/_curry1';
 
 /**
@@ -19,7 +19,7 @@ import _curry1 from './internal/_curry1';
  *      R.thunkify((a, b) => a + b)(25, 17)(); //=> 42
  */
 var thunkify = _curry1(function thunkify(fn) {
-  return _curryN(fn.length, [], function createThunk() {
+  return curryN(fn.length, function createThunk() {
     const fnArgs = arguments;
     return function invokeThunk() {
       return fn.apply(this, fnArgs);
--- a/source/thunkify.js
+++ b/source/thunkify.js
@@ -1,5 +1,5 @@
-import _curry1 from './internal/_curry1';
 import _arity from './internal/_arity';
+import _curry1 from './internal/_curry1';
 
 /**
  * Creates a thunk out of a function. A thunk delays a calculation until
@@ -16,6 +16,7 @@ import _arity from './internal/_arity';
  * @example
  *
  *      R.thunkify(R.identity)(42)(); //=> 42
+ *      R.thunkify((a, b) => a + b)(25, 17)(); //=> 42
  */
 var thunkify = _curry1(function thunkify(fn) {
   return _arity(fn.length, function createThunk() {
--- a/test/thunkify.js
+++ b/test/thunkify.js
@@ -1,0 +1,22 @@
+var R = require('..');
+var eq = require('./shared/eq');
+
+describe('thunkify', function() {
+  it('returns a function with the same arity as the given function', function() {
+    var input = function input(a0, a1) { };
+    var f = R.thunkify(input);
+    eq(typeof f, 'function');
+    eq(f.length, input.length);
+  });
+
+  it('returns a function that expects arguments and returns a new invoker function', function() {
+    var input = function input(a0, a1) { };
+    var f = R.thunkify(input);
+    eq(typeof f(42, 'xyz'), 'function');
+  });
+
+  it('calls the original function with the provided arguments', function() {
+    var f = R.thunkify(R.add(2));
+    eq(f(42)(), 44);
+  });
+});
--- a/test/thunkify.js
+++ b/test/thunkify.js
@@ -19,4 +19,10 @@ describe('thunkify', function() {
     var f = R.thunkify(R.add(2));
     eq(f(42)(), 44);
   });
+
+  it('should support sending partial arguments', function() {
+    var input = function input(a0, a1) { return [a0, a1]; };
+    var f = R.thunkify(input);
+    eq(f(42)(), [42, undefined]);
+  });
 });
--- a/test/thunkify.js
+++ b/test/thunkify.js
@@ -2,6 +2,13 @@ var R = require('..');
 var eq = require('./shared/eq');
 
 describe('thunkify', function() {
+  it('returns a function with the same arity as the given function', function() {
+    var input = function input(a0, a1) { };
+    var thunk = R.thunkify(input);
+    eq(typeof thunk, 'function');
+    eq(thunk.length, input.length);
+  });
+
   it('returns a function that expects arguments and returns a new invoker function', function() {
     var input = function input(a0, a1) { };
     var thunk = R.thunkify(input);
--- a/test/thunkify.js
+++ b/test/thunkify.js
@@ -2,27 +2,14 @@ var R = require('..');
 var eq = require('./shared/eq');
 
 describe('thunkify', function() {
-  it('returns a function with the same arity as the given function', function() {
-    var input = function input(a0, a1) { };
-    var f = R.thunkify(input);
-    eq(typeof f, 'function');
-    eq(f.length, input.length);
-  });
-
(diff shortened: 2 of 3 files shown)


source/thunkify.js
```diff
@@ -0,0 +1,29 @@
...
+import _arity from './internal/_arity';
+
+/**
+ * Creates a thunk out of a function. A thunk delays a calculation until
+ * its result is needed, providing lazy evaluation of arguments.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig (a -> b) -> (a -> b) // FIXME: write appropiate signature
```
Bradcomp: I am opening the discussion on a type sig here. Something like 

`((a, b, ..., j) -> k) -> (a, b, ..., j) -> (() -> k)`

would probably do the trick. I haven't gone back to look at the standard way to notate variadic functions though...

source/thunkify.js
```diff
@@ -0,0 +1,29 @@
...
+ * @param {Function} fn A function to wrap in a thunk
+ * @return {Function} Expects arguments for `fn` and returns a new function
+ *  that, when called, applies those arguments to `fn`.
+ * @see R.partial, R.partialRight
+ * @example
+ *
+ *      R.thunkify(R.identity)(42)(); //=> 42
```
Bradcomp: I'd like to see at least one additional example here that uses more than one parameter.

test/thunkify.js
```diff
@@ -0,0 +1,28 @@
...
+  it('calls the original function with the provided arguments', function() {
+    var f = R.thunkify(R.add(2));
+    eq(f(42)(), 44);
+  });
+
+  it('should support sending partial arguments', function() {
+    var input = function input(a0, a1) { return [a0, a1]; };
+    var f = R.thunkify(input);
+    eq(f(42)(), [42, undefined]);
+  });
```
Bradcomp: We should add a property based test for this as well. 

`forall f, a, b :: f(a, b) == thunkify(f)(a, b)()`

test/thunkify.js
```diff
@@ -0,0 +1,28 @@
...
+    var f = R.thunkify(input);
+    eq(typeof f(42, 'xyz'), 'function');
+  });
+
+  it('calls the original function with the provided arguments', function() {
+    var f = R.thunkify(R.add(2));
+    eq(f(42)(), 44);
+  });
+
+  it('should support sending partial arguments', function() {
```
Bradcomp: I am not sure whether this would be the desired behavior, and it should be discussed. One could argue the function returned by thunkify should be curried, and not return a thunk until all the arguments have been accepted...

source/thunkify.js
nfantone: I went for something similar first time writing it, based on how `flip` [signature was written](https://github.com/ramda/ramda/blob/master/source/flip.js#L13). But desisted since I wasn't quite sure if `()` is a valid notation (isn't it just a scalar value?)

source/thunkify.js
nfantone: 👍. Anything in particular in mind?

test/thunkify.js
nfantone: I'm all for adding additional tests, but unsure what you mean by "property based"?

source/thunkify.js
buzzdecafe: This would be the "aspirational" signature: `(a -> b) -> a -> () -> b`

capturing the variadic-ness of `fn` makes it tricky to right a coherent sig.

`(a... -> b) -> a... -> () -> b`

?

buzzdecafe:
> Not very keen on the `thunkify` name, to be honest.

sounds fine to me, but `enthunk` would be ok as well.

source/thunkify.js
nfantone: @buzzdecafe Based on how previous variadic function's signatures were written, I'd say @Bradcomp is on the right track here (i.e.: `(a, b..., j)`).

So, merging both suggestions, you'll end up with his proposal.

test/thunkify.js
Bradcomp: Property tests are for verifying invariant properties of a function. There are certain things that should be true about a function regardless of the inputs. With unit tests, we are only testing certain examples, but property tests can throw a bunch of random input at the function and make sure those invariants hold.

[Here](https://hackernoon.com/property-based-testing-4330e3e77381) is a really simple overview with JS examples.

[Here](https://github.com/ramda/ramda/blob/master/test/compose.js#L63) is an example in our codebase of some property tests.

tommmyy:
I came up with pointfree version if you are interested:

```js
const getThunk = useWith(o(always), [apply]);
const enthunk = useWith(unapply, [getThunk]);
```

See derivation process: https://goo.gl/Y2oEme

nfantone:
@tommmyy Very nice. Good job! That's similar to what I previously came up with (and mentioned in [original SO question](https://stackoverflow.com/questions/50495874/how-to-create-a-function-that-delays-call-to-given-function?noredirect=1#comment88255490_50495874)). But yours actually handles variadic functions.

source/thunkify.js
CrossEye: I like @Bradcomp's suggestion here.  At some point, I want to switch all internally variadic signatures to an ellipses-terminated list of signatures, but Brad's version is consistent for now with our current variadic signatures.

source/thunkify.js
CrossEye: Even a simple `thunkify((a, b) => a + b)(25, 17)() ==> 42` would help.

test/thunkify.js
CrossEye: I do like the idea of adding property-based testing, but I wouldn't hold up a PR for that.  I would rather do a focused pass through the code to bring it about.

test/thunkify.js
nfantone: Not objecting to that, really. It could very well be appropriate, conceptually fitting way to do it.

I'd only like to point out that my original intention was to make the _thunkified_ function behave as close as the one you provide as possible. In that sense, these would be equivalent:

```js
const fn = (a, b) => { ... };

fn(a) === thunkify(fn)(a)();
```

Whereas if I understood you correctly and we decide on currying the result, that would no longer hold true.
 
The thunk only serves as a "delayer" of the invocation and nothing else: a placeholder for a future call.

source/thunkify.js
nfantone: Changing it to match @Bradcomp's sig, then.

test/thunkify.js
CrossEye: When did I supply a version?  I don't recall doing so.

Ramda generally pays strict attention to the arity of the functions it deals with.  We are loathe to apply only `(a)` to a function expecting `(a, b)`.

Maybe our best bet would be to have a function that supplies a thunk, and not one that creates thunk-generating functions, something like

```
const myThunk = thunkify(expensiveCalc, [a, b])
// ... later
const result = myThunk() //=> expensiveCalc(a, b)
```

Then if you supply fewer arguments than the function has parameters, it's clearly on you.

The more I think about it, the better I like this alternative.

nfantone: > When did I supply a version? I don't recall doing so.
I was referring to your `curry + 1` suggestion. _That_ version.

>  We are loathe to apply only (a) to a function expecting (a, b).
Fair enough. I can also see the benefits of the reasoning. Let's stick to that perspective.

> Maybe our best bet would be to have a function that supplies a thunk, and not one that creates thunk-generating functions
What I don't find attractive about going in that direction is the fact that you lose the ability to simply provide parameters at a later stage if you don't know them at thunk-creation time.

```js
// What if [a, b] are not known here?
const myThunk = thunkify(expensiveCalc, [a, b]);

// I'd like to see something closer to
const myThunk = thunkify(expensiveCalc)([a, b]);
```

Bradcomp: @nfantone because all Ramda functions are curried by default, you'd be able to do either one of the examples you gave.

nfantone: @Bradcomp Certainly. But that's why I wrote "closer to". The semantics of the first example are, to me at least, clear enough; while on the latter, they seem kinda obscure (is `expensiveCalc` expecting a single array or two arguments `a`, `b`?).

Just a thought 👇 

```js
// IMHO, this:
R.thunkify(R.identity)(42)();

// ...looks much cleaner and predictable than:
R.thunkify(R.identity)([42])(); // <- Why is `R.identity` not returning what I'm passing?
```

CrossEye: > ```js
> // What if [a, b] are not known here?
> const myThunk = thunkify(expensiveCalc, [a, b]);
> ```   

So then you're introducing three separate moments in time: one when you know the baseline function you want to thunkify, a second when you know the arguments you want to provide to it but have lost reference to the function itself, and a third when you want to actually apply the function to the arguments and might no longer know either of them.  Is that actually a realistic scenario?

I ask because `thunkify`, even with two distinct moments in time is already an oddity in Ramda, which tries to stick with a functional paradigm.  Functional programming is all about evaluating expressions, and generally you can leave it to the underlying engine to decide when to evaluate individual ones; you don't control sequencing directly.  The existence of [`Monads`][mo]s is in good part about allowing functional programming to actually control some of this sequencing.  In any case, having three separate moments involved seems overkill to me.  But I could be convinced by a good example.

My suggestion of `curry + 1` was off-the-cuff, and while I can see some sense in it, I'm not sure I would really want that API.  I think this demonstrates some issues with it:


```js
const foo1 = curry((a, b) => a + 2 * b)
foo1.length //=> 2
const bar1 = foo1(12)
bar1.length //=> 1
const baz1 = bar1(15)
baz1 //=> 42

const foo2 = thunkify((a, b) => a + 2 * b)
foo2.length //=> ???  2?  3?
const bar2 = foo2(12)
bar2.length //=> ???  1?  2?
const baz2 = bar2(15)
baz2.length //=> ???  0?  1?
baz2() //=> foo2(10, 15) => 42
```

  [mo]: https://github.com/fantasyland/fantasy-land#monad

nfantone: @CrossEye I understand your concern about multiple points in time - but believe  you've stretched the interpretation.

> So then you're introducing three separate moments in time

Actually, only two: _thunk creation time_ and _thunk invocation_. I'm struggling with finding a real world example that fits your three moments.

> Is that actually a realistic scenario?

Not that I can think of, no. But I'm sure there's _bound to be one_ if we think harder.

> My suggestion of curry + 1 was off-the-cuff, and while I can see some sense in it, I'm not sure I would really want that API. 

You raised good points with your arity oddities. The more I think about it, the more I keep coming back to `thunkify(expensiveCalc, [a, b]);` as the better approach.

nfantone: @CrossEye @Bradcomp I'm ready to push this forward and (re-)implement a solution. Are we all inclining to `thunkify(fn, [a, b])` sig here? Any objections/comments?

Bradcomp: I like this one, and I don't see an issue with the length like @CrossEye does:

```
const foo2 = thunkify((a, b) => a + 2 * b)
foo2.length //=> ???  2?  3?
const bar2 = foo2(12)
bar2.length //=> ???  1?  2?
const baz2 = bar2(15)
baz2.length //=> ???  0?  1?
baz2() //=> foo2(10, 15) => 42
```

The left numbers are correct. A nullary function is a perfectly valid return type, and has a length of zero. Even when taking `curry` into account, the length of a function says nothing of the return type.

The best case scenario in my mind is that `thunkify` does two things to the input function - curries the function, and changes the return type from `b` to `() -> b`.

CrossEye: OK, I think I'm convinced.  Describing the thunk as the output and not as one more curried step made me see the light.

At this point, the only thing I see left is that it makes sense to curry `createThunk` instead of simply fixing its arity, that is:

```js
var thunkify = _curry1(function thunkify(fn) {
  return curryN(fn.length, function createThunk() {
    const fnArgs = arguments;
    return function invokeThunk() {
      return fn.apply(this, fnArgs);
    };
  });
});
```

This allows either of these:

```js
thunkify(R.add)(42)(2)() //=> 44
thunkify(R.add)(42, 3)() //=> 45
```

which seems more in line with Ramda.

source/thunkify.js
```diff
@@ -0,0 +1,30 @@
...
+ *      R.thunkify((a, b) => a + b)(25, 17)(); //=> 42
+ */
+var thunkify = _curry1(function thunkify(fn) {
+  return _arity(fn.length, function createThunk() {
```
CrossEye: Can we switch this to `curryN` rather than `arity`?

nfantone: Will do.

test/thunkify.js
nfantone: Done.

nfantone: Done.

PR #2553, opened 2018-06-05
reference R.without in the R.remove docs

chriddyp:
Many, many thanks to everyone who has worked on Ramda 🙇 

I have a hard time remembering `R.without`. I always think "remove item from list" and end up searching for `R.remove`, which isn't really what I was looking for.

This PR adds a `"See also without"` to the `remove` section.


CrossEye:
This makes sense.  Could I ask that you do the converse as well, adding `"See also remove"` to the `without` section?

chriddyp:
@CrossEye - Yup! Done.

CrossEye:
Thank you!

PR #2556, opened 2018-06-07
Update the 'path' method to it could take a string as the 1st param.

Kalita-Roman:
The idea is to make it easier to use the 'path' method. The method can take not only an array as 1st parameter but and a  string which describes a path. The separator for paths' steps is dot `.` 
E.g.
`'a.b.c'` instead `['a', 'b', 'c']`.   
`R.path('a.b', {a: {b: 2}});` instead `R.path(['a', 'b'], {a: {b: 2}});`

### DIFF
--- a/source/path.js
+++ b/source/path.js
@@ -10,6 +10,7 @@ import _curry2 from './internal/_curry2';
  * @category Object
  * @typedefn Idx = String | Int
  * @sig [Idx] -> {a} -> a | Undefined
+ * @sig String -> {a} -> a | Undefined
  * @param {Array} path The path to use.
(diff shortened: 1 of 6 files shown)


CrossEye:
Thank you for your PR.

Unfortunately, I think that API is a non-starter for Ramda.  See #876 for the discussion on where we moved *away* from that API.

tommmyy:
@Kalita-Roman Function is already included in [ramda-extension](https://ramda-extension.firebaseapp.com/docs/#dotPath).

Kalita-Roman:
Thanks for information!

PR #2558, opened 2018-06-12
Cleaner example for reduceBy

CrossEye:
The original is pretty hard to follow.

### DIFF
--- a/source/reduceBy.js
+++ b/source/reduceBy.js
@@ -30,7 +30,7 @@ import _xreduceBy from './internal/_xreduceBy';
  * @example
  *
  *      const groupNames = (acc, student) => acc.concat(student.name)
- *      const toGrade = (student, score = student.score) =>
+ *      const toGrade = ({score}) =>
  *        score < 65 ? 'F' :
  *        score < 70 ? 'D' :
  *        score < 80 ? 'C' :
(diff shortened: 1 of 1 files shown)


source/reduceBy.js
```diff
@@ -29,25 +29,22 @@ import _xreduceBy from './internal/_xreduceBy';
...
- *                      {name: 'Bart', score: 62}];
- *      namesByGrade(students);
- *      // {
- *      //   'A': ['Lucy'],
- *      //   'B': ['Drew']
- *      //   // ...,
- *      //   'F': ['Bart']
- *      // }
+ *      const groupNames = (acc, student) => acc.concat(student.name)
+ *      const toGrade = (student, score = student.score) =>
```
Bradcomp: TIL you could define parameters in terms of other parameters in JS. 

I had to run this in the REPL to make sure I knew what was going on, and you have a function that _appears_ to take two parameters even though the sig calls for one.

Would it be simpler to just do a basic destructuring: `const toGrade = ({score}) => ...`?

CrossEye: Yes it would.  I wasn't thinking straight.  I was changing the existing function and wanted to make it a simple expression, and that version was the first thing that came to my mind.  I will to use simple destructuring.

Issue #2560, opened 2018-06-13
Documentation for ascend incomplete


matthewboman:
Most examples show the result of calling the provided functions. The [documentation for ascend](https://ramdajs.com/docs/#ascend) doesn't provide an example of what the people array would look like or what the output of running `R.sort(byAge, people)` would be.

Suggested fix: 

    var byAge = R.ascend(R.prop('age'));
    var people = [
      { name: 'Emma', age: 70 }, 
      { name: 'Peter', age: 78 },
      { name: 'Mikhail', age: 62 },
    ];
    var peopleByYoungestFirst = R.sort(byAge, people); 
    //=> [{ name: 'Mikhail', age: 62 },{ name: 'Emma', age: 70 }, { name: 'Peter', age: 78 }]

If people feel this is necessary/useful, I'm down to add this to the documentation.

CrossEye:
Yes, if you would be willing to create a PR, we'd be grateful.

crashspringfield:
Great. Is the documentation in the [https://github.com/ramda/ramda.github.io]( https://github.com/ramda/ramda.github.io) project?

Bradcomp:
@crashspringfield You can update the comments in https://github.com/ramda/ramda/blob/master/source/ascend.js and it will automatically update the site upon the next release.

PR #2562, opened 2018-06-14
update ascend, comparator, descend docs

crashspringfield:
I opened issue #2560 to update the ascend documentation. The comparator and descend functions use a similar example so I updated their documenation as well.


buzzdecafe:
:cow2:

PR #2565, opened 2018-06-19
Added functions then and otherwise

Bradcomp:
New and improved `then` PR open for discussion. I tried to incorporate the comments from the previous PR while changing as little as possible otherwise. 

Please review the type signatures, examples, and wording on the `_assertPromise` errors.

For reference, https://github.com/ramda/ramda/pull/1906 is the prior discussion.


source/fail.js
```diff
@@ -0,0 +1,35 @@
...
+ * Returns the result of applying the onFailure function to the value inside
+ * a failed promise. This is useful for handling rejected promises
+ * inside function compositions.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig (e -> b) -> (Promise e a) -> (Promise e b)
+ * @sig (e -> (Promise f b)) -> (Promise e a) -> (Promise f b)
+ * @param {Function} onFailure The function to apply. Can return a value or a promise.
```
buzzdecafe: "Can return a value or a promise of a value."

Does this mean we are signing up for `then` as collapsed `map/chain`?

source/then.js
```diff
@@ -0,0 +1,35 @@
...
+ * Returns the result of applying the onSuccess function to the value inside
+ * a successfully resolved promise. This is useful for working with promises
+ * inside function compositions.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig (a -> b) -> (Promise e a) -> (Promise e b)
+ * @sig (a -> (Promise e b)) -> (Promise e a) -> (Promise e b)
+ * @param {Function} onSuccess The function to apply. Can return a value or a promise.
```
buzzdecafe: as above

source/fail.js
Bradcomp: Well, `then` _is_ a collapsed `map/chain`. I tried to add the signatures for each case. It doesn't make sense to only support one or the other, I can't think of which one we would prioritize.

test/fail.js
```diff
@@ -0,0 +1,38 @@
...
+  it('catches failed promises', function(done) {
+    var p = Promise.reject(1);
+
+    var f = function(n) {
+      eq(n, 1);
+      done();
+    };
+
+    R.fail(f, p);
+  });
```
CrossEye: I personally find this easier to grasp:
```js
R.fail(
  (n) => {eq(n, 1); done()},
  Promise.reject(1)
);
```

But I may be alone in this.

test/fail.js
```diff
@@ -0,0 +1,38 @@
...
+
+  it('does nothing to successfully resolved promises', function(done) {
+    var asyncAddTwo = R.pipe(Promise.resolve.bind(Promise), R.then(R.inc), R.fail(R.inc), R.then(R.inc));
```
CrossEye: What if the `fail` part was `R.fail(R.multiply(0))`?  That would make it more clear what does and doesn't run...

test/then.js
```diff
@@ -0,0 +1,54 @@
...
+  it('invokes then on the promise with the function passed to it', function(done) {
+    var p = Promise.resolve(1);
+
+    var f = function(n) {
+      eq(n, 1);
+      done();
+    };
+
+    R.then(f, p);
+  });
```
CrossEye: As above.  I find this easier to read with the named variables inlined instead.

test/then.js
```diff
@@ -0,0 +1,54 @@
...
+            function() { R.then(R.inc, 1); },
+            function(err) {
+              return err.constructor === TypeError &&
+                       err.message === '`then` expected a Promise, received 1';
+            }
+        );
+
+  });
+
+  it('is not dependent on a particular promise implementation', function(done) {
```
CrossEye: :+1:

bas080:
I have some other naming suggestions. I find the name `fail` not intuitive. Promises can resolve or reject.

`rejected` and `resolved`

or

`resolution` and `rejection`

or 

`caught` and `resolved`

or

`mapCatch/mapRejected` and `mapThen/mapResolved`

CrossEye:
> I find the name `fail` not intuitive. Promises can resolve or reject.

They also can't `then`, but that name seems unequivocally correct.  The trouble is that its counterpart, `catch`, is not available as an identifier.  And Ramda functions are often exposed without the `R` namespace, so we have a conflict.

`fail` seems to me a reasonable compromise, but I certainly could imagine other possibilities.

Of course the best name is `map`, but [that ship has sailed][pf].

  [pf]: https://github.com/promises-aplus/promises-spec/issues/94

bas080:
How about `then & otherwise`..

An example would be a function that increments when `getCount` rejects to get an existing count.

```js
pipe(getCount(), then(inc), otherwise(always(0)))

// vs

pipe(getCount(), then(inc), fail(always(0)))
```

Forgive me for my persistence 😄.

CrossEye:
I don't care between `then`/`fail` and `then`/`otherwise`.

@Bradcomp: what are your thoughts?

Bradcomp:
I could go for that. I am not really wedded to the name `fail` and I think `otherwise` sounds nice when working with piped functions and whatnot.

Personally, I am not 100% sold we _need_ a `fail/otherwise` function other than that it makes sense given that we are adding `then`

CrossEye:
@Bradcomp:

I think there is a good reason for `otherwise`.  There is clearly a desire to in userland to work with Promises.  We've seen that Ramda offers only sporadic support for them.  If we offer `then` and `otherwise` we can simply point to them whenever someone asks about Promise support.  "You can build it using these.  Batteries not included."  And this should cover almost anything people do with Pro,mises, including async functions. (I suppose there's still `finally` to cover if we find a need, but that should be simple enough.)  And we can go back to using Futures or Tasks for our own work.

source/internal/_assertPromise.js
```diff
@@ -0,0 +1,8 @@
+import _isFunction from './_isFunction';
+import _toString from './_toString';
+
+export default function _assertPromise(name, p) {
```
evan-king: I read in earlier issues from the core devs (sorry, don't have references on hand) that Ramda's philosophy regarding undefined behavior was not to throw exceptions of your own but let low-level failures bubble up when users do something that doesn't make sense.

Is that a fair characterization, and if so, what makes this case different?

bas080: @evan-king, I think you have a valid point. It will throw a type error anyways when trying to call the then method on the object.

```js
a.then()
// ReferenceError: a is not defined

a = {}; a.then()
//TypeError: a.then is not a function
```
By using JS default behavior we get more information about the reason of the error.

source/internal/_assertPromise.js
```diff
@@ -0,0 +1,8 @@
...
+  if (p == null || !_isFunction(p.then)) {
+    throw new TypeError('`' + name + '` expected a Promise, received ' + _toString(p, []));
```
bas080: Considering that these helpers are generic enough to work with then-ables, it might make sense to reflect that in the code and documentation.

CrossEye: Perhaps, but at this point, that should be a separate PR.

PR #2567, opened 2018-06-20
New function to recursively transform an objects keys

tomwhale:
Based on the `evolve` function. This recursively transforms the keys of the object. 

```js
const tomato = {firstName: 'John', lastName: 'Smith', emailAddress: 'foo@BAR.com'};
const transformations = {
     firstName: R.toUpper,
     lastName: R.toLower
};
R.evolveKeys(transformations, tomato);
//=> {FIRSTNAME: 'John', lastname: 'Smith', emailAddress: 'foo@BAR.com'}
```

### DIFF
--- a/source/evolveKeys.js
+++ b/source/evolveKeys.js
@@ -31,8 +31,8 @@ var evolveKeys = _curry2(function evolveKeys(transformations, object) {
   for (key in object) {
     transformation = transformations[key];
     type = typeof transformation;
-    newKey = type === 'function'                 ? transformation(key)
-                                                 : key;
+    newKey = type === 'function' ? transformation(key)
+                                 : key;
 
     result[newKey] = transformation && type === 'object' ? evolveKeys(transformation, object[key])
                                                          : object[key];
--- a/source/evolveKeys.js
+++ b/source/evolveKeys.js
@@ -1,0 +1,42 @@
+import _curry2 from './internal/_curry2';
+
+
+/**
+ * Creates a new object by recursively transforming the keys based on the
+ * supplied transformation function.
+ *
+ * A `transformation` function will not be invoked if its corresponding key
+ * does not exist in the evolved object.
+ *
+ * @func
+ * @memberOf R
+ * @category Object
+ * @sig {k: (v -> v)} -> {k: v} -> {k: v}
+ * @param {Object} transformations The object specifying transformation functions to apply
+ *        to the object.
+ * @param {Object} object The object to be transformed.
+ * @return {Object} The transformed object.
+ * @example
+ *
+ *      const tomato = {firstName: 'John', lastName: 'Smith', emailAddress: 'foo@BAR.com'};
+ *      const transformations = {
+ *          firstName: R.toUpper,
+ *          lastName: R.toLower
+ *      };
+ *      R.evolveKeys(transformations, tomato); //=> {FIRSTNAME: 'John', lastname: 'Smith', emailAddress: 'foo@BAR.com'}
+ */
+var evolveKeys = _curry2(function evolveKeys(transformations, object) {
+  var result = {};
+  var transformation, key, newKey, type;
+  for (key in object) {
+    transformation = transformations[key];
+    type = typeof transformation;
+    newKey = type === 'function'                 ? transformation(key)
+                                                 : key;
+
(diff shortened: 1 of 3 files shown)


CrossEye:
I'd like to see some arguments for *why* we should include this.  At a first glance, this function looks well-written, and aside from some possible formatting issues, seems to meet Ramda coding standards.

But why is this important  enough to include in the main library?

Is this something you've reached for often?  I've personally never used anything like this, but perhaps I'm missing something critical.

kurtmilam:
Not necessarily arguing for the inclusion of this function in the library, but I will offer a (simplified) real-world use case based on the project I'm currently working on.

The project I'm currently on includes several ETL (extract, transform, load) services that import into the application database CSV files uploaded by the user. The CSV files have all caps kebab case column headers like `FIRST_NAME`, `LAST_NAME`, `EMAIL`.

The [`csvtojson`](https://github.com/Keyang/node-csvtojson) library we use to consume the CSVs turns each CSV row into an object that look like this:
```
{
  FIRST_NAME: "John",
  LAST_NAME: "Smith",
  EMAIL: "js@example.com"
}
```

One of the ETL services POSTs to an API endpoint once for each row in the CSV. The API service expects a json payload that looks like this:
```
{
  firstName: "John",
  lastName: "Smith",
  email: "js@example.com"
}
```

There are similar use cases, usually having some integration component whereby data from one source needs to be reshaped before it can be passed on to a second source.

adriaan-pelzer:
This would make my life a lot easier.

One of the most common things we do on a day to day basis is data transformation, from some random public data source's schema and conventions to our own (often to merge into existing data.

This always involves using toPairs and fromPairs to separate keys and values into pairs, and then have a spec of transforming functions for the key and the value separately. It's not hard to do, but reads a bit awkwardly, and just always seems too verbose.

This will make it much more readable.

@kurtmilam 's example is a good one.

Bradcomp:
I tend to agree with @CrossEye on this one. It seems _very_ specific. To me all the use cases listed so far could be better handled by [`mapKeys`](https://github.com/ramda/ramda/wiki/Cookbook#map-keys-of-an-object) or [`renameKeys`](https://github.com/ramda/ramda/wiki/Cookbook#rename-keys-of-an-object), both of which are already part of the cookbook, and one of which is in Ramda-Adjunct. For this function to be necessary as it stands, you somehow need to know what the original key name is while at the same time not being able to just specify the new key name. 

The one thing this has that those don't is the ability to be recursive, but none of the given examples make use of that.

bas080:
The ramda wiki has an amazing list of useful functions. They are called recipes. Maybe it's worth adding this to that list.

https://github.com/ramda/ramda/wiki/Cookbook

CrossEye:
I agree with @Bradcomp (in his agreement with me :stuck_out_tongue_winking_eye: ) that the examples presented can be handled well with current functions or Cookbook examples.  But I've done similar transformations fairly often, and I'm wondering if one of my regular functions would make sense for Ramda:

```js
const collect = curry((spec, obj) => map(
  val => is(Array, val) ? path(val, obj) : collect(val, obj), 
  spec
))

// Simpler version if output is flat 
// const collect = curry((spec, obj) => map(path(__, obj), spec))

const person = {
  ID: '12345',
  NAME: {
    FIRST: 'Fred',
    LAST: 'Flintstone',
  },
  'EMAIL-ADDRESS': 'fred@bedrock.com',
  AGE: 27,
  GENDER: 'M',
  OCCUPATION: 'Bronto Crane Operator'
}

const spec = {
  first: ['NAME', 'FIRST'],
  last: ['NAME', 'LAST'],
  email: ['EMAIL-ADDRESS'],
  demographics: {
    age: ['AGE'],
    gender: ['GENDER']
  }
}

collect(spec)(person) //=> 
// {
//     demographics: {
//         age: 27,
//         gender: "M"
//     },
//     email: "fred@bedrock.com",
//     first: "Fred",
//     last: "Flintstone"
// }
```

(My actual code uses single strings such as `first: 'NAME.FIRST'` rather than the array, but that's not a good fit for a more general-purpose Ramda.)

tommmyy:
Hi, we have similar function in ramda-extension - [camelizeKeys](https://github.com/tommmyy/ramda-extension/blob/v0.3.7/packages/ramda-extension/src/camelizeKeys.js)... maybe you can take some inspiration from   our implementation.

PR #2569, opened 2018-06-26
adding similar omit/pick to see also on dissoc/assoc

matthewdaniel:


matthewdaniel:
@buzzdecafe - is this done correctly or does ramda.js need to be updated

Bradcomp:
While I can see how `omit` is very similar to `dissoc`, I don't see the similarity between `assoc` and `pick`.

CrossEye:
This is done just fine.

While I don't have any objection to `assoc: see also 'pick'` -- because more connections such as this help make everything more discoverable, it's a strange one.  The only reason seems to be that `assoc` and `dissoc` are closely related as are `pick` and `omit`, and since `dissoc` and `omit` are also closely related, it seems that `assoc` and `pick` *must* be.  But there really seems to be little commonality.

However, as I said, connections are good.

CrossEye:
Oh weird, I swear the comment from @Bradcomp was not there when I responded...

Bradcomp:
> The only reason seems to be that assoc and dissoc are closely related as are pick and omit, and since dissoc and omit are also closely related, it seems that assoc and pick must be. But there really seems to be little commonality.

See, for me this serves to suggest the opposite conclusion. The fact that `omit` and `pick` have much in common, but not `pick` and `assoc` suggests to me that `dissoc` and `omit` are maybe not as closely related as they seem to be on the surface 😛

CrossEye:
I think that's actually closer to the truth.  `omit` really is the dual to `pick`.  `pick` is the one-off version of Codd's `project`: present these features of our datum in their own structure.  `omit` simply says to present *all but* these features.  `assoc` and `dissoc` are much more closely tied with adding or removing properties.  `assoc` says to create a structure which adds this feature to our datum.  `dissoc` says to create a structure which removes this feature from our datum.

I'm not going looking, but I believe we had a request at some point to combine `dissoc` and `omit`, and it's this feeling that made me little interested in even looking at it, even if we couched our objections in points of API simplicity.  But I don't think I considered then how little sense `pick`/`assoc` made.

I still see no reason to revert.  As I said, adding connections, even if they're rather distant, helps users discover more of the library.  Do you disagree?

Bradcomp:
I see no reason to revert, just thought it was an interesting rabbit hole 😁

PR #2577, opened 2018-07-05
casting to Boolean

roastedfrost-zz:
Currently function doesn't return Boolean - it returns result of predicate-function. Ideally, predicate shouldn't return anything except Boolean, but it's easy to introduce a bug like this.
```javascript
R.propSatisfies(
    prop => prop && prop.a && prop.a.b > 3,
    'property'
);
```

### DIFF
--- a/source/pathSatisfies.js
+++ b/source/pathSatisfies.js
@@ -22,6 +22,6 @@ import path from './path';
  *      R.pathSatisfies(y => y > 0, ['x', 'y'], {x: {y: 2}}); //=> true
  */
 var pathSatisfies = _curry3(function pathSatisfies(pred, propPath, obj) {
-  return propPath.length > 0 && pred(path(propPath, obj));
+  return propPath.length > 0 && Boolean(pred(path(propPath, obj)));
 });
 export default pathSatisfies;
--- a/source/propSatisfies.js
+++ b/source/propSatisfies.js
@@ -21,6 +21,6 @@ import _curry3 from './internal/_curry3';
  *      R.propSatisfies(x => x > 0, 'x', {x: 1, y: 2}); //=> true
  */
 var propSatisfies = _curry3(function propSatisfies(pred, name, obj) {
-  return pred(obj[name]);
+  return Boolean(pred(obj[name]));
 });
 export default propSatisfies;
--- a/test/pathSatisfies.js
+++ b/test/pathSatisfies.js
@@ -2,7 +2,7 @@ var R = require('..');
 var eq = require('./shared/eq');
 
 
-describe('pathSatisfies', function() {
+describe('pathSatisfies returns Boolean taking predicate', function() {
 
   var isPositive = function(n) { return n > 0; };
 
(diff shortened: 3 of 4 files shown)


roastedfrost:
Most likely other similar functions have the same behavior, I can prepare the changes if needed

CrossEye:
:herb:

This looks good to me.  Anyone have objections?  @ramda/core?

CrossEye:
As to other functions, if you want to take a look, we'd love the help!

Bradcomp:
:bowtie:

roastedfrost:
yep, I'll do within this PR

buzzdecafe:
:cow2:

paldepind:
I think this is a fine change. But, I think it should be documented that the predicate function does not have to return `true` or `false` and that `propSatisfies` will return `true` as long as the predicate returns a truthy value.

I think it should be documented both in the written explanation and with an example where the predicate returns a truthy value.

TheLudd:
I do not think ramda should take responsibility for users breaking the contract of the functions. The reasoning in this PR could be expanded to all functions and situations and I think that would create unreasonable demands on the library.

Further on, there are other predicate functions in this lib. What about `either`, `both` etc. 
Is it ok to change this function but not others?
If not, what maintenance demands to we put on the lib with this change?

TheLudd:
IMHO, if you want this function to always return a Boolean, a better implementation would be `pred(obj[name]) === true`

bas080:
@TheLudd, I like what you are proposing. Making that change will require a major version.

I also think there are arguments for casting truthy values to true: 
- This change is less likely to break existing code.
- Supporting truthy values feels more like Javascript.

I have never run into issues because these type of functions don't return true. I tend to call these kind of functions with actual predicate functions. As a result I always get a boolean back.

What this change does is make sure someone doesn't do something really wrong which they should not be doing to begin with.

Static type checking would solve this issue by not allowing anything but predicate functions to be passed.

roastedfrost:
@TheLudd reasonable, but `propSatisfies` neither returns Boolean nor handle that case.. (somehow - throw an error, cast value - doesn't matter).
I think it worth to update documentation at least.
btw, I had that error a while ago

CrossEye:
@TheLudd:

Even if its reputation isn't as shiny as it once was, I would still invoke [Postel's Law][pl] here:

>  Be conservative in what you do, be liberal in what you accept from others.

We have usually documented our aspirational types.  And in this case in particular, I see very little reason to enforce a rule that says users must supply a boolean-returning function.  In JS, there is a longstanding rule on what things are false-y.  I see no reason to ignore that in our JS library.  So "be liberal in what you accept from others."

The current PR fixes the other half of Postel's law.  If we force the return type to be a boolean, we align with "be conservative in what you do."

Postel's law is never going to be the only criterion.  We have to balance this against [remaining simple][se], but when we can follow it, I think it's worth doing.



  [pl]: https://en.wikipedia.org/wiki/Robustness_principle
  [se]: https://www.infoq.com/presentations/Simple-Made-Easy

CrossEye:
@paldepind:

While I would have no problems adding such an example, I don't see a compelling need.  Can you explain why you find it important?  (Or would we simply end up rehashing #1218 and #1335?

CrossEye:
@paldepind:

>> While I would have no problems adding such an example, I don't see a compelling need. Can you explain why you find it important?

> Because I think the behavior of functions should be documented :smile:

Sure, go ahead, get all technical on me!  :smile:


This is a (probably very minor) breaking change.  Perhaps that's enough reason to document it.  But I'm all for documenting how we expect people to use our functions and not making any guarantees about how we handle it if they don't. 

But again, I have no issues with this, especially as this is technically a breaking change.

CrossEye:
As minor a breaking change as this is, we're hoping not to introduce any breaking changes before `v1.0`.  I'm closing this now, but we can resurrect it later.

PR #2580, opened 2018-07-08
Adds Traversable instance for Objects

scott-christopher:
This introduces a new function which can then be used to dispatch `Object` types for `R.traverse` and `R.sequence`.
```
R.traverseWithKey :: Applicative f => (a -> f a) -> ((k, a) -> f b) -> {a} -> f {b}
```

This effectively maps over the object's set of keys and values with the provided function, producing `Applicative` values for each. The object is then rebuilt within the context of the `Applicative`.

```js
const allValid = (k, v) => k.length >= 3 && v > 0 ? Just(v) : Nothing();

R.traverseWithKey(Maybe.of, allValid, { a: 1, b: 2, c: 3 }); //=> Nothing
R.traverseWithKey(Maybe.of, allValid, { foo: 0, bar: 1, baz: 2 }); //=> Nothing
R.traverseWithKey(Maybe.of, allValid, { foo: 1, bar: 2, baz: 3 }); //=> Just({ foo: 1, bar: 2, baz: 3 })
```

```js
R.sequence(Maybe.of, { a: Just(1), b: Just(2), c: Just(3) });   //=> Just({a: 1, b: 2, c: 3})
R.sequence(Maybe.of, { a: Just(1), b: Just(2), c: Nothing() }); //=> Nothing()
```

```js
const safeDiv = n => d => d === 0 ? Maybe.Nothing() : Maybe.Just(n / d)

R.traverse(Maybe.of, safeDiv(10), {a: 2, b: 4, c: 5}); //=> Maybe.Just({a: 5, b: 2.5, c: 2})
R.traverse(Maybe.of, safeDiv(10), {a: 2, b: 0, c: 5}); //=> Maybe.Nothing
```

### DIFF
--- a/source/index.js
+++ b/source/index.js
@@ -220,6 +220,7 @@ export { default as toUpper } from './toUpper';
 export { default as transduce } from './transduce';
 export { default as transpose } from './transpose';
 export { default as traverse } from './traverse';
+export { default as traverseWithKey } from './traverseWithKey';
 export { default as trim } from './trim';
 export { default as tryCatch } from './tryCatch';
 export { default as type } from './type';
--- a/source/internal/_flattenCons.js
+++ b/source/internal/_flattenCons.js
@@ -1,0 +1,9 @@
+// flattens [a, [b, [c, []]]] to [a, b, c]
+export default function _flattenCons(xs) {
+  var flattened = [];
+  while (xs.length > 0) {
+    flattened.push(xs[0]);
+    xs = xs[1];
+  }
+  return flattened;
+}
--- a/source/internal/_sequenceArray.js
+++ b/source/internal/_sequenceArray.js
@@ -1,0 +1,12 @@
+import _flattenCons from './_flattenCons';
+import ap from '../ap';
+import map from '../map';
+import pair from '../pair';
+import reduceRight from '../reduceRight';
+
+
+export default function _sequenceArray(of, xs) {
+  return map(_flattenCons, reduceRight(function(x, acc) {
+    return ap(map(pair, x), acc);
+  }, of([]), xs));
+}
--- a/source/sequence.js
+++ b/source/sequence.js
@@ -1,8 +1,8 @@
 import _curry2 from './internal/_curry2';
-import ap from './ap';
-import map from './map';
-import prepend from './prepend';
-import reduceRight from './reduceRight';
+import _isArrayLike from './internal/_isArrayLike';
+import _isObject from './internal/_isObject';
+import _sequenceArray from './internal/_sequenceArray';
+import traverseWithKey from './traverseWithKey';
 
 
 /**
@@ -10,7 +10,8 @@ import reduceRight from './reduceRight';
  * of [Applicative](https://github.com/fantasyland/fantasy-land#applicative) into an
  * Applicative of Traversable.
  *
- * Dispatches to the `sequence` method of the second argument, if present.
+ * Objects and Arrays are both accepted as types of Traversable, otherwise this will
+ * dispatch to the `sequence` method of the second argument, if present.
  *
  * @func
  * @memberOf R
@@ -20,20 +21,27 @@ import reduceRight from './reduceRight';
  * @param {Function} of
  * @param {*} traversable
  * @return {*}
- * @see R.traverse
+ * @see R.traverse, R.traverseWithKey
  * @example
  *
  *      R.sequence(Maybe.of, [Just(1), Just(2), Just(3)]);   //=> Just([1, 2, 3])
  *      R.sequence(Maybe.of, [Just(1), Just(2), Nothing()]); //=> Nothing()
  *
+ *      R.sequence(Maybe.of, { a: Just(1), b: Just(2), c: Just(3) });   //=> Just({a: 1, b: 2, c: 3})
+ *      R.sequence(Maybe.of, { a: Just(1), b: Just(2), c: Nothing() }); //=> Nothing()
+ *
  *      R.sequence(R.of, Just([1, 2, 3])); //=> [Just(1), Just(2), Just(3)]
  *      R.sequence(R.of, Nothing());       //=> [Nothing()]
  */
 var sequence = _curry2(function sequence(of, traversable) {
-  return typeof traversable.sequence === 'function' ?
-    traversable.sequence(of) :
-    reduceRight(function(x, acc) { return ap(map(prepend, x), acc); },
-                of([]),
-                traversable);
+  if (typeof traversable.sequence === 'function') {
+    return traversable.sequence(of);
+  } else if (_isArrayLike(traversable)) {
+    return _sequenceArray(of, traversable);
+  } else if (_isObject(traversable)) {
+    return traverseWithKey(of, function(v, k) { return v; }, traversable);
+  } else {
+    throw new TypeError('Unsupported traversable instance');
+  }
 });
 export default sequence;
--- a/source/traverse.js
+++ b/source/traverse.js
@@ -9,7 +9,8 @@ import sequence from './sequence';
  * then uses [`sequence`](#sequence) to transform the resulting Traversable of Applicative
  * into an Applicative of Traversable.
  *
- * Dispatches to the `traverse` method of the third argument, if present.
+ * Objects and Arrays are both accepted as types of Traversable, otherwise this will
+ * dispatch to the `traverse` method of the third argument, if present.
  *
  * @func
  * @memberOf R
@@ -20,7 +21,7 @@ import sequence from './sequence';
  * @param {Function} f
  * @param {*} traversable
  * @return {*}
- * @see R.sequence
+ * @see R.sequence, R.traverseWithKey
  * @example
  *
  *      // Returns `Maybe.Nothing` if the given divisor is `0`
@@ -28,6 +29,9 @@ import sequence from './sequence';
  *
  *      R.traverse(Maybe.of, safeDiv(10), [2, 4, 5]); //=> Maybe.Just([5, 2.5, 2])
  *      R.traverse(Maybe.of, safeDiv(10), [2, 0, 5]); //=> Maybe.Nothing
+ *
+ *      R.traverse(Maybe.of, safeDiv(10), {a: 2, b: 4, c: 5}); //=> Maybe.Just({a: 5, b: 2.5, c: 2})
+ *      R.traverse(Maybe.of, safeDiv(10), {a: 2, b: 0, c: 5}); //=> Maybe.Nothing
  */
 var traverse = _curry3(function traverse(of, f, traversable) {
   return typeof traversable['fantasy-land/traverse'] === 'function' ?
--- a/source/traverseWithKey.js
+++ b/source/traverseWithKey.js
@@ -1,0 +1,36 @@
+import _curry3 from './internal/_curry3';
+import _sequenceArray from './internal/_sequenceArray';
+import keys from './keys';
+import map from './map';
+import zipObj from './zipObj';
+
+
+/**
+ * Maps an [Applicative](https://github.com/fantasyland/fantasy-land#applicative)-returning
+ * function over the keys and values of an Object, then constructs a new Object within the
+ * Applicative consisting of each returned Applicative value.
+ *
+ * @func
+ * @memberOf R
+ * @since v0.19.0
+ * @category Object
+ * @sig Applicative f => (a -> f a) -> ((a, k) -> f b) -> {a} -> f {b}
+ * @param {Function} of
+ * @param {Function} f
+ * @param {*} obj
+ * @return {*}
+ * @see R.sequence, R.traverse
+ * @example
+ *
+ *      const allValid = (v, k) => k.length >= 3 && v > 0 ? Just(v) : Nothing();
+ *
+ *      R.traverseWithKey(Maybe.of, allValid, { a: 1, b: 2, c: 3 }); //=> Nothing
+ *      R.traverseWithKey(Maybe.of, allValid, { foo: 0, bar: 1, baz: 2 }); //=> Nothing
+ *      R.traverseWithKey(Maybe.of, allValid, { foo: 1, bar: 2, baz: 3 }); //=> Just({ foo: 1, bar: 2, baz: 3 })
+ */
+var traverseWithKey = _curry3(function traverseWithKey(of, f, obj) {
+  var ks = keys(obj).sort();
+  var fxs = _sequenceArray(of, map(function(k) { return f(obj[k], k); }, ks));
+  return map(zipObj(ks), fxs);
+});
+export default traverseWithKey;
--- a/test/sequence.js
+++ b/test/sequence.js
@@ -33,4 +33,17 @@ describe('sequence', function() {
     eq(R.sequence(R.of, Id([1, 2, 3])), [Id(1), Id(2), Id(3)]);
   });
 
+  it('operates on an object of applicatives', function() {
+    eq(R.sequence(S.Maybe.of, { a: S.Just(3), b: S.Just(4), c: S.Just(5)}), S.Just({a: 3, b: 4, c: 5}));
+    eq(R.sequence(S.Maybe.of, { a: S.Just(3), b: S.Nothing(), c: S.Just(5)}), S.Nothing());
+  });
+
+  it('performs the applicative effects in a consistent order on objects', function() {
+    eq(R.sequence(R.of, { a: [1, 2], b: [3, 4] }),
+       R.sequence(R.of, { b: [3, 4], a: [1, 2] }));
+
+    eq(R.sequence(S.Either.of, { a: S.Left(0), b: S.Left(1) }),
+       R.sequence(S.Either.of, { b: S.Left(1), a: S.Left(0) }));
+  });
+
 });
--- a/test/traverse.js
+++ b/test/traverse.js
@@ -21,6 +21,20 @@ describe('traverse', function() {
     eq(R.traverse(S.Maybe.of, R.map(R.add(10)), [S.Just(3), S.Nothing(), S.Just(5)]), S.Nothing());
   });
 
+  it('operates on an object of applicatives', function() {
+    eq(R.traverse(S.Maybe.of, R.map(R.add(10)), { a: S.Just(3), b: S.Just(4), c: S.Just(5) }),
+       S.Just({ a: 13, b: 14, c: 15}));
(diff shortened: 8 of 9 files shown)


polytypic:
Cool.  I have a few ideas:

* Swap the order of arguments that `traverseWithKey` provides to the user given function.  IOW, `(value, key) => ...` rather than `(key, value) => ...`.  That way it would be more consistent with existing functions like `mapObjIndexed`.

* Implement this so that it has linear, `O(n)`, time complexity instead of quadratic, `O(n*n)`.  You can do that by e.g. accumulating a cons list of the key value pairs, `list = null | [key, value, list]`, and then constructing the object at the end.  This also applies to the existing `traverse` / `sequence` on arrays, which is currently quadratic.

* Consider documenting the order in which `traverse` / `sequence` calls the user given function.  IOW, consider specifying that `R.addIndex(R.traverse(of))` works as one might expect.

* Consider making it so that `R.addIndex` would work for both arrays and objects.  The current implementation already assumes that array elements are iterated in specific order.  If you would make it so that object traversing functions would access the properties of an object in the same order (in the same order with respect to each other), then you could extend `R.addIndex` to also work with objects.  If so, you could actually drop `R.mapObjIndexed` (would be `R.addIndex(R.map)`) and this proposed `R.traverseWithKey` (would be `R.addIndex(R.traverse(of))`).

CrossEye:
@scott-christopher:

This looks very good.

@polytypic:

I like your first three points.  The fourth one scares me a bit.  There was such pushback to **[my idea][ss]** of defining our own consistent iteration order for objects, that I dropped it, although I still would like to do it.  This proposal seems to require that.


  [ss]: https://github.com/ramda/ramda/issues/1067#issuecomment-135923353

scott-christopher:
@polytypic / @CrossEye: I've updated the PR to swap the argument order of the function provided to `traverseWithKey` and updated the implementation to use `zipObj` rather than incurring the overhead by repeatedly calling `assoc`.

I've also split out and modified the `sequence` implementation for arrays into `./internal/_sequenceArray`, such that it now builds up and then flattens a cons list rather than iterating with `prepend`. The reasoning behind moving it out to `./internal` was so `traverseWithKey` can use it while avoiding the circular dependency on `sequence`.

I've avoided changing anything relating to `addIndex` for now, as that can be captured in a separate PR if desired.

CrossEye:
LGTM.  Any other comments, @ramda/core?

davidchambers:
It's good to see you, @scott-christopher! :wave:

scott-christopher:
It's nice to see y'all again too. It's certainly been a while between pull requests. 😅

paldepind:
Maybe I've missed something. But is `reduce` for objects something that we are now comfortable implementing? 

Because, anything that's is a Traversable is also a Foldable. Which means that if `traverse` handles objects then`reduce` should also handle objects.

scott-christopher:
@paldepind This change requires us to be ok with the Applicative effects of each key/value pair occurring in arbitrary order. This feels less prone to confusion than a function provided to `reduce` that could potentially receive its arguments in arbitrary orders. Traversable at least preserves the shape of itself within the context of the Applicative.

If there's concern about calling Object an instance of Traversable then we can update this PR to remove the change in `sequence` and just leave `traverseWithKey` as it is. Mind you, we don't currently dispatch to Traversable instances within `reduce` either.

To still provide Foldable-like support for Objects, a `foldMapObj` or similar could be introduced that just requires the produced values to be commutative monoids.

davidchambers:
I'm still of the opinion that Ramda should depend on [sanctuary-type-classes][1]. If we don't want to add a dependency we could include the source code verbatim and update Ramda's licence accordingly.


paldepind:
@scott-christopher

>  This change requires us to be ok with the Applicative effects of each key/value pair occurring in arbitrary order. This feels less prone to confusion than a function provided to reduce that could potentially receive its arguments in arbitrary orders.

The order isn't arbitrary. Keys are stored in insertion order :wink:

I don't see the difference though. `traverse` is a generalization of `reduce` and `reduce` can be implemented in terms of `traverse`. Thus any argument for not implementing `reduce` applies equally well to `traverse` (and for `traverseWithKey`). Thus if we are ok with `traverse` we should also be ok with `reduce`.

I am in favor of implementing both foldable and traversable for objects since both are immensively useful. But, implementing `traverse` without `reduce` is inconsistent and confusing. We can't have our cake without eating it :smile:

paldepind:
@scott-christopher 

> I say arbitrary because two objects that are deemed equal could produce different results with reduce depending on how they were constructed.

I see your point—but, the very same is true for `traverse` on objects.

> Excluding ordering of effects, traverse doesn't exhibit this behaviour.

That is a big thing to exclude given that, in general, the ordering of effects in applicative does matter and changing the evaluation order does change the result.

> Given two objects that `R.equals` determines to be equal, `traverse` will evaluate to the same resulting value for each,

That is not true. Here is a counter example:

```js
const foo = { a: [1, 2], b: [3, 4] };
const bar = { b: [3, 4], a: [1, 2] };
R.equals(foo, bar)); //=> true

R.traverse(R.of, R.identity, foo)); //=> [ { a: 1, b: 3 }, { a: 1, b: 4 }, { a: 2, b: 3 }, { a: 2, b: 4 } ]
R.traverse(R.of, R.identity, bar)); //=> [ { b: 3, a: 1 }, { b: 3, a: 2 }, { b: 4, a: 1 }, { b: 4, a: 2 } ]

R.equals(
  R.traverse(R.of, R.identity, foo),
  R.traverse(R.of, R.identity, bar)
); //=> false
```

In the example above `R.equals` consider `foo` and `bar` to be equal but it does not consider the value returned by `R.traverse` on those two to be equal. So `R.traverse` suffer from the same thing that `R.reduce` does.

I hope the above example illustrates my point. As `traverse` is just a generalization of `reduce` any argument against allowing `reduce` on object aplies equally well to `traverse`.

functionalStoic:
@paldepind I was exploring your really excellent example and found something odd. The Ramda REPL returns your last `equals` example as `true`. Just an FYI

polytypic:
@functionalStoic, that is because the idea in @paldepind's example is that the objects should be traversed (using the changes in this PR), but the current version of Ramda does not traverse objects.

Partial Lenses provides generic [`traverse`](https://github.com/calmm-js/partial.lenses/#L-traverse) over a given traversal using a given applicative. Here @paldepind's example translated to Partial Lenses:

```js
const foo = { a: [1, 2], b: [3, 4] }
const bar = { b: [3, 4], a: [1, 2] }

log( R.equals(foo, bar) ) //=> true

const ConcatOf = ({empty, concat}) => ({
  map: (_, x) => x,
  of: _ => empty(),
  ap: concat
})

const ArrayApplicative = ConcatOf({empty: R.always([]), concat: R.concat})

R.equals(
  L.traverse(ArrayApplicative, R.identity, L.values, foo),
  L.traverse(ArrayApplicative, R.identity, L.values, bar)
) //=> false
```

Here is [the above code in a playground](https://calmm-js.github.io/partial.lenses/playground.html#MYewdgzgLgBAZiEMC8MDeMCGAuGBtARgBoYAmAXRICNc8BmEgFnJgF8AoUSWKzAJxToYNfAxjMSOfMTIsO7ADYgA5gAoYAJQB0AUwCOAV0wKIqhCGr8AlDBsB6O8gB8MKHwM72ncNBgBhcGBMKAB5OEFVNB0AWwAHKABPEi4gqFYbZxhI9mjMWNxVAH0SAA8MlxKidhA4XEKUFxj4hNUrKrzcFOD2dK8uXwBBPj5MBIHY2IUAS1SpgDcdQQCwVLDIpsTcbWMAd1HTPHI2mC6oLa1T3vZtfSMTVXYAGS03TAW+CB1VIZGxienZgsSNopgATHRgKBTRIkZ5zYweCAkcxtJ4vEbvT7fYajcaTGbBeY6YFaMEQqEwmBwhE6JHCazseyOFxwYyfIA).

Partial Lenses is based on [Static Land](https://github.com/rpominov/static-land), so in the above example we explicitly implement the applicative instance, `ArrayApplicative`, on arrays that is implicitly used in @paldepind's example.  Also, we explicitly specify the [`L.values` traversal](https://github.com/calmm-js/partial.lenses/#L-values) to use.

scott-christopher:
@paldepind My bad, there are actually plenty of applicative effects that `R.equals` won't consider equal when applied in different orders.

So with all the above in mind, do we:

1. Leave this change as it is?
2. Remove support for Objects in `sequence`/`traverse` in this change, keeping `traverseWithKeys`?
3. Add support for Objects in `reduce` and friends?
4. Ensure consistent ordering when traversing objects (e.g. sorted keys)?
5. Abandon this change altogether?

polytypic:
I'd recommend merging the traverse / sequence optimization on arrays first via a separate PR.  Quadratic time complexity is unacceptable for an operation that can be done in linear time.  Just for background, in the project where I'm currently working, people were bitten by quadratic time complexity of Ramda's set operations (like `union` and `difference`) and as a result some developers became rather opposed to using Ramda.  I would recommend making sure that operations work efficiently and in case their time complexities are much worse than optimal it should be documented.

scott-christopher:
I've updated the commit to first sort the keys ensuring the applicative effects are now performed in a consistent order.

CrossEye:
I like this sorting version, and would like to extend this to `keys`, `values`, and the other functions whose values depend on key order.

I think we need to at least consider the [objection raised by @megawac][mw] to the actual sort order.  This may not be an issue here, and we may not ever want to worry about it, if we're sure we would always handle arrays separately, but we should look at it.


  [mw]: https://github.com/ramda/ramda/issues/1067#issuecomment-135935554

CrossEye:
@scott-christopher: Was committing `test/.traverse.js.swp` an oversight?

scott-christopher:
Figured the codebase needed some swap files... (oops, removed) 🙃

scott-christopher:
Regarding the sort order, perhaps it would be best for us to say that the actual order is an implementation detail and it's just the invariant that two objects considered equal will have a consistent sort order. In other words, people shouldn't be relying on a particular sort order.

paldepind:
@scott-christopher 

> Regarding the sort order, perhaps it would be best for us to say that the actual order is an implementation detail and it's just the invariant that two objects considered equal will have a consistent sort order. In other words, people shouldn't be relying on a particular sort order.

That would make those functions significantly less useful. `R.traverse` on objects couldn't be used for the plethora off applicatives where the ordering of effects matter.

IMO sorting is a way too complex and confusing solution to what I think is a non-issue in the first place. It's also bad for performance. It makes an otherwise linear function have `O(n * log(n))` running time.

Objects in JavaScript are key-value pairs stored in insertion order. In some use cases it is useful to think of them as such and other times it is useful to ignore the key order. `R.equals` ignores the key order as that is often what one wants. But, if the decisions in `R.equals` have to stand in the way for functions where using the key order is actually really useful then maybe we should consider changing `R.equals` instead?

_If_ `R.equals` considered key order would you all then be fine with `R.reduce`, `R.traverse`, and `R.keys` depending on key order without sorting?

buzzdecafe:
> So... do we want to make a stance on whether our functions are referentially transparent for objects with the same set of key/value pairs regardless of insertion order, or do we embrace the language as is and accept that insertion order has an impact on how objects are interacted with?

That is a very good summary of a discussion that has previously stretched over several threads. 

I'd be satisfied enough with:

1. a sorted-keys path that guarantees referential transparency for object iteration always; and
2. an unsorted-keys path that only guarantees referential transparency if the function being applied is commutative.

These would have to be clearly labeled to distinguish their purposes. I don't see another way to satisfy the conflicting constraints on this, e.g. ref. transparency, preserving info about key order, performance, consistency, etc.

paldepind:
@polytypic

> Indeed, how frequently people here tend to run into issues due to the key order? Is it something that actually causes real problems? What kind of real problems? What approaches people use to avoid those problems?

Your thoughts reflects mine and I think you're asking very good questions here :+1: Personally key order has never caused me problems and I've written code that depended on key order only rarely. 

@scott-christopher

> So... do we want to make a stance on whether our functions are referentially transparent for objects with the same set of key/value pairs regardless of insertion order

I don't see what this has to due with [referential transparency](https://en.m.wikipedia.org/wiki/Referential_transparency)? Refential transparency is a property of pure code. There's is nothing impure about relying on the order of the keys in an object. So we wouldn't be breaking referential transparency. _If_ relying on key order broke referential transparency I'd definitely be against it. But it doesn't.

scott-christopher:
@paldepind

> If all functions involved in the expression are pure functions and return the same values for **the same arguments**, then the expression is referentially transparent.

This was exactly my point. Do we treat objects with the same key/value pairs as **the same arguments**, or do we treat them as different values.

paldepind:
@scott-christopher 

> This was exactly my point. Do we treat objects with the same key/value pairs as the same arguments, or do we treat them as different values depending on insertion order.

Yes, I think I understand the argument. You're saying that if `R.equals` considers two values to be the same then all other functions in Ramda should also treat them as the same. I think it was @CrossEye who previously expressed that property eloquently as something similar to the following:

> `R.equals(x, y) === true` should, for any Ramda function `f`, imply `R.equals(f(x), f(y)) === true`.

However, the above is _not_ referential transparency. It's a property expressing a relationship between `R.equals` and all other functions in Ramda. I'm not arguing against the above property being desirable. I'm arguing against calling it referential transparency because saying that `R.reduce` breaks referential transparency makes it sound as if it breaks a fundamental property of purely functional programming.

`R.reduce` on objects does not break referential transparency because it is a pure function and pure functions preserve referential transparency. Longer arguments [here](https://github.com/CrossEye/ramda-major-version/issues/2#issuecomment-338494420) and [here](https://github.com/CrossEye/ramda-major-version/issues/2#issuecomment-338763858).

polytypic:
The problem here is indeed that the notion of equality provided by current `R.equals` is not the most discriminating notion of equality available in JS.

I don't claim to be an expert on the EcmaScript standard, but AFAICT, the most discriminating notion of equality available in JS is [`Object.is`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) aka `R.identical` and it holds some very important properties (aside from being reflexive, symmetric, and transitive).  Below are a few of them.

1. `Object.is` holds extensionally.  For all values `x` and `y` and for any (pure) predicate `p`

> `Object.is(x, y)` implies `p(x) === p(y)`.

In other words, given two values that are the same according to `Object.is` there is no way to distinguish `x` and `y` within JS.

This also means that for any (proper) notion of equality in JS, `eq`, it holds that

> `Object.is(x, y)` implies `eq(x, y)`

That is because there is no way to tell apart two values that are the same according to `Object.is`.

2. `Object.is` is *invariant*:

```js
const x = ...
const y = ...

const atTimeIndex1 = Object.is(x, y)

// Arbitrary side-effects

const atTimeIndex2 = Object.is(x, y)

// Now
atTimeIndex1 === atTimeIndex2 // is true
```

As an aside, ever since programming in Standard ML, equality has been one of my litmus tests for assessing programming language quality.  There are very few programming languages whose default notion of equality, [like in Standard ML](http://mlton.org/PolymorphicEquality), is reflexive, symmetric, transitive, holds extensionally, and is invariant.  SML's equality is also robust in that it always returns after a finite amount of computation.  IOW, it cannot get stuck in an infinite loop, for example.

Ramda's `R.equals` doesn't hold either of the above properties.  So, indeed, it is very easy to come up with values `x` and `y` that are equated by `R.equals`, but for which there exists a predicate that tells `x` and `y` apart and expose the fact that `R.equals` does not hold extensionally.  Also, given two distinct objects `x` and `y` that are equated by `R.equals` one can mutate one or both of them to expose the fact that `R.equals` is not invariant.

polytypic:
The fact that `R.equals` does not hold extensionally within JS and is not invariant are not necessarily bad things.  When doing FP in JS, we generally do want to forget that objects are mutable and that any notion of equality, that equates two objects that are not the same according to `Object.is`, is not invariant.  We also often want to forget that object keys are stored in insertion order.

> What should Ramda do?

First of all, that is about making a choice rather than about coming up with the correct answer to a well defined problem.

Should `R.equals` be changed to account for key order?
Should Ramda function operating on objects always use some special ordering for keys?
Should Ramda stop pretending that are only two notions of equality?
...

There are pros and cons to those and many other choices that one could make here.

My personal opinions are as follows:

Sorting keys is too expensive and should not be done.  (I also do not really see that as a solution to the problem once and for all for many reasons.)

Accounting for key order in `R.equals` should not be done.  It might have been a plausible choice earlier, but, in my estimate, making that change would be a *major* breaking change and would probably break almost any application using Ramda.

What I recommend is adding (possibly just documenting) other notions of equality into Ramda.  Seriously.  The idea that there should be only one equality simply does not make sense.  Even with a super nice equality, like in Standard ML, you still want less discriminating notions of equality.  For example, you may want to have a notion of equality that considers two lambda expressions equivalent modulo renaming of bound variables.

`R.equals` simply is not and can never be the only notion of equality that every application should use.  Likewise, I argue that `R.equals` is not the only notion of equality inherent to every function of Ramda and should never be so.  For example, the set operations, like `union` and `difference`, in Ramda essentially operate under a less discriminating notion of equality.  In other words, you can specify a notion of equality, `setEq`, for which

> `setEq(xs1, xs2)` and `setEq(ys1, ys2)` implies `setEq(setFn(xs1, ys1), setFn(xs2, ys2)`

and which is less discriminating than `R.equals`.  That is a notion of equality inherent to a bunch of Ramda functions that (I claim) is different from `R.equals`.

So, there are already two explicitly implemented notions `R.identical` and `R.equals` and, as I claim above, there exist a third one inherent to the set operations.

*Why could there not be other notions of equality inherent to Ramda functions?*  Like notions of equality that are inherent to functions that operate on keys?

CrossEye:

I would much rather work to ensure that Ramda's own notion of equality should be used universally across Ramda (except if we want to keep `R.identical`, which simply reports something different.)  I think this means sorting keys whenever the key order escapes our functions.  I really doubt that the key sorting would be an extreme performance burden.

scott-christopher:
Thinking further about this, if we were to introduce sorting of keys we should probably do so at the origin within `R.keys` itself, such that `R.equals(R.keys({a:1,b:2}), R.keys({b:2,a:1}))` holds.

If we’re not comfortable with that, I’d suggest we remove sorting from `traverseWithKeys` and live with the ordering as determined by the provided objects.

And if we’re not comfortable with that, I’d suggest we pare this back to only include `traverseWithKeys`, leaving `traverse` and friends without support for Objects.

CrossEye:
This will not be possible on the route to `v1.0`.  We can look at it again later.

PR #2581, opened 2018-07-08
Improving the time complexity for the difference function

cuchi:
Before this change, the time complexity for this function was something like O(M x N), being those values the first and the second array lengths.

This can get pretty slow with larger arrays, we learned it the painful way in one of our services :sweat_smile:

This change uses the internal `_Set`, which is O(1)-ish to lookup and insert values. So this function becomes an O(n), which solves our problem.

`differenceWith` remains untouched

### DIFF
--- a/source/difference.js
+++ b/source/difference.js
@@ -1,5 +1,5 @@
-import _includes from './internal/_includes';
 import _curry2 from './internal/_curry2';
+import _Set from './internal/_Set';
 
 
 /**
@@ -26,8 +26,15 @@ var difference = _curry2(function difference(first, second) {
   var out = [];
   var idx = 0;
   var firstLen = first.length;
+  var secondLen = second.length;
+  var toFilterOut = new _Set();
+
+  for (var i = 0; i < secondLen; i += 1) {
+    toFilterOut.add(second[i]);
+  }
+
   while (idx < firstLen) {
-    if (!_includes(first[idx], second) && !_includes(first[idx], out)) {
+    if (toFilterOut.add(first[idx])) {
       out[out.length] = first[idx];
     }
     idx += 1;


bas080:
I tried making a PR to your fork but instead I accidentally created a PR to ramda's repo. I squashed the two commits. You can do this yourself or reset to the branch I have squashed for you.

https://github.com/ramda/ramda/pull/2582

CrossEye:
This looks good to me, but...

Did you do any performance comparisons?  Could you include some information on the results?  Making a change for performance reasons should probably show some before/after information.

cuchi:
@CrossEye I just did made a quick test an put it into [this gist](https://gist.github.com/cuchi/bf6be6436d2e62a5ade1a5c2a649d0c8).

Here is the output I got in Node v8.11.3:
```
Before:
  Many [size 10] arrays: 14.180593
  Many [size 100] arrays: 204.134127
  Some [size 1000] arrays: 1641.455261
  Few [size 10000] arrays: 16040.934775
After:
  Many [size 10] arrays: 47.307934
  Many [size 100] arrays: 152.725264
  Some [size 1000] arrays: 132.193501
  Few [size 10000] arrays: 138.87976
```

source/difference.js
```diff
@@ -26,8 +26,15 @@ var difference = _curry2(function difference(first, second) {
...
+  var secondLen = second.length;
+  var toFilterOut = new _Set();
+
+  for (var i = 0; i < secondLen; i += 1) {
+    toFilterOut.add(second[i]);
+  }
+
   while (idx < firstLen) {
-    if (!_includes(first[idx], second) && !_includes(first[idx], out)) {
+    if (toFilterOut.add(first[idx])) {
```
stefano-b2c2: Any reason not to use the `has` function?

svozza: Any reason not to use the has function?

cuchi: That`s because I need to keep track of every value that was inserted into `out` before (for uniqueness), so `.add` does this job and conveniently let me know if the value is new or repeated.

CrossEye:
Thanks, @cuchi!

That's pretty convincing.  What do you say, @ramda/core?  Also, are there other places that this sort of technique might help?  Our `_Set` was written with the variations of `uniq` in mind, if I remember correctly.  It looks as though it's more widely useful.

There's a part of me that simply wants to export `_Set`, and a corresponding version of `Map`.   But that's outside Ramda's usual practices.  But I know that I've several times pulled it into my projects.  So I can see a need.

Bradcomp:
I like this change a lot. I have had to avoid using the Ramda set operations at times due to the time complexity of them

I think `_Set` would be useful for all our set operations and not just `difference`. 

`union`, `intersection`, and `symmetricDifference` should get it too.

It would be great to get the `with` versions too, but I don't know off the top of my head how we could do that.

cuchi:
I'm glad you both liked it, thanks!

>I think `_Set` would be useful for all our set operations and not just `difference`.
`union`, `intersection`, and `symmetricDifference` should get it too.

I totally agree with that, I just went through the source code and realized that `without` and `intersection` could be improved in the same way. Maybe I could do this in this same PR, thoughts? @CrossEye

Also:
- `union` looks fine, bacause it already uses `_Set` behind the scenes
- `symmetricDifference` uses `difference` which is improved in this PR, so that is a double win already

I am sure the `with` versions could have their time complexity improved, but I didn't think of a solution to those yet.

Bradcomp:
Thanks for putting in the time to investigate all this @cuchi!

paldepind:
Looks good to me! Great work @cuchi.

vitorebatista:
Great work @cuchi !!

PR #2584, opened 2018-07-09
update lensIndex docs on negative indexes

belgac:
The docs on lensIndex didn't specify that you can use negative indexes to start from the end of the array.

### DIFF
--- a/source/lensIndex.js
+++ b/source/lensIndex.js
@@ -5,7 +5,7 @@ import update from './update';
 
 
 /**
- * Returns a lens whose focus is the specified index.
+ * Returns a lens whose focus is the specified index, a negative index starts from the end of the array.
  *
  * @func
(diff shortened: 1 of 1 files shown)


CrossEye:
I'm not sure if we want to commit to that behavior.

The question is whether we think this is only for arrays, in which case, this is fine.  It happens to be built atop `nth` and `update/adjust`, which are built around arrays.  But it's not clear to me that we want this function to always work only on arrays.  It seems to me that it could also work on finite or infinite list implementations, or on streams, 

Perhaps that doesn't make sense.  Indices on such types are more accidental than fundamental, but they seem well-defined enough.  But negative indices are not so well-defined.  Even for finite lists, they are at least somewhat problematic, if doable.

Any thoughts, @ramda/core?

Issue #2525, opened 2018-05-12
support for iterables and iterators (in general)


jmagaram:
My background is C# and I love LINQ. I also have had great experiences with F#. So what's the best alternative for TypeScript? First I tried some npm modules like 'linq' and 'linq-es2015'. I prefer the dot-less more functional approach and have run into a basic problems, like the GroupBy method exploding. I couldn't get my head around underscore at all. I thought Ramda might be the answer. But it doesn't seem to support iterables - only arrays. See the example code below. I would expect ALL the LINQ-like and F# seq operators - groupBy, scan, minBy, orderBy, zip, pairwise, etc. to be supported out of the box, but no. It would also be helpful to have some basic generators, like R.range(1,100). Am I wrong or is this whole class of functionality a natural fit for Ramda? Is so, why isn't it there yet? LINQ is a critical technology for nearly all C# .net development and it has been around for years. Can't imagine doing .net development without it. But in this JavaScript world I can't find it even though modern browsers with ES6 support it. [f# seq module](https://msdn.microsoft.com/visualfsharpdocs/conceptual/collections.seq-module-%5bfsharp%5d) and [.net linq](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable?view=netframework-4.7.2)

    test("use filter on an array", () => {
        let result = R.filter(isEven, [1, 2, 3, 4]);
        expect(result).toEqual([2, 4]);
    });

    function* numbers() {
        yield 1;
        yield 2;
        yield 3;
        yield 4;
    }

    test("use filter on an iterable", () => {
        // no lazy evaluation here. can't get this to compile without first
        // materializing the iterable into an array.
        let result = R.filter(isEven, Array.from(numbers()));
        expect(result).toEqual([2, 4]);
    });

CrossEye:
> Am I wrong or is this whole class of functionality a natural fit for Ramda? 

No, you're not wrong at all.  There is good reason to want these integrations.  There was some recent discussion in #2508.  But nothing has happened yet.  This would be a large change, with a lot of code.  Also, while @semmel made it clear enough in that issue how we would be able to support iterators, I'm less sure about generators.

You ask why this is not in Ramda so far.  That's because it has never bubbled to the top of the concerns for any of the core team.

[...]

Issue #2508, opened 2018-04-16
nth() (and therefore head()) to support Iterators


ukslim:
I'm aware that support for ImmutableJS has been discussed and deemed out of scope: https://github.com/ramda/ramda/issues/1367

However I've noticed that a lot of Ramda works with Immutables' Seq (etc) because Ramda detects that the List is an Iterable. `map()` and `find`, for example, work fine:

    const double = compose(
         x => x * 2,
         tap(() => console.log('called')
    );

    compose(
        find(equals(8)),
        map(double)
    )(immutables.Range());

... logs 5 times then returns `8`.

However, while this returns 0:

    compose(head, map(double))([0,1,2])

This returns `undefined`:

   compose(head, map(double))(immutables.Range(0,3))

... because `head` calls `nth(0)` and `nth` only caters for arrays and strings.

Any comments or objections to my writing a PR to make `nth` (or possibly just `head`) understand iterables? 

I proved the concept with my own `head`:

```
const symIterator = typeof Symbol !== 'undefined' ? Symbol.iterator : '@@iterator';
const { head: oldHead } = require('ramda');

const head = list => {
  if (list[symIterator] != null) {
    return list[symIterator]().next().value;
  };
  return oldHead(list);
};
```

Of course there are more efficient ways to get the nth member of many objects that happen to be Iterables. In the case of an immutables Seq, `get(n)` -- but `iterable` is standardised and often we use low numbers in `nth`

semmel:
I think it would be awkward if `R.head` works on iterables but not the other three `tail`, `init`, `last` of the quartet as well. 
Admitted the indices are low integers, but without a `ReverseIterator`in JS, it feels wrong even attempting it. Also with `R.nth` with negative index one would need to run to the end of the sequence in the implementation. 

For infinite sequences we'd get infinite loops, but this I'd not consider a big problem, since that's Ramda's principle "Garbage in - garbage out".


While I like JS iterables and think that many of Ramda's List functions are in principle suitable candidates (`contains`, `startsWith`, ...)  to work on iterable sequences I think JS iterables do not qualify as full List collection type - it's just an interface for **looping/iterating** of the sequence.

ukslim:
As it happens, `tail` works on an *immutables* object, because Ramda dispatches to the object's `slice`. Hence with my `head` implementation `head(R.tail(imm.Range(5,Infinity)))` returns 6 as you'd want. Ramda is so close to working with these kinds of list.

[...]

Issue #2337, opened 2017-10-13
Major version bump to 1.0


Siilwyn:
After seeing methods being deprecated and reading a bunch of closed issues on plans for version 1.0 I think this is a case of what Germans would call Hauptversionsnummernerhöhungsangst. The fear of a major bump. As Stephan Bönnemann (who is arguing hard for semver and did talks on the subject), loosely quoted:
> Version numbers should be used to avoid dependency hell... There's no reason for keeping the version number low, expect for marketing or maybe emotional attachment to versions. There's this concept in our mind that the version number has something to say about the project or the process or the stability and all of this is wrong... People believe something incredible existing is happening when a major/breaking version is released... Leave your emotions out of this.

Please consider following semver, it doesn't mean the project is suddenly 'done' or has changed. It just gives users context about the change and reliability that patch and minor changes won't break their code base. Please consider this.

polytypic:
My Partial Lenses library is already at version 12 (there is already version 13 in the works), because I have continued to improve the semantics and increased the major version whenever something has changed in a backward incompatible manner even if the changes are unlikely to affect any users.  (Of course, there have been on average more than 10 releases per major version.)

I highly recommend:
* To release 1.0.0.
* To make feature releases more frequent.
* To make releases with deprecations as early as possible and to add e.g. `console.warn` statements to deprecated functions that are run once in development builds when such functions are called.

CrossEye:
@Siilwyn :
I've been wanting to get to 1.0 for far too long.  I agree that we should do this.

I think this will still take a fair bit of work, given how many open issues we have.  But it's worth doing, and I will try to come up with a reasonable means of getting there in the next few days.

@polytypic:

I agree with your points, except for the `console.warn`.  Ramda scrupulously avoids side-effects.

buzzdecafe:
"Hauptversionsnummernerhöhungsangst" is a *fantastic* word

Siilwyn:
@CrossEye thank you for your detailed & fast reply. I'm looking forward to it!

CrossEye:
I dangled a participle in this:

>  I will try to come up with a reasonable means of getting there in the next few days.

That should really read

> In the next few days I will try to come up with a reasonable means of getting there.

:wink:

CrossEye:
We have started this process by having the core team begin to triage the huge list of open issues/prs.  We'll figure out the next steps when we see how big the remaining list is.

Siilwyn:
[...]

PR #2585, opened 2018-07-11
WIP: Add support for iterables and Set

GingerPlusPlus:
This is a proof of concept, not yet ready to be merged, see "possible issues" below. I didn't want to spend too much time on this PR without the idea getting approved first.

Experimenting with transducers, I wrote lazy `transducer -> iterable -> iterable` function (similar to `toIter` mentioned [here](https://github.com/ramda/ramda/issues/695#issuecomment-66914483)). Then I realized that due to Ramda's transducers support, it allows to transform any iterables using Ramda (https://github.com/ramda/ramda/issues/695#issuecomment-66914234).

I was gonna write an issue about it, until I noticed that small addition to `_dispatchable` is all it takes to add iterator support to every function which supports transducers... so I just did it, to prove it's possible, and to showcase how neat it is.

While I was at it, I went ahead and similarly added support for `Set`, but if you feel that it's too much for 1 PR, feel free to remove or revert that commit.

## Possible issues:

1. [ ] **Naming** may be incorrect or sloppy.
2. [ ] **Correctness** &ndash; I'm not certain that `_transduceIterable` is correct in all cases, it was tricky to write, and I'm fairy new to transducers.
3. [ ] **Code duplication**. for now left on purpose to keep things simple.
4. [ ] **Performance** &ndash; every iterator method would be more performant if implemented without transducers, especially `chain`. However, little additional code required is the main selling point of this implementation.
5. [ ] A lot of **missing tests** &ndash; I didn't want to bother covering every list method before the idea is approved.
6. [ ] **Quirky behaviors** &ndash; `R.some`, when passed a `Set`, returns `Set { true }` or `Set { false }` instead of `boolean`, which is just weird in this context. I won't know all the quirks until all tests are written.

## Possible improvements:

7. [ ] Return some kind of **persistent lazy iterable** instead of generator (https://github.com/ramda/ramda/issues/2508#issuecomment-382050531). It won't be a breaking change, if we document that  functions return an unspecified iterable when iterable is passed.
8. [ ] **Dispatch** to `transduce` (?) method of `obj`, if present.
9. [ ] To address performance and quirky behaviors, introduce optional, 4th argument to `_dispatchable` &ndash; a function which'd (if present) be called if `obj` is present, overriding default handling of iterables.

Related: #2525, #2508, #695, possibly other, closed issues.

I'm awaiting your feedback.


buzzdecafe:
this is pretty nice IMO, I'd love to get this in post 1.0 (if that ever happens :stuck_out_tongue:)

CrossEye:
This is fantastic.  This is one of the things I've wanted to have Ramda do for a long time.

I tried once before and fell down on the job, but I think we need a dedicated effort to get 1.0 out the door.  I would like to do much more like this post 1.0.

GingerPlusPlus:
Glad to hear that you like it :)

> post 1.0

Will 1.0 be the next non-patch release?

If so, I'll improve my code and I'll start adding tests. Since I won't remove or replace a single line of code, and `memoize` is the only deprecated function, it shouldn't cause any non-trivial merge conflicts.

> I want to stop worrying about ES3 compatibility, and embrace ES6+. (https://github.com/ramda/ramda/issues/2337#issuecomment-375403848)

I'll ignore CI for versions of Node older than 6.x, which are all end-of-life, and use ES6 in this PR.

Although, it'd be nice if you decided which versions of Node to support in 1.0 and updated CI to match it;  
I'd definitely rebase my work so that CI status visible on Github would be accurate.

trgwii:
Amazing work!

GingerPlusPlus:
My work here is (hopefully) done, I consider the PR ready to be merged.

I didn't add tests and docs for `reduceBy` and derieved `countBy`, `groupBy` and `indexBy` because I can't wrap my head around `reduceBy`.

@CrossEye / @buzzdecafe, I'm waiting for code review.

Bradcomp:
I haven't done a deep dive into the code yet, but I reviewed the tests. I'd like to see some tests that verify that the functions don't mutate the input when passed an iterable.

For instance, with `all`:

```

const it = iter([2,4,6,8,9,10]);
eq(R.all(even, it), false);
eq(R.all(even, it), false);
```

GingerPlusPlus:
@Bradcomp... why?

There are no such tests anywhere in Ramda, and for that reason, I consider them out of the scope of this PR.

Also, `iter` aka [`iterableOnly`](https://github.com/ramda/ramda/pull/2585/files#diff-37)  is impossible to mutate. Equivalent definition:
`const iterableOnly = R.compose(Object.freeze, R.pickAll([Symbol.iterator]), `[`bound`](https://github.com/GingerPlusPlus/Rextester-bot-v3/blob/ca3d993b26e6e1429643b53db929da35604ecca7/utils/misc.js#L11-L16)`)`.  
It returns a frozen object which only exposes iterable protocol.

Bradcomp:
Here is a counterexample, please let me know if I am missing something: https://goo.gl/xSpHEk

GingerPlusPlus:
Thanks for elaborating, I understand your concern now.

> If I pass in a generator, I would expect it to not be mutated. If we're working with all iterables, we need to make sure that happens.

1. It's impossible to read an iterator without mutating it.
2. Unless stateful iterator is passed, Ramda will never produce and return a stateful iterator.

We could refuse to operate on stateful iterators. I like this idea, I'll code it in tomorrow.

And if we ever decide to support mutable iterators, it'd be a non-breaking change.

CrossEye:
Some day, I'll learn to study the code before responding.  Thanks @Bradcomp for your due diligence!

GingerPlusPlus:
I'll rebase this tomorrow to take care of merge conflicts, get rid of repl, and squash some commits.

GingerPlusPlus:
If a transformer returns `reduced`, should its `'@@transducer/result'` method be called?

semmel:
Since `for ... of` loops are [not supported by IE11](http://kangax.github.io/compat-table/es6/#test-for..of_loops) does this mean Ramda will no longer work in IE11?

Out customers (large corporations, not consumers) still work with IE11 which is afaik still shipped with Windows and gets security updates.

CrossEye:
I'm slowly working my way backward through open issues.  This one hasn't made its way to the top of my list yet, but in any case, yes, it will probably have to wait for 1.0, as it's implemented with `for-of`, which is not available in all the pre-1.0 target environments.

GingerPlusPlus:
Please take a look at #2756, I'd love to rebase this on top of it.

trgwii:
> 
> 
> Since `for ... of` loops are [not supported by IE11](http://kangax.github.io/compat-table/es6/#test-for..of_loops) does this mean Ramda will no longer work in IE11?
> 
> Our customers (large corporations, not consumers) still work with IE11 which is afaik still shipped with Windows and gets security updates.

I believe it should be possible to keep supporting IE11 through transpilation, but this would need some further investigation... I don't see a reason why for..of couldn't be transpiled, so there is still hope for the foreseeable future!

semmel:
@trgwii 

> I believe it should be possible to keep supporting IE11 through transpilation, ...

Yes fair enough, I could ship a transpiled `ramda.js` to IE 11, no problem there. 

I just want to note that while `for .. of` is not needed for iterating over iterables, however it's use is certainly left a will of the implementor. But in that case I cannot see why Ramda could not go all ES2019 with arrow (`=>`) functions and stuff. And thus would be a different library with many functions gone. I think that decision is better saved for a major version bump. Rather then quietly introducing changes in the runtime requirements.

> ... existing versions will continue to function as they have always done  ...

Letting source code rot with an old library version just makes it harder to update one day then IE 11 is finally history. That's not an option. 

Also you cannot maintain a zoo of library versions in a medium size code base: What about a *shared component* which depends on Ramda? Must I than implement, test and document two versions of it, one with Ramda pre 1.0 and one with post 1.0? That would kill productivity. So also not an option.

trgwii:
> I think that decision is better saved for a major version bump. Rather then quietly introducing changes in the runtime requirements.

@semmel Ramda is very careful about these things in my experience, just look at [0.26.1 release notes](https://github.com/ramda/ramda/issues/2723). I'm certain no changes to supported runtimes will happen until major bump.

GingerPlusPlus:
I can rebase this to make it mergable, but I'm wondering if implicitly educing iterables is a good idea in the first place. Perhaps exposing `R.educe` to be explicitly would be better?

GingerPlusPlus:
Also, there's a possible optimization: `_educe(g, _educe(f, in)) = _educe(o(f, g), in)`, should I add it?

CrossEye:
@GingerPluPlus:  I very much like the idea here.  After the holidays, I'm going to try to get a discussion going between anyone intertested in how to get 1.0 out the door, and then I'd be very interested in including this.

Also, what is the implication of using the word "educe" here?  I know the English definitions, but I'm not making the connection with this function.

GingerPlusPlus:
Thanks, but reconsider -- this may hinder adding `Map` support. No rush.

At this point, I'll believe in 1.0 only after I see it done :P

"Educe" is the name picked by Rich Hickey, see https://youtu.be/4KqUvG8HPYo?t=2437

CrossEye:
Closing until after `v1.0`

xgbuils (approved):

PR #2588, opened 2018-07-14
Add file extensions to imports and exports in `source/`

### LINKED ISSUE #2587: Cannot import ramda in the browser with native modules
### Test case

Create a html document with the following contents in a directory where ramda is installed via npm.

```
<script type="module">
  import * as Ramda from './node_modules/ramda/es/index.js'
</script>
```

Open the document through a webserver in in latest Chrome, Firefox, Edge or Safari.

### Expected result

Ramda functions are imported and can be used inside the script.

### Actual result

The browser cannot find the files listed in `node_modules/ramda/es/index.js` (eg. `GET http://127.0.0.1:8080/node_modules/ramda/es/F 404 (Not Found)`).

### Error diagnosis

Valid module paths *must* contain the file extension. For example, the first import in `index.js` should say `export { default as F } from './F.js';`. Note the extra `.js` at the end that is currently absent.

### The way forward

I suggest specifying the full paths to files including file extension whereever `import`/`export` is used in the code base. The build system used in this project probably supports this already. This way, developers can import Ramda via native ES2015 modules directly in the browser, a feature that is supported by all modern browsers.

If you agree this should be worked on I'm certainly willing to prepare a pull request.

Thank you for your consideration and for your work on Ramda.

modiho:
This enables a developer to load Ramda via import statements directly in the
browser without bundling/transpilation.

Before, imports in the Ramda source only included the file name without
extension, eg. `import curryN from './curryN'`. Most bundlers will
assume a default extension and find the file regardless. In the browser,
the exact file path that was specified in the import statement will be
requested, and the file can usually only be served by the webserver or
located on the file system if the file path matches exactly. Thus, the
complete file path must be used in import statements to be able to load
Ramda via import statements without some kind of bundling/transpilation
process.

Fixes #2587.

### Method

Applied [this codemod](https://astexplorer.net/#/gist/1c3fe56c12f839a5d5eca10084b39fc1/073cf55c223810f36d098fc7d5e50c8a70e68abf) with [jscodeshift](https://github.com/facebook/jscodeshift) on all files in `source/` because I don't know how to use `sed`.

I was able to load the `ramda/es/index.js` as described in the original issue without problems.


CrossEye:
This looks good to me.  :herb:

@ramda/core: Does anyone object?  See #2587 for the rationale.

GingerPlusPlus:
I have a concern, does this work with Node 10's modules support?

GingerPlusPlus:
I just checked, and I can't get it to work (I haven't used es6 modules in Node before tho).

I'd expect it to work if `.mjs` extension was used by all of the files instead.

Also, `eslint` rule could probably be put in place to ensure that imports use desired extension, whatever it'll be: [`no-restricted-imports`](https://eslint.org/docs/rules/no-restricted-imports).

GingerPlusPlus:
It's probably not a huge deal for es6 module users on Node, since they can import the commonjs version, but I think if this is touched, it should be resolved properly, for all environments.

GingerPlusPlus:
**TL;DR**:  I give up. Feel free to proceed.

<hr />

Disclaimer: I have never used babel directly before, I may be completely wrong.

I spent the last 2 hours looking for an alternative. I give up.

`babel-cli` will apparently religiously convert every file to `.js` extension, but it won't update the `import` statements to match the change (why? why? why?).

What can be done about it? My answer is, nothing:

1. No way to update the extension: I couldn't get [`import-rename`](https://npmjs.com/babel-plugin-import-rename) to transform re-exports in `index`.
2. No way to preserve input's extension: https://github.com/babel/babel/pull/6221, doesn't work for me on latest `babel-cli`.
2. No way to change outputted extension: https://github.com/babel/babel/issues/6222. 

I see no way, with `babel-cli` alone, to have `.mjs` _anywhere_.

Feel free to proceed. Sorry for delaying the merge.

GingerPlusPlus:
I successfully managed to use `jscodeshift` for build process, but it's noticeably slower than babel alone; I'll try modifying the `babel-plugin-import-rename`.

modiho:
Closed in favor of #2591

PR #2589, opened 2018-07-15
[CI] Drop support for Node older than 8.x

GingerPlusPlus:


buzzdecafe:
:cow2:

mattgrande:
It looks like Node 8 is still in Active support until April 2019; Why would you want to remove it from CIs now?

https://github.com/nodejs/Release

GingerPlusPlus:
@mattgrande I didn't remove Node 8.

mattgrande:
Ugh, sorry, I read "older than 8.x" in the title and read it as "get rid of 8.x." Ignore me :)

CrossEye:
:herb:

Any objections, @ramda/core ?

davidchambers:
Do we have browser tests? If not, testing against old versions of Node seems like a good idea.

Bradcomp:
I too am hesitant to remove old node tests without browser tests, as they serve as a mediocre proxy for that. I also don't think we should remove Node 6 until it's off LTS. Somebody should look into getting browser tests set up though so that we can get rid of older node versions without worry.

I think it would behoove us to determine what browsers and environments we want to support and then document that.

buzzdecafe:
> Somebody should look into getting browser tests set up

NOT IT

GingerPlusPlus:
Please review #2599, in which I fixed browser tests.

GingerPlusPlus:
Can this be merged, now that browser tests are up?

Bradcomp:
So we aren't supporting maintenance LTS now, only active?

GingerPlusPlus:
Yes.

Bradcomp:
I think that's really problematic. Browsers are weird in that we don't know how long they have to be supported, so it makes sense to determine an arbitrary cut off. With Node, we have a clear roadmap for what versions are supported and when they die. 

It's really straightforward to know which versions of Node are currently supported. That includes The most recent, the active LTS, and the maintenance LTS. I see no reason we shouldn't support all three.

PR #2593, opened 2018-07-17
Make R.length operate on Set and Map

GingerPlusPlus:
I think making it useful is better than removing it (https://github.com/ramda/ramda/pull/2450#issuecomment-361073664).

I'd like to know why the heck `Set` and `Map` use `.size` instead of `.length` in the first place.

Related to #1055.

### DIFF
--- a/source/length.js
+++ b/source/length.js
@@ -3,21 +3,33 @@ import _isNumber from './internal/_isNumber';
 
 
 /**
- * Returns the number of elements in the array by returning `list.length`.
+ * Returns the number of elements in the array, set or map
+ * by returning `collection.length` or `collection.size`.
  *
  * @func
  * @memberOf R
  * @since v0.3.0
  * @category List
  * @sig [a] -> Number
- * @param {Array} list The array to inspect.
- * @return {Number} The length of the array.
+ * @param {Array|Set|Map} collection The collection to inspect.
+ * @return {Number} The length of the collection.
  * @example
  *
  *      R.length([]); //=> 0
  *      R.length([1, 2, 3]); //=> 3
+ *      R.length(new Set(['x', 'y', 'z'])); //=> 3
+ *      R.length(new Map([['k1', 'v1'], ['k2', 'v2']])); //=> 2
  */
 var length = _curry1(function length(list) {
-  return list != null && _isNumber(list.length) ? list.length : NaN;
+  if (list == null) {
+    return NaN;
+  }
+  if (_isNumber(list.length)) {
+    return list.length;
+  }
+  if (_isNumber(list.size)) {
+    return list.size;
+  }
+  return NaN;
 });
 export default length;
--- a/test/length.js
+++ b/test/length.js
@@ -13,6 +13,18 @@ describe('length', function() {
     eq(R.length('xyz'), 3);
   });
 
+  it('returns the size of a es6 set', function() {
+    if (typeof Set !== 'function') { return; }
+    eq(R.length(new Set()), 0);
+    eq(R.length(new Set(['x', 'y', 'z'])), 3);
+  });
+
+  it('returns the size of a es6 map', function() {
+    if (typeof Map !== 'function') { return; }
+    eq(R.length(new Map()), 0);
+    eq(R.length(new Map([['k1', 'v1'], ['k2', 'v2']])), 2);
+  });
+
   it('returns the length of a function', function() {
     eq(R.length(function() {}), 0);
     eq(R.length(function(x, y, z) { return z; }), 3);


GingerPlusPlus:
@Bradcomp added examples.

CrossEye:
Why not also delegate?  If `myContainer` exposes a `length` method, why not call it?

GingerPlusPlus:
Do you know any library that has an interface like that? I would dispatch only to stuff that people actually use.

Bradcomp:
Yesterday I learned that often ‘size’ is used when the operation can be done in constant time, while ‘length’ is used for the linear time equivalent. I don’t think this is the case in JS, but it might be the origin of the difference.

CrossEye:
@GingerPlusPlus: I haven't gone looking, but I know that I've implemented either `size` or `length` methods for a number of different types over the years.

@Bradcomp: I'd forgotten that.  It seems like long-ago lore that might or might not have a basis in the truth.

Bradcomp:
@CrossEye It's at least the truth in Elixir, which is where I learned of the difference.

CrossEye:
@Bradcomp: It's a sensible distinction, and one I've heard somewhere before.  I don't know how close it is to universal.  But I like it.

GingerPlusPlus:
> If `myContainer` exposes a `length` method, why not call it?

Even if it is a good idea (I'm not convinced), it doesn't serve the goal of this PR, which is to add `Set` and `Map` support to `R.length`.

source/length.js
```diff
@@ -3,21 +3,33 @@ import _isNumber from './internal/_isNumber';
...
+ * by returning `collection.length` or `collection.size`.
  *
  * @func
  * @memberOf R
  * @since v0.3.0
  * @category List
  * @sig [a] -> Number
- * @param {Array} list The array to inspect.
- * @return {Number} The length of the array.
+ * @param {Array|Set|Map} collection The collection to inspect.
```
CrossEye: Our docs are often aspirational rather than pedantically correct, but should we specifically mention `Array|Set|Map` here?  The implementation has little to do with these types but only with having either a `length` or a `size` property.

Of course I'm not sure of the simplest way to document that.  But is there a reason to specifically list these types?

source/length.js
```diff
@@ -3,21 +3,33 @@ import _isNumber from './internal/_isNumber';
...
+  if (list == null) {
+    return NaN;
+  }
+  if (_isNumber(list.length)) {
+    return list.length;
+  }
+  if (_isNumber(list.size)) {
+    return list.size;
+  }
+  return NaN;
```
CrossEye: Should we also handle the `Object` type?  How much difference is there between saying that a `Map` has a size and saying that an `Object` has one?

GingerPlusPlus: I know that this can be done well, but I'm not sure if it should be done.

Reading a property is `O(1)`, iterating over object is `O(n)`.

In the interest of not getting this PR stuck in a limbo, I suggest to not do it in this PR.
If someone thinks this is a good idea, they can open a new issue or PR just for that.

source/length.js
GingerPlusPlus: Well, those are all the builtin type that the function in its current form intentionally supports.

> I don't think we should call out `function` specifically. (https://github.com/ramda/ramda/pull/2450#issuecomment-361025289).

If `Object` support was added, the param type could be `Collection`.

That's the best I came up with. Do you have a better param type in mind?

CrossEye: I think the discussion that we never followed up on there might be better.  This should simply be removed... although there definitely is something to be said for making it work with `Map` **and** `Set` as well as current `Array` and `Function` (and `Int16Array`, etc., and `AsyncFunction`, and who knows how many others?)

The point, though, is that with the plethora of types that this matches, even built-in ones, should we call out those three?

source/length.js
CrossEye: I don't know how the engines implement this, but there is a non-trivial chance that `Object.keys` does not involve iterating anything. 

>  In the interest of not getting this PR stuck in a limbo,...

Unfortunately, it seems a great number of issues and PRs are trapped there.  But if we're going to change this function, rather than just remove it, I'd prefer to do it once.

source/length.js
GingerPlusPlus: > I think the discussion that we never followed up on there might be better. This should simply be removed...

I'm not sure what you mean, please be more clear.

> `Int16Array`

Right, didn't know typed arrays have a length too, never used them.

CrossEye: I was simply saying that the discussion there ended up with something like a consensus on removing `length` altogether.

GingerPlusPlus:
Length can also be derived from `reduce`, perhaps trying to support every type imaginable is futile...

I changed my mind. I don't mind `length` getting deprecated, as long as `keys`, `values`, `toPairs` and `fromParis` also get deprecated instead of receiving `Map` support.

CrossEye:
@GingerPlusPlus:

Would you deprecate them simply *because* they wouldn't support `Map`s?  There are many people still using objects.

GingerPlusPlus:
I'd deprecate them because without `Map` support they have no advantage over builtin `Object.keys`, `Object.values`, `Object.entries` and `Object.fromEntries`.

CrossEye:
@GingerPlusPlus: I'll start a conversation soon about 1.0 and beyond.  When we're no longer supporting older ES environments, these would definitely be good targets for the chopping block.

PR #2595, opened 2018-07-18
clearer example of madd3 for lift/liftN

ted-ted:
Per CrossEye's comment to this stackoverflow answer: https://stackoverflow.com/a/44306719:

> The example from the docs is terrible for understanding. One like this would probably help: madd3([100, 200], [30, 40, 50], [6, 7]); //=> [136, 137, 146, 147, 156, 157, 236, 237, 246, 247, 256, 257]. (And a PR to fix the documentation would be welcome.) – Scott Sauyet Jun 4 '17 at 1:07

This really helped me understand how lift() behaves for arrays.


ted-ted:
Oops; nevermind. There's already other activity in this space: #2190 and #2467.

PR #2599, opened 2018-07-20
Fix and simplify testing

GingerPlusPlus:
Uses babel as a plugin to avoid  having to create commonjs build on disk before running tests.

Also, fixes browser tests (https://github.com/ramda/ramda/pull/2589#issuecomment-405263241), so closes #2597.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,8 +76,9 @@
     "babel-plugin-annotate-pure-calls": "^0.2.0",
     "babel-register": "^6.26.0",
     "babel-types": "^6.26.0",
+    "babelify": "^8.0.0",
     "benchmark": "~1.0.0",
-    "browserify": "10.x.x",
+    "browserify": "https://api.github.com/repos/browserify/browserify/tarball/9ff7c55cc67a7ddbc64f8e7270bcd75fcc72ce2f",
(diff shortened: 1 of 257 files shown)


GingerPlusPlus:
CI fails on Node 8 with npm 5.6.0 due to https://github.com/npm/npm/issues/16726.

CrossEye:
I would definitely like to see browser testing back.  And I have no initial qualms with this.

Elsewhere we are discussing what versions of Node to support.  I think that will drive whether we can do this or not (although possibly different browserify options would make it pass even in older Node?)

GingerPlusPlus:
The discussion is, as far as I'm aware of, whether to support Node 6, which is in maintenance (not active) lts. 

This code fails only on 0.12, which is end-of-life since 2016-12-31.

CrossEye:
Anyone have objections, @ramda/core ?

Bradcomp:
:bowtie:  :shipit:

GingerPlusPlus:
#2589 should be merged first, otherwise every commit will fail CI on Node 0.12.

GingerPlusPlus:
I downgraded Mocha, so that it works on Node 0.12, and this PR can be merged now.

PR #2601, opened 2018-07-22
typos

bliotti:
minor typos


GingerPlusPlus:
"its" is correct here ("arguments of the function", not "it is arguments"). And "needn't" is definitely not a typo.

CrossEye:
As @GingerPlusPlus says, "its" is correct.  While "needn't" is not a typos, I do prefer "need not" here.  I think it's easier to understand.

coinGuyBri:
oops, "its" is correct. Sorry!

PR #2602, opened 2018-07-22
Remove dist/ from git

GingerPlusPlus:
IMO, it should only be published to npm, on each release.
Why was it committed in the first place?


git-in-my-anus:
These files are used in bower distribution, which is fetched directly from github. See the bower config https://github.com/ramda/ramda/blob/master/bower.json

They can't be deleted without deprecating installation through bower first.

GingerPlusPlus:
Damn, that's a shame :(

[Bower hooks](https://bower.io/docs/config/#hooks) could be used to build the `dist/` consumer-side, even tho the docs explicitly mention they're not meant for this.

OTOH, Bower itself recommends people to move on...

Thanks for the answer.

CrossEye:
It's been committed since the very early days, when Ramda was used by only its two initial developers, usually in the browser, long before there was a thought of pushing to NPM, long before we published to CDNs.  We wanted to be able to use it right from the repo.  Of course it wasn't exactly the `dist` folder you see now, but the library as a single-file UMD module.  That was where we developed it.

Even ignoring the bower issue, I'm a little loathe to remove it at the moment.  I haven't checked lately, but last I remember we couldn't build on Windows.  I don't want the git repo to be useful only to those with a *nix OS.  If we're past those, then perhaps I wouldn't mind too much... if we can get past the bower problem.

GingerPlusPlus:
I'm currently on Windows 10, and I have no issues with building `dist/ramda.js` even from `cmd`.

I think that `dist/ramda.min.js` cannot be built anywhere, due to `uglify`'s lack of support for "modern" JS:
`[!] (uglify plugin) Error: Error transforming bundle with 'uglify' plugin: Unexpected token: keyword (const)`

...but, I think that's a completely separate issue.

CrossEye:
> I'm currently on Windows 10, and I have no issues with building `dist/ramda.js` even from `cmd`. 

OK, glad to hear it. 

> I think that `dist/ramda.min.js` cannot be built anywhere, due to `uglify`'s lack of support for "modern" JS.

That had to be since our last version.  I wonder why...

GingerPlusPlus:
Faster option for Bower support would be to use the hook to download `dist/` from a CDN.

And maybe building locally could be used as a fallback.

GingerPlusPlus:
Faster option for Bower support would be to use the hook to download `dist/` from a CDN.

And maybe building locally could be used as a fallback.

CrossEye:
> <ins>Also, support for Bower could be dropped in 1.0</ins>

Yes!

PR #2603, opened 2018-07-22
Improve R.without and R.intersection time complexity

cuchi:
I am using the internal `_Set` in both functions to drop the time complexity from O(n²) to O(n).
That's the same way the `R.difference` is improved in #2581.

Before/after (in milliseconds):
```
-- R.without
Before:
  10k [size 10] arrays: 65.593378
  10k [size 100] arrays: 437.021462
  1k [size 1000] arrays: 1186.243514
  100 [size 10000] arrays: 9127.722465
After:
  10k [size 10] arrays: 37.329974
  10k [size 100] arrays: 102.189432
  1k [size 1000] arrays: 93.958924
  100 [size 10000] arrays: 114.788804

-- R.intersection
Before:
  10k [size 10] arrays: 86.015134
  10k [size 100] arrays: 502.037411
  1k [size 1000] arrays: 1212.673359
  100 [size 10000] arrays: 9114.332054
After:
  10k [size 10] arrays: 45.109575
  10k [size 100] arrays: 175.27189
  1k [size 1000] arrays: 165.56708
  100 [size 10000] arrays: 199.455852
```

### DIFF
--- a/source/intersection.js
+++ b/source/intersection.js
@@ -22,21 +22,12 @@ import uniq from './uniq.js';
  *      R.intersection([1,2,3,4], [7,6,5,4,3]); //=> [4, 3]
  */
 var intersection = _curry2(function intersection(list1, list2) {
-  var lookupList, filteredList;
   var toKeep = new _Set();
 
-  if (list1.length > list2.length) {
-    lookupList = list1;
-    filteredList = list2;
-  } else {
-    lookupList = list2;
-    filteredList = list1;
+  for (var i = 0; i < list1.length; i += 1) {
+    toKeep.add(list1[i]);
   }
 
-  for (var i = 0; i < lookupList.length; i += 1) {
-    toKeep.add(lookupList[i]);
-  }
-
-  return uniq(_filter(toKeep.has.bind(toKeep), filteredList));
+  return uniq(_filter(toKeep.has.bind(toKeep), list2));
 });
 export default intersection;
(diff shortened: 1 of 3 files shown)


source/intersection.js
```diff
@@ -24,13 +23,20 @@ import uniq from './uniq';
...
   }
-  return uniq(_filter(flip(_includes)(lookupList), filteredList));
+
+  for (var i = 0; i < lookupList.length; i += 1) {
+    toKeep.add(lookupList[i]);
+  }
+
+  return uniq(_filter(toKeep.has.bind(toKeep), filteredList));
 });
 export default intersection;
```
CrossEye: I think if we're adding this optimization, then we should consider undoing #1597, which has remained controversial since it was created.

cuchi:
I also committed in two new test cases, considering bigger lists on any side. This may catch a bug if someone ends up refactoring or optimizing this code.

CrossEye:
@cuchi:  Ramda fell way behind in handling its open issues.  This one was one of many that fell through the cracks.  But it seems like a very good idea.  Would you be interested in fixing the conflicts in this one that have developed?

customcommander:
@CrossEye Rebased against the master branch and fixed all the conflicts but this will fail for the same reason master is failing. Will be fixed once #3224 is merged.

customcommander:
@CrossEye All issues fixed. This should be ready to go if we still want this.

cuchi:
@CrossEye @customcommander
Hello there! I'm happy to see the project is alive :heart: that's an awesome effort!

About the changes, they seem to be ok after the conflict fixes, from my POV.

CrossEye (approved):
LGTM

:herb:

PR #2605, opened 2018-07-23
Set function length in _arity using Object.defineProperty

### LINKED ISSUE #2459: dead code elimination cause problems with _arity

wojpawlik:
Fixes #2459.
Alternative to #2460, closes #2460.

`fn` is always a function created by Ramda that is not reused for anything else, so it should be safe to mutate it.

### DIFF
--- a/source/internal/_arity.js
+++ b/source/internal/_arity.js
@@ -1,17 +1,7 @@
 export default function _arity(n, fn) {
-  /* eslint-disable no-unused-vars */
-  switch (n) {
-    case 0: return function() { return fn.apply(this, arguments); };
-    case 1: return function(a0) { return fn.apply(this, arguments); };
-    case 2: return function(a0, a1) { return fn.apply(this, arguments); };
-    case 3: return function(a0, a1, a2) { return fn.apply(this, arguments); };
-    case 4: return function(a0, a1, a2, a3) { return fn.apply(this, arguments); };
-    case 5: return function(a0, a1, a2, a3, a4) { return fn.apply(this, arguments); };
-    case 6: return function(a0, a1, a2, a3, a4, a5) { return fn.apply(this, arguments); };
-    case 7: return function(a0, a1, a2, a3, a4, a5, a6) { return fn.apply(this, arguments); };
-    case 8: return function(a0, a1, a2, a3, a4, a5, a6, a7) { return fn.apply(this, arguments); };
-    case 9: return function(a0, a1, a2, a3, a4, a5, a6, a7, a8) { return fn.apply(this, arguments); };
-    case 10: return function(a0, a1, a2, a3, a4, a5, a6, a7, a8, a9) { return fn.apply(this, arguments); };
-    default: throw new Error('First argument to _arity must be a non-negative integer no greater than ten');
-  }
+  Object.defineProperty(fn, 'length', {
+    value: n,
+    configurable: true
+  });
+  return fn;
 }


GingerPlusPlus:
Even IE supports `Object.defineProperty` since IE9. What is the oldest thing you wish to support?

nfantone:
@CrossEye I'd like to integrate changes proposed here in #2782. Would this be a good idea at this point in time?

nfantone:
@GingerPlusPlus Just realised that this new implementation contradicts `nAry` definition. From the documentation:

> Any extraneous parameters will not be passed to the supplied function.

GingerPlusPlus:
No it doesn't. Currently `nAry` doesn't use `_arity`. Tests pass on Node with `Object.defineProperty`.

I think it's a great idea to `git merge -X theirs` this into #2782.

nfantone:
> Currently `nAry` doesn't use `_arity`. Tests pass on Node with `Object.defineProperty`.

@GingerPlusPlus I know it doesn't. But they're are currently a copy-paste of one another, word by word. Would be only reasonable, specially to newcomers, to think that expectations are the same for both of their outcomes. This would break that implicit contract.

nfantone:
@GingerPlusPlus Not to mention there's also this: https://github.com/ramda/ramda/pull/2611/files#diff-c48e854ef7be86688178177f08744c86R36.

GingerPlusPlus:
After the change you mentioned, `nAry` _uses_ `_arity`.

They're **not** the same, and thus, they behave differently.

I think you're confused. What would you suggest to do about the two?

> newcomers

`_arity` is internal. Newcomers won't even see it.

nfantone:
> They're **not** the same, and thus, they behave differently.

Except they are and they do. Apart from the fact that `nAry` is curried and `_arity` is not, currently they are the same exact thing, with a different name.

I also don't think I'm confused. What about? 

> Newcomers won't even see it.

I was referring to contributors. New and old.

CrossEye:
@nfantone:

They are different.  Look for instance at these two lines:

```js
// _arity.js:
case 3: return function(a0, a1, a2) { return fn.apply(this, arguments); };
// nAry.js:
case 3: return function(a0, a1, a2) { return fn.call(this, a0, a1, a2); };
```

They are designed for different purposes.  The internal `_arity` function is designed to make Ramda a good neighbor.  When Ramda converts a function through partial application, currying, flipping, whatever, we want the resulting function to properly report its arity.  But we don't want to cause issues with functions that themselves don't properly report their arities, so we pass through everything supplied.

`nAry`, on the other hand, is a public function; it's goal is to exactly fix the arity of the function supplied.  So if you need a binary function, you can supply one that takes zero, one, two, three, or forty-two arguments, and call `nAry(2, fn)` to get back one that takes exactly two.  It's really just an obvious extension of `unary` and `binary`.  

The main difference is that `nAry` fixes the actual count of arguments supplied, while `_arity` fixes only the *announced* length.

While for reasons mentioned above, I want to hold off on this one until after v1.0, I don't have the same concerns about #2782.  (I just haven't reviewed the conflicts that @GingerPlusPlus noted there yet.)

nfantone:
@CrossEye Ah, you're right. I see my error now. Completely missed the difference. Thanks for pointing it out! And apologies to you and @GingerPlusPlus for arguing.

CrossEye:
Closing until after `v1.0`

PR #2606, opened 2018-07-23
Add function cloning

GingerPlusPlus:
I don't know if this is a good idea. I did this because I could.

### DIFF
--- a/source/internal/_clone.js
+++ b/source/internal/_clone.js
@@ -1,3 +1,4 @@
+import _cloneFunction from './_cloneFunction';
 import _cloneRegExp from './_cloneRegExp';
 import type from '../type';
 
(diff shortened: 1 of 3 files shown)


CrossEye:
That's fascinating: a entire clone in terms of properties, and referentially transparent, but still not a simple reference, nor with the same debugging or performance profile.

It is still not a complete clone, as those specifics are not the only properties that a given function might have, nor should `equals(fn), clone(fn)` return `true`, but it is pretty nice.

What do people think, @ramda/core?

GingerPlusPlus:
> nor should equals(fn, clone(fn)) return true

I don't think that functions should be exempt from `equals(clone(anything), anything)`.

Function could be considered equal if their `.toString()` is equal (maybe `name` too) and their properties are equal.

> Is there a practical reason for it, though?

I could imagine wanting a function with the same behavior, but different properties.

I could imagine wanting to set a property on a function, but not wanting to mutate it.

`fn.bind()` is fine, but has more limitations: https://stackoverflow.com/a/6772648/3821804.

PR #2607, opened 2018-07-23
In _clone, use Map to store cloned values

wojpawlik:


CrossEye:
A definite clean-up.  Anyone see a problem with this?  @ramda/core

Bradcomp:
The native `Map` uses reference equality instead of value equality. Will that cause an issue for this case?

GingerPlusPlus:
No, it won't. Original code uses "reference equality" (`===`) as well.

For a good reason: 2 equal but not `identical` values will be equal but not `identical` in the clone as well, and 2 `identical` values will remain `identical`:

```js
const baz = {}
const foo = [{}, baz, baz]
assert(foo[0] !== foo[1])
assert(foo[1] === foo[2])
const bar = R.clone(foo)
assert(bar[0] !== bar[1])
assert(bar[1] === bar[2])
```

Bradcomp:
Thanks for humoring me @GingerPlusPlus!

CrossEye:
Using `Map` is a no-go before `v1.0`.  Will look again after we get that out.

PR #2610, opened 2018-07-24
Fix build:umd:min

GingerPlusPlus:
https://github.com/ramda/ramda/pull/2602#issuecomment-406889696

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "ramda": "0.17.x",
     "rimraf": "~2.3.2",
     "rollup": "^0.50.0",
-    "rollup-plugin-uglify": "^2.0.1",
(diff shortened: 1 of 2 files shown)


GingerPlusPlus:
I don't, because uglify doesn't work on current codebase :P

I could compare last release, instead.

I wouldn't expect any difference tho, `terser` is a fork or `uglify`.

GingerPlusPlus:
What exactly does "standalone" :bowtie: mean? I've seen 🐄 and 🌿 used by other members, too.

Bradcomp:
Haha, each member uses their own emoji for a simple sign off. All three of those are the equivalent of LGTM!

PR #2611, opened 2018-07-25
Simplify constructN and nAry

wojpawlik:


CrossEye:
Again, this uses techniques not available in old environments.  This should come post-1.0.

PR #2612, opened 2018-07-25
Drop Bower support

wojpawlik:
https://github.com/ramda/ramda/pull/2602#issuecomment-407504807


GingerPlusPlus:
@CrossEye this in 1.0 or 2.0?

CrossEye:
We still need to have that discussion.  Later this evening I'm planning on trying to republish the docs and figure out a framework for how we make those decisions, and maybe finally start on a roadmap.

customcommander:
@CrossEye: could we consider officially dropping support for Bower? AFAIK this thing is dead; even the maintainers recommend to move away from it.


---

As for recording this kind of decisions, I've alluded to ADRs in a comment. This is the tool that I use: https://github.com/npryce/adr-tools. The decisions are plain old Markdown files that live in the repository.

customcommander:
Although perhaps we should do so by following this:
https://bower.io/blog/2017/how-to-drop-bower-support/

wojpawlik:
Instead of outright removing support, perhaps https://github.com/ramda/ramda/blob/c2cafc2a42fc9d28fa25e17c3df54d67481fbba7/bower.json#L3 could point to ESM `source/index.js`? Build artifacts (`dist/`) wouldn't have to be committed anymore.

customcommander:
This does sound like a reasonable suggestion to me; I'm just not 100% sure about its ramifications (if any) but this is definitely noted. My preference would be to drop support for Bower nonetheless.

Issue #1658, opened 2016-02-20
Easier way to create transducers 


asaf-romano:
See issue #1656 for background.

My experience with transducers so far shows that they are worthless if the accumulator transducer does not avoid creating intermediate objects by itself. This means that something like ``transduce(compose(...), mergeWith(...), {})`` **won't** perform well on a large data sets. Problem is, the transducers protocol, while very powerful, asks for quite an ugly object, that is tiresome to write.

It would thus be nice to have something like Java's ``Collector.of``, which allows initiating Java's analogous of accumulator-transducer, given its building up blocks. See: 
https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collector.html#of-java.util.function.Supplier-java.util.function.BiConsumer-java.util.function.BinaryOperator-java.util.function.Function-java.util.stream.Collector.Characteristics...-

(Ignore combiner and Characteristics)

CrossEye:
I'm trying to make sense of what that would mean in JS.  Do you have a sense of what one might supply to a Ramda function to use something like this?  There doesn't seem to be a simple translation of the Java types.  `Supplier :: Unit -> a` ? `BiConsumer :: b -> a -> Unit` ? For side effects?`, but with the more sensible `BinaryOperator :: a -> a -> a` that you want us to ignore.  I'm just not getting it.

asaf-romano:
It'd be as simple as ``thatNewFunctionName(init, step, result)``. So, for instance:
``thatNewFunctionName(Array, (acc, x) => { acc.push(x); return acc }, uniq)``

This isn't so different from the current object you pass in, but there's a better chance people will bother.

CrossEye:
How much of this do you think we could embed in internal calls?  I really don't like the idea of exposing or asking users to supply mutable structures.  I don't at all mind using them internally.

Or perhaps even if we do offer the general API, could we create a simpler gloss for some types, especially lists?

asaf-romano:
At least in Java (sorry, don't really know Clojure) the whole idea behind collectors is that you sandbox the mutations the way I did in the PR today, so while the mutations are technically mutable, they have no effect on outside-usage of the (finalized) data structure. Now, it's a little hard to demonstrate that in plain JS without in-your-face array.push kind of mutations, because there's no "builder" concept that emphasizes the fact the object is _intermediate by its nature_.

[...]

PR #2613, opened 2018-07-25
Remove _arrayReduce

GingerPlusPlus:


CrossEye:
This looks good.  Did you do any tests for whether this affects performance?

GingerPlusPlus:
I didn't.

GingerPlusPlus:
Benchmark

GingerPlusPlus:
Tests fail in 0.12, because array was not iterable then. Wait with merging until 0.12 compat is not a concern.

GingerPlusPlus:
I haven't noticed that benchmark tests build code, not `source` (PR incoming).

It does significantly degrade performance.

New bench:

```
~/JS/ramda (master *$%=)
$ npm run bench

> ramda@0.26.1 bench D:\Wojciech\JS\ramda
> node scripts/benchRunner

│ reduce                 │ Hertz                  │ Margin of Error        │
│ reduce(add, 0, nums)   │ 439,735                │ 1.39%                  │
│ reduce(add, 0)(nums)   │ 285,705                │ 2.27%                  │
│ reduceAdd(nums)        │ 485,393                │ 1.29%                  │
│ native reduce          │ 4,674,637              │ 0.41%                  │

~/JS/ramda (no-array-reduce *$%=)
$ npm run bench

> ramda@0.25.0 bench D:\Wojciech\JS\ramda
> node scripts/benchRunner

│ reduce                 │ Hertz                  │ Margin of Error        │
│ reduce(add, 0, nums)   │ 312,213                │ 1.12%                  │
│ reduce(add, 0)(nums)   │ 236,533                │ 2.68%                  │
│ reduceAdd(nums)        │ 298,260                │ 1.68%                  │
│ native reduce          │ 4,761,831              │ 0.27%                  │
```

@CrossEye: by supporting `reduced` only for transducers, not for functions, `_reduce` could achieve near-native speeds, would you be fine with that change? Current code already doesn't support `reduced` for fantasyland foldables.

CrossEye:
I care little about `reduced`, honestly.  I would have no issues with that.  Anyone else?  @ramda/core?

scott-christopher:
We should be able to emulate `R.reduced` with `R.thunkify` as a recipe for the cookbook, e.g.

```js
> has3 = R.reduceRight(R.thunkify((x, acc) => { console.log(x); return x == 3 ? true : acc() }), R.F)
> has3([1,2,3,4,5,6])()
1
2
3
true
> has3([1,2,4,5,6])()
1
2
4
5
6
false
```

Bradcomp:
I think I have only used reduced a few times. It's one of those that isn't _that_ useful but can be crucial in a pinch (though as @scott-christopher shows, it can be emulated), but its main value is as an answer for people who complain bout the performance of reduce because you can't break out early.

In addition, we also have `reduceWhile` to break during a reduction.

GingerPlusPlus:
I just benched all the ways in which I could imagine changing `(function, array)` paths of `_reduce`:

```
  4,966,930 -- bare, native .reduce (just for comparison)
A 4,208,003 -- use .reduce in _reduce
A 3,355,363 -- _iterableReduce, but using for of, no reduced, no wrapping
  3,314,015 -- hand-written, no wrapping, no reduced
A 3,162,076 -- _iterableReduce, but using for of and without reduced
B 3,144,069 -- _arrayReduce, reduced switched off (but present)
  3,116,694 -- _arrayReduce, but no reduced
B 3,046,868 -- _iterableReduce, but using for of and reduced switched off
  2,543,624 -- Array.prototype.reduce.call
  1,003,981 -- use .reduce, wrap the function to emulate reduced
    877,481 -- no reduced, just wrapping & unwrapping
    877,364 -- reduced emulation* disabled, just wrapping and unwrapping
    413,357 -- _arrayReduce
    383,294 -- _iterableReduce, but using for of and reduced switched on
    380,261 -- _iterableReduce, but using for of
    354,904 -- _iterableReduce
    325,249 -- _arrayReduce, but reduced is emulated*
    287,915 -- _iterableReduce, but using for of, emulated* reduced 
```

<sup>\* by "emulated reduced", I mean adding `if (acc && acc['@@transducer/reduced']) return acc` to `_xwrap`'s step.</sup>

<sup>Where not specified otherwise, I was benching `_arrayReduce`.</sup>

I encourage to reproduce the bench. Also, perhaps it'd be a good idea to bench in Firefox.

My main idea:
1. Disallow `reduced` for functions, by either using functions that don't support it (`A`) or by adding a flag to existing functions to disable it as an optimization (`B`).
2. Where `reduced` is desirable, pass transformer instead of function (`reduced` is part of transducer spec). Consider exporting `_xwrap` to make the conversion easy (related: #1658).

GingerPlusPlus:
Those are the optimizations I deemed viable:

```
A1 4,208,003 -- use .reduce in _reduce
A2 3,355,363 -- _iterableReduce, but using for of, no reduced, no wrapping
A3 3,303,324 -- _iterableReduce, no reduced, no wrapping
B1 3,222,902 -- _iterableReduce, reduced switched off
B2 3,144,069 -- _arrayReduce, reduced switched off (but present)
```

A2 is only 6.7% faster than B2, so performance is not really a concern when choosing between them
(B2 is 7.1 to 7.6 times faster than current code for arrays, B1 is 9 to 10 times faster than current code for iterables).

Out of those, I'd pick B1 and B2, as they require least amount of code.

<hr />

The fastest, A1 is a 25% faster than A2. However, implementing it would cause inconsistencies with handling sparse arrays between environments with and without `Array.prototype.reduce`:

https://github.com/ramda/ramda/blob/f4c6b6232935e4ce790c34f76ee2fa1201fa19f4/source/reduce.js#L16-L17

<hr />

Disallowing the use of `reduced` in functions is of course a breaking change.

paldepind:
@scott-christopher That still iterates over the whole list though, right? It doesn't execute the thunkified function for all elements, but `thunkify` itself is still called for every element.

scott-christopher:
@paldepind: Correct, it's only really useful if the function itself is expensive. It doesn't solve for iterating over large arrays.

CrossEye:
@GingerPlusPlus: I will try to look over your recent work this weekend.  Thanks for all the contributions.

PR #2614, opened 2018-07-25
Add transducer support to takeLast

GingerPlusPlus:


CrossEye:
Is this just on spec, or do you have a strong need for this function?

That's a lot of additional code for something I've never seen requested.

GingerPlusPlus:
It's essential for adding transducer support to `slice` (https://github.com/ramda/ramda/pull/2342#issuecomment-336936214), supporting negative indexes.

buzzdecafe:
i'd rather remove support for negative indices

$0.02

GingerPlusPlus:
Support for negative indexes could be left up to the passed value.

Like, it'd still work on arrays, strings, and other things having a slice method supporting them.

PR #2616, opened 2018-07-26
Add repl script

GingerPlusPlus:


CrossEye:
This makes sense.  I often pull up the online REPL when I don't need it, only because Ramda is already in scope.  But a simple CLI repl would be useful.

CrossEye:
Thanks for the reminder.  I've finally been catching up on new items before heading back to the older ones.  But I can see no reason not to accept this.

PR #2618, opened 2018-07-30
Add transducer support to slice

GingerPlusPlus:
Alternative to #2342; closes #2342.


CrossEye:

While I don't have a strong objection, is there a real need for this?  Do you have a real-world use-case?

PR #2620, opened 2018-07-30
Adds `R.splitWith`

gpmcadam:
Takes a list and a predicate and returns a list of lists split
by (and including) matches.

### DIFF
--- a/source/splitWith.js
+++ b/source/splitWith.js
@@ -1,0 +1,30 @@
+import _curry2 from './internal/_curry2';
+import _reduce from './internal/_reduce';
+
+/**
+ * Takes a list and a predicate and returns a list of lists split by (and including) matches
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig (a -> Boolean) -> [a] -> [[a]]
+ * @param {Function} pred The predicate that determines where the array is split.
+ * @param {Array} list The array to be split.
+ * @return {Array}
+ * @example
+ *
+ *      R.splitWith(R.equals(2), [1, 2, 3, 1, 2, 3]);   //=> [[1, 2], [3, 1, 2], [3]]
+ */
+var splitWith = _curry2(function splitWith(pred, list) {
+  let currIndex = 0;
+  return _reduce(function(acc, curr) {
+    const clone = acc.length === 0 ? [new Array()] : [...acc];
+    clone[clone.length - 1].push(curr);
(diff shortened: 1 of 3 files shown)


Bradcomp:
The way I see it, there are four different 'reasonable' behaviors for this function. 

1. The way it is right now (include the split value at the end of the sub-lists)
2. Include the split value at the beginning of the sub-lists
3. Add the split value as it's own sublist
4. Exclude the split values from the results

I don't have a strong opinion right now on which way is 'best', it would be nice to hear some discussion on the 'right' API here.

source/splitWith.js
```diff
@@ -0,0 +1,30 @@
...
+ * @param {Array} list The array to be split.
+ * @return {Array}
+ * @example
+ *
+ *      R.splitWith(R.equals(2), [1, 2, 3, 1, 2, 3]);   //=> [[1, 2], [3, 1, 2], [3]]
+ */
+var splitWith = _curry2(function splitWith(pred, list) {
+  let currIndex = 0;
+  return _reduce(function(acc, curr) {
+    const clone = acc.length === 0 ? [new Array()] : [...acc];
```
GingerPlusPlus: There is no point in cloning here. You can just mutate `acc`.

functionalStoic:
No strong opinion but noticed Haskell has a similar function and it goes by behavior 4. https://hackage.haskell.org/package/split-0.2.3.3/docs/Data-List-Split.html#v:splitWhen

GingerPlusPlus:
> there are four different 'reasonable' behaviors for this function

Behavior 3:

> Add the split value as it's own sublist

`splitWhen3` ≈ [`groupWith`](https://ramdajs.com/docs/#groupWith)?

Bradcomp:
@functionalStoic Interesting that our `splitWhen` doesn't work the same way. If it did we wouldn't need this PR.

@GingerPlusPlus The major difference here is that `groupWith` takes a binary function for comparing adjacent items instead of a unary function with each item.

I tend to lean toward option 4 as well.

CrossEye:

I like this function, but with @Bradcomp's option 4 instead.

mudit56:
@CrossEye , @Bradcomp  isn't the 4th option already implemented in splitWhenever.js , which we merged last year.

mudit56:
This PR can be closed now I think as this feature (Point 4) is already merged in splitWhenever

PR #2624, opened 2018-08-05
Add Map and Set support to toString

wojpawlik:
#1055

### DIFF
--- a/source/internal/_toString.js
+++ b/source/internal/_toString.js
@@ -26,10 +26,14 @@ export default function _toString(x, seen) {
       return typeof x === 'object' ? 'new Boolean(' + recur(x.valueOf()) + ')' : x.toString();
     case '[object Date]':
       return 'new Date(' + (isNaN(x.valueOf()) ? recur(NaN) : _quote(_toISOString(x))) + ')';
+    case '[object Map]':
+      return 'new Map(' + recur(Array.from(x)) + ')';
     case '[object Null]':
       return 'null';
     case '[object Number]':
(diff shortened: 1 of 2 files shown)


CrossEye:

We really need to get 1.0 out the door.  This makes sense to me for when the library drops support for <es5.

GingerPlusPlus:
Note that the code doesn't __use__ `Map` and `Set` at all. The code just won't ever run if they aren't defined in the environment. So <es5 compat doesn't prevent merging this.

CrossEye:
Yes, going too fast through recent issues/prs.  This makes sense to me.

:herb:

source/internal/_toString.js
```diff
@@ -26,10 +26,14 @@ export default function _toString(x, seen) {
...
     case '[object Date]':
       return 'new Date(' + (isNaN(x.valueOf()) ? recur(NaN) : _quote(_toISOString(x))) + ')';
+    case '[object Map]':
+      return 'new Map(' + recur(Array.from(x)) + ')';
     case '[object Null]':
       return 'null';
     case '[object Number]':
       return typeof x === 'object' ? 'new Number(' + recur(x.valueOf()) + ')' : 1 / x === -Infinity ? '-0' : x.toString(10);
+    case '[object Set]':
+      return 'new Set(' + recur(Array.from(x).sort()) + ')';
```
Bradcomp: Why are we sorting the keys here? If key order matters then we should maintain the current ordering. If it doesn't matter then there's no reason to change it.

GingerPlusPlus: > If key order matters

It doesn't. Sets are by definition unordered.

Sorting guarantees that if `equals(setA, setB)`, then `eqBy(toString, setA, setB)`.

CrossEye (approved):

PR #2625, opened 2018-08-05
Ignore lock files

GingerPlusPlus:
Just so they don't show up in git, and don't get committed accidentally (https://github.com/ramda/ramda/pull/2186#issuecomment-304993652, https://github.com/ramda/ramda/pull/2506#pullrequestreview-111300673).

Incompatible with #2071, closes #2071.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,9 @@
 /lib/docs/public/
 /node_modules/
 /npm-debug.log
+/package-lock.json
+/shrinkwrap.yaml
+/yarn.lock
 /lib/test/bundle.js
 /tmp
 /bower_components/


Siilwyn:
Normally you would commit the package lock though...

CrossEye:
Why do you want to remove them?  As I understand them, the whole idea is that they enforce consistent builds.  With the lock file, anyone building from the same commit should get the same result.

Is there some benefit to ignoring them?

GingerPlusPlus:
Then, **do commit** one of them, and ignore the other two.

There are only problems if they are neither ignored, nor commited.

This PR solves them, while _keeping_ the files uncommited.

PR #2627, opened 2018-08-07
Change flip's arity from 1 to 3

### LINKED ISSUE #2177: Add left section, right section and infix functions
See https://github.com/sanctuary-js/sanctuary/pull/391#issuecomment-300962716.

To resolve the issues in #1497 regarding associative, non-commutative binary "operators" I propose adding three new functions inspired by how Haskell deals with infix [operators](https://wiki.haskell.org/Section_of_an_infix_operator):

```js
// leftSection :: (a, (a -> b -> c)) -> b -> c
// abbreviated _l
const _3subtract = R._l(3, R.subtract);
_3subtract(5); //=> -2

// rightSection :: ((a -> b -> c), b) -> a -> c
// abbreviated _r
const subtract3 = R._r(R.subtract,  3);
subtract3(5); //=> 2

// infix :: (a, (a -> b -> c), b) -> c
// not sure if it needs abbreviating. possibly _in
R.infix(3, R.subtract, 5); //=> -2
```

Before submitting a PR I'd like to get the semantics locked down. Specifically, should these functions be curried?

My initial thinking is no. These functions are mainly to aid in intent/readability. The examples for the section functions could be rewritten:
```js
const _3subtract = R.subtract(3);
_3subtract(5); //=> -2

const minus3 = R.subtract(R.__, 3); // or R.flip(R.subtract)(3)
minus3(5); //=> 2
```

This shows that there is no use for curried `R._l`, and curried `R._r` is equivalent to `R.flip`. But the intent is lost.

Without the fully qualified noise:
```js
// 3 subtract 5
_l(3, subtract)(5); //=> -2

// subtract 3 from 5
_r(subtract, 3)(5) //=> 2

// 3 subtract x is less than 5
pipe(_l(3, subtract), _r(lt, 5))( 12 / 3 ); //=> true


```
The weakest proposal is for `infix`. After all
```js
infix(1, lt, 2) //=> true

// can just as easily be written
1 < 2 //=> true
```
But it makes more sense once #1900 is released.
```js
infix([1, 2], lt, [2, 1]) //=> true
```

@CrossEye I got bored yesterday 😄

wojpawlik:
### Pros:

+ Simplifies `flip`'s implementation (no implicit currying).
+ Makes `flip` work as right section (#2177).
`applyTo` can be used as left section (and as `infix`, if redefined as `flip(call)`), so resolves #2177.

### Cons:

- Doesn't preserve function's arity.

### Alternative approaches:

* Add `f_` and `_f`, deprecate `flip` and `applyTo`.
* `uncurryN(2)` `flip`?

### What's next?

+ Deprecate placeholder.
+ Remove `mergeLeft` = `flip(merge)`, `mergeRight` = `merge`.
`merge` can stay deprecated, I added support for objects to `concat` at #2622.

### DIFF
--- a/source/flip.js
+++ b/source/flip.js
@@ -1,33 +1,30 @@
-import _curry1 from './internal/_curry1';
 import curryN from './curryN';
 
 
 /**
- * Returns a new function much like the supplied one, except that the first two
- * arguments' order is reversed.
+ * Calls the passed function with first 2 arguments swapped.
  *
  * @func
  * @memberOf R
  * @since v0.1.0
  * @category Function
- * @sig ((a, b, c, ...) -> z) -> (b -> a -> c -> ... -> z)
+ * @sig ((a, b, c, ...) -> z) -> b -> (a, c, ...) -> z
  * @param {Function} fn The function to invoke with its first two parameters reversed.
+ * @param {*} a First argument
+ * @param {*} b Second argument
  * @return {*} The result of invoking `fn` with its first two parameters' order reversed.
  * @example
  *
- *      const mergeThree = (a, b, c) => [].concat(a, b, c);
+ *      const mergeThree = (a, b, c) => [a, b, c];
+ *
+ *      const isPositive = R.flip(R.gt, 0);
  *
  *      mergeThree(1, 2, 3); //=> [1, 2, 3]
  *
  *      R.flip(mergeThree)(1, 2, 3); //=> [2, 1, 3]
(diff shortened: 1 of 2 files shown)


CrossEye:
This came in during my annual two-vacation on a lake.  I never saw it.

The *con* is a major one, since Ramda has always insisted on preserving arity.  Moreover, the implementation uses features not available for Ramda until post `v1.0`.

I'm closing this for now.  We can reexamine it later if we decide.

PR #2629, opened 2018-08-08
Remove placeholder, add useful function names

### LINKED ISSUE #2073: Better stack traces
So we've just shipped new stack traces in RunKit, and I'd really like Ramda to have easier to read stack traces (even outside of our tools). Currently it contains lots of Object.f2s and such, and I did a few experiments with setting functions names.

Here you can see what Ramda looks like currently:

screen shot 2017-02-01 at 9 31 39 pm

( https://runkit.com/tolmasky/ramda-current-traces )

And here is what they could look like:

screen shot 2017-02-01 at 9 31 47 pm

( https://runkit.com/tolmasky/ramda-better-traces-22 )

Basically, the trick is to do something like this:

```javascript
function functionNamed(aFunction, aName)
{
    Object.defineProperty(aFunction, "name", { value: aName });
    aFunction.displayName = aName;

    return aFunction;
}
```

GingerPlusPlus:
Resolves #2073! 😄


Bradcomp:
I'm totally 👍 on the function name stuff. 

Personally I like the placeholder but I won't make it my hill to die on, especially since it's messing with the docs.

buzzdecafe:
I am +:100: on removing the placeholder. 

If the goal of the naming thing is clearer stack traces, I am all for that, too.

CrossEye:
I have mixed feelings about the placeholder.  But I hate `curry1` and I really like the clearer function names.

I'm going to have to look at this and many other recent PRs in more detail over the next few days (now that vacation is over.)

naorzr:
Reviving this pr, why was this closed?

GingerPlusPlus:
#2705.

PR #2635, opened 2018-08-14
mapAccumRight

Leooman:
consistent with docs & I think this is right
`((x, acc) -> (y, acc)) -> acc -> [x] -> ([y], acc)`


CrossEye:
The underlying code has changed and this is no longer relevant, I believe.

PR #2637, opened 2018-08-14
Expand documentation of `memoize` and `memoizeWith`

evan-king:
Expand documentation of `memoize` and `memoizeWith` to clarify their limitations and help users find `memoizeWith(toString)` in lieu of choosing a deprecated function.

I'm assuming for now that the documentation supports multiple @example stanzas per docblock, as the examples added are separate and do not in my opinion stand well as the sole example.


source/memoizeWith.js
```diff
@@ -31,6 +32,27 @@ import _has from './internal/_has';
...
+ *
+ *      const memoizeByRef = R.memoizeWith(function() {
+ *          return weakReferenceId(arguments);
+ *      });
+ *
+ *      // getPrimaryContact :: [{primary: bool}] -> {primary: true}
+ *      const getPrimaryContact = memoizeByRef(R.compose(
+ *          R.find(R.propEq('primary', true)),
+ *          R.prop('contacts')
+ *      ));
```
Bradcomp: What about a simpler example? The one you have is confusing to me, and looks like it's replacing the memoizing capabilities of the function.

If you have a large object, it would be expensive to call `toString` on it. You can use this function to provide a separate version of equality. Here is a very contrived example that should probably be cleaned up a bit:

```
// very contrived
const objFib = (obj) => 
  obj.val <= 1 ? 
  1 : 
  objFib(evolve({val: subtract(__, 1)}, obj)) + objFib(evolve({val: subtract(__, 2)}, obj));
//now we only have to deal with the one property that matters instead of the whole object
const memoVal = compose(toString, prop('val'));
const memoized = memoizeWith(memoVal, objFib);
```

evan-king: I'm open to ideas.  The intent is to demonstrate the responsibilities/concerns of the key generating function without getting bogged down in them.  Your suggestion leads me to think about memoizing a function operating on database records by indexing on the record id.  But like the pre-existing example, it obscures variadic arguments, reference vs value, mutating input, etc.

My hope is that the user will be informed enough to use `memoizeWith` responsibly, alerting them to the pitfalls while simultaneously give the user a reason to take on those concerns rather than blindly use `memoize` or `memoizeWith(toString)`.

Maybe I'm just trying to do too much?

evan-king: I did just notice my signature didn't match my implementation, so I reduced the implementation to the signature.  I don't think it really addresses your concerns though.

Perhaps I should reduce to just the description changes and ditch the new examples.

CrossEye:
I'm simply assuming that we will remove `memoize` in the next release, so there's no good reason I can see to fix its documentation.

And while the changes to the description of `memoizeWith` look useful, I don't get much out of the example.  So I'd be all for reducing this PR to just the changes to the description of `memoizeWith`.

evan-king:
I take it documentation is never updated without a version release?  Part of the problem is that it is in fact still here, without any compelling explanation why not to use it or how to _easily_ replace it.

CrossEye:
We have re-released documentation, but that's usually when there's some problem with the documentation generation.  Otherwise, we don't really have a process for that.  We'd need to keep around branches for all old versions if that were the case.  We're going to have to find a better process for that, but I don't know what it is.  Usually if it's necessary to change something like that, we do a minor release.

MarcNq:
Hi guys,

What's the state on this issue?
I feel the docs on memoizeWith can really improve, since I had to come here to understand that:

- The function passed to memoizeWith is actually meant to return the hash used to cache the results.
- The basic usage of single-cell memoizeWith is `R.memoizeWith(R.toString)`.
- `R.memoizeWith(R.identity)` is a **dangerous misuse** and will make object inputs all be considered equal, no matter the object. It is definitely not the best example for a documentation.

CrossEye:
@MarcNq: You're right, we should include this with the updated documentation for `memoizeWith`.  I'd forgotten about it.

@evan-king: Would you be willing to remove the `memoize` version from this?

@MarqNq: Alternately, do you want to create a separate PR for better `memoizeWith` documentation?

I do think that while another example would be fine, `memoizeWith(identity)` is reasonable for String or Number inputs:

```js
const longRunningCalc = (n) => {
  console.log(`calculating value for ${n}`)
  return 2 * n // ok, so not really long-running!  :-)
}

const double = memoizeWith(identity, longRunningCalc)

console.log (
  double (1),
  double (2),
  double (1)
) //~>
// calculating value for 1
// calculating value for 2
// 2
// 4
// 2
```

MarcNq:
Hi @CrossEye , thanks for your answer!

I will try to open a PR with it. I agree `memoizeWith(identity)` gives no problem if restricted to basic types, and as long as we get the docs to explain it well, it will be fine.

evan-king:
How about just switching to `memoizeWith(Number, ...)` since that demonstrates expected-input awareness, keeps the example simple, and doesn't promote uninformed usage of `identity`?

PR #2646, opened 2018-08-28
Add R.xor (Exclusive OR)

rferromoreno:
_Exclusive OR_, or _exclusive disjunction_ is a logical operation that returns `true` if only one the arguments is `true` and the other is `false`. 

It can be easily implemented with other logical operations like `and`, `or` and `not`, but I guess it would be nice to have it in Ramda to improve the code readability.

Since it's my first pull request here, please let me know if I'm missing something.

Kind regards

### DIFF
--- a/source/xor.js
+++ b/source/xor.js
@@ -1,6 +1,5 @@
 import _curry2 from './internal/_curry2';
 
-
 /**
  * Exclusive disjunction logical operation.
  * Returns `true` if one of the arguments is `true` and the other is `false`.
@@ -22,6 +21,6 @@ import _curry2 from './internal/_curry2';
  *      R.xor(false, false); //=> false
  */
 var xor = _curry2(function xor(a, b) {
-  return (a || b) && !(a && b);
+  return Boolean(a || b) && !(a && b);
 });
 export default xor;
--- a/source/xor.js
+++ b/source/xor.js
@@ -21,6 +21,6 @@ import _curry2 from './internal/_curry2';
  *      R.xor(false, false); //=> false
  */
 var xor = _curry2(function xor(a, b) {
-  return Boolean(a || b) && !(a && b);
+  return Boolean(!a ^ !b);
 });
 export default xor;
--- a/source/xor.js
+++ b/source/xor.js
@@ -1,0 +1,27 @@
+import _curry2 from './internal/_curry2';
+
+
+/**
+ * Exclusive disjunction logical operation.
+ * Returns `true` if one of the arguments is `true` and the other is `false`.
+ * Otherwise, it returns `false`.
+ *
+ * @func
+ * @memberOf R
+ * @category Logic
+ * @sig a -> b -> Boolean
+ * @param {Any} a
+ * @param {Any} b
+ * @return {Boolean} true if one of the arguments is truthy and the other is falsy
+ * @see R.or, R.and
+ * @example
+ *
+ *      R.xor(true, true); //=> false
+ *      R.xor(true, false); //=> true
+ *      R.xor(false, true); //=> true
+ *      R.xor(false, false); //=> false
+ */
+var xor = _curry2(function xor(a, b) {
+  return (a || b) && !(a && b);
+});
+export default xor;
--- a/source/xor.js
+++ b/source/xor.js
@@ -2,7 +2,7 @@ import _curry2 from './internal/_curry2';
 
 /**
  * Exclusive disjunction logical operation.
- * Returns `true` if one of the arguments is `true` and the other is `false`.
+ * Returns `true` if one of the arguments is truthy and the other is falsy.
  * Otherwise, it returns `false`.
  *
  * @func
--- a/source/and.js
+++ b/source/and.js
@@ -12,7 +12,7 @@ import _curry2 from './internal/_curry2';
  * @param {Any} a
  * @param {Any} b
  * @return {Any} the first argument if it is falsy, otherwise the second argument.
- * @see R.both
+ * @see R.both, R.xor
  * @example
  *
  *      R.and(true, true); //=> true
--- a/source/index.js
+++ b/source/index.js
@@ -246,6 +246,7 @@ export { default as when } from './when';
 export { default as where } from './where';
 export { default as whereEq } from './whereEq';
 export { default as without } from './without';
+export { default as xor } from './xor';
 export { default as xprod } from './xprod';
 export { default as zip } from './zip';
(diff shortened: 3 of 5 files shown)


CrossEye:
Welcome!

This looks fine to me.  I think it makes sense so long as we have `and` and `or` to also have `xor`.  But you should know that there is a real argument for removing many such functions after 1.0, as they are trivial to implement with lambdas and arrow functions.

What do others think, @ramda/core?

There is one addition I would suggest: the See Also docs for `or` and `and` should be updated to point to this.

source/xor.js
```diff
@@ -0,0 +1,27 @@
...
+ * Returns `true` if one of the arguments is `true` and the other is `false`.
+ * Otherwise, it returns `false`.
+ *
+ * @func
+ * @memberOf R
+ * @category Logic
+ * @sig a -> b -> Boolean
+ * @param {Any} a
+ * @param {Any} b
+ * @return {Boolean} true if one of the arguments is truthy and the other is falsy
```
davidchambers: As implemented, the actual return type is `Any`. :\

```javascript
> R.xor (0) (0)
0
```

CrossEye: Right.  Perhaps `!!(a || b) && !(a && b)` would be appropriate.

Bradcomp: > !!(...)

I think you spelled `Boolean` wrong 😛

CrossEye: Long habit.  But yes, clearly `Boolean(a || b)` is better.

rferromoreno: @davidchambers Good point, thanks!

@ all: I know there is a long discussion about `and` and `or` returning types. Do you think this should return Any, or be converted to Boolean instead?

CrossEye: I think returning a boolean is fine here.  We're not trying to mimic a native boolean xor function here.

CrossEye:
@davidchambers
> It's too complex given Ramda's loose semantics.

By that logic, which Ramda functions should we keep?  :wink:

source/xor.js
rferromoreno: Okay, I'll change my commit and also add the @see docs to `and` and `or`

CrossEye:
I'm happy with this as it is.  Any other suggestions, comments?

Bradcomp:
If we are going to convert the output to a Boolean, should we add a test for that, or should we leave it as undefined behavior subject to change in the future?

CrossEye:
@Bradcomp: I see no reason to call it out in a test.  That's the signature.

But if you think it's a good idea, I certainly don't object.  We could use the example from @davidchambers to show that it always returns a Boolean.

CrossEye:
Yes, you're right.  My answer was over-hasty.

@rferromoreno: Could you add a few tests that demonstrate this.  We can leave property-based testing for later.

paldepind:
> As I see it, we could either restrict the domain to `bool -> bool`, in which case the tests are fine and we can get by with the old implementation

I'll throw a vote for that :smile:

If we give `xor` the type `Any -> Any -> Any` then what's holding us back from replacing every `Boolean` in every type declaration with `Any`?

rferromoreno:
Hi all, 

Sorry for the delay. I added some more tests for checking the signature.

If you think that more tests are needed, please let me know.

Thanks!

CrossEye:
I'm happy with this.  @Bradcomp?  Anyone else?

source/xor.js
```diff
@@ -0,0 +1,26 @@
...
+ * @see R.or, R.and
+ * @example
+ *
+ *      R.xor(true, true); //=> false
+ *      R.xor(true, false); //=> true
+ *      R.xor(false, true); //=> true
+ *      R.xor(false, false); //=> false
+ */
+var xor = _curry2(function xor(a, b) {
+  return Boolean(a || b) && !(a && b);
```
polytypic: How about a simpler expression like `Boolean(!a ^ !b)`?

rferromoreno: I like it. 
I simply didn't knew about the existence of the xor bitwise operator. I agree that your solution is simpler, so I'm changing it.

CrossEye: I'm not sure I like this so well.  The multiple levels of coercion seem a bit much: coerce both parameters to booleans, invert both, coerce the booleans to Numbers, apply the bitwise operator, then coerce the result again to a boolean.  I think I prefer the explicit approach above.

polytypic: The `^`-operator using version seems to be faster on both Node 8.11.3 and 10.8.0.  Of course, once you add currying, performance drops by a factor of 10 and so does the performance advantage of the `^`-operator using version.

rferromoreno: @polytypic so, I guess now the debate is between which solution is simple/easier to read, and also which performs better. Did you use any profiling tool for the comparison in performance? If so, could you please show me the results? 
Thanks!

polytypic: I used Benchmark.js.  Here is the biggest diff (not curried):

```
      95,043/s    1.00x  vs1000.reduce(xorBitwise)
      70,951/s    1.34x  vs1000.reduce(xorLogical)
```

`vs1000` is an array of 1000 elements (of number `1`, IIRC).  In all cases that I tried the version using `^` (named `xorBitwise` above) was faster.  Once you add Ramda's `curry`, the difference gets much smaller:

```
       9,532/s    1.00x  vs1000.reduce(xorBitwise)
       8,924/s    1.07x  vs1000.reduce(xorLogical)
```

Adding `R.curry` has a cost of one decimal order of magnitude in this case.

CrossEye: I don't have a major objection to the bitwise version.  But there is something odd about using it for Booleans.

source/xor.js
```diff
@@ -0,0 +1,26 @@
...
+ * @see R.or, R.and
+ * @example
+ *
+ *      R.xor(true, true); //=> false
+ *      R.xor(true, false); //=> true
+ *      R.xor(false, true); //=> true
+ *      R.xor(false, false); //=> false
+ */
+var xor = _curry2(function xor(a, b) {
+  return Boolean(!a ^ !b);
```
GingerPlusPlus:
```suggestion
  return !a !== !b;
```

rferromoreno: There was a previous discussion about this here: https://github.com/ramda/ramda/pull/2646#discussion_r217921724

I'm ok with any implementation. Any thoughts @CrossEye @polytypic ?

CrossEye: Yes, and sorry this got dropped.  Merged now.

PR #2649, opened 2018-09-10
Fix test of `API surface`

objectisundefined:
incorrect check `require.resolve !== 'function'` in test/index.js


CrossEye:
Nice catch!  Thank you very much.

PR #2650, opened 2018-09-14
missing a '(' in sig of addIndex

surmind:


CrossEye:
Thank you very much.

At some point, I would like to get rid of the internal ellipses on all our functions, but describing multiple versions of the sigs, each with fixed arities, and put an ellipsis after the list of them.  But this one is more difficult than most, as it is variadic in two different dimensions.  I suppose we could do something like this:

```hs
addIndex 
      :: ((a -> b) -> [x] -> c)) -> ((a, Int) -> b) -> [x] -> c
      :: (((a, b) -> c) -> [x] -> d)) -> ((a, b, Int) -> c) -> [x] -> d
      :: (((a, b, c) -> d) -> [x] -> e)) -> ((a, b, c, Int) -> d) -> [x] -> e
    :: ...
      :: ((a -> b) -> c -> [x] -> d)) -> ((a, Int) -> b) -> c -> [x] -> d
      :: (((a, b) -> c) -> d -> [x] -> e)) -> ((a, b, Int) -> c) -> -> d -> [x] -> e
      :: (((a, b, c) -> d) -> e -> [x] -> f)) -> ((a, b, c, Int) -> d) -> e -> [x] -> f
    :: ...
      :: ((a -> b) -> c -> d ->[x] -> e)) -> ((a, Int) -> b)-> c -> [x] -> d
      :: (((a, b) -> c) -> d -> [x] -> e)) -> ((a, b, Int) -> c) -> d -> e -> [x] -> f
      :: (((a, b, c) -> d) -> e -> f -> [x] -> g)) -> ((a, b, c, Int) -> d) -> e -> f -> [x] -> g
    :: ...
  :: ...
```

With examples like this:

```hs
map              :: (p -> q) -> [p] -> [q]
addIndex         :: ((a -> b) -> [x] -> c)) -> ((a, Int) -> b) -> [x] -> c
addIndex(map)    :: ((p, Int) -> q) -> [p] -> [q]
where  a :: p
       b :: q 
       c :: [q] 
       x :: p


reduce           :: ((q, p) -> q) -> q -> [p] -> q
addIndex         :: (((a, b) -> c) -> d -> [x] -> e)) -> ((a, b, Int) -> c) -> -> d -> [x] -> e
addIndex(reduce) :: ((q, p, Int) -> q) -> q -> [p] -> q
where  a :: q
       b :: p 
       c :: q 
       x :: p
```

But that's getting pretty complex too.

For the moment, of course, this is a great fix.  Thanks!

Issue #2653, opened 2018-09-19
Incorrect documentation for `replace()`


foxbunny:
The `replace()` function is documented as `RegExp|String → String → String → String`, but the second argument can also be a function due to the fact that `replace()` just delegates to `String.prototype.replace()`.

CrossEye:
Interesting.  I've never used it that way, but you're absolutely right.  Want to make a quick PR?

Bradcomp:
Before making a PR, we should ask if that is behavior we want to explicitly support? Just because it's possible doesn't mean it should be documented.

It might be better to just note that it delegates to String.replace, and people can then use that knowledge as they will.

foxbunny:
I'm not opposed to either solution. Let me know what kind of PR you want. :)

foxbunny:
This is (I assume) what the signature would look like if it were modified to match the actual behavior:

```
RegExp|String → (String|((*...) → String))  → String → String
```

Original documentation is [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#Specifying_a_function_as_a_parameter). If groups are used in the RegExp (first argument to `replace()`), the function can take a variable number of arguments, but then two more arguments are always supplied after them. 

I think making the signature more specific would not be very helpful without repeating the specification of the `String.prototype.replace()` function. Therefore, the user might still need to be told the Ramda function delegates to `String.prototype.replace()` so go look that up. In other words, what @Bradcomp suggests would probably be a good enough remedy, and would be necessary either way.

PR #2655, opened 2018-09-20
Improve R.replace() documentation

foxbunny:
- Add a note about R.replace() delegating to String.prototype.replace()
- Change the signature to match actual behavior

(ref #2653)

### DIFF
--- a/source/replace.js
+++ b/source/replace.js
@@ -4,6 +4,10 @@ import _curry3 from './internal/_curry3';
 /**
  * Replace a substring or regex match in a string with a replacement.
  *
+ * The first two parameters correspond to the parameters of the
+ * `String.prototype.replace()` function, so the second parameter can also be a
+ * function.
+ *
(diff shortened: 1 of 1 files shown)


CrossEye:
I personally prefer the first commit.  I think it's simpler, and I really doubt we'd ever change this behavior.

I wouldn't even mind a version that read something like

     This function delegates to the `String.prototype.replace()`, so you could also supply
     a function as replacer.

or some such

foxbunny:
It's getting a bit late in my neck of woods, so I'll redo the PR tomorrow.

foxbunny:
I apologize for the delay. The PR has been updated.

foxbunny:
I missed the point of the @CrossEye comment originally, so amended the last change. Sorry about the push spam. :)

CrossEye:
Thank you for persevering.

PR #2656, opened 2018-09-21
Add list of supported functions in reduced docs

### LINKED ISSUE #1766: Documentation on reduced (v0.20.1)
Reduced API docs state:

`Note: this optimization is unavailable to functions not explicitly listed above. For instance, it is not currently supported by reduceRight.`

But it doesn't list which functions support this optimization above (or below).

morgsmccauley:
resolves #1766


CrossEye:
Thank you very much for persevering with this!

Issue #2668, opened 2018-10-10
keeping simetric behaviour composing or using transducers in arrays and objects


xgbuils:
We have this code:

```  javascript
var mapSquare = R.map(x => x * x);
var mapInc = R.map(R.inc);
var intoArray = R.into([]);
var intoObject = R.into({});

var fn = R.compose(mapSquare, mapInc);
var xFn = R.compose(mapInc, mapSquare);

var arr = [1, 2, 3];
var obj = {a: 1, b: 2, c: 3};

fn(arr) // [4, 9, 16]
intoArray(xFn, arr) // [4, 9, 16]

// However ...
fn(obj) // {a: 4, b: 9, c: 16}
intoObject(xFn, obj) // throwing: reduce: list must be array or iterable
// but I also expect {a: 4, b: 9, c: 16}
```

I guess there are similar issues using `filter`. I could do a PR to fix it.

Best regards!

CrossEye:
I'd have to think carefully about what it means to support transducing on objects.  I think about `take(2)` for instance,  We think in Ramda as `{a: 1, b: 2, c: 3}` as equivalent to `{b: 2, c: 3, a: 1}`.  The thought that `take(2)` would yield different results for them makes me pretty suspicious.

But if you have ideas, please go ahead and share them.

CrossEye:
Sorry, hadn't seen the PR when I responded.  I'll check there.

Issue #2046, opened 2017-01-11
Reduce for Objects?


jethrolarson:
I find myself trying to turn Objects into other types pretty frequently and find myself doing something like:
```
Object.keys(myObj).reduce(reducer, someAccumulator)
```
the other part of the time I want to return early from the iteration and that ends up being the only time I really have to reach for a `for` loop.

What I'm hoping for is a function like:

`reduceObject :: (b -> a -> String -> b) -> Object a -> b`
I know this isn't strictly the same as ramda's reduce so I can see adding a separate function to deal with this case.

CrossEye:
#257, #364, #546, #625, #656, #920, #1067 

The problem is this sort of concern:

```js
var diff = function(x, y) {return y - x;}, 
    obj1 = {a: 1, b: 2, c: 3}, 
    obj2 = {b: 2, a: 1, c: 3};
_.reduce(obj1, diff); //=> 2
_.reduce(obj2, diff); //=> 4
```

`obj1` and `obj2` differ only in the order of key definition.

If we did something like that in Ramda, we'd violate a law that I'd like to maintain:

    if R.equals(x, y), then R.equals(f(x), f(y)) for any f.

#1067 was a long discussion, and involved one proposal that would allow us to do this but was pretty controversial: apply our own sort to keys before any function where it might matter.

paldepind:
@CrossEye 

We've been over this before 😄  But as was discussed in #1067 we're already violating the law:

```js
R.equals({a: 1, b:2}, {b: 2, a: 1}); //=> true
R.equals(R.keys({a: 1, b:2}), R.keys({b: 2, a: 1})); //=> false
```

So, should we remove `R.keys` and the other functions [you mention here](https://github.com/ramda/ramda/issues/1067#issuecomment-135923353)? Why do these function get to stay?

The way I see it our users keeps asking us to add this function and we keep trying to convince them it's bad for them.

To me the use case seems covered by `R.values` and `R.toPairs`

```js
R.reduceRight(([val, acc]) => ..., R.toPairs(object));
```

This also has advantage that `R.toPairs` can be used with `R.reduceWhile` which might be useful for the early return you mention @jethrolarson.

But this also leads me to the question: If we're already giving people something that is just as dangerous but just slightly less convenient why not give them the real thing?

I'm indifferent. I think I fine solution would be to add an example of reducing an object to the documentation for `R.toPairs`.

davidchambers:
[...]

Issue #1817, opened 2016-07-08
Must list functions work with objects?


neerone:
For now we using in our project [normalized](https://github.com/paularmstrong/normalizr) format of data store. We have all data in objects like this:
{1: {id:1, name:"test" ...} ...}

For example, it would be awesome if we could use R.last() on objects too (just get the last key value).
Only(?) R.filter works with objects but even R.forEach does not works as expected. 

Ramda is awesome but it needs to be more universal to use only ramda in our project (for now we use lodash+ramda+own helpers).

davidchambers:
> it would be awesome if we could use R.last() on objects too

Objects are unordered. How would we choose the "last" property? Would we sort the property names in some way and choose the last, or would we—:scream:—rely on enumeration order? If the former I think this is better left for others to define in their own projects, since the order we would choose would be arbitrary and therefore inappropriate in many contexts. If the latter we can no longer consider `{x: 0, y: 0}` and `{y: 0, x: 0}` equal – a sacrifice I'm unwilling to make.

neerone:
Thanks for the response. Now i see that the question have no sense. 
But for the more correctness, for example in V8 object key order match it's insertion order with exception that key doesn't parsed as Int32 value.
Again sorry for the dumb question. We previously used Immutable.OrderedMap() to keep objects key order

davidchambers:
> But for the more correctness, for example in V8 object key order match it's insertion order with exception that key doesn't parsed as Int32 value.

We could rely on insertion order, but it would be a shame for `R.equals({x: 0, y: 0}, {y: 0, x: 0})` to be forced to evaluate `false` (to ensure `R.last(a)` ≍ `R.last(b)` if `a` ≍ `b`).

CrossEye:
And if you want to see a painfully long discussion on this matter, see #1067.  :wink:

PR #2669, opened 2018-10-11
enable mapping and filtering objects with transducers

xgbuils:
Hi!

PR related to #2668

### DIFF
--- a/source/internal/_reduce.js
+++ b/source/internal/_reduce.js
@@ -35,8 +35,12 @@ function _methodReduce(xf, acc, obj, methodName) {
 }
 
 var symIterator = (typeof Symbol !== 'undefined') ? Symbol.iterator : '@@iterator';
+var typeErrorMessage = 'reduce: list must be array or iterable';
 
 export default function _reduce(fn, acc, list) {
+  if (!list) {
+    throw new TypeError(typeErrorMessage);
+  }
   if (typeof fn === 'function') {
     fn = _xwrap(fn);
   }
@@ -56,5 +60,5 @@ export default function _reduce(fn, acc, list) {
     return _methodReduce(fn, acc, list, 'reduce');
   }
 
-  throw new TypeError('reduce: list must be array or iterable');
+  throw new TypeError(typeErrorMessage);
 }
--- a/source/internal/_reduce.js
+++ b/source/internal/_reduce.js
@@ -74,8 +74,12 @@ export default function _reduce(fn, acc, list) {
   if (typeof list.reduce === 'function') {
     return _methodReduce(fn, acc, list, 'reduce');
   }
-  if (fn['@@transducer/commutative'] && _isObject(list)) {
-    return _objectReduce(fn, acc, list);
+  if (_isObject(list)) {
+    if (fn['@@transducer/commutative']) {
+      return _objectReduce(fn, acc, list);
+    } else {
+      throw new TypeError('reduce: unordered input passed for ordered transducer');
+    }
   }
 
   throw new TypeError(typeErrorMessage);
--- a/source/internal/_reduce.js
+++ b/source/internal/_reduce.js
@@ -1,4 +1,6 @@
 import _isArrayLike from './_isArrayLike';
+import _has from './_has';
+import _isObject from './_isObject';
 import _xwrap from './_xwrap';
 import bind from '../bind';
 
@@ -26,6 +28,19 @@ function _iterableReduce(xf, acc, iter) {
       break;
     }
     step = iter.next();
+  }
+  return xf['@@transducer/result'](acc);
+}
+
+function _objectReduce(xf, acc, object) {
+  for (var key in object) {
+    if (_has(key, object)) {
+      acc = xf['@@transducer/step'](acc, object[key], key);
+      if (acc && acc['@@transducer/reduced']) {
+        acc = acc['@@transducer/value'];
+        break;
+      }
+    }
   }
   return xf['@@transducer/result'](acc);
 }
@@ -58,6 +73,9 @@ export default function _reduce(fn, acc, list) {
   }
   if (typeof list.reduce === 'function') {
     return _methodReduce(fn, acc, list, 'reduce');
+  }
+  if (fn['@@transducer/commutative'] && _isObject(list)) {
+    return _objectReduce(fn, acc, list);
   }
 
   throw new TypeError(typeErrorMessage);
--- a/source/internal/_stepCat.js
+++ b/source/internal/_stepCat.js
@@ -1,10 +1,19 @@
 import _objectAssign from './_objectAssign';
 import _identity from './_identity';
 import _isArrayLike from './_isArrayLike';
+import _isString from './_isString';
 import _isTransformer from './_isTransformer';
 import objOf from '../objOf';
 
 
+function convertObjectInput(input, key) {
+  if (_isString(key)) {
+    return objOf(key, input);
+  } else if (_isArrayLike(input)) {
+    return objOf(input[0], input[1]);
+  }
+  return input;
+}
 var _stepCatArray = {
   '@@transducer/init': Array,
   '@@transducer/step': function(xs, x) {
@@ -20,12 +29,13 @@ var _stepCatString = {
 };
 var _stepCatObject = {
   '@@transducer/init': Object,
-  '@@transducer/step': function(result, input) {
+  '@@transducer/step': function(result, input, key) {
     return _objectAssign(
       result,
-      _isArrayLike(input) ? objOf(input[0], input[1]) : input
+      convertObjectInput(input, key)
     );
   },
+  '@@transducer/commutative': true,
   '@@transducer/result': _identity
 };
 
--- a/source/internal/_xfilter.js
+++ b/source/internal/_xfilter.js
@@ -9,8 +9,8 @@ function XFilter(f, xf) {
 }
 XFilter.prototype['@@transducer/init'] = _xfBase.init;
 XFilter.prototype['@@transducer/result'] = _xfBase.result;
-XFilter.prototype['@@transducer/step'] = function(result, input) {
-  return this.f(input) ? this.xf['@@transducer/step'](result, input) : result;
+XFilter.prototype['@@transducer/step'] = function(result, input, key) {
+  return this.f(input) ? this.xf['@@transducer/step'](result, input, key) : result;
 };
 
 var _xfilter = _curry2(function _xfilter(f, xf) { return new XFilter(f, xf); });
--- a/source/internal/_xfilter.js
+++ b/source/internal/_xfilter.js
@@ -5,6 +5,7 @@ import _xfBase from './_xfBase';
 function XFilter(f, xf) {
   this.xf = xf;
   this.f = f;
+  this['@@transducer/commutative'] = xf['@@transducer/commutative'];
 }
 XFilter.prototype['@@transducer/init'] = _xfBase.init;
 XFilter.prototype['@@transducer/result'] = _xfBase.result;
--- a/source/internal/_xmap.js
+++ b/source/internal/_xmap.js
@@ -5,11 +5,12 @@ import _xfBase from './_xfBase';
 function XMap(f, xf) {
   this.xf = xf;
   this.f = f;
+  this['@@transducer/commutative'] = xf['@@transducer/commutative'];
 }
 XMap.prototype['@@transducer/init'] = _xfBase.init;
 XMap.prototype['@@transducer/result'] = _xfBase.result;
-XMap.prototype['@@transducer/step'] = function(result, input) {
-  return this.xf['@@transducer/step'](result, this.f(input));
+XMap.prototype['@@transducer/step'] = function(result, input, key) {
+  return this.xf['@@transducer/step'](result, this.f(input), key);
 };
 
 var _xmap = _curry2(function _xmap(f, xf) { return new XMap(f, xf); });
--- a/test/filter.js
+++ b/test/filter.js
@@ -44,7 +44,7 @@ describe('filter', function() {
     var fn = R.compose(R.filter(even), R.filter(quintuple));
     assert.throws(function() {
       R.into([], fn, {a: 5, b: 10, c: 2});
-    }, throwReduceTypeError);
+    }, throwReduceTypeError('unordered input passed for ordered transducer'));
   });
 
 });
--- a/test/filter.js
+++ b/test/filter.js
@@ -1,9 +1,13 @@
+var assert = require('assert');
+
 var R = require('../source');
 var eq = require('./shared/eq');
+var throwReduceTypeError = require('./helpers/throwReduceTypeError');
 
 
 describe('filter', function() {
   var even = function(x) {return x % 2 === 0;};
+  var quintuple = function(x) {return x % 5 === 0;};
 
   it('reduces an array to those matching a filter', function() {
     eq(R.filter(even, [1, 2, 3, 4, 5]), [2, 4]);
@@ -29,6 +33,18 @@ describe('filter', function() {
   it('dispatches to passed-in non-Array object with a `filter` method', function() {
     var f = {filter: function(f) { return f('called f.filter'); }};
     eq(R.filter(function(s) { return s; }, f), 'called f.filter');
+  });
+
+  it('composes with object inputs using R.into', function() {
+    var fn = R.compose(R.filter(even), R.filter(quintuple));
+    eq(R.into({}, fn, {a: 5, b: 10, c: 2}), {b: 10});
+  });
+
+  it('does not filter with objects inputs and array acummulator using R.into', function() {
+    var fn = R.compose(R.filter(even), R.filter(quintuple));
+    assert.throws(function() {
+      R.into([], fn, {a: 5, b: 10, c: 2});
+    }, throwReduceTypeError);
   });
 
 });
--- a/test/helpers/throwReduceTypeError.js
+++ b/test/helpers/throwReduceTypeError.js
@@ -1,4 +1,5 @@
-export default function throwReduceTypeError(err) {
-  return err.constructor === TypeError &&
-    err.message === 'reduce: list must be array or iterable';
+export default function throwReduceTypeError(expectedMessage) {
+  return function(err) {
+    return err.constructor === TypeError && err.message === 'reduce: ' + expectedMessage;
+  };
 }
--- a/test/helpers/throwReduceTypeError.js
+++ b/test/helpers/throwReduceTypeError.js
@@ -1,0 +1,4 @@
+export default function throwReduceTypeError(err) {
+  return err.constructor === TypeError &&
+    err.message === 'reduce: list must be array or iterable';
+}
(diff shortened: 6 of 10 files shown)


CrossEye:
If you want to understand why I'm nervous about this, see  #257, #364, #546, #625, #656, #920, #1317, #1601, #1817, #2046, and most comprehensively, #1067.

The problem is this:

```js
> R.reduce((a, b) => b - a, 0, {x: 1, y: 2})
1
> R.reduce((a, b) => b - a, 0, {y: 2, x: 1})
-1
```

This scares me.

...

A lot.

buzzdecafe:
:ghost:

xgbuils:
Hi @CrossEye, good point!

I understand you, then, one solution for keeping the simetry in `R.map` and `R.filter` without breaking `reduce` and others is tagging that `_xmap` and `_xfilter` transducers are associative and having a different `_reduce` behaviour depending on the transducer that is passed.

What do you think?

Thanks!

CrossEye:
That's a really nice clean-up.  When I tried to think of clean ways to address the concerns I mentioned earlier, everything I came up with was downright ugly.  This seems pretty clean.

I'm :+1: on this.  I would like to hold it open for a few days to allow others from @ramda/core to have a chance to chime in.  But my take is that this is nice work.  Thanks for the contribution!

xgbuils:
Hi @CrossEye 

I'm sad to say that my second solution it doesn't work well. I added new tests that doesn't work as expected:
```
  it('throws if input is an object and function is a composition with some no associative transducer', function() {
    var throwsATypeError = function(err) {
      return err.constructor === TypeError &&
             err.message === 'reduce: list must be array or iterable';
    };
    assert.throws(function() {R.into({}, R.compose(mapInc, takeOne), {a: 1, b: 2});}, throwsATypeError);
    assert.throws(function() {R.into({}, R.compose(takeOne, mapInc), {a: 1, b: 2});}, throwsATypeError);
  });
```
I have to work a little bit more for having a consistent behaviour with transducer composition. My idea is if some transducer like `takeOne` is not associative, then object input shouldn't work.

Cheers!

xgbuils:
I fixed map and filter composition. I could squash the last two commits after people from @ramda/core give me the approval.

Thanks!

xgbuils:
Hi!

I checked that in `map.js` and `filter.js` there is code that handles the object case:

``` javascript
// filter
      _reduce(function(acc, key) {
        if (pred(filterable[key])) {
          acc[key] = filterable[key];
        }
        return acc;
      }, {}, keys(filterable))

// map
        acc[key] = fn(functor[key]);
        return acc;
      }, {}, keys(functor));
```

I think that, after the changes of this PR, two cases could be refactored to:

``` javascript
// filter
return _reduce(_xfilter(_stepCatObject), {}, filterable);

// map
return _reduce(_xmap(_stepCatObject), {}, functor);
```

Do you think it's good to do the refactor in this PR?

CrossEye:
Yes, I would include that refactoring in this same PR.  Thanks for persevering with this!

CrossEye:
@kedashoe:

You're absolutely right that the interesting property here is commutativity, not associativity.  (It's funny, I have a reasonably strong background in math, and I know clearly what those two mean, but people often make exactly this mistake, and it never triggers my, "Hey, that's wrong" signal.)

You're probably right too about `reduced`.  Just in general, I've never been thrilled with its existence.  In fact, our whole transducers work bothers me a bit.  It feels more clever than useful.  But when it's useful, it's *very* useful.  I'm simply not sure it's worth the complexity.  I've been in favor of the idea of this PR because I think it could occasionally be *very* useful as well.  But it would be a shame to say that we follow the `Functor` laws only if you don't happen to return `reduced`.  So yes, I'm starting to doubt it altogether.

@xgbuils: What do you think about the fact that this would break the [Functor laws][fu]?


  [fu]: https://github.com/fantasyland/fantasy-land#functor

kedashoe:
> Sorry by I don't get it. Could you put an example that breaks the law?

From the functor laws:

> iii. No parts of f's return value should be checked.

Here we are checking to see if the return value is the special `reduced` sentinel.

In any event, the commutativity issue is more to the crux of things I reckon. I've always felt we should allow folding objects regardless, so I'll let @CrossEye take this one.

xgbuils:
I'm trying to link the concepts between `R.reduced` function and `No parts of f's return value should be checked.` Maybe I need a specific example like:

given functions f and g:
``` javascript
var f = function(x) {
...
}

```
and object:
``` javascript
var obj = {...}
```

xgbuils:
I'm really trying to link the concepts between `R.reduced` function and `No parts of f's return value should be checked.` Maybe I need a specific example like:

given functions f and g:
``` javascript
var f = function(x) { ... }
var g = function(x) { ... }

```
and object:
``` javascript
var obj = {...}
```
`R.map` does not follow the law `n`.

Maybe could help me.

davidchambers:
Here's a concrete example, @xgbuils:

```javascript
> jQuery.map ([1, 2, 3, 4, 5], n => n % 2 === 0 ? null : n * n)
[1, 9, 25]
```

Note that `jQuery.map` inspects the value returned by the given function. This is unlawful: `map (x => x)` should be equivalent to `x => x` for any functor, but `jQuery.map (x => x, [null, null, null])` evaluates to `[]` rather than `[null, null, null]`.

xgbuils:
Hi @davidchambers 

Thanks for the explanation. I got it. But your answer it raises another question. Is it really necessary having this implementation rule (`No parts of f's return value should be checked`)? Is not enough being compliant with identity an composition laws for preventing these weird behaviour? For example, jQuery.map is not compliant with composition law.

``` javascript
var f = x => x === 1 ? null : 2* x;
var g = x => x === null ? 5 : 3 * x;

jQuery.map(jQuery.map([1, 2], f), g) // [12]
jQuery.map([1, 2], x => g(f(x))) // [5, 12]
```
Maybe is something to talk on fantasy-land repository.

Hi @kedashoe 

I think I got it you say. An example:

``` javascript
var f = x => x === 1 ? {'@@transducer/reduced': true, '@@transducer/value': 2 * x} : 2 * x
var g = x => 2 * x

// breaking composition law
R.map(g)((R.map(f, [1, 2])) // 4
R.map(x => g(f(x)))([1, 2]) // [NaN, 8]
```

I have to think about it. Thanks!

xgbuils:
After this dialog I learnt about some points.

1)
Tests like `eq(R.into([], x => x + 1, {a: 3, b: 2, c: 1}), [4, 2, 0]);` are wrong because an agreement I perceived is if we consider two equivalent inputs, then, we don't want to apply to the same function and get different outputs. Then, if we consider that `{a: 3, b: 2, c: 1}` is equal to `{b: 2, a: 3, c: 1})`, then we would like that `R.into([], x => x + 1, {a: 3, b: 2, c: 1})` was equal to `R.into([], x => x + 1, {b: 2, a: 3, c: 1})` but it isn't. Then, in my opinion, R.into({}) is the only function able to get the equivalence:

``` javascript
R.into({}, x => x + 1, {a: 3, b: 2, c: 1}) // {a: 4, b: 3, c: 2}
R.into({}, x => x + 1, {b: 2, a: 3, c: 1}) // {b: 3, a: 4, c: 2}
```
Because `{a: 4, b: 3, c: 2}` is equal to `{b: 3, a: 4, c: 2}`.

R.into('') and R.into([]) should be avoided to operate with objects.

2)
`map` is currently functor laws compliant with array an objects. However it is not compliant with transducers:

`R.into([], R.compose(R.map(f), R.map(g)), array)` is not equivalent to `R.into([], R.map(R.compose(g, f), array)` because of functions could use `R.reduced` function.

Then,

> @xgbuils: What do you think about the fact that this would break the Functor laws?

I think that we could keep the same behaviour as now. `map`. How we can do it? An easy way is avoiding to do the proposed refactor (`_reduce(_xfilter(pred, _stepCatObject), {}, filterable)` and `_reduce(_xmap(fn, _stepCatObject), {}, functor)`).

Of course, I'm going to replace `@@transducer/associative` to `@@transducer/commutative` and only `_objectReduce` will have set this property to true.

What do you think?

On the other hand, it comes to me a thought if there are some way to implement lawful transducers. But it's another topic.

GingerPlusPlus:
> I checked that in `map.js` and `filter.js` there is code that handles the object case

@xgbuils: while this is true, keep in mind that the specialization for arrays can also be expressed in terms of transducers (#2622).

> Do you think it's good to do the refactor in this PR?

No.

kedashoe:
The PR still includes [map](https://github.com/ramda/ramda/pull/2669/files#diff-d81ef730e2844e5a3858d4e0b0b304e6R8), but map is not commutative.

xgbuils:
Hi @kedashoe 

> The PR still includes [map](https://github.com/ramda/ramda/pull/2669/files#diff-d81ef730e2844e5a3858d4e0b0b304e6R8), but map is not commutative.

The only transducer that is commutative is [_stepCatObject](https://github.com/ramda/ramda/pull/2669/files#diff-c78dd3dc852a1f37454e6a9c243cf6fcR38)

```
const ts = '@@transducer/step';

_stepCatObject[ts](_stepCatObject[ts](acc, a), b) is equivalent to _stepCatObject[ts](_stepCatObject[ts](acc, b), a)
```

`_xmap` just propagate the commutativity. I mean, if `xf` is commutative, then `new XMap(f, xf)` is commutative.

You can see the related tests:
- [composing map and take (no commutative)](https://github.com/ramda/ramda/pull/2669/files#diff-62207d04e00d3b7775774d12b10b8a38R64)
- [composing map and filter (commutative)](https://github.com/ramda/ramda/pull/2669/files#diff-62207d04e00d3b7775774d12b10b8a38R44)

PS. I noticed that I didn't update the description of tests and it is still written `associative` instead of `commutative`. I will fix it.

kedashoe:
> `_xmap` just propagate the commutativity.

Hmm, why is it propagated in `_xmap` and not in `_xtake`?

GingerPlusPlus:
Because `_xtake` is order-sensitive.

kedashoe:
> Because `_xtake` is order-sensitive.

Map is order sensitive. Is take though? What do you get if you take 5 and then take 1? What if you take 1 and then take 5?

xgbuils:
Hi @kedashoe! 

> Map is order sensitive. Is take though?

It isn't. For example:

``` javascript
const ts = const ts = '@@transducer/step';
const xTakeOne = new XTake(1, _stepCatObject);
let acc = {}

xTakeOne[ts](xTakeOne[ts](acc, {a: 1}), {b: 2}) // _reduced({a: 1})
xTakeOne[ts](xTakeOne[ts](acc, {b: 2}), {a: 1}) // _reduced({b: 2})
```

> What do you get if you take 5 and then take 1? What if you take 1 and then take 5?

When I talk about transducer commutativity, I'm talking about the point indicated above, not about function composition like `R.compose(R.take(1), R.take(5)) ===  R.compose(R.take(5), R.take(1))`

kedashoe:
> not about function composition

Ya I think I was not looking at this correctly.

I don't understand how this addresses the initial concern though. I'm not sure what property we're actually defining here and what it has to do with commutativity. I don't want to hold this up though if others want to move forward.

CrossEye:
I want to look through this one more thoroughly again.  I should have time starting Wednesday.

CrossEye:
I just want to note that I haven't forgotten this.  But I have to say that every time I try to get through this, I end up confused.  Part of this is that while *what* transducers do is pretty well clear to me, it's not always clear *how* they do so.  But part of it is really understanding why this is really important, and what's left of the original concerns that spawned it.  I will spend some more time in the next few days, but I'm afraid I might get woozy again.

xgbuils:
Hi @CrossEye 

It's curious, because I learned about transducers thanks to `ramda` library. A year ago I tried to create a library that manages lazy evaluation using iterables. But after making some performance benchmarks I noticed that `ramda` won using transducers over all the libraries that try to do same aproach than me. Then, I needed to look into *what* are transducers and *how* they work. Indeed, this PR and  the comments I received are really valuable for learning about transducer concepts.

I can try to synthesize what has been talked during these days. First of all I noticed that sometimes I used the word *transducer* when I wanted to say *transformer*. Then, I want to clarify the two concepts: 

- A *transformer* is an object that has some properties like `@@tranducer/init`, `@@tranducer/step`, `@@tranducer/result`, etc.
- A *transducer* is a function that receives a transformer and returns another transformer.

My final proposal can be splitted in two parts:
1. It enables to transduce objects with commutative transformers and throws an error when not.
2. `R.map/R.filter` remain functor lawful compliant with arrays and objects and remain no functor lawful compliant with transducers.

### About point 1:

A commutative transformer is new concept that I defined in this way. A transformer `xf` is commutative if `@@transducer/step` complies this law for each `acc` accumulator and `a` and `b` input items:

``` javascript
const ts = '@@transducer/step';
xf[ts](xf[ts](acc, a), b) is equivalent to xf[ts](xf[ts](acc, b), a)
```

or for each `acc` and <key, value> pairs `<ka, a>` and `<kb, b>`:

``` javascript
const ts = '@@transducer/step';
xf[ts](xf[ts](acc, a, ka), b, kb) is equivalent to xf[ts](xf[ts](acc, b, kb), a, ka)
```

We can notice that at least [_stepCatObject](https://github.com/ramda/ramda/pull/2669/files#diff-c78dd3dc852a1f37454e6a9c243cf6fcR30) is a commutative transformer. And, for each commutative transformer and object input that is traversed by [_objectReduce](https://github.com/ramda/ramda/pull/2669/files#diff-5861836cf1fab2e7793d633617b63889R35), the transduced result is the same regardless the order of the object keys (forget objects with reduced values at the moment, I'm going to talk about it in the point 2).

We can also notice that:
-  `_xmap(f, xf)` is a commutative transformer if `xf` is commutative transformer.
-  `_xfilter(f, xf)` is a commutative transformer if `xf` is commutative transformer.

I took advantadge on it for creating a new `@@transducer/commutative` prop that indicates if transformer is commutative or not:
- `_stepCatObject` has the property  `@@transducer/commutative` set to `true`
- `_xmap(f, xf)` has the property  `@@transducer/commutative` set to `true` if and only if `xf` has the property  `@@transducer/commutative` set to `true`
- `_xfilter(f, xf)` has the property  `@@transducer/commutative` set to `true` if and only if `xf` has the property  `@@transducer/commutative` set to `true`.

@kedashoe pointed that my definition of commutative transformer isn't related to the concept of commutativity. I'm open to change to another better name if you see it right.

I will talk about de point 2 when I have a while.

Cheers!

GingerPlusPlus:
> why this is really important, and what's left of the original concerns that spawned it. 

After getting this merged, I'd modify `_dispatchable` to utilize transducers to map and filter arrays and objects, and I'd get rid of array and object specializations and associated ad-hoc type switching. Then I'd add `Map` suport to `map` and `filter` by adding `_stepCat` and `_reduce` for `Map`.

source/internal/_reduce.js
```diff
@@ -30,13 +32,30 @@ function _iterableReduce(xf, acc, iter) {
...
+  }
+  return xf['@@transducer/result'](acc);
+}
+
 function _methodReduce(xf, acc, obj, methodName) {
   return xf['@@transducer/result'](obj[methodName](bind(xf['@@transducer/step'], xf), acc));
 }
 
 var symIterator = (typeof Symbol !== 'undefined') ? Symbol.iterator : '@@iterator';
+var typeErrorMessage = 'reduce: list must be array or iterable';
```
GingerPlusPlus:
```suggestion
var typeErrorMessage = 'reduce: input must be object or iterable';
```

xgbuils: Hi! I don't like the current error message, but I think the proposal isn't accurate enough when `_reduce(R.take(1), {}, {a: 1, b: 2})` throws an error and `_reduce(R.map(x => 2 * x), {}, {a: 1, b: 2})` doesn't.

GingerPlusPlus:
```js
  if (_isObject(list)) {
    if (fn['@@transducer/commutative']) {
      return _objectReduce(fn, acc, list);
    } else {
      throw new TypeError('reduce: passed transducer requires ordered input');
      // or
      throw new TypeError('reduce: unordered input passed for ordered transducer');
    }
  }
```

xgbuils:
Hi! A month ago I said that:

> 2. `R.map/R.filter` remain functor lawful compliant with arrays and objects 
This part is true because the particular implmentations for arrays and objects are not modified.

> [...] and remain no functor lawful compliant with transducers.
And after thinking a little more, I don't know what means be lawful compliant with transducers or transformers

Assuming that transducer is a function that receives a transformer and returns a transformer. A transducer proposal functor laws could be:

1. Identity law:
`R.map(x => x)` is equivalent to `xf => xf`
2. Composition law:
`R.compose(R.map(f), R.map(g))` is equivalent `R.map(R.compose(g, f))`

Notice that `R.reduce` or `R.into` does not participate on these laws. Then, we have no problem with objects with `@@transducer/reduced` property because `@@transducer/reduced` property is something handled by `_reduce` implementation.

The real problem with transducers are not object inputs: The real problem is using an object accumulator with `R.reduce`:

``` javascript
var f = ([key, value]) => ({[key]: 2 * value})

R.into({}, R.map(f), [['a', 1], ['b', 2]]) // {"a": 2, "b": 4}
R.into({}, R.map(f), [['@@transducer/reduced', 1], ['b', 2]]) // undefined
```
I think it's an issue of javascript transducer protocol.

Cheers!

GingerPlusPlus:
Good catch!

Workaround: store the object in accumulator as array of `[key, value]` pairs and use `fromPairs` as `@@transducer/result`.

xgbuils:
Hi @GingerPlusPlus , 

I forgot this PR. Updated with your proposals.

Thanks!

mneumark:
@GingerPlusPlus Any chance of taking a look again and seeing if its ready for a merge?

GingerPlusPlus:
I would be reluctant to merge this until `@@transducer/reduced` becomes a valid key.

xgbuils:
Hi @GingerPlusPlus 

> I would be reluctant to merge this until `@@transducer/reduced` becomes a valid key.

I don't know how `@@transducer/reduced` issue is related to this issue. As I showed, the problem about `@@transducer/reduced` exists regardless merging this PR. Is there another issue for fixing  `@@transducer/reduced` issue? On the other hand, I'm not sure how to fix it without breaking the transducer protocol. I mean, [the specs says that a value with `@@transducer/reduced` prop should be considered a reduced value](https://github.com/cognitect-labs/transducers-js#reduced).

GingerPlusPlus:
> I don't know how `@@transducer/reduced` issue is related to this PR.

I'm already thinking in terms of reimplementing object-to-object paths with the enhanced transducers, and that issue would cause a (small) regression there. And, in general, I'd be reluctant to merge code with known issues.

> As I showed, the problem about `@@transducer/reduced` exists regardless merging this PR.

Where did you show it?

xgbuils:
> Where did you show it?

Some repplies above I said:

``` javascript
var f = ([key, value]) => ({[key]: 2 * value})

R.into({}, R.map(f), [['a', 1], ['b', 2]]) // {"a": 2, "b": 4}
R.into({}, R.map(f), [['@@transducer/reduced', 1], ['b', 2]]) // undefined instead of {'@@transducer/reduced': 1, b: 2}
```

Then, the issue about `@@transducer/reduced` as valid key is not introduced in this PR.

CrossEye:
I'm going through old issues and trying to triage them.  This one still has my head spinning, and I'll come back to it soon.

Issue #2667, opened 2018-10-10
align `lensPath` with `lensIndex` and `nth` semantics


2beaucoup:
would be nice if `lensPath` would support negative values:
```js
var a = ['a', 'b', 'c'];

var n = R.nth(-2);
var i = R.lensIndex(-2);
var p = R.lensPath([-2]);

n(a); // "b"
R.view(i, a); // "b"
R.view(p, a); // undefined
```

CrossEye:
While I won't guarantee it would be accepted, a PR along these lines would certainly be welcome.

PR #2670, opened 2018-10-12
align lensPath with lensIndex by supporting negative indeces

2beaucoup:
see #2667

### DIFF
--- a/test/path.js
+++ b/test/path.js
@@ -28,6 +28,19 @@ describe('path', function() {
     eq(R.path(['j', 1], obj), undefined);
   });
 
+  it('takes a path that contains indeces into arrays that might be negative', function() {
+    var obj = {
+      a: [[{}], [{x: 'first'}, {x: 'second'}, {x: 'third'}, {x: 'last'}]]
+    };
+    eq(R.path(['a', 0, 0], obj), {});
+    eq(R.path(['a', 0, 1], obj), undefined);
+    eq(R.path(['a', 1, 0, 'x'], obj), 'first');
+    eq(R.path(['a', 1, 1, 'x'], obj), 'second');
+    eq(R.path(['a', 1, -2, 'x'], obj), 'third');
+    eq(R.path(['a', 1, -1, 'x'], obj), 'last');
+    eq(R.path([0], ['A']), 'A');
+  });
+
   it("gets a deep property's value from objects", function() {
     eq(R.path(['a', 'b', 'c'], deepObject), 'c');
     eq(R.path(['a'], deepObject), deepObject.a);
--- a/test/path.js
+++ b/test/path.js
@@ -24,6 +24,7 @@ describe('path', function() {
     eq(R.path([], obj), obj);
     eq(R.path(['a', 'e', 'f', 1], obj), 101);
     eq(R.path(['j', 0], obj), 'J');
+    eq(R.path(['j', -1], obj), 'J');
     eq(R.path(['j', 1], obj), undefined);
   });
 
--- a/test/path.js
+++ b/test/path.js
@@ -24,11 +24,10 @@ describe('path', function() {
     eq(R.path([], obj), obj);
     eq(R.path(['a', 'e', 'f', 1], obj), 101);
     eq(R.path(['j', 0], obj), 'J');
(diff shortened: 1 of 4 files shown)


test/path.js
```diff
@@ -24,6 +24,7 @@ describe('path', function() {
     eq(R.path([], obj), obj);
     eq(R.path(['a', 'e', 'f', 1], obj), 101);
     eq(R.path(['j', 0], obj), 'J');
+    eq(R.path(['j', -1], obj), 'J');
```
Bradcomp: A couple things here. The object at `j` only has a single item in it, so it's not as good of a test of whether the negative indices work.  I would prefer to have a separate test for this behavior rather than just adding a case to this test.

CrossEye:
This looks good to me.  I am not really a fan of supporting negative indices, but I would rather maintain the consistency.

Anyone else?  @ramda/core?

2beaucoup:
added some tests + docs

test/path.js
```diff
@@ -28,6 +28,19 @@ describe('path', function() {
     eq(R.path(['j', 1], obj), undefined);
   });
 
+  it('takes a path that contains indeces into arrays that might be negative', function() {
```
CrossEye: sp: "indices"  (or possibly "indexes")

test/path.js
```diff
@@ -28,6 +28,19 @@ describe('path', function() {
...
+    var obj = {
+      a: [[{}], [{x: 'first'}, {x: 'second'}, {x: 'third'}, {x: 'last'}]]
+    };
+    eq(R.path(['a', 0, 0], obj), {});
+    eq(R.path(['a', 0, 1], obj), undefined);
+    eq(R.path(['a', 1, 0, 'x'], obj), 'first');
+    eq(R.path(['a', 1, 1, 'x'], obj), 'second');
+    eq(R.path(['a', 1, -2, 'x'], obj), 'third');
```
CrossEye: I know that these tests are building upon what was already there, but I think slowly the signal is getting lost in the noise.  There should be other tests for deep paths, so a demonstration of negative indices should be simpler, perhaps something like this:

```js
eq(R.path(['x', -2], {x: ['a', 'b', 'c', 'd']}), 'c')
```
along with

```js
eq(R.path([-1, 'y'], [{x: 1, y: 99}, {x, 2, y: 98}, {x: 3, y: 97}]), 97)
```

test/path.js
```diff
@@ -24,9 +24,23 @@ describe('path', function() {
     eq(R.path([], obj), obj);
     eq(R.path(['a', 'e', 'f', 1], obj), 101);
     eq(R.path(['j', 0], obj), 'J');
+    eq(R.path(['j', -1], obj), 'J');
     eq(R.path(['j', 1], obj), undefined);
   });
 
+  it('takes a path that contains indeces into arrays that might be negative', function() {
```
CrossEye: sp: "indices" (or possibly "indexes")

test/path.js
```diff
@@ -24,9 +24,23 @@ describe('path', function() {
...
+    var obj = {
+      a: [[{}], [{x: 'first'}, {x: 'second'}, {x: 'third'}, {x: 'last'}]]
+    };
+    eq(R.path(['a', 0, 0], obj), {});
+    eq(R.path(['a', 0, 1], obj), undefined);
+    eq(R.path(['a', 1, 0, 'x'], obj), 'first');
+    eq(R.path(['a', 1, 1, 'x'], obj), 'second');
+    eq(R.path(['a', 1, -2, 'x'], obj), 'third');
```
CrossEye: I know that these tests are building upon what was already there, but I think slowly the signal is getting lost in the noise.  There should be other tests for deep paths, so a demonstration of negative indices should be simpler, perhaps something like this:

```js
eq(R.path(['x', -2], {x: ['a', 'b', 'c', 'd']}), 'c')
```
along with

```js
eq(R.path([-1, 'y'], [{x: 1, y: 99}, {x, 2, y: 98}, {x: 3, y: 97}]), 97)
```

2beaucoup:
Thanks! I added a test that should address your feedback.

CrossEye:
Looks good to me.  :herb:

@Bradcomp?

PR #2675, opened 2018-10-21
Enhance `map` To Handle `AsyncFunction` (+ more), add docs for `type`

willheslam:
This is intended as an addendum to https://github.com/ramda/ramda/pull/2513, but the previous PR is consistent without these changes and can be merged before this one.

The main change I've made is to `map`, which can take a function as a functor.
I've included the same fix @rjhilgefort did here: https://github.com/ramda/ramda/commit/b5a3cb6563d22d29b5798bfeca02938ab94c2761
with some tests to show what it would look like.

The only other changes are superficial - adding documentation and tests to prove the behaviour of `type`: https://github.com/ramda/ramda/commit/92b96c72f179c452d4894b7bc9642c1a17cabb87
which unfortunately requires making a change to eslint to parse async functions:
https://github.com/ramda/ramda/commit/764b105dba86ed3b256e1bb46a4057e187157ddc
I'd appreciate any feedback on whether this is OK or if there are any reasons to hold back on changing eslint settings.

Is `(async (->) r)` the right signature for an async function?

I've kept the `type` docs/tests and eslintrc changes separate so the `map` changes can be easily merged without them.

Please let me know if anything else needs changing!

### DIFF
--- a/source/map.js
+++ b/source/map.js
@@ -45,6 +45,9 @@ import keys from './keys';
 var map = _curry2(_dispatchable(['fantasy-land/map', 'map'], _xmap, function map(fn, functor) {
   switch (Object.prototype.toString.call(functor)) {
     case '[object Function]':
+    case '[object AsyncFunction]':
+    case '[object GeneratorFunction]':
+    case '[object AsyncGeneratorFunction]':
       return curryN(functor.length, function() {
         return fn.call(this, functor.apply(this, arguments));
       });
--- a/source/type.js
+++ b/source/type.js
@@ -24,6 +24,8 @@ import _curry1 from './internal/_curry1';
  *      R.type([]); //=> "Array"
  *      R.type(/[A-z]/); //=> "RegExp"
  *      R.type(() => {}); //=> "Function"
+ *      R.type(async () => {}); //=> "AsyncFunction"
+ *      R.type(function* () {}); //=> "GeneratorFunction"
  *      R.type(undefined); //=> "Undefined"
  */
 var type = _curry1(function type(val) {
--- a/test/map.js
+++ b/test/map.js
@@ -42,10 +42,10 @@ describe('map', function() {
   });
 
   it('interprets generator as a functor', function() {
-    var f = function(gen) { return [gen.next().value - 1, gen.next().value - 2, gen.next().value - 3]; };
-    var g = function* (b) { yield b * 1; yield b * 2; yield b * 3; };
+    var f = function(gen) { return gen.next().value - 1; };
+    var g = function* (b) { yield b * 2; };
     var h = R.map(f, g);
-    eq(h(10), [(10 * 1) - 1, (10 * 2) - 2, (10 * 3) - 3]);
+    eq(h(10), (10 * 2) - 1);
   });
 
   it('dispatches to objects that implement `map`', function() {
--- a/test/map.js
+++ b/test/map.js
@@ -31,6 +31,23 @@ describe('map', function() {
     eq(h(10), (10 * 2) - 1);
   });
 
+  it('interprets (async (->) r) as a functor', function() {
+    var f = function(a) { return a.then(function(a) { return a - 1; }); };
+    var g = async function(b) { return b * 2; };
+    var h = R.map(f, g);
+    return h(10)
+      .then(h => {
+        eq(h, (10 * 2) - 1);
+      });
+  });
+
+  it('interprets generator as a functor', function() {
+    var f = function(gen) { return [gen.next().value - 1, gen.next().value - 2, gen.next().value - 3]; };
+    var g = function* (b) { yield b * 1; yield b * 2; yield b * 3; };
+    var h = R.map(f, g);
+    eq(h(10), [(10 * 1) - 1, (10 * 2) - 2, (10 * 3) - 3]);
+  });
+
   it('dispatches to objects that implement `map`', function() {
     var obj = {x: 100, map: function(f) { return f(this.x); }};
     eq(R.map(add1, obj), 101);
--- a/test/type.js
+++ b/test/type.js
@@ -45,4 +45,15 @@ describe('type', function() {
     eq(R.type(undefined), 'Undefined');
   });
 
+  it('"Function" if given a function', function() {
+    eq(R.type(() => {}), 'Function');
+  });
+
+  it('"AsyncFunction" if given an async function', function() {
+    eq(R.type(async() => {}), 'AsyncFunction');
+  });
+
+  it('"Generator Function" if given a generator function', function() {
+    eq(R.type(function* () {}), 'GeneratorFunction');
+  });
 });
--- a/test/type.js
+++ b/test/type.js
@@ -53,7 +53,7 @@ describe('type', function() {
     eq(R.type(async() => {}), 'AsyncFunction');
   });
 
-  it('"Generator Function" if given a generator function', function() {
+  it('"GeneratorFunction" if given a generator function', function() {
     eq(R.type(function* () {}), 'GeneratorFunction');
   });
 });
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,7 @@
 {
   "parserOptions": {
-    "sourceType": "module"
+    "sourceType": "module",
+    "ecmaVersion": 2017
   },
   "env": {
     "node": true,


CrossEye:
Wouldn't this cause a problem with mapping several async functions?

```js
> const R = require('./dist/ramda')
> const f1  = (foo) => `${foo} - bar`
> const g1 = (s) => R.toUpper(s)
> const h1 = R.map(f1, g1)
> h1('baz')
'BAZ - bar'

> const f2  = async (foo) => `${foo} - bar`
> const g2 = async (s) => R.toUpper(s)
> const h2 = R.map(f2, g2)
> h2('baz')
TypeError: h2 is not a function
    ...
> h2
  Promise('undefined - bar')
```

> which unfortunately requires making a change to eslint to parse async functions:
> 764b105
> I'd appreciate any feedback on whether this is OK or if there are any reasons to hold back on changing eslint settings.

I don't think there are any concerns with that.

> Is (async (->) r) the right signature for an async function?

I hadn't given it any thought until now; Ramda has had no dealings with async functions.  Perhaps it would be better if when we added `async` to `(a -> b)` that we simply change it to `(a -> Promise b)`.

willheslam:
Good points!
I've tried replicating your example with my PR - hopefully I'm interpreting it correctly.
```
> const f2 = async (foo) => `${foo} - bar`
> const g2 = async (s) => R.toUpper(s)
> const h2 = R.map(f2, g2)
> h2('baz')
Promise("[object Promise] - bar")
> h2
function
```
The reason is exactly the same as if you replaced `map` with `compose` - `g2` returns a Promise and `f2` tries to combine it (stringifying) with `" - bar"`, returning that value wrapped in a Promise (being async itself).

If `f2` was to do what you'd expect, it'd need to be:
```
const f2 = async (foo) => `${await foo} - bar`
//or, without await
const f2 = async (foo) => foo.then(foo => `${foo} - bar`)
```
This is the case with `map` and also `compose`.
In the case of compose you could get the same effect with `composeP`:
`composeP(f2, g2)('baz')` //Promise('BAZ - bar')

Being completely honest I'm still wrapping my head around the exact mathematical implications of functions as functors - I mainly use `map(f, g)` as a way of composing n-ary functions with unary functions ala `(a, b) => f(g(a, b))` in a way that can be partially applied.
e.g.
```map(add(1), (x, y) => x + y)(10)(20)``` //31
vs
```compose(add(1), (x, y) => x + y)(10)(20)``` //compose(...)(...) is not a function

(In my PR there is no equivalent of `mapP` (e.g.  `mapP(f, g)(a)(b)` - to `composeP` in the same way `map` is to `compose`) in the same way there is no `mapK`.)

The main aim here is to make sure `map` treats `async () => 42` in the same way it treats `() => Promise.resolve(42)`, which it doesn't right now due to `Object.prototype.toString`'s (unfortunate?) perception of the former as `[object AsyncFunction]` and the latter as `[object Function]`

That actually also applies to your next point about the function signature:
`it('interprets (async (->) r) as a functor', function() {` //in test/map.js
There is potentially an issue with using `(a -> Promise b)` to represent an async function - `map` *already* treats `(a -> Promise b)` as a functor - it's the specific case of `(async a -> b)` that `map` sees differently.
Personally I feel that async functions *should* be seen as `(a -> Promise b)` as they behave identically to my knowledge, but `Object.prototype.toString` leaks the implementation details of the function. IMO this is another point in `typeof`'s favour, because it sees them both as just `'function'`. :)

I'm happy to put together a PR that uses `typeof` instead, as posited by @rjhilgefort here https://github.com/ramda/ramda/pull/2513#issuecomment-411116312

**Comment by CrossEye** [comment]
>>     > h2('baz')
>>     TypeError: h2 is not a function

vs.

>     > h2('baz')
>      Promise("[object Promise] - bar")

This difference is interesting.  I simply tested what I thought was the correct branch from your fork, and that was the result I got in the Node.js REPL.

The behavior you report is less scary to me. 

As to the Functor behavior of map, I think it's not as complicated as it first looks.

`R.map` has the signature:

```hs
Functor f => (a → b) → f a → f b
```

For arrays, we can swap the `f` with `[]` to get:

```hs
(a → b) → [a] → [b]
```

For functions that receive some argument `r`, it becomes:

```hs
(a → b) → (r → a) → (r → b)
```

which is the same signature as `compose`.

> Personally I feel that async functions should be seen as (`a -> Promise b`) as they behave identically to my knowledge, but `Object.prototype.toString` leaks the implementation details of the function. IMO this is another point in `typeof`'s favour, because it sees them both as just `'function'`. :)

Once I take my foot out of my mouth and recognize that this code:

```js
```
 is branching on the functor and not the calling function, once I get hit over the head with the reason for this change, I can see nothing but good in it.

I'm going to wait for my head to settle while I look for comments from other member of @ramda/core, but I like this.

I'm not sure about `typeof`.  The reason that someone (kangax?) years ago brought out the `Object.prototype.toString.call(foo)` was because of problems with pretty much any other attempt to test a type.  I'd have to look at whether any of the issues would affect us here.

----------

Lesson learned: don't comment under the influence of narcotics, even ones prescribed by your oral surgeon!

source/type.js
```diff
@@ -24,6 +24,8 @@ import _curry1 from './internal/_curry1';
  *      R.type([]); //=> "Array"
  *      R.type(/[A-z]/); //=> "RegExp"
  *      R.type(() => {}); //=> "Function"
+ *      R.type(async () => {}); //=> "Async Function"
```
GingerPlusPlus:
```suggestion
 *      R.type(async () => {}); //=> "AsyncFunction"
```

source/type.js
```diff
@@ -24,6 +24,8 @@ import _curry1 from './internal/_curry1';
...
+ *      R.type(function* {}); //=> "Generator Function'"
```
GingerPlusPlus:
```suggestion
 *      R.type(function* () {}); //=> "GeneratorFunction'"
```

source/map.js
```diff
@@ -45,6 +45,8 @@ import keys from './keys';
 var map = _curry2(_dispatchable(['fantasy-land/map', 'map'], _xmap, function map(fn, functor) {
   switch (Object.prototype.toString.call(functor)) {
     case '[object Function]':
+    case '[object AsyncFunction]':
+    case '[object GeneratorFunction]':
```
GingerPlusPlus: Does not support `AsyncGeneratorFunction`.

source/internal/_isFunction.js
```diff
@@ -1,3 +1,6 @@
 export default function _isFunction(x) {
-  return Object.prototype.toString.call(x) === '[object Function]';
+  var type = Object.prototype.toString.call(x);
+  return type  === '[object Function]' ||
+    type === '[object AsyncFunction]' ||
+    type === '[object GeneratorFunction]';
```
GingerPlusPlus: Does not support `AsyncGeneratorFunction`.

source/invoker.js
```diff
@@ -27,6 +27,16 @@ import toString from './toString';
...
+ *
+ *      const dog = {
+ *        speak: async () => 'Woof!'
+ *      };
+ *      const person = {
+ *        speak: async () => 'Hi!'
+ *      };
+ *      const speak = R.invoker(0, 'speak');
+ *      speak(dog).then(R.identity) //=> 'Woof!'
+ *      speak(person).then(R.identity) //=> 'Hi!'
```
GingerPlusPlus: `.then(R.identity)` does literally nothing. I'd personally use `.then(console.log)` in the examples.

source/map.js
GingerPlusPlus:
```js
case '[object AsyncGeneratorFunction]':
```

willheslam:
Cheers @GingerPlusPlus, I really appreciate your feedback - I've made the changes you pointed out.

Hopefully now rjhilgefort's PR has been merged this one can go through too!

source/map.js
willheslam: Have added AsyncGeneratorFunction as an extra case, thanks for pointing this out!

source/type.js
willheslam: Great spot, this is now fixed

source/type.js
willheslam: This is now fixed

willheslam:
@CrossEye Yep, fixed - sorry, should have noticed this failed earlier!

I also did a bit of git rebasing to clean up the commits. This should be good to go now.
(The only thing it doesn't have is a test example in map for AsyncGeneratorFunctions, but the version of node I have doesn't support that yet.)

test/type.js
```diff
@@ -45,4 +45,15 @@ describe('type', function() {
...
 
+  it('"Function" if given a function', function() {
+    eq(R.type(() => {}), 'Function');
+  });
+
+  it('"AsyncFunction" if given an async function', function() {
+    eq(R.type(async() => {}), 'AsyncFunction');
+  });
+
+  it('"Generator Function" if given a generator function', function() {
```
CrossEye: Typo: extra space in "Generator Function"

test/map.js
```diff
@@ -31,6 +31,23 @@ describe('map', function() {
...
+        eq(h, (10 * 2) - 1);
+      });
+  });
+
+  it('interprets generator as a functor', function() {
+    var f = function(gen) { return [gen.next().value - 1, gen.next().value - 2, gen.next().value - 3]; };
+    var g = function* (b) { yield b * 1; yield b * 2; yield b * 3; };
+    var h = R.map(f, g);
+    eq(h(10), [(10 * 1) - 1, (10 * 2) - 2, (10 * 3) - 3]);
+  });
```
CrossEye: This test looks over-complicated to me.  Could we switch to two `yield`s/`next`s instead of three?

But it also makes me wonder whether a generator really is a functor.  Statefunesslscares me here.  But I'm a little foggy-brained (possible flu) to think it through right now.

willheslam: That's a good question... to my understanding, in terms of the way a generator behaves (ignoring any internal implementation details), a generator function is a pure function that returns an object that has state and an impure method (giving you different results each time you call .next()).
(a generator object: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator/throw )

In the test example, you could invoke `g` as many times as you like and get a new generator - it's no different to `g` returning an object that has a function that may or may not have a side effect.
`map` itself doesn't interact with the generator object in any way, it just passes it through to `f`.

I can simplify the test to just two `yield`s, but if you think the impure nature is distracting from the proof that to map, 'GeneratorFunctions' are equivalent to 'Functions', then it can be replaced altogether with just one, something like this:
```
    var f = function(gen) { return gen.next().value - 1 };
    var g = function* (b) { yield b * 2; };
    var h = R.map(f, g);
    eq(h(10), (10 * 2) - 1);
```
which doesn't have any typically generator-y behaviour, but gives it much more symmetry with the original function-as-functor test "interprets (async (->) r) as a functor" above.

CrossEye: I definitely prefer this alternate demonstration.

My brain is less foggy now (not much more than it's usual fogginess!)  And I still can't quite convince myself that generator functions obey the composition rule for functors.  But I'm also not coming up with a counterexample, so maybe I'm still more foggy than I'd like to think! :stuck_out_tongue_closed_eyes: 

Still, I this looks good to me.  Anyone have objections, @ramda/core?

source/map.js
scott-christopher: As this currently stands it is not a valid functor implementation because the provided `fn` is forced to receive a Generator of `a`, or Promise of `a` rather than accepting functions that work for any `a`.

I.e. it should be possible to do something like `R.map(a => a + 1, function*(a) { yield a * 2 })` 

By passing in the generator or promise, you’re providing more flexibility or power than functors alone should have access to.

We could however update it with their own seperate implementations with something like:
```
case '[object AsyncFunction]':
return async a => fn(await functor(a))

case '[object GeneratorFunction]':
return function*() {
  for (let a of functor()) yield fn(a);
}
```
The only downside of this is that it then makes JS runtimes with async and generator support a requirement to use Ramda, even if you don’t plan on using the functionality.

CrossEye: Yes, of course.  I was looking at the entirely wrong bit of rules for problems.

I like your alternative... but probably only for v1.0+

willheslam: This is interesting, however I wonder if treating AsyncFunctions (typeof fn === 'AsyncFunction') as separate from functions that return Promises (typeof fn === 'Function') could have unintended consequences.

It would mean babel can change your program's behaviour (e.g. if you start or stop transpiling, `map` will treat your transpiled [Special]Functions differently!), but also it breaks eta abstraction/reduction:
```
const f = tap(console.log)
const asyncFunction = async x => x

map(f, async x => x)(42) //"42"
map(f, x => asyncFunction(x))(42) //Promise<42>
//because
Object.prototype.toString.call(asyncFunction) //'[object AsyncFunction]'
Object.prototype.toString.call(x => asyncFunction(x)) //'[object Function]'
```
(This also extends to `compose`, `pipe` etc.)

In a typeless language like JS where we really can't know what a function will return and there's no way at runtime time to truly know if `x => f(x)` really is the same as `f` then I feel like we should rely on the extensional equality of values rather than the intensional properties that `Object.prototype.toString.call` sometimes exposes, but other times can't.

At the moment, Ramda already treats AsyncFunctions as sometimes separate entities.
In the transpiled ramda repl:
```
const asyncFn = async x => x
map(tap(console.log), asyncFn)(5) //[object Promise]
```
but in an ES6 native environment like node:
```
const asyncFn = async x => x
map(tap(console.log), asyncFn) //undefined
map(tap(console.log), asyncFn)(5)
//TypeError: R.map(...) is not a function
map(tap(console.log), x => asyncFn(x))(5)
//Promise { 5 }
map(tap(console.log), compose(asyncFn))(5)
//Promise { 5 }
```

My PR as it stands intentionally treats AsyncFunctions and GeneratorFunctions as just normal function functors, which is transpilation and eta-abstraction proof. Same goes for any other functor that has a consistent perceived type regardless of abstraction - e.g. arrays, objects or something that has a delegateable map e.g. eithers and maybes.

If we do need to treat AsyncFunctions and GeneratorFunctions (or, abstracted, functions that return Promises and functions that return Generators) differently when mapping, it seems safer to have a separate `map` (or `mapWith`...) that treats them consistently regardless of the environment.

test/map.js
willheslam: Thanks for the feedback - I have simplified the test with https://github.com/ramda/ramda/pull/2675/commits/f3cbb4be1843f3dd36dd1fcf119fba2af2b140ff

test/type.js
willheslam: Fixed with https://github.com/ramda/ramda/pull/2675/commits/b1056f4c1c9509d8417aae1f3c3a40bda135c8f5

willheslam:
@CrossEye I'm pretty confident this is up to scratch now, would be good to get your or @ramda/core's feedback. :)

CrossEye:
I'm still having mixed feelings about this.  I understand that right now the REPL is not doing what might be appropriate.  But that is a much smaller concern for me than making sure the code is right.  Even the problems with transpilers doesn't bother me so much.

I would really like it if the async functions were `mapped` as `(->) Promise`, that is, as though the type was `map :: (a -> b) -> (->) Promise a -> (->) Promise b`.  Treating it as the same type as `(->)`  seems likely to lead to problems.

Similarly, with generators, I would think this would be most appropriate: `map :: (a -> b) -> Generator a -> Generator b`.

But this is just a quick thought.  I really haven't dedicated any time to this.

willheslam:
Hello @CrossEye, any more thoughts on this? Does it need anyone else from the ramda team to have a closer look?

I'm personally still feeling that Async Functions should be treated the same as Functions (that return Promises), in much the same way that a Function shouldn't be treated differently if it is wrapped inside another function, e.g.
```
function foo(x) { ... }

function bar(x) { return foo(x) }
```
It seems to me any ramda operation (e.g. `map`) on foo should be identical to an operation on bar (they retain the same arity, for example).
In much the same way it would be for compose:
```
function foo(x){ ... }

const bar = compose(foo)
```

The same is, in my opinion, true of an Async Function or a Generator Function (i.e. a function that *returns a generator object*, not a function that has implicit state):
```
async function foo(x) { ... }

function bar1(x) { return foo(x) }
const bar2 = compose(foo)
```

In the same way, I expect the following function foo to have exactly the same behaviour as bar and baz (which wraps an async function, but no calling code can tell that), and be treated in the same way by `map`.

```
function foo() {
  return Promise.resolve(42)
}

async function bar() {
  return 42
}

function baz() {
  return bar()
}
```

I am open minded, but also keen to see this go in - interested to know what you or the other ramda maintainers think. :)

CrossEye:
I did drop this and entirely forget it.  Sorry.  I think it was because every time I tried to think about it my brain hurt.  It still does.

I wonder if some of the smart people who occasionally hang around here have advice on this?

@ramda/core? @buzzdecafe? @davidchambers? @DrBoolean? @joneshf? @GingerPlusPlus? @Avaq?

Briefly, the issue is how we treat the JavaScript types such as AsyncFunction and GeneratorFunction as functors.  The [comment above][sc] from @scott-christopher makes it clear that if we want proper parametricity, then just as we have

```js
map(square, n => n + 1)          //=> n => square(n + 1)
```

we should also have

```js
map(square, async n => n + 1)    //=> async n => square(n + 1) 
```

and 

```js
const foo = function * () {yield 1; yield 2;}
map(square, foo)                 //=> function * () {yield square(1); yield square(2);}
```

But as @willheslam [points out][wh], not treating these like any other function leads to odd contradictions in equational reasoning, in which `x => foo(x)` acts differently from `foo`.

Is there a good solution to this?




  [sc]: https://github.com/ramda/ramda/pull/2675/files#r249646227
  [wh]: https://github.com/ramda/ramda/pull/2675#discussion_r249889371

willheslam:
...and at the risk of further muddying the waters, here are the following interpretations of `Object.prototype.toString` of various manipulations of some async functions:
```
async function foo(a, b){
  return a + b
}

const fooA = curry(foo)
const fooB = fooA(42)

Object.prototype.toString.call(foo) //"[object AsyncFunction]"
Object.prototype.toString.call(fooA) //"[object Function]"
Object.prototype.toString.call(fooB) //"[object Function]"
Object.prototype.toString.call(R.binary(foo)) //"[object Function]"

const bar = a => async b => {
  return a + b
}

Object.prototype.toString.call(bar) //"[object Function]"
Object.prototype.toString.call(bar(42)) //"[object AsyncFunction]"
Object.prototype.toString.call(R.uncurryN(2, bar)) //"[object Function]"
Object.prototype.toString.call(R.uncurryN(2, bar)(42)) //"[object Function]"
```

(This won't work in the ramda repl which transpiles the AsyncFunction away into a Function, so they all become "[object Function]", but will work in the Chrome Dev Tools console on the same page.)

The point being as it is, to achieve the parametricity, Ramda would need to treat Async/GeneratorFunctions consistently throughout the entire library, and even then, any non-ramda abstraction (e.g. wrapping in a `(a, b) => foo(a)(b)`) would break this - more of the contradictory equational reasoning issues. :)

joneshf:
Sorry, I don't know enough about how async functions work in JS. But the from the reasoning above, it sounds like this isn't going to work. I defer to someone else to help out here.

willheslam:
For what it's worth @joneshf and @CrossEye, I don't think you need to know a whole lot about async functions when considering this PR.

This is the best way I can explain it. It'll be a long one, but I've been thinking about this a lot and by not missing anything out I hope I can make my perspective as clear as possible, and any mistakes or errors can be pointed out:

We agree `map` takes a (mapping) function and a functor.
When you map over a functor, you get back the same kind of functor with the contents modified in some way - i.e. an array is mapped into another array, an observable is mapped into another observable.

When the functor is a *function*, the contents of that functor is the *result of the function* - and you are mapping a function into another function. 

```
const doubleAndIncrement = map(x => x + 1, x => x * 2)
// map `x => x * 2` into a new function that also increments the result by 1

doubleAndIncrement(5) // 11
```
[repl link
](https://ramdajs.com/repl/#?const%20doubleAndIncrement%20%3D%20map%28x%20%3D%3E%20x%20%2B%201%2C%20x%20%3D%3E%20x%20%2A%202%29%0A%2F%2F%20map%20%60x%20%3D%3E%20x%20%2A%202%60%20into%20a%20new%20function%20that%20also%20increments%20the%20result%20by%201%0A%0AdoubleAndIncrement%285%29)

So far we all agree.
We also, I hope, agree that the function we're mapping over might also return a functor itself - for example, `of` is a function that takes a value and returns an array with that value inside.

```
const contrived = map(concat([1, 2]), of)
// we've mapped of into a new function that prepends [1, 2] to the result 

contrived(3) // [1, 2, 3]
```
[repl link
](https://ramdajs.com/repl/#?const%20contrived%20%3D%20map%28concat%28%5B1%2C%202%5D%29%2C%20of%29%0A%0Acontrived%283%29)

Note that the mapping function `concat([1, 2])` is applying to the *contents* of `of` - that is, the array - not the *contents of the **contents*** of `of`, the integer. If we want to access the *contents of the **contents*** of `of`, we use nested map:

```
const contrived = map(map(add(1)), of)
// we're mapping over the contents of the contents of `of`, the integer

contrived(3) // [4]
```
[repl link](https://ramdajs.com/repl/#?const%20contrived%20%3D%20map%28map%28add%281%29%29%2C%20of%29%0A%0Acontrived%283%29)

This is exactly like mapping over an array of arrays (a functor of functors).
If you want to manipulate each inner array, you use a single map.
If you want to manipulate each value in the inner array, you use a nested map.
```
const single = map(concat([1, 2]), [[3, 4, 5], [6, 7, 8]])
const nested = map(map(add(1)), [[3, 4, 5], [6, 7, 8]])

const result = {
  single,
  nested
}
result // {"nested": [[4, 5, 6], [7, 8, 9]], "single": [[1, 2, 3, 4, 5], [1, 2, 6, 7, 8]]}
```
[repl link](https://ramdajs.com/repl/#?const%20single%20%3D%20map%28concat%28%5B1%2C%202%5D%29%2C%20%5B%5B3%2C%204%2C%205%5D%2C%20%5B6%2C%207%2C%208%5D%5D%29%0Aconst%20nested%20%3D%20map%28map%28add%281%29%29%2C%20%5B%5B3%2C%204%2C%205%5D%2C%20%5B6%2C%207%2C%208%5D%5D%29%0A%0Aconst%20result%20%3D%20%7B%0A%20%20single%2C%0A%20%20nested%0A%7D%0Aresult)

That is to say, from map's perspective, any function that returns a functor is the same as a functor of functors.

You want `map(mappingFn, functionAsFunctor)` to be applying `mappingFn` to the *contents* of `functionAsFunctor`- its result.

You definitely *don't* want to be applying `mappingFn` to the *contents of the **contents*** of `functionAsFunctor`, in the same way you don't want `map(mappingFn, arrayOfArrays)` to be
applying `mappingFn` to the contents of the inner most array.
For that, you want to use `map(map(mappingFn), arrayOfArrays)`.

We all agree on the above, and that's exactly how Ramda's map works today.

The current map implementation also currently treats a Function that returns a Promise the same as an Array of Promises.

I'm now going to argue that an Async Function should be treated by map just as a Function that returns a Promise (that is to say, a Functor of a Functor). Just like an Array of Promises.
This is what this PR addresses: treating Async Functions (+ Generator Functions) as Functions.

This is what it would look like:
```
const contrived = map(x => Promise.all([Promise.resolve(1), Promise.resolve(2), x]), async () => 3)

contrived().then(console.log) // logs: [1,2,3]
```
[repl link](https://ramdajs.com/repl/#?const%20contrived%20%3D%20map%28x%20%3D%3E%20Promise.all%28%5BPromise.resolve%281%29%2C%20Promise.resolve%282%29%2C%20x%5D%29%2C%20async%20%28%29%20%3D%3E%203%29%0A%0Acontrived%28%29.then%28console.log%29)

This is, roughly speaking, the same as the earlier example using `of`:
```
const contrived = map(concat([1, 2]), of)

contrived(3) // [1, 2, 3]
```

If Ramda adopted @CrossEye's example above of parametricity:
```
map(square, async n => n + 1)    //=> async n => square(n + 1) 
```
then doing valid and useful operations like concating Promises:
```
map(x => Promise.all([Promise.resolve(1), Promise.resolve(2), x]), async () => 3)
```
would be impossible. Furthermore, it would be, IMO, *incorrect*
because the contents of `async n => n + 1` is *not* `n + 1`.
The contents of `async n => n + 1` is `Promise<n + 1>`,
the same way it is for `n => Promise.resolve(n + 1)`.

Ultimately if this was true:
```
map(square, async n => n + 1)    //=> async n => square(n + 1) 
```
it would also be like the equivalent of the following:
```
map(square, [[1, 2, 3], [4, 5, 6]])    //[[1, 4, 9], [16, 25, 36]]
```
Which instead is, rightly, a runtime error!
What you really need is `map(map(square), [[1, 2, 3], [4, 5, 6]]))`.

The right format for @CrossEye 's example would, IMO, actually be:
```
const square = x => x * x
const squareAsync = map(then(square), async n => n + 1)

squareAsync(4).then(console.log) // logs: 25
```
[link to repl](https://ramdajs.com/repl/#?const%20square%20%3D%20x%20%3D%3E%20x%20%2A%20x%0Aconst%20squareAsync%20%3D%20map%28then%28square%29%2C%20async%20n%20%3D%3E%20n%20%2B%201%29%0A%0AsquareAsync%284%29.then%28console.log%29)
(this only works currently in the ramda repl because the JS is transpiled to ES5 - and babel transpiles async functions, correctly, into functions that return promises - which, then, IMO, Ramda treats in the right way.)

If Promises could be mapped over (which is a separate idea) then you could get away with simply:
```
const squareAsync = map(map(square), async n => n + 1)
```


The same goes for Generator Functions.
In the example above:
```
const foo = function * () {yield 1; yield 2;}
map(square, foo)        //=> function * () {yield square(1); yield square(2);}
```

The contents of `foo` is *not* `1, 2`, the contents of `foo` is a [Generator object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator) that just so happens to generate `1, 2` when iterated over
(i.e. if a generator is a functor, then `foo` is not a Functor of (1, 2), it is a Functor of a Functor of (1, 2)).

The mapping function should be able to transform or alter the *Generator* returned by the Generator Function in the same way we should be able to transform or alter the Promise returned by an Async Function.
For example wrapping the generator in another generator.
I cannot represent this faithfully in the current Babel repl, as Ramda's internal/_map.js is mangling GeneratorFunctions into empty arrays, so I have used an Array of a Generator *Object*, which is, according to my argument, equivalent to a Generator Function.
```
const contrived = map(x => function *(){
     yield* [1, 2]
     yield* x
  },
  [(function*() { 
    yield 3;
  })()])

const generator = contrived[0]()

console.log(generator.next().value); // logs: 1
console.log(generator.next().value); // logs: 2
console.log(generator.next().value); // logs: 3
```
[link to repl](https://ramdajs.com/repl/#?const%20contrived%20%3D%20map%28x%20%3D%3E%20function%20%2A%28%29%7B%0A%20%20%20%20%20yield%2A%20%5B1%2C%202%5D%0A%20%20%20%20%20yield%2A%20x%0A%20%20%7D%2C%0A%20%20%5B%28function%2A%28%29%20%7B%20%0A%20%20%20%20yield%203%3B%0A%20%20%7D%29%28%29%5D%29%0A%0Aconst%20generator%20%3D%20contrived%5B0%5D%28%29%0A%0Aconsole.log%28generator.next%28%29.value%29%3B%0Aconsole.log%28generator.next%28%29.value%29%3B%0Aconsole.log%28generator.next%28%29.value%29%3B)

If Generators could be mapped over (which is a separate idea) then @CrossEye 's example of:
```
const foo = function * () {yield 1; yield 2;}
map(square, foo)       //=> function * () {yield square(1); yield square(2);}
```

would be actually realised as:
```
const foo = function * () {yield 1; yield 2;}
map(map(square), foo)          //=> function * () {yield square(1); yield square(2);}
```


This brings me back to @scott-christopher's comment [above](https://github.com/ramda/ramda/pull/2675/files#r249646227):

> By passing in the generator or promise, you’re providing more flexibility or power than functors alone should have access to.

I believe this idea has confused which functor we are referring to.
I completely agree with him that, if you could, mapping over a Generator or a Promise should not be able to modify the Generator or the Promise itself - it should just modify the contents. This should be the same as mapping over an Array, or Object, etc - any Functor.

However, when calling `map(fn, function*() { ... })` or `map(fn, async () => { ... })`, you are not mapping over a Generator or a Promise.

You are mapping over a Function that's contents just so happen to be another Functor - in the same way mapping over a Function that returns an Array (`map(fn, of)`) or an Array of Arrays (`map(fn, [[1, 2, 3]]`) gives you access to the inner Array, not the *contents* of that Array.

That's why in the mapping function, you can only change the results of the function, not the properties of the function itself (e.g. swapping it out for a different one, changing its arity, etc.).

His suggestion of
```
return async a => fn(await functor(a))
```
and
```
return function*() {
  for (let a of functor()) yield fn(a);
}
```
works well for mapping directly over a Generator Object or a Promise, where you *should* only have access to the contents, but it doesn't make sense for a Generator Function or an Async Function.
That alternative proposal as it stands would be like requiring `map(fn, [[1, 2, 3], [4, 5, 6]])` to always behave as `map(map(fn), [[1, 2, 3], [4, 5, 6]])`, which is incorrect, and precludes all the things you'd naturally want to do - such as `map(filter(fn), [[1, 2, 3], [4, 5, 6]])`. (imagine the generator/promise-equivalents)

This has been a long post and I think I've described my perspective as clearly as I can. :) Thanks for staying with me if you got this far!

Not only do I believe that it is more correct to treat Async and Generator Functions the same as normal Functions, but it is more congruent with the rest of Ramda (because it requires no changes to `curry`, `compose`, or any other higher order functions that would need to propagate the 'function type' through, unlike with the alternative proposal).

It's also, IMO, more idiomatic with JS in general:
If you wrap an Async/GeneratorFunction in an `x => foo(x)`, namely, standard eta abstraction, then you can no longer get the same results from `Object.prototype.toString`, despite it being behaviourally exactly the same. 
Async/GeneratorFunctions even `typeof` to "Function", just like Functions.

Lastly, it provides less surprising behaviour with Babel and other transpilers - right now, Ramda treats transpiled AsyncFunctions correctly, but provides nonsense values (`[undefined]`) if you remove Babel. My PR ensures the behaviour is consistent and backwards compatible.

The alternative proposal as it stands by @scott-christopher, treating AsyncFunctions and GeneratorFunctions not as higher order functors (i.e. functors that contain functors), but just as standard functors, not only changes IMO, the expected behaviour of map (to only map one level deep), but also could potentially require pervasive changes to Ramda's library of functions to propagate `toString` behaviour, breaks vanilla JS eta abstraction/reduction, and would introduce incompatible behaviour between transpiling/non-transpiling. I think his suggestion works very well for mapping over Generator Objects and Promises directly, though, and IMO could be addressed in a future PR or version of Ramda.

paldepind:
I find your arguments convincing @willheslam. I think that you've at the very least shown your proposal to, indeed, be a valid function implementation that doesn't violate "parametricity" or anything else.

However, after struggling to understand @scott-christopher's point and after having given this some thought I actually think that there are two valid ways to look this problem:

**Perspective number one**: An async function is the same as a function that returns a promise `Function<A, Promise<B>>` (TypeScript like notation) or as a `a -> Promise b` (Haskell like notation). With this point of view the behavior implemented by this PR is the only right way to implement things. Mapping over a `Function<A, B>` should pass a `B` to the mapping function and hence mapping over a `Function<A, Promise<B>>` should pass a `Promise<B>` to the mapping function.

**Perspective number two**: An async function is a completely separate type: `AsyncFunction<A, B>` or `AsyncFunction a b`. With this perspective mapping over an async function should not pass a promise to the mapping function, it should pass the plain value returned inside the async function. With this point of view one a naturally lead to the behavior [illustrated above](https://github.com/ramda/ramda/pull/2675#issuecomment-496758156) by @CrossEye 

To me, both of these seems to be completely valid functor implementations. Hence, I think the choice between them should be justified by what is most useful and what makes most sense to us JS developers.

To sum it up I believe the question is this: Is it most useful and natural to consider an async function to be a `a -> Promise b` or to consider it being a `AsyncFunction a b`.

CrossEye:
@willheslam:

That was an excellent exposition of your suggestion.

I agree 100% with @paldepind, though, that even with that, this comes down to the basic question of what Functor we want to expose with Async/Generator functions.  That is why I've struggled so hard with this.

If it weren't for your point about `x => asyncFoo (x)` ending up different from `asyncFoo`, and to a lesser extent the REPL/Babel issues, I think I would come down on the other side, what @paldepind calls `AsyncFunction a b`.  But these issues are swinging me toward your view.

My trouble is that the other behavior is the one that I would personally choose.  And I'm wondering if the best be would be to accept this PR but also create functions like `mapAsync`/`mapGenerator` with the alternative behavior.

willheslam:
@CrossEye  I've created https://github.com/ramda/ramda/pull/2863 as a speculative implementation of your suggestion to create `mapAsync`/`mapGenerator` functions... it would be good to know what you think, and whether or not it relieves this PR of the need do to anything but have `R.map` treat all types of functions as "Function Functors".

I don't expect that new PR to be merged any time soon - as you mentioned earlier, it would place new minimum runtime requirements on Ramda, plus it raises a few more questions than it answers!

Thanks for all your help and patience with this so far, it seems these problems don't have fully satisfactory answers, but do have lasting ramifications, especially when considering backwards compatibility.

This stuff feels like we're dancing on the often awkward boundary between concepts developed in and facilitated by languages with complex type systems, and the flexible but rather stark environment that JS gives us, and I'm cognisant of Ramda's focus on JS-centric ideas, not porting the Haskell standard library!

CrossEye:
Postponing until after 1.0.  There are some really good ideas here, and in #2863.

PR #2676, opened 2018-10-22
Added new function "evolveTo" that returns new object of transformations

allexysleeps:
Hi, Ramda Team. 
I'm often facing issues when I need to create, a new object from other object or any other data.
And I thought that that would be nice to have something like evolve but for new objects using the transformations map approach from evolve func.

I would be happy if you would like this function and add it to the library. I have some more ideas on evolving function. Let me know if you are interested in further developing in this direction.

Please let me know if there are needed some improvements from my side. 

Best regards, Alex.

### DIFF
--- a/source/evolveTo.js
+++ b/source/evolveTo.js
@@ -1,0 +1,45 @@
+import _curry2 from './internal/_curry2';
+
+
+/**
+ * Creates a new object by recursively applying transformation's map.
+ * It calls transformation functions with source data, which can take any form,
+ * and put their response in corresponding keys of the new object
+ *
+ * The new object will contain only mentioned keys, that may not be present in source `object`
+ * If transformation won't be a `function` or an `object` the key will be set to `null`
+ *
+ * @func
+ * @memberOf R
+ * @category object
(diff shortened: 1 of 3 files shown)


foxbunny:
What's the difference between this function and `applySpec()`?

Taking your example from tests:

```
var transf = {
  sum: R.prop('total'),
  values: R.prop('numbers'),
  multi: R.compose(R.reduce(R.multiply, 1), R.prop('numbers'))
}
var src = {total: 10, numbers: [1, 4, 2, 3]}

R.applySpec(transf)(src)
// => {multi: 24, sum: 10, values: [1, 4, 2, 3]}
```

It looks to be the same outcome. Am I missing something?

allexysleeps:
@foxbunny  You are right it looks kinda the same. I just couldn't find it 😞.
Maybe it should be added [https://github.com/ramda/ramda/wiki/What-Function-Should-I-Use%3F](here).
Naming is a little bit confusing as always in programming 😃.

Ok, I think this should be closed 😃.
Thanks for fast answer.

CrossEye:
I doubt anyone particularly likes that name.  Suggestions are welcome!

PR #2677, opened 2018-10-23
remove mentions of .js extension from imports

vsimko:
In the file `source/pipeWith.js`, somebody forgot to remove the explicit .js extensions from the imports.


CrossEye:
Good catch!  Thanks!

PR #2591, opened 2018-07-17
Add js extrension to imports and named re-exports in es build

GingerPlusPlus:
Fixes #2587.

Unlike #2588, it doesn't introduce conflicts with pending pull requests (which use no extension in imports and named exports).

PR #2678, opened 2018-10-26
support native modules in node when using `.mjs` extension

vsimko:
This pull request fixes the problem when using native modules from node (`--experimental-modules`) or `esm` module loader together with `.mjs` extensions.
```js
// a sample script `01.mjs`
import {length} from 'ramda'
console.log(length("hello world"))
```

Here are some examples of what didn't work before this pull request:

### Using esm module loader
```console
$ node -r esm 01.mjs
SyntaxError: Missing export name 'length' in ES module:
    file:///.../node_modules/ramda/src/index.js
```

### Using node --experimental-modules 
```console
$ node --experimental-modules 01.mjs
(node:18860) ExperimentalWarning: The ESM module loader is experimental.
SyntaxError: The requested module does not provide an export named 'length'
    at ModuleJob._instantiate (internal/loader/ModuleJob.js:85:17)
    at <anonymous>
```

After this fix, the following works:
```console
$ node --experimental-modules 01.mjs
(node:19885) ExperimentalWarning: The ESM module loader is experimental.
11
```
```console
$ node -r esm 01.mjs
11
```


CrossEye:
This makes sense to me.  Anyone else, @ramda/core ?

GingerPlusPlus:
Incompatible with #2600, closes #2600.

As long as Ramda isn't planning to acquire any dependencies, I'd rather completely get rid of CJS build (#2600) and alter ES build to work in both browser in Node.

GingerPlusPlus:
Even tho it's hacky and I still don't like it, it is the simplest solution.
Feel free to proceed, sorry for holding you back.

Also about ESM, @CrossEye, please get to #2591.

martinmckenna:
Hey @vsimko I'm having some issues because of the built `.mjs` file. Apparently Jest cannot parse `.mjs` files at all, so I'm getting errors such as:

```
/node_modules/ramda/src/index.mjs:1
 ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){export { default as F } from './F.js';
                                                               ^^^^^^
SyntaxError: Unexpected token export
    8 | var _axios = _interopRequireDefault(require("axios"));
    9 |
 > 10 | var _ramda = require("ramda");
```

I'm wondering what was the main motivation behind this and if there is anyway to get a commonjs version of index.js?

ivan-kleshnin:
I'm trying Ramda with Node 12 experimental modules.

**package.json**

```json
{
  "type": "module"
}
```

**terminal**
```
$ node --experimental-modules xxx.js
```

All Ramda exports end up under the `default` key:

**xxx.js**
```js
import * as R from "ramda"

console.log(R) // {default: {==ramda exports==}}
// instead of {==ramda exports==}
```

Am I doing something wrong?

CrossEye:
I'm afraid I don't have any expertise here.  I expect to be looking at it over the next few months, but I don't know enough now to offer an opinion.  Anyone else?

Issue #2681, opened 2018-10-30
applySpec returns undefined when 'map' key is specified 


Maushundb:
When  `applySpec` is given an object with the key `map`, `[undefined]` is returned:

```
R.applySpec({
  map: R.always(1)
})({a: 1}); 
// [undefined]
```

https://ramdajs.com/repl/?v=0.25.0#?R.applySpec%28%7B%0A%20%20map%3A%20R.always%281%29%0A%7D%29%0A%28%7Ba%3A%201%7D%29%3B

Bradcomp:
It looks like this is desired behavior for `map`, in that it dispatches to the `map` method of the second parameter if present. It seems like it would be very undesirable behavior for `applySpec`, so I'd say it's definitely a bug / abstraction leak due to `applySpec` using the public `map` function.

Would you be interested in trying your hand at a fix and opening a PR?

Maushundb:
Yeah sure, let me see what I can do.

Maushundb:
https://github.com/ramda/ramda/pull/2682

PR #2682, opened 2018-10-30
Fix err with applySpec "map" keys

Maushundb:
## Summary
This PR addresses the issue described at https://github.com/ramda/ramda/issues/2681. `applySpec` was using ramda's `map` function, which calls out to a functions `.map` method if it is present. When a spec object with a `map` key was defined, it would think the value was a function it needed to call out to, producing undefined behavior. 

## Test Plan
TDD - Added a failing unit test. Wrote some code. Unit test passes.

### DIFF
--- a/source/applySpec.js
+++ b/source/applySpec.js
@@ -1,12 +1,20 @@
 import _curry1 from './internal/_curry1';
 import apply from './apply';
 import curryN from './curryN';
-import map from './map';
 import max from './max';
 import pluck from './pluck';
 import reduce from './reduce';
+import keys from './keys';
 import values from './values';
 
+// Use custom mapValues function to avoid issues with specs that include a "map" key and R.map
+// delegating calls to .map
+function mapValues(fn, obj) {
+  return keys(obj).reduce(function(acc, key) {
+    acc[key] = fn(obj[key]);
+    return acc;
+  }, {});
+}
 
 /**
  * Given a spec object recursively mapping properties to functions, creates a
@@ -34,12 +42,16 @@ import values from './values';
  * @symb R.applySpec({ x: f, y: { z: g } })(a, b) = { x: f(a, b), y: { z: g(a, b) } }
  */
 var applySpec = _curry1(function applySpec(spec) {
-  spec = map(function(v) { return typeof v == 'function' ? v : applySpec(v); },
-             spec);
-  return curryN(reduce(max, 0, pluck('length', values(spec))),
-                function() {
-                  var args = arguments;
-                  return map(function(f) { return apply(f, args); }, spec);
-                });
+  spec = mapValues(
+    function(v) { return typeof v == 'function' ? v : applySpec(v); },
+    spec
+  );
+
+  return curryN(
+    reduce(max, 0, pluck('length', values(spec))),
+    function() {
+      var args = arguments;
+      return mapValues(function(f) { return apply(f, args); }, spec);
+    });
 });
 export default applySpec;
--- a/test/applySpec.js
+++ b/test/applySpec.js
@@ -21,6 +21,10 @@ describe('applySpec', function() {
        { unnested: 0, nested: { sum: 3 } });
   });
 
+  it('works with a spec defining a map key', function() {
+    eq(R.applySpec({map: R.prop('a')})({a: 1}), {map: 1});
+  });
+
   it('retains the highest arity', function() {
     var f = R.applySpec({ f1: R.nAry(2, R.T), f2: R.nAry(5, R.T) });
     eq(f.length, 5);
@@ -29,5 +33,4 @@ describe('applySpec', function() {
   it('returns a curried function', function() {
     eq(R.applySpec({ sum: R.add })(1)(2), { sum: 3 });
   });
-
 });


source/applySpec.js
```diff
@@ -1,12 +1,19 @@
...
 import values from './values';
+import fromPairs from './fromPairs';
+import toPairs from './toPairs';
 
+// Use the built-in map function to avoid issues with specs what include a "map" key
+function mapValues(fn, obj) {
+  return fromPairs(toPairs(obj).map(function(pair) {
+    return [pair[0], fn(pair[1])];
+  }));
+}
```
CrossEye: I wonder if we shouldn't use an implementation more like [what's in `map`][ma], but with the native `reduce`:

```js
function mapValues(fn, obj) {
  return keys(obj).reduce(function(acc, key) {
    acc[key] = fn(obj[key]);
    return acc;
  }, {})
}
```
Ramda tries not to build on its external functions too much, preferring somewhat ugly imperative code to improve performance.  We would use `keys` here, as Ramda through 1.0 is supporting older engines that don't necessarily have `Object.keys`.  But if we can avoid `fromPairs`/`toPairs`, it would (presumably) be faster.

   [ma]: https://github.com/ramda/ramda/blob/master/source/map.js#L51-L55

source/applySpec.js
```diff
@@ -1,12 +1,19 @@
...
-import map from './map';
 import max from './max';
 import pluck from './pluck';
 import reduce from './reduce';
 import values from './values';
+import fromPairs from './fromPairs';
+import toPairs from './toPairs';
 
+// Use the built-in map function to avoid issues with specs what include a "map" key
+function mapValues(fn, obj) {
```
Bradcomp: Instead of writing our own function here, can you use the same process `map` does for mapping Objects: https://github.com/ramda/ramda/blob/master/source/map.js#L52?

test/applySpec.js
```diff
@@ -21,6 +21,10 @@ describe('applySpec', function() {
        { unnested: 0, nested: { sum: 3 } });
   });
 
+  it('works with a spec defining a map key', function() {
+    eq(R.applySpec({map: R.always(1)})({}), {map: 1});
```
Bradcomp: I'd like it if this test actually used the value being passed in so we can verify the behavior is as expected.

Something like `R.applySpec({map: R.prop('a')})({a: 1})`

source/applySpec.js
CrossEye: The only problem with that is that it depends upon `_reduce`, which adds transducing behaviors that I don't think we want here.

Maushundb: Happy to do something along the lines of 
```
function mapValues(fn, obj) {
    acc[key] = fn(obj[key]);
    return acc;
  }, {}, keys(obj));
}
```

Unless there's some transducing behavior here that we don't want?

Maushundb: Ah, comment below works as well

CrossEye:
Thank you very much, @Maushundb!

jordie23:
Hello,

So turns out, previous to this, `applySpec` worked seemlessly with arrays/lists, not just objects. Since this change, it has broken that. Just wondering whether `applySpec` should be supporting arrays/lists or whether that was accidental previously? This has broken some of our code - not a problem, we can update it - but just wondering whether we were relying on a quirk in the code or something that was actually supposed to be supported?

e.g.

```const kvs = [
  {
    Key: 'Tag1',
    Value: 'TestVal1'
  },
  {
    Key: 'Tag2',
    Value: 'TestVal2'
  },
  {
    Key: 'Tag3',
    Value: 'TestVal3'
  }
]

applySpec([map(prop('Key')), map(prop('Value'))])(kvs)

// result < 0.26

[ 
  [ 'Tag1', 'Tag2', 'Tag3', 'Tag4' ],
  [ 'TestVal1', 'TestVal2', 'TestVal3', 'TestVal4' ]
]
  
// result >= 0.26

{
  "0": [ 'Tag1', 'Tag2', 'Tag3', 'Tag4' ],
  "1": [ 'TestVal1', 'TestVal2', 'TestVal3', 'TestVal4' ]
}
```

chearmstrong:
We have the same issue as @jordie23 - We are using `R.applySpec` for arrays in a lot of our code.

What's the best way to update our code so that we can upgrade?

CrossEye:
Somehow I don't recall seeing @jordie23's comment.

This is definitely something we should fix.  Anyone have a suggested change?

Maushundb:
Hmm it feels like we're trying to overload `.applySpec` to work on too many input types - objects, arrays, ...args. A bandaid could be to do a test for `isArray(input) ? delegate to .map : do what we're doing`. I'm not sure what ramda's philosophy is on input type overloading, but having separate methods for `applyObjSpec` vs `applyArraySpec` etc could be at least worth consideration by the maintainers imho. Although it is nice for it to "just work" with whatever you give it.

CrossEye:
One main reason for fixing this would be to make it consistent with things like `assoc`/`assocPath`, as well as `map`, `filter`, and a number of others that work on arrays and objects equally well.  JS treats arrays as objects with some slightly special handling around the `length` parameter.  I think when we can, it would be nice to do so too.

If no one gets to it before then, I'll try to create a PR next weekend.  It should be an easy fix.

PR #2685, opened 2018-11-01
Add relevant 'see' to always

jamesmarlowe:


CrossEye:
I don't understand the reason for this.  Could you please explain *why*?

While `T` and `F` are implemented in terms of `always` (and probably shouldn't be, in fact), their API, such as it is, has little to do with `always`.

jamesmarlowe:
My thinking was definitely influenced by their implementation, but further confirmed by `always` inclusion in the `T` and `F` "see" sections.

As they both "always" return true and false respectively, it seemed appropriate that upon seeing always for the first time, you'd be pointed in the direction of the two functions which always return something. Searching for "A function that always returns" gives these three functions and the other two already reference the others. I only meant to make them consistent.

foxbunny:
> always inclusion in the T and F "see" sections

Yeah, I always (pun intended) thought it weird that T and F reference always but not the other way around.

CrossEye:
I don't have a major objection, but I still don't see the point.

I think the `see also`-graph could reasonably be directed and not always bidirectional.  But part of this is that I don't think the `T` -> `always` link should be included.  (And I'm in favor of many *more* links between functions!)  I don't like having the implement dictate such things.

Moreover, I think we should rewrite `T` and `F` to lose the dependency.  This is plenty simple:

```js
var T = function T() {return true;}
```

And after 1.0, when we embrace modern syntax, this will be simpler still:

```js
const T = () => true
```

So I'd rather move away from these links.  But again, I don't feel strongly about this.

jamesmarlowe:
I can get behind that idea as well, would this be more appropriate? https://github.com/ramda/ramda/pull/2686

CrossEye:
Yes, thank you!

PR #2686, opened 2018-11-01
simplify T and F

jamesmarlowe:
Following discussion in https://github.com/ramda/ramda/pull/2685 simplifying the implementation of `R.T` and `R.F` to be self contained function definitions and removing link to always

### DIFF
--- a/source/F.js
+++ b/source/F.js
@@ -1,4 +1,3 @@
-import always from './always';
 
 
 /**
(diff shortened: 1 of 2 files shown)


CrossEye:
Thank you!

def14nt:
Why hasn't this been merged? Should be that tiny bit faster.

CrossEye:
> Why hasn't this been merged?

Because we're slow around here.  Merged now.

PR #2687, opened 2018-11-01
Add dispatching to R.of. Fixes #2502.

### LINKED ISSUE #2502: Rename/alias R.of --> R.listOf
For me personally [`R.of`](http://ramdajs.com/docs/#of) has a confusing name. Every time I meet it in code I spend extra cognitive payload on it. Unlike `R.objOf` that is rather clear.

What do you think about renaming/aliasing it to `R.listOf` / `R.toArray`?

I've read that ramda [should be 100% without aliases](https://github.com/ramda/ramda/pull/574#issuecomment-64129603). I understand that this change can break existing codebase. But, from the other side, I believe that clear function naming is an important point. Especially for newcomers.

ben-eb:


CrossEye:
Thank you.  I'll need to spend a little time on this.

This is a breaking change; it's probably worth doing, but I need to think through the consequences.

CrossEye:
We don't do breaking changes without serious thought, though.

def14nt:
Make a separate method then everyone's happy @CrossEye :)

CrossEye:
@def14nt:

Well, we don't do *methods* around here, but plain *functions*.  :smile: 

We don't add functions willy-nilly.  We try to be cautious about our API.  There is little reason for two functions that do the same thing.  But `of` has been a thorn in our side for a while.  It does need fixing.  `of` is not a particularly widely-used function.  But it could be if written correctly. This is a function that we want to conform to the [FantasyLand specification][fl].  And FantasyLand offers all sorts of generic ways to work with types.  That is why we want to do this.  We want to be cautious about it because... well, if nothing else, we really should long ago have been at 1.0, as people are depending upon this in production.



customcommander:
Hi @ben-eb 

We recently revisited this issue and we concluded that we would welcome that change. If you don't mind reviving this pull request I can guarantee you that you'll get a fair and quick review. Sorry it took so long!

Thank you.

customcommander:
For the record: since the branch for that pull request no longer exists I had to redo one from scratch and I have almost copy/pasted this (minus a few changes).

PR #2688, opened 2018-11-02
Add R.splitWith

jamesmarlowe:
Finishing work started in https://github.com/ramda/ramda/pull/2620

### DIFF
--- a/source/splitWith.js
+++ b/source/splitWith.js
@@ -15,24 +15,23 @@ import _reduce from './internal/_reduce';
  * @return {Array}
  * @example
  *
- *      R.splitWith(R.equals('a'), [1, 2, 3, 1, 2, 3]); //=> [[1], [3, 1], [3]]
+ *      R.splitWith(R.equals(2), [1, 2, 3, 4, 2, 1]); //=> [[1], [3, 4], [1]];
+ *      R.splitWith(R.equals('a'), 'bananas'); //=> [['b'],['n'],['n'],['s']]
  */
+
 var splitWith = _curry2(function splitWith(pred, list) {
-  let currIndex = 0;
-  return _reduce(function(acc, curr) {
-    if (pred(curr)) {
-      if (currIndex !== list.length - 1) {
-        acc.push([]);
-      }
-    } else {
-      if (acc[acc.length - 1]) {
-        acc[acc.length - 1].push(curr);
-      } else {
-        acc.push([curr]);
-      }
+  var idx = 0;
+  var len = list.length;
+  var ret = [];
+  while (idx < len) {
+    var entry = [];
+    while (idx < len && !pred(list[idx])) {
+      entry = entry.concat(list[idx]);
+      idx += 1;
     }
-    currIndex += 1;
-    return acc;
-  }, [], list);
+    ret = entry.length ? ret.concat([entry]) : ret;
+    idx += 1;
+  }
+  return ret;
 });
 export default splitWith;
--- a/source/splitWith.js
+++ b/source/splitWith.js
@@ -1,0 +1,38 @@
+import _curry2 from './internal/_curry2';
+import _reduce from './internal/_reduce';
+
+
+/**
+ * Takes a list and a predicate and returns a list of lists split by matches
+ *
+ * @func
+ * @memberOf R
+ * @since v0.25.0
+ * @category List
+ * @sig (a -> Boolean) -> [a] -> [[a]]
+ * @param {Function} pred The predicate that determines where the array is split.
+ * @param {Array} list The array to be split.
+ * @return {Array}
+ * @example
+ *
+ *      R.splitWith(R.equals('a'), [1, 2, 3, 1, 2, 3]); //=> [[1], [3, 1], [3]]
+ */
+var splitWith = _curry2(function splitWith(pred, list) {
+  let currIndex = 0;
+  return _reduce(function(acc, curr) {
+    if (pred(curr)) {
+      if (currIndex !== list.length - 1) {
+        acc.push([]);
+      }
+    } else {
+      if (acc[acc.length - 1]) {
+        acc[acc.length - 1].push(curr);
+      } else {
+        acc.push([curr]);
+      }
+    }
+    currIndex += 1;
+    return acc;
+  }, [], list);
+});
+export default splitWith;
--- a/source/splitWith.js
+++ b/source/splitWith.js
@@ -1,5 +1,4 @@
 import _curry2 from './internal/_curry2';
-import _reduce from './internal/_reduce';
 
 
 /**
--- a/source/splitWith.js
+++ b/source/splitWith.js
@@ -15,7 +15,7 @@ import _curry2 from './internal/_curry2';
  * @example
  *
  *      R.splitWith(R.equals(2), [1, 2, 3, 4, 2, 1]); //=> [[1], [3, 4], [1]];
- *      R.splitWith(R.equals('a'), 'bananas'); //=> [['b'],['n'],['n'],['s']]
+ *      R.splitWith(R.equals('a'), ['b','a','n','a','n','a','s']); //=> [['b'],['n'],['n'],['s']]
  */
 
 var splitWith = _curry2(function splitWith(pred, list) {
--- a/test/splitWith.js
+++ b/test/splitWith.js
@@ -2,7 +2,7 @@ var R = require('../source');
 var eq = require('./shared/eq');
 
 
-describe.only('splitWith', function() {
+describe('splitWith', function() {
   it('splits an array at every instance of predicate matching', function() {
     eq(R.splitWith(R.equals(2), [1, 2, 3, 4, 2, 1]), [[1], [3, 4], [1]]);
   });
--- a/test/splitWith.js
+++ b/test/splitWith.js
@@ -2,10 +2,10 @@ var R = require('../source');
 var eq = require('./shared/eq');
 
 
-describe.only('splitWith', function() {
+describe('splitWith', function() {
   it('splits an array at every instance of predicate matching', function() {
     eq(R.splitWith(R.equals(2), [1, 2, 3, 4, 2, 1]), [[1], [3, 4], [1]]);
-    eq(R.splitWith(R.equals('a'), 'bananas'), [['b'],['n'],['n'],['s']]);
+    eq(R.splitWith(R.equals('a'), ['b','a','n','a','n','a','s']), [['b'],['n'],['n'],['s']]);
     eq(R.splitWith(R.gt(5), [1, 7, 5, 4, 3, 7, 6]), [[7, 5], [7, 6]]);
     eq(R.splitWith(R.propEq('x', 5), [{x:1}, {x:5}, {x:7}]), [[{x:1}], [{x:7}]]);
   });
--- a/test/splitWith.js
+++ b/test/splitWith.js
@@ -2,11 +2,14 @@ var R = require('../source');
 var eq = require('./shared/eq');
 
 
-describe('splitWith', function() {
+describe.only('splitWith', function() {
   it('splits an array at every instance of predicate matching', function() {
     eq(R.splitWith(R.equals(2), [1, 2, 3, 4, 2, 1]), [[1], [3, 4], [1]]);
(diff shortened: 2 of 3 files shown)


source/splitWith.js
```diff
@@ -0,0 +1,38 @@
...
+ * @memberOf R
+ * @since v0.25.0
+ * @category List
+ * @sig (a -> Boolean) -> [a] -> [[a]]
+ * @param {Function} pred The predicate that determines where the array is split.
+ * @param {Array} list The array to be split.
+ * @return {Array}
+ * @example
+ *
+ *      R.splitWith(R.equals('a'), [1, 2, 3, 1, 2, 3]); //=> [[1], [3, 1], [3]]
```
CrossEye: Presumably `R.equals(2)` not `R.equals('a')`.

source/splitWith.js
```diff
@@ -0,0 +1,38 @@
...
+ */
+var splitWith = _curry2(function splitWith(pred, list) {
+  let currIndex = 0;
+  return _reduce(function(acc, curr) {
+    if (pred(curr)) {
+      if (currIndex !== list.length - 1) {
+        acc.push([]);
+      }
```
CrossEye: This only works if there is a single matching value.  With this implementation,

```js
splitWith(equals(2), [2, 2, 1, 2, 3, 4, 1, 2, 2]) //=> [[], [1], [3, 4, 1], []]
```

So something needs to change.

test/splitWith.js
```diff
@@ -0,0 +1,21 @@
...
+  it('returns sane default if an empty list is given', function() {
+    eq(R.splitWith(R.equals(1), []), []);
+  });
+  it('handles first and last elements matching predicate', function() {
+    eq(R.splitWith(R.equals(1), [1, 2, 3, 4, 1]), [[2, 3, 4]]);
+  });
+  it('handles strings', function() {
+    eq(R.splitWith(R.equals('a'), 'bananas'), [['b'],['n'],['n'],['s']]);
+  });
+});
```
CrossEye: I would want a few additional tests.  

  - Two for the issue mentioned above with two or more predicate-matching values at the ends
  - One for a non-`equals` predicate.  (`splitWith(n => n > 5, [2, 4, 6, 1, 3, 7, 8, 0])`, for instance.
  - And possibly one for an `equals` with something other than numbers or strings, perhaps `{x: 42}` or some such.

On the other hand, the `handles strings` test is not important in itself.  I would just do one or two with numbers, and some others with Strings.

test/splitWith.js
```diff
@@ -0,0 +1,21 @@
...
+var eq = require('./shared/eq');
+
+
+describe('splitWith', function() {
+  it('splits an array at every instance of predicate matching', function() {
+    eq(R.splitWith(R.equals(2), [1, 2, 3, 4, 2, 1]), [[1], [3, 4], [1]]);
+  });
+  it('retains all original elements', function() {
+    eq(R.splitWith(R.equals(2), [1, 1, 1]), [[1, 1, 1]]);
+  });
```
CrossEye: `'splitWith' retains all original elements` doesn't tell me anything.  You need another clause.  Perhaps, `when the predicate matches nothing.`

xgbuils:
Hi!

If I'm not wrong, `splitWith` should behave analogously to `String#split`. Then:

``` javascript
''.split(anyString) // ['']
splitWith(anyPredicate, []) // [[]]

'222'.split('2') // ['', '', '', '']
splitWith(() => true, [1, 2, 3]) // [[], [], [], []]

'222'.split('1') // ['222']
splitWith(() => false, [1, 2, 3]) // [[1, 2, 3]]

'112233'.split('2') // ['11', '', '33']
splitWith(R.equals(2), [1, 1, 2, 2, 3, 3]) // [[1, 1], [], [3, 3]]

'112233'.split('3') // ['1122', '', '']
splitWith(R.equals(3), [1, 1, 2, 2, 3, 3]) // [[1, 1, 2, 2], [], []]
```

I think that the code could be simplified. Here an implementation in es6:
``` javascript
var push = (value, arr) => {
    arr.push(value);
    return arr;
};
var step = pred => (acc, value) => pred(value)
    ? [push(acc[1], acc[0]), []]
    : [acc[0], push(value, acc[1])];

var result = (acc) => push(acc[1], acc[0]);

var splitWith = (pred, arr) => result(arr.reduce(step(pred), [[], []]));
```

Cheers!

test/splitWith.js
```diff
@@ -0,0 +1,23 @@
+var R = require('../source');
+var eq = require('./shared/eq');
+
+
+describe.only('splitWith', function() {
+  it('splits an array at every instance of predicate matching', function() {
+    eq(R.splitWith(R.equals(2), [1, 2, 3, 4, 2, 1]), [[1], [3, 4], [1]]);
+    eq(R.splitWith(R.equals('a'), 'bananas'), [['b'],['n'],['n'],['s']]);
```
CrossEye: If this is the behavior, I'm not sure I want to promote it in a test case.  To my mind, if we accept Strings intentionally rather than as an accident of implementation, then this should really return `['b', 'n', 'n', 's']`  -- although I like the example and the cross-function consistency in this example.

jamesmarlowe: That makes sense, I changed it to an array of strings (which has the same output which highlights the issue)

CrossEye:
@xgbuils:

> If I'm not wrong, `splitWith` should behave analogously to `String#split`

Why?  Those features where `String#split` differs from this are precisely those features I particularly dislike in `String#split`.  Is there a strong reason you think we should follow that API?

CrossEye:
This looks good to me.  Any other thoughts, @ramda/core?

xgbuils:
Hi @CrossEye 

> Why? Those features where String#split differs from this are precisely those features I particularly dislike in String#split.

Hi! I was trying to understand the behaviour of `R.splitWith` because the point in #2620 is not clear to me. 

> 4. Exclude the split values from the results

For example, I interpreted that behaves like `R.split` or `String#split`.

> Is there a strong reason you think we should follow that API?

I don't know if they are strong but some thoughts here:

1. Consistency. `R.splitWith` behaves different from `R.split`. Then, it could be less intuitive for the `ramda` users.

2. Single responsability. If we implement `R.splitWith` with `[]` holes and some user want the version without holes it could compose `R.filter` and `R.splitWith`: 
``` javascript
var splitWithNoHoles = (pred) => R.compose(
    R.filter(arr => !R.equals([], arr)),
    R.splitWith(pred)
);
``` 
We can't do the same in the reverse way.

3. I always liked the solution of counting specific characters of string using `String#split`:
``` javascript
const count = (c, str) => str.split(c).length - 1
```
We can get analogous `countByPredicate` using `splitWith`. However I know that this point is very weak.

Cheers!

CrossEye:
@xgbuils:

Of these arguments, I think the second one is the strongest.

Consistency with one other `split*` function is not particularly convincing when we also have `splitWhen`, `splitEvery`, and `splitAt`, which don't really have anything to do with strings.  I think of this function as more related to these ones and to ones like `aperture`, `partition`, `intersperse`, and the like.  One of the other nice things about `split` is its inverse in `join`:  `compose(join(s), split(s))` is equivalent to `identity` on its domain, for any `s`, as is the reverse.  But we fundamentally cannot do something like that using `splitWith`.

And as you note, the third argument doesn't add much, just noting a particular short implementation of a useful function.  In Ramda, the predicate version would be shorter still: `compose(length, filter)`.

But the second point is more compelling.  I don't think single-responsibility is in itself a strong argument in a library that includes `pluck` as well as `map` and `prop`, one that would rather take a pragmatic approach to including derivable functions if they're useful.  But the fact that this is one-way gives me a bit of pause.  That the un-hole-y (:smile:)  version can be derived from the hole-y one but not the reverse means that we should think about this a bit more.

Still, I have never wanted a hole-y result like that.  Do you have practical use-cases where you have wanted such results?  (Besides counting characters in a string!)  I personally have often used an equivalent function to generate the un-hole-y version.  But never the hole-y one.  I'd like to know when it's actually been useful.

xgbuils:
@CrossEye 

Currently I can't find any example for needing hole-y version. Then, un-hole-y version is :+1:  for me.

paldepind:
```js
R.splitWith(R.equals('a'), ['b','a','n','a','n','a','s']); //=> [['b'],['n'],['n'],['s']]
```
Maybe it's just me, but I find that behavior confusing. I would expect the following:
```js
R.splitWith(R.equals('a'), ['b','a','n','a','n','a','s']); //=> [['b'], ['a', 'n'], ['a', 'n'], ['a', 's']]
```
For all the existing `R.split*` functions it is the case that they don't remove elements. I.e. for each splitting function `R.split*` it is the case that `R.join(R.split*(...))` gets you back to where you started. To me the name "split" indicates a function that seperates a sequence into several sequences but such that the sequences together contain the original elements.

In fact, the least confusing behavior, to me, whould be:
```js
R.splitWith(R.equals('a'), ['b','a','n','a','n','a','s']); //=> [['b'], ['a', 'n', 'a', 'n' , 's']]
```
Becaues that seems to be most consisting with the existing split functions. The only current split function that does not split the array into exactly two arrays is `R.splitEvery` but for that function the "every" clearly communicates that there are many splits. With the current behavior of the function in this PR I whould expect and prefer it to be called `R.splitEveryWith`.

CrossEye:
@paldepind:

> For all the existing `R.split*` functions it is the case that they don't remove elements. I.e. for each splitting function `R.split*` it is the case that `R.join(R.split*(...))` gets you back to where you started. 

Well, this is not precisely the case for `R.split`.  You need to pass the same split value to both `split` and `join`.  Perhaps we need to bikeshed a bit on the name.  `splitWhenever` appeals to me much more than `splitEveryWith`.  I wish we had separated with names the two concepts of breaking apart linear structures based on when a new group *starts* and when a value *punctuates* the subgroups.  We could, for instance, have used `spit*` and `break*` to distinguish them.  And of course we still could, but that's another breaking change.

foxbunny:
> Yes, you're right. I forgot about R.split. That definitely weakens my argument, but it's still true that 3/4 of the existing R.split* functions doesn't throw away elements.

That's because the other `R.split*` functions are not doing the same thing to begin with. `R.split()` takes a separator and it breaks the iterable down along those. Others break the iterables _without_ a separator. The sound of `splitWhere()` appears to be closer to the other functions than `split()` proper. In fact, it sounds like a more version of `splitWhen()` that goes all the way.

Also some more food for thought:

Python has two standard methods on strings, `split()` and `partition()`. The difference is as follows:

```
>>> 'bananas'.split('a')
['b', 'n', 'n', 's']
>>> 'bananas'.partition('a')
('b', 'a', 'nanas')
```

The `partition()` methods behaves similar to `splitWhen()` but does include the separator. Since it always returns a 3-tuple, the inclusion of a separator is OK, because it's easy to work with. If we are talking about functions that are easy to build on, I would say it should be modeled after this kind of behavior. Do just ONE split, and return the separator. If you need to go all the way through the rest, just do it in an iteration, and you have full control over what is done with the separator.

paldepind:
@foxbunny

> That's because the other R.split* functions are not doing the same thing to begin with. R.split() takes a separator and it breaks the iterable down along those. Others break the iterables without a separator. The sound of splitWhere() appears to be closer to the other functions than split() proper.

Yes. As far as I can tell we're pretty much saying the same thing. Considering what `split` does it makes sense to remove elements. But due to the nature of the other split functions they don't throw away elements. And this new function is a lot closer to the other split functions and hence it shouldn't remove elements either.


@CrossEye

> Perhaps we need to bikeshed a bit on the name. splitWhenever appeals to me much more than splitEveryWith.

I agree. That name is great :heart:. The "when" part clearly communicates that the argument is a predicate and the "ever" part indicates that the splitting is done many times.

> I wish we had separated with names the two concepts of breaking apart linear structures based on when a new group starts and when a value punctuates the subgroups. We could, for instance, have used spit* and break* to distinguish them.

What about `partition`? Would it make sense to consider this and other split functions a variant of `partition`?

CrossEye:
@jamesmarlowe:

Could you change the name to `splitWhenever`?  I think this conversation has played out, and I'd like to see this change make it in.

CrossEye:
Done in #2884

Issue #2692, opened 2018-11-08
[Feature] Move item in a list


MadDeveloper:
Unless I'm mistaken, or have not found the function in the documentation yet, is it possible to add a function which permits to move and item in a list?
The simplest solution I found for now:
```javascript
const move = curry((at, to, list) => insert(to, prop(at, list), remove(at, 1, list)))
```

I'm new to Ramda, so my function could be simplified, but even if it's possible I think that a `move` function could be nice 😉

**EDIT**
An issue had been opened (https://github.com/ramda/ramda/issues/2218) but I don't see a solution. It's not exactly the same needs, but it's similar.

CrossEye:
This is more general than what was suggested in #2218.  You could easily build `swap` atop such a `move`.  This is barely tested, and might fail on some boundary conditions, but I think it's close:

```js
const move = curry((at, to, list) => insert(to, nth(at, list), remove(at, 1, list)))

const swap = curry((idx1, idx2, list) => move(
  idx2 + (idx1 > idx2 ? 1 : -1), 
  idx1, 
  move(idx1, idx2, list))
)
```

I'm still less interested in `swap`, but there have been more requests for it, and `move` seems even more useful.  So, while there's no guarantee it would be accepted, I for one would love to see a PR for this.

MadDeveloper:
Indeed the `move` permits to built from it some other functions, for examples: (I took back your swap function):
```javascript
const swap = curry((a, b, list) => move(b + (a > b ? 1 : -1), a, move(a, b, list)))
const rotateN = curry((n, list) => n === 0 ? list : rotateN(n - 1, move(-1, 0, list)))
const rotate = list => rotateN(1, list)

console.log(rotate(list)) // [4,1,2,3]
console.log(rotateN(2, list)) // [3,4,1,2]
console.log(swap(2, 3, list)) // [1,2,4,3]
```
If I have to write other functions (like swap, rotate, rotateN (or only fews of them)), do I need to make one PR for each, or I can bulk all in one?

semmel:
`move`, `swap`, `rotate*`, ... would all be nice to have. :+1: 

This is how I do 1x rotate though:

```javascript
rotateRight = R.converge(R.prepend, [R.last, R.init]),
rotateLeft = R.converge(R.append, [R.head, R.tail])
```

MadDeveloper:
@semmel thank's for `rotate*`, I noticed I'd forgotten the side of the rotation! I've updated my previous comment.

[...]

PR #2694, opened 2018-11-09
feat(List): add move function

MadDeveloper:
As discussed here: https://github.com/ramda/ramda/issues/2692

### DIFF
--- a/source/move.js
+++ b/source/move.js
@@ -20,11 +20,12 @@ import _curry3 from './internal/_curry3';
 var move = _curry3(function(from, to, list) {
   var result = list.slice();
   var item = result.splice(from, 1);
+  var positiveTo = to < 0 ? list.length + to : to;
 
   return []
-    .concat(result.slice(0, to))
+    .concat(result.slice(0, positiveTo))
     .concat(item)
-    .concat(result.slice(to, list.length));
+    .concat(result.slice(positiveTo, list.length));
 });
 
 export default move;
--- a/source/move.js
+++ b/source/move.js
@@ -1,0 +1,30 @@
+import _curry3 from './internal/_curry3';
+
+/**
+ * Move an item, at index `from`, to index `to`, in a list of elements.
+ * A new list will be created containing the new elements order.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig Number -> Number -> [a] -> [a]
+ * @param {Number} from The source index
+ * @param {Number} to The destination index
+ * @param {Array} list The list which will serve to realise the move
+ * @return {Array} The new list reordered
+ * @example
+ *
+ *      R.move(0, 2, [1,2,3]); //=> [2,3,1]
+ *      R.move(-1, 0, [1,2,3]); //=> [3,1,2] list rotation
+ */
+var move = _curry3(function(from, to, list) {
+  var result = list.slice();
+  var item = result.splice(from, 1);
+
+  return []
+    .concat(result.slice(0, to))
+    .concat(item)
+    .concat(result.slice(to, list.length));
+});
+
+export default move;
--- a/source/move.js
+++ b/source/move.js
@@ -14,8 +14,8 @@ import _curry3 from './internal/_curry3';
  * @return {Array} The new list reordered
  * @example
  *
- *      R.move(0, 2, [1,2,3]); //=> [2,3,1]
- *      R.move(-1, 0, [1,2,3]); //=> [3,1,2] list rotation
+ *      R.move(0, 2, [8, 6, 7, 5, 3, 0, 9]); //=> [ 6, 7, 8, 5, 3, 0, 9 ]
+ *      R.move(-1, 0, [8, 6, 7, 5, 3, 0, 9]); //=> [ 6, 7, 5, 3, 0, 9, 8 ] list rotation
  */
 var move = _curry3(function(from, to, list) {
   var result = list.slice();
--- a/source/move.js
+++ b/source/move.js
@@ -1,4 +1,6 @@
 import _curry3 from './internal/_curry3';
+import max from './max';
+import min from './min';
 
 /**
  * Move an item, at index `from`, to index `to`, in a list of elements.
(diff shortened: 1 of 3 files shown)


source/move.js
```diff
@@ -0,0 +1,30 @@
...
+ */
+var move = _curry3(function(from, to, list) {
+  var result = list.slice();
+  var item = result.splice(from, 1);
+
+  return []
+    .concat(result.slice(0, to))
+    .concat(item)
+    .concat(result.slice(to, list.length));
+});
```
CrossEye: Although I'm the one who suggested that this be done on the native methods, I'm having some second thoughts.  Would this do well working on all indexable types?  Strings would be an obvious fit, but should it also work on `NodeList`, on `ArgumentList`, and so on? 

What do you think, @ramda/core?

source/move.js
```diff
@@ -0,0 +1,30 @@
...
+ * @category List
+ * @sig Number -> Number -> [a] -> [a]
+ * @param {Number} from The source index
+ * @param {Number} to The destination index
+ * @param {Array} list The list which will serve to realise the move
+ * @return {Array} The new list reordered
+ * @example
+ *
+ *      R.move(0, 2, [1,2,3]); //=> [2,3,1]
+ *      R.move(-1, 0, [1,2,3]); //=> [3,1,2] list rotation
```
CrossEye: I think `[1, 2, 3]` could too easily cause confusion, especially given how close it is to indices.  I also think it's too short by at least one.  Either `['a', 'b', 'c', 'd']` or `[8, 6, 7, 5, 3, 0, 9]` would demonstrate things better.

test/move.js
```diff
@@ -0,0 +1,19 @@
...
+var eq = require('./shared/eq');
+
+describe.only('move', function() {
+  it('should correctly moves an element from an index to another', function() {
+    eq(R.move(0, 1, [1,2,3]), [2,1,3]);
+    eq(R.move(2, 1, [1,2,3]), [1,3,2]);
+    eq(R.move(-1, 0, [1,2,3]), [3,1,2]);
+  });
+
+  it('should move the element at the end when the destination index is outside liste bounds', function() {
```
CrossEye: typo: "liste" should be "list"

CrossEye: But is this the correct behavior?  Why do we distinguish how it behaves with the source index and the destination one?

test/move.js
```diff
@@ -0,0 +1,19 @@
+var R = require('../source');
+var eq = require('./shared/eq');
+
+describe.only('move', function() {
+  it('should correctly moves an element from an index to another', function() {
```
CrossEye: Ramda usually skips "should" in the test description, and I don't see a good reason for "correctly" either.

```js
it('moves an element from one index to another', function() { /* ... */ })
```

seems cleaner to me.

test/move.js
```diff
@@ -0,0 +1,19 @@
...
+    eq(R.move(0, 1, [1,2,3]), [2,1,3]);
+    eq(R.move(2, 1, [1,2,3]), [1,3,2]);
+    eq(R.move(-1, 0, [1,2,3]), [3,1,2]);
+  });
+
+  it('should move the element at the end when the destination index is outside liste bounds', function() {
+    eq(R.move(0, 20, [1,2,3]), [2,3,1]);
+  });
+
+  it('should do nothing when source index is outside liste bounds', function() {
```
CrossEye: see above

test/move.js
MadDeveloper: Because of the flexibility of `move()` and more precisely `splice` and `slice`, we are able to make a list rotation, but it has some unexpected behaviour, like this one you pointed out. How do you think prevent this behaviour without breaking the rotation one?

test/move.js
```diff
@@ -0,0 +1,22 @@
+var R = require('../source');
+var eq = require('./shared/eq');
+
+var list = [8, 6, 7, 5, 3, 0, 9];
+
+describe.only('move', function() {
```
mattgrande: Just a note to remove the `.only` before this gets merged :)

MadDeveloper: Sorry, an oversight 😕

test/move.js
CrossEye: I'm not quite sure.  Perhaps we want to allow negative outbounds only down to the first element.  I really don't like this asymmetry:

```js
move(-50, 2, ['a', 'b', 'c', 'd', 'e', 'f']) //=> ['b', 'c', 'a', 'd', 'e', 'f']
move( 50, 2, ['a', 'b', 'c', 'd', 'e', 'f']) //=> ['a', 'b', 'c', 'd', 'e', 'f']
```

MadDeveloper: Maybe these ones are enough:
```js
move(0, -1, ['a', 'b', 'c']) /// ['b', 'c', 'a']
move(-1, 0, ['a', 'b', 'c']) /// ['c', 'a', 'b']
```

MadDeveloper: So? Should Ramda be more strict on indexes, or keep that asymmetric behaviour?

CrossEye: I have long wished that we never started supporting negative indices.  They add complexity that is rarely warranted, even though they do slightly simplify the use in a few cases.  But given that we have them, perhaps we should use something conceptually like this for any index we work with:

```js
const realIndex = (index, {length}) => index < 0 
  ? max(length + index, 0)
  : min(length - 1, index) 
```

I think that's already how `slice` works -- just by defaulting to `Array.prototype` or `String.prototype`.  Could we do that here with both `from` and `to`?

That would yield:

```js
move( -2,  1, ['a', 'b', 'c', 'd', 'e', 'f']) //=> ['a', 'e', 'b', 'c', 'd', 'f']
move(  1, -2, ['a', 'b', 'c', 'd', 'e', 'f']) //=> ['a', 'c', 'd', 'e', 'b', 'f']
move(-50,  2, ['a', 'b', 'c', 'd', 'e', 'f']) //=> ['b', 'c', 'a', 'd', 'e', 'f']
move( 50,  2, ['a', 'b', 'c', 'd', 'e', 'f']) //=> ['a', 'b', 'f', 'c', 'd', 'e']
```
----------

Better to my mind, though, is never falling off the end like that.  If either of the indices (allowing for small negative values) is not present, then we return an unmodified copy of the list.

That might involve creating a `positiveFrom` like the `positiveTo` and, before the `splice`, adding

```js
if (positiveFrom < 0 || positiveFrom >= list.length ||
    positiveTo   < 0 || positiveTo   >= list.length) {
  return result;
}
```

yielding

```js
move( -2,  1, ['a', 'b', 'c', 'd', 'e', 'f']) //=> ['a', 'e', 'b', 'c', 'd', 'f']
move(  1, -2, ['a', 'b', 'c', 'd', 'e', 'f']) //=> ['a', 'c', 'd', 'e', 'b', 'f']
move(-50,  2, ['a', 'b', 'c', 'd', 'e', 'f']) //=> ['a', 'b', 'd', 'd', 'e', 'f']
move( 50,  2, ['a', 'b', 'c', 'd', 'e', 'f']) //=> ['a', 'b', 'c', 'd', 'e', 'f']
```
Does that seem correct?

CrossEye: Do you want to do that update?  I'd be inclined to accept this with that change, although I would still like some feedback from @ramda/core .

MadDeveloper: Updated. I didn't add the `realIndex` function in the `internal` directory, as I don't know if it should be added in a separate PR. Let me know if you want to see it appears in this PR.

CrossEye: Well, I'm afraid that this code combines two of what I think of contradictory suggestions of mine.

I was suggesting something like this except with

```js
  var positiveFrom = from < 0 ? list.length + from : from;
  var positiveTo = to < 0 ? list.length + to : to;
```

so that  

```js
move( -2,  1, ['a', 'b', 'c', 'd', 'e', 'f']) //=> ['a', 'e', 'b', 'c', 'd', 'f']
move(  1, -2, ['a', 'b', 'c', 'd', 'e', 'f']) //=> ['a', 'c', 'd', 'e', 'b', 'f']
move(-50,  2, ['a', 'b', 'c', 'd', 'e', 'f']) //=> ['a', 'b', 'd', 'd', 'e', 'f']
move( 50,  2, ['a', 'b', 'c', 'd', 'e', 'f']) //=> ['a', 'b', 'c', 'd', 'e', 'f']
```

The idea is that a negative index only does something up to the length of the list.  After that, it becomes a no-op.  I think that stopping at the end of the array is far too surprising.

MadDeveloper: Ok, I think there was a misunderstanding... I fixed it and I hope everything is ok now.

CrossEye:
Merged.  Thank you for your persistence.

PR #2696, opened 2018-11-10
Add rus translate

Guck111:


.gitignore
```diff
@@ -10,3 +10,4 @@
 /dist/gh-pages/
 /tmp-test-bundle.js
 /coverage/
+.idea/*
```
CrossEye: I'd rather not start committing this for everyone's choice of editor.

Could you remove this entry?

PR #2700, opened 2018-11-17
Lint cleanup

CrossEye:
I was updating some dependencies, and ESLint seems to have gotten pickier.


davidchambers:
@CrossEye, did you mean to add __package-lock.json__? I suggest we remove it (see #3214).

CrossEye:
No, I didn't.  I understand why `package-lock.json` was added for npm; however, it's been nothing but a thorn in my side since it was introduced.

PR #2701, opened 2018-11-18
Deprecating {compose|pipe}{P|K}

CrossEye:


antonazgarovich:
Why is deprecated?

CrossEye:
@antonazgarovich:

They're being replaced with `pipeWith/composeWith`, a more generic version that can handle these and other possible piping operations.  See #2515 and related discussions.

PR #2703, opened 2018-11-18
Disallow tests to depend on build

GingerPlusPlus:


CrossEye:
Thanks.  I always miss something on a release!

Any objections, @ramda/core?

PR #2704, opened 2018-11-19
update propOr documentation following #2394

johannes-scharlach:
In MR #2394 the behaviour of `propOr` was changed to not only consider own properties
and to work with `null` and `undefined`.

I've also expanded the examples. Let me know if this is too verbose or you'd like something changed.

### DIFF
--- a/source/propOr.js
+++ b/source/propOr.js
@@ -3,9 +3,8 @@ import pathOr from './pathOr';
 
 
 /**
- * If the given, non-null object has an own property with the specified name,
- * returns the value of that property. Otherwise returns the provided default
- * value.
+ * If the given object has an property with the specified name, returns the value of that
+ * property. Otherwise returns the provided default value.
  *
  * @func
  * @memberOf R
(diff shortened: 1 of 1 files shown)


source/propOr.js
```diff
@@ -22,11 +21,19 @@ import pathOr from './pathOr';
...
+ *        name: 'BOB',
+ *        age: 61,
+ *        favoriteLibrary: 'Ramda@0.26.0'
+ *      };
  *      const favorite = R.prop('favoriteLibrary');
  *      const favoriteWithDefault = R.propOr('Ramda', 'favoriteLibrary');
  *
  *      favorite(alice);  //=> undefined
  *      favoriteWithDefault(alice);  //=> 'Ramda'
+ *      favoriteWithDefault(Object.create(alice));  //=> 'Ramda'
```
CrossEye: What is this line trying to prove?  It seems to say that if Alice doesn't have the property neither will a clone of Alice; that shouldn't be a surprise.

While I can at least conceive of a reason to do this for Bob, where the value will be found on the prototype, I can see no reason to do it for Alice.

CrossEye: I think this one line should simply be removed.

source/propOr.js
```diff
@@ -3,9 +3,8 @@ import pathOr from './pathOr';
 
 
 /**
- * If the given, non-null object has an own property with the specified name,
- * returns the value of that property. Otherwise returns the provided default
- * value.
+ * If the given object has an property with the specified name, returns the value of that
+ * property. Otherwise returns the provided default value.
```
CrossEye: Definitely.  Thank you.

source/propOr.js
```diff
@@ -3,9 +3,8 @@ import pathOr from './pathOr';
 
 
 /**
- * If the given, non-null object has an own property with the specified name,
- * returns the value of that property. Otherwise returns the provided default
- * value.
+ * If the given object has an property with the specified name, returns the value of that
```
kurtmilam:
```suggestion
 * If the given object has a property with the specified name, returns the value of that
```

Replace 'an' with 'a'.

source/propOr.js
johannes-scharlach: You're right, this is nonsense in this form.

What I'd like to document is that it _does_ check the prototype for the property. This is where the behaviour has changed compared to version 0.25, which only considered own properties. Would it be more useful if I write the following (as the last example):

```js
favoriteWithDefault(Object.create(bob));  //=> 'Ramda@0.26.0'
```

CrossEye: While I think that would be better, I'm not convinced that it's necessary.  Our examples are not meant to be exhaustive, nor to show version changes.  However, I wouldn't object if you feel strongly about it.

CrossEye:
This has since been patched.

PR #2705, opened 2018-11-19
Remove placeholder

wojpawlik:
Closes #2629, I think I tried to do too much in single PR.


source/internal/_curry3.js
```diff
@@ -12,38 +10,5 @@ import _isPlaceholder from './_isPlaceholder';
...
-              : _isPlaceholder(b) && _isPlaceholder(c)
-                ? _curry2(function(_b, _c) { return fn(a, _b, _c); })
-                : _isPlaceholder(a)
-                  ? _curry1(function(_a) { return fn(_a, b, c); })
-                  : _isPlaceholder(b)
-                    ? _curry1(function(_b) { return fn(a, _b, c); })
-                    : _isPlaceholder(c)
-                      ? _curry1(function(_c) { return fn(a, b, _c); })
-                      : fn(a, b, c);
-    }
```
CrossEye: I especially like the idea of seeing this go away!

CrossEye:
I'd like to hear back from @ramda/core, especially @Bradcomp and @davidchambers, who have expressed fondness for the placeholder.

I think this might be a better fit for a 2.0 branch, now that we might actually be on the way to cutting a 1.0 one.  (And I'm not looking forward to maintaining multiple branches, I'll tell you!)

davidchambers:
Let's remove `R.__`. We dropped `S.__` in sanctuary-js/sanctuary#499 and haven't looked back. :)

Bradcomp:
Here's my weak argument for keeping it. In any instance where one would use `flip` I think the placeholder is a more readable solution. If we want a practical library I think it makes more sense to remove `flip` as it's more useful for point free tricks than for writing understandable code. The placeholder is a more readable and versatile solution of the same problem.

That being said, it isn't too often that I need to use the placeholder, and a lambda expression will do the same thing, so I am not too sold on keeping it given the implementation complications it causes. If we eventually move to 'simple' currying then the problem becomes a moot point anyways.

CrossEye:
@Bradcomp: Yes, especially when used without the `R.` prefix, it is more readable than `flip`, but it is little or no more more readable than the comparable lambda.  My take is that it is not worth the overhead.

Anyone have thoughts on trying to get this in right away or moving it to a 2.x branch?  My first thought is that I just want to get 1.0 out with most of the same features we have now and any minor outstanding additions, deletions, and bug-fixes.  This should then go in the 2.0 branch.  Other opinions?  @buzzdecafe?

buzzdecafe:
> Anyone have thoughts on trying to get this in right away or moving it to a 2.x branch? My first thought is that I just want to get 1.0 out with most of the same features we have now and any minor outstanding additions, deletions, and bug-fixes. This should then go in the 2.0 branch. Other opinions? @buzzdecafe?

Easiest & least disruptive IMO is to ship 1.0 as is, and remove `__` ASAP in the next release branch. I'm very happy to see it on the chopping block. Was an interesting experiment that was worth a try, but IMO the lib is better off without it.

GingerPlusPlus:
> In any instance where one would use `flip` I think the placeholder is a more readable solution. If we want a practical library I think it makes more sense to remove `flip`

@Bradcomp: glad you brought up `flip` :) I think I have a better idea than removing it: #2627.
After that change: `const isAdult = R.flip(R.gte, 18)`.


> Anyone have thoughts on trying to get this in right away or moving it to a 2.x branch?

> This should then go in the 2.0 branch. Other opinions?

@CrossEye: if we provide an alternative other than function expression (such as modified `flip`, #2627), I see no reason to wait.

CrossEye:
@GingerPlusPlus:
I don't remember even seeing #2627.  I will look at it more carefully soon, but I don't see how `flip(gte, 18)` is much of an advantage over `flip(gte)(18)`.  I also don't remember discussion about removing `flip`.  I would not be happy about that, unless we replaced it with something more powerful.

>  if we provide an alternative other than function expression (such as modified flip, #2627), I see no reason to wait.

I'm hesitant to make any significant changes that we might want to back out as we try to hit 1.0.  That's what worries me.  Ramda, our little science fair experiment, has become a widely used tool; it's harder to make changes.  Post-1.0 should be a bit liberating.

GingerPlusPlus:
Note that re-adding the placeholder support would be a non-breaking change, wouldn't have to wait for a major release to do it.

And current implementation is messy.

CrossEye:
My concern is in merging this pre-1.0, publishing 1.0, then finding it was a bad idea.  Once we get 1.0 out, I'm hoping the only real changes on that branch will be minor bug fixes.  If we move this to after the 1.0 release, then we will, I'm all too afraid, be releasing a number of major versions.  We don't keep very still here, and I will want to take semver more seriously at that point.

GingerPlusPlus:
> Once we get 1.0 out, I'm hoping the only real changes on that branch will be minor bug fixes.

Ah, you want 1.x to be more of a not-breaking-much continuation of 0.x, just under semver? Fair enough, provided that you'll start working on (breaking) 2.x immediately after releasing 1.x.

CrossEye:
That's my goal.  What I'm really interested in is the larger breaking changes that I would want to consider for 2.0+.  As soon as we can get a reasonably stable 1.0 out, I'd like to start looking into that.

CrossEye:
@tommmyy:

You wouldn't, although I would actually think about including something like lodash's [`rearg`][ra] at some point.

But point-free should not be a goal on its own.  Do what's most readable.  To me, this is already more readable:

```js
const ifElseF = (a, b) => ifElse(a, b, F)
```


  [ra]: https://lodash.com/docs/4.17.11#rearg

tommmyy:
@CrossEye 

but to complete the impolementation you one must use `curry`:

```js
const ifElseF = curry((a, b) => ifElse(a, b, F))
```

and then version below can be useful refactor:

```js
const ifElseF = ifElse(__, __, F)
```

CrossEye:
@tommmyy:

Absolutely.  There's no perfect replacement.  But the placeholders slow down the code, offer only a slight benefit in the ES6+ world, and lead to implementation issues such as [`curry3`][c3].

That's why, even though I still use them occasionally, I would be very happy to see them go.


  [c3]: https://github.com/ramda/ramda/blob/b7a4f09f66ed04c7c18d3e86015ad780c6a529ef/source/internal/_curry3.js#L14-L49

CrossEye:
Revisit after `v1.0`

Issue #2706, opened 2018-11-20
Combine `groupBy` and `values`?


customcommander:
Following up on this [post in Stack Overflow](https://stackoverflow.com/questions/53372012/javascript-remove-duplicate-in-a-list-of-objects-merging-some-properties) and @CrossEye suggestion:

>  I actually think Ramda needs a function something like (fn) => (list) => values(groupBy(fn, list)). It's such a common combination.

The idea (if I got it correctly) is to facilitate this transformation:

From:

```javascript
[
  {id: 'foo', amount: 10},
  {id: 'foo', amount: 20},
  {id: 'foo', amount: 30},
  {id: 'bar', amount: 10},
  {id: 'bar', amount: 10},
  {id: 'baz', amount: 10}
]
```

To:

```javascript
[
  [ {id: 'foo', amount: 10},
    {id: 'foo', amount: 20},
    {id: 'foo', amount: 30} ],

  [ {id: 'bar', amount: 10},
    {id: 'bar', amount: 10} ],

  [ {id: 'baz', amount: 10} ]
]
```

One way of doing this could be, as @CrossEye suggested:

```javascript
const groupValuesBy = curry((groupFn, list) => values(groupBy(groupFn, list)));
```

Would it make sense to offer this "shortcut" as part of Ramda's toolkit?

CrossEye:
Yes, I think it's worth talking about. 

The example might be misleading, though.  I think we should get the same results from this:

```js
[
  {id: 'foo', amount: 10},
  {id: 'bar', amount: 10},
  {id: 'foo', amount: 20},
  {id: 'foo', amount: 30},
  {id: 'baz', amount: 10},
  {id: 'bar', amount: 10}
]
```

There would be only one `foo` group, one `bar` group, etc., even if the elements are not grouped together in the initial list.  This is as opposed to [`groupWith`][gw], which groups only sequential elements.

I don't particularly like the name `groupValuesBy`; it seems more to do with this quick implementation than with the underlying goal.  But I don't have a ready alternative.  I'll think about that.


  [gw]: https://ramdajs.com/docs/#groupWith

customcommander:
Oh I see what you mean about groupWith, good catch.  Yeah the name isn't great I agree.

Here's a couple of other suggestions:

- nestWith
- partitionGroup
- arrangeBy

def14nt:
arrange is cool. Goes with ramda's nice function names

CrossEye:
`collectBy` sounds good to me.

[...]

PR #2715, opened 2018-11-25
Implement R.collectBy

customcommander:
Follow up to #2706 

I wanted to keep the implementation to a bare (perhaps naive) minimum and get some early feedback. I'm not too sure about a few things:

1. Compatibility with transducers.
2. Compatibility with any relevant fantasy land specs.
3. I'm not too familiar with the Hindley-Milner notation.
4. Is it something that the collective would consider a nice addition to the library? :) 

Thank you

### DIFF
--- a/source/collectBy.js
+++ b/source/collectBy.js
@@ -1,0 +1,44 @@
+import _curry2 from './internal/_curry2';
+import _reduce from './internal/_reduce';
+
+/**
+ * Splits a list into sub-lists, based on the result of calling a String-returning function on each element,
+ * and grouping the results according to values returned.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig (a -> String) -> [a] -> [[a]]
+ * @param {Function} fn Function :: a -> String
+ * @param {Array} list The array to group
+ * @return {Array}
+ *    An array of arrays where each sub-array contains items for which
+ *    the String-returning function has returned the same value.
+ * @see R.groupBy, R.partition
+ * @example
+ *      R.collectBy(R.prop('type'), [
+ *        {type: 'breakfast', item: '☕️'},
+ *        {type: 'lunch', item: '🌯'},
+ *        {type: 'dinner', item: '🍝'},
+ *        {type: 'breakfast', item: '🥐'},
+ *        {type: 'lunch', item: '🍕'}
+ *      ]);
+ *
+ *      // [ [ {type: 'breakfast', item: '☕️'},
+ *      //     {type: 'breakfast', item: '🥐'} ],
+ *      //   [ {type: 'lunch', item: '🌯'},
+ *      //     {type: 'lunch', item: '🍕'} ],
+ *      //   [ {type: 'dinner', item: '🍝'} ] ]
+ */
+var collectBy = _curry2(function collectBy(fn, list) {
+  var group = _reduce(function(o, x) {
+    var tag = fn(x);
+    if (o[tag] === undefined) { o[tag] = []; }
+    o[tag].push(x);
+    return o;
+  }, {}, list);
+  var newList = [];
+  for (var tag in group) { newList.push(group[tag]); }
+  return newList;
+});
+export default collectBy;
--- a/test/collectBy.js
+++ b/test/collectBy.js
@@ -1,0 +1,37 @@
+var {all, compose , difference , equals , head , identity , is , isEmpty , length , uniq , unnest , collectBy} = require('../source');
+var {property} = require('jsverify');
+var {spy} = require('sinon');
+
+describe('collectBy', function() {
+
+  property('returns a list of lists', '[nat]', function(xs) {
+    var check = all(is(Array));
+    var ys = collectBy(identity)(xs);
+    return check(ys);
+  });
+
+  property('groups items but neiter adds new ones nor removes any', '[nat]', function(xs) {
+    var check = compose(isEmpty, difference(xs), unnest);
+    var ys = collectBy(identity)(xs);
+    return check(ys);
+  });
+
+  property('groups related items together', '[bool]', function(xs) {
+    var ys = collectBy(identity)(xs);
+    var check = all(compose(equals(1), length, uniq));
+    return check(ys);
+  });
+
+  property('invokes the tag function for each item in the list', '[nat]', function(xs) {
+    var id = spy(x => 42);
+    collectBy(id)(xs);
+    var check = compose(isEmpty, difference(xs));
+    return check(id.getCalls().map(call => call.args[0]));
+  });
+
+  property('groups items according to the tag value', '[nat]', function(xs) {
+    var ys = collectBy(x => 42)(xs);
+    var check = compose(isEmpty, difference(xs), head);
+    return isEmpty(xs) && isEmpty(ys) ? true : check(ys);
+  });
+});
--- a/source/index.js
+++ b/source/index.js
@@ -26,6 +26,7 @@ export { default as call } from './call';
 export { default as chain } from './chain';
 export { default as clamp } from './clamp';
 export { default as clone } from './clone';
+export { default as collectBy } from './collectBy';
 export { default as comparator } from './comparator';
 export { default as complement } from './complement';
 export { default as compose } from './compose';


source/collectBy.js
```diff
@@ -0,0 +1,37 @@
...
+import values from './values';
+
+/**
+ * Splits a list into sub-lists, based on the result of calling a String-returning function on each element,
+ * and grouping the results according to values returned.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig (a -> String) -> [a] -> {String: [a]}
```
customcommander: I'm not familiar with the Hindley-Milner notation

source/collectBy.js
```diff
@@ -0,0 +1,37 @@
...
+ * @param {Function} fn Function :: a -> String
+ * @param {Array} list The array to split into sub-arrays.
+ * @return {Array}
+ *    An array of arrays where each sub-array contains items for which
+ *    the String-returning function has returned the same value.
+ * @see R.groupBy
```
customcommander: Should `R.partition` be mentioned?

CrossEye: Yes, I think that's a good idea.  It's quite related.  In fact, `collectBy(pred)` will group a list into the same groups that `partition(pred)` does, although the order might not be the same.

source/collectBy.js
```diff
@@ -0,0 +1,37 @@
...
+ * @example
+ *      R.collectBy(R.prop('type'), [
+ *        {type: 'breakfast', item: '☕️'},
+ *        {type: 'lunch', item: '🌯'},
+ *        {type: 'dinner', item: '🍝'},
+ *        {type: 'breakfast', item: '🥐'},
+ *        {type: 'lunch', item: '🍕'}
+ *      ]);
```
CrossEye: I think these should be in a slightly different order so that the result could not be mistaken for something equivalent to `compose(transpose, splitEvery(3))`

test/collectBy.js
```diff
@@ -0,0 +1,25 @@
...
+      [{type: 'breakfast', item: '☕️'}, {type: 'breakfast', item: '🥐'}],
+      [{type: 'lunch', item: '🌯'}, {type: 'lunch', item: '🍕'}],
+      [{type: 'dinner', item: '🍝'}]
+    ]);
+  });
+
+  it('returns an empty list if given an empty array', function() {
+    eq(R.collectBy(R.prop('x'), []), []);
+  });
+
```
CrossEye: I think we might want a test-case with a function that returns numbers.  (Maybe my modulus example above.)

source/collectBy.js
```diff
@@ -0,0 +1,37 @@
...
+ *
+ *      // [ [ {type: 'breakfast', item: '☕️'},
+ *      //     {type: 'breakfast', item: '🥐'} ],
+ *      //   [ {type: 'lunch', item: '🌯'},
+ *      //     {type: 'lunch', item: '🍕'} ],
+ *      //   [ {type: 'dinner', item: '🍝'} ] ]
+ */
+var collectBy = _curry2(function collectBy(fn, list) {
+  return values(groupBy(fn, list));
```
CrossEye: If you have some time, could you compare the performance to a iteration-based approach?  There may not be much difference, in which case, we can keep this simple code.  But if the difference is substantial (perhaps: a factor of more than 2), we should probably switch to more performant code.

customcommander: How do you usually do this? Do you use any benchmarking library?

test/collectBy.js
customcommander: Makes sense

source/collectBy.js
customcommander: Ok will do

source/collectBy.js
customcommander: sig tag updated with your suggestion

source/collectBy.js
customcommander: Mentioned `partition` in see tag

source/collectBy.js
customcommander: Added more objects and "shuffled" the array to avoid seeing patterns where there are none. What do you think?

source/collectBy.js
customcommander: If a grouping function returns a non-string value must cause `collectBy` to return an empty list, then I'd probably have to implement with classic loop statements. Do you still want to me to measure the performance in that case and what would be an acceptable threshold?

test/collectBy.js
CrossEye: I am starting to find review conversations harder to follow on GitHub.  I thought I had made a suggestion for a simpler example, but I can't find it now.  The version of this in the test suite seems perfect, but here, something like this might be simpler:

```js
collectBy (n => n % 10, [5, 11, 12, 22, 23, 25, 32, 35, 43])
//=> [[11], [12, 22, 32], [23, 43], [5, 25, 35]]
```

source/collectBy.js
CrossEye: > Do you use any benchmarking library?

Usually a quick test in [jsperf][jp] or [JSBench.me][jb] is plenty.  This doesn't have to live long.

  [jb]:https://jsbench.me/
  [jp]: https://jsperf.com/

CrossEye: I would simply use the default object behavior for keys, which should take no effort: 

```js
const foo = {}
foo[true] = 1
foo[{a: 1}] = 2
foo[/[a-z]/] = 3
foo[(n) => n * n] = 4
// etc.

foo //=>
// {
//   "/[a-z]/": 3,
//   "[object Object]": 2,
//   "function (n) {\n  return n * n;\n}": 4, // (or whatever)
//   "true": 1
// }
```

If you peruse our source code, many functions which could easily be written in terms of other ones are written natively, often with `while` loops.  The rationale is that the library does this nasty stuff so you don't have to.

customcommander:
@CrossEye I'll finish this!

CrossEye:
@customcommander:

I'd forgotten all about it, I'm afraid.

test/collectBy.js
customcommander: I resorted to property-based testing since you `jsverify` is available.

source/collectBy.js
customcommander: I ended up writing the function "natively". Would you still see some performance benchmark?

customcommander:
@CrossEye There you go :-) I resolved all comments. Can you check and see if there's room for improvements?

source/collectBy.js
CrossEye: No.  The only reason for that was that if we were skipping our usual style, I wanted to ensure that it didn't hurt performance.

test/collectBy.js
CrossEye: That's fine.  I have very mixed feelings about property-based testing.  It's clearly a better way to catch edge cases, and that's the most important facet of unit tests.  But a strong secondary feature of unit tests is to serve as an extra API guide.  When I look to see how a project or a new feature works, I usually look first to the test suite rather than the docs.  That doesn't work nearly as well with property-based testing.

CrossEye:
:herb:

This looks great to me.  Anyone else? @ramda/core

source/collectBy.js
```diff
@@ -0,0 +1,44 @@
...
+import _reduce from './internal/_reduce';
+
+/**
+ * Splits a list into sub-lists, based on the result of calling a String-returning function on each element,
+ * and grouping the results according to values returned.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig (a -> String) -> [a] -> [[a]]
```
hitmands, line 11: >  based on the result of calling a String-returning function on each element

I am not sure about the constraints of a string returning function, this might be required for `groupBy` but seems to not be strictly needed here.

`collectBy` really looks like `partition` without the need for the returned tuple to be a pair.

```ts
// Tuple length is 2
// eq to partition as fn is a predicate
const [evens, odds] = R.collectBy((n) => n % 2 === 0, [1, 2, 3, 4, 5]);
``` 

```ts
const bySpeed = (speed) => {
  if (speed < 30) { return 0; }
  if (speed < 90) { return 1; }
  return 2;
};

const [slow, average, fast] = R.collectBy(bySpeed, [10, 30, 70, 250]);
```

Probably better to reword

source/collectBy.js
```diff
@@ -0,0 +1,44 @@
...
+    var tag = fn(x);
+    if (o[tag] === undefined) { o[tag] = []; }
+    o[tag].push(x);
+    return o;
+  }, {}, list);
+  var newList = [];
+  for (var tag in group) { newList.push(group[tag]); }
+  return newList;
+});
+export default collectBy;
```
hitmands, line 44: this is more of a question, 
but why the implementation isn't aliasing `groupBy . values` ?

source/collectBy.js
customcommander: I _think_ the string requirement is just a clue for the developer. Related items are grouped under the same key of an object and keys are always strings. The function can indeed return anything but it will be automatically casted into a string.

@CrossEye and I did discuss this already. See https://github.com/ramda/ramda/pull/2715#discussion_r236656183

source/collectBy.js
customcommander: This was actually the case in my first attempt. I believe @CrossEye had some questions about the performance and hinted at implementing from "scratch".

source/collectBy.js
hitmands: sounds fair, in the end it's more about wording.

I'd personally prefer "A function returning a valid property key".

- [Typescript / PropertyKey](https://github.com/microsoft/TypeScript/blob/master/src/lib/es5.d.ts#L85)
- [Symbols as PropertyKey](https://2ality.com/2014/12/es6-symbols.html#symbols-as-property-keys)

CrossEye: I missed this thread when I merged.  @hitmands: I think you're right about the wording.  We should probably change it that way.  Somewhere else we use a signature like

```hs
somefunc :: Index a => (b -> a) -> ?? -> Object
Index == String | Int | Symbol
```

We should probably get that cleaned up and make this a habit for describing functions that return Object keys.  And the description should match.

That can be a separate task, though.

PR #2717, opened 2018-11-26
Update docs for memoizeWith

peoplenarthax:
The documentation for memoizeWith was not updated when memoize was removed.

This PR simply removes the reference to memoize.

### DIFF
--- a/source/memoizeWith.js
+++ b/source/memoizeWith.js
@@ -4,11 +4,10 @@ import _has from './internal/_has';
 
 
 /**
- * A customisable version of [`R.memoize`](#memoize). `memoizeWith` takes an
- * additional function that will be applied to a given argument set and used to
- * create the cache key under which the results of the function to be memoized
- * will be stored. Care must be taken when implementing key generation to avoid
- * clashes that may overwrite previous entries erroneously.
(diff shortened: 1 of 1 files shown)


CrossEye:
Thank you.  Good catch!

source/memoizeWith.js
```diff
@@ -4,11 +4,10 @@ import _has from './internal/_has';
 
 
 /**
- * A customisable version of [`R.memoize`](#memoize). `memoizeWith` takes an
- * additional function that will be applied to a given argument set and used to
- * create the cache key under which the results of the function to be memoized
- * will be stored. Care must be taken when implementing key generation to avoid
- * clashes that may overwrite previous entries erroneously.
+ * Takes an additional function that will be applied to a given argument set
```
Bradcomp: Additional is no longer applicable here now that Memoize doesn't exist. Perhaps we should adapt the docs from `memoize` since we can't reference them anymore:

```
/**
 * Creates a new function that, when invoked, caches the result of calling `fn`
 * for a given argument set and returns the result. Subsequent calls to the
 * memoized `fn` with the same argument set will not result in an additional
 * call to `fn`; instead, the cached result for that set of arguments will be
 * returned.
*/
```

PR #2718, opened 2018-11-26
Reference to Ramda Conventions wiki page

tommmyy:
Regarding to https://github.com/ramda/ramda/issues/2706#issuecomment-441712047


CONTRIBUTING.md
```diff
@@ -12,17 +12,19 @@
...
 
-4.  Make one or more atomic commits. Each commit should have a descriptive
+4.  Before designing a new function please read our [Ramda Conventions](https://github.com/ramda/ramda/wiki/Ramda-Conventions) article.
+
+5.  Make one or more atomic commits. Each commit should have a descriptive
     commit message, wrapped at 72 characters. Do not commit changes to
     `dist/ramda.js`.
 
-5.  Run `npm test` (or `make test lint`) and address any errors. It will install
+6.  Run `npm test` (or `make test lint`) and address any errors. It will install
```
Bradcomp: Markdown will let us define an ordered list without hard-coding the numbers in. If you just put in `1.` for _all_ the numbers it'll keep us from needing to tweak things in the future.

Bradcomp: I think you have a valid point, and I certainly don't want to cause a ruckus, but I think there is more to consider. By leaving the numbers in, we make it easier to make a mistake in the future. By making them all the same, we have eliminated the numbering as a source for problems, while leaving the final product the same.

In other words, I think that using the same sigil makes our code more resilient and reliable, and justifies using it over a slightly more readable code convention that reduces our code's resiliency.

CrossEye: As I said, I was just whining.

Most people view this content in some form generated from the raw Markdown, and then it's no issue at all.  I personally do a lot of Markdown documentation (lately using VuePress, but in all sorts of environments) and I try to ensure that the source is as readable as possible.  So for instance, I always do `[my text][mt]` with this in the footer `[mt]: https://example.com` rather than `[my text](https://example.com)` because it seems more readable to me.

But I certainly can't get worked up about this.

PR #2720, opened 2018-11-27
Change const to var

screendriver:
Because `const` will break on older browsers like in iOS 9.3


CrossEye:
:herb:

Thank  you.  I don't know how these snuck in.

screendriver:
Thank you. That was quick. This should be released as soon as possible because without that fix some applications will break without knowing what's going on 😉 (I had this kind of issue for example)

CrossEye:
I'll try to get to it tonight.  There's some other things that should make it into a quick release.

screendriver:
Great! 🎉

PR #2724, opened 2018-11-28
Use uglify instead of terser - this can act as smoke test against shi…

Andarist:
…pping es2015+ code


CrossEye:
So would you suggest reverting this change post 1.0 when we will want to embrace ES2015+ ?

Andarist:
If you plan to ship es2015+ in your main entry point (in the file pointed by "main" in package.json) then - yeah, you should revert it and minify with terser (which is uglify fork which handles es2015+)

CrossEye:
I think this makes sense, but would love to hear from @ramda/core .

paldepind:
Sounds very reasonable :+1:

Bradcomp:
I see no problem here. Is there a significant difference in bundle size? I would assume not...

Andarist:
I havent checked - there should be no difference or it should be insignificant. The sole purpose of this PR is uglify acting as a smoke test against including es2015+ syntax, not the bundle optimization.

CrossEye:
As always, thanks @Andarist!

Bradcomp:
I agree the smoke test is a good idea, but not if it comes at the expense of bundle size. I'd at least like to know we are still in the ballpark.

CrossEye:
@Andarist:

That sounds like every developer ever! :smile:

Andarist:
```
gzip -c dist/ramda.min.js | wc -c
```
terser 12574
uglify 12508

So actually its slightly better in favour of uglifyJS 😄

Issue #2723, opened 2018-11-28
v0.26.1 Upgrade Guide


CrossEye:
[v0.26.0...v0.26.1](https://github.com/ramda/ramda/compare/v0.26.0..v0.26.1)

:bulb: Changes:

- #2720 removed some `const`s that had snuck in before v1.0

:clipboard: Documentation fixes: #2708, #2709, #2717, #2722

:heavy_check_mark: Testing improvements:  #2703, #2710, #2711

----------

Many thanks to all the contributors supporting this release, especially to those helping with improvements to tests and documentation :tada:

Captainlonate:
@CrossEye 
Would it make sense to add that `move` was added in 26**.1**, to the "Changes" notes? Was actually a little challenging to track down exactly when that was added.

CrossEye:
Yes it would, if that's what happened.  I'll try to research this evening.

PR #2725, opened 2018-11-28
Removes reference to 'memoize' from documentation

ptolemybarnes:


CrossEye:
I think this is covered by #2722.  While I published this in the [new version][nv], I ran into some issue last night trying to upgrade the documentation..  I will try again tonight.


  [nv]: https://github.com/ramda/ramda/issues/2723

PR #2728, opened 2018-11-29
Fixed passing banner option to rollup

Andarist:
I've missed this when upgrading rollup in https://github.com/ramda/ramda/pull/2724


CrossEye:
Thanks!  I really need to learn more about rollup!

Issue #2733, opened 2018-12-03
(pipeWith) transforming function is not applied on the 1st function


franzwong:
From the source code of `pipeWith`, it said `R.pipeWith(f)([g, h, i])(...args) = f(i, f(h, f(g, ...args)))`. 

I have created a test.

```
const R = require('ramda')

const trace = (f, res) => {
  const newRes = f(res)
  console.log(`old: ${res}, new: ${newRes}`)
  return newRes
}

const incrementThree = R.pipeWith(trace)([R.inc, R.inc, R.inc])

console.log('Try incrementTree')
incrementThree(10)

console.log('Try trace(R.inc, trace(R.inc, trace(R.inc, 10)))')
trace(R.inc, trace(R.inc, trace(R.inc, 10)))

console.log('Try trace(R.inc, trace(R.inc, R.inc(10)))')
trace(R.inc, trace(R.inc, R.inc(10)))
```

Here is the result

```
Try incrementTree
old: 11, new: 12
old: 12, new: 13
Try trace(R.inc, trace(R.inc, trace(R.inc, 10)))
old: 10, new: 11
old: 11, new: 12
old: 12, new: 13
Try trace(R.inc, trace(R.inc, R.inc(10)))
old: 11, new: 12
old: 12, new: 13
```

It looks like `pipeWith` is not working as its comment says. I don't know what the expected behaviour is.

CrossEye:
I think it's working as expected.  If you can think of a way to clean up the comments, please feel free.  (See #2512 for the motivation.)

The idea is that now we can replace these

```js
pipeP(f, g, h)
pipeK(f, g, h)
```
with

```js
pipeWith(then, [f, g, h])
pipeWith(chain, [f, g, h])
```

as well as writing

```js
pipeWith(notNil, [f, g, h])
```

or whatever else we want to do between function invocations.  We could theoretically implement `pipe` and `compose` on top of these, using `identity`, but choose not to as they are some of the most central functions in the library.

franzwong:
Updated with more accurate comment

PR #2734, opened 2018-12-04
Update the comment about pipeWith

franzwong:
See #2733

### DIFF
--- a/source/pipeWith.js
+++ b/source/pipeWith.js
@@ -10,7 +10,8 @@ import identity from './identity';
  * Performs left-to-right function composition using transforming function. The leftmost function may have
  * any arity; the remaining functions must be unary.
  *
- * **Note:** The result of pipeWith is not automatically curried.
+ * **Note:** The result of pipeWith is not automatically curried. Transforming function is not used on the
+ * leftmost function.
  *
  * @func
  * @memberOf R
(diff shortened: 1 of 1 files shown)


source/pipeWith.js
```diff
@@ -10,7 +10,8 @@ import identity from './identity';
  * Performs left-to-right function composition using transforming function. The leftmost function may have
  * any arity; the remaining functions must be unary.
  *
- * **Note:** The result of pipeWith is not automatically curried.
+ * **Note:** The result of pipeWith is not automatically curried. Transforming function is not used on the
```
CrossEye: I would rather love to find some alternative to `"leftmost" and "rightmost" (here and in `pipe`, `compose` and `composeWith`.)  While it's probably clear enough most of the time, it's not really accurate, in that I often lay out `pipe` functions vertically.

If you have any suggestions, please let me know.

franzwong: Updated for `composeWith`. How about simply `first argument` and `last argument`?

CrossEye: Thank you.

Yes, I think that `first argument` and `last argument` would be better.  

It's not ideal, but anything I can come up with that's really accurate sounds like a legalism.

franzwong: Updated. I've updated `compose`, `composeP`, `pipe`, `pipeP` as well. :)

PR #2735, opened 2018-12-05
fixes missing 'b' in readme.md

lukeaus:


CrossEye:
Well, first off, we need a little more information on *why* you are suggesting this.  What problems does `bit` solve for Ramda users or for the Ramda team?  Conversely, what headaches might it add?

Secondly, we're not replacing our README.md with something dedicated to `bit`.  :laughing: 

Third, does `bit` mandate exposing the internal functions?  While I would expect to include them in their dependent public functions, they are not meant for direct public consumption.  If we were to accept this, could we expose only the public functions?

Fourth, what's with the `utils` folder?  Ramda doesn't have such a folder; is this a necessary artifact of `bit`?

Finally, this issue really needs a better title!

lukeaus:
Sorry. Cancel this - wrong repo. Apologies.

CrossEye:
Ok.  You now have me investigating `bit`.  Was that part of your nefarious plan?  :smile:

lukeaus:
lol...no - they forked Ramda to use it as an [example](https://github.com/teambit/ramda) of how to use Bit.

PR #2737, opened 2018-12-06
First time

SONG-PC:
make _curryN clearer


foxbunny:
@SONG-PC You should look at the warnings made by travis. Also, probably be more clearer in the description about what in your changes makes `_curryN` 'clearer'.

PR #2739, opened 2018-12-08
fix(hasPath): forces hasPath to be safe

### LINKED ISSUE #2738: hasPath is not safe
Would it be possible to make `hasPath` function safe ? It works as a predicate and I'd expect the predicates to be safe.

```js
hasPath(['a', 'b'], null); // throws Error
```
In [ramda-adjunct](https://github.com/char0n/ramda-adjunct/blob/master/src/hasPath.js) we have this function for more that a year, although is has recursive implementation it does the same but it's safe. The tests doesn't reflect what happens if the 2nd argument is not an object: https://github.com/ramda/ramda/blob/master/test/hasPath.js

char0n:
The function now doesn't throw Errors anymore when supplied with non-object.
Also fixes R.has.

Closes #2738

Signed-off-by: Vladimir Gorej <vladimir.gorej@gmail.com>

### DIFF
--- a/test/has.js
+++ b/test/has.js
@@ -22,4 +22,15 @@ describe('has', function() {
     eq(R.has('age', bob), false);
   });
 
+  it('tests paths on non-objects', function() {
+    eq(R.has('a', undefined), false);
+    eq(R.has('a', null), false);
+    eq(R.has('a', true), false);
+    eq(R.has('a', ''), false);
+    eq(R.has('a', /a/), false);
+  });
+
+  it('tests currying', function() {
+    eq(R.has('a')({ a: { b: 1 } }), true);
+  });
 });
--- a/test/has.js
+++ b/test/has.js
@@ -22,7 +22,7 @@ describe('has', function() {
     eq(R.has('age', bob), false);
   });
 
-  it('tests paths on non-objects', function() {
+  it('returns false for non-objects', function() {
     eq(R.has('a', undefined), false);
     eq(R.has('a', null), false);
     eq(R.has('a', true), false);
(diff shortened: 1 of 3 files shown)


test/hasPath.js
```diff
@@ -50,4 +59,16 @@ describe('hasPath', function() {
   it('returns false for empty path', function() {
     eq(R.hasPath([], obj), false);
   });
+
+  it('tests paths on non-objects', function() {
```
CrossEye: perhaps `it('returns false for non-objects', ...)`

'tests' is very non-specific.

test/has.js
```diff
@@ -22,4 +22,15 @@ describe('has', function() {
     eq(R.has('age', bob), false);
   });
 
+  it('tests paths on non-objects', function() {
```
CrossEye: I'd like to find better phrasing here, too, although it's harder to come up with something.

test/hasPath.js
char0n: yeah we can do that, or we can utilize mochas contexts and [gherkin DSL](http://docs.behat.org/en/v2.5/guides/1.gherkin.html)
```js
context('given second argument is a non-object', function() {
  specify('should return false', function() {
    assertions
  });
});
```
@CrossEye let me know if this is something you like. `context` is alias of `describe`, `specify` of `it`. In this case we have a top level `describe`, we have inner `context` inside `describe` and `specify` inside it.

CrossEye: I was only worried about the actual text of the test.  This reads fine to me, but I don't have any real preference between nested `describe`s and this `context`.

char0n:
@CrossEye addressed code review comments.

CrossEye:
This LGTM.  :herb:

Any other comments?  @ramda/core?

Bradcomp:
:bowtie:

CrossEye:
Thanks, @char0n !

PR #2741, opened 2018-12-12
Don't curry transducers

GingerPlusPlus:


source/internal/_dispatchable.js
```diff
@@ -32,7 +32,7 @@ export default function _dispatchable(methodNames, xf, fn) {
         idx += 1;
       }
       if (_isTransformer(obj)) {
-        var transducer = xf.apply(null, args);
+        var transducer = transformer => new xf(...args, transformer);
```
CrossEye: Pre-1.0 our code (although not the examples) is not using ES6 features such as arrow functions.

trgwii: When will we ES6 everything? I think the transpiler toolchain is mature enough for Ramda to embrace the "new javascript" while still retaining support for at least ES5

CrossEye: I'm trying to write a suggested roadmap document; if possible I will finish it this weekend and present it to the core team.  I'm hoping then to actually release a 1.0 within a few weeks.  After that, our main development will embrace more modern JS, although I don't know with exactly what toolchain.

GingerPlusPlus:
Reason? It's a small simplification with no drawbacks.

It also allows `_dispatchable` to access the transducers classes' directly.

CrossEye:
I will take a closer look this weekend.  There is a build error to deal with, probably because of the arrow function.

source/internal/_dispatchable.js
trgwii: 🎉 Would be great to have a public roadmap, so far, roadmap has been scouring issues for things marked with 1.0 and similar, some definitions of what high-level changes we want would be great

GingerPlusPlus:
Arrow function can be replaced and remain neat, argument spread cannot.

<hr/>

I could do the same thing I did to transducers to `_xwrap`.

Issue #2740, opened 2018-12-10
Proposal: Can we have an operator similar to https://lodash.com/docs/4.17.11#at


minaseem:

char0n:
You have it in ramda-adjunct : https://char0n.github.io/ramda-adjunct/2.13.0/RA.html#.paths

CrossEye:
While I'd love to hear some more feedback, it seems to me that this would complete this square:

|                   | Single node        | Multiple nodes       |
| ----------------- | ------------------ | -------------------- |
| **Scalar input**  | [`prop`][pr]       | [`path`][pa]         |
| **Array input**   | [`props`][ps]      | `??`                 |

Obviously it's not essential.  It's fairly simple to write as `compose(juxt, map(path))`, or to wrap that in `uncurryN(2)` if that exact API is wanted.  But it does seem sensible to add.

@minaseem: Are you interested in creating a PR for this?

  [pa]: https://ramdajs.com/docs/#path
  [pr]: https://ramdajs.com/docs/#prop
  [ps]: https://ramdajs.com/docs/#props

Bradcomp:
I think the name, per @char0n, should be `paths` rather than `at` to mirror `props`. Other than that this seems like a good idea to me.

minaseem:
@CrossEye Can you please review PR - https://github.com/ramda/ramda/pull/2742

PR #2742, opened 2018-12-12
feature: adding paths operator - #2740

minaseem:
Link for proposal discussion - https://github.com/ramda/ramda/issues/2740

### DIFF
--- a/source/index.js
+++ b/source/index.js
@@ -160,6 +160,7 @@ export { default as partial } from './partial';
 export { default as partialRight } from './partialRight';
 export { default as partition } from './partition';
 export { default as path } from './path';
+export { default as paths } from './paths';
 export { default as pathEq } from './pathEq';
 export { default as pathOr } from './pathOr';
 export { default as pathSatisfies } from './pathSatisfies';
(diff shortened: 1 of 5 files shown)


minaseem:
Thank you @CrossEye for reviewing this.
If I understand correctly, below is the proposed change:

Implement `paths` natively/using internal helper functions, and then reimplement`prop` and `path` using `paths`. Please tell me whether my understanding is correct.

**Question:**
Since you said, `Ramda writes most of its object and array handling natively, without depending on other of its public functions`, I see `nth` being used in the implementation of [path](https://github.com/ramda/ramda/blob/master/source/path.js#L3). So, it is ok to use `nth` while writing `paths` natively?

minaseem:
Hi @CrossEye 
Changes suggested by you have been completed. Please review and let me know if it ok.

PS: Apologies for the delay

CrossEye:
This looks good to me.  Any other responses, @ramda/core?

Bradcomp:
:bowtie: 👍

CrossEye:
Thanks, @minaseem!

minaseem:
Glad to be able to contribute

PR #2744, opened 2018-12-17
Refactor transpose.js to make it readable

ChanwheKim:
I just changed while-loop to for-loop of transpose.js beause I thinks it looks a little bit more readable. :)
* Change while-loop to for-loop of transpose.js
* Change var variables to const and let variables

### DIFF
--- a/source/transpose.js
+++ b/source/transpose.js
@@ -25,20 +25,18 @@ import _curry1 from './internal/_curry1';
  * @symb R.transpose([[a, b], [c, d]]) = [[a, c], [b, d]]
  * @symb R.transpose([[a, b], [c]]) = [[a, c], [b]]
  */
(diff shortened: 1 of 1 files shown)


CrossEye:
While the Travis build failed because we're still not allowing `const` and other ES6 syntax into the code before v1.0, the bigger problem is that house style is the `while` loops.  One of the founding members of Ramda has made it a personal goal to never write another `for` loop.  And when Ramda started, there was still some truth to the folklore that `while` loops were more performant.  I think that time has now passed, but you shouldn't find *any* `for` loops in the Ramda code.

Bradcomp:
Not a founding member, but I too will never write a for loop in javascript, and will always suggest an alternative during code review.

PR #2750, opened 2018-12-19
Added function ifElseP

sebyul2:


GingerPlusPlus:
Isn't this equivalent to `curry(compose(then, ifElse))`?

CrossEye:
@GingerPlusPlus:

Perhaps maybe I misread what you were suggesting.  I believe that @sebyul2  was looking for this:

```js
// delay :: (a -> b) -> a -> Promise b
const delay = (fn) => (a) => new Promise((res => setTimeout(() => res(fn(a)), 1000)))

// ifElseP :: (a -> Promise Bool) -> (b -> c) -> (b -> c) -> a -> Promise c
const ifElseP = (f, g, h) => (a) => then(b => (b ? g : h)(a), f(a))

const fn = ifElseP(
  delay(a => a > 10),
  Math.sqrt,
  identity
)

fn(25).then(console.log) //~> Promise(5)
fn(9).then(console.log)  //~> Promise(9)
```
And I thought you were suggesting something like this:

```js
// ifElseP :: (a -> Bool) -> (b -> Promise c) -> (b -> Promise c) -> a -> Promise c
const ifElseP = (f, g, h) => (a) => f(a) ? g(a) : h(a)

const fn = ifElseP(
  a => a > 10,
  delay(Math.sqrt),
  delay(identity)
)
```

But perhaps you meant something like this:

```js
// ifElseP :: (a -> Promise Bool) -> (b -> Promise c)  -> (b -> Promise c) -> a -> Promise c
const ifElseP = (f, g, h) => (a) => then(b => b ? g(a) : h(a), f(a))

const fn = ifElseP(
  delay(a => a > 10),
  delay(Math.sqrt),
  delay(identity)
)
```

In any case, none of these seems to match `curry(compose(then, ifElse))`.  Am I missing something?

sebyul2:
Thank you for your many comments.

@CrossEye It worked exactly for me. I have studied. Thank you.

GingerPlusPlus:
Please close this PR then.

In general, @CrossEye, please close PRs that have been "verbally rejected" by a core member and have received no activity since, they are just cluttering the list of open PRs.

If you're not feeling like it, I volunteer &mdash; during the holidays I'll sift through all the PRs, close the ones that without a doubt should be closed, and attempt to properly label the rest. I'd need a (temporary) write access to the repo. I promise to not use it for anything else.

<hr />

Back to the topic of `curry(compose(then, ifElse))`, @CrossEye, your 2nd snippet is subset of `ifElse`.
I assume this is the signature of `ifElseP2` you meant to write (`onTrue` and `onFalse` take the same type as the predicate):

```
ifElseP2 :: (a -> Bool) -> (a -> Promise c) -> (a -> Promise c) -> a -> Promise c
```
If we restrict the signature of `ifElse` to unary functions

```
ifElse :: (a -> Bool) -> (a -> b) -> (a -> b) -> a -> b
```

Substitute `b = Promise c`, and you get the signature of `ifElseP2`.

<hr />

3rd one... uh, I think you meant `b => b(a) ? g(a) : h(a)`, not `b => b ? g(a) : h(a)`.

<hr />

> In any case, none of these seems to match `curry(compose(then, ifElse))`. Am I missing something?

My point is that you can use `then` to lift any unary function to take and return a promise:

```js
// p, f :: a -> Promise b | b
// foo :: a -> Promise b | b
const foo = ifElse(p, t, f);

// then :: (a -> Promise b | b) -> Promise a -> Promise b

// fooP :: Promise a -> Promise b
const fooP = then(foo)
```

And the `curry(compose` just... composes `then` with `ifElse`, duh.

CrossEye:
@GingerPlusPlus: Thank you.  I'm going through the issues and will close a great number of them soon, and open up a discussion about which of the others we want to include in 1.0, and how we want to proceed from there.  I may call for help in a few days if it's not going well.  But I should have some time, as I finish my current job tomorrow and don't start my new one until the new year.

----------

As to the code under discussion, you are right about the signatures -- I stopped part-way through a transformation from a more specific case to something generic.  I really should have noticed.  And I also hadn't noticed that the second one was in fact just a version of `ifElse`.  :smile:

As to the third one, no, I'm pretty sure that what I had is right.  An equivalent with `Promise.prototype.then` rather than `R.then` would be

```js
const ifElseP = (f, g, h) => (a) => f(a).then(b => b ? g(a) : h(a))
```

For those following along at home:

  - **[Version 1][v1]**:

      ```js
      // ifElseP :: (a -> Promise Bool) -> (a -> b) -> (a -> b) -> a -> Promise b
      const ifElseP = (f, g, h) => (a) => then(b => (b ? g : h)(a), f(a))
      ```
  - **[Version 2][v2]**:

      ```js
      // ifElseP :: (a -> Bool) -> (a -> Promise b) -> (a -> Promise b) -> a -> Promise b
      const ifElseP = (f, g, h) => (a) => f(a) ? g(a) : h(a)
      ```

  - **[Version 3][v3]**:

      ```js
      // ifElseP :: (a -> Promise Bool) -> (a -> Promise b)  -> (a -> Promise b) -> a -> Promise b
      const ifElseP = (f, g, h) => (a) => then(b => b ? g(a) : h(a), f(a))
      ```

And what you're suggesting, I guess, is

  - **[Version 4][v4]**:
      ```js
      // ifElseP :: (a -> Bool) -> (a -> Promise b)  -> (a -> Promise b) -> Promise a -> Promise b
      const ifElseP = curry(compose(then, ifElse))
      ```

which is used a bit differently from the others, in that the function returned from `ifElseP` accepts a `Promise a`, not just an `a`.  While that might be an interesting function on it's own, I don't think it has much to do with the original request, which seems to be something like [version 1][v1].

  [v1]: https://goo.gl/gRjgcD
  [v2]: https://goo.gl/aXZjAt
  [v3]: https://goo.gl/x9e9Nr
  [v4]: https://goo.gl/ULxYPL

CrossEye:
Yes, a cookbook implementation makes sense.

Since this is the very first request I've seen for something like this, I'm not inclined to spend any effort on `ifElseWith`, but if someone created a PR, we could certainly give a look.

It's funny after all those versions that I still hadn't noticed that 1 and 3 were the same.  It's that damn `chain`/`map` combination.  If `then` were more sensible, a specialization of version 1 where `b` is `Promise c`  would yield `Promise Promise c`.

PR #2752, opened 2019-01-02
Add `omitArgs`

juusaw:
Adds the function `omitArgs` which wraps a function of any arity into a (left) padded function with `n` parameters that are ignored before the original parameters.

The code is not very readable due to backwards compatibility and array length property but the following does the same in more idiomatic JS: 

```javascript
const omitArgs = (num, fn) => (...args) => fn.apply(null, args.slice(num))
```

I've previously found the function useful for example when writing callbacks for asynchronous functions that pass the callback some data it doesn't need. `omitArgs` enables omitting unnecessary parameters from the callback signature.

I'd be happy to hear if there are any suggestions for a more intuitive name for the function.

### DIFF
--- a/source/omitArgs.js
+++ b/source/omitArgs.js
@@ -32,17 +32,17 @@ var omitArgs = _curry2(function omitArgs(n, fn) {
   }
   var arity = n + fn.length;
   switch (arity) {
-    case 0: return function() { return fn.apply(this, Array.from(arguments).slice(n)); };
-    case 1: return function(a0) { return fn.apply(this, Array.from(arguments).slice(n)); };
-    case 2: return function(a0, a1) { return fn.apply(this, Array.from(arguments).slice(n)); };
-    case 3: return function(a0, a1, a2) { return fn.apply(this, Array.from(arguments).slice(n)); };
-    case 4: return function(a0, a1, a2, a3) { return fn.apply(this, Array.from(arguments).slice(n)); };
-    case 5: return function(a0, a1, a2, a3, a4) { return fn.apply(this, Array.from(arguments).slice(n)); };
-    case 6: return function(a0, a1, a2, a3, a4, a5) { return fn.apply(this, Array.from(arguments).slice(n)); };
-    case 7: return function(a0, a1, a2, a3, a4, a5, a6) { return fn.apply(this, Array.from(arguments).slice(n)); };
-    case 8: return function(a0, a1, a2, a3, a4, a5, a6, a7) { return fn.apply(this, Array.from(arguments).slice(n)); };
-    case 9: return function(a0, a1, a2, a3, a4, a5, a6, a7, a8) { return fn.apply(this, Array.from(arguments).slice(n)); };
-    case 10: return function(a0, a1, a2, a3, a4, a5, a6, a7, a8, a9) { return fn.apply(this, Array.from(arguments).slice(n)); };
+    case 0: return function() { return fn.apply(this, Array.prototype.slice.call(arguments).slice(n)); };
+    case 1: return function(a0) { return fn.apply(this, Array.prototype.slice.call(arguments).slice(n)); };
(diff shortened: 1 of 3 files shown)


CrossEye:
I'd be curious to see what others in @ramda/core think.

While I can see the use of this function, I don't find it particularly compelling.  It's possible I'd like it better with a different name, but I don't have a suggestion.

@juusaw: Do you have any argument to convince me that this is important to Ramda users?

juusaw:
I believe the function complements `nAry` nicely and by having both, it is possible to wrap a function so that arguments in the beginning and end can be omitted. The rationale for including the function is somewhat the same as for `nAry`.

For a real world example of usage, `passport.js` authentication middleware's OAuth2 strategies require a verify callback that receives the user and session information. The function signature is `(accessToken, refreshToken, profile, cb) => void`. In many cases only the profile and cb are used. We can define a verify callback like `const verify = (profile, cb) => cb(null, profile)` and provide it to the middleware as `R.omitArgs(2, verify)`. This helps in keeping the verify function signature clean improves readability and reusability.

As for the naming ideas, some alternatives include skipArgs (suggested by a colleague), dropArgs and padArgs.

robyoder:
I'm not part of ramda/core and don't have a compelling use case, but wanted to comment on the name. `omitArgs(2, verify)` feels like `nAry(0, verify)` to me. It feels like I'm ignoring the two params to verify, not adding two ignored params. I think `padArgs(2, verify)` makes the most sense there. 

I was trying to come up with a way to get `nAryLeft(4, verify)` to work because that makes sense to me and could be extended to `nAryLeft(1, verify)`, but then what would that mean? Pass `undefined` as the first param to `verify`?

source/omitArgs.js
```diff
@@ -0,0 +1,50 @@
...
+ * @param {Function} fn The function to wrap.
+ * @return {Function} A new function wrapping `fn`.
+ * @see R.nAry
+ * @example
+ *
+ *      const takesTwoArgs = (a, b) => a + b;
+ *
+ *      takesTwoArgs(1, 2); //=> 3
+ *
+ *      const ignoresTwoArgs = R.omitArgs(2, takesTwoArgs);
```
CrossEye: I would make `n` either 1 or 3 here, so that it's easily distinguised from the underlying functions's arity.

source/omitArgs.js
```diff
@@ -0,0 +1,50 @@
...
+    case 3: return function(a0, a1, a2) { return fn.apply(this, Array.prototype.slice.call(arguments).slice(n)); };
+    case 4: return function(a0, a1, a2, a3) { return fn.apply(this, Array.prototype.slice.call(arguments).slice(n)); };
+    case 5: return function(a0, a1, a2, a3, a4) { return fn.apply(this, Array.prototype.slice.call(arguments).slice(n)); };
+    case 6: return function(a0, a1, a2, a3, a4, a5) { return fn.apply(this, Array.prototype.slice.call(arguments).slice(n)); };
+    case 7: return function(a0, a1, a2, a3, a4, a5, a6) { return fn.apply(this, Array.prototype.slice.call(arguments).slice(n)); };
+    case 8: return function(a0, a1, a2, a3, a4, a5, a6, a7) { return fn.apply(this, Array.prototype.slice.call(arguments).slice(n)); };
+    case 9: return function(a0, a1, a2, a3, a4, a5, a6, a7, a8) { return fn.apply(this, Array.prototype.slice.call(arguments).slice(n)); };
+    case 10: return function(a0, a1, a2, a3, a4, a5, a6, a7, a8, a9) { return fn.apply(this, Array.prototype.slice.call(arguments).slice(n)); };
+    default: throw new Error('The increased arity of the function passed to omitArgs must be no greater than ten');
```
CrossEye: Wow, this makes me really want to move on to more modern JS!

CrossEye:
We're finally working to clean up old issues and push toward `v1.0`.  

@juusaw: If you're interested in still pursuing this, could you rename to `dropArgs` and clean up the conflicts?  If not, feel free to close it.

PR #2756, opened 2019-01-06
Remove _xall, _xany, _xfind(Last)?(Index)?

wojpawlik:
Transducers are useful for transforming a sequence into another sequence.

These are not that. These are about reducing a sequence to a single value.

They break `anyFunction(anyArray) === into([], anyFunction, anyArray)`, which causes [confusion](https://github.com/ramda/ramda/pull/2689#issuecomment-435619988) and [special casing](https://github.com/ramda/ramda/pull/2585/commits/cda6f0b356a6150cf680447292a0e3ee94b7e057).

What are they useful for, anyway?

The only use-case I can think about is using them with something like [Highland](https://highlandjs.org/#transduce), in which indexes don't make sense. It also defines its own [`find`](https://highlandjs.org/#find) that Ramda can dispatch to.


CrossEye:
This is another one that escaped initial detection.  At a quick glance, it looks like a good idea.  But we're trying to make no breaking changes, so I'm going to put it on hold until after `v1.0`.

PR #2757, opened 2019-01-08
prod: reduceBy splits into mutation-free groups

Beraliv:


source/reduceBy.js
```diff
@@ -50,7 +51,7 @@ var reduceBy = _curryN(4, [], _dispatchable([], _xreduceBy,
   function reduceBy(valueFn, valueAcc, keyFn, list) {
     return _reduce(function(acc, elt) {
       var key = keyFn(elt);
-      acc[key] = valueFn(_has(key, acc) ? acc[key] : valueAcc, elt);
+      acc[key] = valueFn(_has(key, acc) ? acc[key] : _clone(valueAcc, [], [], false), elt);
```
Beraliv: Here I shallowly copy `valueAcc`
However, it can be `deep`

test/reduceBy.js
```diff
@@ -41,6 +41,36 @@ describe('reduceBy', function() {
...
+      {name: 'Eddy', score: 58},
+      {name: 'Fred', score: 67},
+      {name: 'Gillian', score: 91},
+      {name: 'Hannah', score: 78},
+      {name: 'Irene', score: 85},
+      {name: 'Jack', score: 69}
+    ];
+    var byGrade = function(student) {return grade(student.score || 0);};
+    var collectNames = function(acc, student) {
+      acc.push(student.name);
```
Beraliv: Here the example with `push`

source/reduceBy.js
Bradcomp: I don't think it's worth it to do a deep clone here. Shallow clone is enough to eliminate the 'bug' and if a user is doing deep mutation of their accumulator I tend to feel like that's on them...

CrossEye:
Looks great.  Thank you very much!

PR #2758, opened 2019-01-09
Benchmark source instead of build, misc

GingerPlusPlus:


lib/bench/either.bench.js
```diff
@@ -1,4 +1,4 @@
-var R = require('../..');
+var R = require('../../source');
```
Bradcomp: We have two separate ways we're requiring the function in these tests, and we should standardize on one if we're gonna go in and change them all anyways. I _think_ I prefer `var f = require('../../source').f` over this alternative, but I would defer to others if they have a different opinion because I don't have a strong preference.

Bradcomp:
Thanks for this @GingerPlusPlus. The changes look good and I really appreciate the PRs you've been opening to improve the codebase.

Isn't it good to run the benches against the final build for accuracy? What benefit do we get from running it against the source?

GingerPlusPlus:
> What benefit do we get from running it against the source?

Mainly being unable to forget to build (happened at #2613), but also faster iteration time &mdash; I can change the source and immediately bench it.

An alternative would be to build in `prebench`.

lib/bench/either.bench.js
GingerPlusPlus: I prefer `var f = require(../../source/f)`.

CrossEye:
:+1: for `prebench`.

PR #2663, opened 2018-10-05
Add reduced support to reduceBy and reduceRight

morgsmccauley:
#2662

PR #2762, opened 2019-01-11
Ignore reduced from functions passed to reduce (3.9x to 7.2x speedup)

wojpawlik:
Changed `_reduce` to allow `reduced` to be returned only by transformers, not functions.

This is a breaking change, but it speeds up `reduce` by a factor of 3.9 to 7.2.

<details>
<summary>Benchmark</summary>
Before:
<pre>
│ reduce                 │ Hertz                  │ Margin of Error        │
│ reduce(add, 0, nums)   │ 411,807                │ 2.37%                  │
│ reduce(add, 0)(nums)   │ 400,479                │ 1.69%                  │
│ reduceAdd(nums)        │ 508,636                │ 1.19%                  │
│ native reduce          │ 4,752,316              │ 0.26%                  │
</pre>
After:
<pre>
│ reduce                 │ Hertz                  │ Margin of Error        │
│ reduce(add, 0, nums)   │ 2,911,028              │ 1.25%                  │
│ reduce(add, 0)(nums)   │ 1,520,781              │ 0.55%                  │
│ reduceAdd(nums)        │ 2,138,148              │ 0.16%                  │
│ native reduce          │ 4,843,689              │ 0.28%                  │
</pre>
</details>

Analysis of possible `_reduced` optimizations: https://github.com/ramda/ramda/pull/2613#issuecomment-453164165.

<hr />

Also, `_reduce` no longer tries to dispatch transformers to `fantasy-land/reduce`, because `reduced` support is forbidden by [the spec](https://github.com/fantasyland/fantasy-land#reduce-method):

   >  iii. `f` must return a value of the same type as `x`.
   >  iv.  No parts of `f`'s return value should be checked.

### DIFF
--- a/source/internal/_reduce.js
+++ b/source/internal/_reduce.js
@@ -3,12 +3,12 @@ import _xwrap from './_xwrap';
 import bind from '../bind';
 
 
-function _arrayReduce(xf, acc, list) {
+function _arrayReduce(xf, acc, list, reduced) {
   var idx = 0;
   var len = list.length;
   while (idx < len) {
     acc = xf['@@transducer/step'](acc, list[idx]);
-    if (acc && acc['@@transducer/reduced']) {
+    if (reduced && acc && acc['@@transducer/reduced']) {
       acc = acc['@@transducer/value'];
       break;
     }
@@ -17,11 +17,11 @@ function _arrayReduce(xf, acc, list) {
   return xf['@@transducer/result'](acc);
 }
 
-function _iterableReduce(xf, acc, iter) {
+function _iteratorReduce(xf, acc, iter, reduced) {
   var step = iter.next();
   while (!step.done) {
     acc = xf['@@transducer/step'](acc, step.value);
-    if (acc && acc['@@transducer/reduced']) {
+    if (reduced && acc && acc['@@transducer/reduced']) {
       acc = acc['@@transducer/value'];
       break;
     }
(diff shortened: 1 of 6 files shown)


source/reduced.js
```diff
@@ -7,7 +7,7 @@ import _reduced from './internal/_reduced';
  * box: the internal structure is not guaranteed to be stable.
  *
  * Note: this optimization is only available to the below functions:
- * - [`reduce`](#reduce)
+ * - [`reduce`](#reduce) (**when passed a transformer**)
  * - [`reduceWhile`](#reduceWhile)
```
Bradcomp: Will `reduced` still work on `reduceWhile`? I would think we should take it out there too. Thoughts?

test/reduce.js
```diff
@@ -71,7 +72,7 @@ describe('reduce', function() {
   });
 
   it('short circuits with reduced', function() {
```
Bradcomp: We should change the name of this test to mention transducers

test/reduce.js
```diff
@@ -71,7 +72,7 @@ describe('reduce', function() {
...
-    var addWithMaxOf10 = function(acc, val) {return acc + val > 10 ? R.reduced(acc) : acc + val;};
+    var addWithMaxOf10 = _xwrap(function(acc, val) {return acc + val > 10 ? R.reduced(acc) : acc + val;});
```
Bradcomp: Is there a way of setting this test up without using internal functions?

test/reduced.js
```diff
@@ -18,11 +19,11 @@ describe('reduced', function() {
     // black box test.
     eq(
       R.reduce(
-        function(acc, v) {
+        _xwrap(function(acc, v) {
```
Bradcomp: Same as above, I'd prefer not to resort to testing on stuff that isn't exported.

source/internal/_reduce.js
```diff
@@ -37,20 +37,25 @@ function _methodReduce(xf, acc, obj, methodName) {
 var symIterator = (typeof Symbol !== 'undefined') ? Symbol.iterator : '@@iterator';
 
 export default function _reduce(fn, acc, list) {
+  var reduced = true;
```
Bradcomp: We should change the name of this variable. It mirrors a public function, there's no indication that it's a predicate, and it doesn't actually say whether `reduced` has been called (I think). Maybe something like `shouldUseReduced`?

source/reduced.js
Bradcomp: On a second read through, it appears `reduceWhile` depends on `reduced`, but I don't necessarily think we should advertise that fact.

source/reduced.js
```diff
@@ -7,7 +7,7 @@ import _reduced from './internal/_reduced';
  * box: the internal structure is not guaranteed to be stable.
  *
  * Note: this optimization is only available to the below functions:
```
Bradcomp: This note should flat out say that this optimization is only for transducers.

test/reduce.js
GingerPlusPlus: No, unless we export or duplicate it.

source/internal/_reduce.js
GingerPlusPlus: My vote is on `useReduced`.

source/reduced.js
GingerPlusPlus: That'd conflict with #2663.

test/reduce.js
Bradcomp: There should be a way to test for everything using our public API. Otherwise we are in a situation where users can't properly use a function without knowing the internals of our system - in this case _xwrap. 

If a change is solely internal it shouldn't affect the existing tests at all. If it changes the API, we should be able to capture that change in a test that demonstrates the correct behavior as it pertains to the public API change.

GingerPlusPlus:
> we have wrappers for handling dispatching, but `reduce` doesn't use them. This might be due to the necessity of handling stuff like `reduced`

That's not it. Unlike `_reduce`, dispatching is the first thing that wrappers try to do.

> `reduced` is becoming less useful

No it isn't. You can get the old behavior by wrapping the function in a transformer.

Agree with your last 2 paragraphs.

PR #2763, opened 2019-01-12
Changed type checking more strictly

moonformeli:


CrossEye:
Can you explain *why* you want to do this?

PR #2764, opened 2019-01-16
fix `@since` in `includes`

ipanasenko:
`includes` is only available since 0.26.0, need to set correct version in docs


CrossEye:
Thank you!

PR #2767, opened 2019-01-17
Don't dispatch transformers to 'fantasy-land/reduce'

GingerPlusPlus:
`_reduce` no longer tries to dispatch transformers to `fantasy-land/reduce`, because `reduced` support is forbidden by [the spec](https://github.com/fantasyland/fantasy-land#reduce-method):

   >  iii. `f` must return a value of the same type as `x`.
   >  iv.  No parts of `f`'s return value should be checked.

<hr />

Extracted from #2762, discussing it over there could be messy.

This solution avoids unnecessary `_xwrap`. An [alternative solution](https://github.com/ramda/ramda/pull/2762/files#diff-5861836cf1fab2e7793d633617b63889R51) avoids nesting.

### DIFF
--- a/source/internal/_reduce.js
+++ b/source/internal/_reduce.js
@@ -38,13 +38,13 @@ var symIterator = (typeof Symbol !== 'undefined') ? Symbol.iterator : '@@iterato
 
 export default function _reduce(fn, acc, list) {
   if (typeof fn === 'function') {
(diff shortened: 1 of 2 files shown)


CrossEye:
:herb:

I can't see a reason not to do this, and it's certainly a simpler fix than the alternative.  Any other thoughts, @ramda/core ?

PR #2772, opened 2019-01-23
fix: rename then to andThen

### LINKED ISSUE #2751: Addition of 'then' function in 0.26.0 makes Ramda an incomplete thenable
Code to replicate:

```js
const R = require("ramda");
(async function() {
  return await R;
})();
// (node:46376) UnhandledPromiseRejectionWarning:
// TypeError: `then` expected a Promise, received function () { [native code] }
```

We’re running into this issue with Observable, in which we `await` library loads from dynamic `import()` calls, but it’ll also happen for anyone who’s using the `esm` module to do dynamic import() in Node.js now, and for anyone who uses Node.js dynamic import whenever they ship it. Here’s the dynamic import failure case:

```js
(async function() {
  const R = await import("ramda/es/index.js");
})().catch(e => console.error(e));
// TypeError: `then` expected a Promise, received function () { [native code] }
// at _assertPromise (/Users/tmcw/tmp/a/node_modules/ramda/es/internal/_assertPromise.js:6:11)
```

This code works properly in `0.25.0` and breaks in `0.26.0` and beyond, because `await` works like:

1. Await the imported code
2. Await the result of that import

Before, because Ramda was an object, the second await didn't do anything. Now, because it's a 'thenable object' - an object with a then method - `await` tries to await its eventual value and triggers the error.

---

I'm not a maintainer or contributor of this project, so my 2c on how to resolve this is likely to miss something - my only real guess for a solution would be to rename `then` to something else.

tmcw:
BREAKING CHANGE

The addition of R.then in Ramda 0.26 made the R object
a Promise-like object, and meant that Promise.resolve() will attempt
to resolve it by calling the then() method. R is not a Promise, and
to reassure JavaScript of this fact, R.then is renamed to R.andThen.

Fixes #2751

### DIFF
--- a/source/andThen.js
+++ b/source/andThen.js
@@ -24,12 +24,12 @@ import _assertPromise from './internal/_assertPromise';
  *      var getMemberName = R.pipe(
  *        makeQuery,
  *        fetchMember,
- *        R.then(R.pick(['firstName', 'lastName']))
+ *        R.andThen(R.pick(['firstName', 'lastName']))
  *      );
  */
-var then = _curry2(function then(f, p) {
-  _assertPromise('then', p);
+var andThen = _curry2(function andThen(f, p) {
+  _assertPromise('andThen', p);
 
   return p.then(f);
 });
-export default then;
+export default andThen;
--- a/source/index.js
+++ b/source/index.js
@@ -214,7 +214,7 @@ export { default as takeLastWhile } from './takeLastWhile';
 export { default as takeWhile } from './takeWhile';
 export { default as tap } from './tap';
 export { default as test } from './test';
-export { default as then } from './then';
+export { default as andThen } from './andThen';
 export { default as times } from './times';
 export { default as toLower } from './toLower';
 export { default as toPairs } from './toPairs';
--- a/test/andThen.js
+++ b/test/andThen.js
@@ -4,9 +4,9 @@ var R = require('../source');
 var eq = require('./shared/eq');
 
 
-describe('then', function() {
+describe('andThen', function() {
   it('invokes then on the promise with the function passed to it', function(done) {
-    R.then(
+    R.andThen(
       function(n) {
         eq(n, 1);
         done();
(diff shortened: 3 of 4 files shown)


source/andThen.js
```diff
@@ -14,7 +14,7 @@ import _assertPromise from './internal/_assertPromise';
  * @sig (a -> (Promise e b)) -> (Promise e a) -> (Promise e b)
  * @param {Function} onSuccess The function to apply. Can return a value or a promise of a value.
  * @param {Promise} p
- * @return {Promise} The result of calling `p.then(onSuccess)`
+ * @return {Promise} The result of calling `p.andThen(onSuccess)`
```
CrossEye: Oops.  All the other `then`s were correctly changed to `andThen`s.  But this one shouldn't be.

CrossEye:
:herb:

I'm all for this.  I would like to be able to keep the minimal `Promise` support we recently added, and leave them alone thereafter.  This bug was just a distraction... and one more reason to hate Promises!

Bradcomp:
I am :expressionless: on this one. I think it’s good to get it fixed right away though, and this is a good way to do that. Thanks @tmcw!

mbostock:
Was this intended to be released in 0.26.1, or will it be available in a future release?

It looks like it was intended to be part of 0.26.1, but didn’t quite make it. There are two tags in the [releases](https://github.com/ramda/ramda/releases): [0.26.1](https://github.com/ramda/ramda/releases/tag/0.26.1) which includes this change, and [v0.26.1](https://github.com/ramda/ramda/releases/tag/v0.26.1) which does not, and it looks like it was the latter that was published to npm:

https://unpkg.com/ramda@0.26.1/src/then.js

mbostock:
Thanks, @CrossEye!

vukhanhtruong:
@CrossEye You're still not fixing it. Could you please update it? Thanks.

Issue #2775, opened 2019-01-28
`R.set`/`R.over` both lose the prototype of manipulated object


Heimdell:
Some preparation:

```javascript
> let R = require('./node_modules/ramda')
> class X { constructor(a) { this.a = a } }
> x = new X(2)
X { a: 2 }
```

What I expect:
```javascript
> R.set(R.lensPath(["a"]), 4, x)
X { a: 4 }
```

What I get:
```javascript
> R.set(R.lensPath(["a"]), 4, x)
{ a: 4 }
```

P.S.:
I had to do the following to circumvent the issue:
```javascript
let patchRambdaSetter = action => (lens, value, source) =>
    Object.assign(
        Object.create(source.__proto__),
        action(lens, value, source)
    )

let Rset  = patchRambdaSetter(R.set)
let Rover = patchRambdaSetter(R.over)
```

Can you do the prototype setup on the Ramda side, please?
I can't think of current behavior as of intended one. The semantics of `set`/`over` are "modify an object", and not "modify object and set prototype to `{}`".

CrossEye:
I'm afraid I'm :-1: on this.

We've discussed similar ideas before, and generally come to the conclusion that Ramda can't keep trying to support OOP ideas alongside its functional concerns.

One of the most important points about JS is that it's flexible.  It's like the old Perl mantra: *There's more than one way to do it.* . There are far too many ways that people do OOP for us to keep track of.  If we were to concern ourselves with the prototype chain in all its guises, we'd never get anything else done.

Here's [an example][ex], tweaked from its earlier incarnation in a [similar discussion][sd] nearly three years ago:

```js
class Foo {
  constructor(_val) {
    let changeCount = 0;
    let val = _val;
    Object.defineProperties(this, {
      x: {
        get: function() {return val;},
        set: function(newVal) {changeCount++; val = newVal;},
        enumerable: true,
        configurable: false
      },
      changes: {
        get: function() {return changeCount;},
        enumerable: true,
        configurable: false
      }
    });
    
  }
}

let foo = new Foo(42);
console.log(foo)  //~> {x: 42, changes: 0}

foo.x = 43
console.log(foo)  //~> {x: 43, changes: 1}

foo.x = 100
console.log(foo)  //~> {x: 100, changes: 2}

const bar = Rset(lensPath('x'), 50, foo)    // Using enhanced `set`
console.log(bar)  //~> {x: 50, changes: 2}  // Hmm, shouldn't `changes` be 0?

[...]

PR #2776, opened 2019-01-28
Fixes over/set losing prototype (#2775), adds tests

Heimdell:
I made so `assoc` starts not from `{}` as accumulator, but with `Object.create(obj.__proto__)`, when `obj` isn't `null` or `undefined`.

This makes `R.set`/`R.over` not lose the prototype of object being changed.

Fixes issue #2775 .

I added tests for `assoc` and `set`/`over`.

### DIFF
--- a/source/assoc.js
+++ b/source/assoc.js
@@ -22,7 +22,7 @@ import _curry3 from './internal/_curry3';
  *      R.assoc('c', 3, {a: 1, b: 2}); //=> {a: 1, b: 2, c: 3}
  */
 var assoc = _curry3(function assoc(prop, val, obj) {
-  var result = obj === undefined || obj === null ? {} : Object.create(obj.__proto__);
+  var result = obj === undefined || obj === null ? {} : Object.create(Object.getPrototypeOf(obj));
   for (var p in obj) {
     result[p] = obj[p];
   }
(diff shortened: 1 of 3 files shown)


source/assoc.js
```diff
@@ -22,7 +22,7 @@ import _curry3 from './internal/_curry3';
  *      R.assoc('c', 3, {a: 1, b: 2}); //=> {a: 1, b: 2, c: 3}
  */
 var assoc = _curry3(function assoc(prop, val, obj) {
-  var result = {};
+  var result = obj === undefined || obj === null ? {} : Object.create(obj.__proto__);
```
newyankeecodeshop: I think `Object.getPrototypeOf()` is the recommended way to access the prototype, and it has broad browser support.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getPrototypeOf#Browser_compatibility

CrossEye:
Thank you for your contribution.  You can see my response to the idea in #2775.

But note that if we do decide to do this, there is more that needs to be changed.  Obviously `dissoc` as well, but also possibly the array ones (`insert`, `update`, `adjust`, and `remove`) if we want to cover custom extensions to `Array.prototype`.  There are probably other places as well.  But I don't think we want to go this route at all, for reasons described in https://github.com/ramda/ramda/issues/2775#issuecomment-458190103.

PR #2786, opened 2019-02-10
Update README.md / Install specific functions

guysaar223:
Added a link to all Ramda functions as individual functions.

I'm both a maintainer of bit (teambit/bit) and a member of its open source community.

I'm one of many people in this community who's been looking for Ramda functions. So, I've shared Ramda as individual functions, which can now be discovered, played-with and installed via npm/yarn.

Currently, these functions will continue to update automatically on every new tag in this repo. If you prefer, I can hand over ownership and help implement the updates (guysaq@gmail.com).

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,9 @@ It is possible to build Ramda with a subset of the functionality to reduce its f
 
 This requires having Node/io.js installed and ramda's dependencies installed (just use `npm install` before running partial build). 
 
+### Install specific functions
 
+[Install individual functions](https://bitsrc.io/ramda/ramda) with bit, npm and yarn without installing the whole library.
 
 Documentation
 -------------


CrossEye:
I got in touch with the `Bit` team over Gitter a few months ago, and have been meaning to look into it more thoroughly.  In the meantime, this looks all right to me, although we might want to emphasize that this is not the only way to get individual functions -- you can require or  import them directly from the repository, and tree shaking should work reasonably well.

What do others in @ramda/core think?

FredyC:
Hey, we just recently discovered that tree-shaking doesn't work at all if we basically reexport ramda from another package. I am not sure if it's an issue of `create-react-app` configuration, but we started looking for other solutions.

The Bit seems like an interesting way to go, but if I am not mistaken, TypeScript will be an issue here since types are for `ramda` as a whole module. It would be necessary for each individual function in Bit to export its own types.

CrossEye:
@FredyC: The Ramda team has not been involved in the Typescript typings.  Perhaps one day one of us will be curious enough to investigate, but for now we leave that to those who know what they are doing.

If you understand those typings, we're always looking for volunteers!

FredyC:
@CrossEye Well, this isn't really about understanding, it's a simple fact that these Bit exports cannot work with it because it's not part of the package. I am not aware of "easy" way how to tackle this, unfortunately, but I will try to ask in Bit community.

CrossEye:
@FredyC: I know little enough about Typescript (even though I sometimes have to use it) that I don't know what it would mean for it to be "part of the package".  I know TS is quite popular, but I've never felt the need.

FredyC:
As I have shown in the case of material-ui, there are an `index.d.ts` with each exported component. I believe the Ramda would have to do the same. I am not sure if there is some better way, I like to use TS, but never got into figuring how it works.

TS is a great win with Ramda, because it can guard for you eg. `R.prop("missing")` will cause a TS error if that prop actually do not exist on the object. It's a great way to prevent typos and you don't have to rely on thorough tests and/or manual testing.

PR #2787, opened 2019-02-16
assoc/dissoc support operate array

bolasblack:
Related https://github.com/ramda/ramda/issues/773#issuecomment-462097072

I noticed `dissocPath` has code:

https://github.com/ramda/ramda/blob/efd899ba81cc5290330f5ca0833746a7ecbe165c/source/dissocPath.js#L38-L39

So

```js
;(a => R.dissocPath([3], a) === a)([0, 1])
;(a => R.dissocPath(['b'], a) === a)({a: 1})
// and
;(a => R.dissoc(3, a) !== a)([0, 1])
;(a => R.dissoc('b', a) !== a)({a: 1})
```

Should their behavior be consistent in this scenario?

Another question, the previous version `dissocPath` use `R.update` to update array input:

https://github.com/ramda/ramda/blob/efd899ba81cc5290330f5ca0833746a7ecbe165c/source/dissocPath.js#L38-L44

In my version, I use 'R.assoc` uniformly:

https://github.com/ramda/ramda/blob/528df9777200a31c754117cdfd2e64871985f4d3/source/dissocPath.js#L33-L37

The behavior of `R.update` is different from `R.assoc` in some scenarios:

```js
R.update(2, 'a', []) // => []
R.assoc(2, 'a', []) // => [undefined, undefined, 'a']
```

Should I keep current version or rollback?

### DIFF
--- a/source/assoc.js
+++ b/source/assoc.js
@@ -1,5 +1,5 @@
 import _curry3 from './internal/_curry3.js';
-
+import assocPath from './assocPath.js';
 
 /**
  * Makes a shallow clone of an object, setting or overriding the specified
@@ -11,8 +11,9 @@ import _curry3 from './internal/_curry3.js';
  * @memberOf R
  * @since v0.8.0
  * @category Object
- * @sig String -> a -> {k: v} -> {k: v}
- * @param {String} prop The property name to set
+ * @typedefn Idx = String | Int
+ * @sig Idx -> a -> {k: v} -> {k: v}
+ * @param {String|Number} prop The property name to set
  * @param {*} val The new value
  * @param {Object} obj The object to clone
  * @return {Object} A new object equivalent to the original except for the changed property.
@@ -21,12 +22,5 @@ import _curry3 from './internal/_curry3.js';
  *
  *      R.assoc('c', 3, {a: 1, b: 2}); //=> {a: 1, b: 2, c: 3}
  */
-var assoc = _curry3(function assoc(prop, val, obj) {
-  var result = {};
-  for (var p in obj) {
-    result[p] = obj[p];
-  }
-  result[prop] = val;
-  return result;
-});
+var assoc = _curry3(function assoc(prop, val, obj) { return assocPath([prop], val, obj); });
 export default assoc;
--- a/source/dissocPath.js
+++ b/source/dissocPath.js
@@ -1,11 +1,30 @@
 import _curry2 from './internal/_curry2.js';
+import _dissoc from './internal/_dissoc.js';
 import _isInteger from './internal/_isInteger.js';
 import _isArray from './internal/_isArray.js';
 import assoc from './assoc.js';
-import dissoc from './dissoc.js';
-import remove from './remove.js';
-import update from './update.js';
 
+/**
+ * Makes a shallow clone of an object. Note that this copies and flattens
+ * prototype properties onto the new object as well. All non-primitive
+ * properties are copied by reference.
+ *
+ * @private
+ * @param {String|Integer} prop The prop operating
+ * @param {Object|Array} obj The object to clone
+ * @return {Object|Array} A new object equivalent to the original.
+ */
+function _shallowCloneObject(prop, obj) {
+  if (_isInteger(prop) && _isArray(obj)) {
+    return [].concat(obj);
+  }
+
+  var result = {};
+  for (var p in obj) {
+    result[p] = obj[p];
+  }
+  return result;
+}
 
 /**
  * Makes a shallow clone of an object, omitting the property at the given path.
@@ -27,18 +46,20 @@ import update from './update.js';
  *      R.dissocPath(['a', 'b', 'c'], {a: {b: {c: 42}}}); //=> {a: {b: {}}}
  */
 var dissocPath = _curry2(function dissocPath(path, obj) {
+  if (obj == null) {
+    return obj;
+  }
+
   switch (path.length) {
     case 0:
       return obj;
     case 1:
-      return _isInteger(path[0]) && _isArray(obj) ? remove(path[0], 1, obj) : dissoc(path[0], obj);
+      return _dissoc(path[0], obj);
     default:
       var head = path[0];
       var tail = Array.prototype.slice.call(path, 1);
       if (obj[head] == null) {
-        return obj;
-      } else if (_isInteger(head) && _isArray(obj)) {
-        return update(head, dissocPath(tail, obj[head]), obj);
+        return _shallowCloneObject(head, obj);
       } else {
         return assoc(head, dissocPath(tail, obj[head]), obj);
       }
--- a/source/internal/_dissoc.js
+++ b/source/internal/_dissoc.js
@@ -1,0 +1,28 @@
+import _isInteger from './_isInteger';
+import _isArray from './_isArray';
+import remove from '../remove';
+
+/**
+ * Returns a new object that does not contain a `prop` property.
+ *
+ * @private
+ * @param {String|Number} prop The name of the property to dissociate
+ * @param {Object|Array} obj The object to clone
+ * @return {Object} A new object equivalent to the original but without the specified property
+ */
+export default function _dissoc(prop, obj) {
+  if (obj == null) {
+    return obj;
+  }
+
+  if (_isInteger(prop) && _isArray(obj)) {
+    return remove(prop, 1, obj);
+  }
+
+  var result = {};
+  for (var p in obj) {
+    result[p] = obj[p];
+  }
+  delete result[prop];
+  return result;
+}
--- a/source/assocPath.js
+++ b/source/assocPath.js
@@ -1,10 +1,8 @@
 import _curry3 from './internal/_curry3.js';
 import _has from './internal/_has.js';
-import _isArray from './internal/_isArray.js';
 import _isInteger from './internal/_isInteger.js';
-import assoc from './assoc.js';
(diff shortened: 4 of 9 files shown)


source/assoc.js
```diff
@@ -22,6 +23,12 @@ import _curry3 from './internal/_curry3';
  *      R.assoc('c', 3, {a: 1, b: 2}); //=> {a: 1, b: 2, c: 3}
  */
 var assoc = _curry3(function assoc(prop, val, obj) {
+  if (_isInteger(prop) && _isArray(obj)) {
+    var arr = [].concat(obj);
```
CrossEye: I know that this was moved from elsewhere in the code, but still, I can see no reason for `[].concat(obj)` over `[obj]`.

CrossEye:
> Should their behavior be consistent in this scenario?

I think they should compare correctly with `equals`.  I'm not worried about reference equality.

> In my version, I use 'R.assoc` uniformly:

I think that's fine.  `dissocPath` was written when `assoc` did not handle arrays and their integer keys, only plain objects with their string keys.  Now that it does both, this is a useful simplification.

> The behavior of `R.update` is different from `R.assoc` in some scenarios:

Again, I'm not worried.  Perhaps now that `assoc` does handle arrays, it's time to get rid of `update`.  But the fact that they don't act the same is not an issue.

source/assoc.js
bolasblack: > I know that this was moved from elsewhere in the code, but still, I can see no reason for `[].concat(obj)` over `[obj]`.

I‘m not sure if I understand your meaning correctly. 

```js
var obj = [1, 2, 3]

// `[].concat(obj)` is cloning a array shallowly, 
;[].concat(obj) // => [1, 2, 3]

// [obj] can not
;[obj] // => [1, 2, 3]
```

bolasblack:
> As we've done elsewhere with `foo` and `fooPath`, I would like to have `foo` be a simple call to `fooPath`.

Do you mean that I can add a file like `internal/_assoc.js`, then `assocPath.js` depend on `_assoc.js`, and `assoc.js` implemented as `assocPath([idx], val, obj)`?

> I think they should compare correctly with `equals`. I'm not worried about reference equality.

As far as my selfishness is concerned, I hope they can be strict equal, because deep comparison requires more resources

source/assoc.js
CrossEye: Never mind.  Brain fart.

CrossEye:
> Do you mean that I can add a file like `internal/_assoc.js`, then `assocPath.js` depend on `_assoc.js`, and `assoc.js` implemented as `assocPath([prop], val, obj)`?

Yes; I had been thinking of a more disruptive rewrite.  That would be a simpler solution.  And if we do something like this also with `dissoc[Path]`, it should address the question of those behavior differences, too.

bolasblack:
Hi @CrossEye, I have completed the changes

> I think they should compare correctly with `equals`. I'm not worried about reference equality.

I added the code to check if the attribute exists 😁

https://github.com/ramda/ramda/blob/f424de89dbd1afe031b123cc0b71042ac8a64cb1/source/internal/_dissoc.js#L13-L18

source/internal/_dissoc.js
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @private
+ * @param {String|Number} prop The name of the property to dissociate
+ * @param {Object|Array} obj The object to clone
+ * @return {Object} A new object equivalent to the original but without the specified property
+ */
+export default function _dissoc(prop, obj) {
+  if (obj[prop] == null) {
+    return obj;
+  }
```
CrossEye: That's going to be a problem:

```js
_dissoc('a', {a: null, b: 2}) //=> {a: null, b: 2}
```

I think you would need to switch to 

```js
  if (!(prop in obj)) {
```

I'm not sure I see much of a good reason for a special case, though, for when the property doesn't exist.  Although we stress that value equality is all we care about, we get complaints any time we sometimes return an object unchanged and sometimes return a new object.  Users who expect the latter behavior get frustrated by the former, thinking that Ramda handles it for them.  That consistency is perhaps a good thing.

bolasblack: I changed the code in

* `dissocPath.js` https://github.com/ramda/ramda/pull/2787/files#diff-229ca62287d81f3b117695acff8e6d5fR33
* `_dissoc.js` https://github.com/ramda/ramda/pull/2787/files#diff-2ffae1703b1b496b4bef5eb8428bed6bR14

bolasblack: > I'm not sure I see much of a good reason for a special case, though, for when the property doesn't exist. Although we stress that value equality is all we care about, we get complaints any time we sometimes return an object unchanged and sometimes return a new object. Users who expect the latter behavior get frustrated by the former, thinking that Ramda handles it for them. That consistency is perhaps a good thing.

In fact, I am not sure which way is better. At present, my understanding is that if the input will not be modified, then the original object should be returned. If it is modified, then the new object should be returned, just like immutable.js or ClojureScript.

CrossEye: The difference with Immutable/ClojureScript is that the Ramda results are not actually immutable.  The problems people have had is that they work for a while with

```js
const newVal = R.foo(bar, oldVal);
```

and things are fine, as `foo` is returning something new.  Then something happens and there is no change, but

```js
alter(newVal)
```

then also alters `oldVal`.

While many Ramda users will never have issues with this, especially if most modifications use Ramda functions, people have had this problem.  I simply don't see any good reason to continue with it while we're changing this function.

bolasblack: Yes, that's why I say I'm not sure which way is better. Because my point of view is based on my understanding and habits, may not suitable for other Ramda users. So I listen to you, I believe that at least this will ensure that these two functions are designed to be consistent with Ramda's overall design philosophy.

So you hope assoc/dissoc can always return a new object, no matter it will or will not be modified right?

CrossEye: Yes, exactly.

source/dissocPath.js
```diff
@@ -27,18 +46,20 @@ import update from './update';
...
-      return _isInteger(path[0]) && _isArray(obj) ? remove(path[0], 1, obj) : dissoc(path[0], obj);
+      return _dissoc(path[0], obj);
     default:
       var head = path[0];
       var tail = Array.prototype.slice.call(path, 1);
       if (obj[head] == null) {
-        return obj;
-      } else if (_isInteger(head) && _isArray(obj)) {
-        return update(head, dissocPath(tail, obj[head]), obj);
+        return _shallowCloneObject(head, obj);
```
bolasblack: I cloned `obj` when path can not deep in

But I can not clone `obj` when path is empty, because I have no way to judge how to clone it (Ramda clone `obj` as an array is according to `_isInteger(prop) && _isArray(obj)`, but I can not get `prop` in that clause)

source/internal/_dissoc.js
bolasblack: Hi @CrossEye, how about now 😆

bolasblack:
Hi @CrossEye, how do you feel about this PR now?

CrossEye:
@bolasblack:

Sorry I missed the earlier ping.  I'm wondering if it's my fuzzy brain or if this is not yet addressed: https://github.com/ramda/ramda/pull/2787#discussion_r257751043.

Other than that, it looks good to me.

bolasblack:
Hi @CrossEye, I think maybe I have resolved the problem: https://github.com/ramda/ramda/pull/2787/files#diff-6787d2bc79f3b7d8d865c601133bd002R10

bolasblack:
Hi @CrossEye, I really need your reply 🤣

CrossEye:
Sorry, I have taken an unexpected break from Ramda.  I should be back to it soon.  I'll look at this in a few days.

bolasblack:
@CrossEye 👌

bolasblack:
Hi @CrossEye, are you interested in seeing this PR again?

bolasblack:
This PR needs some love 😏

CrossEye:
@bolasblack: yes, that happens far too often around here.  I'll try to look soon.

bolasblack:
Hi CrossEye, sorry for my last comment, I'm not trying to blame you, I am very grateful and respect for your work on ramda, it's really useful and helped me a lot, I just want to make sure this PR is still remembered

CrossEye:
@bolasblack: Not at all.  I seem by default to be in charge of this project, but I never intended to sign up for that responsibility and it gets whatever time I can spare from work, family, relaxation, home improvement, the horse farm I co-own, and numerous other programming projects that occupy my hours.  So things often fall by the wayside.  It's good when someone reminds me of them!  In fact, if things are to happen around here, it might be imperative.

CrossEye:
Thank you for persevering @bolasblack!

bolasblack:
> @bolasblack: Not at all. I seem by default to be in charge of this project, but I never intended to sign up for that responsibility and it gets whatever time I can spare from work, family, relaxation, home improvement, the horse farm I co-own, and numerous other programming projects that occupy my hours. So things often fall by the wayside. It's good when someone reminds me of them! In fact, if things are to happen around here, it might be imperative.

@CrossEye Haha, I like your attitude towards the project (although I feel a little anxious when I am the creator of PR 🤪)

bolasblack:
> :bowtie:
> This looks like an improvement, thanks!



> :bowtie:
> This looks like an improvement, thanks!

@Bradcomp Thank you very much also 😆

PR #2793, opened 2019-03-02
docs: Add @since where it is missing

WaldoJeffers:
Hello,

Really not a big deal but some `@since` are missing for some functions introduced in `v0.26.0`, which makes it hard to know when they were introduced in the docs. This should fix it :)

See the screenshot below:
image

### DIFF
--- a/source/composeWith.js
+++ b/source/composeWith.js
@@ -12,6 +12,7 @@ import reverse from './reverse';
  *
  * @func
  * @memberOf R
+ * @since v0.26.0
  * @category Function
  * @sig ((* -> *), [(y -> z), (x -> y), ..., (o -> p), ((a, b, ..., n) -> o)]) -> ((a, b, ..., n) -> z)
  * @param {...Function} ...functions The functions to compose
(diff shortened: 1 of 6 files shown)


CrossEye:
Hmm, that should have been added by the build script.  I'll have to investigate why it wasn't, but for the moment, this LGTM.  :herb:

Anyone else?

WaldoJeffers:
@CrossEye Yes, I suspected it wasn't added manually but was clueless about where to look :/ In the meantime, I think it will avoid simple mistakes like this one: in my team, we have a lot of projects using ramda (😍), some of them use v0.25 & others v0.26.1. I was writing some new code, and needed to merge things, I remembered something about `merge` being deprecated, so I went to the docs and saw `mergeRight/Left`. Because I didn't the "Added in vXXX" mention, I did not think twice about it and used these new functions. Turns out I was working on one the projects using Ramda v0.25 😆

Bradcomp:
:bowtie:  LGTM!

Issue #2784, opened 2019-02-07
ApplySpec not properly handling properties with arrays of anonymous objects


ryanbas21:
```
applySpec({
  test: [{ a : inc }]
})(2);

// {"test": {"0": {"a": 3}}} test should have a value of an array, not of an object.
```
https://ramdajs.com/repl/#?applySpec%28%7B%0A%20%20test%3A%20%5B%7B%20a%20%3A%20inc%20%7D%5D%0A%7D%29%282%29%3B%0A

What is happening is that when applySpec recursively goes into the array, it changes the array to an object with a key "0" and then applies the function, where previously it handled the array appropriately. 

I am seeing this issue in the new `0.26.1`/ `0.26` and not 0.25.

CrossEye:
Hmm, I wonder what caused this regression.  I don't recall any recent work on `applySpec`.  I'll have to check this weekend.

ryanbas21:
@CrossEye The only thing we noticed was a PR merged about the `map` key with applySpec, although I could not tell how that would affect this regression

ryanbas21:
Actually, that function that was added is the issue. ITs taking the keys and not retransforming it back to an array

CrossEye:
#2681 #2682

bensbigolbeard:
I've submitted a potential fix for this issue (☝️)

@CrossEye Let me know if there are any issues with this approach.

jordie23:
I posted this in another issue before I noticed this, so reposting here because I think it applies, however my issue was the root object being an array not necessarily a nested property.

Original comment:

------------------------

Hello,

So turns out, previous to this, `applySpec` worked seemlessly with arrays/lists, not just objects. Since this change, it has broken that. Just wondering whether `applySpec` should be supporting arrays/lists or whether that was accidental previously? This has broken some of our code - not a problem, we can update it - but just wondering whether we were relying on a quirk in the code or something that was actually supposed to be supported?

e.g.

```const kvs = [
  {
    Key: 'Tag1',
    Value: 'TestVal1'
  },
  {
    Key: 'Tag2',
    Value: 'TestVal2'
  },
  {
    Key: 'Tag3',
    Value: 'TestVal3'
  }
]

applySpec([map(prop('Key')), map(prop('Value'))])(kvs)

// result < 0.26

[ 
  [ 'Tag1', 'Tag2', 'Tag3', 'Tag4' ],
[...]

PR #2795, opened 2019-03-08
fix(applySpec): reintroduce support for arrays of nested specs and functions

bensbigolbeard:
Addresses #2784

Prior to 0.26.x, this was a valid structure, where an array could be passed at an arbitrary depth and could contain other values/functions to be traversed:
```javascript
// Ex. 1
R.applySpec([ R.sum ])(1,2) // [ 3 ]

// Ex. 2
R.applySpec({ nested: [{ sum: R.sum }])(1,2) // { nested: [{ sum: 3 }]}
```
After the PR to address #2681, the output changed to:
```javascript
// Ex. 1
{ "0": [ 3 ]}

// Ex. 2
{ nested: { "0": { sum: 3 }}}
```

These changes add a condition to `applySpec`'s private function, `mapValues`, to just add a check if the current `obj` is an array before branching to `.map()` or `.keys().reduce()` operations based on the result.

### DIFF
--- a/source/applySpec.js
+++ b/source/applySpec.js
@@ -1,4 +1,5 @@
 import _curry1 from './internal/_curry1';
+import _isArray from './internal/_isArray';
 import apply from './apply';
 import curryN from './curryN';
 import max from './max';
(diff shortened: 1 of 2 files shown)


CrossEye:
:herb:

Thank you.  This looks good to me.  Anyone else, @ramda/core?

PR #2802, opened 2019-03-20
Add browser tests and lint job to CI

wojpawlik:


CrossEye:
This looks good to me.  Other thoughts, @ramda/core?

davidchambers:
> Add browser tests and lint job to CI

It looks as though CI already does linting.

GingerPlusPlus:
Yes, but has no separate tasks; if lint fails, tests don't run, and you need to read the output to distinguish failed lint from failed test.

PR #2805, opened 2019-04-02
fixed description of and function

kace91:
Hi there,

I noticed a typo in the 'and' function's description while reading the docs, so here's a quick fix!

Cheers.


CrossEye:
Actually, that's not a typo.  The type of `and(a, b)` is the type of `a` if `a` is truth-y OR the type of `b` if it is not.

PR #2806, opened 2019-04-03
add an example which covers error and value

sunnystatue:
Update the tryCatch examples to include an item which shows the way to catch the error message.


CrossEye:
Thank you!

PR #2810, opened 2019-04-24
fixes jsdoc annotations for internal _Set class

pdeona:
These changes attempt to fix the jsdoc type annotations associated with the internal `_Set` class and its associated methods.

Screen Shot 2019-04-24 at 4 14 54 PM


CrossEye:
Very sorry @pdeona.  A number issues from this period were entirely ignored.  We're back and trying to get caught up.  I'm curious about this one.  I don't even know why that object has JSDoc tags, as it's not part of the public API.  Do they get exposed through some JSDoc magic I don't know about?

PR #2813, opened 2019-04-28
[feat]: scope omit behavior & new inter fn.

Tommy-White:
In my opinion, there is a problem with the existing `omit` method:
1. The `names` parameter is expected to be an array but there is no type checking inside.
2. When only one attribute needs to be omitted, the `names` parameter must be written in the array or an unexpected usage will occur.

So make it more safety, I made some change below:
1. Add a new internal function `_ArrayWrapper` that use to check whether an object is an array and wrapped, it returns an array.
2. First, preprocess the `names` parameter of the `omit` method as: `names =  _ArrayWrapper(names)` to compatible with the original usage, and can pass a single `String` property directly
3. Add some new unit test to make sure the method behavior is supposed to.

### DIFF
--- a/source/internal/_arrayWrapper.js
+++ b/source/internal/_arrayWrapper.js
@@ -1,4 +1,3 @@
-import _curry1 from './_curry1';
 import _isArray from './_isArray';
 
 /**
@@ -22,5 +21,5 @@ import _isArray from './_isArray';
 //           Object.prototype.toString.call(val) === '[object Array]');
 // };
 export default function _arrayWrapper(val) {
-  return _isArray(val) ? val : [val]
-};
+  return _isArray(val) ? val : [val];
+}
--- a/source/internal/_arrayWrapper.js
+++ b/source/internal/_arrayWrapper.js
@@ -15,11 +15,6 @@ import _isArray from './_isArray';
  *
  */
 
-// export default Array.isArray || function _isArray(val) {
-//   return (val != null &&
-//           val.length >= 0 &&
-//           Object.prototype.toString.call(val) === '[object Array]');
-// };
 export default function _arrayWrapper(val) {
   return _isArray(val) ? val : [val];
 }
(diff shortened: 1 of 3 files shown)


CrossEye:
Thank you for the contribution, but I'm afraid that I'm :-1: on this.

> 1. The names parameter is expected to be an array but there is no type checking inside.

Ramda has always preferred to be a no hand-holding library.  We do little type-checking for the user, assuming she knows what she's doing.  

> 2. When only one attribute needs to be omitted, the names parameter must be written in the array or an unexpected usage will occur.

Ramda strives for simplicity, especially in its public API.  If you haven't seen the classic Rich Hickey talk, *[Simple Made Easy][rh]*, I would highly recommend it.  He describes *complex* as intertwined, and *simple* as the opposite.  Right now, `omit` has the signature `omit :: [String] -> Object -> Object`.  What you're proposing is that it also work as `omit :: String -> Object -> Object`.  One function doing two things means that those concerns are intertwined, and thus the function would be more complex than we prefer.

That, for instance, `map` works on Arrays, Objects, and Functions as well as delegating to the `map` methods on types that have one does not contradict this goal of simplicity.  `map` works on the more abstract type of [`Functor`][fu] -- Ramda merely supplies implementations for Arrays, Objects, and Functions.  Note that it doesn't work on Strings because Strings are not Functors.  (An accident of implementation means that it does in fact return a slightly reasonable value, but it doesn't return a String, as `map` should; this again speaks to our no-hand-holding philosophy: *garbage in, garbage out*.)  But there is no obvious type that encompasses both Strings and Arrays of Strings.


  [fu]: https://github.com/fantasyland/fantasy-land#functor

Bradcomp:
> When only one attribute needs to be omitted, the names parameter must be written in the array or an unexpected usage will occur.

There is another function in the library, `R.dissoc`, for this purpose. While it might seem odd to have two functions that are so similar, there are a couple reasons for it.

First is the simplicity @CrossEye mentioned above.

Secondly it allows for some nice symmetry between sets of functions that allows for writing more semantically correct code, and helps discoverability.

`dissoc` has the opposite function `assoc`, for adding and removing an entry from an object.
`omit` has the opposite function `pick` for whitelisting and black listing object properties.

PR #2817, opened 2019-05-05
enabling R.scan to be used as transducer.

### LINKED ISSUE #2816: enabling R.scan to be used as transducer.
Hi!

I noticed that `R.scan` doesn't work as transducer and I think could be useful for performing reducer operations lazily.

xgbuils:
It fixes #2816


CrossEye:
:herb:

I'm happy with this (as much as I am with anything extending our transducers.)

Anyone else?  @ramda/core?

xgbuils:
Cool, @CrossEye ! 

After self-reviewing my PR I'm going to remove `['scan']` in dispatchable because I think it's out of the scope of this PR.

CrossEye:
As with many things around this time, this PR was inadvertently dropped.  This makes sense, and I wonder, @xgbuils, if you want to resolve the conflicts so that we can merge?

source/scan.js
```diff
@@ -22,7 +24,7 @@ import _curry3 from './internal/_curry3.js';
  *      const factorials = R.scan(R.multiply, 1, numbers); //=> [1, 1, 2, 6, 24]
  * @symb R.scan(f, a, [b, c]) = [a, f(a, b), f(f(a, b), c)]
  */
-var scan = _curry3(function scan(fn, acc, list) {
+var scan = _curry3(_dispatchable([], _xscan, function scan(fn, acc, list) {
```
customcommander, line 27: If I understand this correctly then `R.scan` can act as a transducer when given a transformer as a list. Perhaps we should mention that in the description?

customcommander (changes requested):

source/scan.js
xgbuils: Thanks for the observation, @customcommander. I've added that in the description.

customcommander (dismissed):
Approving on the basis that this pull request has been approved before but since there's been some conflicts I think it would make sense for @CrossEye to have a quick re-review.

Thank you @xgbuils and sorry for the delay ;)

xgbuils:
No problem. My pleasure. I have to review documentation about transducers again (I had to review my PR accurately to understand what I did 2 years ago). I think that there are another functions that works with lists that can be adapted to work with transducers and I could review it in the future.

CrossEye (dismissed):

customcommander (dismissed):

CrossEye:
Hmm, just coming back to this to see there are conflicts.  Going to bed now, but I hope to look at these soon.

PR #2818, opened 2019-05-11
Add isNotNil

LarsKoelpin:
I feel like writing complement(isNil) a lot to do null/undefned checking.

I haven't found an appropriate alternative, so I think it might be a good addition to the library to add
a function named 

isDefined: Object -> Bool 

which does the null/undefined checking.

What do you think of this addition?

### DIFF
--- a/source/isNotNil.js
+++ b/source/isNotNil.js
@@ -1,0 +1,22 @@
+import isNil from './isNil';
+import _curry1 from './internal/_curry1';
+
+
+/**
+ * Checks if the input value is not `null` and not `undefined`.
+ *
+ * @func
+ * @memberOf R
+ * @category Type
+ * @sig * -> Boolean
+ * @param {*} x The value to test.
+ * @return {Boolean} `true` if `x` is not `undefined` or not `null`, otherwise `false`.
+ * @example
+ *
+ *      R.isNotNil(null); //=> false
+ *      R.isNotNil(undefined); //=> false
+ *      R.isNotNil(0); //=> true
+ *      R.isNotNil([]); //=> true
+ */
+var isNotNil = _curry1(function isNotNil(x) { return !isNil(x); });
+export default isNotNil;
--- a/source/isNotNil.js
+++ b/source/isNotNil.js
@@ -1,5 +1,5 @@
-import isNil from './isNil';
-import _curry1 from './internal/_curry1';
+import isNil from './isNil.js';
+import _curry1 from './internal/_curry1.js';
 
 
 /**
--- a/source/index.js
+++ b/source/index.js
@@ -99,6 +99,7 @@ export { default as invoker } from './invoker.js';
 export { default as is } from './is.js';
 export { default as isEmpty } from './isEmpty.js';
 export { default as isNil } from './isNil.js';
+export { default as isNotNil } from './isNotNil.js';
 export { default as join } from './join.js';
 export { default as juxt } from './juxt.js';
 export { default as keys } from './keys.js';
--- a/test/isNotNil.js
+++ b/test/isNotNil.js
@@ -1,0 +1,15 @@
+var R = require('../source');
+var eq = require('./shared/eq');
+
+describe('isNotNil', function() {
+  it('tests a value for `null` or `undefined`', function() {
+    eq(R.isNotNil(void 0), false);
+    eq(R.isNotNil(undefined), false);
+    eq(R.isNotNil(null), false);
+    eq(R.isNotNil([]), true);
+    eq(R.isNotNil({}), true);
+    eq(R.isNotNil(0), true);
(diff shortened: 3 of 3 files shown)


CrossEye:
This sounds like a good idea.  We probably should have done this long ago.  I'm not certain of the name.  `isNotNil` seems more likely to me.

source/isDefined.js
```diff
@@ -0,0 +1,23 @@
...
+ * @param {*} x The value to test.
+ * @return {Boolean} `true` if `x` is not `undefined` or not `null`, otherwise `false`.
+ * @example
+ *
+ *      R.isDefined(null); //=> false
+ *      R.isDefined(undefined); //=> false
+ *      R.isDefined(0); //=> true
+ *      R.isDefined([]); //=> true
+ */
+var isDefined = complement(isNil);
```
CrossEye: I see no reason to depend on `complement` here.   Just do it manually.

bas080:
> This sounds like a good idea. We probably should have done this long ago. I'm not certain of the name. `isNotNil` seems more likely to me.

```js
if notNil(person) { /* ... */ };
if isNotNil(person) { /* ... */ };
if defined(person) { /* ... */ };
if isDefined(person) { /* ... */ };
if exists(person) { /* ... */ };
```

I don't minds being a bit more expressive instead of using `not`.

LarsKoelpin:
In function names, I like to prevent "not", because in my optnion it is an unnecessary mental overhead and in the past often created unnecessary confusions for me.

That's why I try to name things in a "positive" way.
I also like "exists". Shall be "isNotNil" the still the final name?

I will be adding the test and incorporating your changes soon when I'm back home again.

Greetings

LarsKoelpin:
* Testcase Added: Please review if the specification matches your thoughts
* Removed dist/ commit from history
* Removed dependecy on comeplemt

source/isDefined.js
LarsKoelpin: Removed in https://github.com/ramda/ramda/pull/2818/commits/0789ed369b2e0c256a76adeda38bbdc31d956919

CrossEye:
I don't share your objections to using 'not' in the name; `nil` to me means simply having one of the two nil types in JS, `null` or `undefined`.  But that was just a quick suggestion; `isNotNil` is not a name I care about.  But `isDefined` does not resonate for me.  I didn't guess what it did from the name.  When I saw the title, I thought that maybe you were simply looking for [`R.has`][ha].

So does anyone have a another candidate for a name for this function?  @ramda/core?



  [ha]: https://ramdajs.com/docs/#has

LarsKoelpin:
I changed the function name to isNotNil. 
If everything is fine for you l'd squash the WIP commits.

Grohden:
Why not use ramda adjunct?
https://char0n.github.io/ramda-adjunct/2.18.0/RA.html#.isNotNil

I know that it adds a lot o functions.. but it has most of the functions that i find myself repeating across projects using ramda

Btw: I thought that ramda was meant to have only "base" functional tools, that's why i think that we still don't have the "isNot*" functions. If I'm wrong, would be nice to have `isNotEmpty` in the lib too.

tommmyy:
And also there is a https://ramda-extension.firebaseapp.com/docs/#isNotNil

ben-eb:
I'm not sure about this one, I did use to use something similar when first starting out with Ramda but then found it to be unnecessary:

```js
const guardFn = when(isDefined);
// same as:
const guardFn = unless(isNil);
```

And:

```
const byDefined = filter(isDefined);
// same as:
const byDefined = reject(isNil);
```

I'd say 99% of the time I would want to do a null check, I'd use one of these two patterns.

source/isNotNil.js
```diff
@@ -0,0 +1,22 @@
...
+ * @param {*} x The value to test.
+ * @return {Boolean} `true` if `x` is not `undefined` or not `null`, otherwise `false`.
+ * @example
+ *
+ *      R.isNotNil(null); //=> false
+ *      R.isNotNil(undefined); //=> false
+ *      R.isNotNil(0); //=> true
+ *      R.isNotNil([]); //=> true
+ */
+var isNotNil = _curry1(function isNotNil(x) { return !isNil(x); });
```
floydspace: `_curry1(compose(not, isNil))`

customcommander:
@LarsKoelpin @CrossEye This pull request could not be merged because of a merge conflict with `source/index.js` and linter issues as it was predating the decision to add `.js` extensions to our import statements. That's the only thing I changed in this pull request, everything else is exactly as it was when you approved it @CrossEye.

If we still want this, this should now be ready to go.

customcommander (approved):

CrossEye (approved):

Issue #2819, opened 2019-05-12
invoker.js documentation


salis010:
A suggestion: maybe the documentation should make it clear that only the target object's methods can be invoked, and not any function.

Replace this: @param {String} method Name of the method to call.

With this: @param {String} method Name of any of the target object's methods to call.

CrossEye:
If you want to create a PR for this, I think we'd accept it.  If you found this confusing, probably others did too.  For me, I don't see a reason for confusion; the very word "method" already implies all that.  But I don't object.

PR #2821, opened 2019-05-13
updated invoker.js documentation

salis010:
#2819 

A suggestion: maybe the documentation should make it clear that only the target object's methods can be invoked, and not any function.

Replace this: @param {String} method Name of the method to call.

With this: @param {String} method Name of any of the target object's methods to call.


CrossEye:
Thank you!

:herb:

Issue #2771, opened 2019-01-23
hasPath does not work if one attribute in the path is undefined 


sespinoza-dev:
## hasPath does not work if one attribute in the path is undefined 

If one of the object attributes in the path is undefined, hasPath throws an error. it should return false.

```console
R.hasPath(['a', 'b'], { a: undefined });
=> TypeError: Cannot convert undefined or null to object
// expected false
```

CrossEye:
This is definitely a bug.  Are you interested in creating a PR?

sespinozj:
I'll give it a try.

Greettings!

machad0:
Hi @sespinozj, are you still into this issue? If you don't, I'd like to try

cheers

sespinozj:
@machad0 do please go ahead. :)

machad0:
Hi again,

Well, actually this bug was fixed at  #2738, it's just an outdated bundle.

```javascript
var hasPath = _curry2(function hasPath(_path, obj) {
...
  while (idx < _path.length) {
    if (_has(_path[idx], val)) {
      val = val[_path[idx]];
      idx += 1;
    } else {
      return false;
    }
 
```
the method doesn't contains the updated `hasPath` version, should we bump it?

sespinozj:
yes, thanks for checking that out @machad0 

We use `pathOr` as a workaround in some cases thou.

Greetings!

CrossEye:
Damn, I'm so caught up in what I want to do soon that I've forgotten what we've done recently!

gouegd:
Just got bit by this bug, I don't think the fix is in the latest version (or it has not worked):
`R.hasPath(['a', 'b'], { a: undefined })`
and
`R.hasPath(['a', 'b'], { a: { b: undefined } })`
both crash

gouegd:
@mattgrande That means the fix for #1 is on master but not published, doesn't it ?

mattgrande:
That's what it seems like, yes.

CrossEye:
I've been away from Ramda for a few weeks, and am trying to get back.  I will try to publish again soon, and this change should then be included.

gahabeen:
Seems like there is still an issue. If the sub value isn't an Object (Object or Array), it fails.

I ended up with this exact error message:
```
C:\[PATH]\ramda\src\internal\_has.js:2
  return Object.prototype.hasOwnProperty.call(obj, prop);
                                         ^
TypeError: Cannot convert undefined or null to object
[...]

PR #2825, opened 2019-05-14
hasPath return false for non-object checks

LorneCurrie:
Used to throw an Error for non-object children tests
Added test for object checking if val is an object when testing a path
Added tests testing false is returned for non-objects children tests
Related to GitHub issue #2771
Fixed based on the discussion in ticket

### DIFF
--- a/source/hasPath.js
+++ b/source/hasPath.js
@@ -1,7 +1,6 @@
 import _curry2 from './internal/_curry2';
 import _has from './internal/_has';
 
-import is from './is';
 import isNil from './isNil';
 
 
@@ -33,7 +32,7 @@ var hasPath = _curry2(function hasPath(_path, obj) {
   var val = obj;
   var idx = 0;
   while (idx < _path.length) {
-    if (!isNil(val) && is(Object, val) && _has(_path[idx], val)) {
+    if (!isNil(val) && _has(_path[idx], val)) {
       val = val[_path[idx]];
       idx += 1;
     } else {
--- a/source/hasPath.js
+++ b/source/hasPath.js
@@ -1,6 +1,7 @@
 import _curry2 from './internal/_curry2';
 import _has from './internal/_has';
 
+import is from './is';
 import isNil from './isNil';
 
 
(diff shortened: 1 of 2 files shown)


Bradcomp:
Hi @LorneCurrie! Thanks for the PR! Are you sure this wasn't fixed [here](https://github.com/ramda/ramda/commit/4cd4e4bb85ab9d7a532fbb45e5e70009cdbc2098)? Does the test you wrote fail on master before your changes are applied?

If this still needs to happen, can you please fix the linting errors? You should be able to run the tests and linting locally before you push up.

Thanks again!

LorneCurrie:
@Bradcomp No, I used the latest version of ramda and had an issue where I had an object to the below example.
```
import * as R from 'ramda';

const obj = {
  root: {
    foo: {
      bar: null,
      zar: {
        a: 'a',
        b: 'b',
      }
    }
  }
}

console.log(R.hasPath(['root', 'foo', 'bar', 'far', 'd'], obj));
```
I would have expected `false` to be returned.
`Error`:
```
xxx/node_modules/ramda/src/internal/_has.js:2
  return Object.prototype.hasOwnProperty.call(obj, prop);
                                         ^
TypeError: Cannot convert undefined or null to object
    at hasOwnProperty (<anonymous>)
    at _has (xxxpoc/node_modules/ramda/src/internal/_has.js:2:42)
    at hasPath (/xxxpoc/node_modules/ramda/src/hasPath.js:35:9)
    at Object.f2 [as hasPath] (/xxxpoc/node_modules/ramda/src/internal/_curry2.js:29:14)
    at Object.<anonymous> (/xxxpoc/fibre/client.ts:15:15)
    at Module._compile (internal/modules/cjs/loader.js:701:30)
    at Module.m._compile (/home/lcurrie/.config/yarn/global/node_modules/ts-node/src/index.ts:439:23)
    at Module._extensions..js (internal/modules/cjs/loader.js:712:10)
    at Object.require.extensions.(anonymous function) [as .ts] (/home/lcurrie/.config/yarn/global/node_modules/ts-node/src/index.ts:442:12)
    at Module.load (internal/modules/cjs/loader.js:600:32)
```

LorneCurrie:
Yea, I just looked at it again, and the fix you mentioned worked.  I have removed my change to the hasPath function but left my test cases in as I feel they are valid cases that would happen

CrossEye:
I agree.  These look good.

PR #2827, opened 2019-05-15
fix: Optimize `pathSatisfies` to handle specified `path` not exists

adispring:


source/pathSatisfies.js
```diff
@@ -23,6 +23,10 @@ import path from './path';
  *      R.pathSatisfies(R.is(Object), [], {x: {y: 2}}); //=> true
  */
 var pathSatisfies = _curry3(function pathSatisfies(pred, propPath, obj) {
-  return pred(path(propPath, obj));
+  var result = false;
+  try {
+    result = pred(path(propPath, obj));
```
Qquanwei: IMO. There is the best way to using `pathOr` instead of try...catch

CrossEye:
I'm afraid that I'm :-1: on this.  I know some recent `path` changes have not yet been released.  I'm not sure if they would fix the error that led you to this PR, but I'll see what I can do about getting them out soon.

But what I really don't like about this is that it might suppress an error in the user's `pred` function.  We really don't want to swallow an error that we didn't create.

source/pathSatisfies.js
adispring: Yes, `pathOrSatisfies` does not swallow an error throw from predication:

```js
var pathOrSatisfies = _curry3(function pathOrSatisfies(defaultValue, pred, propPath, obj) {
  result = pred(pathOr(defaultValue, propPath, obj));
  return result;
});
```
@CrossEye

PR #2832, opened 2019-05-23
Update dependencies

kibertoad:


Bradcomp:
Not to derail, but one of our dev dependencies is an old version of Ramda... That seems troublesome. I think we should look into that.

CrossEye:
@Bradcomp:

Yes we should.  I'm pretty sure that we can just replace it with @-latest.  We use Ramda in the build.  If our build can't use a recent version of our library, either the library or the build should get some serious attention.

But as you said, that doesn't need to derail this PR.  It looks good to me.  If you agree, then I think we can go ahead and merge it.

Bradcomp:
:bowtie:

PR #2630, opened 2018-08-09
Added pipeWith/composeWith functions

skallet:
Squashed commits from #2515.

PR #2835, opened 2019-05-27
removing files deprecated in previous release

CrossEye:
Removing the function :floppy_disk: deprecated in v0.26.0:

- #2365 :floppy_disk: `merge` - in favor of new `mergeRight`
- #2452 :floppy_disk: `contains` - renaming to `includes`
- #2630 :floppy_disk: `composeP` - in favor of `composeWith(then)`
- #2630 :floppy_disk: `pipeP` - in favor of `pipeWith(then)`
- #2630 :floppy_disk: `composeK` - in favor of `composeWith(chain)`
- #2630 :floppy_disk: `pipeK` - in favor of `pipeWith(chain)`


anka-213:
Could we add a short note in the documentation about what the deprecated functions are replaced with?

CrossEye:
Just getting back to Ramda after some time away...

@anka-213:  Do you have something in mind, something beyond what we put in our update guides?

morsedav:
@CrossEye if you could just add each of the above "in favor of" and "renamed to" to the docs next to each function it'd be sufficient. Bonus points for a link to the discussion that lead to the decision to deprecate.

e.g. the current docs for [pipeP](https://ramdajs.com/docs/#pipeP) just say that it's depracated and don't suggest an equivalent replacement.

Issue #2836, opened 2019-05-28
Questions about Internal functions _isArrayLike 


itboos:
Recently, our team area analysing  ramda source code , when I see _isArrayLike.js 's
internal implementation， I am counfued。

version: "0.26.1"
```javascript
var _isArrayLike = _curry1(function isArrayLike(x) {
  if (_isArray(x)) { return true; }
  if (!x) { return false; }
  if (typeof x !== 'object') { return false; }
  if (_isString(x)) { return false; }
  if (x.nodeType === 1) { return !!x.length; } // What this word mean? Is it a bug?
  if (x.length === 0) { return true; }
  if (x.length > 0) {
    return x.hasOwnProperty(0) && x.hasOwnProperty(x.length - 1);
  }
  return false;
});
export default _isArrayLike;
```
```javascript
 // if (x.nodeType === 1) { return !!x.length; }  this code result in the following problems

_isArrayLike({nodeType: 1}) // => false
_isArrayLike({nodeType: 1, length: 0}) // => false
_isArrayLike({nodeType: 1, length: 1}) // => true
_isArrayLike({nodeType: 1, length: 2}) // => true

// Is  it not a bug ?
```

CrossEye:
I'm not sure if it actually makes sense.  The idea is clearly to look at DOM elements such as NodeList and Node.  But I think Nodes have `nodeType`, but not `length`, while NodeLists have `length` but not `nodeType`, but there could well be some such case that I'm not seeing.

itboos:
well, I also guess this may use to look at Dom elements such as NodeList and Node. But,  think about the code bellow:
 ```javascript
   _isArrayLike(document.querySelector("html")) // false
   _isArrayLike(document.querySelectorAll('div')) // true
  // this is the right result
```
if we remove  the bellow code:
```
 ( if (x.nodeType === 1) { return !!x.length; }  )
```
becasue NodeList as param, the code runs  to  x.length > 0
 we can get the same result , so  is it redundant ？

CrossEye:
It might well be redundant.  If you want to create a PR, we might get a few more eyes on this.

PR #2837, opened 2019-05-29
Remove _isArrayLike.js's redundant code

itboos:
As we talk about in the issue， I think we shoud remove the redundant code。
[Questions about Internal functions _isArrayLike](https://github.com/ramda/ramda/issues/2836)


CrossEye:
:herb:

Discussion in #2836.

Anyone else, @ramda/core?  Or can anyone think of a case that was covered with this line but not without it?

Bradcomp:
:bowtie: This looks good. Thanks @itboos ❗️

PR #2841, opened 2019-06-05
hasPath: returns true for empty paths

### LINKED ISSUE #2839: R.hasPath returns false for the empty array as a path
## Bug Report

R.hasPath returns false for the empty array as a path

## To Reproduce

```
const R = require('ramda');

console.log(R.hasPath([], { value: 10 })); // false
console.log(R.path([], { value: 10 })); // { value: 10 }
```

## Expected behavior

R.hasPath should returns `true` for the empty array as a path

zouxuoz:
Closes #2839

### DIFF
--- a/test/hasPath.js
+++ b/test/hasPath.js
@@ -63,10 +63,6 @@ describe('hasPath', function() {
     eq(R.hasPath(['toString'], bob), false);
   });
 
-  it('returns false for non-objects', function() {
-    eq(R.hasPath([], obj), false);
-  });
-
   it('tests paths on non-objects', function() {
     eq(R.hasPath(['a', 'b'], undefined), false);
     eq(R.hasPath(['a', 'b'], null), false);
@@ -77,5 +73,14 @@ describe('hasPath', function() {
 
   it('tests currying', function() {
     eq(R.hasPath(['a', 'b'])({ a: { b: 1 } }), true);
+  });
+
+  it('returns true for empty path array', function() {
+    eq(R.hasPath([], obj), true);
+    eq(R.hasPath([], undefined), true);
+    eq(R.hasPath([], null), true);
+    eq(R.hasPath([], true), true);
+    eq(R.hasPath([], ''), true);
+    eq(R.hasPath([], /a/), true);
   });
 });
--- a/source/hasPath.js
+++ b/source/hasPath.js
@@ -26,9 +26,12 @@ import isNil from './isNil';
  *      R.hasPath(['a', 'b'], {});                  // => false
  */
 var hasPath = _curry2(function hasPath(_path, obj) {
-  if (_path.length === 0 || isNil(obj)) {
+  if (_path.length === 0) {
+    return true;
+  } else if (isNil(obj)) {
     return false;
   }
+
   var val = obj;
   var idx = 0;
   while (idx < _path.length) {


test/hasPath.js
```diff
@@ -63,10 +63,6 @@ describe('hasPath', function() {
     eq(R.hasPath(['toString'], bob), false);
   });
 
-  it('returns false for non-objects', function() {
```
zouxuoz: I removed this test because it uses an empty path, cases for non-objects already have covered by tests below

Bradcomp:
What _should_ happen in the following case: `R.hasPath([], null)`?

My intuitions conflict here, and lead me to wonder whether we should be explicitly testing and accounting for a non-object second parameter (I know this has been litigated before, but the situation has changed with this PR.)

I am a lot more comfortable saying the empty path should always return true than I am saying a non object testee should always return false.

zouxuoz:
@Bradcomp You're absolutely right. I added checks for situations with an empty path and non-object values.

CrossEye:
@Bradcomp: I'm really loathe to throw errors if we don't have to.  I like the idea of this change, but hadn't considered the `hasPath([], null)` scenario.

In mathematics there are sometimes such contradictions between two different expectations that end up having to be decided in a somewhat arbitrary way.  Raising zero to any power always returns zero.  And raising any value to the zeroth power always returns one.  So what then is zero raised to the zeroth power?  It's answered by making the definitions match the more useful of these expectations.  (Zero to the zeroth power is one)  I think we can do the same here.  If that breaks backward compatibility, we won't release it in a point release.  I would rather get it right than maintain backward compat.

My feeling is that `hasPath([], val) //=> true` is more fundamental than `hasPath(nodes, null) //=> false`, but I am certainly open to discussion on it.

zouxuoz:
@Bradcomp @CrossEye:

I think we should consider all functions which working with the path and they should work consistently. After a quick review, I found one more interesting behavior: 

```
const R = require('ramda');

console.log(R.assocPath([], 42, null)); // 42
```

CrossEye:
@zouxuoz:  Are you suggesting that this behavior is not correct?  It makes sense to me.  If the path has three nodes, `assocPath` sets the value three layers deep (`{foo: {bar: {baz: 42}}}` )  If it has two nodes, `assocPath` sets it two layers deep (`{foo: {bar: 42}}`.).  If it has one node , `assocPath` sets the value one layer deep (`{foo: 42}`.).  So if the path has no nodes at all, I would expect it so set the value no layers deep, i.e. at the top (`42`.)  Is there some other behavior that makes more sense to you?

Bradcomp:
> My feeling is that hasPath([], val) //=> true is more fundamental than hasPath(nodes, null) //=> false, but I am certainly open to discussion on it.

I agree 100% with this assessment.

> I'm really loathe to throw errors if we don't have to.

I get that, and I know this used to throw before somebody opened a PR to change it.  One subtlety which may be irrelevant is that I am not actually suggesting we throw, I am saying GIGO.

If a person passes a non-object in to `hasPath` I think they are either intentionally relying on ambiguous behavior, or else they have made a coding mistake. Do we want to swallow that?

That being said, I am not gonna die on this hill. I _can_ see how people would argue for returning false and that seems OK to me too. If that's the way we want to go I am :bowtie:  :+1: on this one.

CrossEye:
I still want to think about this for a day or two.

Bradcomp:
@CrossEye have you given this one any more thought?

CrossEye:
@Bradcomp: No, I dropped the ball.  Will try to look this evening.

CrossEye:
I haven't forgotten, just gotten busy again.  I will get to it this weekend, though.

CrossEye:
A breaking change, so punting until after `v1.0`.

PR #2845, opened 2019-06-11
Remove Ramda devDependency

kibertoad:


Bradcomp:
I am :bowtie:  on this, but I'd like at least one other person to weigh in on using `latest`.

kibertoad:
Comes from discussion here: https://github.com/ramda/ramda/pull/2832

Bradcomp:
>  I'm pretty sure that we can just replace it with @-latest. 

Sounds good enough for me! However, I just did a quick repo search and I am not sure that dependency is being used at all. It's possible we could remove it entirely.

@CrossEye Do you know where we are using it?

CrossEye:
I can't find it, nor do I remember where we were using it before.  Perhaps it had to do with the docs before they were separated.  I think it can be safely removed.

Bradcomp:
@kibertoad Can you try removing the dependency completely and see if that effects anything?

kibertoad:
@Bradcomp Done!

CrossEye:
Thank you @kibertoad!

kibertoad:
@CrossEye np! Btw, any timeline on next Ramda release?

PR #2849, opened 2019-06-17
Suggested Diverge Function

suni-masuno:
Finding converge a little challenging to use I've become fond of a different pattern...

```
pipe(
  diverge([add(1), subtract(2)]),
  apply(multiply)
);
```
rather than
```
converge(multiply, [add, subtract])(1, 2);
```
This is a suggestion to enshrine that breakdown as an available function.

I also wonder if `fork` or `branch` might be better names, but diverge appears to describe the relation to converge more clearly.

Please forgive my first attempt at a PR for this project.

### DIFF
--- a/source/diverge.js
+++ b/source/diverge.js
@@ -1,0 +1,32 @@
+import _curry2 from './internal/_curry2';
+import _map from './internal/_map';
+
+
+/**
+ * Accepts an array of arity 1 functions which will be applied to the supplied
+ * data. The results will be returned as an array in the original order.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig ([fns], a) -> [b]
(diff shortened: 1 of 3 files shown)


CrossEye:
Does [`just`][ju] already do what you want?

Also, when #2795 is published, `applySpec` may again do what you would like here.

  [ju]: https://ramdajs.com/docs/#juxt

suni-masuno:
I think `juxt` is in fact sufficient, this was more a matter of name/examples and my failure to understand what it can do.

Thank you for the correction.

CrossEye:
> Please forgive my first attempt at a PR for this project.

I forgot to say this yesterday: that's nothing to forgive, but something to celebrate!  Thank you!

PR #2852, opened 2019-06-22
add P combinator

### LINKED ISSUE #2851: Add the P combinator: R.on ?
I'd love to have the P combinator available in Ramda.

``` javascript
const on = R.curry((f, g, a, b) => f(g(a))(g(b)))
```

thoughts?

Edit: some examples:

``` javascript
const eqBy = on(R.equals)

const mergeOn = on(R.mergeLeft)
const mergeProp = p => mergeOn(R.propOr({}, p))

const searchInsensitive = on(R.includes, R.toLower)
```

DanielFGray:
closes #2851

### DIFF
--- a/source/on.js
+++ b/source/on.js
@@ -1,0 +1,31 @@
+import curryN from './internal/_curryN';
+
+
+/**
+ * Takes a binary function `f`, a unary function `g`, and two values.
+ * Applies `g` to each value, then applies the result of each to `f`.
+ *
+ * Also known as the P combinator.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig ((a, a) -> b) -> (c -> a) -> c -> c -> b
+ * @param {Function} f a binary function
+ * @param {Function} g a unary function
+ * @param {any} a any value
+ * @param {any} b any value
+ * @return {any} The result of `f`
+ * @example
+ *
+ *      const eqBy = R.on((a, b) => a === b);
+ *      eqBy(R.prop('a'), {b:0, a:1}, {a:1}) //=> true;
+ *
+ *      const containsInsensitive = R.on(R.contains, R.toLower);
+ *      containsInsensitive('o', 'FOO'); //=> true
+ * @symb R.on(f, g, a, b) = f(g(a), g(b))
+ */
+var on = curryN(4, [], function on(f, g, a, b) {
+  return f(g(a), g(b));
+});
+export default on;
--- a/test/on.js
+++ b/test/on.js
@@ -1,0 +1,10 @@
+var R = require('../source');
+var eq = require('./shared/eq');
+
+
+describe('on', function() {
+  it('transforms and combines two values', function() {
+    const containsInsensitive = R.on(R.includes, R.toLower);
+    eq(containsInsensitive('o', 'FOO'), true);
+  });
+});
--- a/source/index.js
+++ b/source/index.js
@@ -147,6 +147,7 @@ export { default as o } from './o';
 export { default as objOf } from './objOf';
 export { default as of } from './of';
 export { default as omit } from './omit';
+export { default as on } from './on';
 export { default as once } from './once';
 export { default as or } from './or';
 export { default as otherwise } from './otherwise';


DanielFGray:
Apparently I don't understand the testing system well enough to write tests that pass...

DanielFGray:
Passing, but still needs work.
Docs could be improved, and I don't think my usage of `curry` matches the rest of the codebase.

source/on.js
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Takes a binary curried function `f`, a unary function `g`, and two values.
+ * Applies `g` to each value, then applies the result of each to `f`.
+ *
+ * Also known as the P combinator.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig f -> g -> a -> b -> c
```
CrossEye: I believe this should read:

     * @sig ((b, b) => c) -> (a -> b) -> a -> a -> c

Since `on` is a curried form of (f, g, x, y) => f ( g (x), g (y) ), we can map `a` as the type of `x` and `y`, `b` as the type of `g (x)` and of `g (y)`, and then `g` has the type `a  -> b`, and `f` has the type `(b, b) -> c`  Another way to think about it is that we have three fundamental types, the input type (`a) for which we'll supply to a function (`f`) to convert two values into an output type (`c`) by way of a conversion function (`g`) that generates an intermediate type (`b`) and is applied to each input.

Wadler's famous paper, [Theorems for Free][pw] should demonstrate that this signature strongly restricts the number of possible  implementations that can match.  If I'm not mistaken, after currying it could only be one of these:

```js
(f) => (g) => (x) => (y) => f ( g (x), g (x) )
(f) => (g) => (x) => (y) => f ( g (x), g (y) )
(f) => (g) => (x) => (y) => f ( g (y), g (x) )
(f) => (g) => (x) => (y) => f ( g (y), g (y) )
```

 ... and that is one advantage of such type signatures.


  [pw]: https://people.mpi-sws.org/~dreyer/tor/papers/wadler.pdf

DanielFGray:
I think this should be better? Despite Travis failing for it's own weird reasons.. tests pass on my end anyway.

source/on.js
```diff
@@ -0,0 +1,31 @@
...
+ *
+ *      const eq = R.curry((a, b) => a === b);
+ *      const eqBy = R.on(eq);
+ *      eqBy(R.prop('a'), {b:0, a:1}, {a:1}) //=> true;
+ *
+ *      const containsInsensitive = R.on(R.contains, R.toLower);
+ *      containsInsensitive('o', 'FOO'); //=> true
+ */
+var on = curryN(4, [], function on(f, g, a, b) {
+  return f(g(a))(g(b));
```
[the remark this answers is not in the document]
DanielFGray: It might even be possible to test whether `f` is a function with `f.length === 2` and then automatically curry it, so it would still work either way?

source/on.js
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Takes a binary curried function `f`, a unary function `g`, and two values.
+ * Applies `g` to each value, then applies the result of each to `f`.
+ *
+ * Also known as the P combinator.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig (b -> b -> c) -> (a -> b) -> a -> a -> c
```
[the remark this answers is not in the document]
DanielFGray: I actually just stole the signature from Sanctuary, I haven't much experience with HM types. I'm not sure I even understand what could be ordered differently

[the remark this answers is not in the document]
CrossEye: The point is not the signature block, although that would have to change as well.

This is just switching the first two parameters.  It's the difference between 

```js
curry (f, g, a, b) => f ( g(a ) ) ( g(b) ) )
```

and 

```js
curry (g, f, a, b) => f ( g(a ) ) ( g(b) ) )
```

which would change the signature from

```hs
on :: (b -> b -> c) -> (a -> b) -> a -> a -> c
```

to  

```hs
on :: (a -> b) -> (b -> b -> c) -> a -> a -> c
```

source/on.js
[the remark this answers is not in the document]
CrossEye: Although some day I want to consider changing this across all of Ramda, I agree that a binary function here is more compatible with the rest of Ramda.

source/on.js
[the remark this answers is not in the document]
Bradcomp: > I actually just stole the signature from Sanctuary

Heh, I guess I know where I stood on this a little while ago at least 😛 .

I am OK with the merging function going first as it matches with current behavior of converge and useWith. The other way draws me due to `on(prop('date'), sort)` reading better to me than `on(sort, prop('date'))`, but I am OK with it as it stands.

source/on.js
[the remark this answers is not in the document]
DanielFGray: Fixed

source/on.js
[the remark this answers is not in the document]
DanielFGray: Swapping position of `g` and `f` params seems really odd to me? It would also break expectations if you know `on` from Haskell or Sanctuary..

Maybe under a different name in that case? `by` might also work.

test/on.js
```diff
@@ -0,0 +1,15 @@
+var R = require('../source');
+var eq = require('./shared/eq');
+
+
+describe('on', function() {
+  it('compares identity using identity', function() {
+    const eqBy = R.on((a, b) => a === b);
+    eq(eqBy(R.identity, R.identity, R.identity), true);
+  });
```
CrossEye: I would drop this test.  I find it confusing, and it's hard to see what `on` is doing here.  And of course `eqBy(identity)` is already covered with `identical`.

test/on.js
```diff
@@ -0,0 +1,15 @@
...
+
+  it('finds a substring using lower case', function() {
+    const containsInsensitive = R.on(R.includes, R.toLower);
+    eq(containsInsensitive('o', 'FOO'), true);
+  });
+});
```
CrossEye: This test, on the other hand is precisely what's needed to demonstrate the behavior of the function.

But the test description is still off.  These BDD-style test descriptions are meant to be read as sentences.  For instance, these are the test results for `filter`: 

```
    filter
    ✓ reduces an array to those matching a filter
    ✓ returns an empty array if no element matches
    ✓ returns an empty array if asked to filter an empty array
    ✓ filters objects
    ✓ dispatches to passed-in non-Array object with a `filter` method
```

They describe what `filter` does, not what the test does.  The test should simply be a demonstration of the that description.  Here is what `on` looks like:

```
  on
    ✓ compares identity using identity
    ✓ finds a substring using lower case
 ```

It describes the tests, not the function.

Perhaps something like this for the description?:

```
  on
    ✓ transforms two values with a common function and supplies their results to a combining function
 ```

It is a bit of a mouthful, but this explicitly describes the function.  This might also work, but is even longer: "combines two values by calling a common transformation function on each and a combining function on the results".

source/on.js
```diff
@@ -0,0 +1,30 @@
...
+ * @param {any} b any value
+ * @return {any} The result of `f`
+ * @example
+ *
+ *      const eqBy = R.on((a, b) => a === b);
+ *      eqBy(R.prop('a'), {b:0, a:1}, {a:1}) //=> true;
+ *
+ *      const containsInsensitive = R.on(R.contains, R.toLower);
+ *      containsInsensitive('o', 'FOO'); //=> true
+ */
```
CrossEye: I think it would be useful to add a `@symb` tag as well.  Here it would very much describe the implementation, but that's fine.  See [converge][co] or [map][ma] for examples.


  [co]: https://github.com/ramda/ramda/blob/c29a727b350ad1d37c428ec94ccf9f1ef31ff6c1/source/converge.js#L35
  [ma]: https://github.com/ramda/ramda/blob/c29a727b350ad1d37c428ec94ccf9f1ef31ff6c1/source/map.js#L41-L43

test/on.js
DanielFGray: How about

>combines two values using a transformer function on each value

Or, perhaps more tersely:

>combines and transforms two values

?

source/on.js
DanielFGray: If I understand correctly, this would be

    @symb R.on(f, g, a, b) = f(g(a), g(b))

?

CrossEye: Yes, exactly.

test/on.js
CrossEye: Sure, I like the first one a lot, but either would do.

CrossEye: I think it's fine.  The test reports are meant to be readable, but they are not the main documentation.  I just like the idea that they can serve as secondary documentation.  I personally often look first to the unit test suite when trying to understand a new system.

CrossEye:
This looks great to me.  @Bradcomp?

source/on.js
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Takes a binary function `f`, a unary function `g`, and two values.
+ * Applies `g` to each value, then applies the result of each to `f`.
+ *
+ * Also known as the P combinator.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig ((a, a) -> b) -> (c -> a) -> c -> c -> b
```
davidchambers: I'd prefer `((b, b) -> c) -> (a -> b) -> a -> a -> c`. I like the types to flow from `a` to `b` to `c`.

PR #2855, opened 2019-06-28
feat(applySpec): support literal values

nfantone:
I saw the [introduction of new changes in `applySpec`](https://github.com/ramda/ramda/commit/f3023ea9fc16b2665be890e73d80746cdf2f4112) and was happy to find out that one of the two things that always bugged me about `applySpec` was fixed: support for spec functions in arrays (thanks @bensbigolbeard!). This PR is for the other one I wish made the cut: support for literal values.

The current behaviour is kinda odd and unexpected:

```js
> applySpec({ foo: 'bar', bar: {}, baz: 42, meh: [42] })(2)
{ foo: {}, bar: {}, baz: {}, meh: { '0': {} } }
```

Yes, I'm aware of the whole "crap-in-crap-out" philosophy - but this is just silly, IMHO. Forcing `always` wrappers everywhere seems unnecessary.

After changes in this PR, the above would become:

```js
> applySpec({ foo: 'bar', bar: {}, baz: 42, meh: [42] })(2)
{ foo: 'bar', bar: {}, baz: 42, meh: [ 42 ] }
```

### DIFF
--- a/source/applySpec.js
+++ b/source/applySpec.js
@@ -1,5 +1,6 @@
 import _curry1 from './internal/_curry1';
 import _isArray from './internal/_isArray';
+import always from './always';
 import apply from './apply';
 import curryN from './curryN';
 import max from './max';
@@ -39,15 +40,19 @@ function mapValues(fn, obj) {
  * @example
  *
  *      const getMetrics = R.applySpec({
+ *        list: [R.add, 'value']
  *        sum: R.add,
- *        nested: { mul: R.multiply }
+ *        some: 'value',
+ *        nested: { mul: R.multiply, any: 'value' }
  *      });
- *      getMetrics(2, 4); // => { sum: 6, nested: { mul: 8 } }
+ *      getMetrics(2, 4); // =>  {  list: [6, 'value'], sum: 6, some: 'value', nested: { mul: 8, any: 'value' } }
  * @symb R.applySpec({ x: f, y: { z: g } })(a, b) = { x: f(a, b), y: { z: g(a, b) } }
  */
 var applySpec = _curry1(function applySpec(spec) {
   spec = mapValues(
-    function(v) { return typeof v == 'function' ? v : applySpec(v); },
+    function(v) {
+      return typeof v === 'function' ? v : typeof v === 'object' ? applySpec(v) : always(v);
+    },
     spec
   );
 
--- a/test/applySpec.js
+++ b/test/applySpec.js
@@ -41,6 +41,21 @@ describe('applySpec', function() {
     [['a1', 'a2'], ['b1', 'b2']]);
   });
 
+  it('should retain literal values on the spec', () => {
+    eq(R.applySpec({ v: [42, 'foo'], u: 'bar' })(2), {
+      v: [42, 'foo'],
+      u: 'bar'
+    });
+  });
+
+  it('should support mixing literal values and spec functions', () => {
+    eq(R.applySpec({ v: [x => x + 40, 'foo'], u: 'bar', w: x => 2 * x })(2), {
+      v: [42, 'foo'],
+      u: 'bar',
+      w: 4
+    });
+  });
+
   it('works with a spec defining a map key', function() {
     eq(R.applySpec({map: R.prop('a')})({a: 1}), {map: 1});
   });


Bradcomp:
One of the important tenets of functional programming is that functions are first class - you can treat them the same way you treat every other value. Much of Ramda is based on this concept, utilizing functions as parameters and return values all the time, passing them around, etc.

I worry about any change that would treat a special type of data differently 'under the hood'. What if a  user wants to add a method to an object using `applySpec`? They would have to know that if you want a _function_ to be added they have to wrap it in `always`, but values get passed through. 

If we were to make this change, we would also need to alter the type signature to match the new functionality, but I suspect that type signature would be complicated.

I am 👎  on this unless those issues can be resolved.

nfantone:
@Bradcomp Hi 👋. Thanks for replying. I understand the general point you're raising, but I'm not sure I fall on the same side.

> One of the important tenets of functional programming is that functions are first class - you can treat them the same way you treat every other value.

I see where you're coming from, but how would this PR change the fact that functions are treated as first class citizens in `ramda`, exactly? I sense only a superfluos connection here.

> I worry about any change that would treat a special type of data differently 'under the hood'. 

Understandable. But `applySpec` [already does that](https://github.com/ramda/ramda/blob/03817fe578048652b9532ae71fd1f1705da4de60/source/applySpec.js#L50) by explicitly checking the `typeof` of its input - so, in this case, the counter argument feel stronger to me.

> What if a user wants to add a method to an object using applySpec?

This is a valid objection. It honestly didn't occur to me and have never felt the need to do that - but I do see the discrepancy in the API. The only reply I can think of is that, even with the current approach, adding a function to an object doesn't necessarily require using `always`. This PR adds support for "literal" values. Since the API for `applySpec` applies _functions_, the only way to set a function as a spec is to have a function returning function: `() => () => ...` (a.k.a.: `always(() => ...)` - which is the exact thing you would do if this PR gets merged. Nothing would change.

Bradcomp:
> I see where you're coming from, but how would this PR change the fact that functions are treated as first class citizens in ramda, exactly? I sense only a superfluos connection here.

Currently, `applySpec` has a type signature that restricts the leaves of the spec to be functions. After this change, the leaves could be anything. Once that happens, we will be special casing the leaves that are functions and not treating them the same as the other value leaves.

> But applySpec already does that by explicitly checking the typeof of its input - so, in this case, the counter argument feel stronger to me.

See above. The input is a tree. it has branches and leaves that are treated differently. This is a separate concern from treating the values of leaves differently based on their type. 

The code assumes that there will only be functions as leaves, which is what leads to the odd behavior of your original example. While it might be better to throw if it gets something that's a leaf but not a function, I think GIGO is a reasonable alternative.

> Nothing would change.

A change in the type signature requires rethinking what the reasonable behavior is. Currently (and this would need to be changed too), the first line of the description is this:

> Given a spec object recursively mapping properties to functions

This is no longer true, as we are now accepting a spec recursively mapping properties to anything. but then we aren't treating that anything consistently, instead giving it a special meaning depending on what it is.

The question in modifying the behavior of a function (as opposed to fixing a bug) shouldn't be what changes there are from previous, but what is the ideal behavior, and how can we get that. In this case, I think the tradeoffs aren't really worth it.

nfantone:
I agree with all your premises - just not on the conclusion. My opinion, although obviously personal and small, is that you're leaning a bit too much towards academic formality vs. practicality.

> Currently, applySpec has a type signature that restricts the leaves of the spec to be functions. After this change, the leaves could be anything. Once that happens, we will be special casing the leaves that are functions and not treating them the same as the other value leaves.

Yes. I don't see an issue there. Again, `applySpec` already treats non-function arguments as if they were something special. And still, that doesn't mean functions are less than first class citizens because of that.

> The question in modifying the behavior of a function (as opposed to fixing a bug) shouldn't be what changes there are from previous, but what is the ideal behavior, and how can we get that. 

I honestly fail to see how inexplicably transforming a boolean value to `{}` can be "ideal behavior".

```js
applySpec({ b: true })(42); // { b: {} } 🤔 
```

Yes, it's bad input. Yes, it's not complying with the function signature. Is it **ideal**? I'd argue it's far from it. Is the solution proposed in this PR _ideal_? I don't know - I'm opening the debate. But, in my view, it's a lot closer to being reasonable.

I'd rather have a function signature rethought and modified accordingly than trying to debug an API endpoint that started returning things like `{ '0': {} }` for no obvious reason after a sneaky bug in some release.

My main concern here is that in order to pursue a purist form, we are relegating use cases and situations that actually _happen_ on a daily basis and have a real impact. 

Let's think for a moment on that hypothetical scenario I started describing above: how many software engineers/developers, when confronted with that kind of bug, do you think are going to come forward and say something like: _"Oh, I got this! This must be the output of `applySpec`, which uses a special, internal `mapValues` function that calls `keys` to iterate on keys as if all non-function inputs where objects! That's where the `'0': {}` comes from!"_. 

The fact that you _need_ to know the internal workings of `ramda` to even be able to diagnose many of the problems is, to me, one of the major pain points of this library (which I ❤️ and use every single day 🙏). I believe we should focus on ironing those and aiding users, rather than punishing when a signature wasn't appropriately followed.

CrossEye:
> The fact that you *need* to know the internal workings of `ramda` to even be able to diagnose many of the problems is, to me, one of the major pain points of this library (which I :heart: and use every single day :pray:). 

Fixing issues that cause such pain points is fine.  Specific fixes need to be balanced against the the Ramda GIGO philosophy, though.  It's always been a no-hand-holding library, and we tend to err toward the side of "let the chips fall where they may" if the user supplies bad input.  But in many cases, when there is a clear way to return something useful in such cases, we will ignore the above and do some sort of type-checking or other ugly code branching.


> I believe we should focus on ironing those and aiding users, rather than punishing when a signature wasn't appropriately followed.

It's not a matter of punishing users, only of focusing chiefly on those who use Ramda's functions as designed and documented.

----------

But much more central to Ramda's philosophy is to keep things *simple*.  A function should do one thing and one thing only.

Often a Ramda function will operate on multiple types, but only if there is a common abstraction that they share.  Thus `map` works on arrays, objects, functions, and other types exactly because they all share the common type of `Functor`.

`applySpec` is pretty clear now.  It accepts an object representing a tree whose leaf nodes are functions, using this structure to create an object of the same structure by applying those functions to our input values.  The suggested changes add complexity, not only to the implementation, but more importantly, to the API.

The concerns from @Bradcomp are real; there would be legitimate fall-out from such a change.  But to me the fundamental issue is one of corrupting a simple function by adding unnecessary complexity.

My feeling is that this sort of usage of `applySpec` is just barely ok:

```js
const minimum = apply (Math.min), maximum = apply (Math.max)

const transform = applySpec ({
  title: always ('Calc Results'),
  type: always ('Statistical'),
  values: identity,
  statistics: {length, sum, mean, maximum, minimum},
})

transform ([8, 6, 7, 5, 3, 0, 9]) //=> 
// {
//     title: "Calc Results",
//     type: "Statistical",
//     values: [8, 6, 7, 5, 3, 0, 9],
//     statistics: {
//         length: 7,
//         maximum: 9,
//         mean: 5.428571428571429,
//         minimum: 0,
//         sum: 38
//     }
// }
```

I think this would be cleaner and clearer:

```js
const transform = pipe (
  applySpec ({
    values: identity,
    statistics: {length, sum, mean, maximum, minimum},
  }),
  mergeRight ({title: 'Calc Results', type: 'Statistical'})
)
```

But that this would be merely confusing:

```js
const transform = alternateApplySpec ({
  title: 'Calc Results',
  type: 'Statistical',
  values: identity,
  statistics: {length, sum, mean, maximum, minimum},
})
```

Reading that code, I have no idea what sort of value `mean` or `identity` is.

So I'm afraid I'm definitely :-1: on this.

Bradcomp:
It's been a while, I am gonna close this. Feel free to reopen if you feel it warrants more discussion.

PR #2856, opened 2019-06-30
Faster _curryN function

Jorge-Gonzalez:
I found this implementation to be a bit faster.

### DIFF
--- a/source/internal/_curryN.js
+++ b/source/internal/_curryN.js
@@ -14,25 +14,35 @@ import _isPlaceholder from './_isPlaceholder';
  */
 export default function _curryN(length, received, fn) {
   return function() {
-    var combined = [];
+    var receivedLen = received.length;
+    var holdersCount = 0;
+    var argsLen = arguments.length;
+    var argsHoldersCount = 0;
+    var left;
+    var combined;
+    var idx = -1;
     var argsIdx = 0;
-    var left = length;
-    var combinedIdx = 0;
-    while (combinedIdx < received.length || argsIdx < arguments.length) {
-      var result;
-      if (combinedIdx < received.length &&
-          (!_isPlaceholder(received[combinedIdx]) ||
-           argsIdx >= arguments.length)) {
-        result = received[combinedIdx];
-      } else {
-        result = arguments[argsIdx];
-        argsIdx += 1;
-      }
-      combined[combinedIdx] = result;
-      if (!_isPlaceholder(result)) {
-        left -= 1;
-      }
-      combinedIdx += 1;
+    for (var i = 0; i < receivedLen; i++) {
+      if (_isPlaceholder(received[i])) holdersCount++;
+    }
+    for (var i = 0; i < argsLen; i++) {
+      if (_isPlaceholder(arguments[i])) argsHoldersCount++;
+    }
+    left = length - receivedLen + holdersCount - argsLen + argsHoldersCount;
+    combined = Array(
+      argsLen > holdersCount
+        ? receivedLen + argsLen - holdersCount
+        : receivedLen
+    );
+    while (++idx < receivedLen) {
+      combined[idx] = _isPlaceholder(received[idx])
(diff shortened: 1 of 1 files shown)


CrossEye:
I'm perfectly fine with changing this function if it improves performance, but if we're doing so, do you have some benchmarks we can see?  Or could you create one to demonstrate?

Note that this is [failing linting][li] for a few reasons:

```
/home/travis/build/ramda/ramda/source/internal/_curryN.js
  22:5   error  Combine this with the previous 'var' statement with uninitialized variables \
                                                                                  one-var
  25:38  error  Unary operator '++' used                                          no-plusplus
  26:7   error  Expected { after 'if' condition                                   curly
  26:40  error  Unary operator '++' used                                          no-plusplus
  28:34  error  Unary operator '++' used                                          no-plusplus
  29:7   error  Expected { after 'if' condition                                   curly
  29:41  error  Unary operator '++' used                                          no-plusplus
  37:12  error  Unary operator '++' used                                          no-plusplus
  40:23  error  Unary operator '++' used                                          no-plusplus
  45:16  error  Unary operator '++' used                                          no-plusplus
  45:35  error  Unary operator '++' used                                          no-plusplus
```
But there is one more stylistic change that I would request.  House style uses `while` loops, not `for` loops.  One of the founders of Ramda has made it a personal goal to never write another explicit `for` loop, and it has made it into our codebase.  It would better to keep this consistency.


  [li]: https://travis-ci.org/ramda/ramda/jobs/552330172

Bradcomp:
@CrossEye I didn't know the origin of that convention! That founder's goal is close to my heart as well.

CrossEye:
@Bradcomp: I agree with @buzzdecafe about it too, although I'm perhaps less adamant about it.  I think I was the one who wrote the original pass where we used `while` loops, but that had to do as much with ancient lore about the relative performance of `while` and `for` as with this distaste for `for`.  But I like not seeing any ugly C-style `for` loops in our code-base.

That's a more minor objection, but it's real.  The syntax of the `for(init; test; step) {body}` is pretty horrible.  Just watching the order control flow moves between `init`, `test`, `step`, and `body` helps explain why many find computer programming arcane.  Even using the same semicolon used otherwise as a statement terminator as an internal separator is odd.  I have much less problem with `for-of` statements.

But of course the main objection is that unless it's for side effects, pretty well every `for` loop I see can be more clearly and explicitly expressed as a `filter`, `find`, `map`, `some`, `every`, or `reduce`.   The objection is simply one of code clarity.  And if it's for side-effects, `forEach` has cleaner syntax.

Jorge-Gonzalez:
Yes for loops apart from having that eccentric structure are really hard to read, they are kind of the essence of imperative code I guess.

I apologize for not having linted the code with the settings of the project, and and I appreciate your patience about that.

I have been doing some tests and I noticed that my implementation is not as fast as it seemed according to the first tests. That is because although been faster in V8, in SpiderMonkey it is a bit slower.

At https://jsperf.com/curry-fn-comparison/27 you can see the test. I also, added an option that detects if there is a placeholder to handle, or if you can just concatenate the arguments. Interestingly, this option is not always faster than the one that deals indiscriminately with each call.

Personally I think that optimizations should be framed in the most common use cases.

In my first test the numbers doubled the performance of the function which made me think that it was worth to make the pull request.

CrossEye:
@Jorge-Gonzalez:

I want to thank you for your effort on this.  It's always worth investigating whether we can improve performance of core functions.  But looking at the results of that test, I don't think this is a clear winner, and as it adds to the code-base, I don't see any real reason to proceed.

@ramda/ramda-core: Do others agree?

I won't close this right now.  If you feel strongly about it, feel free to continue arguing for it.


Feel free

Jorge-Gonzalez:
With pleasure, thanks to you for the time and effort dedicated to this excellent library.
I Agree with you, and do not see results that support this change.

PR #2862, opened 2019-07-16
clone: Update description

### LINKED ISSUE #2850: Clone does  not do a deep clone
Steps to reproduce: 
1. Go to https://ramdajs.com/repl/?v=0.26.1 and open your developer console. 
2. Copy the below snippet and run it in console

```javascript
(() => {
     //
    let initialObj = [{
        attr1: 'value1',
        attr2: 'value2',
        attr3: 'value3',
        attr4: 'value4',
        attr5: 'value5',
    }];

    // the new constructed object
    let foo = {
        bar: initialObj[0],
        baz: initialObj[0]
    }


    let clonedFoo = R.clone(foo);

    console.log(clonedFoo);
    clonedFoo.bar[0] = 'valueChanged';
    console.log(clonedFoo);

})()
```

The expected behaviour: 

When the attribute(`bar`) of the cloned object(`clonedFoo`) is changed it will change the other attribute(`baz`) as well. While *deep* clone should be a complete new clone of the object...

DanielRX:
Updates the clone description to match the way the function behaves (preserving structural isomorphism with the source object). (fixes #2850)

### DIFF
--- a/source/clone.js
+++ b/source/clone.js
@@ -3,9 +3,9 @@ import _curry1 from './internal/_curry1';
 
 
 /**
- * Creates a deep copy of the value which may contain (nested) `Array`s and
- * `Object`s, `Number`s, `String`s, `Boolean`s and `Date`s. `Function`s are
- * assigned by reference rather than copied
+ * Creates a deep copy of the source that can be used in place of the source object without retaining any references to it.
+ * The source object may contain (nested) `Array`s, `Object`s, `Number`s, `String`s, `Boolean`s and `Date`s.
+ * `Function`s are assigned by reference rather than copied.
  *
  * Dispatches to a `clone` method if present.
  *
--- a/source/clone.js
+++ b/source/clone.js
@@ -3,11 +3,17 @@ import _curry1 from './internal/_curry1';
 
 
 /**
- * Creates a deep copy of the source that can be used in place of the source object without retaining any references to it.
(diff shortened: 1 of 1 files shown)


source/clone.js
```diff
@@ -3,9 +3,9 @@ import _curry1 from './internal/_curry1';
 
 
 /**
- * Creates a deep copy of the value which may contain (nested) `Array`s and
- * `Object`s, `Number`s, `String`s, `Boolean`s and `Date`s. `Function`s are
- * assigned by reference rather than copied
+ * Creates a deep copy of the source that can be used in place of the source object without retaining any references to it.
+ * The source object may contain (nested) `Array`s, `Object`s, `Number`s, `String`s, `Boolean`s and `Date`s.
+ * `Function`s are assigned by reference rather than copied.
```
CrossEye: Could you drop the column width to match other documentation blocks?

source/clone.js
```diff
@@ -3,9 +3,9 @@ import _curry1 from './internal/_curry1';
...
  *
```
CrossEye: Do we want an additional, more explicit paragraph here?  Perhaps something like:

>      * Note that if the source object has multiple nodes that share a reference,
>      * the returned object will have the same structure, but the data the new 
>      * references point to will be a *copy* of the original.

CrossEye:
@DanielRX: Thank you very much!

PR #2871, opened 2019-07-29
[documentation] Change order of examples

nabati:
Makes more pedagogical sense that the examples have the same order as the supplied functions.


CrossEye:
Agreed.  this is definitely a better order.  Thanks!

PR #2882, opened 2019-08-16
fix(_clone): refFrom and refTo assignment wrong

nishino-tsukasa:
```js
var len = refFrom.length;
var idx = 0;
while (idx < len) {
  if (value === refFrom[idx]) {
    return refTo[idx];
  }
  idx += 1;
}
refFrom[idx + 1] = value;
refTo[idx + 1] = copiedValue;
```

idx is already equal to refFrom.length.

```js
refFrom[refFrom.length + 1] = value;
refTo[refFrom.length + 1] = copiedValue;
```
Obviously, is wrong

### DIFF
--- a/source/internal/_clone.js
+++ b/source/internal/_clone.js
@@ -22,8 +22,8 @@ export default function _clone(value, refFrom, refTo, deep) {
       }
       idx += 1;
     }
-    refFrom[idx + 1] = value;
-    refTo[idx + 1] = copiedValue;
+    refFrom[idx] = value;
+    refTo[idx] = copiedValue;
     for (var key in value) {
       copiedValue[key] = deep ?
         _clone(value[key], refFrom, refTo, true) : value[key];


Bradcomp:
> Obviously, is wrong

Hi @nishino-tsukasa, thanks for the PR!

Can you help me understand what is wrong about it. Is there a test that can expose what the bug is? This PR doesn't seem to change any of the existing tests, and it isn't based on a bug report, so I am struggling to understand what this fixes.

nishino-tsukasa:
> > Obviously, is wrong
> 
> Hi @nishino-tsukasa, thanks for the PR!
> 
> Can you help me understand what is wrong about it. Is there a test that can expose what the bug is? This PR doesn't seem to change any of the existing tests, and it isn't based on a bug report, so I am struggling to understand what this fixes.

From a functional point of view, it's not a bug, because test cases can pass. 
I think this is a "semantic" bug. These two arrays should not be sparse arrays.
eg.
```js
// test case: clones deep object
it('clones deep object', function() {
  var obj = {a: {b: {c: 'ramda'}}};
  var clone = R.clone(obj);
  obj.a.b.c = null;
  eq(clone, {a: {b: {c: 'ramda'}}});
});
// print refFrom
[
  empty item,
  { a: { b: {c: 'ramda'} } },
  empty item>,
  { b: { c: 'ramda' } },
  empty item,
  { c: 'ramda' } 
]
```
It's confusing.

CrossEye:
Nice catch.  Thank you!

Issue #2881, opened 2019-08-15
Suggesting a new function to split an array into multiple at a given value


mudit56:
Several times while working with different data set, I felt the need for a function to do the partitions like below : 

[1, 2, 3 ,2, 4, 5, 2, 6, 7]  ( break at 2 ) ===>  [ [1] , [3] , [4,5] , [6,7] ]
[1, 2, 1]  (break at 2) ===> [ [1] ,  [1] ]
[1, 2, 1] (break at 1) ===> [ [2] ]

I have also worked on the same which also support whether to exclude or include the delimiter. I have followed the contribution guidelines, and willing to provide the PR.

CrossEye:
I could definitely imagine including something like `splitWhenever` or something.

```hs
splitWhenever :: (a -> Bool) -> [a] -> [[a]]
```

Would you be interested in creating a PR for it?  No guarantees it would be accepted, but I promise at least a fair hearing.

mudit56:
I have opened the PR as suggested.

PR #2884, opened 2019-08-19
Splitwhenever fixes#2881

mudit56:
Hi, this is in regards to issue #2881 , I have raised the PR as suggested. please have a look. 
I have tried to follow the contribution, however do suggest if some changes are required. 

 * @sig a -> [a] -> [[a]]
 * @sig a -> [a] -> Boolean -> [[a]]

R.splitWhenever(2, [1, 2, 3 ,2, 4, 5, 2, 6, 7]); //=> [[1],[3],[4,5],[6,7]]
R.splitWhenever(2, [1, 2, 3 ,2, 4, 5, 2, 6, 7],true); //=> [[1],[2,3],[2,4,5],[2,6,7]]

Supports whether to include or exclude the delimiter. 
 

@CrossEye


buzzdecafe:
https://ramdajs.com/docs/#splitWhen

is more general

CrossEye:
@buzzdecafe:

`splitWhen`  is more general in one sense, taking a predicate rather than a value.  But it breaks the list only on the first match.  The idea is to break multiple times.  I think we need a combination of the above with `splitWhen`.

test/splitWhenever.js
```diff
@@ -0,0 +1,21 @@
+var assert = require('assert');
+var R = require('../source');
+var eq = require('./shared/eq');
+
+
+describe('splitWhenever', function() {
+
+  it('Splits an array into slices on every occurrence of a value', function() {      
+    eq(R.splitWhenever(2,[1, 2, 3 ,2, 4, 5, 2, 6, 7]), [[1],[3],[4,5],[6,7]]);
```
CrossEye: House style could add a space after each comma.

```js
eq(R.splitWhenever(2, [1, 2, 3, 2, 4, 5, 2, 6, 7]), [[1], [3], [4, 5], [6, 7]]);
```

source/splitWhenever.js
```diff
@@ -0,0 +1,37 @@
...
+ * @param {Number} n
+ * @param {Array} list
+ * @param {Boolean} incDelimiter
+ * @return {Array}
+ * @example
+ *
+ *      R.splitWhenever(2, [1, 2, 3 ,2, 4, 5, 2, 6, 7],true); //=> [[1],[2,3],[2,4,5],[2,6,7]]
+ *      R.splitWhenever(2, [1, 2, 3 ,2, 4, 5, 2, 6, 7]); //=> [[1],[3],[4,5],[6,7]]
+ */
+var splitWhenever = _curryN(2, [], function splitWhenever(n, list, incDelimiter) {
```
CrossEye: Ramda is allergic to optional parameters.  If this parameter is necessary, then it should come early and allow the user to partially apply it, something like this:

```js
const splitInclusive = R.splitWhenever(true)
const splitExclusive = R.splitWhenever(false)

splitInclusive(2. [1, 2, 3, 5, 2, 7]) //=> [[1, 2], [3, 5, 2], [7]]
splitExclusive(2. [1, 2, 3, 5, 2, 7]) //=> [[1], [3, 5], [7]]
```

But this still offers another question: if we are thinking of including the delimiter, is there a strong reason to say that it *ends* a group rather than *begins* one.  That is, why choose 

```js
splitInclusive(2. [1, 2, 3, 5, 2, 7]) //=> [[1, 2], [3, 5, 2], [7]]
```

over

```js
splitInclusive(2. [1, 2, 3, 5, 2, 7]) //=> [[1], [2, 3, 5], [2, 7]]
```

?

This makes me think that the inclusive option might not be worth the trouble at all.

mudit56:
Sure, we can remove the inclusive option, It won't be of much use if I think about it. 
That will resolve two questions here, the one with optional parameter and also the position of delimiter. 

So Now, if I understand it correctly, we need to implement predicate instead of value, right? if that so I will update the PR.

CrossEye:
> So Now, if I understand it correctly, we need to implement predicate instead of value, right? if that so I will update the PR.

Yes, that would be much more in keeping with Ramda's design.

(I may have more to discuss after you do so.  I really haven't looked at your implementation yet, only the tests.)

mudit56:
Requested changes has been done. it now supports predicates instead of values.

mudit56:
Predicate examples: 
eq(R.splitWhenever(R.equals(2), [1, 2, 3, 2, 4, 5, 2, 6, 7]), [[1], [3], [4,5], [6,7]]);
eq(R.splitWhenever((e) => e % 2 === 0 ? true : false, [1, 2, 3, 2, 4, 5, 2, 6, 7]),[[1], [3], [5], [7]]);

source/splitWhenever.js
```diff
@@ -0,0 +1,33 @@
+import _curryN from './internal/_curryN';
+/**
+ * Splits an array into slices on every occurrence of a value.
+ *
+ * @func
+ * @memberOf R
+ * @since v0.26.1
+ * @category List
+ * @sig a -> [a] -> [[a]]
+ * @sig a -> [a] -> Boolean -> [[a]]
```
CrossEye: I believe the only signature should be `(a -> Boolean) -> [a] -> [[a]]`

test/splitWhenever.js
```diff
@@ -0,0 +1,17 @@
+var R = require('../source');
+var eq = require('./shared/eq');
+
+
+describe('splitWhenever', function() {
+  it('Splits an array into slices on every occurrence of a value', function() {
+    eq(R.splitWhenever(R.equals(2), [1, 2, 3, 2, 4, 5, 2, 6, 7]), [[1], [3], [4,5], [6,7]]);
```
CrossEye: More space: `... [4, 5], [6, 7] ...`

test/splitWhenever.js
```diff
@@ -0,0 +1,17 @@
...
+    eq(R.splitWhenever(R.equals(2), [1, 2, 1]), [[1], [1]]);
+    eq(R.splitWhenever(R.equals(1), [1, 2, 1]), [[2]]);
+  });
+});
+describe('splitWhenever', function() {
+  it('Splits an array into slices, Supports conditions defined by predicate fuctions ', function() {
+    eq(R.splitWhenever((e) => e % 2 === 0 ? true : false, [1, 2, 3, 2, 4, 5, 2, 6, 7]),[[1], [3], [5], [7]]);
```
CrossEye: More space: `... 2, 6, 7]), [[1], [3]...`

source/splitWhenever.js
```diff
@@ -0,0 +1,33 @@
...
+ *      R.splitWhenever(2, [1, 2, 3 ,2, 4, 5, 2, 6, 7],true); //=> [[1],[2,3],[2,4,5],[2,6,7]]
+ *      R.splitWhenever(2, [1, 2, 3 ,2, 4, 5, 2, 6, 7]); //=> [[1],[3],[4,5],[6,7]]
+ */
+var splitWhenever = _curryN(2, [], function splitWhenever(pred, list) {
+  var acc = [];
+  var curr = [];
+  for (var i = 0;i < list.length;i = i + 1) {
+    curr.push(list[i]);
+    if (i < list.length - 1 && pred(list[i + 1]) || i === list.length - 1) {
+      (pred(curr[0]) && curr.length > 1) && curr.shift();
```
CrossEye: Please shift to an `if` statement

source/splitWhenever.js
```diff
@@ -0,0 +1,33 @@
...
+      !(curr.length === 1 && pred(curr[0])) && acc.push(curr);
```
CrossEye: Ditto

mudit56:
Fixed the if statements and the spaces

CrossEye:
I'm good with this.  I personally keep reinventing this function, so I see a need.

What do others think, @ramda/core ?

source/index.js
```diff
@@ -155,6 +155,7 @@ export { default as pair } from './pair';
 export { default as partial } from './partial';
 export { default as partialRight } from './partialRight';
 export { default as partition } from './partition';
+export { default as splitWhenever } from './splitWhenever';
```
Bradcomp: This should be moved to the proper place alphabetically, right after `splitWhen`

test/splitWhenever.js
```diff
@@ -0,0 +1,17 @@
...
+
+describe('splitWhenever', function() {
+  it('Splits an array into slices on every occurrence of a value', function() {
+    eq(R.splitWhenever(R.equals(2), [1, 2, 3, 2, 4, 5, 2, 6, 7]), [[1], [3], [4, 5], [6, 7]]);
+    eq(R.splitWhenever(R.equals(2), [1, 2, 1]), [[1], [1]]);
+    eq(R.splitWhenever(R.equals(1), [1, 2, 1]), [[2]]);
+    eq(R.splitWhenever(R.equals(3), [1, 2, 1]), [[1, 2, 1]]);
+  });
+});
+describe('splitWhenever', function() {
```
Bradcomp: Can't we have both of these tests in the same `describe` block?

test/splitWhenever.js
```diff
@@ -0,0 +1,17 @@
+var R = require('../source');
+var eq = require('./shared/eq');
+
+
+describe('splitWhenever', function() {
```
Bradcomp: I'd like to see a few more tests in here. Specifically, we should be testing what happens with an empty array, as well as where the first or last element matches the predicate, as well as when no element of the list matches the predicate.

mudit56:
Done with the requested changes. 

Also for the new tests.

- In case of empty array, returns an empty array 
```javascript
 eq(R.splitWhenever(R.equals(3), []), []);
```
- When first or last element matches the predicate, returns one slice of all the other elements
```javascript
 eq(R.splitWhenever(R.equals(1), [1, 2, 3]), [[2, 3]]);
```
- When no element matches the predicate, returns one slice of the original array (doesn't return the original array to be consistent with the design)
```javascript
 eq(R.splitWhenever(R.equals(4), [1, 2, 3]), [[1, 2, 3]]);
```

CrossEye:
Thank you very much @mudit56 !

mudit56:
@CrossEye @Bradcomp - Thanks Guys, Glad to be able to contribute.

Issue #2799, opened 2019-03-15
R.isEmpty(Buffer.from("")) returns false


abeMedia:
It appears `isEmpty` does not work for Buffers. Since Buffers have the `length` property, my suggestion would be to first check for the length property and return true if this exists and is `0`. This would already cover almost everything but objects and would probably also be more efficient than creating an empty var to compare against.

CrossEye:
Are you interested in creating a PR for this?  I think there might be some significant discussion, and I can't guarantee that it would be accepted, but it's easier to talk about a concrete proposal.

DanielRX:
@abeMedia Looking at the test cases for `isEmpty` I see `isEmpty({length: 0}) -> false` so I don't think you could check it that way (unless you first rule out that case)
Also just to note, `Array(1)` produces an empty array with length 1, so this also would not catch that case

canytra:
I took a stab at this...

Instead of just Buffer, implemented it for all TypedArrays.

CrossEye:
@cantyra: Thank you!  The PR looks great to me!

canytra:
No problem! And thanks! Happy to contribute

canytra:
@CrossEye I think this can safely be closed now. Thank you!

PR #2885, opened 2019-08-20
Add support for typed arrays for empty and isEmpty

canytra:
Added support for TypedArrays for R.empty which also allowed TypedArrays to work for R.isEmpty.

#2799

### DIFF
--- a/source/internal/_isTypedArray.js
+++ b/source/internal/_isTypedArray.js
@@ -1,0 +1,23 @@
+/**
+ * Tests whether or not an object is a typed array.
+ *
+ * @private
+ * @param {*} val The object to test.
+ * @return {Boolean} `true` if `val` is a typed array, `false` otherwise.
+ * @example
+ *
+ *      _isTypedArray(new Uint8Array([])); //=> true
+ *      _isTypedArray(new Float32Array([])); //=> true
+ *      _isTypedArray([]); //=> false
+ *      _isTypedArray(null); //=> false
+ *      _isTypedArray({}); //=> false
+ */
+export default function _isTypedArray(val) {
+  var type = Object.prototype.toString.call(val);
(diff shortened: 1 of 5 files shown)


test/empty.js
```diff
@@ -30,6 +30,13 @@ describe('empty', function() {
     eq(R.empty([1, 2, 3]), []);
   });
 
+  it('returns empty typed array of equivalent type given typed array', function() {
+    eq(R.empty(Uint8Array.from('123')), Uint8Array.from(''));
+    eq(R.empty(Uint8Array.from('123')).constructor.name, 'Uint8Array');
```
canytra: Not sure if this is necessary.

I only added it because `eq(R.empty(Uint8Array.from('123')), Float32Array.from(''));` was passing.

source/internal/_isTypedArray.js
```diff
@@ -0,0 +1,23 @@
...
+ *
+ *      _isTypedArray(new Uint8Array([])); //=> true
+ *      _isTypedArray(new Float32Array([])); //=> true
+ *      _isTypedArray([]); //=> false
+ *      _isTypedArray(null); //=> false
+ *      _isTypedArray({}); //=> false
+ */
+export default function _isTypedArray(val) {
+  var type = Object.prototype.toString.call(val);
+  return type === '[object Uint8ClampedArray]' ||
```
canytra: I wasn't sure if this or `[...].indexOf(type)` was preferable.

CrossEye: This looks fine to me

test/empty.js
CrossEye: Anything that helps prevent a possible regression seems fine to me.

CrossEye:
Thank you very much @canytra!

PR #2888, opened 2019-08-30
Replace defaultTo with undefined check in pathOr

Tworley96:


CrossEye:
I think what we have documented is what we probably want to support, even if that breaks backward compatibility: `[prop|path]Or` should return whatever value is there at the property -- even if it's a nil value -- so long as there is such a named property, and the default value otherwise.

That this is different from `defaultTo` doesn't bother me at all; that one is explicitly about handling nil values.

test/path.js
```diff
@@ -54,8 +54,8 @@ describe('path', function() {
     eq(R.path(['a', 'b'], {a: null}), undefined);
   });
 
-  it('works with falsy items', function() {
-    eq(R.path(['toString'], false), Boolean.prototype.toString);
+  it('returns undefined for non-object items', function() {
```
Tworley96: toString in false errors, so this test no longer makes sense

Tworley96:
It's been awhile but I believe that this has fixed all of the stated issues. If not let me know.

CrossEye:
Yes, it does.  Sorry this got dropped.  Thanks for the contribution!  :herb:

Bradcomp:
@Tworley96 @CrossEye  I've reverted this merge because it breaks our test suite. Can we get an update so we can add this to the codebase?

Tworley96:
I've updated the code so that the test suite passes in #2995. The code that was reverted will undo some of the optimizations in #2962. Personally, I think some of these changes might not be the best and there might be some more discussion to be had about how these functions handle null vs undefined.

CrossEye:
> I've reverted this merge because it breaks our test suite.

Weird, how did I miss that!?

PR #2893, opened 2019-09-18
chore(differenceWith): updated code example for difference with to be…

BobbieBarker:
…tter illuminate the behaviour of the function through the documentation

### DIFF
--- a/source/differenceWith.js
+++ b/source/differenceWith.js
@@ -23,6 +23,9 @@ import _curry3 from './internal/_curry3';
  *      const l1 = [{a: 1}, {a: 2}, {a: 3}];
  *      const l2 = [{a: 3}, {a: 4}];
  *      R.differenceWith(cmp, l1, l2); //=> [{a: 1}, {a: 2}]
+ *
+ *      R.differenceWith(R.equals, [1, 2, 3, 3, 3], []); //=> [1, 2, 3]
+ *      R.differenceWith(R.equals, [1, 2, 3, 3, 3], [1]); //=> [2, 3]
  */
 var differenceWith = _curry3(function differenceWith(pred, first, second) {
   var out = [];


source/differenceWith.js
```diff
@@ -23,6 +23,10 @@ import _curry3 from './internal/_curry3';
  *      const l1 = [{a: 1}, {a: 2}, {a: 3}];
  *      const l2 = [{a: 3}, {a: 4}];
  *      R.differenceWith(cmp, l1, l2); //=> [{a: 1}, {a: 2}]
+ *      // Additionally, if the second set is empty or smaller than the first set
```
CrossEye: I don't think this has anything to do with the second set being smaller:

```js
R.differenceWith(R.equals, [1, 2, 3, 3, 3], [4, 5, 6, 7, 8, 9]) //=> [1, 2, 3]
```

source/differenceWith.js
```diff
@@ -23,6 +23,10 @@ import _curry3 from './internal/_curry3';
...
+ *      // the first set will be used to deduplicate itself using the supplied predicate.
+ *      R.differenceWith(R.eq, [1, 2, 3, 3, 3], []); //=> [1, 2, 3]
+ *      R.differenceWith(R.eq, [1, 2, 3, 3, 3], [1]); //=> [2, 3]
```
CrossEye: `eq` was renamed to `equals` long ago.  Can you change it here?

Bradcomp:
This looks good to me once @CrossEye's changes are made. A few things I considered were whether this behavior is tested for (it is) and whether we need to consider this for other set operations. I don't think we do because all other operations are symmetrical.

:bowtie:

source/differenceWith.js
kurtmilam: That's correct - the resulting array is deduplicated via the supplied predicate in all cases.

kurtmilam:
Not to bikeshed, but I wonder whether a different set of examples would be more helpful. The problem with these examples is that they use `R.equals`, which may not make it clear to the reader that the predicate is being used to dedupe the result array.

Here's another suggestion, worked into the existing docs:
```
 *      const cmp = (x, y) => x.a === y.a;
 *      const l1 = [{a: 1}, {a: 2}, {a: 3}];
 *      const l2 = [{a: 3}, {a: 4}];
 *      R.differenceWith(cmp, l1, l2); //=> [{a: 1}, {a: 2}]
 *
 *      const l3 = [{a: 2}, {a: 2}, {a: 3}];
 *      R.differenceWith(cmp, l3, l2); //=> [{a: 2}]
 *      R.differenceWith(cpm, l3, []); //=> [{a: 2}, {a: 3}]
```

CrossEye:
@kurtmilam: Not to bikeshed your bikeshed :smile:, but I have a difficult time distinguishing between 

```js
 *      R.differenceWith(cmp, l3, l2); //=> [{a: 2}]
```

and

```js
 *      R.differenceWith(cmp, 13, 12); //=> [{a: 2}]
```

("ell-3" versus "thirteen")

I like this example, but perhaps with different variable names.

kurtmilam:
I completely agree about the variable names. I just kept the `l1` and `l2` currently in the docs and extended it to `l3`, and I can report that it was a little confusing to keep them straight.

How about swapping 'l' for 'a (for 'array'):
```
 *      const cmp = (x, y) => x.a === y.a;
 *      const a1 = [{a: 1}, {a: 2}, {a: 3}];
 *      const a2 = [{a: 3}, {a: 4}];
 *      R.differenceWith(cmp, a1, a2); //=> [{a: 1}, {a: 2}]
 *
 *      const a3 = [{a: 2}, {a: 2}, {a: 3}];
 *      R.differenceWith(cmp, a3, a2); //=> [{a: 2}]
 *      R.differenceWith(cpm, a3, []); //=> [{a: 2}, {a: 3}]
```

Or maybe `list123`, `list34` and `list223`:
```
 *      const cmp = (x, y) => x.a === y.a;
 *      const list123 = [{a: 1}, {a: 2}, {a: 3}];
 *      const list34 = [{a: 3}, {a: 4}];
 *      R.differenceWith(cmp, list123, list34); //=> [{a: 1}, {a: 2}]
 *
 *      const list223 = [{a: 2}, {a: 2}, {a: 3}];
 *      R.differenceWith(cmp, list223, list34); //=> [{a: 2}]
 *      R.differenceWith(cpm, list223, []); //=> [{a: 2}, {a: 3}]
```

CrossEye:
@kurtmilam: Sorry, I hadn't remembered that we had `l1`/`l2` anywhere in the docs.

I really like your last examples.  I think those show it off better than anything else we've tried.

You are probably right that this behavior was intended.  When I sit down to think of what I'd want, I can't come up with anything better.  I really don't remember the history of this function, but I think there's no reason to change it unless someone can do better.

CrossEye (approved):

PR #2897, opened 2019-09-26
Fix documentation for complement

rgeraldporter:
The documentation of `complement` leaves `isNil` in the example without a clear definition, and the REPL is picking out `isNil` from the scope of `R` (so it does work...). However, that seems unclear to someone reading the example without running the REPL as this var seems to come out of nowhere.


CrossEye:
Absolutely.  Thank you very much!

PR #2898, opened 2019-09-28
Fix a bug in `R.values` test file

mapkts:
It makes no sense to test for `R.keys` in `R.values` testfile. `R.keys` accidently passed the test clause, but it needs to fix for sure.

### DIFF
--- a/test/keys.js
+++ b/test/keys.js
@@ -20,10 +20,15 @@ describe('keys', function() {
   });
 
   it('works for primitives', function() {
-    var result = R.map(function(val) {
-      return R.keys(val);
-    }, [null, undefined, 55, '', true, false, NaN, Infinity, , []]);
-    eq(result, R.repeat([], 10));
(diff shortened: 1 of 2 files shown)


CrossEye:
You're right that this needs to be fixed.  But I wonder if it would be clearer altogether to skip the list and call each one explicitly:

```js
  it('returns an empty object for primitives', function() {
    eq(R.values(null), [])
    eq(R.values(undefined), [])
    eq(R.values(55), [])
    eq(R.values('foo'), [])
    eq(R.values(true), [])
    eq(R.values(false), [])
    eq(R.values(NaN), [])
    eq(R.values(Infinity, [])
    eq(R.values([]), [])
  });
```

mapkts:
I agree with you it's more easy to read when writing in this way. If we want to update this, it's better to update `R.keys` counterpart simultaneously.

CrossEye:
Perfect.  Thank you very much!

Issue #2895, opened 2019-09-25
Feature: allow logic functions to call functions and pass value to them


Santinell:
It can be useful to combine several logic functions.
For example to filter array:

```javascript
const arr = [3, 8, true, 's', 'he', undefined, null, undefined];
const result = R.filter(
  R.or(R.is(Number), R.is(Boolean))
)(arr); // => [3, 8, true]
```
Function `R.or` could looks like:
```javascript
R.or = R.curry(function or(a, b) {
  if (a instanceof Function && b instanceof Function) {
    return function (v) {
      return a(v) || b(v);
    }
  }
  return a || b;
});
```

CrossEye:
I'm assuming this was closed because you realized Ramda already has `or`, and also what you want here, `either`.  If not, well it does!  :smile:

Santinell:
Yes, you are right, I have found `both` for `and`, `either` for `or`and `complement` for 'not'.
But I want to ask 2 questions
1) Why we need 2 almost the same functions?
2) Why `both`, `either` and `complement` and use only 2 arguments?
My variant:

```
function or (...args) {
  return function () {
    return args.some(v => R.is(Function, v) ? v.apply(this, arguments) : v);
  };
}

function and (...args) {
  return function () {
    return args.every(v => R.is(Function, v) ? v.apply(this, arguments) : v);
  };
}

function not (...args) {
  return function () {
    return !args.some(v => R.is(Function, v) ? v.apply(this, arguments) : v);
  };
}
```

And now you can check multiple logic conditions
```
const arr = [8000, 1, 2, 4, 0, -12, -6, null, {}, [], [1], {pi: 3.14}]
R.filter(
  R.or(R.isNil, R.and(R.is(Number), R.gt(R.__, 0), R.lt(R.__, 100)), R.isEmpty)
)(arr) // => [1, 2, 4, null, {}, []]
```

BTW it's strange to see that `R.gt` and `R.lt` using first argument as value.

CrossEye:
@Santinell:

> Why we need 2 almost the same functions?

The most obvious ones, `and`, `or`, `not` are used for values:

```js
or (false, true) //=> true
```

The alternatives are used for functions that return values:

```js
const fn = either(x => x < 10, x => x > 20)
fn(7) //=> true
fn(13) //=> false
fn(29) //=> true
```

For the unary case, these could be implemented with `lift`, as `either = lift(or)`, `both = lift(and)`.  But these functions are polyadic.

[...]

PR #2900, opened 2019-10-02
Remove duplicate test cases

mapkts:


mapkts:
I'm a newbie to functional programming and recently I notice something weird to me when looking into Ramda source code. Some methods of Ramda seem not favour pointfree style, like gt, substract and concat. The argument order of concat I expected is concat(what, data), but Ramda's implementation is more likely to be translated as "a concat b". When using Ramda's concat in partially application or function pipeline, we write code like:
```javascript
const concatRamda = flip(concat('ramda'))
compose(toUpperCase, concatRamda)('enter ') //=> 'ENTER RAMDA'
```
In my opinion concat(what, data) is more nature and intutive. We does not need to introduce extra flip and it's more consistent with the remaining codebase. Does Ramda design these methods for a reason?

CrossEye:
I can't tell whether you think these tests should be removed because you think the API should change or if you see real duplication in the tests as your issue title implies.  If the latter, what do you see duplicated?

As to this:

> Does Ramda design these methods for a reason?

I'm just going to repost part of my answer from https://github.com/ramda/ramda/issues/2895#issuecomment-536237789:

>> BTW it's strange to see that R.gt and R.lt using first argument as value.

> You're asking about things that have a long history of controversy in Ramda.  I'm afraid I don't have the strength to look up the references, as I will start re-reading all our old arguments and tedious discussions about them.  But if you search for "operator", you will find at least a dozen issues related to this.
> 
> Briefly, there is a serious tension between, for instance,
> 
> ```
> R.subtract(10, 3) //=> clearly should be 7
> ```
> 
> and 
> 
> ```
> const less3 = R.subtract(3)
> 
> less3(10) //=> we want this to be 7, but it's -7.
> ```
> 
> This is true for `gt` as well, and for any binary operator where `op(a, b) !== op(b, a)`.  We have never really found a solution we liked for this.

mapkts:
Sorry I didn't provide enough imformation about this PR. There is real duplication in the test:
```javascript
  it('adds combines the elements of the two lists', function() {
    eq(R.concat(['a', 'b'], ['c', 'd']), ['a', 'b', 'c', 'd']);
    eq(R.concat([], ['c', 'd']), ['c', 'd']);
  });

  var z1 = {
    x: 'z1',
    concat: function(that) { return this.x + ' ' + that.x; }
  };
  var z2 = {
    x: 'z2'
  };

  it('adds combines the elements of the two lists', function() {
    eq(R.concat(['a', 'b'], ['c', 'd']), ['a', 'b', 'c', 'd']);
    eq(R.concat([], ['c', 'd']), ['c', 'd']);
  });
```
Now it's obvious I supposed.

CrossEye:
Weird, I swear I looked at the file and didn't see it.  I must have looked at the one from your commit...

Thank you.  This is clearly needed.

PR #2902, opened 2019-10-06
ramda#2890: Add isUndefined

### LINKED ISSUE #2890: Feature request: isUndefined
There is `isNil` but It would be useful to have `isUndefined` for those cases where you want to distinguish `null` and `undefined`.  (example use case: `reject(isUndefined, obj)` to remove undefined props from object).

TheHandsomeCoder:
If accepted this PR will add a new function to ramda to allow a user to check if a value is `undefined`

closes #2890

### DIFF
--- a/source/index.js
+++ b/source/index.js
@@ -97,6 +97,7 @@ export { default as invoker } from './invoker';
 export { default as is } from './is';
 export { default as isEmpty } from './isEmpty';
 export { default as isNil } from './isNil';
+export { default as isUndefined } from './isUndefined';
 export { default as join } from './join';
 export { default as juxt } from './juxt';
 export { default as keys } from './keys';
--- a/source/isUndefined.js
+++ b/source/isUndefined.js
@@ -1,0 +1,20 @@
+import _curry1 from './internal/_curry1';
+
+
+/**
+ * Checks if the input value is `undefined`.
+ *
+ * @func
+ * @memberOf R
+ * @category Type
+ * @sig * -> Boolean
+ * @param {*} x The value to test.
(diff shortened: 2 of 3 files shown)


Bradcomp:
Hi @TheHandsomeCoder, thanks for the PR. As mentioned in the issue bringing this up, I think this might be a bit too specific for Ramda, though I would be open to hearing your thoughts if you feel otherwise.

For reference, this function does exist in [Ramda Adjunct](https://char0n.github.io/ramda-adjunct/2.21.0/RA.html#.isUndefined)so it's there if there is an immediate need.

Bradcomp:
Closing due to lack of interest.

CrossEye:
Reopening because the discussion has reactivated.

I'm with @Bradcomp.  My first impression is that this is not widespread enough, and it's easy to write on your own.

Bradcomp:
I guess for me the question is why to add this function in particular? Is it better to add this than `isDefined` instead? If we add `isUndefined` should we add `isNull` as well?

nfantone:
@Bradcomp @CrossEye  My two (unrequested) cents:

`isUndefined` would just be `equals(undefined)` and doesn't look, in my opinion, like it would add much more value than saving 6 characters.

> Is it better to add this than `isDefined` instead? 

`isDefined` sounds kinda of vague. Something being "defined" might be tied to its semantic context. `isUndefined` is much more clear in its intent and has a direct JS correlation.

> If we add `isUndefined` should we add `isNull` as well?

Yes. If you don't, it'd be like an inside itch that you can't scratch.

CrossEye:
@nfantone: 

> My two (unrequested) cents

Unrequested, perhaps, but very welcome!  I agree on all points.

PR #2908, opened 2019-10-27
Make `R.split` work with arrays.

### LINKED ISSUE #2619: `R.split` should work on lists and not just strings
Does that seem reasonable? I am a little hesitant due to how weird strings are in JS with no underlying char class.

diasbruno:
Fixes #2619.

I wrote it with a `reduce` to allow other iterables to also work,
but it can be written to use the Iterable protocol.

Let me know if there is something to improve on this PR.

### DIFF
--- a/source/split.js
+++ b/source/split.js
@@ -1,18 +1,20 @@
+import _curry2 from './internal/_curry2';
 import invoker from './invoker';
+import is from './is';
+import equals from './equals';
 
 
 /**
- * Splits a string into an array of strings based on the given
- * separator.
+ * Splits a list or string based on the given separator.
  *
  * @func
  * @memberOf R
  * @since v0.1.0
- * @category String
- * @sig (String | RegExp) -> String -> [String]
- * @param {String|RegExp} sep The pattern.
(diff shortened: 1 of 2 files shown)


diasbruno:
If this is going works with other iterables, the signature would be something like `a -> f a -> [f a]`. (?)

CrossEye:
Just getting back to Ramda after some time away...

Could you provide some more context on *why* you want this?  Is it just that it seems a natural extension or did you have a real-world problem that would have been solved if only `split` also handled arrays?  How do you propose to handle the oddity that this would then accept a regex separator for strings, but not for arrays?

diasbruno:
It's natural when we consider strings a list of chars, than it would be nice to allow it to work on any kind of list. 

One problem with this implementation is the idea that we need to check before execute the correct implementation (the regex case included). Maybe, it would be better to let `split` unchanged, since it's more connected to the String interface and make a new function to work with array.

zacnomore:
I agree with separating this functionality from split, it would avoid confusion.

PR #2909, opened 2019-10-29
[chore] update the dist with latest version

abhi18av:
This PR updates the `ramda.js` and `ramda.min.js` bundled in the `dist`.


CrossEye:
Missed this at that time, but newer releases have changed the need for it.

PR #2914, opened 2019-11-11
make addIndex's signature more clear

adispring:


CrossEye:
Sure.  That's an improvement.  At some point I want to go through these variadic signatures and replace the ellipses with multiple versions, showing an ellipsis only between signatures.  But `addIndex` may be a particularly nasty one regardless.

PR #2916, opened 2019-11-13
Improve lift documentation to be more intuitive

gravitypersists:
By using different orders of magnitude, lift's combinatorial nature is a bit more clear


CrossEye:
Funny, I thought we'd done this long ago.

It's definitely worth doing.

PR #2919, opened 2019-11-18
change see of 'either' and 'both' to match type signature

ni-ko-o-kin:
Currently R.both uses R.and for "see also" but R.and accepts two booleans and both two functions. I propose to **change the @see from R.and to R.either**.

And the same goes for R.either:  **change the @see from R.or to R.both**.

### DIFF
--- a/source/and.js
+++ b/source/and.js
@@ -12,7 +12,7 @@ import _curry2 from './internal/_curry2';
  * @param {Any} a
  * @param {Any} b
  * @return {Any} the first argument if it is falsy, otherwise the second argument.
- * @see R.both, R.xor
+ * @see R.both, R.or
  * @example
  *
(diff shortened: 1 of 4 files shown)


CrossEye:
`or` and `both` are each related to `and` in different ways.  I think better would be if `and`, `both`, `or`, and `either` all pointed both horizontally and vertically in this table:

|                 | scalar | function |
|-----------------|--------|----------|
| **conjunction** | and    | both     |
| **disjunction** | or     | either   |

So in `and.js`:

```js
 * @see R.both, R.or
```

Etc.

CrossEye:
Thank you very much.  This is definitely an improvement.

ni-ko-o-kin:
No problem! I'm very happy I'm able to contribute a little bit to this wonderful library!

PR #2922, opened 2019-11-20
add @see: pluck, project and map

adispring:


CrossEye:
I think better than this would be if these four would refer either all to one another or at least all refer to their horizontal and vertical neighbors (so that, for instance,, `pluck` might refer to `prop` and `project` but possibly not to `props`):

|                         | single object | multiple objects |
|-------------------------|---------------|------------------|
| **single property**     | prop          | pluck            |
| **multiple properties** | props         | project          |

They're not entirely parallel, since they have different return types, but they make sense as a group that would all make reference to one another.

I don't think that `map` belongs in this group.  It does something else, even if it can be used to implement, say `pluck` in terms of `prop`.

adispring:
Modified @see by your suggestion, @CrossEye

CrossEye:
:herb:

Thank you very much!

Issue #2923, opened 2019-11-20
feat: whereAny


hitmands:
### Feature Request
Implement `whereAny` as opposite of `where`.

### Motivation
Ramda has a lot of complementary apis, such as `allPass` and `anyPass`, but `where` feels like it's missing its counterpart.
as stated in the doc:
> `where` returns true if all the predicates return true, false otherwise. 

but we don't have an api that, given a spec object, returns `true` if at least one of the predicates returns true.

### Use Case
Let us build a simple search of a list of users:
```typescript
const users = [
  { name: 'Giuseppe', surname: 'Mandato' },
  { name: 'Jimi', surname: 'Hendrix' }, 
  { name: 'Freddie', surname: 'Mercuri' }, 
];

const find = (value) => R.whereAny(
  name: R.equals(value),
  surname: R.equals(value), 
);

R.filter(find('Giuseppe'), users);
// [ { name: 'Giuseppe', surname: 'Mandato' } ]


R.filter(find('Charles'), users);
// []
```

### Alternatives
1. we could enhance the `anyPass` api to also take a spec object, but this would then be in contrast with `allPass`

### Notes
**I'd love to provide a pull request** to implement this api if it'd be of interest.

### Resources
 - [`where`](https://ramdajs.com/docs/#where)
 - [`allPass`](https://ramdajs.com/docs/#allPass)
 - [`anyPass`](https://ramdajs.com/docs/#anyPass)

CrossEye:
:+1:

This sounds like a good idea.  I agree that this is better as a separate function than as an extension to `anyPass`.  I'd love to see a PR.

Also, thank you very much for a well-written and clear feature request!

PR #2925, opened 2019-11-22
Feature/2923  implement whereAny

hitmands:
### Implements
- #2923 

### TS Declarations
-  https://github.com/types/npm-ramda/pull/443

### Tasks
- [ ] TS Declarations

### DIFF
--- a/source/whereAny.js
+++ b/source/whereAny.js
@@ -1,0 +1,46 @@
+import _curry2 from './internal/_curry2';
+import _has from './internal/_has';
+
+
+/**
+ * Takes a spec object and a test object; each of the spec's own properties must be a predicate function.
+ * Each predicate is applied to the value of the corresponding property of the
+ * test object. `whereAny` returns true if at least one of the predicates return true,
+ * false otherwise.
+ *
+ * `whereAny` is well suited to declaratively expressing constraints for other
+ * functions such as [`filter`](#filter) and [`find`](#find).
+ *
+ * @func
+ * @memberOf R
+ * @category Object
+ * @sig {String: (* -> Boolean)} -> {String: *} -> Boolean
+ * @param {Object} spec
+ * @param {Object} testObj
+ * @return {Boolean}
+ * @see R.propSatisfies, R.where
+ * @example
+ *
+ *      // pred :: Object -> Boolean
+ *      const pred = R.whereAny({
+ *        a: R.equals('foo'),
+ *        b: R.complement(R.equals('xxx')),
+ *        x: R.gt(R.__, 10),
+ *        y: R.lt(R.__, 20)
+ *      });
+ *
+ *      pred({a: 'foo', b: 'xxx', x: 8, y: 34}); //=> true
+ *      pred({a: 'xxx', b: 'xxx', x: 9, y: 21}); //=> false
+ *      pred({a: 'bar', b: 'xxx', x: 10, y: 20}); //=> false
+ *      pred({a: 'foo', b: 'bar', x: 10, y: 20}); //=> true
+ *      pred({a: 'foo', b: 'xxx', x: 11, y: 20}); //=> true
+ */
+var whereAny = _curry2(function whereAny(spec, testObj) {
+  for (var prop in spec) {
+    if (_has(prop, spec) && spec[prop](testObj[prop])) {
+      return true;
+    }
+  }
+  return false;
+});
+export default whereAny;
--- a/test/whereAny.js
+++ b/test/whereAny.js
@@ -1,0 +1,61 @@
+var R = require('../source');
+var eq = require('./shared/eq');
+
+
+describe('whereAny', function() {
+
+  it('returns true if the test object satisfies the spec', function() {
+    var spec = {x: R.equals(0), y: R.equals(2)};
+    var test1 = {x: 0, y: 200};
+    var test2 = {x: 0, y: 10};
+    var test3 = {x: 1, y: 101};
+    var test4 = {x: 1, y: 2};
+    eq(R.whereAny(spec, test1), true);
+    eq(R.whereAny(spec, test2), true);
+    eq(R.whereAny(spec, test3), false);
+    eq(R.whereAny(spec, test4), true);
+  });
+
+  it('does not need the spec and the test object to have the same interface (the test object will have a superset of the specs properties)', function() {
+    var spec = {x: R.equals(100)};
+    var test1 = {x: 20, y: 100, z: 100};
+    var test2 = {w: 1, x: 100, y: 100, z: 100};
+
+    eq(R.whereAny(spec, test1), false);
+    eq(R.whereAny(spec, test2), true);
+  });
+
+  it('matches specs that have undefined properties', function() {
+    var spec = {x: R.equals(undefined)};
+    var test1 = {};
+    var test2 = {x: null};
+    var test3 = {x: undefined};
+    var test4 = {x: 1};
+    eq(R.whereAny(spec, test1), true);
+    eq(R.whereAny(spec, test2), false);
+    eq(R.whereAny(spec, test3), true);
+    eq(R.whereAny(spec, test4), false);
+  });
+
+  it('is false for an empty spec', function() {
+    eq(R.whereAny({}, {a: 1}), false);
+  });
+
+  it('matches inherited properties', function() {
+    var spec = {
+      toString: R.equals(Object.prototype.toString),
+      valueOf: R.equals(null)
+    };
+    eq(R.whereAny(spec, {}), true);
+  });
+
+  it('does not match inherited spec', function() {
+    function Spec() { this.y = R.equals(6); }
+    Spec.prototype.x = R.equals(5);
+    var spec = new Spec();
+
+    eq(R.whereAny(spec, {y: 6}), true);
+    eq(R.whereAny(spec, {x: 5}), false);
+  });
+
+});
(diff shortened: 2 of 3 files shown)


source/whereAny.js
```diff
@@ -0,0 +1,52 @@
...
+ * Each predicate is applied to the value of the corresponding property of the
+ * test object. `whereAny` returns true if at least one of the predicates return true, false
+ * otherwise.
+ *
+ * `whereAny` is well suited to declaratively expressing constraints for other
+ * functions such as [`filter`](#filter) and [`find`](#find).
+ *
+ * @func
+ * @memberOf R
+ * @since v0.27.0
```
hitmands: To be defined

source/whereAny.js
```diff
@@ -0,0 +1,52 @@
...
+ *        y: R.lt(R.__, 20)
+ *      });
+ *
+ *      pred({a: 'foo', b: 'xxx', x: 8, y: 34}); //=> true
+ *      pred({a: 'xxx', b: 'xxx', x: 9, y: 21}); //=> false
+ *      pred({a: 'bar', b: 'xxx', x: 10, y: 20}); //=> false
+ *      pred({a: 'foo', b: 'bar', x: 10, y: 20}); //=> true
+ *      pred({a: 'foo', b: 'xxx', x: 11, y: 20}); //=> true
+ */
+var whereAny = _curry2(function where(spec, testObj) {
```
hitmands: `whereAny` probably needs a better naming

source/whereAny.js
```diff
@@ -0,0 +1,52 @@
+import _curry2 from './internal/_curry2';
+import _has from './internal/_has';
+import _isEmpty from './isEmpty';
```
hitmands: there is no internal for `isEmpty`

source/whereAny.js
CrossEye: Yes, the build process should add this automatically, so let's take it out.

source/whereAny.js
CrossEye: I'm not a big fan of the name either, but I haven't come up with a good alternative.  Any suggestions?

source/whereAny.js
```diff
@@ -0,0 +1,52 @@
...
+ *
+ *      pred({a: 'foo', b: 'xxx', x: 8, y: 34}); //=> true
+ *      pred({a: 'xxx', b: 'xxx', x: 9, y: 21}); //=> false
+ *      pred({a: 'bar', b: 'xxx', x: 10, y: 20}); //=> false
+ *      pred({a: 'foo', b: 'bar', x: 10, y: 20}); //=> true
+ *      pred({a: 'foo', b: 'xxx', x: 11, y: 20}); //=> true
+ */
+var whereAny = _curry2(function where(spec, testObj) {
+  if (_isEmpty(spec)) {
+    return true;
```
CrossEye: I don't think so.  This is a disjunction, which means that the vacuous case should default to `false`, no?

That is, where `all` could be written as `reduce(and, true)`, this is more like `any`, which is `reduce(or, false)`.  It asks the question "Do any of these properties hold?"  If there are no properties, it's false.

test/whereAny.js
```diff
@@ -0,0 +1,61 @@
...
+    var test4 = {x: 1};
+    eq(R.whereAny(spec, test1), true);
+    eq(R.whereAny(spec, test2), false);
+    eq(R.whereAny(spec, test3), true);
+    eq(R.whereAny(spec, test4), false);
+  });
+
+  it('is true for an empty spec', function() {
+    eq(R.whereAny({}, {a: 1}), true);
+  });
```
CrossEye: No, see above.

source/whereAny.js
hitmands: - [ ] rely on automatic versioning

source/whereAny.js
hitmands: nope... 

`some` as the [ES `Array#some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) ?

source/whereAny.js
hitmands: @CrossEye  any comment on this? **is it okay to import from non-internal**?

source/whereAny.js
hitmands: i can agree with this, makes sense...

I took this case from the [`where spec`](https://github.com/ramda/ramda/blob/a52431a9d24f272d220fba8daf2d03c74248c474/test/where.js#L40:42)

Happy to change it

test/whereAny.js
hitmands: replied

source/whereAny.js
CrossEye: I'm sure we do so in other places.  We have the internals mostly for when the auto-currying adds nothing but could conceivably slow things down.

source/whereAny.js
CrossEye: Sounds good.

source/whereAny.js
CrossEye: I'd prefer `whereAny` to just `some`, and `whereSome` doesn't do anything for me either.

Uggh!  Naming is always hard.

hitmands: - [ ] use `some`

source/whereAny.js
hitmands: can be resolved, `isEmpty` is no longer being used.

source/whereAny.js
hitmands: let's see if someone helps to clear this out :), other comments have been addressed.


- [ ] define naming

hitmands: @CrossEye any news here? 
Should we keep `whereAny`?

CrossEye: I haven't heard any other good candidates.  Let's go with it.

.gitignore
```diff
@@ -10,3 +10,4 @@
 /dist/gh-pages/
 /tmp-test-bundle.js
 /coverage/
+.idea
```
Bradcomp: I don't think this needs to be in the commit

source/whereAny.js
```diff
@@ -0,0 +1,47 @@
+import _curry2 from './internal/_curry2';
+import _has from './internal/_has';
+
+
+/**
+ * Takes a spec object and a test object; returns true if the test satisfies
+ * the spec. Each of the spec's own properties must be a predicate function.
```
Bradcomp: I think everything in the first sentence after the semicolon needs to go. It's misleading to say that the test object has to satisfy the spec. Also, it looks like the docs for `where` have a typo in that same spot (test -> test object)

source/whereAny.js
```diff
@@ -0,0 +1,47 @@
...
+ *        y: R.lt(R.__, 20)
+ *      });
+ *
+ *      pred({a: 'foo', b: 'xxx', x: 8, y: 34}); //=> true
+ *      pred({a: 'xxx', b: 'xxx', x: 9, y: 21}); //=> false
+ *      pred({a: 'bar', b: 'xxx', x: 10, y: 20}); //=> false
+ *      pred({a: 'foo', b: 'bar', x: 10, y: 20}); //=> true
+ *      pred({a: 'foo', b: 'xxx', x: 11, y: 20}); //=> true
+ */
+var whereAny = _curry2(function where(spec, testObj) {
```
Bradcomp: The function name is still `where` inside the call to `curry2`

hitmands: - [ ] rename function's name

source/whereAny.js
hitmands: @CrossEye what do you think here?

.gitignore
hitmands: it is a separate commit though, 
I could drop it eventually...

source/whereAny.js
hitmands: - [ ] use whereAny

.gitignore
hitmands: - [ ] drop .gitignore updates

source/whereAny.js
hitmands: - [  ] remove `'returns true if the test satisfies the spec.'`

hitmands: - [  ] remove 'returns true if the test satisfies the spec.'

hitmands:
@Bradcomp @CrossEye 

A quick follow up, 
are we ready to merge this?

source/whereAny.js
```diff
@@ -0,0 +1,47 @@
+import _curry2 from './internal/_curry2';
+import _has from './internal/_has';
+
+
+/**
+ * Takes a spec object and a test object;
+ * Each of the spec's own properties must be a predicate function.
```
Bradcomp: Because the prior phrase ends in a semicolon, `Each` should be `each`. We can also move it up onto the same line, breaking in the typical spot.

Alternately, we could use a period, or rewrite the sentence, but I think that changing E to e is easiest.

Bradcomp:
I made one tiny comment, but other than that I am :bowtie:

source/whereAny.js
hitmands: - [ ] fix documentation

hitmands: @Bradcomp :) we should be good to go!! please, take a look at my latest changes :)

hitmands:
@CrossEye @Bradcomp going to merge this in **two days** if no further comments.

Next follow up on Wed 10 Dec 2019

CrossEye:
:herb:!  Thanks, @hitmands!

hitmands:
```typescript
const gentleReminder = R.unless(
  furtherActivity, 
  mergeBy(endOfToday),
);

await publish(gentleReminder);
```

Bradcomp:
Thanks @hitmands! :bowtie:

Issue #2906, opened 2019-10-23
isVowel is not defined in repl for groupWith documentation


vasinkevych:
so correct I suppose we have to add isVowel check here https://github.com/ramda/ramda/blob/v0.26.1/source/groupWith.js#L30
the correct example should be:
```javascript
R.groupWith(R.equals, [0, 1, 1, 2, 3, 5, 8, 13, 21])
//=> [[0], [1, 1], [2], [3], [5], [8], [13], [21]]

R.groupWith((a, b) => a + 1 === b, [0, 1, 1, 2, 3, 5, 8, 13, 21])
//=> [[0, 1], [1, 2, 3], [5], [8], [13], [21]]

R.groupWith((a, b) => a % 2 === b % 2, [0, 1, 1, 2, 3, 5, 8, 13, 21])
//=> [[0], [1, 1], [2], [3, 5], [8], [13, 21]]

const isVowel =(s)=> (/^[aeiou]$/i).test(s);
R.groupWith(R.eqBy(isVowel), 'aestiou')
//=> ['ae', 'st', 'iou']
```

Bradcomp:
Hi @vasinkevych,

Thanks for pointing this out! Would you be interested in submitting a PR?

We may want to use our test function for isVowel:

`R.test(/[aeiou]/i)`

vasinkevych:
@Bradcomp sure, I'd glad to help

PR #2927, opened 2019-11-24
isVowel is not defined in groupsWith

vasinkevych:
https://github.com/ramda/ramda/issues/2906


CrossEye:
Thank you very much @vasinkevych!

Issue #2937, opened 2019-12-19
Should 'pathEq/propEq' and 'pathSatisfies/propSatisfies' parameters' order be consistent


adispring:
`pathEq`'s path and `propEq`'s prop is the first parameter, but `pathSatisfies` and `propSatisfies` is the second parameter, should their parameters' order be the same order?

CrossEye:
Probably.  I often use `propEq` as my demonstration that although it's usually clear enough what order to put parameters in, once in a while, it simply seems to be an arbitrary decision.  Here these related functions (including `propIs`) seem pretty clear: the predicate/type/value is the thing least likely to change and so should go first.  We should fix `propEq` to match.

Are you interested in writing up a PR for this?


This also shows that we seem to be missing one entry:

|               | Single        |  Multiple     |
|---------------|---------------|---------------|
| **equals**    | propEq        | pathEq        |
| **predicate** | propSatisfies | pathSatisfies |
| **type**      | propIs        |      ??       | 

(Or perhaps that we have one too many.  I don't know if I've *ever* used `propIs` myself.)

JamesRamm:
Just came across this change in the v0.29.0 release. IMO, the predicate/type/value is the *most* likely thing to change and follows the 'data last' philosophy of ramda. 

Here is an example usage of `propEq` for us:
```js
const users = [{ name: 'John',  team: '' }, { name: 'Pete', }, { name: 'Olivia'}, {name: 'Violette'}]
const someRecord = { author: 'Olivia', someProp: 'blah' ;
const findByName = compose(find, propEq('name'));

findByName(someRecord.author)(users);
```

Our usage of `propEq` and `pathEq` is 90% of time geared towards this kind of stuff - filtering lists and such. It is *always* the value that changes, not the property. 

Do you have any other examples or justification of putting the value first?

semmel:
> Just came across this change in the v0.29.0 release. IMO, the predicate/type/value is the most likely thing to change and follows the 'data last' philosophy of ramda.

Yes @JamesRamm, post v0.29 there is some resentment in the air, about this change of parameter order. See also the [recent comments](https://github.com/ramda/ramda/pull/2938#issuecomment-1521946369) in the PR thread.

Gastonite:
It was not a good idea IMO =(

because :
1. It's way more easier to read `propEq('hair', 'brown')` than `propEq('brown', 'hair')`
2. a function `propEq('hair')` is more useful than `propEq('brown')`
[...]

Issue #2428, opened 2018-01-05
R.propIs() and R.propEq() have inconsistent argument order


helios1138:
- `R.propIs()` and `R.propEq()` are named similarly enough that it's expected that they would place the prop name argument in the same place
- but `R.propIs()` takes prop name as the last argument while `R.propEq()` takes it as the last one
- ideally, `R.propIs()` should reverse it's argument order or reverse its name to something like `R.isProp()` or vice versa
- I think these functions are not very frequently used with partial application, they usually make sense as part of some conditional statements where readability is important
- same problem with `R.propOr()` I guess

CrossEye:
It's a good question.  I believe `propOr`, which doesn't in general return a boolean, is a different case.  But `propSatisfies` should also be discussed.

I remember waffling on the argument order for `propEq`.  For most functions, it's simply obvious what order is best for currying.  That one wasn't.  When we later added `propIs` and `propSatisfies`, the order for those seemed obvious again:

```js
propIs(Number) ~~> name => obj => is(Number, prop(name, obj))
propSatisfies(isOdd) ~~> name => obj => idOdd(prop(name, obj))
```

These seem so obviously right that I'm loathe to change them.  `propEq` doesn't have as clear an answer to me.  But if we were to switch it to

```js
propEq(42) ~~> name => obj => equals(42, prop(name, obj))
```

we would certainly add some consistency.  I think it would be worth hearing from others whether they think this consistency is worth breaking backward compatibility.  What do you say @ramda/core?

TheLudd:
`propEq` is one of out most used functions from ramda and its argument order is just the way it needs to be.

```
const hasEmail = propEq('email')
const isMe = hasEmail('ludwig@mediatool.com')
const me = find(isMe, userArray)
```

The order of `propEq` was actually changed in one release and then changed back because of many complaints.

The prop you are checking is the most static argument, the value to compare against is the second most static. Often the second value will be based on user input or database values but sometimes they are constants. And the object you compare against is almost always user input.

For some function, argument order can be tricky. I *really* do not think this is in the case for `propEq` but in any case, to me the argument for a specific order should never be "there is a similarly named function, lets copy the order from that one".

CrossEye:
@TheLudd:

The reverse seems equally likely in my experience:

```js
const uninitialized = flip(propEq)(null)
const noName = uninitialized('name')

noName({age: 10, name: null}) //=> true
```

[...]

PR #2938, opened 2019-12-20
change propEq/pathEq parameters order

adispring:
Make `propEq/pathEq` parameters order be consistent to `propSatisfies/pathSatisfies` parameters order.

related to #2937

### DIFF
--- a/source/pathEq.js
+++ b/source/pathEq.js
@@ -12,22 +12,23 @@ import path from './path.js';
  * @since v0.7.0
  * @category Relation
  * @typedefn Idx = String | Int | Symbol
- * @sig [Idx] -> a -> {a} -> Boolean
- * @param {Array} path The path of the nested property to use
+ * @sig a -> [Idx] -> {a} -> Boolean
  * @param {*} val The value to compare the nested property with
+ * @param {Array} path The path of the nested property to use
  * @param {Object} obj The object to check the nested property in
  * @return {Boolean} `true` if the value equals the nested object property,
  *         `false` otherwise.
+ * @see R.whereEq, R.propEq, R.pathSatisfies, R.equals
  * @example
  *
  *      const user1 = { address: { zipCode: 90210 } };
  *      const user2 = { address: { zipCode: 55555 } };
  *      const user3 = { name: 'Bob' };
  *      const users = [ user1, user2, user3 ];
- *      const isFamous = R.pathEq(['address', 'zipCode'], 90210);
+ *      const isFamous = R.pathEq(90210, ['address', 'zipCode']);
  *      R.filter(isFamous, users); //=> [ user1 ]
  */
-var pathEq = _curry3(function pathEq(_path, val, obj) {
+var pathEq = _curry3(function pathEq(val, _path, obj) {
   return equals(path(_path, obj), val);
 });
 export default pathEq;
--- a/source/propEq.js
+++ b/source/propEq.js
@@ -6,18 +6,20 @@ import equals from './equals.js';
 /**
  * Returns `true` if the specified object property is equal, in
  * [`R.equals`](#equals) terms, to the given value; `false` otherwise.
- * You can test multiple properties with [`R.whereEq`](#whereEq).
+ * You can test multiple properties with [`R.whereEq`](#whereEq),
+ * and test nested path property with [`R.pathEq`](#pathEq).
  *
  * @func
  * @memberOf R
  * @since v0.1.0
  * @category Relation
- * @sig String -> a -> Object -> Boolean
- * @param {String} name
- * @param {*} val
- * @param {*} obj
- * @return {Boolean}
- * @see R.whereEq, R.propSatisfies, R.equals
+ * @sig a -> String -> Object -> Boolean
+ * @param {*} val The value to compare the property with
+ * @param {String} name the specified object property's key
+ * @param {*} obj The object to check the property in
+ * @return {Boolean} `true` if the value equals the specified object property,
+ *         `false` otherwise.
+ * @see R.whereEq, R.pathEq, R.propSatisfies, R.equals
  * @example
  *
  *      const abby = {name: 'Abby', age: 7, hair: 'blond'};
@@ -25,10 +27,10 @@ import equals from './equals.js';
  *      const rusty = {name: 'Rusty', age: 10, hair: 'brown'};
  *      const alois = {name: 'Alois', age: 15, disposition: 'surly'};
  *      const kids = [abby, fred, rusty, alois];
- *      const hasBrownHair = R.propEq('hair', 'brown');
+ *      const hasBrownHair = R.propEq('brown', 'hair');
  *      R.filter(hasBrownHair, kids); //=> [fred, rusty]
  */
-var propEq = _curry3(function propEq(name, val, obj) {
+var propEq = _curry3(function propEq(val, name, obj) {
   return equals(val, prop(name, obj));
 });
 export default propEq;
--- a/test/pathEq.js
+++ b/test/pathEq.js
@@ -14,23 +14,23 @@ describe('pathEq', function() {
   };
 
   it('returns true if the path matches the value', function() {
-    eq(R.pathEq(['a'], 1, obj), true);
-    eq(R.pathEq(['b', 1, 'ba'], 3, obj), true);
+    eq(R.pathEq(1, ['a'], obj), true);
+    eq(R.pathEq(3, ['b', 1, 'ba'], obj), true);
   });
 
   it('returns false for non matches', function() {
-    eq(R.pathEq(['a'], '1', obj), false);
-    eq(R.pathEq(['b', 0, 'ba'], 3, obj), false);
+    eq(R.pathEq('1', ['a'], obj), false);
+    eq(R.pathEq(3, ['b', 0, 'ba'], obj), false);
   });
 
   it('returns false for non existing values', function() {
-    eq(R.pathEq(['c'], 'foo', obj), false);
-    eq(R.pathEq(['c', 'd'], 'foo', obj), false);
+    eq(R.pathEq('foo', ['c'], obj), false);
+    eq(R.pathEq('foo', ['c', 'd'], obj), false);
   });
 
   it('accepts empty path', function() {
-    eq(R.pathEq([], 42, {a: 1, b: 2}), false);
-    eq(R.pathEq([], obj, obj), true);
+    eq(R.pathEq(42, [], {a: 1, b: 2}), false);
+    eq(R.pathEq(obj, [], obj), true);
   });
 
   it('has R.equals semantics', function() {
@@ -39,10 +39,10 @@ describe('pathEq', function() {
       return x instanceof Just && R.equals(x.value, this.value);
     };
 
-    eq(R.pathEq(['value'], 0, {value: -0}), false);
-    eq(R.pathEq(['value'], -0, {value: 0}), false);
-    eq(R.pathEq(['value'], NaN, {value: NaN}), true);
-    eq(R.pathEq(['value'], new Just([42]), {value: new Just([42])}), true);
+    eq(R.pathEq(0, ['value'], {value: -0}), false);
+    eq(R.pathEq(-0, ['value'], {value: 0}), false);
+    eq(R.pathEq(NaN, ['value'], {value: NaN}), true);
+    eq(R.pathEq(new Just([42]), ['value'], {value: new Just([42])}), true);
   });
 
 });
--- a/test/propEq.js
+++ b/test/propEq.js
@@ -7,18 +7,18 @@ describe('propEq', function() {
   var obj2 = {name: 'Fred', age: 12, hair: 'brown'};
 
   it('determines whether a particular property matches a given value for a specific object', function() {
-    eq(R.propEq('name', 'Abby', obj1), true);
-    eq(R.propEq('hair', 'brown', obj2), true);
-    eq(R.propEq('hair', 'blond', obj2), false);
+    eq(R.propEq('Abby', 'name', obj1), true);
+    eq(R.propEq('brown', 'hair', obj2), true);
+    eq(R.propEq('blond', 'hair', obj2), false);
   });
 
   it('handles number as property', function() {
     var deities = ['Cthulhu', 'Dagon', 'Yog-Sothoth'];
-    eq(R.propEq(0, 'Cthulhu', deities), true);
-    eq(R.propEq(1, 'Dagon', deities), true);
-    eq(R.propEq(2, 'Yog-Sothoth', deities), true);
-    eq(R.propEq(-1, 'Yog-Sothoth', deities), true);
-    eq(R.propEq(3, undefined, deities), true);
+    eq(R.propEq('Cthulhu', 0, deities), true);
+    eq(R.propEq('Dagon', 1, deities), true);
+    eq(R.propEq('Yog-Sothoth', 2, deities), true);
+    eq(R.propEq('Yog-Sothoth', -1, deities), true);
+    eq(R.propEq(undefined, 3, deities), true);
   });
 
   it('has R.equals semantics', function() {
@@ -27,15 +27,15 @@ describe('propEq', function() {
       return x instanceof Just && R.equals(x.value, this.value);
     };
 
-    eq(R.propEq('value', 0, {value: -0}), false);
-    eq(R.propEq('value', -0, {value: 0}), false);
-    eq(R.propEq('value', NaN, {value: NaN}), true);
-    eq(R.propEq('value', new Just([42]), {value: new Just([42])}), true);
+    eq(R.propEq(0, 'value', {value: -0}), false);
+    eq(R.propEq(-0, 'value', {value: 0}), false);
+    eq(R.propEq(NaN, 'value', {value: NaN}), true);
+    eq(R.propEq(new Just([42]), 'value', {value: new Just([42])}), true);
   });
 
   it('returns false if called with a null or undefined object', function() {
-    eq(R.propEq('name', 'Abby', null), false);
-    eq(R.propEq('name', 'Abby', undefined), false);
+    eq(R.propEq('Abby', 'name', null), false);
+    eq(R.propEq('Abby', 'name', undefined), false);
   });
 
 });


CrossEye:
Thank you very much.  I think this is a very positive change, but I'd like to leave it open for a bit to hear back from others.  Any objections, @ramda/core?

Bradcomp:
This is a big breaking change. I think it's good but I think it's not fair to make this without bumping to 1.0.

Bradcomp:
@CrossEye What do you think about using this one as an impetus to get to 1.0?

CrossEye:
@Bradcomp: That sounds like a very good idea.

lambert-velir:
Any news on this? I just spent 15 minutes debugging why this didn't work: 

```javascript
R.propSatisfies("name", R.test(queryRegex))
```

CrossEye:
This is one of the few open issues that I think might be worth a breaking change in `v1.0` -- or possibly a short-lived `v0.29`.

customcommander:
> This is one of the few open issues that I think might be worth a breaking change in `v1.0` -- or possibly a short-lived `v0.29`.

My personal preference would be to stop releasing breaking changes in minor v0 releases especially if we're committed to release v1.0 as soon as possible. As much as I prefer having a non-breaking v1.0, I think these changes bring some consistency and v1.0 is probably the best time to introduce such breaking change.

I'm in favour of this change but would prefer to release it as part of v1.0

customcommander:
@adispring I took the liberty to update your branch:

1. Rebased against origin/master to fix a conflict
2. Fixed a failing test (somehow one test for `propEq` wasn't updated to reflect the change of parameter position)

@CrossEye I think this change was welcomed by pretty much everybody in the team. Could we merge this now and ship it in v1.0? I already added that milestone :)

customcommander:
@adispring I took the liberty to update your branch:

1. Rebased against origin/master to fix a conflict
2. Fixed a failing test (somehow one test for `propEq` wasn't updated to reflect the change of parameter position)

@CrossEye I think this change was welcomed by pretty much everybody in the team. Could we merge this now and ship it in v1.0? I already added that milestone :)

adispring:
Is it more natural to keep current `propEq`/`pathEq`'s parameter order? And instead, we can change `propSatisfies` and `pathSatisfies`'s parameter order.

customcommander:
> Is it more natural to keep current `propEq`/`pathEq`'s parameter order? And instead, we can change `propSatisfies` and `pathSatisfies`'s parameter order.

I think your change makes more sense as it focus on the value first as opposed to where the value is:

```javascript
var isJohn = propEq('john');

isJohn('name', {name: 'john'});
isJohn('first_name', {first_name: 'john'});
```

With the current parameters order you would have to flip the function first.

adispring:
> I think your change makes more sense as it focus on the value first as opposed to where the value is:

OK.

customcommander (approved):

semmel:
@customcommander wrote
> I think your change makes more sense as it focus on the value first as opposed to where the value is...

How to order the arguments is not so clear cut as your comment suggests. Is not 
> put the least changing arguments first

the principle of Ramda? 

With that principle, the "focus" should sharpen with the position of the argument – so the most general would be first. 
What that is, may depend on the particular context. E.g. I use more often the flipped version of `contains` to determine that a value is contained in a known set (enumeration). In that case the enumeration is clearly the least-changing argument and should come first. However, it would be perhaps unpractical for Ramda to introduce it's functions in all permutations. For the there is `flip` and `__`.
 
If this change comes through, we may eliminate the inconsistency with `propSatisfies`, but how many other will be introduced?
E.g. `assoc`, `adjust` and `update` are authored in the way that least-changing is *where* the value is.

customcommander:
I don't think this is necessarily going against the philosophy of putting data last. We simply have different opinions about what the least-changing parameters are. 

For me `propEq(42)` is the same as `propSatisfies(equals(42))` as I consider `42` as a form of predicate. So in my view the parameters order should be consistent.

You make reasonable observations (as always) but we don't need to put everything into the same bag either. I don't see `propEq` and the like as related to functions such as `assoc` (just my opinion here).

This PR has been opened for more than two years now so we ought to make a decision at some point. Do you feel strongly against this change?

hitmands (approved):

semmel:
> This PR has been opened for more than two years now so we ought to make a decision at some point. Do you feel strongly against this change?

No, I have anticipated (feared) this breaking change might come eventually and moved away from `propEq`. The trivial task it solves is not worth introducing potential bugs. Instead I write `({key}) =>  key === value`.

I was just asking if someone else had a consistent conception about the order of parameters in Object-accepting functions of Ramda. This PR might be the right place to ask I thought.

CrossEye:
> I was just asking if someone else had a consistent conception about the order of parameters in Object-accepting functions of Ramda. This PR might be the right place to ask I thought.

I don't think there is any universal answer to these questions.  For the majority of Ramda functions the order feels immediate, and obviously correct.  But some, like `includes` and `propEq` seem to have equal weight, and no wonderful answer.  Here's where I really miss Haskell's left- and right-sections.

lidorcg:
Hi, this is a change in the API of a popular function in Ramda, shouldn't it call for an update to the major version?

whitelizard:
I think this change was a huge misstake and a horrible breaking change.

**But the misstake was made long ago!**

To explain, I'm thinking of the process of adding functions to an API.
_After a function need ("step 0") is identified_:

**Step 1. Signature**
Here, the often repeated "put the least changing / most general arguments first", also called the "data-last principle", is to be used. Worth noting is that a function would never be considered "data".

**Step 2. Naming**
Here comes the hard part.
- If the function is a general & commonly understood concept, like "map", "filter", "reduce", use that name. Or perhaps another general and "fluffy" word would fit (ex: "converge", "scan"), great -- one should rely on the documentation.
- If the name is to be invented and **will somehow include or reference any of the arguments, the references must come in the opposite order as the signature**. This is because of currying and readability (and consistency with the absolute majority of the ramda library). (Ex: `R.minBy(square)`, `R.mergeWith(R.concat)`, `R.has("name")`)

---

**Example 1: A function that checks a predicate for an object prop**
Step 1: 
(a → Boolean) → String → {String: a} → Boolean
Note, as mentioned above, a function is to be considered highly general and "far away" from being "data". Therefor the predicate is best put first.

Step 2:
There is no commonly known concept to apply here as a name, but `propSatisfies` would be great because the last word references the first argument, and the first word references the second! **So the meaning of the "sentence" _points_ to the beginning of the argument list.**

**Example 2: A function that checks an object prop aganst a value**
Step 1: 
String → a → Object → Boolean
`a` is a value closer to being "data" than the name of the prop, therefor it should come later in the argument list.

Step 2:
This is an equality check, and it references a prop in an object. But if our name is to include argument references, the order is important. `valueAtProp` has the right order of arg references so it's a great candidate (but perhaps is missing the word "equals"/"eq")? `eqAtProp` or even `eqProp` then? `propEq` is **not** a good candidate however, because it implies that the first argument should be the equality check somehow. 

---

If the current `propEq` should have been named correctly, there would be no inconsistency with `propSatisfies`. The signature being the most logical and user friendly should come before the naming. The issue here was therefor the name and not the (great) signature of `propEq`. **The question is: What is a good name for the needed signature?**

This has been pointed out before, mainly by "TheLudd" and "helios1138" [here](https://github.com/ramda/ramda/issues/2428), and I also read there that **"The order of propEq was actually changed in one release and then changed back because of many complaints."**. Perhaps a repeat on that one ;P.

Is it time to correct this the right way, for version 1.0? :)

semmel:
@whitelizard Thanks for working out the concept of naming functions. It is easy to remember and brings consistency to the vast API surface. 

Your Steps 1 and 2 should 
- Go prominently in the Readme. It they were, it could have reduced the many times I had to visit the documentation to look up a function signature,
- Become the criterion for adding new functions

I also agree that naming `propEq`  badly in the first place, brought us into this mess here.

I briefly checked the list of Ramda functions, if they comply with your Steps 1 and 2. I came up with *just a few* mismatches!

|existing fn| signature | consistent name |
|-----------|-----------|------------|
| allPass   | `[(*… → Boolean)] → *… → Boolean` | passesAll   |
| anyPass  | `[(*… → Boolean)] → *… → Boolean` | passesAny   |
| nthArg   | `Number → *… → *` | argNth |
|          | `[Idx] → a → {a} → Boolean`  | eqPath |
|          | `Idx → a → {a} → Boolean`  | eqProp |

KurtMar:
This was quite the unexpected breaking change for a minor version.

yuercl:
> This was quite the unexpected breaking change for a minor version.
> 
> While I understand the benefit, unlike with removed or renamed functions the change from e.g. `R.propEq('id', '123')` to `R.propEq('123', 'id')` will cause many hard to find bugs in non-typescript codebases.
> 
> It would be fun if people would comment here the amount of time they spent tracking the bug before finding the culprit and the amount of time they spent refactoring their codebases. :)

i spend 3 hours in my case today morning, and will spend 4 hours afternoon to refactor my code. 😭

semmel:
@yuercl Don't forget that `R.of` has also changed.

A full text search 'n replace over all my code seems inevitable. That's why, I am staying on v0.28 where I can.

crshmk:
I often want to 
```javascript 
const idEq = propEq('id')
```

I never want to 
```javascript 
const somePropEqualsOne = propEq(1)
```

Lambdaphile:
@crshmk Is there a workaround for this?

crshmk:
> @crshmk Is there a workaround for this?

@Lambdaphile  the easiest solution is to use version `0.28`
```json
"dependencies": {
  "ramda": "=0.28.0"
}
```

However, if you use ramda in many different apps, managing two different incompatible versions across them is unwieldy at best, and impossible if they share code. If you need a current version, I suggest creating your own `propEq` locally.


```javascript 
// propEq.js, kept in local utils
import { curry, equals, prop } from 'ramda'

export const propEq = curry((name, val, obj) => 
  equals(val, prop(name, obj)))

```


```javascript 
// app.js
import { filter } from 'ramda'
import { propEq } from './utils'
```

I have my own everyday ramda [utils](https://github.com/crshmk/utils) I've built over the years, so I just [added](https://github.com/crshmk/utils/commit/7a9aca0e7e1b9c8f5db0cf487afc7cc08163acba) it to those. In this way, all my apps can use version 0.28 or 0.29 of ramda as long as they grab the current version of my utils and import `propEq` from there (as above). 

Finally, respectfully, the best solution would be to admit that we already [litigated](https://github.com/ramda/ramda/pull/1286) this and to revert, preferably quickly.

Gastonite:
It was not a good idea IMO =(

because :
1. It's way more easier to read `propEq('hair', 'brown')` than `propEq('brown', 'hair')`
2. a function `propEq('hair')` is more useful than `propEq('brown')`
3. It creates hard to find bugs
4. It's **just** a cosmetic change that has no other real motivations

Sometime it's better to simply rename a function instead of changing its signature

Pyrolistical:
While I agree with this change, these kinds of subtle breaking changes is what will prevent large adoption of this library. I wish this project to immediately release 1.0 and follow strict semver. Breaking changes are fine. Just make a new major release. That is what that is for!

whitelizard:
> While I agree with this change, these kinds of subtle breaking changes is what will prevent large adoption of this library. I wish this project to immediately release 1.0 and follow strict semver. Breaking changes are fine. Just make a new major release. That is what that is for!

I agree. Every lib should actually always start with v1.0.0, from the very beginning, for the npm semver system to work well. I recommend to read e.g. this article on this subject: https://medium.com/javascript-scene/software-versions-are-broken-3d2dc0da0783

Harris-Miller:
First, as a reminder for everyone, per the rules of semver, for **major version zero**, minor bumps should be considered breaking. Ref: https://semver.org/#spec-item-4

That being said, IMHO `ramda` should have reached `1.0.0` long ago. Regardless of function names and signatures, what `ramda` offers as an API and how each function works are very stable. I personally don't see a reason to keep it as `0.x`. That would have alleviated the main issue of the "breaking" expectation

To address the argument order of `propEq`, I agree with the signature change made in this MR. With that change, the following functions all share the same argument order:
* pathEq
* pathOr
* pathSatisfies
* propEq
* propIs
* propOr
* propSatisfies

I'm going to argue that this consistency is the most important part here. What the argument order _should_ be is completely subjective.

As for function naming, we have a pretty decent convention going already, for the aforementioned functions, the convention is `subject -> prop/path -> object -> boolean`. In contrast, we have these functions
* assoc
* assocPath
* dissoc
* dissocPath
* adjust
* update

If we fully qualify the names, we get this:
* assocProp
* assocPath
* dissocProp
* dissocPath
* adjustIndex
* updateIndex

And the convention for these are `prop/path -> subject -> object -> object`.

To summarize, when prop/path comes first in the function name, prop/path is the second argument, and when prop/path comes second in the function name, prop/path is the first argument

Now, that convention could also apply to the fact that prop/path second is "mutate" and prop/path first is "evaluate against". We could keep that naming convention for that purpose and have them all share the same general signature of `prop/path -> subject -> object -> result` (this is what I meant about the argument order being subjective).

In the end, I think consistency is the most important part here. Therefore I vote _not_ to revert this MR.

If it is decided that the argument order should be changed, then I would propose that, as part of a long overdue `1.0.0` release, change **ALL** of their signatures to be `prop/Path -> subject -> object -> boolean` **TOGETHER** as part of a single release.

whitelizard:
Every function should stand on its own and not be dependent on or ruled by other functions and their most logical & convenient names and signatures (arg order). I don't think that arg order is always subjective, but the name should reflect the order, that is most important.
You could probably find many different groupings among all ramda functions, but if a grouping is based on names, and we adjust signatures based on the group of names, then we are working backwards! Signature comes first, then a name that correctly reflects what the function does AND (almost more importantly) the argument order -- regardless of other functions in the library.
(See my previous post)

Harris-Miller:
> regardless of other functions in the library.

Good APIs have consistent and predictable behavior. And that includes function signatures and argument order. My point is that they should be consistent. I'm not particular about which of the 2 orders here is correct, `subject -> prop/path -> object -> boolean` or `prop/Path -> subject -> object -> boolean`.

>  I don't think that arg order is always subjective

What I mean to say is the reasoning behind why the arguments should be in the order they are is subjective. Using your own example here
> **Example 1: A function that checks a predicate for an object prop**
Step 1:
(a → Boolean) → String → {String: a} → Boolean
Note, as mentioned above, a function is to be considered highly general and "far away" from being "data". Therefor the predicate is best put first.

and

> **Example 2: A function that checks an object prop aganst a value**
Step 1:
String → a → Object → Boolean
a is a value closer to being "data" than the name of the prop, therefor it should come later in the argument list.

Let's rewrite these 2 signatures generically.
* Example 1: `X -> Key -> Object -> Boolean`
* Example 2: `Key -> X -> Object -> Boolean`

The idea that a "function" is further away from "data" than "value", so `Key` should come second in one case and second in the other is _subjective_. I could argue that the `Key` is closer to the `Object` in both cases.

Besides, `propEq = (value, key, obj) => propSatisfies(x => x === value, key, obj)`

Whatever the order is, just have it be consistent. `propEq` was inconsistent before, and now it is.

I don't disagree with you that `propEq` should be `String -> a -> Object -> Boolean`, all I'm saying is that if we choose to do that, we should for `propSatisfies`, `propIs`, etc, as well.

Harris-Miller:
For any future argument changes, I think the community would benefit greatly from codemods that `ramda` could supply as part of it's release process. Similar to what react supplies: https://github.com/reactjs/react-codemod

crshmk:
1.)
```javascript 
const idEq = propEq('id')
```
is useful. 
```javascript 
const somePropEqualsOne = propEq(1)
```
is not useful. 

It seems to me that the API should stem from utility;  not from unmeasurable pontifications of what kinds of arguments are more general. 

2.) 
A few people, otherwise astute and transgressing accidentally rather than in trade, above and in related threads, have proposed all sorts of breaks - not simple swaps but name collisions - in order to align a signature with what they presumed it would be when they first read a function name. When I get a param order wrong in the same situation, I consult the well-organized documentation and quickly solve the problem in about a minute. I do not insist other users spend days changing their code and memorizing a new api for no benefit to them. I can't ever recall using `concat` without a placeholder as the first argument. I never found it appropriate to suggest the entire world adopt my way of viewing concat; I just got used to writing `pipe(concat(__, newItems)` because changing the signature would cause so many complex problems that obviously outweigh my own trivial concerns.

whitelizard:
@Harris-Miller I feel that you don't get my point. 

The **naming** is wrong, not the signature. 

As I wrote earlier: e.g. "valueAtProp" would be the better name if the prop name (key) comes first in the argument list (`Key -> X -> Object -> Boolean`). 

It is not the start of the function name that refers to the first argument, it is the last part of the function name that should refer to the first function argument. This is because of currying.

```js
const ageIs = R.valueAtProp('age');
const isPerson20yearsOld = ageIs(20);
const celebratePerson = isPerson20yearsOld(person);
```

Do you agree that the intention of ramda functions is to be able to use currying in the way like above? And that the names should read well with the argument that follows, like above?

If the rules that I wrote in my first long post is followed, then the API would be entirely consistent.

Harris-Miller:
> Do you agree that the intention of ramda functions is to be able to use currying in the way like above? And that the names should read well with the argument that follows, like above?

@whitelizard Yes I do agree. My arguments for why `propEq` should have been changed as described in this MR from the start is solely for that and that alone. That does not mean that I disagree with you on the fact that the names _should_ be different to better represent their signatures, or that the signatures _should_ be different to better match the current names.

My goal is for consistency in behavior and expectation in the API.

> If the rules that I wrote in my first long post are followed, then the API would be entirely consistent.

Also agree, as it aligns with my stated goal

whitelizard:
So, the solution could be simple. 

**`propEq` is not really a useful function**, as @crshmk has pointed out multiple times above. It could stay in the library, with arg order changed as by this MR, but the library should have the **more useful** function `valAtProp` (or whatever would be the best name). Also change the other inconsistent functions (their names!) according to @semmel.

Release 1.0.0.

Again, the order of designing any API function should be: 

**Need -> Signature -> Name**

Never change the signature of a function to match the faulty name it unfortunately got. The signature was designed out of **greatest utility**. The name should change if it got a bad one. Changing the name is not initially a breaking change, it could start with a minor release of **adding** the new name to the API and flagging the old as "will be removed in the next major". Ramda could have this as part of its _change strategy_.

Gastonite:
100% agree with @whitelizard
And I would add that I think it's preferable to consider the users of the library and their (often large) code base when planning such a huge change. By this I mean the difficulty of the migration needed when inverting the parameters of a curried function, which could be used in several ways, compared to the simplicity of the migration if we just changed its name.

I also agree with having consistent functions in a library is a good thing, but maybe it would be a good thing also to add another function (temporary or not) to replace with when migrating.

Harris-Miller:
@whitelizard just to be clear, to extend on [this table](https://github.com/ramda/ramda/pull/2938#issuecomment-1528863527)

existing fn | signature | consistent name
-- | -- | --
allPass | [(*… → Boolean)] → *… → Boolean | passesAll
anyPass | [(*… → Boolean)] → *… → Boolean | passesAny
append | a -> [a] -> [a] | appendValue
appendFlipped | [a] -> a -> [a] | appendTo
nthArg | Number → *… → * | argNth
  | [Idx] → a → {a} → Boolean | eqPath
  | Idx → a → {a} → Boolean | eqProp
included | [a] → a → Boolean | includedIn
pathEq | a → [Idx] → {a} → Boolean | valAtPath
propEq | a → Idx → {a} → Boolean | valAtProp

While obviously incomplete, the table just represents the idea of deprecating existing fn columns, adding their replacements in the consistent name column, as well as adding new functions that don't currently exist

For the sake of typescript I personally like having both `eqProp` and `valAtPath`. There is currently no good way to type `flip` for functions of arity 3+. Having both varieties of all these functions would help a lot in that regard.

CC: @kedashoe

semmel:
> For the sake of typescript I personally like having complementary functions eqProp/valAtProp, append/appendValue, etc. There is currently no good way to type flip for functions of arity 3+. Having both varieties of all these functions would help a lot in that regard.

Yes, and I would extend that to a few *binary* functions too. They have crept up in [Ramda-adjunct](https://char0n.github.io/ramda-adjunct/4.0.0/index.html) eg. `R.includes ≡ flip(RA.included)`, `R.append ≡ flip(RA.appendFlipped)`, but in Ramda too e.g. `R.difference ≅ flip(R.without)`. 

This is always when *"…the reasoning behind why the arguments should be in the order they are is subjective"*.

Another example which comes to my mind is `concat`. Written in a function pipeline `pipe(() =>a, concat(b))()` I always need to imagine the binary form `concat(x, y) = [...x, ...y]` to find out, that the pipeline kind of *appends* all of `a` to `b` and not the other way round. That reasoning is ok. Is is not too hard to figure out and consistent in Ramda (`difference`, `subtract`).

On the other hand, while you @Harris-Miller mention it, `flip` for functions of arity 3+ gives my headaches — and not because of TypeScript. I think its an anti-pattern. (Just as a side-note)

jktravis:
I know this has been beat to death, but I'd just like to add my two cents. I really hate this change. I just propEq and pathEq way more than propSatisfies. Plus, the naming, as other has suggested, practically tells you how to use it: prop, then equals. I've always had to look up the docs when needing propSatisfies simply because of the naming inconsistency. But even still, I think it would have made more sense to change propSatisfies to match the other two. /shrug

papaver:
would it be possible to add this change to the `propEq` documentation at least?  it's extremely confusing for anyone using 0.28.0.  i just wasted several hours trying to figure out why my code was working in the doc page but not in my code.

thanks.

michaek:
Ouch. It seems pretty onerous have to change any and all calls to propEq/pathEq for such a flimsy and subjective reason. I get that the library will probably not change back to its historical API at this point, but this seems hostile to longtime users, and it certainly makes me consider the wisdom of moving away from a dependency on Ramda.

whitelizard:
> @whitelizard just to be clear, to extend on [this table](https://github.com/ramda/ramda/pull/2938#issuecomment-1528863527)
> 
> existing fn	signature	consistent name
> allPass	[(*… → Boolean)] → *… → Boolean	passesAll
> anyPass	[(*… → Boolean)] → *… → Boolean	passesAny
> append	a -> [a] -> [a]	appendValue
> appendFlipped	[a] -> a -> [a]	appendTo
> nthArg	Number → *… → *	argNth
>  	[Idx] → a → {a} → Boolean	eqPath
>  	Idx → a → {a} → Boolean	eqProp
> included	[a] → a → Boolean	includedIn
> pathEq	a → [Idx] → {a} → Boolean	valAtPath
> propEq	a → Idx → {a} → Boolean	valAtProp
> While obviously incomplete, the table just represents the idea of deprecating existing fn columns, adding their replacements in the consistent name column, as well as adding new functions that don't currently exist
> 
> For the sake of typescript I personally like having complementary functions `eqProp`/`valAtProp`, `append`/`appendValue`, etc. There is currently no good way to type `flip` for functions of arity 3+. Having both varieties of all these functions would help a lot in that regard.
> 
> CC: @kedashoe

I believe all number comparison functions are also the wrong order: `lt`, `gt`, `lte`, `gte`.

kedashoe:
> I believe all number comparison functions are also the wrong order

https://github.com/ramda/ramda/issues/1497

Issue #2921, opened 2019-11-19
uncurryN returns a ramda style curried function from a standard curried function?


adispring:
```js
const addFour = a => b => c => d => a + b + c + d;
const uncurriedAddFour = R.uncurryN(4, addFour);
uncurriedAddFour(1, 2, 3, 4); //=> 10
```

This also works:
```js
uncurriedAddFour(1, 2, 3)(4); //=> 10
```

Should this function add some additional explanation like "Note that, the returned function is an n-ary ramda style curried function" ?

CrossEye:
Yes, that would be an improvement.  Do you want to create a PR or should I do it?

adispring:
OK, I will create a PR.

PR #2939, opened 2019-12-20
add note: uncurryN actually return a ramda style curried function.

adispring:
related to #2921


CrossEye:
Thank  you very much!

PR #2940, opened 2019-12-20
_createPartialApplicator: avoid passing in wrong count of arguments

adispring:
If I have a function created by `partialRight`, and passing in wrong count of arguments, this will lead to wrong behavior of original function. For example:

```js
const greet = (salutation, title, firstName, lastName) =>
  salutation + ', ' + title + ' ' + firstName + ' ' + lastName + '!';

const greetMsJaneJones = R.partialRight(greet, ['Ms.', 'Jane', 'Jones']);

greetMsJaneJones('Hello', 'Mr.' 'Green'); //=> 'Hello, Mr. Green Jones!'
```

So  function created by `partialRight`, should be restricted to the right count of parameters?

### DIFF
--- a/source/internal/_createPartialApplicator.js
+++ b/source/internal/_createPartialApplicator.js
@@ -4,8 +4,10 @@ import _curry2 from './_curry2';
 
 export default function _createPartialApplicator(concat) {
   return _curry2(function(fn, args) {
-    return _arity(Math.max(0, fn.length - args.length), function() {
-      return fn.apply(this, concat(args, arguments));
+    var arity = Math.max(0, fn.length - args.length);
+    return _arity(arity, function() {
+      var neededArgs = Array.prototype.slice.call(arguments, 0, arity);
+      return fn.apply(this, concat(args, neededArgs));
     });
   });
 }


CrossEye:
This is definitely a problem.  But I don't think it's the right fix.  I think the behavior should be something like this:

```js
const foo = (a, b, c, d, ...rest) => ({a, b, c, d, rest})

foo (1, 2, 3, 4, 5, 6)
//=> {a: 1, b: 2, c: 3, d: 4, rest: [5, 6]}

partialRight(foo, [100, 200])(1, 2, 3, 4)
//=> {a: 1, b: 2, c: 100, d: 200, rest: [3, 4]}
```

Right now that would return `{a: 1, b: 2, c: 3, d: 4, rest: [100, 200]}`.  I'm guessing the easiest way would be to simply abandon that helper, which isn't doing much anyway, and rewrite it in some ES5 version of 

```js
const partialRight = (fn, origArgs) => (...args) => 
  fn (...insertAll (fn.length - origArgs.length, origArgs, args))
```

CrossEye:
closing in favor of #2941

PR #2941, opened 2019-12-22
optimize partial and partialRight to correctly process rest args

adispring:
optimize partial and partialRight as you suggested

related to #2940


test/partial.js
```diff
@@ -21,4 +21,9 @@ describe('partial', function() {
     eq(g.length, 1);
   });
 
+  it('correctly processes rest arguments', function() {
+    var foo = (a, b, c, d, ...rest) => ({ a, b, c, d, rest });
+    var f = R.partial(foo, [100, 200]);
+    eq(f(1, 2, 3, 4), { a: 100, b: 200, c: 1, d: 2, rest: [3, 4] });
+  });
```
CrossEye: So long as we're supporting ES5, I don't think we can include this test.

test/partialRight.js
```diff
@@ -21,4 +21,9 @@ describe('partialRight', function() {
     eq(g.length, 1);
   });
 
+  it('correctly processes rest arguments', function() {
+    var foo = (a, b, c, d, ...rest) => ({ a, b, c, d, rest });
+    var f = R.partialRight(foo, [100, 200]);
+    eq(f(1, 2, 3, 4), { a: 1, b: 2, c: 100, d: 200, rest: [3, 4] });
+  });
 });
```
CrossEye: Ditto

test/partial.js
adispring: So how should I modify the test, should these test files stay as before?

CrossEye: I think we should remove the tests for `...rest` and add one for the version of `greet` that inspired #2940.  That should show the behavior well enough.

adispring: Like this?:

```
const greet = (salutation, title, firstName, lastName) =>
  salutation + ', ' + title + ' ' + firstName + ' ' + lastName + '!';

const greetMsJaneJones = R.partialRight(greet, ['Ms.', 'Jane', 'Jones']);

greetMsJaneJones('Hello', 'Mr.' 'Green'); //=> 'Hello, Ms. Jane Jones!'
```

CrossEye: exactly.

adispring: Has changed the test as you suggested.

CrossEye (dismissed):

CrossEye:
This looks good to me.  Others?  @ramda/core?

CrossEye:
@adispring: Another issue that's been dropped.  This should go in.  Could you resolve the conflicts with the current HEAD?

adispring:
Ok, I will resolve the conflict soon.

adispring:
The conflict has been resolved.

customcommander (approved):
LGTM

PR #2943, opened 2019-12-25
Make R.eqProps safe for null/undefined arguments

adispring:


CrossEye:
This looks great to me.  I see no downside.

Any one else, @ramda/core?

PR #2944, opened 2019-12-27
make unionWith's description more clear: use first element from first list if duplicate

adispring:
make `unionWith`'s description more clear, use first element from first list if duplicate, for example:

```js
const l1 = [{a: 1, b: 1}, {a: 1, b: 2}, {a: 2}];
const l2 = [{a: 1}, {a: 4}];
R.unionWith(R.eqProps('a'), l1, l2); 
//=> [{a: 1, b: 1}, {a: 2}, {a: 4}]
```


CrossEye:
Thank you!

:herb:

PR #2945, opened 2019-12-27
introduce husky to guard ramda

adispring:
Run npm test and lint before git push.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3047,6 +3047,12 @@
       "integrity": "sha512-5tabW/i+9mhrfEOUcLDu2xBPsHJ+X5Orqy9FKpale3SjDA17j5AEpYq5vfy3oAeAHGcvANRCO3NV3d2D6q3NiA==",
       "dev": true
     },
+    "@types/normalize-package-data": {
+      "version": "2.4.0",
+      "resolved": "http://bnpm.byted.org/@types/normalize-package-data/download/@types/normalize-package-data-2.4.0.tgz",
+      "integrity": "sha1-5IbQ2XOW15vu3QpuM/RTT/a0lz4=",
+      "dev": true
+    },
     "Base64": {
       "version": "0.2.1",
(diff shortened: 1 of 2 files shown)


CrossEye:
That's probably a good idea.  I do it in all my work projects these days.

Any objections, @ramda/core?

davidchambers:
I dislike commit hooks, but I have not committed to this project for a long time so don't mind me.

On an unrelated note, why do we include __package-lock.json__ in the repository? I've never seen a reason for libraries to include such files.

GingerPlusPlus:
It's pre-push, not pre-commit.

👎 from me &ndash; this checks working directory, instead of commits that are about to be pushed.

About lockfile: https://github.com/ramda/ramda/pull/2625#issuecomment-412553820.

adispring:
> 👎 from me – this checks working directory, instead of commits that are about to be pushed.

Using `lint-staged` , we can lint only changed files at commit stage, and only run unit test at push stage.

GingerPlusPlus:
I wouldn't oppose `lint-staged`, even tho I'm not sure if it [properly](https://gist.github.com/luuuis/e41fd71134ce88ac5e9359cbdbfb6273) handles partial stages.

Naive `npm test` in `pre-push` disallows untracked files in `source/` and makes no sense when using `git push <remote> <branch>`. I don't know how to do this properly, I don't think this is the right place.

CI already lints and runs tests, correctly.

adispring:
> Naive `npm test` in `pre-push` disallows untracked files in `source/` and makes no sense when using `git push <remote> <branch>`. I don't know how to do this properly, I don't think this is the right place.
> 
You can stash the untracked files before push. 

Run lint and unit test locally can find some obvious errors earlier, maybe this can save some time than remote CI.

GingerPlusPlus:
Hooks are supposed to enhance or ensure valid workflows, not break them.

PR #2946, opened 2019-12-28
add lensEq and lensSatisfies

jabbla:
add ``lensEq`` and ``lensSatisfies``.
```js
const obj = { a: 1, b: 2, c: { d: 3 } };
const dLens = R.lensPath(['c', 'd']);

R.lensEq(dLens, 3, obj);  //=> true
R.lensSatisfies(n => n <= 3, dLens, obj);  //=> true
```
``lens`` is very useful. we should strengthen its capabilities.


test/lensEq.js
```diff
@@ -0,0 +1,57 @@
+var R = require('../source');
+var eq = require('./shared/eq');
+
+
+describe('lensEq', function() {
+  var obj = {name: 'Abby', age: 7, hair: 'blond', family: { father: 'Jones' }};
+  var array = [1, 2, 3, 4];
+
+  it('able to use with lens', function() {
```
CrossEye: Two things here:

  - I don't think there's any reason to call out tests for different implementations of lenses.  Lenses are their own abstraction which can be tested in their own implementations (`lens`, `lensIndex`, `lensPath`, etc.)  While it would probably be a good idea if the various test cases used a few different lens creators, there is no particular reason to call attention to it in the test descriptions.

  - in this pseudo-BDD style of tests, I really prefer if the the description read as sentences when appended to the `describe` value.  That is

    ```
    describe ('My noun', () => {
      it ('has properties described by this phrase')
    })
     ```

    which of course is read in the output as "My noun has properties described by this phrase."

test/lensSatisfies.js
```diff
@@ -0,0 +1,25 @@
+var R = require('../source');
+var eq = require('./shared/eq');
+
+
+describe('lensSatisfies', function() {
+
+  var isPositive = function(n) { return n > 0; };
+
+  it('returns true if the specified object property, the lens focused on, satisfies the given predicate', function() {
```
CrossEye: These test descriptions seem fine, by contrast.

source/lensEq.js
```diff
@@ -0,0 +1,39 @@
...
+import equals from './equals';
+
+
+/**
+ * Returns `true` if the property, path, index, the lens focused on, is equal, in
+ * [`R.equals`](#equals) terms, to the given value; `false` otherwise.
+ *
+ * @func
+ * @memberOf R
+ * @since v0.27.1
```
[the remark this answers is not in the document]
CrossEye: That's correct.  This is auto-generated.

Bradcomp:
With partial.lens in existence for those who want to go all in, and a solid but basic base for people who are interested in using lenses to try them out, I don't know if we should be adding more lens functions. 

I can see the potential in replacing (reviving?) [Ramda Lens](https://github.com/ramda/ramda-lens) which is currently way out of date. We could then remove the core lenses from Ramda and put them in there, and there would be a clearer case for adding these functions. 

With that being said, I'd want to know what our library would have to offer in terms of philosophy or functionality that we can't get from the Calmm library.

I have found lenses useful on a rare occasion but they've never been a huge piece of my toolkit. If there are others that use Ramda's lenses on a regular basis or prefer them to partial.lens I would be interested in hearing from them.

jabbla:
@CrossEye @GingerPlusPlus @Bradcomp 
Thanks for all your reply. I've learned a lot from the discussions above. 😊

PR #2955, opened 2020-01-08
make tap's symb description more clearly

adispring:
`tap` run the side-effect with input first, and return the input. 

`@symb R.tap(f, a) = (f(a), a)` describes what `R.tap` do more clearly.


customcommander (approved):
Personally I think the documentation at the top is clear enough so not 100% convinced we need this change. Having said that I don't mind at all. So 🌻

customcommander:
@ramda/ramda-docs Good to go?

CrossEye (approved):

PR #2956, opened 2020-01-08
Add clarification example to compose not auto-currying

josh-byster:
This is a rather small change so I understand if it isn't worth adding given that there's already the note, but for someone rather new to functional programming like myself, seeing the consequence of the fact that `compose` and `pipe` do not auto-curry is helpful.


CrossEye:
This is absolutely not too small.  It looks to me an excellent additional form of documentation.

Thanks!

:herb:

PR #2958, opened 2020-01-21
optimize call signature & its example's indent

adispring:


CrossEye:
As always, thank  you very much, @adispring !  :herb:

PR #2959, opened 2020-01-22
wrap R.call with _curry1 instead of curry

adispring:
`call` actually uses `_curry1` in `curry`'s implementation, so it can use `_curry1` directly.


CrossEye:
Thanks once again!  :herb:

PR #2962, opened 2020-01-25
"prop" optimization (10x faster)

ku8ar:
In the current version of the code, "prop" uses "paths" underneath. This is only done so as not to duplicate **one** line of logic.
I copied the logic choosing property directly to "prop".
The overhead on "prop" was so great that this function started to work **10x faster**!

Why is this optimization so important? Most programmers use "prop" as a substitute for the null safety "**?.**" syntax. This PR will drastically accelerate all React projects where Ramda is used during every rerender (e.g. in selectors).

https://jsperf.com/ramda-optimized-prop

Zrzut ekranu 2020-01-25 o 16 53 23

### DIFF
--- a/source/prop.js
+++ b/source/prop.js
@@ -1,5 +1,6 @@
 import _curry2 from './internal/_curry2';
-import path from './path';
+import _isInteger from './internal/_isInteger';
+import nth from './nth';
 
 
 /**
@@ -24,5 +25,10 @@ import path from './path';
  *      R.compose(R.inc, R.prop('x'))({ x: 3 }) //=> 4
  */
 
-var prop = _curry2(function prop(p, obj) { return path([p], obj); });
+var prop = _curry2(function prop(p, obj) {
+  if (obj == null) {
+    return;
+  }
+  return _isInteger(p) ? nth(p, obj) : obj[p];
+});
 export default prop;
--- a/source/propOr.js
+++ b/source/propOr.js
@@ -1,5 +1,6 @@
 import _curry3 from './internal/_curry3';
-import pathOr from './pathOr';
+import defaultTo from './defaultTo';
+import prop from './prop';
 
 
 /**
@@ -29,6 +30,6 @@ import pathOr from './pathOr';
  *      favoriteWithDefault(alice);  //=> 'Ramda'
  */
 var propOr = _curry3(function propOr(val, p, obj) {
-  return pathOr(val, [p], obj);
+  return defaultTo(val, prop(p, obj));
 });
 export default propOr;
--- a/test/prop.js
+++ b/test/prop.js
@@ -11,6 +11,36 @@ describe('prop', function() {
     eq(nm(fred), 'Fred');
   });
 
+  it('shows the same behaviour as path for a nonexistent property', function() {
+    var propResult = R.prop('incorrect', fred);
+    var pathResult = R.path(['incorrect'], fred);
+    eq(propResult, pathResult);
+  })
+
+  it('shows the same behaviour as path for an undefined property', function() {
+    var propResult = R.prop(undefined, fred);
+    var pathResult = R.path([undefined], fred);
+    eq(propResult, pathResult);
+  })
+
+  it('shows the same behaviour as path for a null property', function() {
+    var propResult = R.prop(null, fred);
+    var pathResult = R.path([null], fred);
+    eq(propResult, pathResult);
+  })
+
+  it('shows the same behaviour as path for a valid property and object', function() {
+    var propResult = R.prop('age', fred);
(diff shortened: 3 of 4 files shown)


ku8ar:
For desktop apps, this optimization is not important. But currently most of JS code is executed on smartphones.

Suppose we have such React component:
```
const fields = ['name', 'surname', 'nickname', 'age', 'status', 'active']
const Item = ({item}) => (
<View>{map(field =>  <Text>{prop(field, item)}</Text>, fields)}</View>
)
```
Of course, this code is maximally simplified. Usually Ramda is directly in selectors.
And we display this component 50 times (item list).

Finally, 300 calls come out for each single application rendering.

And now please look how efficient the prop() function is on Android:
83171325_853379038455037_7202341597667655680_n

It gives us: (1 / 282169) * (6 * 50) * 1000 = 1 ms
And with optimization it gives us: (1 / 19790334) * (6 * 50) * 1000 = 0,01 ms

Finally, we waste 1ms on simply retrieving value from object, which could be done using: `field && field.value`

How many such rerenders will the app have? Will there be only one list called at time? Is it worth adding this code to save 1ms? I do not know.

Thinking about such micro-optimization begins to be difficult when we understand that in addition to displaying, such a list of items could also be used for something else. E.g. a hook that does something:
```
useEffect(() => {
  track('active items showed', pipe(filter(prop('active'), map(prop('id'))))(items))
}, [])
```
Or when it turns out that there are 1000 elements in store, and we only have to show 50:
```
<Items items={filter(prop('isBelongToMe'), allItems)} />
```

And suddenly it turns out that it only took 15 ms to get properties from objects during first rendering. And that's 1 FPS lost...

CrossEye:
As I said, I think these numbers are compelling.  Do you have any thoughts on how to write some regression tests to ensure that the API keeps in sync with `path`?

Also, we could probably optimize `path` quite a bit by not depending on (in my impression the rarely used) `paths`.  We could easily write the dependency the other way by mapping.  I wonder how much of the lost performance we would gain by doing so.

I rarely use `prop` or `path` unless the code-base is already heavily using Ramda; they are so easy to write as custom functions.  But we do want both to make them efficient and to make them work the same as much as feasible.  It would be great to be able to do the latter automatically, but it's probably not worth it with this much performance degradation on the line.

Bradcomp:
>  Do you have any thoughts on how to write some regression tests to ensure that the API keeps in sync with path?

I'll fall back on my old standard suggestion here and recommend property based tests.

For any given object and property, `prop(p, o) === path([p], o`.

ku8ar:
I added tests :)

source/prop.js
```diff
@@ -24,5 +25,10 @@ import path from './path';
  *      R.compose(R.inc, R.prop('x'))({ x: 3 }) //=> 4
  */
 
-var prop = _curry2(function prop(p, obj) { return path([p], obj); });
+var prop = _curry2(function prop(p, obj) {
+  if (obj == null) {
+    return;
+  }
+  return _isInteger(p) ? nth(p, obj) : obj[p];
```
Bradcomp: Wouldn't `obj[p]` work for both instances? Is there a case where nth would return something different?

ku8ar: This is the code copied from "paths". I didn't want to change any logic. Add tests to this case?

Bradcomp: Interesting. I don't know the history behind this code. For both cases I would think we don't need the check, however I wouldn't want to block the PR on that. I am OK leaving it as-is.

datenreisender: `nth` does support negative indices, which normal array access does not: 

```js
const fruit = ['banana', 'apple', 'peach'];

[
  fruit[-1], 
  nth(-1, fruit),
  fruit[-1] === nth(-1, fruit),
]
```
[REPL evaluates it to `[undefined, "peach", false]`](https://ramdajs.com/repl/#?const%20fruit%20%3D%20%5B%27banana%27%2C%20%27apple%27%2C%20%27peach%27%5D%3B%0A%0A%5B%0A%20%20fruit%5B-1%5D%2C%20%0A%20%20nth%28-1%2C%20fruit%29%2C%0A%20%20fruit%5B-1%5D%20%3D%3D%3D%20nth%28-1%2C%20fruit%29%2C%0A%5D)

ku8ar: We will need to write tests for this logic. And review rest of code. Becasue, for example, `propIs` doesn't have `nth()` inside...

CrossEye: Support for negative indices keeps causing issues!

masoud-msk:
`props` is still using `path` in its implementation. As discussed here, I think it should use `prop` for both readability and performance.

Issue #2957, opened 2020-01-20
Typo in tryCatch documentation


jonhallur:
https://brians.wsu.edu/2016/05/31/catched/

"catched" should be "caught"

CrossEye:
I wonder how that snuck through.  Do you want to create a quick PR to fix it?

B1Z0N:
does anybody closing it?

CrossEye:
It was fixed in #2968.  The documentation will be updated the next time the library is published.

PR #2968, opened 2020-02-02
update try catch docs to use caught instead of catched

ryanbas21:
#2957 - quick docs fix to change catched to caught.


CrossEye:
Thank you very much @ryanbas21!  :herb:

PR #2983, opened 2020-02-13
Update and rename head.js to first.js

tanrax:


CrossEye:
Thank you very much for the PR.

Note that it is broken.  You can see the results [on Travis][tr].

I'm not much interested in renaming this.  `head`/`tail` has a long history.  While adding some symmetry with `last`, it would remove the symmetry with `tail`; and to my mind, `head` and `tail` are the much more important functions than `init`/`last`.


  [tr]: https://travis-ci.org/ramda/ramda/jobs/650081948

tanrax:
First I want to thank you for the effort, I'm a big fan of the project.
I usually work with Clojure and I was confused that there was a last but not a first. In case there is no equivalent... delete last or create an alias for first?
Thanks

CrossEye:
@tanrax: Ramda has taken a stand against any aliases.  We do have *See Also* links in the documentation for `last` to `init`, `head`, and `tail`.

PR #2984, opened 2020-02-15
add number property tests to prop* and fix propIs (when property is negative number)

ku8ar:
https://github.com/ramda/ramda/pull/2962/files#r377375414


CrossEye:
:herb:

LGTM.  Anyone else?

PR #2985, opened 2020-02-17
improve clone performance

danillosl:
Changed _clone to use a Map to cache and retrieve the cloned objects.
this is NOT a breaking change and in my tests reduce the time to clone this [object](https://paste.ee/p/LNvsR) from 93.2ms to 6.8ms on average after 100 iterations.

### DIFF
--- a/source/internal/_clone.js
+++ b/source/internal/_clone.js
@@ -10,7 +10,10 @@ import type from '../type.js';
  * @param {Boolean} deep Whether or not to perform deep cloning.
  * @return {*} The copied value.
  */
-export default function _clone(value, deep, map = new Map()) {
+export default function _clone(value, deep, map) {
+
+  map || (map = new _ObjectMap());
+
   // this avoids the slower switch with a quick if decision removing some milliseconds in each run.
   if (_isPrimitive(value)) {
     return value;
@@ -23,6 +26,7 @@ export default function _clone(value, deep, map = new Map()) {
     if (cachedCopy) {
       return cachedCopy;
     }
+
     map.set(value, copiedValue);
 
     for (var key in value) {
@@ -58,3 +62,61 @@ function _isPrimitive(param) {
   var type = typeof param;
   return param == null || (type != 'object' && type != 'function');
 }
+
+function _ObjectMap() {
+  this.map = {};
+  this.length = 0;
+}
+
+_ObjectMap.prototype.set = function(key, value) {
+  const hashedKey = this.hash(key);
+
+  let bucket = this.map[hashedKey];
+  if (!bucket) {
+    this.map[hashedKey] = bucket = [];
+  }
+
+  bucket.push([key, value]);
+  this.length += 1;
+};
+
+_ObjectMap.prototype.hash = function(key) {
+  let hashedKey = [];
+  for (var value in key) {
+    hashedKey.push(key[value]);
+  }
+  return hashedKey.join();
+};
+
+_ObjectMap.prototype.get = function(key) {
+
+  /**
+   * depending on the number of objects to be cloned is faster to just iterate over the items in the map just because the hash function is so costly,
+   * on my tests this number is 180, anything above that using the hash function is faster.
+   */
+  if (this.length <= 180) {
+
+    for (const p in this.map) {
+      const bucket = this.map[p];
+
+      for (let i = 0; i < bucket.length; i += 1) {
+        const element = bucket[i];
+        if (element[0] === key) {return element[1];}
+      }
+
+    }
+    return;
+  }
+
+  const hashedKey = this.hash(key);
+  const bucket = this.map[hashedKey];
+
+  if (!bucket) {return;}
+
+  for (let i = 0; i < bucket.length; i += 1) {
+    const element = bucket[i];
+    if (element[0] === key) {return element[1];}
+  }
+
+};
+
--- a/source/internal/_clone.js
+++ b/source/internal/_clone.js
@@ -7,30 +7,32 @@ import type from '../type.js';
  *
  * @private
  * @param {*} value The value to be copied
- * @param {Array} refFrom Array containing the source references
- * @param {Array} refTo Array containing the copied source references
  * @param {Boolean} deep Whether or not to perform deep cloning.
  * @return {*} The copied value.
  */
-export default function _clone(value, refFrom, refTo, deep) {
+export default function _clone(value, deep, map = new Map()) {
+  // this avoids the slower switch with a quick if decision removing some milliseconds in each run.
+  if (_isPrimitive(value)) {
+    return value;
+  }
+
   var copy = function copy(copiedValue) {
-    var len = refFrom.length;
-    var idx = 0;
-    while (idx < len) {
-      if (value === refFrom[idx]) {
-        return refTo[idx];
-      }
-      idx += 1;
+    // Check for circular and same references on the object graph and return its corresponding clone.
+    var cachedCopy = map.get(value);
+
+    if (cachedCopy) {
+      return cachedCopy;
     }
-    refFrom[idx] = value;
-    refTo[idx] = copiedValue;
+    map.set(value, copiedValue);
+
     for (var key in value) {
       if (value.hasOwnProperty(key)) {
-        copiedValue[key] = deep ? _clone(value[key], refFrom, refTo, true) : value[key];
+        copiedValue[key] = deep ? _clone(value[key], true, map) : value[key];
       }
     }
     return copiedValue;
   };
+
   switch (type(value)) {
     case 'Object':  return copy(Object.create(Object.getPrototypeOf(value)));
     case 'Array':   return copy([]);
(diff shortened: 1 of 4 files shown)


newyankeecodeshop:
How about implementing an internal `_Map` that uses the global Map if available but falls back to the existing implementation for older runtimes?

CrossEye:
@newyankeecodeshop:

That might work.  There is a real tension, though, between trying to make an internal _Map that might gain some of this speedup and one which would work as expected in Ramda (that is, one focused on value equality rather than reference equality.)

danillosl:
@CrossEye:

After thinking about the options I implemented the _ObjectMap which is a private hashmap that uses the object values as a hash, using this class is not as fast as Map but still we get a significant improvement in performance.

this are the values that i get using the clone function now:
max/min/avg time in ms after 100 runs
ramda: new clone              max: 71.066 min: 23.946 avg: 30.496
ramda: old clone              max: 440.424 min: 411.802 avg: 421.692
lodash cloneDeep():           max: 26.248 min: 9.779 avg: 11.207
JSON.parse(JSON.stringify()): max: 16.261 min: 9.313 avg: 10.276

source/internal/_clone.js
```diff
@@ -12,29 +11,104 @@ import type from '../type';
...
+  this.length += 1;
+};
+
+_ObjectMap.prototype.hash = function(key) {
+  let hashedKey = [];
+  for (var value in key) {
+    hashedKey.push(key[value]);
+  }
+  return hashedKey.join();
+};
```
CrossEye: I think there's a problem with any generic hash here.  For this simple one, note that 

```js
hash ({foo: 1, bar: 2}) == hash ({bar: 1, foo: 2}) //=> true
hash ({foo: {a: 42}}) == hash ({foo: {b: false}})  //=> true
```

That second one is especially problematic.  We could easily end up with data that's all hash collisions.

But even a more sophisticated hash can still have such problems.  I suppose if we get sophisticated enough, then we'd probably lose much of the benefit.

source/internal/_clone.js
```diff
@@ -12,29 +11,104 @@ import type from '../type';
...
+   * on my tests this number is 180, anything above that using the hash function is faster.
+   */
+  if (this.length <= 180) {
+
+    for (const p in this.map) {
+      const bucket = this.map[p];
+
+      for (let i = 0; i < bucket.length; i += 1) {
+        const element = bucket[i];
+        if (element[0] === key) {return element[1];}
```
CrossEye: I think this would have to be `if (equals (element [0], key)) ...`

Ramda is all about value equality, not reference equality.

source/internal/_clone.js
danillosl: Yes, the hash function is very simple and if we made it more sophisticated we end up losing performant, but I think this hash function is just good enough to split the data a little bit to make it performant, and yes we could in a worst-case scenario end up with a collection full of collisions, but in that case, the get function will just behave the way the old algorithm behaved.

source/internal/_clone.js
danillosl: While I understand that Ramda is about value equality I think that the clone function should preserve the structure of the object, let's say:
```js
const d = {foo:1, bar:"2"};
const a = {foo: d, c:{ bar: d}};
a.foo === a.c.bar // == true;

//the clone method should preserve the structure.

const clone = R.clone(a);
clone.foo === clone.c.bar // == should also be true;
```
That's the way the R.clone() is behaving and lodash cloneDeep() also behave like this and I also believe it should behave like this.

danillosl:
Because _ObjectMap is very specific for this problem (for reasons that I explain above) I don't think this should be separated into its own file. If there is a need for a _Map we should create an "abstract class" where the user could extend and override specific pieces of code (like the equality function and the hash function) to suit specific needs kind of how the Map works in java.

source/internal/_clone.js
CrossEye: I suppose you're right that this is how it's working now.  I never worked closely with this code and was thinking that we had the stronger guarantee of preserving the value equality structure (as if the above with `x === y` replaced with `equals (x, y)`).  Part of me would still like that.  But that would further degrade performance, not enhance it!

source/internal/_clone.js
CrossEye: I'm not convinced.

I'd love to see some benchmarks with different styles of input objects.

This hash seems too tuned to that input structure.

CrossEye:
> Because _ObjectMap is very specific for this problem 

That worries me.  We have an internal _Set, written to solve one problem, but available to be used in other functions, and arguably something worth making public.  Why would we not want to do the same here?  If it's too tuned to `clone`, is it also perhaps too tuned to a specific sort of data?

> If there is a need for a _Map we should create an "abstract class" where the user could extend and override specific pieces of code (like the equality function and the hash function)

The only reason I can think of for exposing these would be that they are useful alternatives, for those who like Ramda's value equality model, to the built in `Set` and `Map` types.  I would not expect to offer users ways to tweak their behavior.

source/internal/_clone.js
danillosl: But `(x === y); //=== true` then `equals(x,y) //must also be true`. So by preserving reference equality we are also preserving value equality.

CrossEye: I was suggesting that we might go further than that, but on further reflection I realized that it's not only unnecessary but actively harmful to the goal here.

source/internal/_clone.js
danillosl: After further inspection, I realize I made a mistake on my hash function, instead of using just the value I should use the propertyName + value as the hash, but regardless, the idea still the same.
About the benchmarks with different inputs I used the following array:
````js
const obja = { foo: "foo1", bar: "bar1" };
const objb = { foo: "foo2", bar: "bar2" };
const objc = { foo: "foo3", bar: "bar3" };
const objd = { foo: "foo4", bar: "bar4" };


const arr = [
  { obj1: obja, obj2: { foo: "foo1", bar: "bar1" }, obj3: objc, obj4: { foo: "foo2", bar: "bar2" }, obj5: obja, obj6: objb },
  { obj1: obja, obj2: { foo: "foo1", bar: "bar1" }, obj3: objd, obj4: { foo: "foo2", bar: "bar2" }, obj5: obja, obj6: objb },
  { obj1: obja, obj2: { foo: "foo1", bar: "bar1" }, obj3: objc, obj4: { foo: "foo2", bar: "bar2" }, obj5: obja, obj6: objb },
  { obj1: obja, obj2: { foo: "foo1", bar: "bar1" }, obj3: objd, obj4: { foo: "foo2", bar: "bar2" }, obj5: obja, obj6: objb },
];
````

and got the following results:

max/min/avg time in ms after 100 runs
ramda: new clone:             max: 0.815 min: 0.030 avg: 0.059 
ramda: old clone:             max: 0.816 min: 0.029 avg: 0.063 
lodash cloneDeep():           max: 1.913 min: 0.038 avg: 0.114 
JSON.parse(JSON.stringify()): max: 0.804 min: 0.013 avg: 0.030 

Because it is a small object to be cloned it is the best-case scenario for the old clone() but still, the new one shows better results, it only gets better as the object size goes up.
I am happy to test other examples of objects or even show the code that I'm using to benchmark so the numbers could be confirmed.

danillosl:
>  Why would we not want to do the same here? If it's too tuned to `clone`, is it also perhaps too tuned to a specific sort of data?

I wouldn't say it is too tuned for specific data but for this specific problem, _ObjectMap is focused on reference equality which is ideal for this problem, but I think it won't be as useful to have as a separate file in ramda because the framework has a focus on value equality.

>  I would not expect to offer users ways to tweak their behavior.

When I said user I meant the person that would use the code, not the framework user. What I'm proposing is to create an "abstract class" lest say a _HashTable.js and then create two classes that will inherit from _hashTable one focused on value equality and the other focused on reference equality.

GingerPlusPlus:
Closes #2607.

nfantone:
Any movements here? What's the hold up? Can I help in any way?

CrossEye:
@nfantone: We are way overdue for a release.  And we should get this into it, if it's ready.  I will try to take a look soon.

adispring:
> @nfantone: We are way overdue for a release. And we should get this into it, if it's ready. I will try to take a look soon.

We should publish a new release, the last release (v0.27.1) is 10 months ago: https://www.npmjs.com/package/ramda .

nfantone:
@CrossEye @adispring Would be happy to contribute in any way I can to help make that happen.

CrossEye:
When Ramda went through a period of little attention, this was inadvertently dropped.  @danillosl: Are you interested in resolving the conflicts so that we can get this in before `v1.0`?

customcommander:
@CrossEye Rebased and fixed the conflicts. However I'd prefer if you could review this again thoroughly as I'm not quite sure I got everything right. Some other pull requests merged after this one touched the same files and I wanted to preserve them too.

customcommander:
@CrossEye Merge, lint and tests error fixed. Ready for re-review.

CrossEye (approved):

CrossEye:
I'm going to merge this, but tag it for follow-up, as I think we still need to think about exposing `_Set` and it would make sense to have a version of `_Map` to go with it.

PR #2986, opened 2020-02-18
hasIn: protection against undefined object

ku8ar:
sometimes people write code like this:
```
let obj = null
hasIn('test', obj) && prop('test', obj)
```


CrossEye:
Yes.  We're slowly allowing data-verification to slip into the library.  But this passes my smell test: *if we think of the function as answering a question, can we reasonably answer it given this bad data?*  Yes, we know that `hasIn(foo, null)` should be `false`.  So I"m all for this.

Issue #2980, opened 2020-02-12
Calling functions in succession


anomaaly:
Uncurry seemed like the best way to accomplish this `(a, b) => f(a)(b)`

`
const cache = {
  u: {id:'1'},
  v: {id:'2'},
  w: {id:'3'},
  x: {id:'4'},
  y: {id:'5'},
  z: {id:'6'},
}
const ids = {
  a: ['x','y','z'],
  b: ['u','v','w']
}
const fx = compose(map, flip(prop));

const foo = compose(mapObjIndexed, fx);
//foo(cache)(ids) //works

const bar = uncurryN(2, foo)
bar(cache, ids) //just returns a function
`
If we were given the function `foo`, say from a library, is there a way to accomplish this using ramda?

CrossEye:
Perhaps the documentation for `uncurry` should be expanded.  The single-word parenthetical remark in 

> Returns a function of arity `n` from a (manually) curried function.

may not be explicit enough to make it clear that this doesn't work with Ramda-curried functions.

I honestly don't remember if this was simply fall-out of the implementation or something fundamental, and if you want to look into changing it, please feel free.

Of course, you can always do this:

```js
const bar = (cache, ids) => foo (cache) (ids)
```

so you can certainly work around this.

----------

Also, a quick question:

Why would you do 

```js
const fx = compose(map, flip(prop));
const foo = compose(mapObjIndexed, fx);
```

instead of just 

```js
const foo = compose (mapObjIndexed, map, flip(prop))
```

?

adispring:
I have just added some more description about `uncurryN` before:

https://github.com/ramda/ramda/blob/d70b3cd6e59c4cffddb8a263712187995a63fe97/source/uncurryN.js#L6-L9

Maybe `manually` can be expanded as `accept only one argument each time` ?

CrossEye:
@adispring: I'm not seeing that change anywhere.

adispring:
It's on the master branch, maybe have not published in v0.27.0:

https://github.com/ramda/ramda/blob/master/source/uncurryN.js

anomaaly:
The work-around is what I had originally, which is what prompted this question. Maybe `uncurry` was never meant to solve this problem, but it seems like there should be basic function that does the following:

```
const callSuccessively = (f, ...args) => {
    while(a = args.shift()) {
         f = f(a);
    }
    return f;
};
```

[...]

PR #2987, opened 2020-02-18
new function: recall

anomaaly:
This is in reference to #2980. It supports a particular class of function: A unary function that if returns a function, is also unary with the same rule.

### DIFF
--- a/source/index.js
+++ b/source/index.js
@@ -177,6 +177,7 @@ export { default as propOr } from './propOr';
 export { default as propSatisfies } from './propSatisfies';
 export { default as props } from './props';
 export { default as range } from './range';
+export { default as recall } from './recall';
 export { default as reduce } from './reduce';
 export { default as reduceBy } from './reduceBy';
 export { default as reduceRight } from './reduceRight';
--- a/source/recall.js
+++ b/source/recall.js
@@ -1,0 +1,34 @@
+import uncurryN from './uncurryN';
+
+/**
+ * Collapses a high-order unary function. The order N of the function
+ * is implicit. A maximum number of N values are singularly fed.
+ *
+ * @func
+ * @category Function
+ * @sig ((a -> b -> * -> c), *...) -> d
+ * @param {Function} fn The N-order function to be collapsed.
+ * @param {...*} args Up to N input values.
+ * @return {*}
+ * @see R.call
+ * @example
+ *
+ *      const xs = [{a: 1}, {a: 2}, {a: 3}];
+ *      const f = R.compose(R.find, R.propEq('a'));
+ *      R.recall(f, 2, xs); //=> {"a": 2}
+ *
+ * @symb R.recall(f, a, b) = f(a)(b)
+ */
+var recall = uncurryN(2, function(fn) {
+  return function() {
+    var numArgs = arguments.length;
+    var value = fn;
+    var i = 0;
+    while(typeof value == 'function' && i < numArgs) {
+      value = value(arguments[ i++ ]);
+    }
+    return value;
+  };
+});
+
+export default recall;
--- a/test/recall.js
+++ b/test/recall.js
@@ -1,0 +1,22 @@
+var R = require('../source');
+var eq = require('./shared/eq');
+
+var fn = function(a) {
+  var x = a + 1;
+  return function(b) {
+    var y = x * b;
+    return function(c) {
+      return y * c;
+    }
+  }
+};
+
+describe('recall', function() {
+  it('collapses a high-order unary function', function() {
+    eq(R.recall(fn, 1, 2, 3), 12);
+  });
+
+  it('accepts no more than N arguments for an N-order function', function() {
+    eq(R.recall(fn, 1, 2, 3), R.recall(fn, 1, 2, 3, 4));
+  });
+});


source/recall.js
```diff
@@ -0,0 +1,35 @@
...
+ * @see R.call
+ * @example
+ *
+ *      const xs = [{a: 1}, {a: 2}, {a: 3}];
+ *      const f = R.compose(R.find, R.propEq('a'));
+ *      R.recall(f, 2, xs); //=> {"a": 2}
+ *
+ * @symb R.recall(f, a, b) = f(a)(b)
+ */
+var recall = uncurryN(2, function(fn) {
```
CrossEye: I'm extremely loathe to add new variadic functions to Ramda.

I would much rather deal with an API such as this:

```js
recall (f) (a, b) //=> f (a) (b)
```

(Even that probably isn't possible.  I'm quite certain that we will need to supply the final arity.  But that's a different comment.)

I know that the generated function is going to be polyadic.  That's the whole point.  And it's not even as though it always returns functions of a specific arity.  But I'm hoping to make `pipe`, `compose` and other leftover variadic functions into fixed-arity ones.

source/recall.js
```diff
@@ -0,0 +1,35 @@
...
+  return function() {
+    var numArgs = arguments.length;
```
CrossEye: I'm quite sure that we need to supply the arity here as we do with `uncurryN`.

What happens if our target output is a function?  What if we want to convert

```js
const makeHandler = (a) => (b) => (evt) => ({do: 'something', with: 'a, b, and evt'})
```

so that it can be used as 

```js
something .addHandler (recall (makeHandler, 42, false) )
```

where `addHandler` accepts a function?

test/recall.js
```diff
@@ -0,0 +1,22 @@
...
+
+describe('recall', function() {
+  it('collapses a high-order unary function', function() {
+    eq(R.recall(fn, 1, 2, 3), 12);
+  });
+
+  it('accepts no more than N arguments for an N-order function', function() {
+    eq(R.recall(fn, 1, 2, 3), R.recall(fn, 1, 2, 3, 4));
+  });
+});
```
CrossEye: I don't feel that this tests proves what it claims to prove.  Perhaps with an innermost function in `fn` that has an optional parameter and acts differently based on its presence.

```js
var fn = function(a) {
  var x = a + 1;
  return function(b) {
    var y = x * b;
    return function(c, d = 0) {
      return y * c + d;
    };
  };
};
```

where

```js
fn (1) (2) (3)          //=> 12
fn (1) (2) (3, 4)       //=> 16
recall (fn, 1, 2, 3, 4) //=> 12
```

But then that makes me wonder if this is the correct behavior.  It seems to me that if we do this as `:: Int -> (a -> b -> ... -> n -> x) -> ((a, b, ..., n) -> x)`, then we could write this so that 

```js
recall (3, fn) (1, 2, 3, 4) //~> fn (1) (2) (3, 4)
``` 

which feels more in keeping with other Ramda functions.

CrossEye:
Thank your for the PR.  I'm afraid I forgot to do the research into why `uncurryN` doesn't already handle this.  I will try to do so soon.

Did you look into updating `uncurryN` to also handle this case?  Because if there's not some compelling reason not to do so, it seems that would be a much better solution.

source/index.js
```diff
@@ -177,6 +177,7 @@ export { default as propOr } from './propOr';
 export { default as propSatisfies } from './propSatisfies';
 export { default as props } from './props';
 export { default as range } from './range';
+export { default as recall } from './recall';
```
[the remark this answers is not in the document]
anomaaly: I'm terrible at naming. The idea is that it calls, then re-calls, etc. `collapse` also works.

source/recall.js
anomaaly: I feel the same way, I was just mirroring `call`. The list of values can be passed as an array instead, as in `apply`.

source/recall.js
anomaaly: This is a good use. The value returned can be whatever it needs to be, including just another function. If we supply a function `a => b => c => d => a + b * c + d` and want a function a particular depth, we just provide values up to that depth. For example, `[ 2, 3 ]` if we want the function `c => d => 5 * c + d`.

test/recall.js
anomaaly: It's intended to support a special class of high-order function, those composed of unary functions. E.g. `const f = g => h => v`, where `f`, `g`, and `h` are unary functions. In `compose`, all functions must be unary except the last because the inputs to the last function are outside the workings of `compose`. Similarily, the last value of the high-order unary function can be anything, including a polyadic function, because it's considered to be the value and outside the workings of `recall`. The tests themselves could use some work though, I agree.

anomaaly:
@CrossEye No problem. I gave a cursory look into `uncurryN`. It seemed fine. I didn't notice anything glaring and figured I'd need to dig further and see how currying was being handled. I'll need to make some time for that.

source/recall.js
CrossEye: Yeah, the `call` API was an early mistake as we felt our way around this space.

source/recall.js
CrossEye: I think I misread before and didn't notice `&& i < numArgs`.  My initial reading suggested that this use case wouldn't be possible with that implementation.

test/recall.js
CrossEye: Yes, I was responding to the test case.  I don't think it really demonstrated that it "accepts no more than N arguments for an N-order function". But I am curious about which behavior we actually want.  I'm thinking that it might be a little different from what's in this implementation.

source/recall.js
anomaaly: The idea is to call a function with a set of values in the fashion of `call` and `apply`. The difference here is there is no notion of arity since the n-order function is composed of unary functions. The list of values supplied doesn't represent an arity of a function, rather a serial list of singular inputs fed to the n-order function. So returning a function defeats the purpose in which case I'd rather just use `uncurryN` somehow.

CrossEye: Perhaps then we're at cross purposes.  I was thinking that the goal was a to bring behavior similar to `uncurryN` to work with Ramda-styles curried functions. I thought we were mostly just removing the word "manually" from the docs.

I'm not sure what your goal is hers.

anomaaly: Ideally `uncurryN` would be made to work but I'd still want this functionality, otherwise I'd have just submitted a patch. Is anyone looking into it?

CrossEye: I don't think so.  I've been pulled into something else, and haven't had any real free time for Ramda in the past week.  If no one does it before then, I will try to get to it this weekend.  (You can do it yourself.  Any conversations of importance about this would be in the issues here; so a GitHub search should probably show anything of importance that isn't already captured in the source or the tests.)

PR #2989, opened 2020-02-20
add @see: R.view and R.set

adispring:
`R.view` and `R.set` are Lens's getter and setter functions. Add @see to each other may bring some convenience for people to read.

### DIFF
--- a/source/over.js
+++ b/source/over.js
@@ -22,7 +22,7 @@ var Identity = function(x) {
  * @param {*} v
  * @param {*} x
  * @return {*}
- * @see R.prop, R.lensIndex, R.lensProp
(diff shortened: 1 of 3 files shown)


Bradcomp:
Thanks @adispring! I think it might be worthwhile to remove `R.prop` from the `@see` list too, as they don't really fit here. Thoughts?

CrossEye:
Yes on the change, yes to @Bradcomp's extension, but there's one thing missing.  We should also include `over` among these.  `view`, `set`, and `over` form a useful trio.

adispring:
@Bradcomp @CrossEye 
Ok, I have removed `R.prop` from their @see lists, and add two more apis:  `R.lens` and `R.lensPath`.

CrossEye:
:herb:

Looks great to me!

PR #2993, opened 2020-03-08
`pipe` - throw on non-function arguments

brekk:
Hey there, I was recently bitten by this -- I had a function composition that I thought looked like:

```js
pipe(
  curriedTernaryFunction(a, b),
  someOtherFn,
  anotherFn
)('blah')
```

But the function that I thought was ternary was actually binary, so instead of being a function it was the result of calling that function inside the pipe. This results in an error `TypeError: g.call is not a function.` -- as I was several levels above this it took a while to chase this down.

I'd love if this or something similar could land in `ramda` as it is a core library for me, also happy to make any changes.

### DIFF
--- a/source/pipe.js
+++ b/source/pipe.js
@@ -1,7 +1,10 @@
 import _arity from './internal/_arity';
+import _isFunction from './internal/_isFunction';
 import _pipe from './internal/_pipe';
+import complement from './complement';
 import reduce from './reduce';
 import tail from './tail';
+import toString from './toString';
 
 
 /**
@@ -31,6 +34,12 @@ import tail from './tail';
 export default function pipe() {
   if (arguments.length === 0) {
     throw new Error('pipe requires at least one argument');
+  }
+  const firstNonFunction = [].slice.call(arguments).find(complement(_isFunction));
+  if (firstNonFunction) {
+    throw new TypeError(
+      'pipe: ' + toString(firstNonFunction) + ' is not a function'
+    );
   }
   return _arity(
     arguments[0].length,
--- a/source/pipe.js
+++ b/source/pipe.js
@@ -1,7 +1,6 @@
 import _arity from './internal/_arity';
 import _isFunction from './internal/_isFunction';
 import _pipe from './internal/_pipe';
-import complement from './complement';
 import reduce from './reduce';
 import tail from './tail';
 import toString from './toString';
@@ -35,11 +34,15 @@ export default function pipe() {
   if (arguments.length === 0) {
     throw new Error('pipe requires at least one argument');
   }
-  const firstNonFunction = [].slice.call(arguments).find(complement(_isFunction));
-  if (firstNonFunction) {
-    throw new TypeError(
-      'pipe: ' + toString(firstNonFunction) + ' is not a function'
-    );
+  var args = [].slice.call(arguments);
+  var idx = 0;
+  while (idx < args.length) {
+    if (!_isFunction(args[idx])) {
+      throw new TypeError(
+        'pipe: ' + toString(args[idx]) + ' is not a function'
+      );
+    }
+    idx += 1;
   }
   return _arity(
     arguments[0].length,
--- a/test/pipe.js
+++ b/test/pipe.js
@@ -56,4 +56,13 @@ describe('pipe', function() {
     eq(g(1, 2, 3), [1, 2, 3]);
   });
 
+  it('throws given non-function arguments', function() {
+    assert.throws(
+      function() { R.pipe(x => x, x => x, 'not-a-function'); },
+      function(err) {
+        return err.constructor === TypeError &&
+               err.message === 'pipe: "not-a-function" is not a function';
+      }
+    );
+  });
 });


brekk:
Also, there is a `propOr` test on this branch which is failing but on some code I have not touched.

Bradcomp:
@brekk I've reverted the failing test on master. Can you rebase this? It should pass CI now.

source/pipe.js
```diff
@@ -32,6 +32,9 @@ export default function pipe() {
   if (arguments.length === 0) {
     throw new Error('pipe requires at least one argument');
   }
+  if ([].slice.call(arguments).filter(fn => typeof fn !== 'function').length > 0) {
```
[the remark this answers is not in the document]
brekk: Sure, I'd be happy to take a look for a common parallel. Otherwise I can use `find` here instead, as it short-circuits early?

brekk:
@Bradcomp Thanks for the feedback, I hope the newest stuff is to your liking.

I dug around in the codebase, and there's an established precedent for the more verbose `Array.prototype.slice.call` in a few places -- `liftN` / `call` / `useWith` / `invoker` to name a few. I'd be happy to make my code `[].slice.call` consistent with that if preferred.

On the formatting of the error, I used your suggestion mainly, but added a prefix of `pipe:` to the error string (which has a precedent in `_internal/reduce.js`) as the error.stack downstream is often-manipulated but the error message rarely is.

source/pipe.js
```diff
@@ -32,6 +35,12 @@ export default function pipe() {
   if (arguments.length === 0) {
     throw new Error('pipe requires at least one argument');
   }
+  const firstNonFunction = [].slice.call(arguments).find(complement(_isFunction));
```
brekk: I wasn't sure whether I should be using ramda functions for this internal check, so I'd be happy to make this `ff => !_isFunction(ff)` if that's preferable.

CrossEye: Those internal functions are there precisely to be used in places like this, so... yes!

Bradcomp: We can't use Arrow functions until we drop support of ES5. I think `complement` is OK here.

Bradcomp:
> I dug around in the codebase, and there's an established precedent for the more verbose Array.prototype.slice.call in a few places -- liftN / call / useWith / invoker to name a few. I'd be happy to make my code [].slice.call consistent with that if preferred.

This would definitely help, but isn't quite what I was getting at. Typically with `arguments` we are just passing them along to another function using apply or call. That happens in a really large number of our functions, and how we do that should definitely be consistent. But here we are iterating through them and performing [generic] validation for each argument. The iteration and validation is the piece I'd like to discuss, because to my knowledge (unless we find something) that isn't done in our codebase.

brekk:
@Bradcomp  Ah ok; I see the distinction you're making — in my mind composition is a very specific operation on functions and this PR is just a way of ensuring that `pipe` is operating on functions — it's not generic validation so much as basic type safety; would there be a better way of approaching this problem which would ameliorate your concerns?

source/pipe.js
Bradcomp: For me at least I'd like to see it more similar to how we perform list iterations in general insead of using the built in Array functions.

what about this:
```
var args = Array.prototype.slice.call(arguments);
var idx = 0;
  while (idx < args.length) {
    if (!_isFunction(args[idx])) {
          throw new TypeError(
            'pipe: ' + toString(args[idx]) + ' is not a function'
    );
    }
    idx += 1;
  }
return blah;
```

@brekk @CrossEye @team?

source/pipe.js
```diff
@@ -32,6 +34,16 @@ export default function pipe() {
...
+  var args = [].slice.call(arguments);
+  var idx = 0;
+  while (idx < args.length) {
+    if (!_isFunction(args[idx])) {
+      throw new TypeError(
+        'pipe: ' + toString(args[idx]) + ' is not a function'
+      );
+    }
+    idx += 1;
+  }
```
brekk: @Bradcomp Sorry for the delay! How does this feel?

source/pipe.js
brekk: Updated.

CrossEye:
I'm sorry.  I somehow didn't pay much attention to this as it moved along.

And I'm less than certain that it's a good idea.  Ramda has generally not added type-checking unless we can do something more useful with the results than throwing an exception.  Perhaps it's time to revisit our let-it-fail philosophy, but I think we would need to have such a discussion before merging this.

brekk:
@CrossEye I'm fine with literally any solution that doesn't leave me asking what `g.call is not a function` is referring to (thrown from `internal/_pipe`). When you're several levels above the piped function, it's not especially clear what `g.call` refers to. And I understand where y'all are at w/r/t let-it-fail, but this isn't very usable as is -- it's not type-checking in my mind so much as the basic requirements for composition: the composition of a function and a non-function is unresolvable garbage.

If you're not into this PR, I'm happy to publish it as a standalone library, but at that point I'll just fork all of `ramda` rather than use two different libraries for the same purpose (which I am loathe to do, as `ramda` solves nearly all problems out of the box).

CrossEye:
@brekk: Please see the discussion in #2998.

danielbom:
Has some discussion about enable some basic type checking, like when create sanctuary env? Ex.

file to expose ramda configuration:
```javascript
let R = require("ramda");
const basicCheck = require("ramda-basic-check"); // example lib

// Add basic assertion
If (process.env.NODE_ENV === "development") {
  // based of "ramda-debug" configuration
  R = basicCheck.wrap(R);
  // or made this something internal, I don't know.
}

module.exports = R;
```

CrossEye:
@danielbom:  Please see the discussion in #2998.

customcommander:
@CrossEye Close for now until #2998 is resolved?

Issue #2974, opened 2020-02-07
Unexpected `pathOr` return value while path is [-1] (in version 0.27)


Andrey-Bazhanov:
Code bellow will return `3` in ramda v0.27. In v0.26 it worked as expected and returned `null`

```
const arr = [1, 2, 3];
const result = pathOr(null, [-1], arr);
```

Here a demo: https://codesandbox.io/s/agitated-archimedes-3ktv6

adispring:
Because `R.pathOr`'s implementation  depends on `R.path`, and `R.path`'s implementation changed in 0.27.0. 

This is `R.path` 0.26.1's implementation: https://github.com/ramda/ramda/blob/v0.26.1/source/path.js#L29

and this is `R.paths` 0.27.0's implementation: https://github.com/ramda/ramda/blob/ed191e6a476330e37db259a5c2d04bfba0b2d63d/source/paths.js#L32

The object path rule changes:

```js
// 0.26.1
val = val[paths[idx]];

// 0.27.0
      p = paths[idx];
      val = _isInteger(p) ? nth(p, val) : val[p];
```

CrossEye:
I really wish we'd never done anything to support negative indices.  I think they end up a useful convenience that makes too little logical sense.

Bradcomp:
This is a change in behavior. Should we revert and discuss?

CrossEye:
@Bradcomp: This has been bouncing around far too much.  The reason for those changes was a drastic performance improvement.  But now there is one more thing to worry about.  How interested are you in adding some of those property-based tests we've discussed?  That might help us clean some of this up.

Bradcomp:
We can keep the performance gain and drop the change by not using `nth`, instead just using a regular property access. 

I think property tests could help us to get this stuff locked down once we come up with the properties we want to satisfy.

I think we need to shave a few yaks first.

It sounds like one question is whether to support negative indices in functions that do property access on Arrays, the other is what all functions do we want to 'track' path / prop. I'd rather settle those questions first and then get it all taken care of for existing functions. 

It's very likely this will lead to breaking changes to get everything consistent.

ku8ar:
what was the point of supporting negative values at all? Any specific use case? Because it's a bit funny. Probably almost nobody uses it.

CrossEye:
[...]

PR #2995, opened 2020-03-09
Standardize prop or and path or functionality to match documentation

Tworley96:
Follow up to #2888 
Reverts change to propOr that caused issue in test suite


ku8ar:
It all goes the wrong way. Adding `nth` support to `prop`/`path` causes unnecessary complexity. And it's very inefficient...

CrossEye:
@ku8ar:

Do you have suggestions.  I think the main constraints are that things like `path(foo)` and `path([foo])` should always have the same behavior, and that all of these, but especially the simplest ones like `path` should be as performant as possible.

ku8ar:
@CrossEye Yes of course. I added a suggestion here: https://github.com/ramda/ramda/issues/2974#issuecomment-602273378
That suggestion will simplify this code. In short: remove `nth` from `prop` and `path`. But this is a hard decision, considering that there is already version 0.27 with this feature...

CrossEye:
I have absolutely no problem changing the behavior back in the next version if it turns our we made a mistake.  I don't have much time now to investigate *why* this was added, and we need to do that before deciding whether it's worth the costs it adds.

Issue #2991, opened 2020-02-28
In React app: Cannot read property 'fantasy-land/reduce' of undefined


trentlarson:
I just added ramda in my React app and I'm able to use it in my tests, but after bundling and hitting a page with it in a browser I get the following:

```_reduce.js:54 Uncaught TypeError: Cannot read property 'fantasy-land/reduce' of undefined
    at _reduce (_reduce.js:54)
    at Module.f3 (_curry3.js:46)
    at createStepsById (workflowUtils.js:20)
    at Workflow (Workflow.jsx:96)
    at renderWithHooks (react-dom.development.js:16260)
    at mountIndeterminateComponent (react-dom.development.js:18794)
    at beginWork$1 (react-dom.development.js:20162)
    at HTMLUnknownElement.callCallback (react-dom.development.js:336)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:385)
    at invokeGuardedCallback (react-dom.development.js:440)```

I've run `npm install` and even `npm ci`.

package.json is attached
[package.json.txt](https://github.com/ramda/ramda/files/4268283/package.json.txt)

I didn't see any other "fantasy-land" issues that looked similar to this.  Any help is appreciated.  Thanks.

trentlarson:
I've tried different ways of importing:

const R = require('ramda');
import R from 'ramda'
import * as R from 'ramda';
import { assoc, map, reduce } from 'ramda';

Bradcomp:
Ramda is being imported fine here, but there is an error in the way reduce is being called. Can you provide some example code?

Given the stack trace, I would start by looking here: `workflowUtils.js:20`

kiyutink:
i think @trentlarson 's usage is something like this: https://ramdajs.com/repl/?v=0.27.0#?R.reduce%28R.sum%2C%200%2C%20undefined%29 ; The usage of react is not relevant though.

I've noticed the same problem with R.map: https://ramdajs.com/repl/?v=0.27.0#?R.map%28R.identity%2C%20undefined%29

Doesn't look like this behavior is intended

CrossEye:
@kiyutink:

That sounds like a good guess.  While this behavior is not particularly intended, Ramda has always taken a no-hands-holding approach.  If you use it as documented, it should do as you expect.  But if you pass unexpected data (such as `undefined` where an object, array, function, or fantasy-land-compatible functor is expected), then anything can happen.

[...]

PR #2997, opened 2020-03-22
reduce - null safety

ku8ar:
fix this bug: https://github.com/ramda/ramda/issues/2991

I also get sometimes this bug in project.
All in all, it came to the situation that we started  in the team writing `defaultTo` before each `reduce`:

```js
const getRating = pipe(
 defaultTo([]),
 reduce(add, 0)
)
```
After fix we can just write:

```js
const getRating = reduce(add, 0)
```

We use Ramda to select values from API, so null/undefined sometimes can happen...

@CrossEye fix is very efficient. In the case of happy path, it doesn't matter.
Before comparing to null, the `reduce` function uses type checking in `_isArrayLike`, so V8 already knows the type (hidden class)

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,6 +239,8 @@ Translations
 - [Ukrainian(Українська)](https://github.com/ivanzusko/ramda)
 - [Portuguese(BR)](https://github.com/renansj/ramda)
 - [Russian(Русский)](https://github.com/Guck111/ramda)
+- [Spanish(ES)](https://github.com/wirecobweb/ramda)
+
 
 
 
--- a/source/hasIn.js
+++ b/source/hasIn.js
@@ -1,4 +1,5 @@
 import _curry2 from './internal/_curry2';
+import isNil from './isNil';
 
 
 /**
@@ -28,6 +29,9 @@ import _curry2 from './internal/_curry2';
  *      R.hasIn('area', square);  //=> true
  */
 var hasIn = _curry2(function hasIn(prop, obj) {
+  if (isNil(obj)) {
+    return false;
+  }
   return prop in obj;
 });
 export default hasIn;
--- a/source/internal/_reduce.js
+++ b/source/internal/_reduce.js
@@ -43,6 +43,9 @@ export default function _reduce(fn, acc, list) {
   if (_isArrayLike(list)) {
     return _arrayReduce(fn, acc, list);
   }
+  if (list == null) {
+    return acc;
+  }
   if (typeof list['fantasy-land/reduce'] === 'function') {
     return _methodReduce(fn, acc, list, 'fantasy-land/reduce');
   }
--- a/source/over.js
+++ b/source/over.js
@@ -22,7 +22,7 @@ var Identity = function(x) {
  * @param {*} v
  * @param {*} x
  * @return {*}
- * @see R.prop, R.lensIndex, R.lensProp
+ * @see R.view, R.set, R.lens, R.lensIndex, R.lensProp, R.lensPath
  * @example
  *
  *      const headLens = R.lensIndex(0);
--- a/source/set.js
+++ b/source/set.js
@@ -17,7 +17,7 @@ import over from './over';
  * @param {*} v
  * @param {*} x
  * @return {*}
- * @see R.prop, R.lensIndex, R.lensProp
+ * @see R.view, R.over, R.lens, R.lensIndex, R.lensProp, R.lensPath
  * @example
  *
  *      const xLens = R.lensProp('x');
--- a/source/view.js
+++ b/source/view.js
@@ -19,7 +19,7 @@ var Const = function(x) {
  * @param {Lens} lens
  * @param {*} x
  * @return {*}
- * @see R.prop, R.lensIndex, R.lensProp
+ * @see R.set, R.over, R.lens, R.lensIndex, R.lensProp, R.lensPath
  * @example
  *
  *      const xLens = R.lensProp('x');
--- a/test/hasIn.js
+++ b/test/hasIn.js
@@ -26,4 +26,8 @@ describe('hasIn', function() {
     eq(R.hasIn('name', anon), false);
   });
 
+  it('returns false when non-existent object', function() {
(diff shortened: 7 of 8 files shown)


CrossEye:
This looks good to me.

Anyone else?  @ramda/core?

My test for whether we can go beyond our regular no-hands-holding approach is whether we have enough information to give a reasonable answer.  Here like with `defaultTo` I think we do.  The accumulator is what we should always return when we have no data.

Bradcomp:
I do think this is reasonable, but I am hesitant. My concern is less of a hands holding one and more of an 'avoid magic' one. 

Any time we make an assumption about what our user means based on what they don't tell us, we are hiding a potential error that could result in a bug. 

In most cases, if `reduce` is called with `undefined` as the foldable, it means the developer has made a mistake. It could very well mean that they will always get undefined there instead of whatever they are expecting to get. Is it helpful to hide this from the user?

ku8ar:
@Bradcomp The question of how Ramda is used. For frontend/mobile this library is practical because some functions are null safety.
In addition (as @CrossEye mentioned), `reduce` differs from `map`/`filter` in that it has the initial state (accumulator) given.

Of course, you can take the second approach, in which Ramda is the equivalent of the Haskell lens.
But in this case, why the types are checked anywhere? Why `prop(null, {})` doesn't return an error?
In this approach, we should also not show "fantasy ground issue" errors (this is very magical error for some developers), but something more usefull, e.g "list is null" ;)

davidchambers:
> In this approach, we should also not show "fantasy ground issue" errors (this is very magical error for some developers), but something more usefull, e.g "list is null" ;)

This is absolutely the correct approach, in my view. :)

CrossEye:
@Bradcomp:

My feeling is that we generally keep the Garbage In - Garbage Out mantra.  If you supply types of data not supported by Ramda's contracts, then you should not count on the behavior whatsoever.  But as developers and users of the library, we should also want to prevent unexpected exceptions as best we can.  When we can't, I feel fine letting it fail in whatever way it happens to fail -- throwing an exception, launching the missiles, insulting your children.  But when we can, I think it's best that we do. 

I edited my previous comment, because I mentioned `defaultTo`, when I should have been speaking of `propOr`.  In `propOr` we know exactly what result the user wants if the property is nil.  We could throw an exception in the case the object is nil, and that would certainly fit best with our let-it-fail philosophy.  But that seems to ignore what the user told us: that if the property isn't there, we should return the default.  It seems to cause almost no harm to return the default.  It will serve as a failsafe for those who want it.  And for those who don't, we already supply no guarantees of behavior when they supply bad data; so we haven't hurt them.  If they wanted a specific error in this case, then they would need to generate it on their own.

And that's how I feel about this one too.  By supplying the initial accumulator, the user has already told us what to return if there's no data.  We hurt no one and help some by returning this on a nil `list`.

I guess I don't think of it as magic so much as giving the best answer we can to the question asked.  The data is simply not bad enough that we have to throw our hands up.

@ku8ar, @davidchambers:

While I would certainly not object to having better error messages, I'm quite loathe to spend the effort on the code required to implement them or to take the performance hit of all the checks necessary to report them.  Do you have suggestions that might ease my fears?

davidchambers:
> Do you have suggestions that might ease my fears?

Ramda *could* depend on [sanctuary-def][1]. This would add complexity, though, as type checking would need to be configurable.


[1]: https://github.com/sanctuary-js/sanctuary-def

CrossEye:
@davidchambers:

> Ramda *could* depend on [sanctuary-def][sd]. This would add complexity, though, as type checking would need to be configurable.

And it would add five dependencies (`sanctuary-def` itself as well as the recursive dependencies of `sanctuary-either`, `sanctuary-show`, `sanctuary-type-classes`, and `sanctuary-type-identifiers`) to a library that now has *no* run-time dependencies.  So far, I'm still unwilling to go that route.

[sd]: https://github.com/sanctuary-js/sanctuary-def

CrossEye:
Please see the discussion in #2998.

Bradcomp:
> But in this case, why the types are checked anywhere? Why prop(null, {}) doesn't return an error?

I believe I have asked the exact same question before. I think that is a mistake, but others disagree. Part of why it's a mistake is because it leads to these sorts of questions of how far to go and what principles do we bring to bear?

> And for those who don't, we already supply no guarantees of behavior when they supply bad data; so we haven't hurt them.

I take a little issue with this. We are removing a helpful error on a function with a clear signature and long history. I think depending on one's view of safety this definitely makes `reduce`, which is super widely used and understood, more dangerous.

This isn't GIGO. GIGO is to let it fail. This is to prevent it from failing in the case that it should.

The principal we want to look at here is probably this one: https://en.wikipedia.org/wiki/Robustness_principle

I think there could be an argument made either way here. I think that having predictable functions that do what you expect is a benefit, at least in how I tend to write code. 

```
const getRating = pipe(
 defaultTo([]),
 reduce(add, 0)
)
```
 
Personally, for me, this code is way clearer than `reduce(add, 0)` for the given use case because it breaks up the piece that handles invalid input from the piece that actually performs the calculation.

ku8ar:
> My feeling is that we generally keep the Garbage In - Garbage Out mantra.

@CrossEye perfect!

> I believe I have asked the exact same question before. I think that is a mistake, but others disagree.

@Bradcomp but I don't agree with that either...
I mean, I agree, but only partially.

In a perfect world, I would expect the perfect code. But currently Ramda is used in projects that are created quickly, by a business that expects results as soon as possible.
Errors in such projects will happen often.
And now we have a choice:
- return "null" for incorrect data
- call exception in runtime

For frontend/mobile, the "garbage in - garbage out" approach is a much better option. We do not harm. And developers have another reason to use `R.map` instead of `Array.map`.

And in my personal opinion, every exception is a violation of the principle of functional programming. "throw" is like a linear jump. The functional code should handle garbage as follow: "you gave me a null to which there is no logical answer, so please, here is your null"

CrossEye (approved):

PR #2999, opened 2020-03-26
Fix for importing Ramda in ESM-compatible nodes

### LINKED ISSUE #2988: Problem with ECMAScript modules import in nodejs and browser
Hi!

I want run the same script both in nodejs and browser.
I use latest nodejs 13.9.0 which supports ECMAScript modules. And I use create-react-app to build application for browser.

`import * as R from 'ramda';` works well in browser but in nodejs it is imported as `{ default: { ...R functions}}`

`import R from 'ramda';` works well in nodejs but in browser it returns error 
`Attempted import error: 'ramda' does not contain a default export (imported as 'R').`

I'm not a fan of default imports so if it will work as `import * as R from 'ramda';` everywhere it will be awesome.

Is it possible to do something about it? Maybe some solution or workaround?

UPD. I made a very straightforward hack but it is ugly.
```
import * as R2 from 'ramda';

let R;

// need to hide the "default" call from webpack.
// Otherwise it returns: 'ramda' does not contain a default export 
function getter(obj, prop) {
  return obj[prop];
}

if (Object.keys(R2).length === 1) {
  R = getter(R2, 'default');
} else {
  R = R2;
}
```

Andarist:
closes #2988

This all is quite new in general. node has added support for ESM just recently so keep in mind that I can't guarantee that this covers everything ideally, but I've done the research and some testing and I'm positive about this working correctly. If any problems arise from this - just ping me and I will prepare appropriate fixes (hopefully this won't be necessary though).

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,8 +45,18 @@
     "type": "git",
     "url": "git://github.com/ramda/ramda.git"
   },
-  "module": "es/index.js",
-  "main": "src/index",
+  "main": "./src/index.js",
+  "exports": {
+    ".": {
+      "require": "./src/index.js",
+      "import": "./es/index.js",
+      "default": "./src/index.js"
+    },
+    "./es/": "./es/",
+    "./src/": "./src/",
+    "./dist/": "./dist/"
+  },
+  "module": "./es/index.js",
   "unpkg": "dist/ramda.min.js",
   "jsdelivr": "dist/ramda.min.js",
   "files": [
@@ -58,12 +68,11 @@
     "prebench": "npm run --silent build:cjs",
     "bench": "node scripts/benchRunner",
     "bookmarklet": "node scripts/bookmarklet",
-    "build:es": "cross-env BABEL_ENV=es babel source --out-dir es",
+    "build:es": "cross-env BABEL_ENV=es babel source --out-dir es && node ./scripts/addModulePackageScope.js",
     "build:cjs": "cross-env BABEL_ENV=cjs babel source --out-dir src",
-    "build:mjs": "cross-env BABEL_ENV=es babel source/index.js --out-file src/index.mjs",
     "build:umd": "cross-env NODE_ENV=development rollup -c -o dist/ramda.js",
     "build:umd:min": "cross-env NODE_ENV=production rollup -c -o dist/ramda.min.js",
-    "build": "npm run build:es && npm run build:cjs && npm run build:umd && npm run build:umd:min && npm run build:mjs",
+    "build": "npm run build:es && npm run build:cjs && npm run build:umd && npm run build:umd:min",
     "partial-build": "node ./scripts/partialBuild",
     "clean": "rimraf es/* src/* dist/* coverage/*",
     "prepare": "npm run clean && npm run build",
--- a/scripts/addModulePackageScope.js
+++ b/scripts/addModulePackageScope.js
@@ -1,0 +1,6 @@
+const fs = require('fs');
+const path = require('path');
+
+const PACKAGE_SCOPE_PATH = path.join(__dirname, '..', 'es', 'package.json');
+
+fs.writeFileSync(PACKAGE_SCOPE_PATH, JSON.stringify({ type: 'module' }));


scripts/addModulePackageScope.js
```diff
@@ -0,0 +1,6 @@
+const fs = require('fs');
```
Andarist: I've not done this in shell to support Windows contributors.

package.json
```diff
@@ -58,12 +68,11 @@
     "prebench": "npm run --silent build:cjs",
     "bench": "node scripts/benchRunner",
     "bookmarklet": "node scripts/bookmarklet",
-    "build:es": "cross-env BABEL_ENV=es babel source --out-dir es",
+    "build:es": "cross-env BABEL_ENV=es babel source --out-dir es && node ./scripts/addModulePackageScope.js",
     "build:cjs": "cross-env BABEL_ENV=cjs babel source --out-dir src",
-    "build:mjs": "cross-env BABEL_ENV=es babel source/index.js --out-file src/index.mjs",
```
Andarist: I don't think this was particularly needed and should be even more obsolete thanks to the changes here.

package.json
```diff
@@ -45,8 +45,18 @@
...
     "url": "git://github.com/ramda/ramda.git"
   },
-  "module": "es/index.js",
-  "main": "src/index",
+  "main": "./src/index.js",
+  "exports": {
+    ".": {
+      "require": "./src/index.js",
+      "import": "./es/index.js",
+      "default": "./src/index.js"
```
Andarist: `"require"` and `"default"` are serving the same purpose - CJS consumers, problem is that there was an experimental version of node@13 which had support for `"import"` but had no notion of the `"require"` key, so this "duplication" handles both versions at the same time as that experimental version of node can just fallback to the `"default"` whereas the new version will match against `"require"`

package.json
```diff
@@ -45,8 +45,18 @@
...
+    },
+    "./es/": "./es/",
```
Andarist: Those are here to allow for "deep" imports like `import add from "ramda/es/add.js"`. This isn't needed per se and you could explore hiding those files entirely through listing all allowed entries in this `"exports"` field (which could result in things being accessible like this: `"ramda/add"`), but I've added this for backward compatibility for now.

package.json
```diff
@@ -58,12 +68,11 @@
     "prebench": "npm run --silent build:cjs",
     "bench": "node scripts/benchRunner",
     "bookmarklet": "node scripts/bookmarklet",
-    "build:es": "cross-env BABEL_ENV=es babel source --out-dir es",
+    "build:es": "cross-env BABEL_ENV=es babel source --out-dir es && node ./scripts/addModulePackageScope.js",
```
Andarist: the type of interpretation is set based on the nearest `package.json` file, so we can just add a `package.json` with `{"type": "module"}` to a directory to make files in that directory being treated as modules even if the root is treated as commonjs (and still use `.js` for all files!)

scripts/addModulePackageScope.js
Bradcomp: Not on Windows, but I really appreciate this consideration. It helps reduce the barrier to entry for contributing.

Bradcomp:
This looks really good to me. 

I have one concern, and that is that the changes are fairly inscrutable without your fantastic comments. Somebody who isn't already well versed in `package.json` files wouldn't know why things are set up the way they are.

Is this an issue? Should we look at [hacking in some comments](https://stackoverflow.com/questions/14221579/how-do-i-add-comments-to-package-json-for-npm-install)?

Andarist:
> Is this an issue? Should we look at hacking in some comments?

Sure, I can write this down somewhere. Not sure though if package.json itself is the right place to do so 😅 If you want to keep it there I'm OK with that - just let me know where should I put it.

Bradcomp:
I honestly don't know where we should put it `¯\_(ツ)_/¯`

CrossEye:
Anyone understand why GitHub is saying that some checks are incomplete, but that Travis is saying they are complete?

One had failed before for what looked like spurious reasons.  I restarted, and it looks fine on Travis, but GitHub is saying otherwise.  Any ideas?

Other than that, this looks great to me.  I think I would rather point to this pull than try to hack comments into `package.json`, although I really do wish that commenting JSON were easier to do.

:herb:

Bradcomp:
@CrossEye Can we just merge anyway since we can still visually verify the CI completed successfully?

CrossEye:
@Bradcomp: I was planning on giving it two days in case someone who understood what happened with Travis pointed out some flaw I didn't understand.  But yes, I think we can merge soon.

Andarist:
If this causes any problem (hopefully not) then please just ping me and i will try to find a fix for it

Issue #3001, opened 2020-03-27
filter lacks fantasy-land support v0.27


krzysztofpniak:
Filter source looks:
```js
var filter = _curry2(_dispatchable(['filter'], _xfilter, function(pred, filterable) {
  return (
    _isObject(filterable) ?
      _reduce(function(acc, key) {
        if (pred(filterable[key])) {
          acc[key] = filterable[key];
        }
        return acc;
      }, {}, keys(filterable)) :
    // else
      _filter(pred, filterable)
  );
}));
export default filter;
```
it should be(only first line changed):
```
var filter = _curry2(_dispatchable(['filter', 'fantasy-land/filter'], _xfilter, function(pred, filterable) {
  return (
    _isObject(filterable) ?
      _reduce(function(acc, key) {
        if (pred(filterable[key])) {
          acc[key] = filterable[key];
        }
        return acc;
      }, {}, keys(filterable)) :
    // else
      _filter(pred, filterable)
  );
}));
export default filter;
```

`map` implementation is correct

Bradcomp:
@krzysztofpniak Good catch! Would you be interested in opening a PR to fix this?

krzysztofpniak:
Of course

krzysztofpniak:
@Bradcomp done. I've also added the test for it.

PR #3002, opened 2020-03-27
Filter fix

krzysztofpniak:
it is the fix for of https://github.com/ramda/ramda/issues/3001


CrossEye:
Looks great!  thank you very much!

:herb:

PR #3005, opened 2020-04-06
optimize binary signature

adispring:


CrossEye:
While the current signature is awful, I'm not sure that's much of an improvement.

Maybe

```hs
:: (a -> b -> c -> ... -> z) -> (b -> a -> c -> ... -> z)
```

At some point, I'd really rather switch over to multiple signatures that show the possible patterns, with no internal ellipsis, although possibly one between or after signatures.

So that might be

```hs
:: (a -> b -> c) -> (b -> a -> c)
:: (a -> b -> c -> d) -> (b -> a -> c -> d)
:: (a -> b -> c -> d -> e) ->  (b -> a -> c -> d -> e)
:: ...
```

Bradcomp:
@CrossEye This is `binary`, not `flip`?

source/binary.js
```diff
@@ -11,7 +11,7 @@ import nAry from './nAry';
  * @memberOf R
  * @since v0.2.0
  * @category Function
- * @sig (* -> c) -> (a, b -> c)
+ * @sig (* -> c) -> ((a, b) -> c)
```
Bradcomp, line 14: I think this is a good improvement. Would this be better?

(a -> b -> c -> ... -> z) -> ((a, b) -> z)

CrossEye:
Oh damn, I'm brain-dead!  Yes, this is an improvement, and's Brad's is a bit better.

source/binary.js
adispring: rewrite `binary` & `unary`'s signatures according to your suggestion. @Bradcomp

adispring:
Maybe `nAry`'s signature should also be rewritten, but I can not figure out how to rewrite that.

Bradcomp:
:bowtie: Thanks @adispring !!!

Issue #3007, opened 2020-04-08
Unexpected `path` edge cases for objects with numbers as keys.


AndrewSouthpaw:
I proverbially stubbed my toe against some unexpected hard edges with `path` recently...

```js
path('-5', { "-5": "boo"} ) // undefined
path('5', { "5": "boo"} ) // 'boo'
path([-5], { "-5": "boo"}) // undefined
path([5], { "5": "boo"}) // 'boo'
```

([repl](https://ramdajs.com/repl/#?%0Apath%28%27-5%27%2C%20%7B%20%22-5%22%3A%20%22boo%22%7D%20%29%20%2F%2F%20undefined%0Apath%28%275%27%2C%20%7B%20%225%22%3A%20%22boo%22%7D%20%29%20%2F%2F%20%27boo%27%0Apath%28%5B-5%5D%2C%20%7B%20%22-5%22%3A%20%22boo%22%7D%29%20%2F%2F%20undefined%0Apath%28%5B5%5D%2C%20%7B%20%225%22%3A%20%22boo%22%7D%29%20%2F%2F%20%27boo%27%0A))

Granted, I'm now all too painfully familiar with how passing in `5` and `'5'` have *potentially very different behaviors* for functions that involve path, e.g. `assocPath` (we've actually created a `assocPathStr` because we mess it up so often and always want strings anyway)... but I really didn't expect the presence of a negative value to make such a difference.

Is this ... a bug? Expected?

adispring:
According to `path`'s signature

```js
[Idx] → {a} → a | Undefined
Idx = String | Int
```

`path` only accepts array as its first parameter, and the item in the array should be either `String` or `Int`. So the the following writings are wrong:

```js
path('-5', { "-5": "boo"} ) // undefined
path('5', { "5": "boo"} ) // 'boo'
```

If item in the array is an negative Integer, then `path`‘s second parameter related part should be an array. 

`path` implemented by `paths` and `nth`, this two apis support (negative) number, and already cause some problems.

CrossEye:
The only one of those examples that feels even slightly surprising is

```js
path([-5], { "-5": "boo"}) // undefined
```

I suppose that our signature should really be 

```hs
:: [Idx] → {a} → a | Undefined
:: Idx = String | NonNegativeInt
```

What happens if you don't pass an array for the path is undefined, and, as always,  *caveat emptor*. 

This is actually the most interesting one:

```js
path([5], { "5": "boo"}) // 'boo'
```

[...]

PR #3010, opened 2020-04-10
Add docs to `path`.

AndrewSouthpaw:
Addresses discussion and suggestions in #3007 to clarify documentation of `path`.

I'm not sure if any of these doc edits would be welcome, but I thought I'd at least offer them. Feel free to tweak or reject if it doesn't match the style.

### DIFF
--- a/source/path.js
+++ b/source/path.js
@@ -2,7 +2,9 @@ import _curry2 from './internal/_curry2.js';
 import paths from './paths.js';
 
 /**
- * Retrieve the value at a given path.
+ * Retrieve the value at a given path. String `Idx` that could be parsed as
+ * non-negative integers has the same lookup behavior as their Integer
+ * counterparts. Non-negative integer `Idx` only work on an Array `obj`.
  *
  * @func
  * @memberOf R
@@ -11,7 +13,7 @@ import paths from './paths.js';
  * @typedefn Idx = String | Int | Symbol
  * @sig [Idx] -> {a} -> a | Undefined
  * @param {Array} path The path to use.
- * @param {Object} obj The object to retrieve the nested property from.
+ * @param {Object} obj The object or array to retrieve the nested property from.
  * @return {*} The data at `path`.
  * @see R.prop, R.nth
  * @example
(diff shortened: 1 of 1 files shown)


source/path.js
```diff
@@ -2,7 +2,9 @@ import _curry2 from './internal/_curry2';
 import paths from './paths';
 
 /**
- * Retrieve the value at a given path.
+ * Retrieve the value at a given path. String `Idx` that could be parsed as
+ * non-negative integers has the same lookup behavior as their Integer
+ * counterparts. Non-negative integer `Idx` only work on an Array `obj`.
```
CrossEye, line 7: While I agree that this needs some better explanation, I think this suggestion is too tied to your use-case, which I think is fairly unusual.  I'll try to think about it this weekend.

CrossEye:
Such suggestions are always welcome!

In this case, I disagree on the details.  I'll try to think of some more comprehensive text this weekend.

AndrewSouthpaw:
👍  Let me know if there's more I can do to help.

CrossEye:
Yeah, sorry.  I dropped this entirely.  I will look again.

source/path.js
CrossEye: How about something like this?

> Retrieves the value at a given path.  The nodes of the path can be arbitrary strings or non-negative integers.  For anything else, the value is unspecified.  Integer paths are meant to index arrays, strings are meant for objects.

source/path.js
```diff
@@ -11,7 +13,7 @@ import paths from './paths';
  * @typedefn Idx = String | Int
  * @sig [Idx] -> {a} -> a | Undefined
```
CrossEye, line 14: Perhaps this instead?:

```js
/*
 * @sig [Idx] → {a} → a | Undefined
 * @sig Idx = String | NonNegativeInt
 */
```

source/path.js
```diff
@@ -11,7 +13,7 @@ import paths from './paths';
...
  * @param {Array} path The path to use.
- * @param {Object} obj The object to retrieve the nested property from.
+ * @param {Object} obj The object or array to retrieve the nested property from.
  * @return {*} The data at `path`.
  * @see R.prop, R.nth
```
CrossEye, line 18: We should also mention `assocPath`/`dissocPath` here.

customcommander:
@CrossEye I rebased against the master branch and added your suggestions.

source/path.js
customcommander: Done. Can you check please?

source/path.js
customcommander: Done.

CrossEye (approved):

AndrewSouthpaw:
Oops I totally forgot about this, thanks for carrying it across the finish line, @customcommander!

PR #3013, opened 2020-04-15
Update includes.js

richlloydmiles:


CrossEye:
Yes, thank you.  That is definitely better!

PR #3018, opened 2020-04-22
arguments.length will always be the same as n

adispring:
https://github.com/ramda/ramda/blob/998f798d9ac6297bb9f5b45136e3ff06a56df09a/source/constructN.js#L48-L49

`arguments.length` will always be the same as `n`, so there's no need to dynamically calculate arguments' length. And `switch(n)` may be more clear to source code readers?

### DIFF
--- a/source/constructN.js
+++ b/source/constructN.js
@@ -46,7 +46,7 @@ var constructN = _curry2(function constructN(n, Fn) {
     return function() { return new Fn(); };
   }
   return curry(nAry(n, function($0, $1, $2, $3, $4, $5, $6, $7, $8, $9) {
-    switch (arguments.length) {
+    switch (n) {
       case  1: return new Fn($0);
       case  2: return new Fn($0, $1);
       case  3: return new Fn($0, $1, $2);


CrossEye:
I *think* you're right, because of the currying, but I would like to think it over a bit more to make sure there are no corner cases.

adispring:
`nAry` but not `curry`, guarantees that `n` equals to `arguments.length`. 
`nAry` can limit number of parameters and arguments more strictly than `_arity` and `curryN`.

CrossEye:
I was wondering about the case where someone used this with a consructor that has optional parameters, ones that are baked in by `constructN`.  When called with fewer than `n` arguments, I had been thinking, there might be some difference in behavior because `arguments.length` is less than `n`.  But because of the currying, that would just return a function expecting the remaining ones.

It was because I considered this, even though I quickly saw that it didn't matter, that I want to spend a little time ensuring that there are no additional such cases that actually matter.

CrossEye (approved):

Issue #3024, opened 2020-05-11
Update description/doc for mergeAll


klequis:
Modified description to make clear which values will be in the returned object. This was likely pretty obvious to those already well versed in Ramda but seems a good add for those less so.

**Current description:**

Merges a list of objects together into one object

**New description:**

Creates a new object with the own properties from a list of objects. If a key exists in more than one object, the value from the last object it exists in will be used.

CrossEye:
Thank you!!

PR #3025, opened 2020-05-11
updated mergeAll description per issue #3024

klequis:
Modified description to make clear which values will be in the returned object. This was likely pretty obvious to those already well versed in Ramda but seems a good add for those less so.

see: #3024


CrossEye:
:fireworks: 

Thank you @klequis!

PR #3029, opened 2020-05-19
removing recursion from _clone

webstormrage:
Current _clone implementation is recursive and implements O(nˆ2) solution for reference counting (to deal with objects with circular dependencies). The time complexity for refs counting makes performance on large object graphs poor. In the case of very deep nested objects (new test case for _clone) _clone does not work at all because of maximum call stack size exceeded error. 

This PR fixes these two issues.


CrossEye:
Yes, as soon as we get v1.0 out the door, and are no longer supporting pre-ES6 environments, I think this is a great idea!

CrossEye:
To be revived after 1.0

Issue #3026, opened 2020-05-13
R.traverse doesn't dispatch to `traverse` method, contrary to documentation


joshtch:
The [documentation for `R.traverse`](https://ramdajs.com/docs/#traverse) says:

<pre>
Maps an Applicative-returning function over a Traversable, then uses `sequence` to
transform the resulting Traversable of Applicative into an Applicative of Traversable.

<b>Dispatches to the `traverse` method of the third argument, if present.</b>
</pre>

But [the code](https://github.com/ramda/ramda/blob/ed191e6a476330e37db259a5c2d04bfba0b2d63d/source/traverse.js#L33-L35) does not dispatch as described:

```js
var traverse = _curry3(function traverse(of, f, traversable) {
  return typeof traversable['fantasy-land/traverse'] === 'function' ?
    traversable['fantasy-land/traverse'](f, of) :
    sequence(of, map(f, traversable));
});
```

It instead dispatches to `fantasy-land/traverse` only.

My expectation had been that `R.traverse` would dispatch to the fantasy-land prefixed method, if present, then the non-prefixed method, if present, then fall back to the implementation. This behavior would make it consistent with `map`, `ap`, and a few other functions.

Whatever the case, though, either the code or the documentation needs modification so the two agree.

CrossEye:
Agreed.  Are you interested in creating a PR for this?

joshtch:
Hi Scott -- yes, certainly. 

Just to confirm, are you agreeing/do you agree that the correct functionality should be to dispatch to either of `fantasy-land/traverse` or `traverse` if they exist, preferring the former over the latter?

PR #3031, opened 2020-05-20
Fix dispatching to `traverse` method 

joshtch:
This makes R.traverse dispatch to `traverse` if it's available and `fantasy-land/traverse` is not, as noted in issue #3026.

### DIFF
--- a/source/traverse.js
+++ b/source/traverse.js
@@ -30,8 +30,12 @@ import sequence from './sequence.js';
  *      R.traverse(Maybe.of, safeDiv(10), [2, 0, 5]); //=> Maybe.Nothing
  */
 var traverse = _curry3(function traverse(of, f, traversable) {
-  return typeof traversable['fantasy-land/traverse'] === 'function' ?
-    traversable['fantasy-land/traverse'](f, of) :
-    sequence(of, map(f, traversable));
+  return (
+    typeof traversable['fantasy-land/traverse'] === 'function'
+      ? traversable['fantasy-land/traverse'](f, of)
+      : typeof traversable.traverse === 'function'
+        ? traversable.traverse(f, of)
+        : sequence(of, map(f, traversable))
+  );
 });
 export default traverse;
--- a/test/traverse.js
+++ b/test/traverse.js
@@ -28,9 +28,33 @@ describe('traverse', function() {
     eq(R.traverse(S.Either.of, R.identity, [S.Left('XXX'), S.Left('YYY')]), S.Left('XXX'));
   });
 
-  it('dispatches to `sequence` method', function() {
-    eq(R.traverse(Id, R.map(R.negate), [Id(1), Id(2), Id(3)]), Id([-1, -2, -3]));
-    eq(R.traverse(R.of, R.map(R.negate), Id([1, 2, 3])), [Id(-1), Id(-2), Id(-3)]);
+  it('dispatches to `traverse` method', function() {
+    const mockTraversable = { traverse(_1, _2) { return 'traverse called'; } };
+
+    eq(R.traverse(Id, R.identity, mockTraversable), 'traverse called');
+  });
+
+  it('dispatches to `fantasy-land/traverse` method', function() {
+    const mockTraversable2 = {
+      ['fantasy-land/traverse'](_1, _2) { return 'fantasy-land/traverse called'; }
+    };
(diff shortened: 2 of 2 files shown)


joshtch:
I added some tests to ensure the dispatch methods get used when appropriate.

Bradcomp:
I believe we can update this function to use the _dispatchable decorator:
https://github.com/ramda/ramda/blob/9aa12b4db06855e6bd55a498479d666dd973e31a/source/internal/_dispatchable.js

joshtch:
Sure, that seems reasonable. I'll need a little guidance, though: how am I to handle the transducer case? The docs do not suggest to me that `traverse` can be used as a transducer, so should it fail if it's given a transformer?

joshtch:
@Bradcomp will you please provide the additional information I requested in my previous comment, or unblock this PR?

I grasp that for someone who does understand it, `_dispatchable` is a much cleaner implementation. But to me, it's inscrutable, so I feel deterred from contributing by having to figure out how it works.


But as far as I know, your suggestion does not change the behavior at all. And it introduces the added

joshtch:
@Bradcomp, I've given your suggestion a go, and with some thought, I think I've changed my mind about using `_dispatch` here. I would argue that using `_dispatch` is, in fact, out of the scope of this particular PR.

There aren't any instances I could find of [`_dispatchable` being used](https://github.com/ramda/ramda/search?p=3&q=_dispatchable&unscoped_q=_dispatchable) to implement functions that cannot be used as transducers. So no established convention exists that I can follow. In fact, to use `_dispatchable` for a function that has no documented use as a transducer is contrary to the existing conventions of this library.

Given this, if I were to refactor to using `_dispatchable`, I would have to figure out what the "right" value to use for the transducer argument of `_dispatchable` is. Should it be a function? Or `undefined`? Or `null`? Or something else? If it should be a function, what should the function do? Should it be a runtime error? If so, should the error be thrown, or should it be returned? What should the error message be? If it should not be an explicit error, should it return `undefined`? Or `null`? Or the passed in transformer, unaltered? Something else?

This PR was opened because there was a missing dispatch case in `traverse`. Not because `traverse` needed a refactor. Perhaps it does need a refactor. But this is supposed to be a bug fix, not a refactor, is it not? I'm not trying to be snarky, I mean this sincerely, and I apologize if I'm coming across that way.

Furthermore, I'm not convinced there is a substantive advantage to using `_dispatchable`, anyway. It is added complexity: now a potential contributor has to figure out what the function does, and they have to be careful about the transducer argument even though `traverse` isn't one. Which seems like a potential footgun `_dispatchable` also seems to add to the call stack at runtime, since it [appears to not get optimized away in the build](https://github.com/ramda/ramda/blob/v0.27.0/dist/ramda.js)? The benefit seems to be that it unifies the codebase toward one cleaner, more consistent style, but given there are more functions that *don't* use `_dispatchable` than do, I'm not sure if this PR really achieves that.

It's been a month since this PR was passing all checks, including test cases I provided that addressed the missing functionality that prompted me to open this issue and #3026. As far as I know, @Bradcomp, your suggestion does not change the functionality, isn't that correct? It proposes a refactor, which, on its face, seems perfectly reasonable. But once you dig a little deeper, it turns out to be a meaningfully more involved undertaking. No precedent exists for what you are describing, so a new convention has to be established, albeit a small one.

As such, in my opinion, I think it would be more appropriate for this PR to comply with the existing convention of only using `_dispatchable` with transducers. If the maintainers decide it would be better to do an overhaul and use `_dispatchable` for everything, transducer or not, that's fine. But that's out of scope here. I just wanted to fix a small, niggling issue that was bothering me and maybe others, and has not yet been fixed for reasons unrelated to the extensional functionality of the proposed fix.

So with due respect to @Bradcomp who suggested it, and @CrossEye who :+1: it, I do not think it makes sense to refactor `traverse` to use `_dispatchable` in this PR. I appreciate the suggestion, and would happily comply if it made sense in the context of the existing conventions of the codebase. However, I have not seen evidence to support that. You are welcome to convince me otherwise. But until such time as you have done so, I respectfully decline, and request that this PR be merged in its current form, if there are no other concerns to address.

joshtch:
Oh man, I did not mean to close this. I was rebasing onto the latest changes in Ramda's master branch.

joshtch:
Great! Thanks, guys!

Issue #3030, opened 2020-05-19
Github Actions


hitmands:
Hi, 

there isn't probably a real value in that, 
but we could introduce github actions so that we could help keepings things all in github.

We could create a workflow:

- lint
- test
- build

and set it as CI check in prs

__

this would probably mean decomissioning github

crash7:
I can give a hand here :)

@hitmands what about adding a semantic-release step too?


PS: I think you meant "decomissioning travis"

hitmands:
it'd be interesting to hear @CrossEye here

Bradcomp:
I really like the idea of having everything in one place. I don't know anything about our current Travis setup though.

CrossEye:
The Travis setup was a long time ago, but I have no interest in maintaining it.  If we can do it all in GitHub, I think that would be great.  At work these days, we do everything in GitLab, and I love how little friction there is!

PR #3033, opened 2020-05-31
3030/GitHub actions

hitmands:
Issue: https://github.com/ramda/ramda/issues/3030

This pull request aims to bring the CI inside the github repo via github actions.


.github/workflows/build.yml
```diff
@@ -0,0 +1,26 @@
...
+    strategy:
+      matrix:
+        node: ['14', '12', '10', '8']
+    name: Node ${{ matrix.node }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Install Dependencies
+        run: yarn install
```
hitmands, line 17: - [ ] to use npm

hitmands:
@CrossEye we could use `code-climate` to hoist code coverage, 
I could set this up in this pull request.

**would that be of interest?**

CrossEye:
This looks great to me.  I'm all for moving this into the platform, especially as Travis has become a little flaky lately.

:+1: also on code coverage.  I would want a report from it, though, not a failing build.

.github/workflows/build.yml
```diff
@@ -0,0 +1,26 @@
+name: Build
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: ['14', '12', '10', '8']
```
customcommander, line 10: Minor: is there a specific reason for Node.js 8? According to https://nodejs.org/en/about/releases/ people are encouraged to run production code on LTS/Active releases; Node.js 8 is neither active nor maintained.

hitmands: I totally agree with this point, 
I'd also add that since we _transpile_ ramda, there isn't that much value in making sure you can run a build over multiple node versions...

This was more a porting, so that's why I kept node 8...

@CrossEye what's your view on riding of node 8 here?

Bradcomp: Part of why we test old versions of Node is that we don't have browser tests. At least for now we are committed to supporting ES5 environments, and so we were using older versions of Node as a proxy for that. I don't know how relevant that is at this point in time.

CrossEye: It's probably not all that relevant.  But I think it's worth keeping until we cut away from that support.  And also, as this is meant to be a port, it seems the fewer changes, the better.

But I'm a little worried about code-climate stopping the build.  It might be worth doing, fixing what it notices right now and then keeping it as part of our regular commit process.  I'm much less worried about the errors shown -- we really should fix them -- than about a coverage report stopping our build.  I've been burned by that before, sometimes by removing one or two buggy tests I can no longer build; I'd rather not have that here.

hitmands: for now none of the checks are required to pass,
this is a configuration setting and can be done at another time...

image

____

I am okay to leave node 8 and plan for a removal after 1.0 is out

PR #3037, opened 2020-06-11
Add promap

forrest-akin:
Adds fantasy-land compliant [`promap()`](https://github.com/fantasyland/fantasy-land#fantasy-landpromap-method). `promap()` provides an interface for wrapping a process with a pre- and post-process, which has famously found application in the development of optics. It also could be useful for building and composing SerDes.

I took some inspiration from Tom Harding's [Fantas, Eel, and Specification 18: Bifunctor and Profunctor](http://www.tomharding.me/2017/06/26/fantas-eel-and-specification-18/) for the `profunctor.promap()` test case.

### DIFF
--- a/source/internal/_promap.js
+++ b/source/internal/_promap.js
@@ -1,0 +1,5 @@
+export default function _promap(f, g, profunctor) {
+  return function(x) {
+    return g(profunctor(f(x)));
+  };
+}
--- a/source/promap.js
+++ b/source/promap.js
@@ -1,0 +1,37 @@
+import _curry3 from './internal/_curry3';
+import _dispatchable from './internal/_dispatchable';
+import _promap from './internal/_promap';
+import _xpromap from './internal/_xpromap';
+
+
+/**
+ * Takes two functions as pre- and post- processors respectively for a third function,
+ * i.e. `promap(f, g, h)(x) === g(h(f(x)))`.
+ *
+ * Dispatches to the `promap` method of the third argument, if present,
+ * according to the [FantasyLand Profunctor spec](https://github.com/fantasyland/fantasy-land#profunctor).
+ *
+ * Acts as a transducer if a transformer is given in profunctor position.
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig (a -> b) -> (c -> d) -> (b -> c) -> (a -> d)
+ * @sig Profunctor p => (a -> b) -> (c -> d) -> p b c -> p a d
+ * @param {Function} f The preprocessor function, a -> b
+ * @param {Function} g The postprocessor function, c -> d
+ * @param {Profunctor} profunctor The profunctor instance to be promapped, e.g. b -> c
+ * @return {Profunctor} The new profunctor instance, e.g. a -> d
+ * @see R.transduce
+ * @example
+ *
+ *      const decodeChar = R.promap(s => s.charCodeAt(), String.fromCharCode, R.add(-8))
+ *      const decodeString = R.promap(R.split(''), R.join(''), R.map(decodeChar))
+ *      decodeString("ziuli") //=> "ramda"
+ *
+ * @symb R.promap(f, g, h) = x => g(h(f(x)))
+ * @symb R.promap(f, g, profunctor) = profunctor.promap(f, g)
+ */
+var promap = _curry3(_dispatchable(['fantasy-land/promap', 'promap'], _xpromap, _promap));
+
+export default promap;
--- a/test/promap.js
+++ b/test/promap.js
@@ -1,0 +1,46 @@
+var assert = require('assert');
+
+var R = require('../source');
+var eq = require('./shared/eq');
+var Pair = require('./shared/Pair');
+
+
+describe('promap', function() {
+  var fromCharCode = String.fromCharCode;
+  function charCodeAt(s) { return s.charCodeAt(); }
+  function Costar(run) {
+    return {
+      promap: function promap(f, g) {
+        return Costar(R.pipe(R.map(f), run, g));
+      },
+      run: run
+    };
+  }
+
+  it(`dispatches to pronfuctor["fantasy-land/promap"]() if present`, function() {
+    var setJson = R.promap(JSON.parse, JSON.stringify);
+    var pair = setJson(
+      Pair(R.assoc("left-promapped", true), R.assoc("right-promapped", true))
+    );
+    function mergeWithJson(json) {
+      return function(left, right) { return right(left(json)); };
+    };
+
+    eq(pair.merge(mergeWithJson("{}")), `{"left-promapped":true,"right-promapped":true}`);
+  });
+
+  it(`dispatches to pronfuctor.promap() if present`, function() {
+    var is1337Change = R.promap(R.multiply(100), R.equals(1337), Costar(R.sum));
+    var data = [10, 3, 0.3, 0.07];
+
+    eq(is1337Change.run(data), true);
+  });
+
+  it('composes two functions, f and g, before and after the final function respectively', function() {
+    var decodeChar = R.promap(charCodeAt, fromCharCode, R.add(-8));
+    var decodeString = R.promap(R.split(''), R.join(''), R.map(decodeChar));
+    var code = "ziuli";
+
+    eq(decodeString(code), "ramda");
+  });
+});
--- a/test/shared/Pair.js
+++ b/test/shared/Pair.js
@@ -1,0 +1,45 @@
+var R = require('../../source');
+
+
+function Pair(left, right) {
+  return {
+    '@@type': 'ramda/Pair',
+    'fantasy-land/equals': function equals(pair) {
+      return pair != null
+        && pair['@@type'] === this['@@type']
+        && R.equals(left, pair.left)
+        && R.equals(right, pair.right);
+    },
+    'fantasy-land/concat': function concat(pair) {
+      return Pair(R.concat(left, pair.left), R.concat(right, pair.right));
+    },
+    'fantasy-land/map': function map(f) { return Pair(left, f(right)); },
+    'fantasy-land/ap': function ap(pair) {
+      return Pair(R.concat(left, pair.left), right(pair.right));
+    },
+    'fantasy-land/reduce': function reduce(f, x) { return f(x, right); },
+    'fantasy-land/traverse': function traverse(of, f) {
+      return R.map(
+        function(x) { return Pair(left, x); },
+        of(f(right))
+      );
+    },
+    'fantasy-land/chain': function chain(f) {
+      var pair = f(right);
+      return Pair(R.concat(left, pair.left), pair.right);
+    },
+    'fantasy-land/bimap': function bimap(f, g) { return Pair(f(left), g(right)); },
+    'fantasy-land/promap': function promap(f, g) {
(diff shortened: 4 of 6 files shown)


source/promap.js
```diff
@@ -0,0 +1,13 @@
+import _curry3 from './internal/_curry3';
+import pipe from './pipe';
+
+
+var promap = _curry3(function promap(f, g, profunctor) {
+  return typeof profunctor['fantasy-land/promap'] === 'function'
+    ? profunctor['fantasy-land/promap'](f, g)
+    : typeof profunctor.promap === 'function'
+      ? profunctor.promap(f, g)
```
davidchambers, line 9: I don't think we should dispatch to `promap`.

source/promap.js
```diff
@@ -0,0 +1,13 @@
...
+      : pipe(f, profunctor, g);
```
davidchambers, line 10: I would prefer not to use `pipe` in the implementation.

source/promap.js
forrest-akin: I noticed some other methods dispatch similarly, so included it just in case. I have no opinion in this regard so fine with excluding it

source/promap.js
forrest-akin: In favor of `compose` or a lambda?

source/promap.js
CrossEye: @davidchambers: Why not?  That is still so far how Ramda functions tend to work.  Is there something about `promap` that is different, or is this a more general objection?

However, I would rather use the mechanism we have in place, using `_dispatchable`.  See, for example [`chain`][ch].


  [ch]: https://github.com/ramda/ramda/blob/master/source/chain.js

source/promap.js
CrossEye: A lambda.  Although we have some helper functions that get used internally, we tend to write our internals with as little dependency as we can upon the public API.  The idea is to use as little of the end-user's performance budget as we can in the library code.

test/promap.js
```diff
@@ -0,0 +1,39 @@
...
+
+
+describe.only('promap', function() {
+  var fromCharCode = String.fromCharCode;
+  function charCodeAt(s) { return s.charCodeAt(); }
+  function Costar(run) { return { promap: promap, run: run };
+    function promap(f, g) {
+      return Costar(R.pipe(R.map(f), run, g));
+    }
+  }
```
CrossEye, line 15: Could we reformat this please?  Either something like this:

```js
  function Costar(run) { 
    return { promap: promap, run: run };

    function promap(f, g) {
      return Costar(R.pipe(R.map(f), run, g));
    }
  }
```

or like this:

```js
  function Costar(run) { 
    return { 
      promap: function promap(f, g) {
        return Costar(R.pipe(R.map(f), run, g));
      }, 
      run: run 
    };
  }
```

test/promap.js
```diff
@@ -0,0 +1,39 @@
...
+  });
+
+  it('composes two functions, f and g, before and after the final function respectively', function() {
+    var decodeChar = R.promap(charCodeAt, fromCharCode, R.add(-8));
+    var decodeString = R.promap(R.split(''), R.join(''), R.map(decodeChar));
+    var code = "ziuli";
+
+    eq(decodeString(code), "ramda");
+  });
+});
```
CrossEye, line 39: I love these examples!

test/promap.js
forrest-akin: hehe, fair - I knew I was taking a gamble with that one 😛

forrest-akin:
@CrossEye requested changes have been pushed!

re: functions not using `_dispatchable` - I actually found quite a few! I searched in `/source` for _fantasy-land_ and found the following functions directly dispatching a method on the final argument (excluding internal functions): `ap`, `concat`, `empty`, `sequence`, `traverse`

CrossEye:
@forrest-akin 
> re: functions not using `_dispatchable` - I actually found quite a few! I searched in `/source` for _fantasy-land_ and found the following functions directly dispatching a method on the final argument (excluding internal functions): `ap`, `concat`, `empty`, `sequence`, `traverse`

I think they simply never got converted.  But I may be forgetting something.  I'll try to investigate soon.

source/internal/_promap.js
```diff
@@ -0,0 +1,5 @@
+export default function _promap(f, g, profunctor) {
+  return function(x) {
+    return g(profunctor(f(x)));
+  };
+}
```
davidchambers, line 5: :thumbsup:

test/shared/Maybe.js
```diff
@@ -43,6 +43,11 @@ Maybe.prototype['fantasy-land/ap'] = function(maybe) {
   return this.isJust && maybe.isJust ? Just(maybe.value(this.value)) : Nothing;
 };
 
+//  Maybe#promap :: Maybe (a -> b) ~> (c -> a, b -> d) -> Maybe (c -> d)
```
davidchambers, line 46: `Maybe a` cannot satisfy the requirements of [Profunctor][1] for two reasons, the simpler of which is that `Maybe a` has only one type variable.

Let's see why `Function a b` *can* satisfy the requirements of Profunctor, starting from the generalized `fantasy-land/promap` signature:

```haskell
fantasy-land/promap :: Profunctor p => p b c ~> (a -> b, c -> d) -> p a d
```

Let's replace `Profunctor p => p` with `Function`:

```haskell
fantasy-land/promap :: Function b c ~> (a -> b, c -> d) -> Function a d
```

Finally, let's use the more familiar `(a -> b)` syntax:

```haskell
fantasy-land/promap :: (b -> c) ~> (a -> b, c -> d) -> (a -> d)
```

Given a function from `a` to `b` and a function from `c` to `d`, can we transform `Function b c` into `Function a d`? Yes. We can return a function that receives an `a`, we can get a `b` from the `a`, we can then get a `c` from the `b`, and we can then get a `d` from the `c`.

Let's replace `Profunctor p => p` with `Pair`:

```haskell
fantasy-land/promap :: Pair b c ~> (a -> b, c -> d) -> Pair a d
```

This is impossible, as we do not have an `a` and we have no way to get an `a` from a `b` or a `c`. For this reason `Pair a b` cannot satisfy the requirements of Profunctor.


[1]: https://github.com/fantasyland/fantasy-land#profunctor

test/shared/Id.js
```diff
@@ -12,6 +12,7 @@ module.exports = function Id(value) {
     'fantasy-land/chain': function(f) { return f(value); },
     'fantasy-land/reduce': function(f, x) { return f(x, value); },
     'fantasy-land/traverse': function(f, of) { return R.map(Id, f(value)); },
+    'fantasy-land/promap': function(f, g) { return Id(R.pipe(f, value, g)); },
```
davidchambers, line 15: `Id a` cannot satisfy the requirements of Profunctor for the same reasons that `Maybe a` cannot.

source/promap.js
davidchambers: > That is still so far how Ramda functions tend to work.

I thought this was due to the way Fantasy Land worked four years ago.

test/shared/Maybe.js
forrest-akin: I see, so any Profunctor instance needs to be parameterized as `p a b`; I can't fudge it by providing something else parameterized by `a b` in the instances `a` slot, i.e. `Identity (a -> b)`. I was thinking it might work something like how `Monad a` works with functions by partially applying the input param of `(->)`: `Monad ((->) r)`. I suppose this works because it's still `Monad a`, while `Identity (a -> b)` is not `Profunctor a b`. I'll remove these and use a valid Profunctor instance!

forrest-akin: I suppose `Pair (-> a) (a ->)` could work?

forrest-akin: indeed - `Pair` has a `Profunctor` instance for `(Profunctor p, Profunctor q) => Pair p q`. PR updated 👍

source/promap.js
```diff
@@ -0,0 +1,9 @@
+import _curry3 from './internal/_curry3';
+import _dispatchable from './internal/_dispatchable';
+import _promap from './internal/_promap';
+import _xpromap from './internal/_xpromap';
+
+
+var promap = _curry3(_dispatchable(['fantasy-land/promap', 'promap'], _xpromap, _promap));
```
customcommander, line 7: Perhaps this could include some JSDoc and examples?

test/shared/Pair.js
```diff
@@ -0,0 +1,45 @@
...
+    toString: function toString() {
+      return `Pair(${R.toString(left)}, ${R.toString(right)})`;
+    },
+    toArray: function toArray() { return [left, right]; },
+    left: left,
+    right: right,
+  };
+}
+
+module.exports = Pair;
```
davidchambers, line 45: Another option is to depend on [sanctuary-pair][1].


[1]: https://github.com/sanctuary-js/sanctuary-pair

source/promap.js
CrossEye: Yes, it should.  I looked at the test cases, and didn't even notice that the JSDocs were missing!

test/shared/Pair.js
CrossEye: It's hard to find a point with small changes where it's worth introducing a first dependency into a dependency-free library.  With a large change (say sanctuary-def) it would be a very different matter.

customcommander: For what it's worth (probably not much ;) I do appreciate libraries that make a conscious effort to keep their dependency tree to a minimum. Ramda has none and it's truly _amazing_. I'm sure that must be a contributing factor when it comes to adopting a library. (Admittedly not the most important one nor the only one neither.)

davidchambers: It would only be a development dependency, of which Ramda already has many.

customcommander: Ah sorry @davidchambers I didn't realise this was in a test file!

CrossEye: I wasn't noticing that either.  I don't object to a dependency here.  Nor do I feel any great need.

forrest-akin: I'm not sure that I can use santuary-pair in this case since I wanted to provide the Profunctor instance of Pair for the `fantasy-land/promap` dispatch test case, which it doesn't look like I'd get from sanctuary-pair (unless there's a standard mechanism for providing such implementations to Sanctuary offerings). I chose Pair in this case because it's the simplest non-Function Profunctor instance I could think of 😅

source/promap.js
forrest-akin: Oh, that's my bad; I misread the contributing guidelines saying to exclude `@since` as excluding JSDoc annotations - will get them in today!

forrest-akin: added JSDocs 👍

test/promap.js
Bradcomp: Agreed! They helped me to see some of the usefulness 😄

CrossEye:
Thank you @forrest-akin!  This looks like a great addition.  Any more FantasyLand functions you feel like adding?  :smile:

forrest-akin:
I'll look into knocking out `contramap` and `bimap`, then we'll be map-complete!

forrest-akin:
also, I noticed at the last minute that the `_xpromap` transformer doesn't dispatch to `.promap()` for non-function inputs - lmk if you think that's important and I'll submit a patch

Issue #3028, opened 2020-05-18
Remove ES Module build with a source change


Pyrolistical:
Right now the source is already a well organized set of ES Modules, thank you!

But there is one non-standard right now and its that its omitting the .js extension when importing. This is added back during the build and the files are published under the `es/` folder.

We can see this in action by looking at https://unpkg.com/ramda@0.27.0/es/uniq.js
```js
import identity from "./identity.js";
import uniqBy from "./uniqBy.js";
/**
 * Returns a new list containing only one copy of each element in the original
 * list. [`R.equals`](#equals) is used to determine equality.
 *
 * @func
 * @memberOf R
 * @since v0.1.0
 * @category List
 * @sig [a] -> [a]
 * @param {Array} list The array to consider.
 * @return {Array} The list of unique items.
 * @example
 *
 *      R.uniq([1, 1, 2, 1]); //=> [1, 2]
 *      R.uniq([1, '1']);     //=> [1, '1']
 *      R.uniq([[42], [42]]); //=> [[42]]
 */

var uniq =
/*#__PURE__*/
uniqBy(identity);
export default uniq;
```

Here's where we can do better. If all we did was add .js to all the imports, then we don't even need a build step. The tagged source from github is already usable. We can see https://raw.githubusercontent.com/ramda/ramda/v0.27.0/source/uniq.js is almost there
```js
import identity from './identity';
import uniqBy from './uniqBy';


/**
 * Returns a new list containing only one copy of each element in the original
 * list. [`R.equals`](#equals) is used to determine equality.
 *
 * @func
 * @memberOf R
 * @since v0.1.0
 * @category List
 * @sig [a] -> [a]
 * @param {Array} list The array to consider.
 * @return {Array} The list of unique items.
 * @example
 *
 *      R.uniq([1, 1, 2, 1]); //=> [1, 2]
 *      R.uniq([1, '1']);     //=> [1, '1']
 *      R.uniq([[42], [42]]); //=> [[42]]
 */
var uniq = uniqBy(identity);
export default uniq;
```

Btw, I noticed this when trying to use Ramda in https://github.com/denoland/deno

CrossEye:
Although we would have to run it by people who better understand the various module builds, we would certainly welcome a PR on this.

Andrew-Webb:
It would certainly be great to ensure Ramda works nicely with [Deno](https://deno.land/).  Then it can be added to Deno's [third-party modules list](https://deno.land/x), as `lodash` is.

[...]

PR #3042, opened 2020-06-16
Add `.js` extension to source files

sebastienfilion:
Omitting the file extension when importing a module is a quirk of the NodeJS' require importing method and should be discouraged -- As standard ES module is getting ground (Browser import, Deno import).

Resolve: https://github.com/ramda/ramda/issues/3028


sebastienfilion:
Should I commit the rebuilt distribution file?

Bradcomp:
You don't need to commit the dist file https://github.com/ramda/ramda/blob/master/CONTRIBUTING.md

sebastienfilion:
@Bradcomp I added a plugin to enforce file extension on import, but Codeclimate is complaining although I tested it locally -- Is it possible that they don't support ESLint plugins? If that's the case, what's your preferred way to handle this?

CrossEye:
pinging @hitmands: Do you have any experience with this?  I'm still very new to Code Climate.

hitmands:
Need to take a look into this,
We might need to disable codeclimate.eslint if this plugin isn’t supported

hitmands:
It seems that (unless differently specified) [`eslint-3` is used by default](https://docs.codeclimate.com/docs/eslint), while this repo is using [`esling@5`](https://github.com/ramda/ramda/blob/master/package.json#L100).

I'd suggest to set `channel-5` [here](https://github.com/ramda/ramda/blob/master/.codeclimate.yml#L19), 
image


- https://docs.codeclimate.com/docs/eslint

______

My initial thoughts are that the version of `plugins/import` available for `eslint@3` didn't have any `ignorePackages` option.

image

> Is it possible that they don't support ESLint plugins?

@sebastienfilion it seems that the plugin is supported, 
but the specific config `ignorePackages` isn't.

- https://codeclimate.com/github/ramda/ramda/pull/3042

sebastienfilion:
Thanks for all the help; I've switched the setting from `ignorePackages` to `always` -- it works for my purpose because the source files are self-dependent. Let me know what you think.

sebastienfilion:
If you guys are okay, I'm gonna prepare a PR on the Deno "database" to add Ramda as an official third-party!?

Bradcomp:
@sebastienfilion I have no problem with that, however typescript compatibility isn't a priority for the library, and there is at least one function we have that is a syntax error (`R.type`).

CrossEye:
@sebastienfilion: Absolutely.  Go for it!

sebastienfilion:
@Bradcomp i will mention it on the PR; but I don’t think that FP users care much about typed-languages. We’ll see what Ryan says.

sebastienfilion:
For those that are interested: I submitted Ramda as Deno third-party. https://github.com/denoland/deno_website2/pull/1174

PR #3046, opened 2020-06-25
fix(deps): update lockfile

hitmands:
Fix pipeline


hitmands:
@CrossEye we added eslint plugin import in https://github.com/ramda/ramda/pull/3042, however, we forgot to update lockfiles and caused the build to fail.

Issue #1258, opened 2015-07-09
A curried function for functions with options objects.


JAForbes:
Would you guys be interested in a PR that allows you to curry a function using options objects?

Here is an example: https://gist.github.com/JAForbes/edee9adda777c4ec646e

In the gist, there is a function `request` that takes three arguments.  Two of which are optional (`protocol` and `port`) and 1 is required (`url`).

This request function will continue to accept invocations with options objects, until it receives an options object that contains the required field (`url`).

You can have as many required fields as you would like.  But I kept it to one.

This function side steps the currying argument order issue.  With the added benefit that options objects self document the intention of the arguments.

```js
//example function
function request( protocol, port, url ){
  return protocol + "://" + url + (port == 80 ? "" : ":"+port)
}
 
//creating keyword argument curried function
var kRequest = kwargs(request, {
  port: 80,
  protocol: "http",
  url: undefined
})
 
//tests
 
 
;([
  kRequest
    ({})
    ({})
    ({url: "google.com"}) == "http://google.com"
 
  ,
 
  kRequest
    ({ port: 3000 })
    ({url: "localhost"})  == "http://localhost:3000",
 
  kRequest
    ({ port: 3000 })
    ({ protocol: "https", port: 3001 })
    ({ url: "github.com"}) == "https://github.com:3001"
 
]).every( function(val){ return val} ) && "All tests passed"
```

I don't really like the name `kwargs` by the way, it was just the first thing that came to my head.  I was referencing python kwargs.

davidchambers:
We're better off drawing inspiration from Haskell than from Python in this case, since curried functions are everywhere in Haskell. I suggest something along these lines:

```javascript
const requestWithOpts = R.curry((protocol, hostname, port) =>
  protocol + '://' + hostname + (port === 80 ? '' : ':' + port)
);
const requestHttp  = requestWithOpts('http');
const requestHttps = requestWithOpts('https');
const request      = requestHttp(R.__, 80);
```

JAForbes:
>We're better off drawing inspiration from Haskell than from Python

I found this comment to be quite condescending.  Not sure if that was intended?

asaf-romano:
For functions that take more than 4 or 5 arguments, I tend to prefer named-arguments using ES6 destructing, like this:
```js
function f({ a, b, c, d, e, f })
{
  return a + b * c - d + e / f;
}
f({ a: 1, ... })
```

[...]

PR #3047, opened 2020-06-27
feat: add partialObject

hejhi:
I know it wasn't ever a sure thing this would be considered, but I thought I'd submit a PR to add a new function called namedCurryN, which would be a curryN fn, but with named parameters (as per this thread: https://github.com/ramda/ramda/issues/1258). It supports everything that curryN supports, and parameter order doesn't matter (I remember that being a concern in the thread).

I added a bunch of tests as well, drawing inspiration from the curry and curryN tests.

It was a slightly easier implementation than just a plain named curry where i'd have to infer the arguments in the object, but thought that could come later if this seemed useful. 

Totally understand if it doesn't make sense for Ramda, but I've found it useful with React components and other scenarios. Plus it was fun to make. Thanks for an awesome lib!


source/namedCurryN.js
```diff
@@ -0,0 +1,76 @@
...
+
+function filterPlaceholders(mappedPairs) {
+  return filter(function([, val]) {
+    return !_isPlaceholder(val);
+  }, mappedPairs);
+}
+
+function prepareEntries(acc, mappedPairs) {
+  return compose(curry(concat)(acc), filterPlaceholders)(mappedPairs);
+}
```
CrossEye, line 56: This feels especially strange for a Ramda implementation.  Besides the fact that Ramda uses internal `curry2`, `curry3`, etc. helpers, I can see no reason to write `compose (/* fns */) (/* args */)`.  Here I see no advantage to this over

```js
  return concat(acc, filterPlaceholders (mappedPairs));
```

or something similar with the native `concat`

source/namedCurryN.js
```diff
@@ -0,0 +1,76 @@
+import curryN from './curryN.js';
+import curry from './curry.js';
+import fromPairs from './fromPairs.js';
+import toPairs from './toPairs.js';
+import _isPlaceholder from './internal/_isPlaceholder.js';
+import filter from './filter.js';
+import concat from './concat.js';
+import compose from './compose.js';
+import keys from './keys.js';
+
```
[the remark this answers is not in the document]
hejhi: gotcha, makes sense

source/namedCurryN.js
hejhi: Fair enough, I had native `concat` originally but thought I'd try point-free. Though I think in my refactor I got lost a little in the abstraction, and you're right, I don't think it's necessary. I'll fix.

CrossEye: Before spending time on this particular fix, though, do note the main issue above.  It seems you're matching on the wrong things.

hejhi:
It's interesting. I think you have a good point, and I've definitely thought about this. There's a few approaches I've looked into to match args, which all work with this implementation with small changes, but are less than ideal:
1. Passing argument names as an array to the function (let's call it the angular inject-y road, which is fraught with peril when it comes to minification and feels hacky).
2. Parse a function with a regex to get the params (feels un-ramda to me, though I could be wrong, and also fraught with peril).
3. Divide a function into multiple phases, or have multiple functions that handle accepting of which arguments to take (which in the thread also seemed to not be the ideal option).

So I thought counting could be an interesting approach. I know it could still lead to breaking, but is it the functions responsibility with `namedCurryN` to make sure that the user provides the arguments it needs by name, or is the responsibility of the user? I would say if it was a `curry` then it maybe should, but with `curryN` it seems to be doing what it's expected to do—once you pass it 4 "arguments", it executes the function. If you don't pass the function the arguments it requires, then, the provided function might break and the developer should probably guard for that as there's clearly a chance that within 4 provided arguments, the required arguments for the function may not be provided. I would think that follows a similar path as this example for `curryN` in sentiment:
```
var f = curryN(4, function(a, b, c, d, e, f) {
    return `${JSON.stringify({a, b, c, d, e, f})} => ${(a + b * c) / d + e}`;
}); 
console.log(f(12,3,6,2,4,2)) //=> 19

var g = f(12,3);
console.log(g(6,2,4,2)) //=> 19 // good

console.log(g(6, 2)) //=> {"a":12,"b":3} => NaN
```
Here, `curryN` could _know_ that the function has an arity of 6, but lets the user proceed with just 4 as it's not the responsibility of `curryN`  to know that the function will break when all args aren't provided. I know it's not 1:1 with a named `curryN` approach, but I would say that `namedCurryN` provides benefit, while a (harder to implement!) `namedCurry` would perhaps be better suited to matching?

source/namedCurryN.js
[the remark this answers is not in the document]
hejhi: I notice quite a few relying on the public `curry` and `curryN`, are those alright? I could also use `Object.entries`/`Object.fromEntries` in lieu of `toPairs`/`fromPairs`, but wasn't sure if IE11 was an issue. or i could write my own `reduce` but who wants to do that with the aforementioned 4 ways to get the same done? 😉

CrossEye:
> 1. Passing argument names as an array to the function (let's call it the angular inject-y road, which is fraught with peril when it comes to minification and feels hacky).

I think this would be less fraught, since we're dealing with named object properties, which presumably will not be minified, rather than parameter names, which often will.  (My code samples below choose this route, in two different ways.)

I agree that your other two possibilities are non-starters.

> So I thought counting could be an interesting approach. I know it could still lead to breaking, but is it the functions responsibility with `namedCurryN` to make sure that the user provides the arguments it needs by name, or is the responsibility of the user?

I think there is a huge difference, though.  With Ramda's normal currying, we are depending on a cleanly-ordered set: the parameters to the base function.  With named properties, there is no useful intrinsic ordering.  And counting is really problematic for cases where a derived function wants to supply some new values as well as overrides for previously-chosen values or simple optional values.

But I think, if we're really interested in them, we can do named parameters in a not-too horrible way.  Scott Christopher mentioned one possibility in that earlier thread.  (https://github.com/ramda/ramda/issues/1258#issuecomment-120190987).  That only works for root-level required fields.  I have two different possibilities for deeper values.   The first one is configured with a list of required paths:

```js
const namedCurry = (reqs, fn) =>  {
  const foo = (obj) => reqs .every (
    p => is (Object, path (init (p), obj)) && has (last (p), path (init (p), obj)) 
  ) ? fn (obj)
    : (obj2) => foo (mergeDeepRight (obj, obj2))
  return foo
}

const f = ({foo, bar, baz: {qux: {grault, corge = 42}}}) => 
  [foo, bar, grault, corge]

const g = namedCurry ([
  ['foo'], 
  ['bar'], 
  ['baz', 'qux', 'grault']
], f)

g ({foo: 1, baz: {qux: {corge: 10}}})   // returns function
  ({bar: 2})                            // returns function
  ({baz: {qux: {grault: 99}}})          //=> [1, 2, 99, 10]
```

(You can see this [in the REPL][re1].)

The second one uses a configuration object with a Symbol representing the required values:

```js
function * getPaths(o, p = []) {
  if (Object(o) !== o || Object .keys (o) .length == 0) yield p 
  if (Object(o) === o)
    for (let k of Object .keys (o))
      yield * getPaths (o[k], [...p, Number.isInteger (Number (k)) ? Number (k) : k])
}

const namedCurry = Object .assign ((sample, fn) =>  {
  const [reqs, bases] = partition (p => path (p, sample) == namedCurry .REQUIRED ) ([...getPaths (sample)])
  const base = reduce((o, p) => assocPath(p, path(p, sample), o), {}, bases)
  const foo = (obj) => reqs .every (
    p => is (Object, path (init (p), obj)) && has (last (p), path (init (p), obj)) 
  ) ? fn (obj)
    : (obj2) => foo (mergeDeepRight(mergeDeepRight(base, obj), obj2))
  return foo
}, {REQUIRED: Symbol ()})

const f = ({foo, bar, baz: {qux: {grault, corge = 42}}}) => 
  [foo, bar, grault, corge]

const REQUIRED = namedCurry.REQUIRED

const g = namedCurry ({
  foo: REQUIRED, 
  bar: REQUIRED, 
  baz: {qux: {grault: REQUIRED, corge: 101}}
}, f)


g ({foo: 1})                          // returns function
  ({bar: 2})                          // returns function
  ({baz: {qux: {grault: 99}}})        //=> [1, 2, 99, 10]
 

const h = namedCurry ({
  foo: REQUIRED, 
  bar: 25, 
  baz: {qux: {grault: REQUIRED, corge: 101}}
}, f)

h ({foo: 1})                          // returns function
  ({baz: {qux: {grault: 99}}})        //=> [1, 25, 99, 10]
 
h ({foo: 1})                          // returns function
  ({bar: 2})                          // returns function
  ({baz: {qux: {grault: 99}}})        //=> [1, 2, 99, 10]
```
(This is also available [in the REPL][re2].)

There might well be some cleaner way to pass around the `REQUIRED` value, as well.  I haven't really thought that through.

Please don't take either one as actual example code, only as proofs-of-concept for two different APIs.  Both would likely need some real clean-up.

I still don't have strong feelings either way about whether this is a good fit for Ramda.  But it does seem something we *could* do if we were interested.


  [re1]: https://ramdajs.com/repl/?v=0.27.0#?console%20.clear%20%28%29%0A%0Aconst%20namedCurry%20%3D%20%28reqs%2C%20fn%29%20%3D%3E%20%20%7B%0A%20%20const%20foo%20%3D%20%28obj%29%20%3D%3E%20reqs%20.every%20%28%0A%20%20%20%20p%20%3D%3E%20is%20%28Object%2C%20path%20%28init%20%28p%29%2C%20obj%29%29%20%26%26%20has%20%28last%20%28p%29%2C%20path%20%28init%20%28p%29%2C%20obj%29%29%20%0A%20%20%29%20%3F%20fn%20%28obj%29%0A%20%20%20%20%3A%20%28obj2%29%20%3D%3E%20foo%20%28mergeDeepRight%20%28obj%2C%20obj2%29%29%0A%20%20return%20foo%0A%7D%0A%0Aconst%20f%20%3D%20%28%7Bfoo%2C%20bar%2C%20baz%3A%20%7Bqux%3A%20%7Bgrault%2C%20corge%20%3D%2042%7D%7D%7D%29%20%3D%3E%20%0A%20%20%5Bfoo%2C%20bar%2C%20grault%2C%20corge%5D%0A%0Aconsole.log%28%0A%20%20f%20%28%7Bfoo%3A%201%2C%20bar%3A%202%2C%20baz%3A%20%7Bqux%3A%20%7Bgrault%3A%203%7D%7D%7D%29%0A%29%20%2F%2F~%3E%20%5B1%2C%202%2C%203%2C%204%5D%0A%0Aconst%20g%20%3D%20namedCurry%20%28%5B%0A%20%20%5B%27foo%27%5D%2C%20%0A%20%20%5B%27bar%27%5D%2C%20%0A%20%20%5B%27baz%27%2C%20%27qux%27%2C%20%27grault%27%5D%0A%5D%2C%20f%29%0A%0Aconsole%20.log%20%28%0A%20%20g%20%28%7Bfoo%3A%201%2C%20baz%3A%20%7Bqux%3A%20%7Bcorge%3A%2010%7D%7D%7D%29%20%20%20%2F%2F%20returns%20function%0A%20%20%20%20%28%7Bbar%3A%202%7D%29%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%2F%2F%20returns%20function%0A%20%20%20%20%28%7Bbaz%3A%20%7Bqux%3A%20%7Bgrault%3A%2099%7D%7D%7D%29%20%20%20%20%20%20%20%20%20%20%2F%2F~%3E%20%5B1%2C%202%2C%2099%2C%2010%5D%0A%29
  [re2]: https://ramdajs.com/repl/?v=0.27.0#?console%20.clear%20%28%29%0A%0Afunction%20%2A%20getPaths%28o%2C%20p%20%3D%20%5B%5D%29%20%7B%0A%20%20if%20%28Object%28o%29%20%21%3D%3D%20o%20%7C%7C%20Object%20.keys%20%28o%29%20.length%20%3D%3D%200%29%20yield%20p%20%0A%20%20if%20%28Object%28o%29%20%3D%3D%3D%20o%29%0A%20%20%20%20for%20%28let%20k%20of%20Object%20.keys%20%28o%29%29%0A%20%20%20%20%20%20yield%20%2A%20getPaths%20%28o%5Bk%5D%2C%20%5B...p%2C%20Number.isInteger%20%28Number%20%28k%29%29%20%3F%20Number%20%28k%29%20%3A%20k%5D%29%0A%7D%0A%0Aconst%20namedCurry%20%3D%20Object%20.assign%20%28%28sample%2C%20fn%29%20%3D%3E%20%20%7B%0A%20%20const%20%5Breqs%2C%20bases%5D%20%3D%20partition%20%28p%20%3D%3E%20path%20%28p%2C%20sample%29%20%3D%3D%20namedCurry%20.REQUIRED%20%29%20%28%5B...getPaths%20%28sample%29%5D%29%0A%20%20const%20base%20%3D%20reduce%28%28o%2C%20p%29%20%3D%3E%20assocPath%28p%2C%20path%28p%2C%20sample%29%2C%20o%29%2C%20%7B%7D%2C%20bases%29%0A%20%20const%20foo%20%3D%20%28obj%29%20%3D%3E%20reqs%20.every%20%28%0A%20%20%20%20p%20%3D%3E%20is%20%28Object%2C%20path%20%28init%20%28p%29%2C%20obj%29%29%20%26%26%20has%20%28last%20%28p%29%2C%20path%20%28init%20%28p%29%2C%20obj%29%29%20%0A%20%20%29%20%3F%20fn%20%28obj%29%0A%20%20%20%20%3A%20%28obj2%29%20%3D%3E%20foo%20%28mergeDeepRight%28mergeDeepRight%28base%2C%20obj%29%2C%20obj2%29%29%0A%20%20return%20foo%0A%7D%2C%20%7BREQUIRED%3A%20Symbol%20%28%29%7D%29%0A%0Aconst%20f%20%3D%20%28%7Bfoo%2C%20bar%2C%20baz%3A%20%7Bqux%3A%20%7Bgrault%2C%20corge%20%3D%2042%7D%7D%7D%29%20%3D%3E%20%0A%20%20%5Bfoo%2C%20bar%2C%20grault%2C%20corge%5D%0A%0Aconsole.log%28%0A%20%20f%20%28%7Bfoo%3A%201%2C%20bar%3A%202%2C%20baz%3A%20%7Bqux%3A%20%7Bgrault%3A%203%7D%7D%7D%29%0A%29%20%2F%2F~%3E%20%5B1%2C%202%2C%203%2C%2042%5D%0A%0Aconst%20REQUIRED%20%3D%20namedCurry.REQUIRED%0A%0Aconst%20g%20%3D%20namedCurry%20%28%7B%0A%20%20foo%3A%20REQUIRED%2C%20%0A%20%20bar%3A%20REQUIRED%2C%20%0A%20%20baz%3A%20%7Bqux%3A%20%7Bgrault%3A%20REQUIRED%2C%20corge%3A%20101%7D%7D%0A%7D%2C%20f%29%0A%0Aconsole.log%28%0A%20%20g%20%28%7Bfoo%3A%201%7D%29%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%2F%2F%20returns%20function%0A%20%20%20%20%28%7Bbar%3A%202%7D%29%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%2F%2F%20returns%20function%0A%20%20%20%20%28%7Bbaz%3A%20%7Bqux%3A%20%7Bgrault%3A%2099%7D%7D%7D%29%20%20%20%20%20%20%20%20%2F%2F~%3E%20%5B1%2C%202%2C%2099%2C%2010%5D%0A%29%20%0A%0Aconst%20h%20%3D%20namedCurry%20%28%7B%0A%20%20foo%3A%20REQUIRED%2C%20%0A%20%20bar%3A%2025%2C%20%0A%20%20baz%3A%20%7Bqux%3A%20%7Bgrault%3A%20REQUIRED%2C%20corge%3A%20101%7D%7D%0A%7D%2C%20f%29%0A%0Aconsole.log%28%0A%20%20h%20%28%7Bfoo%3A%201%7D%29%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%2F%2F%20returns%20function%0A%20%20%20%20%28%7Bbaz%3A%20%7Bqux%3A%20%7Bgrault%3A%2099%7D%7D%7D%29%20%20%20%20%20%20%20%20%2F%2F~%3E%20%5B1%2C%2025%2C%2099%2C%2010%5D%0A%29%20%20%0A%0Aconsole.log%28%0A%20%20h%20%28%7Bfoo%3A%201%7D%29%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%2F%2F%20returns%20function%0A%20%20%20%20%28%7Bbar%3A%202%7D%29%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%2F%2F%20returns%20function%0A%20%20%20%20%28%7Bbaz%3A%20%7Bqux%3A%20%7Bgrault%3A%2099%7D%7D%7D%29%20%20%20%20%20%20%20%20%2F%2F~%3E%20%5B1%2C%202%2C%2099%2C%2010%5D%0A%29%20

hejhi:
> With named properties, there is no useful intrinsic ordering. And counting is really problematic for cases where a derived function wants to supply some new values as well as overrides for previously-chosen values or simple optional values.
Great point.

> I think this would be less fraught, since we're dealing with named object properties, which presumably will not be minified, rather than parameter names, which often will.
Of course, didn't think that through 🤦 I also didn't consider deeper values when I was previously passing an array of parameters.

I'll keep playing around with this, those POCs you have there are interesting. I had considered the second approach with Symbols but didn't get around to testing it out. I dig your use of `partition`.

> I still don't have strong feelings either way about whether this is a good fit for Ramda. But it does seem something we could do if we were interested.

well, even if it's not right for Ramda in the end, I'll be happy to have it for my own usage. so no worries.

CrossEye:
One more variant of that second approach will keep `REQUIRED` out higher-level scopes, at the cost of requiring a function for the configuration parameter:

```js
const namedCurry = (makeSample, fn) =>  {
  const REQUIRED = Symbol()
  const sample = makeSample(REQUIRED)
  const [reqs, bases] = partition (p => path (p, sample) == REQUIRED ) ([...getPaths (sample)])
  const base = reduce((o, p) => assocPath(p, path(p, sample), o), {}, bases)
  const foo = (obj) => reqs .every (
    p => is (Object, path (init (p), obj)) && has (last (p), path (init (p), obj)) 
  ) ? fn (obj)
    : (obj2) => foo (mergeDeepRight(mergeDeepRight(base, obj), obj2))
  return foo
}

const f = ({foo, bar, baz: {qux: {grault, corge = 42}}}) => 
  [foo, bar, grault, corge]

const g = namedCurry (
  (REQUIRED) => ({foo: REQUIRED, bar: REQUIRED, baz: {qux: {grault: REQUIRED, corge: 101}}}),
  f
)
```
([REPL][re])

I have no thoughts yet on which variant I prefer.  But I like having such alternatives.

  [re]: https://ramdajs.com/repl/?v=0.27.0#?console%20.clear%20%28%29%0A%0Afunction%20%2A%20getPaths%28o%2C%20p%20%3D%20%5B%5D%29%20%7B%0A%20%20if%20%28Object%28o%29%20%21%3D%3D%20o%20%7C%7C%20Object%20.keys%20%28o%29%20.length%20%3D%3D%200%29%20yield%20p%20%0A%20%20if%20%28Object%28o%29%20%3D%3D%3D%20o%29%0A%20%20%20%20for%20%28let%20k%20of%20Object%20.keys%20%28o%29%29%0A%20%20%20%20%20%20yield%20%2A%20getPaths%20%28o%5Bk%5D%2C%20%5B...p%2C%20Number.isInteger%20%28Number%20%28k%29%29%20%3F%20Number%20%28k%29%20%3A%20k%5D%29%0A%7D%0A%0Aconst%20namedCurry%20%3D%20%28makeSample%2C%20fn%29%20%3D%3E%20%20%7B%0A%20%20const%20REQUIRED%20%3D%20Symbol%28%29%0A%20%20const%20sample%20%3D%20makeSample%28REQUIRED%29%0A%20%20const%20%5Breqs%2C%20bases%5D%20%3D%20partition%20%28p%20%3D%3E%20path%20%28p%2C%20sample%29%20%3D%3D%20REQUIRED%20%29%20%28%5B...getPaths%20%28sample%29%5D%29%0A%20%20const%20base%20%3D%20reduce%28%28o%2C%20p%29%20%3D%3E%20assocPath%28p%2C%20path%28p%2C%20sample%29%2C%20o%29%2C%20%7B%7D%2C%20bases%29%0A%20%20const%20foo%20%3D%20%28obj%29%20%3D%3E%20reqs%20.every%20%28%0A%20%20%20%20p%20%3D%3E%20is%20%28Object%2C%20path%20%28init%20%28p%29%2C%20obj%29%29%20%26%26%20has%20%28last%20%28p%29%2C%20path%20%28init%20%28p%29%2C%20obj%29%29%20%0A%20%20%29%20%3F%20fn%20%28obj%29%0A%20%20%20%20%3A%20%28obj2%29%20%3D%3E%20foo%20%28mergeDeepRight%28mergeDeepRight%28base%2C%20obj%29%2C%20obj2%29%29%0A%20%20return%20foo%0A%7D%0A%0Aconst%20f%20%3D%20%28%7Bfoo%2C%20bar%2C%20baz%3A%20%7Bqux%3A%20%7Bgrault%2C%20corge%20%3D%2042%7D%7D%7D%29%20%3D%3E%20%0A%20%20%5Bfoo%2C%20bar%2C%20grault%2C%20corge%5D%0A%0Aconsole.log%28%0A%20%20f%20%28%7Bfoo%3A%201%2C%20bar%3A%202%2C%20baz%3A%20%7Bqux%3A%20%7Bgrault%3A%203%7D%7D%7D%29%0A%29%20%2F%2F~%3E%20%5B1%2C%202%2C%203%2C%2042%5D%0A%0A%0Aconst%20g%20%3D%20namedCurry%20%28%0A%20%20%28REQUIRED%29%20%3D%3E%20%28%7Bfoo%3A%20REQUIRED%2C%20bar%3A%20REQUIRED%2C%20baz%3A%20%7Bqux%3A%20%7Bgrault%3A%20REQUIRED%2C%20corge%3A%20101%7D%7D%7D%29%2C%0A%20%20f%0A%29%0A%0Aconsole.log%28%0A%20%20g%20%28%7Bfoo%3A%201%7D%29%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%2F%2F%20returns%20function%0A%20%20%20%20%28%7Bbar%3A%202%7D%29%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%2F%2F%20returns%20function%0A%20%20%20%20%28%7Bbaz%3A%20%7Bqux%3A%20%7Bgrault%3A%2099%7D%7D%7D%29%20%20%20%20%20%20%20%20%2F%2F~%3E%20%5B1%2C%202%2C%2099%2C%2010%5D%0A%29%20%0A%0Aconst%20h%20%3D%20namedCurry%20%28%0A%20%20REQUIRED%20%3D%3E%20%28%7Bfoo%3A%20REQUIRED%2C%20bar%3A%2025%2C%20baz%3A%20%7Bqux%3A%20%7Bgrault%3A%20REQUIRED%2C%20corge%3A%20101%7D%7D%7D%29%2C%0A%20%20f%0A%29%0A%0Aconsole.log%28%0A%20%20h%20%28%7Bfoo%3A%201%7D%29%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%2F%2F%20returns%20function%0A%20%20%20%20%28%7Bbaz%3A%20%7Bqux%3A%20%7Bgrault%3A%2099%7D%7D%7D%29%20%20%20%20%20%20%20%20%2F%2F~%3E%20%5B1%2C%2025%2C%2099%2C%2010%5D%0A%29%20%20%0A%0Aconsole.log%28%0A%20%20h%20%28%7Bfoo%3A%201%7D%29%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%2F%2F%20returns%20function%0A%20%20%20%20%28%7Bbar%3A%202%7D%29%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%2F%2F%20returns%20function%0A%20%20%20%20%28%7Bbaz%3A%20%7Bqux%3A%20%7Bgrault%3A%2099%7D%7D%7D%29%20%20%20%20%20%20%20%20%2F%2F~%3E%20%5B1%2C%202%2C%2099%2C%2010%5D%0A%29%20

hejhi:
I didn't get to spend time this week on this due to personal events, but got some time today and just updated with a new approach. There may be technical issues, so I'll fix those (and need to add some more testing scenarios for nesting), but wanted to see what you thought. I liked your suggested signature approach of a straightforward `namedCurry` with the first argument being the paths—it matches by name, and handles deep paths as well. I think this code is actually simpler to reason about than the previous as well. It also works with placeholders and will allow you to curry till all paths have been met. If you add extra params, those will be forwarded as well.

test/namedCurry.js
```diff
@@ -0,0 +1,111 @@
...
+  it('allows further currying of a curried function', function() {
+    var f = R.namedCurry(['a', 'b', 'c', 'd'], function({a, b, c, d}) {return (a + b * c) / d;}); // f(12, 3, 6, 2) == 15
+    var g = f({ a: 12 });
+    eq(g({ b: 3, c: 6, d: 2 }), 15);
+    var h = g({ b: 3 });
+    eq(h({ c: 6, d: 2 }), 15);
+    eq(g({ b: 3, c: 6})({ d: 2}), 15);
+  });
+
+  it('properly reports the length of the curried function', function() {
```
CrossEye, line 28: I don't think this is necessary, given that the length should *always* be one, correct?

test/namedCurry.js
```diff
@@ -0,0 +1,111 @@
...
+    var f = R.namedCurry(['a', 'b', 'c', 'd'], function({a, b, c, d}) {return (a + b * c) / d;});
+    eq(f.length, 1);
+    var g = f({a: 12});
+    eq(g.length, 1);
+    var h = g({b: 3});
+    eq(h.length, 1);
+    eq(g({b: 3, c: 6}).length, 1);
+  });
+
+  it('preserves context', function() {
```
CrossEye, line 38: I think we want to move away from ensuring this.  So I wouldn't add any new tests about preserving context.  If we do away with it, it will of course be a breaking change, but the fewer places it would break, the better.  (Note that I'm more concerned here with *documenting* that this is how it works, even through a unit test, than about making it work this way.  I have no strong feelings on that.)

test/namedCurry.js
```diff
@@ -0,0 +1,111 @@
...
+    var ctx = {x: 10};
+    var f = function({a, b}) { return a + b * this.x; };
+    var g = R.namedCurry(['a', 'b'], f);
+
+    eq(g.call(ctx, {a: 2, b: 4}), 42);
+    eq(g.call(ctx, {a: 2}).call(ctx, {b: 4}), 42);
+  });
+
+  it('supports R.__ placeholder', function() {
```
CrossEye, line 47: I don't understand this at all.  We have named parameters; using the technique to help reorder parameters seems irrelevant; I can see no reason at all to go out of our way to support them.

Instead of 

```js
 g ({a:_, b:2, c:3}) ({a:1});
```

Why wouldn't the user simply do this?:

```js
 g ({b:2, c:3}) ({a:1});
```

test/namedCurry.js
```diff
@@ -0,0 +1,111 @@
...
+    eq(g({a:_, b:2, c:_})({a:1, c:3}), [1, 2, 3]);
+    eq(g({a:_, b:_, c:3})({a:1, b:2}), [1, 2, 3]);
+    eq(g({a:1, b:_, c:_})({b:_, c:3})({b:2}), [1, 2, 3]);
+    eq(g({a:_, b:2, c:_})({a:_, c:3})({a:1}), [1, 2, 3]);
+    eq(g({a:_, b:_, c:3})({a:_, b:2})({a:1}), [1, 2, 3]);
+    eq(g({a:_, b:_, c:_})({a:_, b:_})({c:_})({a:1, b:2, c:3}), [1, 2, 3]);
+    eq(g({a:_, b:_, c:_})({a:1, b:_, c:_})({b:_, c:_})({b:2, c:_})({c:_})({c:3}), [1, 2, 3]);
+  });
+
+  it('supports @@functional/placeholder', function() {
```
CrossEye, line 68: Ditto, of course.  And in any case, we would have no need for both tests that I can see.  The only property of `R.__` that is important is precisely that it has `@@functional/placeholder`.

source/namedCurry.js
```diff
@@ -0,0 +1,60 @@
...
+    const hasPath = path(p, props);
+    return !hasPath || _isPlaceholder(hasPath);
+  })
+}
+
+function parsePaths(paths, fn, acc, props) {
+  const needed = accumulate(paths, props);
+  const nextObj = Object.assign({}, acc, props);
+
+  return needed.length ? curry(parsePaths)(needed, fn, nextObj) : fn.call(this, nextObj);
```
CrossEye, line 53: We almost never use `curry` internally; there's usually a better alternative.

hejhi:
Yeah, I need to add tests for this—currently the tests are copied over from `curry` and `curryN` just to quickly make sure I was covering curry base cases, but I will need to adjust them (to test paths for instance) and will remove unnecessary ones.

> If we go this route, and use properties rather than full paths, how likely do you think it is that users will come back and ask for it to be applied to paths?

Hmm...if I understand you correctly, it should already work with paths? Right now both will work:

```
const g = namedCurry ([
  'foo', 
  'bar', 
  'baz'
], f)
```
```
const g = namedCurry ([
  ['foo'], 
  ['bar'], 
  ['baz', 'qux', 'grault']
], f)
```
It will continue to curry until you match all paths specified. If resolving the path returns a placeholder, it will continue to curry until you match the placeholder. That might not be the correct usage of placeholder, I can remove it. I didn't want to presume people's use cases.

source/namedCurry.js
hejhi: cool I'll take a look at the internal functions

test/namedCurry.js
hejhi: All of these tests were copied from `curry` and `curryN` to make sure I was at least supporting base cases as I worked. I'll adjust/remove them to fit better with this function.

hejhi:
I adjusted the tests, removing the ones you pointed out seemed irrelevant and added a more relevant test for paths. Happy to add more if there's additional scenarios :) I also removed `curry` entirely, as I was just using it for convenience. The only public functions I'm using are `paths` and `hasPaths`, which I know is not ideal and am open to suggestion; I'm sure I could write an internal function if we'd like, I didn't feel like re-inventing the wheel and didn't see any existing internal functions I could leverage instead...perhaps there's an alternative approach to what I'm doing to check paths, but this seemed the most straightforward.

Thanks for the continuous feedback and helpful suggestions, by the way.

CrossEye:
> Hmm...if I understand you correctly, it should already work with paths?

Right.  I did misread a bit.  However...

> Right now both will work:

Did you test this?  Because it shouldn't work.  I would expect this:

```js
const f = ({foo, bar, baz}) => [foo, bar, baz]

const g = namedCurry ([
  'foo', 
  'bar', 
  'baz'
], f)

g ({foo: 'this'}) ({bar: `doesn't`}) ({baz: 'work'}) 
//=> function still looking for correct arguments
```

And this:


```js
const f = ({foo, bar, baz}) => [foo, bar, baz]

const g = namedCurry ([
  ['foo'], 
  ['bar'], 
  ['baz']
], f)

g ({foo: 'but'}) ({bar: `this`}) ({baz: 'does'}) 
//=> ["but", "this", "does"]
```

That's because `hasPath` is not being used properly.  It doesn't fail, just doesn't do what you'd expect when you pass it a string.  That's array-like enough that `hasPath` treats one like an array:

```js
hasPath('foo', {foo: 1, bar: 2})            //=> false
hasPath('foo', {f: {o: {o: 42}, bar: 101}}) //=> true
```

If any of your tests used property names of more than one character, this would have become obvious.

So for this to work, a user would have to supply an array of arrays of Strings.  (Or Numbers or Symbols, I guess.)  That's not a wonderful API, but I can't think of anything better, except possibly the `REQUIRED` examples I gave above.

CrossEye:
> The only public functions I'm using are `paths` and `hasPaths`, which I know is not ideal and am open to suggestion

If we decide we want to do this, then we can create internal versions of each of these and make the public functions things like

```js
export default hasPath = _curry2(_hasPath)
```

or some such.  But let's not bother with that until we make a decision about whether this function should be included in the first place, and if so, what exact API it will have.

hejhi:
> Did you test this? Because it shouldn't work.

Ah yes, my response to your comment was before I pushed up the change to use `hasPath` (instead of using `path` only), so yeah it would need to be an array of array of strings. I also realize I didn't update the old tests to reflect that (though the new one added is written correctly). I personally think an array of arrays of strings as paths is intuitive but open to opinions.

I'll fix the tests and push it up.

hejhi:
Everything is pushed up, if you noticed anything else

test/namedCurry.js
```diff
@@ -0,0 +1,49 @@
+var R = require('../source');
+var eq = require('./shared/eq');
+
+describe('namedCurry', function() {
+  it('curries a single value', function() {
+    var f = R.namedCurry([['a'], ['b'], ['c'], ['d']], function({a, b, c, d}) {return (a + b * c) / d;}); // f(12, 3, 6, 2) == 15
```
CrossEye, line 6: Note that theses comments here are wrong. It's a very minor thing, but that's no longer the correct signature of the functions..

We could change them to be more accurate, but I think it's just as good simply to dump them.

test/namedCurry.js
```diff
@@ -0,0 +1,49 @@
...
+    var g = f({ a: 12 });
+    eq(g({ b: 3, c: 6, d: 2 }), 15);
+  });
+
+  it('curries multiple values', function() {
+    var f = R.namedCurry([['a'], ['b'], ['c'], ['d']], function({a, b, c, d}) {return (a + b * c) / d;}); // f(12, 3, 6, 2) == 15
```
CrossEye, line 12: ditto.  and several more below.

Bradcomp:
> So the question -- and I want to address this to all of @ramda/core and others who are interested -- is whether this API feels like a good match for Ramda. It clearly has its utility, and it's clearly something that is complex enough to make a library implementation useful. But does it match Ramda.

I have some comments I want to make on the PR itself, but I wanted to first answer this question. 

TLDR: I don't see any issue with adding this - it's close enough to our philosophy, but would really only be helpful for codebases that like currying to be able to interface with libraries of differing philosophies.

I probably won't use this function. I can see its usefulness though. I typically think Ramda style currying (and thus the code I write when using Ramda) works best when functions have a static arity with well ordered parameters. This guides a lot of the work I do when I am designing function APIs - what order makes sense for the parameters? What parameters do I actually need? How is this going to be called? Etc.

Using Objects to simulate named parameters is a good idea, but I typically don't use them that way. When I have Objects that represent more than a single parameter  (as opposed to passing in a record or a map) it is generally because either the data comes to me that way or it's an optional config variable where none of the properties are required (if they are required I will put them in a positional parameter).

There are functions in the world and in other libraries that use named parameters. In those cases I think this would be useful to convert those into something that would be more in line with how the rest of a codebase does things. Even then, would it be useful to do it this way, or to take a function with named parameters and convert it into a curried one that takes positional parameters?

All that being said, I can see how this could work and how it could be used to make people's lives easier, and I don't think it really goes against anything in our philosophy so I think it's good. 

I have no insight on a name, but I will let it kick around in my head for a bit.

source/namedCurry.js
```diff
@@ -0,0 +1,53 @@
...
+ *      const addFourNumbers = ({ a, b, c, e }) => a + b + c.d + e;
+ *
+ *      const curriedAddFourNumbers = R.namedCurry([['a'], ['b'], ['c', 'd'], ['e']], addFourNumbers);
+ *      const f = curriedAddFourNumbers({ b: 1, c: { d: 2 } });
+ *      const g = f({ a: 3 });
+ *      g({ e: 4 }); //=> 10
+ */
+
+function accumulate(paths, props) {
+  return paths.filter((p) => {
```
Bradcomp, line 35: We should use our internal `_filter` here: https://github.com/ramda/ramda/blob/master/source/internal/_filter.js

source/namedCurry.js
```diff
@@ -0,0 +1,53 @@
...
+    return !hasPath(p, props) || _isPlaceholder(path(p, props));
```
Bradcomp, line 36: I went through the current discussion, and I still don't see a need for this function to work with placeholders. Placeholders are specifically designed to help with positional parameters. Since positions is irrelevant, we should remove placeholder support unless I am missing something.

source/namedCurry.js
```diff
@@ -0,0 +1,53 @@
...
+function curryPaths(paths, fn, acc) {
+  return (props) => {
+    const needed = accumulate(paths, props);
+    const nextObj = Object.assign({}, acc, props);
+
+    return needed.length ? curryPaths(needed, fn, nextObj) : fn.call(this, nextObj);
+  };
+}
+
+function namedCurry(paths = [], fn) {
```
Bradcomp, line 49: Ramda doesn't use default parameters. Even after the codebase is ES6 we still won't be using them. This function must take an Array of Path as the first parameter.

source/namedCurry.js
```diff
@@ -0,0 +1,53 @@
...
+  return curryPaths(paths, fn, {});
+}
+
+export default namedCurry;
```
Bradcomp, line 53: We should wrap the public function in `_curry2`: https://github.com/ramda/ramda/blob/master/source/internal/_curry2.js

test/namedCurry.js
```diff
@@ -0,0 +1,49 @@
...
+  });
+
+  it('supports nested paths', function() {
+    var f = R.namedCurry([['a'], ['b', 'c', 'd'], ['e'], ['f', 'g']], function({a, b, e, f}) {return (a + b.c.d * e) / f.g;}); // f(12, 3, 6, 2) == 15
+    var g = f({ a: 12, b: { c: { d: 3 } } });
+    eq(g({ e: 6, f: { g: 2 } }), 15);
+    var h = f({ a: 12, b: { c: { d: 3 } }, e: 6 });
+    eq(h({ f: { g: 2 } }), 15);
+  });
+});
```
Bradcomp, line 49: Assuming we move forward, I would like to add some property tests for this. There are examples for `curry` here: https://github.com/ramda/ramda/blob/master/test/curry.js#L125

The property we want to test is that the fully applied curried function always returns the same thing as the uncurried version of the function for the same set of named arguments.

Another property is that the argument order to the curried function is irrelevant:
`forall a, b, c: f({a})({b})({c}) == f({b})({c})({a}) == ...`

source/namedCurry.js
hejhi: I think you're probably right. I figured it made sense to include and remove rather than not consider and have to refactor later, but I don't have any strong opinions either way. If people want it, it wouldn't be hard to add back.

hejhi:
Alright, I _think_ I got everything mentioned above fixed in the latest push—thanks for feedback. I also added two property tests, similar to the curry examples, that test argument order irrelevance and multiple values (though perhaps the one for multiple values is redundant as it's very similar to the order test).

I can see the points around how this could use a different name, so open to suggestions. And if you spot anything else, just let me know.

source/namedCurry.js
```diff
@@ -0,0 +1,53 @@
+import hasPath from './hasPath.js';
+import _filter from './internal/_filter';
+import _curry2 from './internal/_curry2';
+
+/**
+ * Returns a curried equivalent of the provided function with named parameters.
+ * Has all the capabilities of curry, but with named parameters. Supports nested paths as well.
+ * All paths entered as the first argument must be fulfilled before the function is executed.
+ * The following are equivalent, though returned order is not guaranteed:
```
Bradcomp, line 9: I would take out the part here that says "though returned order is not guaranteed". That is really dependent on the behavior of `g` rather than the behavior of this function.

source/namedCurry.js
```diff
@@ -0,0 +1,53 @@
...
+ *   - `g({ b: 2, c: 3 })({ a: 1 })`
+ *   - `g({ c: 3, b: 2, a: 1 })`
+ *
+ * @func
+ * @memberOf R
+ * @category Function
+ * @sig (* -> a) -> (* -> a)
+ * @param {Array} paths Array of required paths in object argument.
+ * @param {Function} fn The function to curry.
+ * @return {Function} A new, curried function.
```
Bradcomp, line 22: When we decide on a name that should drive what these two lines say.

source/namedCurry.js
```diff
@@ -0,0 +1,53 @@
...
+function accumulate(paths, props) {
+  return paths.filter((p) => {
+    return !hasPath(p, props) || _isPlaceholder(path(p, props));
+  });
+}
+
+function curryPaths(paths, fn, acc) {
+  return (props) => {
+    const needed = accumulate(paths, props);
+    const nextObj = Object.assign({}, acc, props);
```
CrossEye, line 43: Let's use the internal [_objectAssign][oa] shim here for consistency.


  [oa]: https://github.com/ramda/ramda/blob/master/source/internal/_objectAssign.js

source/namedCurry.js
CrossEye: I agree.  Let's simply drop them.

source/namedCurry.js
```diff
@@ -0,0 +1,53 @@
...
+
+function accumulate(paths, props) {
+  return paths.filter((p) => {
+    return !hasPath(p, props) || _isPlaceholder(path(p, props));
+  });
+}
+
+function curryPaths(paths, fn, acc) {
+  return (props) => {
+    const needed = accumulate(paths, props);
```
CrossEye, line 42: Can we simply inline the `filter` call here and drop the `accumulate` helper?

test/namedCurry.js
CrossEye: Let's remove them.  They are no longer accurate, and they never were very helpful.

CrossEye:
I'm still stuck for a good name.  While not absolutely opposed to using "curry" in the name, my preference, as discussed above, would be to do something different.

CrossEye:
I created [a Stack Overflow question][so] regarding the function name.


  [so]: https://stackoverflow.com/q/62882449

CrossEye:
One alternative brought up on StackOverflow is not to treat this like `curry`, but more like `partial`, something akin to

```js
const partialObject = (fn, obj) => (o) =>
  fn ({...obj, ...o}) // really a deep merge, but this should explain the idea.

const f = ({a, b, c, d}) => (a + b * c) / d

partialObject 
  (f, {a: 12, b: 3}) //=> new function
  ({c: 6, d: 2})     //=> 15
```

This of course doesn't let us specify the required structure of the output object, which seemed one of the main reasons for this PR.  But that is also its problem.  Having a function that is curried generally makes it clear what the signatures of the subsequent partially applied functions are.  Here it seems quite fuzzy.  But with partial application, the library is no longer trying to pretend to offer you that, only to make it easier to create a version of the function which bakes in an incomplete object so you no longer have to specify those properties..

hejhi:
Yeah, you know what, that makes a lot of sense to me–I'll give it a tinker tonight, it's a solid suggestion.

hejhi:
This is interesting. Do you think it would be worthwhile to provide an optional merge function as a third argument for added control with a default of `mergeDeepRight`? Or does that feel outside the scope of the function?

CrossEye:
> Do you think it would be worthwhile to provide an optional merge function as a third argument for added control with a default of `mergeDeepRight`? 

Well, even if Ramda didn't have a hard stance against optional arguments, this would not feel right to me.  It loses too much of its simplicity.  If there is a strong argument for this, we could consider offering multiple functions, one that accepts such an initial merge argument, one that curries in `mergeDeepRight` for that and possibly one that curries in `mergeDeepLeft`.

But I wouldn't do that yet; we can always refactor to that.  I don't see much call for `mergeDeepLeft` or other merge capabilities.  Yes, it would offer an API that lets you progressively fix parameters that can no longer be overridden.  But since the initial function is still presumably available, that control seems superfluous, simply an unnecessary complication.

hejhi:
I see—makes sense. I just pushed up a `partialObject` implementation, curious about what you all think of it. It does seem a lot cleaner than specifying a path, and (at least for the use cases I have in mind) it would be very useful.

I carried over a test from the `partial` function, and created two more (one to test order, one to test deep merge). Finally, I know it's probably not great, but I imported the public `mergeDeepRight` as I didn't see an internal implementation.

Also, I know it's called `partialObject` now, but do we think it should be `partialObjectDeep`? I wonder if there would be a use case for `partialObject` on its own where it doesn't merge deep.

hejhi:
Sounds great! Yes I agree, the simplicity of it ended up being quite nice I think :) I cleaned up and labelled the test better, so hopefully it's in a good spot.

I also realized in my PR that there were a bunch of changes to the package-lock which was odd...so those are gone from the PR now too.

source/partialObject.js
```diff
@@ -0,0 +1,32 @@
...
+ *      const greet = ({ salutation, title, firstName, lastName }) =>
+ *        salutation + ', ' + title + ' ' + firstName + ' ' + lastName + '!';
+ *
+ *      const sayHello = R.partialObject(greet, { salutation: 'Hello' });
+ *      const sayHelloToMs = R.partialObject(sayHello, { title: 'Ms.' });
+ *      sayHelloToMs({ firstName: 'Jane', lastName: 'Jones' }); //=> 'Hello, Ms. Jane Jones!'
+ * @symb R.partial(f, { a, b })({ c, d }) = f({ a, b, c, d })
+ */
+
+export default _curry2((f, o) => (props) => f.call(this, mergeDeepRight(o, props)));
```
aaditmshah, line 32: Do you really need this function? Note that `partialObject(f)` is equivalent to `curry(compose(f, mergeDeepRight))`.

Here's a [demo](https://ramdajs.com/repl/?v=0.27.0#?const%20discriminant%20%3D%20%28%7B%20a%2C%20b%2C%20c%20%7D%29%20%3D%3E%20b%20%2A%20b%20-%204%20%2A%20a%20%2A%20c%3B%0Aconst%20f%20%3D%20curry%28compose%28discriminant%2C%20mergeDeepRight%29%29%3B%0A%0A%2F%2F%20it%20caches%20the%20initially%20supplied%20arguments%0A%0Aconsole.log%28f%28%7B%20a%3A%203%20%7D%29%28%7B%20b%3A%207%2C%20c%3A%204%20%7D%29%29%3B%20%2F%2F%201%0Aconsole.log%28f%28%7B%20a%3A%203%2C%20b%3A%207%20%7D%29%28%7B%20c%3A%204%20%7D%29%29%3B%20%2F%2F%201%0A%0A%2F%2F%20it%20works%20regardless%20of%20property%20order%0A%0Aconsole.log%28f%28%7B%20b%3A%207%20%7D%29%28%7B%20c%3A%204%2C%20a%3A%203%20%7D%29%29%3B%20%2F%2F%201%0Aconst%20g%20%3D%20curry%28compose%28f%28%7B%20a%3A%203%20%7D%29%2C%20mergeDeepRight%28%7B%20c%3A%204%20%7D%29%29%29%3B%0Aconsole.log%28g%28%7B%20b%3A%207%20%7D%29%29%3B%20%2F%2F%201%0A%0A%2F%2F%20it%20merges%20deeply%20right%0A%0Aconst%20nestedDiscriminant%20%3D%20%28%7B%20a%2C%20c%2C%20d%20%7D%29%20%3D%3E%20d.b%20%2A%20d.b%20-%204%20%2A%20a%20%2A%20c%3B%0Aconst%20h%20%3D%20curry%28compose%28nestedDiscriminant%2C%20mergeDeepRight%29%29%3B%0Aconst%20j%20%3D%20curry%28compose%28h%2C%20mergeDeepRight%28%7B%20a%3A%200%2C%20d%3A%20%7B%20b%3A%200%20%7D%20%7D%29%29%29%3B%0Aconsole.log%28j%28%7B%20a%3A%203%2C%20d%3A%20%7B%20b%3A%207%20%7D%20%7D%29%28%7B%20c%3A%204%20%7D%29%29%3B%20%2F%2F%201%0A). I think `curry(compose(...))` offers more flexibility because you can choose which merging function to use.

```javascript
const discriminant = ({ a, b, c }) => b * b - 4 * a * c;
const f = curry(compose(discriminant, mergeDeepRight));

// it caches the initially supplied arguments

console.log(f({ a: 3 })({ b: 7, c: 4 })); // 1
console.log(f({ a: 3, b: 7 })({ c: 4 })); // 1

// it works regardless of property order

console.log(f({ b: 7 })({ c: 4, a: 3 })); // 1
const g = curry(compose(f({ a: 3 }), mergeDeepRight({ c: 4 })));
console.log(g({ b: 7 })); // 1

// it merges deeply right

const nestedDiscriminant = ({ a, c, d }) => d.b * d.b - 4 * a * c;
const h = curry(compose(nestedDiscriminant, mergeDeepRight));
const j = curry(compose(h, mergeDeepRight({ a: 0, d: { b: 0 } })));
console.log(j({ a: 3, d: { b: 7 } })({ c: 4 })); // 1
```

The point is that you really don't need a `partialObject` or even a `namedCurry` function.

hejhi: That's a valid point, thanks for the examples.

I suppose the question, in the words of @CrossEye , is whether the utility of this is enough to make a library implementation useful?

I personally think the utility of having this function is worthwhile, and doing it manually with `curry`, `compose` and a merge function is a great option when I need something more specific. But of course I'm curious to hear other thoughts as well.

CrossEye: @aaditmshah:

Yes, it's clear that this function is not necessary.  But that's the nature of a library: anything it does can be done directly in your own code.

Ramda strives to strike a balance between several different factors, chiefly how widely useful a function is and how difficult or easy it would be for users if we kept it in userland.

One of my favorite examples is that we have both `pipe` and `compose`.  We could obviously keep only one of them as either is trivial to write in terms of the other.  But they are both so useful and they are often used by different audiences, so we supply both of them.  We can contrast that with [diffobjs][do], a much less trivial-to-write function but which seems to have a much more restricted audience.  In that case, we don't supply it, although we keep a link in the [Cookbook][cb].

I was mixed about whether we should include this functionality in the earlier guises: one which specified what properties/paths needed to be included before applying the underlying function.  But with this formulation, I see it immediately as something I've written many times in one way or another.  It strikes me as one likely to get a fair bit of use.



  [do]: https://github.com/ramda/ramda/wiki/Cookbook#diffobjs---diffing-objects-similar-to-guavas-mapsdifference

hejhi:
Renamed the PR to be more accurate—if there's anything else anyone notices, please let me know!

source/partialObject.js
```diff
@@ -0,0 +1,32 @@
...
+ *      const double = R.partialObject(multiply2, { a: 2 });
+ *      double({ b: 2 }); //=> 4
+ *
+ *      const greet = ({ salutation, title, firstName, lastName }) =>
+ *        salutation + ', ' + title + ' ' + firstName + ' ' + lastName + '!';
+ *
+ *      const sayHello = R.partialObject(greet, { salutation: 'Hello' });
+ *      const sayHelloToMs = R.partialObject(sayHello, { title: 'Ms.' });
+ *      sayHelloToMs({ firstName: 'Jane', lastName: 'Jones' }); //=> 'Hello, Ms. Jane Jones!'
+ * @symb R.partial(f, { a, b })({ c, d }) = f({ a, b, c, d })
```
Bradcomp, line 29: This should be `R.partialObject` I think.

hejhi:
@Bradcomp typo fixed! Any chance of getting that thumbs up one more time?

CrossEye:
@hejhi: Thank you for persevering with this.  I know it was a slow process.  But I think it turned out a function both useful and well-written.  A very nice addition!

:herb:

hejhi:
My pleasure—happy to contribute to an excellent lib. Thanks for all the feedback!

PR #3050, opened 2020-06-29
Update rollup, plugins, and config

benbryant0:
Brings ~1.5 years of updates, lower build times (12.5s -> 10s on my machine), and cuts out 2 transitive dependencies.

The non-minified build output is only subtly different, with some internal variable names having characters appended to them. The minified output does come out differently (I'm not diving into that 😅) but is the same number of bytes.


rollup.config.js
```diff
@@ -1,27 +1,26 @@
...
   output: {
     format: 'umd',
     name: 'R',
     exports: 'named',
     banner: banner
   },
   plugins: [
-    babel({ presets: [['@babel/preset-env', { targets: { ie: '11' } }]]})
+    babel({
+      babelHelpers: 'bundled',
```
benbryant0, line 21: This is recommended when using the newer version of this plugin: https://www.npmjs.com/package/@rollup/plugin-babel#babelhelpers

As no babel helpers are used by ramda, this has no effect.

CrossEye:
This looks fine to my untutored eyes.

Is there someone in @ramda/core with some more Babel experience who can look this over?

rollup.config.js
```diff
@@ -1,27 +1,26 @@
-'use strict';
-
-var babel = require('rollup-plugin-babel');
+import babel from '@rollup/plugin-babel';
 // uglify handles only es5 code, so this also acts as smoke test against shipping es2015+ syntax
-var uglify = require('rollup-plugin-uglify').uglify;
-var pkg = require('./package.json');
+import { uglify } from "rollup-plugin-uglify";
+import pkg from './package.json';
```
hitmands, line 4: currently we produce builds on node `8+`,
using import in plain js files would break that.

I'd suggest to keep using require for the time being.

rollup.config.js
```diff
@@ -38,4 +37,4 @@ if (process.env.NODE_ENV === 'production') {
   );
 }
 
-module.exports = config;
+export default config;
```
hitmands, line 40: same as above, 
I'd suggest to stick to commonjs for the time being

rollup.config.js
benbryant0: Ah, I didn't realise. It might be worth indicating that in `package.json`: https://stackoverflow.com/a/29349773/12895058
This means the latest rollup >=v2 will not work. I've tried 1.32.1 and that works fine, so I'll step it down to that.
What version of 8 is sufficient? Is support for 8.0.0 required, or is something like 8.10.0 acceptable?

Also, `import` statements work in at least node 7 as far as I can tell from testing it locally.

hitmands: esmodules are still marked as experimental in node 14:
https://nodejs.org/api/esm.html#esm_ecmascript_modules

___

I think changing that is slightly out of scope for this pr, 
we'll have to discuss whether to keep node 8 support or not moving forward, 
but I guess this could be a topic for a different story...

benbryant0: The main reason for the change was to conform with the recommended rollup config format, but if we're dropping it to 1.X then that's no longer the case anyway. I'll revert it.

I still need to know what the absolute minimum engine version is. If 8.3.0 is acceptable then only rollup has to be lowered. If it's 8.0.0, then a rollup plugin will also have to be lowered.

rollup.config.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-var babel = require('rollup-plugin-babel');
+var babel = require('@rollup/plugin-babel').babel;
```
hitmands, line 3: what about `var { babel } = require('...')`?

benbryant0:
Would it be possible to get this wrapped up? I have no problem with closing it if there's not much interest or it's not felt that it's worth it.

Bradcomp:
This must have just fallen by the wayside. I think it's fine to merge...

benbryant0:
:tada: Thanks!

PR #3054, opened 2020-07-01
Add support for Nest.land - a package manager for Deno

sebastienfilion:


egg.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "ramda",
+  "description": "A practical functional library for JavaScript programmers.",
+  "stable": true,
+  "entry": "./source/index.js",
+  "files": [
+    "./source/**/*.js",
+    "./README.md"
+  ],
+  "version": "0.27.0",
```
Bradcomp, line 10: Will this need to be manually updated when we do releases?

sebastienfilion: That would be ideal; but my understanding is that new versions needs to be published any. I can take care of that.

CrossEye: I think we can include that in our build script.

tbaumer22: @CrossEye It would be very easy for you all to connect this to our `publish` workflow [here](https://github.com/nestdotland/eggs-ci) for new releases!

CrossEye: @tbaumer22: Thanks.  We'll take a look.

sebastienfilion:
Hey guys, first I would like to know what is the next step here? This thread has been dead for a while.
Also, I'm hijacking the PR to ask if it would be possible to add a webhook to deno.land? https://deno.land/x#add

t8:
@sebastienfilion I'm here to help too. Let me know what steps I can take to help ramda with nest.land integration.

PR #3060, opened 2020-07-17
Unexpected infinite recursion

Mr-LJQ:
Calling the method of the object does not deal with "this", which is very dangerous, in fact, it does cause infinite recursion.
`R.transduce(R.dropLastWhile(R.gt(3)),R.add,0,[1,2,3,4,5])`

### DIFF
--- a/source/internal/_xdropLastWhile.js
+++ b/source/internal/_xdropLastWhile.js
@@ -1,7 +1,7 @@
 import _curry2 from './_curry2.js';
 import _reduce from './_reduce.js';
 import _xfBase from './_xfBase.js';
-import bind from '../bind.js'
+import bind from '../bind.js';
 
 
 function XDropLastWhile(fn, xf) {
--- a/source/internal/_xdropLastWhile.js
+++ b/source/internal/_xdropLastWhile.js
@@ -1,7 +1,7 @@
 import _curry2 from './_curry2.js';
 import _reduce from './_reduce.js';
 import _xfBase from './_xfBase.js';
-import bind '../bind.js'
+import bind from '../bind.js'
 
 
 function XDropLastWhile(fn, xf) {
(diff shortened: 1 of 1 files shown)


CrossEye:
Somehow this got skipped and I never saw it.  Transducers have always confused me.  Can you explain what's going on and how your fix changes it?

Mr-LJQ:
The problem is this, when I saw this code at the time, I felt that there might be a problem with it, because it references a method of an object, but does not bind this (the most important thing is that this method uses to this). There is bound to be a bug caused by this pointing. I tried it, and sure enough, infinite recursion appeared. Take this example as an illustration `R.transduce(R.dropLastWhile(R.gt(3)), R.add,0,[1,2,3,4,5])`.

First, when `R.transduce(...)` is executed, `R.add` is wrapped by `XWrap`.

```javascript
var transduce = curryN(4, function transduce(xf, fn, acc, list) {
  return _reduce(xf(typeof fn === 'function' ? _xwrap(fn) : fn)//first wrapped
  , acc, list);
});
```

And when `R.dropLastWhile(...)` is executed, `XWrap(R.add)` is wrapped by `XWrap` again in the `_reduce()` logic.

```javascript
XDropLastWhile.prototype.flush = function(result, input) {
  result = _reduce(//Here _reduce is wrapped twice
    this.xf['@@transducer/step'],
    result,
    this.retained
  );
  this.retained = [];
  return this.xf['@@transducer/step'](result, input);
};
function _reduce(fn, acc, list) {
  if (typeof fn === 'function') {
    fn = _xwrap(fn);//where secondary wrapping happens
  }
  //....
```

The situation can be simplified to：

```javascript
function Wrap(fn) {
  this.f = fn;
}
Wrap.prototype["step"] = function (acc, val) {
  return this.f(acc, val);
};
let fn = new Wrap(new Wrap(R.add).step);
fn.f(1, 2);
```
Essentially, the above code ends up forming a structure like this：
```javascript
this.f = function (acc,val) {
  return this.f(acc.val)
}
```
Obviously, the error mainly occurs. Two this, which should point to different objects, point to the same object, which is the cause of infinite recursion.

Issue #3066, opened 2020-08-01
R.clone Cloning the prototype property to own property


andykao1213:
As the title says, the function `clone` will copy the properties in the prototype to its own property. Like the example below:
```javascript
const PersonPrototype = {
  name: "",
  test: function() {
    console.log(this.name);
  }
};

const person = Object.create(PersonPrototype);
console.log(person.hasOwnProperty("test"));  // false
const clonePerson = R.clone(person);
console.log(clonePerson)
console.log(clonePerson.hasOwnProperty("test"));  // true
```
Here's the [Ramda REPL](https://ramdajs.com/repl/?v=0.27.0#?const%20PersonPrototype%20%3D%20%7B%0A%20%20name%3A%20%22%22%0A%7D%3B%0A%0Aconst%20person%20%3D%20Object.create%28PersonPrototype%29%3B%0Aconsole.log%28person.hasOwnProperty%28%22name%22%29%29%3B%0Aconst%20clonePerson%20%3D%20R.clone%28person%29%3B%0Aconsole.log%28clonePerson.hasOwnProperty%28%22name%22%29%29%3B).
I found out that the problem is in [_clone.js](https://github.com/ramda/ramda/blob/master/source/internal/_clone.js):
```javascript
for (var key in value) {
    copiedValue[key] = deep ?
    _clone(value[key], refFrom, refTo, true) : value[key];
}
```
The for loop here should check `hasOwnProperty` to fix this issue.
If this is a real issue, I would like to open a pull request to fix this.
Thank you!

andykao1213:
I've opened a pull request in #3067 , please check

PR #3067, opened 2020-08-01
Fix Clone Function Copy Prototype Issue

andykao1213:
To fix the issue mention in #3066 .

1. Check the property key with `hasOwnProperty` when copy to avoid copying the prototype from the original value.
2. Make the new value's prototype point to the origin value's prototype.

### DIFF
--- a/source/internal/_clone.js
+++ b/source/internal/_clone.js
@@ -25,9 +25,12 @@ export default function _clone(value, refFrom, refTo, deep) {
     refFrom[idx] = value;
     refTo[idx] = copiedValue;
     for (var key in value) {
-      copiedValue[key] = deep ?
-        _clone(value[key], refFrom, refTo, true) : value[key];
+      if (value.hasOwnProperty(key)) {
+        copiedValue[key] = deep ?
+          _clone(value[key], refFrom, refTo, true) : value[key];
+      }
     }
+    copiedValue.__proto__ = value.__proto__;
     return copiedValue;
   };
   switch (type(value)) {
--- a/source/internal/_clone.js
+++ b/source/internal/_clone.js
@@ -30,11 +30,10 @@ export default function _clone(value, refFrom, refTo, deep) {
           _clone(value[key], refFrom, refTo, true) : value[key];
       }
     }
-    copiedValue.__proto__ = value.__proto__;
     return copiedValue;
   };
   switch (type(value)) {
-    case 'Object':  return copy({});
+    case 'Object':  return copy(Object.create(Object.getPrototypeOf(value)));
     case 'Array':   return copy([]);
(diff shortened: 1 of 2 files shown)


Cameri:
@andykao1213 could you please add unit tests for this issue?

source/internal/_clone.js
```diff
@@ -25,9 +25,12 @@ export default function _clone(value, refFrom, refTo, deep) {
...
     refTo[idx] = copiedValue;
     for (var key in value) {
-      copiedValue[key] = deep ?
-        _clone(value[key], refFrom, refTo, true) : value[key];
+      if (value.hasOwnProperty(key)) {
+        copiedValue[key] = deep ?
+          _clone(value[key], refFrom, refTo, true) : value[key];
+      }
     }
+    copiedValue.__proto__ = value.__proto__;
```
Bradcomp, line 33: Took a look at MDN and it looks like using `__proto__` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/proto). Is there an alternative way to accomplish this?

andykao1213: @Bradcomp 
Thanks for the advice!
I try to use the build-in function `Object.prototype.create` and `Object.prototype.getPrototypeOf` to replace this.
I fixed it in eaa0ac9, please check!

andykao1213:
@Cameri @Bradcomp 
Thank you for your reply!
I've added unit tests in 7e02905
And I also fix the issue using deprecate `__proto__` in eaa0ac9c327db0c002dc0fd7703c3bd222f56b91
Please check if theese make sense to you.
Thank you!

Cameri:
@andykao1213 Looking good, just one linter issue to fix!

andykao1213:
@Cameri Linter issue fixed!

andykao1213:
@Bradcomp Thank you for the approval!

Here's my thought about the concern you have mentioned.
If you mean the case that altering the prototype property through the object, it will be fine.
Because javascript will copy the property on the prototype to the original object, and the prototype will keep the same:
```javascript
const prototype = {
  prop: 'prototype prop'
}

const obj = Object.create(prototype)
obj.prop = 'altered prop'
console.log(obj)          // {"prop":"altered prop"}
console.log(prototype)    // {"prop":"prototype prop"}
```
Here's a [Ramda REPL](https://ramdajs.com/repl/#?const%20prototype%20%3D%20%7B%0A%20%20prop%3A%20%27prototype%20prop%27%0A%7D%0A%0Aconst%20obj%20%3D%20Object.create%28prototype%29%0Aobj.prop%20%3D%20%27altered%20prop%27%0Aconsole.log%28obj%29%0Aconsole.log%28prototype%29%0A)

PR #3069, opened 2020-08-05
update CONTRIBUTING and package.json to prevent linting errors

Bradcomp:
I noticed that running `npm test` doesn't do linting, and that our CONTRIBUTING.md file doesn't specify linting as a separate step. As a result a contributor following our process could introduce linting errors. This has happened recently.

I'd like to avoid issues like this in the future.

I updated our CONTRIBUTING.md to be more true to life, and I updated our package.json to run the linting on `npm test`. 

Is there a preferred alternative to these changes? 
Should we update CI so it doesn't run linting twice?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,10 +18,11 @@
     commit message, wrapped at 72 characters. Do not commit changes to
     `dist/ramda.js`.
 
-1.  Run `npm test` (or `make test lint`) and address any errors. It will install
-    needed dependencies locally.  Preferably, fix commits in place using `git
-    rebase` or `git commit --amend` to make the changes easier to review and to
-    keep the history tidy.
+1.  Run `npm install` to install needed local dependencies.
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -18,10 +18,11 @@
     commit message, wrapped at 72 characters. Do not commit changes to
     `dist/ramda.js`.
 
-1.  Run `npm test` (or `make test lint`) and address any errors. It will install
```
Bradcomp, line 21: Side note: I looked at the makefile and it doesn't look like it is linting either. I am not super experienced with `make` though so I don't really want to update it.

CrossEye: I think we need to spend some real time working through our build.  What are the goals?  What do we use to reach them?  How can we do so consistently?

Right now this part has grown very haphazardly.  I think it's time to it under control.

Issue #3070, opened 2020-08-06
clone non enumerable properties


felipe-gustavo:
`clone` doen't works with non enumerable properties, such below e.g.:

```js
class e {
  id;
  data;
  constructor(data) {
    this.id = Math.floor(Math.random() * 25);
    this.data = data;
    Object.defineProperty(this, "gTr", {
      enumerable: false,
      value: function () {
        return "it's me " + this.id + "!";
      }
    });
  }
}

const i = new e({ a: 1, b: 2 });
const u = Object.assign(
  Object.create(i),
  Object.getPrototypeOf(Object.create(i))
);
const o = clone(i);

u.id = 28;
u.data.a = 3;
o.data.a = 4;

console.log(
  "i:",
  "   " + (i.gTr && i.gTr()),
  "   " + i.data.a,
  "o:",
  "   " + (o.gTr && o.gTr()),
  "   " + o.data.a,
  "u:",
  "   " + (u.gTr && u.gTr()),
  "   " + u.data.a
);
```

output:
```
i:
   it's me! 23 
   3
o: 
   undefined
   4
u:
   it's me! 28 
   3

```

Bradcomp:
Hi @felipe-gustavo! Thanks for the report!

It looks like we may be able to use [`getOwnPropertyNames`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyNames) to solve this. Would you be interested in submitting a PR?

felipe-gustavo:
@Bradcomp, thank for the quicky answer.

Are you talking about [this row](https://github.com/ramda/ramda/blob/1cd34a0dc028c1eaf44f22b9900c4bde8e1b7848/source/internal/_clone.js#L27)?
![image](https://user-images.githubusercontent.com/55498940/89595162-335a3f80-d82a-11ea-8e84-a07f48096ce2.png)

felipe-gustavo:
> Hi @felipe-gustavo! Thanks for the report!
> 
> It looks like we may be able to use [`getOwnPropertyNames`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyNames) to solve this. Would you be interested in submitting a PR?

[...]

PR #3072, opened 2020-08-10
Change clone to assign non-enumerables properties

felipe-gustavo:
This will solve this issue #3070, I also add if the user added non-enumerable properties in array, [this line](https://github.com/ramda/ramda/commit/2f8e135902115f9110b12027c20360ce436b2d83#diff-3e22b8fb0b7b9c9e80982d8493e966efR35)


Bradcomp:
Thanks for the PR @felipe-gustavo!

Can you add in a test for this?

felipe-gustavo:
done, man!

source/internal/_clone.js
```diff
@@ -24,15 +24,15 @@ export default function _clone(value, refFrom, refTo, deep) {
     }
     refFrom[idx] = value;
     refTo[idx] = copiedValue;
-    for (var key in value) {
+    for (var key of Object.getOwnPropertyNames(value)) {
```
CrossEye, line 27: Most of Ramda will work -- believe it or not -- with ES3.  But I'm certain some ES5 has slipped in.

After 1.0, we will skip that and move to more modern JS.

While I like this change, It's not clear whether we should do it now or after 1.0.

test/clone.js
```diff
@@ -212,3 +212,61 @@ describe('Let `R.clone` use an arbitrary user defined `clone` method', function(
   });
 
 });
+
+describe('Clone object with non-enumerable properties in Object and Array, preserve constructor', function() {
+
+  it('clone non-enumerable in object', function() {
+
+    function EfYP7(data) {
```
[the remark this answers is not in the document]
felipe-gustavo: Doesn't invoke anything, I just work with random function name to tests, I'm sorry for paradigm break

source/internal/_clone.js
felipe-gustavo: Thanks for advice, I didn't read the entire documentation(in my language is the last note [pt-br](https://developer.mozilla.org/pt-BR/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyNames)), but just [`Object.create`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create) resolve this issue, I also noticed that this is already in production (#3067), so, can you close this PR and issue(#3070)?

felipe-gustavo: Thanks for advice, I didn't read the entire documentation(in my language is the last note [pt-br](https://developer.mozilla.org/pt-BR/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyNames)), but just [`Object.create`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create) resolve this issue, I also noticed that this is already in production (#3067), so, I'm closing this PR

PR #3081, opened 2020-09-02
evolve: ignores transform if the input value is not Array and Object

adispring:
If the input is a primitive value, `evolve` will always return an empty object -- `{}`. 

So if the input is a primitive value, it's better to return itself than return an empty object?


CrossEye:
Agreed.  This makes sense. :herb:

PR #3082, opened 2020-09-12
fix: update readme.md ramda version before publish

adispring:


CrossEye:
This looks like a very good idea.  I'd actually like to entirely remove `make` from our build dependencies, and just used node/npm scripts.  It certainly makes it easier to build cross-environments.

But I don't have time to test this today.  I'll try to get to it in the next few days.

CrossEye (approved):

Issue #3084, opened 2020-09-15
Documentation For `propOr` still describes the old behaviour


orthanc:
The documentation for [propOr](https://ramdajs.com/docs/#propOr) says

> If the given, non-null object **has an own property with the specified name**, returns the value of that property. Otherwise returns the provided default value.

Emphasis mine. However #2394 changed this behavior to align with [pathOr](https://ramdajs.com/docs/#pathOr) and return the default if the value is falsey. 

The documentation is now inconsistent with this behavior because there are now cases where the object has an own property but the default is returned. E.g.

```
R.propOr('Ramda', 'favoriteLibrary')({favoriteLibrary: null})
```

The new behavior is more likely what people are expecting, but contradicts the documentation.

orthanc:
#3085 is my suggestion for fixed documentation

PR #3085, opened 2020-09-15
Update documentation of propOr to describe defaulting beahviour.

orthanc:
fixes: #3084 

The documentation for [propOr](https://ramdajs.com/docs/#propOr) says

> If the given, non-null object **has an own property with the specified name**, returns the value of that property. Otherwise returns the provided default value.

Emphasis mine. However #2394 changed this behavior to align with [pathOr](https://ramdajs.com/docs/#pathOr) and return the default if the value is falsey. 

The documentation is now inconsistent with this behavior because there are now cases where the object has an own property but the default is returned. E.g.

```
R.propOr('Ramda', 'favoriteLibrary')({favoriteLibrary: null})
```

The new behavior is more likely what people are expecting, but contradicts the documentation.

This updates the documentation to align with the documentation of `prop` and
`defaultTo` which is what is used in the implementation so should be an
accurate description of the behavior.

### DIFF
--- a/source/propOr.js
+++ b/source/propOr.js
@@ -4,9 +4,10 @@ import prop from './prop.js';
 
 
 /**
- * If the given, non-null object has an own property with the specified name,
- * returns the value of that property. Otherwise returns the provided default
- * value.
+ * Return the specified property of the given non-null object if the property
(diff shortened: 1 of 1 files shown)


Bradcomp (approved):
Thanks for doing this! :bowtie:

Issue #2509, opened 2018-04-18
How can I clear memoizeWith items?


kudorori:
How can I clear memoizeWith items?

CrossEye:
I'm afraid you can't.  I doubt Ramda will ever be interested in supporting that.  Functional programming is about pure functions.  If you supply the same arguments, you get the same value, and there are no side effects.  The only reason for `memoizeWith` is to allow you to avoid repeating expensive calculations.  The values stored are not the sort of things that should be invalidated.

Of course it's quite easy to write a simple `memoize` decorator of your own, especially if you know the types of parameters involved.

Izhaki:
> There are only two hard things in Computer Science: cache invalidation and naming things. -- Phil Karlton

@kudorori I agree with @CrossEye - cache invalidation is a complex business. You have popular solutions like [memoize-one](https://www.npmjs.com/package/memoize-one), but even these can be are not suitable sometimes.

I've been working on this project that had a function called `getAbsoluteBoundingRect` which was *really* expensive to work out; `memoizeWith` would do the trick for static data (say with mouse-detection), but soon as the data change `memoizeWith` was out of the question. I ended up with bespoke cache invalidation.

There are just too many specific scenarios to support - I'd also encourage you to write one that best suits your needs.

CrossEye:
... or my favorite variant:

> There are only two hard things in Computer Science: cache invalidation, naming things, and off-by-one errors.  -- *Some Wag*

meszaros-lajos-gyorgy:
My quick and dirty solution is to create a factory for the memoized function and call it every time you need a fresh, clean instance of memoization:

```javascript
const fnFactory = () => {
  return memoizeWith(toString, (arg1, arg2) => {
    // ... some expensive thing to be called/calculated
  })
}

// since the function reference will change over time
// it's needed to be referenced via on object or some other wrapper
const state = {
  fn: fnFactory()
}

// something happened, need to clear cache
document.getElementById('clear-cache').addEventListener('click', () => {
  state.fn = fnFactory()
})

document.body.addEventListener('click', () => {
  console.log(state.fn('arg1', 'arg2'))
})
```

Qquanwei:
Can we add a new LRU Cache memorize function for ramda ? it will be helpful for high-frequence calc situation.

CrossEye:
[...]

PR #3087, opened 2020-09-22
feat: add memoizeLRUWith

Qquanwei:
Resolve memoizeWith in memory sensetive situation.

## Before
will be bring memory leak.

```javascript
function aBigObject() {
  return {}; 
}

var fn = memoizeWith(R.identity, aBigObject);

fn(1);
fn(2);
...
```

## After

always hold 100 items in memory.
```javascript
function aBigObject() {
  return {}; 
}

var fn = memoizeLRUWith(100, R.identity, aBigObject);

fn(1);
fn(2);
...
```

### DIFF
--- a/source/index.js
+++ b/source/index.js
@@ -255,3 +255,4 @@ export { default as zip } from './zip.js';
 export { default as zipObj } from './zipObj.js';
 export { default as zipWith } from './zipWith.js';
 export { default as thunkify } from './thunkify.js';
+export { default as memoizeLRUWith } from './memoizeLRUWith.js';
--- a/source/internal/_lru.js
+++ b/source/internal/_lru.js
@@ -1,0 +1,85 @@
+function createNode(key, value) {
+  return {
+    newer: null,
+    older: null,
+    key: key,
+    value: value
+  };
+}
+
+function LRU(size) {
+  this.size = size;
+
+  this.length = 0;
+  this.head = null;
+  this.last = this.head;
+
+  this.map = new Map();
+}
+
+LRU.prototype.has = function(key) {
+  return this.map.has(key);
+};
+
+LRU.prototype._hoistNode = function(node) {
+  if (this.head === node) {
+    return;
+  }
+
+  if (this.last === node) {
+    this.last = node.newer || node;
+    this.last.older = null;
+  }
+
+  if (node.older) {
+    node.older.newer = node.newer;
+  }
+  if (node.newer) {
+    node.newer.older = node.older;
+  }
+
+  node.older = this.head;
+  node.newer = null;
+
+  if (this.head) {
+    this.head.newer = node;
+  }
+
+  if (!this.last) {
+    this.last = node;
+  }
+  this.head = node;
+};
+
+LRU.prototype.get = function(key) {
+  if (!this.has(key)) {
+    return undefined;
+  }
+
+  var node = this.map.get(key);
+
+  this._hoistNode(node);
+
+  return node.value;
+};
+
+LRU.prototype.set = function(key, value) {
+  var node;
+  if (this.has(key)) {
+    node = this.map.get(key);
+    node.value = value;
+  } else if (this.length < this.size) {
+    node = createNode(key, value);
+    this.map.set(key, node);
+    this.length += 1;
+  } else {
+    node = this.last;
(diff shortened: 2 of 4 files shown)


Qquanwei:
Or using inject cache strategy `memoizeBy(new LRU(100), identity, fn)` maybe better ?

CrossEye:
At a first read-through, this looks like very well-written, easily maintainable code.  Kudos.  If we decide to go this way, we can investigate further the decision about using an injectible caching strategy.

But I'm not particularly convinced that this is a good fit for Ramda.  I teeter even on `memoizeWith`.  Ramda generally avoids maintaining any internal state; memoization inherently involves state.  Granted the state is not observable, except perhaps by timing calls to the cache.  But it's still state, and that feels off to me.  If we didn't have it from our early days, we probably wouldn't add it now.  This PR adds even more state management.  That bothers me.

On the other hand, this is an undeniably useful function and it's non-trivial to write oneself -- two factors that help us decide whether to include something in Ramda.

I'd love to hear more from you, @Qquanwei about why you think this is a good idea.

And I'd love to hear thoughts from @kudorori, @Izhaki, @meszaros-lajos-gyorgy, who participated in #2509 and of course from others interested, especially @ramda/core.  Is this something we want to add to Ramda.  Also @char0n: if this isn't something Ramda wants to take on, do you think ramda-adjunct might be interested?

Qquanwei:
In practice, pure functions tend to work well with memory functions, so a memory function with state is essential for Ramda.I often use the memoizeWith function in some optimization scenarios, but it can cause problems due to memory leaks. #2509 is a good solution but it lacks sufficient intelligence, so I would like to provide a convenient caching strategy so we have more options for performance optimization.This PR was not a deliberate choice, I would prefer a lodash like Memoize function that allows for cache policy injection.

char0n:
> Also @char0n: if this isn't something Ramda wants to take on, do you think ramda-adjunct might be interested?

@CrossEye, @Qquanwei although in [ramda-adjunct](https://github.com/char0n/ramda-adjunct) we adhere to same general principles as ramda, our bar is lower and function like `memoize` would be acceptable. But it would need to be as flexible as the one provided by lodash with resolver, full cache controll and policy injection.

CrossEye:
So I had never looked at the lodash `memoize` function.  This is the current description.  

> `_.memoize(func, [resolver])`
> 
> Creates a function that memoizes the result of `func`. If `resolver` is provided, it determines the cache key for storing the result based on the arguments provided to the memoized function. By default, the first argument provided to the memoized function is used as the map cache key. The `func` is invoked with the `this` binding of the memoized function.
> 
> **Note**: The `cache` is exposed as the cache property on the memoized function. Its creation may be customized by replacing the `_.memoize.Cache` constructor with one whose instances implement the [Map][ma] method interface of `clear`, `delete`, `get`, `has`, and `set`.

There is definitely something in there to like.  If Ramda was to take this on, I think hanging the `cache` property off the generated function would make sense for us, and probably reusing the `Map` interface would also make sense.  (I wouldn't use a constructor but a plain function; but that's minor.)   I would keep with our explicit equivalent of `resolver`, and I would make the cache strategy a required parameter; certainly I wouldn't add the global state of `_memoize.Cache` used in lodash.  We could retain backward compatibility in `memoizeWith` as the partial application of the current strategy to a new `memoize` function.

This is still not to say that I'm sure we want to do so.

But if we do, I can see writing it this way, keeping our current `memoizeWith` for backward compatibility, but simply implementing it as `const memoizeWith = memoize(_currentNaiveCacheStrategy)`, so that now `memoizeWith.cache` has the `Map` instance interface.  (Not that `clear` or `delete` would do anything in the current strategy.)

I'm still slightly queasy about the notion, though.  Perhaps it's no big deal, as it's not Ramda offering a stateful tool, only allowing users to bring their own along.  But it's still an odd direction for Ramda.

I'd still really like to hear from others.  @buzzdecafe?  @Bradcomp? @davidchambers? 

  [ma]: http://ecma-international.org/ecma-262/7.0/#sec-properties-of-the-map-prototype-object

davidchambers:
One option is to suggest that people use [lodash.memoize][1].


[1]: https://www.npmjs.com/package/lodash.memoize

PR #3088, opened 2020-09-22
chore: rename xf to transducerCreator

adispring:
When I looked into some source code of ramda related to transducer, and some other articles about transducer, I found out that `xf` in `_dispatchable` is not a transducer, but a transducer constructor.

So in order to make the description more clear, I rename `xf` to `transducerCreator`.

### DIFF
--- a/source/internal/_dispatchable.js
+++ b/source/internal/_dispatchable.js
@@ -7,16 +7,17 @@ import _isTransformer from './_isTransformer.js';
  * object in list position (last argument). If it is an array, executes [fn].
  * Otherwise, if it has a function with one of the given method names, it will
  * execute that function (functor case). Otherwise, if it is a transformer,
- * uses transducer [xf] to return a new transformer (transducer case).
+ * uses transducer created by [transducerCreator] to return a new transformer
+ * (transducer case).
  * Otherwise, it will default to executing [fn].
  *
  * @private
  * @param {Array} methodNames properties to check for a custom implementation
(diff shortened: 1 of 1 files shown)


source/internal/_dispatchable.js
```diff
@@ -7,16 +7,17 @@ import _isTransformer from './_isTransformer.js';
...
  * execute that function (functor case). Otherwise, if it is a transformer,
- * uses transducer [xf] to return a new transformer (transducer case).
+ * uses transducer created by [transducerCreator] to return a new transformer
+ * (transducer case).
  * Otherwise, it will default to executing [fn].
  *
  * @private
  * @param {Array} methodNames properties to check for a custom implementation
- * @param {Function} xf transducer to initialize if object is transformer
+ * @param {Function} transducerCreator transducer constructor if object is transformer
```
CrossEye, line 16: Very minor nit.  I would use `factory` here instead of `constructor`.  This is not a constructor function (called with `new`), so I would avoid the word to avoid confusion.

CrossEye (changes requested):
Except for a minor nit in the comment wording, this makes sense to me.

Any objectsions, @ramda/core?

source/internal/_dispatchable.js
adispring: Ok, I have changed `transducer constructor` to `transducer factory`

Bradcomp (approved):
:bowtie:

CrossEye (approved):

PR #3095, opened 2020-10-09
Move existing PBT tests to fast-check and add some more

dubzzz:
As seen in issue #1540, I am opening this Pull Request in order to:
- migrate ramda to fast-check instead of jsverify - _indeed a year ago jsverify owner announced that he stopped the maintenance of the library and the most recent commit was done the 6th of September 2019_
- add some others for `symmetricDifference`, `identical` and `reverse`


test/identical.js
```diff
@@ -26,4 +27,10 @@ describe('identical', function() {
     eq(R.identical(new Number(0), new Number(0)), false);
   });
 
+  it('perfect clones should be considered identical', function() {
+    fc.assert(fc.property(fc.dedup(fc.anything({ maxDepth: 0 }), 2), function(values) {
+      eq(R.identical(values[0], values[1]), true);
+    }));
+  });
```
Bradcomp, line 34: Can you explain this test? Identical shouldn't work for clones, as we have `equal` for that. What does `fc.dedup` do?

test/symmetricDifference.js
```diff
@@ -44,4 +45,78 @@ describe('symmetricDifference', function() {
     eq(R.symmetricDifference(M, M2), []);
   });
 
+  // Remove duplicated values (if any) with respect to R.equals
```
Bradcomp, line 48: We have `uniq` which removes duplicate items from an Array. We should just use that if it works rather than creating a new function.

Bradcomp (commented):
I really appreciate you doing this!

test/symmetricDifference.js
dubzzz: Will do!

test/identical.js
dubzzz: `dedup` (badly named, I need to issue a version to rename it), is supposed to produce clones.

Indeed, it seems that I should move this test to equal instead 🤔 (and remove the `maxDepth` if equality is deep)

CrossEye:
I'll have to look through this in more detail soon, but I've very happy to see these changes.  Thanks, @dubzzz !

test/symmetricDifference.js
dubzzz: Done

test/identical.js
dubzzz: I moved this test to equal and add another one to ensure that equal is symmetric. Regarding this newly added test, in the past I used it to find bugs on Jest (but it needed several runs as inputs causing bugs where pretty rare given the scope of inputs we authorize for this property).

test/curry.js
```diff
@@ -123,28 +122,38 @@ describe('curry', function() {
...
-      g(_, b, c)(a),
-      g(a, _, _)(_, c)(b),
-      g(a, b, _)(c)
-    ]);
+  it('curries with placeholder', function() {
+    fc.assert(fc.property(fc.func(fc.anything()), fc.anything(), fc.anything(), fc.anything(), function(f, a, b, c) {
+      var _ = {'@@functional/placeholder': true, x: Math.random()};
+      var f3 = function(a, b, c) {
+        return f(a, b, c);
+      };
```
Bradcomp, line 147: Why do we need to generate a new function here instead of currying `f` directly?

dubzzz: I initially tried to do that but actually ramda seems to use the `.length` of `f` during currying. The function `f` generated by fast-check looks something like `(..args) => {...}` so the `f.length === 0`. When you try to curry such function it does not work well.

```js
const {curry} = require('ramda');

const f = (...args) => args.reduce((a, b) => a + b, 0);
const g = (a, b, c) => a + b + c;

console.log(curry(f)(1, 2)) // =3
//console.log(curry(f)(1, 2)(3)) // does not work, as curry(f)(1, 2) evaluates to 3 and not to a function

console.log(curry(g)(1, 2)) // =function
console.log(curry(g)(1, 2)(3)) // =6
```
See https://runkit.com/dubzzz/5f8dd0a5e0a9d20019e56c1c

Bradcomp (dismissed):
:bowtie: I really appreciate this, seems like a gain all around. I may even try writing some tests once this gets merged in.

dubzzz:
> :bowtie: I really appreciate this, seems like a gain all around. I may even try writing some tests once this gets merged in.

@Bradcomp Thanks a lot for the approval, I'll also try to add some others too

Bradcomp:
> I'll also try to add some others too

This is fine as it is, but if you're interested here is the one I was thinking about:

https://github.com/ramda/ramda/pull/2962#issuecomment-579000410

dubzzz:
@Bradcomp I am going to have a look to #2962. Do you want me to do it in this PR or to open another one?

dubzzz:
@Bradcomp I finally took some time to implements some PBT tests for https://github.com/ramda/ramda/pull/2962#issuecomment-579000410. I also refreshed the PR to includes the latest version of fast-check

CrossEye (approved):
This looks good to me.

Any thoughts, @ramda/core ?

Bradcomp (approved):
I apologize. I thought I had already approved this. Thanks for doing this!

CrossEye:
Thank you @dubzzz!

dubzzz:
Thanks for the reviews 👍

PR #3096, opened 2020-10-09
fix: partial build

cviejo:
It seems like partial builds broke with the recent rollup update,  output is now formatted as chunks.
```
events.js:187
      throw er; // Unhandled 'error' event
      ^

TypeError [ERR_INVALID_ARG_TYPE]: The "chunk" argument must be one of type string or Buffer. Received type undefined
    at validChunk (_stream_writable.js:268:10)
    at SyncWriteStream.Writable.write (_stream_writable.js:303:21)
    at ramda/scripts/partialBuild.js:15:18
Emitted 'error' event on SyncWriteStream instance at:
    at emitErrorNT (internal/streams/destroy.js:92:8)
    at emitErrorAndCloseNT (internal/streams/destroy.js:60:3)
    at processTicksAndRejections (internal/process/task_queues.js:80:21) {
  code: 'ERR_INVALID_ARG_TYPE'
}
```

### DIFF
--- a/scripts/partialBuild.js
+++ b/scripts/partialBuild.js
@@ -12,7 +12,8 @@ rollupConfig.plugins.push(partialBuildPlugin);
 rollup.rollup(rollupConfig).then(function(bundle) {
   return bundle.generate(rollupConfig.output);
 }).then(function(result) {
-  process.stdout.write(result.code);
+  var chunk = result.output[0];
+  process.stdout.write(chunk.code);
 });
 
 function partialBuild(options) {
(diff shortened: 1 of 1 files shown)


scripts/partialBuild.js
```diff
@@ -12,7 +12,8 @@ rollupConfig.plugins.push(partialBuildPlugin);
 rollup.rollup(rollupConfig).then(function(bundle) {
   return bundle.generate(rollupConfig.output);
 }).then(function(result) {
-  process.stdout.write(result.code);
+  var chunk = result.output[0];
+  process.stdout.write(chunk.code);
```
Bradcomp, line 16: Will there always be exactly one chunk? Would it be better to iterate through the chunks calling `write` for each one?

cviejo: I suspect in this case it will always output one chunk, but iterating is definitely better. Changed

Bradcomp (approved):
:bowtie: Thanks for doing this!

CrossEye:
Thank you, @cviejo!  :herb:

PR #3100, opened 2020-10-23
feat: using npm-run-all -p to run scripts parallel

adispring:
[npm-run-all](https://github.com/mysticatea/npm-run-all/blob/HEAD/docs/npm-run-all.md) can run scripts parallel(or sequential), which can save some time.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,15 +76,15 @@
     "build:cjs": "cross-env BABEL_ENV=cjs babel source --out-dir src",
     "build:umd": "cross-env NODE_ENV=development rollup -c -o dist/ramda.js",
     "build:umd:min": "cross-env NODE_ENV=production rollup -c -o dist/ramda.min.js",
-    "build": "npm run build:es && npm run build:cjs && npm run build:umd && npm run build:umd:min",
+    "build": "npm-run-all --parallel build:**",
     "partial-build": "node ./scripts/partialBuild",
     "clean": "rimraf es/* src/* dist/* coverage/*",
-    "prepare": "npm run clean && npm run build",
+    "prepare": "npm-run-all --sequential clean build",
     "coverage": "BABEL_ENV=cjs nyc --reporter=lcov mocha -- --require @babel/register",
     "lint": "eslint scripts/bookmarklet scripts/*.js source/*.js source/internal/*.js test/*.js test/**/*.js lib/sauce/*.js lib/bench/*.js",
     "browser_test": "testem ci",
     "spec": "cross-env BABEL_ENV=cjs mocha --require @babel/register --reporter spec",
-    "test": "npm run spec && npm run lint"
+    "test": "npm-run-all --sequential spec lint"
   },
   "dependencies": {},
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -72,15 +72,15 @@
     "build:cjs": "cross-env BABEL_ENV=cjs babel source --out-dir src",
     "build:umd": "cross-env NODE_ENV=development rollup -c -o dist/ramda.js",
     "build:umd:min": "cross-env NODE_ENV=production rollup -c -o dist/ramda.min.js",
-    "build": "npm run build:es && npm run build:cjs && npm run build:umd && npm run build:umd:min",
+    "build": "npm-run-all --parallel build:**",
     "partial-build": "node ./scripts/partialBuild",
     "clean": "rimraf es/* src/* dist/* coverage/*",
-    "prepare": "npm run clean && npm run build",
+    "prepare": "npm-run-all --sequential clean build",
```
CrossEye, line 78: While I understand the rationale for the change above, why do you think this is better than the `&&` version?  Do you just see greater consistency by always using `npm-run-all`, or is it that this version is more explicit?

package.json
```diff
@@ -72,15 +72,15 @@
...
     "coverage": "BABEL_ENV=cjs nyc --reporter=lcov mocha -- --require @babel/register",
     "lint": "eslint scripts/bookmarklet scripts/*.js source/*.js source/internal/*.js test/*.js test/**/*.js lib/sauce/*.js lib/bench/*.js",
     "browser_test": "testem ci",
     "spec": "cross-env BABEL_ENV=cjs mocha --require @babel/register --reporter spec",
-    "test": "npm run spec && npm run lint"
+    "test": "npm-run-all --sequential spec lint"
```
CrossEye, line 83: Ditto.

CrossEye (commented):
This makes sense to me.  My note is just a question, as I have no objection to it.

CrossEye (dismissed):
This makes sense to me.  My comment was just a question.  But I'm fine with this either way.

package.json
adispring: There's no advantage to use `npm-run-all --sequential` than `&&` version, just looks consistent.
I have changed `npm-run-all --sequential` back to `&&`, and make `test` run `spec` and `lint` parallel.

CrossEye (dismissed):

CrossEye (dismissed):

CrossEye:
@adispring: Do you have time to look into the build problems?  (I manually resolved the `package-lock.json` differences and may have caused something here.)

adispring:
https://github.com/ramda/ramda/pull/2368#discussion_r790201868

adispring:
This PR(https://github.com/ramda/ramda/pull/3158) introduce `s` regex symbol which is an es2018 feature, eslint will not parse `/***/s` by default.

customcommander (approved):

customcommander:
@CrossEye: no changes since you last review other than the lock file needed to be regenerated. The rebase allowed to get the changes that fixed the linter errors.

Issue #3106, opened 2020-11-07
Add "modify" and "modifyPath" that applies a function to a property or path of an object


daniel7grant:
This issue proposes the addition of a new function to the default Ramda library.

## Proposal

Introduce a new pair of functions to the library that makes it possible to change a property (or path) of an object by applying a function to the value. It should change the value at the property (or path) on the passed object immutably, while maintaining referential equality for all other properties. If the property (or path) does not exist, or the last parameter is not an object or array, the function returns the input value unchanged.

The proposed name is `modify` and `modifyPath`, with the signatures:

```
modify ::     Idx -> (v -> v) -> {k: v} -> {k: v}
modifyPath :: [Idx] -> (v -> v) -> {k: v} -> {k: v}
```

## Motivation

Piggybacking off of the issue #2317, I feel like there is a missing link in the library for applying functions to a property or path of an object.

Consider how you can read and update list, objects, and through lenses.

| | List | Object | Lens |
|--|:--:|:--:|:--:|
|retrieve data | `nth` | `prop / path` | `view` |
|set data | `update` | `assoc / assocPath` | `set` |
|apply fn to data | `adjust` | ??? / ??? | `over` |

There is no simple way to change a value of a property with a function. Currently it can be done either using `over . lensProp` or `evolve`, hovewer these feel clunky comparing to, for example, `adjust`.

In the original issue, the new function is contrasted to `evolve({ [prop]: newVal })`. It is correct that functionally the two is the same, however `evolve` seems overkill for a simple update on one property. Consider saying that `merge({ [prop]: newVal })` offers the same functionality as `assoc(prop, newVal)`, when it is apparent that the latter feels more intuitive, and reduces the mental overhead. Adding a `modify` would also play nicely with the introduction of `modifyPath`, which just feels way more natural than `evolve({ prop1 : { prop2 : { /* ... */ }}})`, and provides updates with dynamic path generation.

So this change would mostly be syntactic sugar, however it would also give the additional benefit of partial application. Similarly how `assoc(prop)` can be used to create multiple updates, `modify(prop)` could be applied to create multiple function updates.

### Basic usage
```js
const person = { name: 'James', age: 20, pets: ['dog', 'cat'] };
R.modify('age', R.add(1), person); //=> { name: 'James', age: 21, pets: ['dog', 'cat'] }
R.modify('pets', R.append('turtle'), person); //=> { name: 'James', age: 20, pets: ['dog', 'cat', 'turtle'] }
```

[...]

PR #3107, opened 2020-11-09
new functions: modify and modifyPath

daniel7grant:
Referencing to issue #3106.

These changes introduce two new functions: `modify` and `modifyPath`. These functions change a value of a property (or path) of an object by applying a function to it. It modifies the prop (or path) in the original array immutably (returning a shallow copy), but copies all other keys by reference. If the given property (or path) does not exist on the object, then it returns the original object.

These functions are similar to `evolve`, however this function does not validate whether the second parameter is a function, it simply fails on call. Functionally it is mostly equivalent to `over(lensProp)` or `over(lensPath)`, however based on my benchmarks (Linux, Node v15 with [benchmark.js](https://gist.github.com/daniel7grant/b5a9410d1a260f8ece3912cccb7a99ea)) it is much faster than either `evolve` and lenses, being up to two times faster for long paths. For multiple path parts, `modifyPath` is even faster than using `assocPath`.

If you have any questions or suggestions, feel free to take them, this is my first contribution, I hope I'm doing everything correctly.


semmel:
Since `assoc` and `assocPath` [now](#2787) operate on array, how about `modify`/`modifyPath` support array in the same way too?

source/index.js
```diff
@@ -136,6 +136,8 @@ export { default as mergeWithKey } from './mergeWithKey.js';
 export { default as min } from './min.js';
 export { default as minBy } from './minBy.js';
 export { default as modulo } from './modulo.js';
+export { default as modify } from './modify.js';
```
CrossEye, line 139: A very minor nit: these should move *before* `modulo` too keep everything alphabetic.

CrossEye (changes requested):
Outside of one very minor nit, this looks great to me.  I think this is a good idea, and the implementation, at a brief look, seems quite well done.

I will need to spend a little more time on this soon, but I'm very happy with it.  :+1:

Others?  @ramda/core?

daniel7grant:
> Since `assoc` and `assocPath` since #2787 operate on array, how about `modify`/`modifyPath` support array in the same way too?

Under the hood `modifyPath` uses the same internal function as `assocPath` (`_assoc`), and its changed version for functions (`_modify`). Both of these handle the array case separately, so these functions should work for array indices in any point in the path.

source/index.js
daniel7grant: Ah, definitely, it was a mistake on my part.

source/internal/_modify.js
```diff
@@ -0,0 +1,29 @@
...
+  if (_isInteger(prop) && _isArray(obj)) {
+    var arr = [].concat(obj);
+    arr[prop] = fn(arr[prop]);
+    return arr;
+  }
+
+  var result = {};
+  for (var p in obj) {
+    result[p] = obj[p];
+  }
```
hitmands, line 26: Unless intentional, 
I'd suggest to wrap the assignment into a `hasOwnProperty` 
to avoid performing the operation in the prototype chain

https://steemit.com/programming/@heejin/javascript-hasownproperty-in-for-in-loops

```ts
var obj = {
  a: 1,
  b: 2
};

Object.prototype.haha = 3;

for (var k in obj) {
  console.log(k); // prints a, b, haha
}
```

daniel7grant: Thank you Giuseppe for your suggestion. I am aware that many linters advise adding own property checks in for loops, however in this day and age (and in this paradigm) I am unsure what is the benefit of this. Prototypal inheritance is wildly discouraged, and does not really fit into the functional paradigm Ramda promotes. I am uncertain about the impact on the performance, but I'm guessing it slows down the loop considerably.

Other than this, the `_modify` function is a pound for pound copy of the `_assoc` function, so I think this question might be beyond the scope of this change. I looked through some similar functions (e. g. `evolve`), and none of them tested for inherited properties. Probably @CrossEye can help whether we should look into it or not.

CrossEye: Note that `assoc` already works this way.  From [the docs][as]:

> Note that this copies and flattens prototype properties onto the new object as well. 

So I would stick with that for `modify`.

  [as]: https://ramdajs.com/docs/#assoc

hitmands: agreed, let's follow `assoc`'s implementation

CrossEye:
@hitmands: This [is failing][cc] a CodeClimate check, one which I don't think is relevant to us.  I don't really understand the tool.  Obviously I can force-merge this to ignore the issue, but does CodeClimate have some mechanism to say "No, this time it's ok."?

  [cc]: https://codeclimate.com/github/ramda/ramda/pull/3107

source/modifyPath.js
```diff
@@ -0,0 +1,55 @@
...
+ *
+ *      const person = {name: 'James', address: { zipCode: '90216' }};
+ *      R.modifyPath(['address', 'zipCode'], R.reverse, person); //=> {name: 'James', address: { zipCode: '61209' }}
+ *
+ *      // Can handle arrays too
+ *      const person = {name: 'James', addresses: [{ zipCode: '90216' }]};
+ *      R.modifyPath(['addresses', 0, 'zipCode'], R.reverse, person); //=> {name: 'James', addresses: [{ zipCode: '61209' }]}
+ */
+var modifyPath = _curry3(function modifyPath(path, fn, object) {
+  if (!_isObject(object) && !_isArray(object) || path.length === 0) {
```
CrossEye, line 35: I'd like to see an extra pair of parentheses here.  Yes, `A && B || C` is specified to be `(A && B) || C`, but I think it's too easy to imagine it as `A && (B || C)`.

CrossEye (changes requested):
Another minor nit.  But otherwise this looks good to me.

:herb:

hitmands:
> @hitmands: This [is failing](https://codeclimate.com/github/ramda/ramda/pull/3107) a CodeClimate check, one which I don't think is relevant to us. I don't really understand the tool. Obviously I can force-merge this to ignore the issue, but does CodeClimate have some mechanism to say "No, this time it's ok."?


Hey @CrossEye yes, you have a couple of options:

1. Mark the issue as `not an issue`
2. Approve the PR


image
abc

Bradcomp (dismissed):

CrossEye:
@hitmands:

That's what I was expecting.

I wonder if I don't have edit authority on codeclimate.  That "New" dropdown you were seeing just looked like a tag label, and didn't seem like I could perform an action there.  Not too worried about it at the moment.  Next time I'll try harder.

CrossEye:
@daniel7grant: do you want to make those minor tweaks?  I'm ready to merge this when those are done.  It seems both useful and well done.

hitmands:
@CrossEye let me check

CrossEye:
@hitmands: Thanks.  If there's a way to give authority to a team, I'd go with all of @ramda/core.

hitmands:
@CrossEye let's take this on a separate thread, however you should have admin access on cc too... 

Can u see the settings tab in cc?

image

daniel7grant:
> @daniel7grant: do you want to make those minor tweaks? I'm ready to merge this when those are done. It seems both useful and well done.

Thanks @CrossEye, I finished fixing the minor issues. All green across the board (including adding one semicolon to the partialBuild script). I'm glad I could contribute, really enjoying the library.

source/modifyPath.js
daniel7grant: I feel like I removed it, as my linter marked it as not useful

CrossEye (approved):

CrossEye:
Thanks @daniel7grant -- this looks like a very worthy addition.  Hopefully we'll push a build soon.

CrossEye:
@hitmands: This is my view:

Main CC page

And the user profile page:

Profile CC page

(The other tabs don't seem to have anything useful either.)

PR #3111, opened 2020-11-22
chore: rename _iterableReduce to _iteratorReduce

adispring:
According to [MDN document](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols), `iterable` an `iterator ` are two different thing:

* `iterable` is an object that can create an `iterator` by `@@iterator` method;
* `iterator` is  an object that implements a `next()` method, which can produce a sequence of values;

So in the `_reduce`'s implementation,  `_iterableReduce` processes `iterator` instead of `iterable` actually. I think it's better to be called as `_iteratorReduce`.

### DIFF
--- a/source/internal/_reduce.js
+++ b/source/internal/_reduce.js
@@ -56,5 +56,5 @@ export default function _reduce(fn, acc, list) {
     return _methodReduce(fn, acc, list, 'reduce');
   }
 
-  throw new TypeError('reduce: list must be array or iterable');
+  throw new TypeError('reduce: list must be array, iterable or iterator');
 }
--- a/source/internal/_reduce.js
+++ b/source/internal/_reduce.js
@@ -17,15 +17,15 @@ function _arrayReduce(xf, acc, list) {
   return xf['@@transducer/result'](acc);
 }
 
-function _iterableReduce(xf, acc, iter) {
-  var step = iter.next();
+function _iteratorReduce(xf, acc, iterator) {
+  var step = iterator.next();
   while (!step.done) {
     acc = xf['@@transducer/step'](acc, step.value);
     if (acc && acc['@@transducer/reduced']) {
       acc = acc['@@transducer/value'];
       break;
     }
-    step = iter.next();
+    step = iterator.next();
   }
   return xf['@@transducer/result'](acc);
 }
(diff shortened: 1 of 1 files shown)


CrossEye:
I think this raises a much broader point.

*Should* we be accepting iterators?  That strikes me very much against the spirit of Ramda.  Accepting iterables keeps the mutation inside our function.  But accepting iterators seems to lead to potential problems with referential transparency.  Iterators are inherently stateful, and I really don't like this:

```
const arr= [1, 2, 3]
const foo = arr [Symbol.iterator] ()

reduce (add, 0, foo) //=> 6
reduce (add, 0, foo) //=> 0  // Whoa!!!
```

With iterables, this just makes sense:

```
const bar = {
  [Symbol.iterator]: (i = 0) => ({
    next: () => i < 3 ? {done: false, value: ++i} : {done: true}      
  })
}

reduce (add, 0, bar) //=> 6
reduce (add, 0, bar) //=> 6
```

I believe we need to think this through pretty carefully.

adispring:
Yes, iterator will change itself's state when calling `next()`. So, let's remove iterator from `reduce`?

CrossEye:
@adispring:

> Yes, iterator will change itself's state when calling `next()`. So, let's remove iterator from `reduce`?

That's my current thought.  But I believe this needs significantly more deliberation.

Anyone from @ramda/core want to chime in?  How about @buzzdecafe or @kevinbeaty ?

CrossEye:
ping @ramda/core:  Any thoughts on this?

customcommander:
My preference is to accept iterables only.

kevinbeaty:
I support removing iterator support  (i.e. only support iterables).

I don't really remember why we supported iterators, but I think it might have been to support existing libraries that added iterators before iterables were well defined. Similar to "thenables" vs `Promise`.

CrossEye:
Ok, then I think we have a plan.  @cusomcommander: Do you think the change belongs in `1.0` or some post-`1.0` release?

customcommander:
I think v1.0 is likely to contain some breaking changes so I'd prefer we have as many as possible in v1.0. So yeah I'd suggest this is part of the v1.0 release. Perhaps we could also record that decision (iterators vs iterables) somewhere? I'm a big fan of ADRs.

CrossEye:
@customcommander: I haven't used ADR's in a GitHub project.  Do you know of a good built-in mechanism that supports them?  Obviously we can just record them in the wiki or some such, if no common method exists.

PR #3116, opened 2020-12-06
revert "align lensPath with lensIndex by supporting negative indeces'

ku8ar:
This PR removes support of negative indices in functions: lensIndex, prop, path, paths, pathEq, pathOr, propEq, propIs, propOr.

The first PR in which this logic was added: https://github.com/ramda/ramda/pull/2670

API inconsistency that was reported as an bug: https://github.com/ramda/ramda/issues/2974

An example that shows why this logic is not compatible with the rest of the functions in Ramda (and JS itself):
```javascript
const ar = ['a', 'b']
const idxOfC = findIndex(equals('c'))(ar)
console.log( prop(idxOfC, ar) ) // => 'b'
```

Interest: @2beaucoup @CrossEye @Bradcomp @Andrey-Bazhanov

### DIFF
--- a/source/lensIndex.js
+++ b/source/lensIndex.js
@@ -15,7 +15,7 @@ import update from './update.js';
  * @sig Number -> Lens s a
  * @param {Number} n
  * @return {Lens}
- * @see R.view, R.set, R.over, R.nth
+ * @see R.view, R.set, R.over
  * @example
  *
  *      const headLens = R.lensIndex(0);
--- a/source/path.js
+++ b/source/path.js
@@ -13,7 +13,7 @@ import paths from './paths.js';
  * @param {Array} path The path to use.
  * @param {Object} obj The object to retrieve the nested property from.
  * @return {*} The data at `path`.
- * @see R.prop, R.nth
+ * @see R.prop
  * @example
  *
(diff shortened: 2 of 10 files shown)


CrossEye:
I'm wondering about `slice` too.  While I'd hate to lose that -- I use the vanilla equivalent all the time -- it might be time to simply gain this consistency across the board.  There may be others too; I'll try to look this evening.

ku8ar:
@CrossEye I don't understand. R.slice is basically the same as Array.slice...
```
var slice = _curry3(_checkForMethod('slice', function slice(fromIndex, toIndex, list) {
  return Array.prototype.slice.call(list, fromIndex, toIndex);
}));
```
So in this case, removing the negative indices handler will be additional logic...

Another question: Is there a plan/calendar for release Ramda 1.0?

CrossEye:
@ku8ar:
> So in this case, removing the negative indices handler will be additional logic...

Yes it would.  But it might be nice to be able to answer the question, "Which Ramda functions allow for negative indices?" with "None of them."

> Another question: Is there a plan/calendar for release Ramda 1.0?

It's always, *some day*.  But we never seem to get to it.  If you are interested in chipping in to make that happen, please let us know.

PR #3119, opened 2020-12-14
Update partial.js docs

carlosmori:
As a newbie to Ramda, I find it clearer this way


CrossEye (approved):
I have no problem merging this.  I'm curious why you find this clearer.  Is it because of the abundance of `2`s in the example, including `multiply2`, `[2]`, and `double`?

carlosmori:
> I have no problem merging this. I'm curious why you find this clearer. Is it because of the abundance of `2`s in the example, including `multiply2`, `[2]`, and `double`?

Thanks!
Yes exactly!

CrossEye:
@carlosmori: Thank you very much for your contribution!

:herb:

PR #3126, opened 2020-12-31
if n equals 0, _xtake use _reduced to short circuit the iteration.

adispring:


CrossEye:
Could you add an explanation, for those of us who never worked closely with the transducer implementation?

What is the advantage of the change?  Is there a failing edge case, or a larger failure altogether?  Or do you believe this is simply cleaner code?

Also if I read this correctly:, 

```js
  var ret = this.n === 0 ? _reduced(result) : this.xf['@@transducer/step'](result, input);
  return this.n >= 0 && this.i >= this.n ? _reduced(ret) : ret;
```

we could end up returning `_reduced (_reduced (result))`.  Is that a problem?

adispring:
```js
  var ret = this.n === 0 ? _reduced(result) : this.xf['@@transducer/step'](result, input);
  return this.n >= 0 && this.i >= this.n ? _reduced(ret) : ret;
```

Though there's no problem to do `_reduced(_reduced(result)`, there's also no need to wrap `result` early.

`_reduced` will only wrap the real value once:

```js
export default function _reduced(x) {
  return x && x['@@transducer/reduced'] ? x :
    {
      '@@transducer/value': x,
      '@@transducer/reduced': true
    };
}
```

I thought it will iterate more than 1 time when `n` is 0, but I'm wrong, it will exit after the first iteration, as follows:

```js
return this.n >= 0 && this.i >= this.n ? _reduced(ret) : ret;
```

`n` equals 0, and `i` is 0 too, so the `ret` will be wrapped with `_reduced` in the first iteration, and exit immediately.

PR #3129, opened 2021-01-14
Added unwindAll function with unwindAll test files and minor correction in unwind function.

Jayesh-30:
#3125


CrossEye:
I'm trying to decide how necessary this seems.  We just added `unwind` mostly on speculation.  It looks like a useful function, but should we simply leave this wrapper for users to define if they need it?  It wouldn't be hard, perhaps something like:

```js
const unwindAll = (keys) => (obj) => 
  keys .reduce ((os, k) => os .flatMap (unwind (k)), [obj])
```

Can you offer a strong argument for including it?

PR #3134, opened 2021-02-06
adjust code refactoring : using modulo to cycle through indices 

9kartik:
Takes care of positive and negative indices, getting rid of the start logic; adding array length as a variable as it was being reused often.

adjust test cases pass too.


CrossEye (approved):
This is a minor clean-up, but clearly well worth doing.

Anyone else have feedback?

:herb:

Bradcomp (approved):
:bowtie:

CrossEye:
Merged.  Thank you very much 9kartik!

PR #3135, opened 2021-02-19
fix: clone typed arrays

Liam-Tait:
When cloning a typed array, the value is a reference to the original value.
This can be fixed by cloning the array using `.slice()`

An example to see how clone is currently behaving
```js
import * as R from "ramda";

const a = new Int8Array([1]);
const b = R.clone(a);

b[0] = 100;

console.log(a[0]); // 100
console.log(b[0]); // 100
```

### DIFF
--- a/source/internal/_clone.js
+++ b/source/internal/_clone.js
@@ -35,7 +35,19 @@ export default function _clone(value, refFrom, refTo, deep) {
     case 'Object':  return copy(Object.create(Object.getPrototypeOf(value)));
     case 'Array':   return copy([]);
     case 'Date':    return new Date(value.valueOf());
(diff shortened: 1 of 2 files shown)


kedashoe:
Nice catch!

Looks like you have `Uint16Array` tests twice and are missing BigInt?

CrossEye (approved):
Some day we'll get 1.0 out the door and allow ourselves to properly support modern JS features.  But this should still work in older versions, as we're only using the *names* of those types.  So it looks good to me.

:herb:

Anyone else?  @ramda/core?

Bradcomp (approved):
:bowtie:

PR #3136, opened 2021-03-03
chore: There is no need to check if acc is empty every time.

adispring:


CrossEye (approved):
That's a bizarre implementation you're fixing here.  I'd be curious to trace the history and see why we did it this way... except that I'm afraid I'd find out it was my fault!  :smiling_imp:

mattgrande:
I was curious, so I looked into the history, and you're off the hook, @CrossEye  :)

https://github.com/ramda/ramda/pull/1704#discussion_r58328270

davidchambers:

> I haven't been following this pull request closely, so this may already have been discussed, but why must we use `null` as the initial value rather than `[]`?

asaf-romano:

> Because the array instance is shared otherwise. That's what you get for mutability. Another option is to special case the empty array case, but I think passing null is cleaner, as far as "clean" can get in this case.

That being said, I still feel like this change is an improvement.

adispring:
Everytime `reduceBy` generates a new subgroup, it will deep clone the initial accumulator value passed to `reduceBy` as its initial value, as following:

https://github.com/ramda/ramda/blob/84af16091e8c19518f189a8cf087d1318ff57d60/source/reduceBy.js#L58

Though `acc.push(item)` will make acc mutable, but acc is not shared between subgroups.

So, I think it is ok to replace `null` with `[]`.

Bradcomp (approved):
:bowtie:

PR #3140, opened 2021-04-02
fix: isOdd: n % 2 may be 1 and -1.

adispring:
Negative integer may also be odd number, and negative integer moduloed by 2 returns -1.

So the correct way to check whether an integer is odd or even, is as follows:

```js
const isOdd = n => n % 2 !== 0;
const isEven = n => n % 2 === 0;
```


CrossEye (approved):
Yes, a good idea, and a nice catch!

Issue #3143, opened 2021-04-23
count function?


Nondv:
Hi!

I didn't find anything similar to `count(pred, list)` function that would count items with `pred(x) == true`.

Did I miss something?

I think it's a very common function and it makes sense to have it already defined and curried.

Usage example: 

```javascript
const isEven = n => n % 2 === 0;

R.count(isEven, [1, 2, 3, 4]); //=> 2
```

(obviously, it's quite nice to use in pipelines)

I added this ineffective definition to my code:

```javascript
const count = curry(compose(length, filter))
```

A better implementation would be something like:

```javascript
const count = curry((pred, list) => reduce((a, e) => pred(e) ? a + 1 : a, 0, list))
```

or with a transducer:

```javascript
const count = curry((pred, list) => transduce(filter(pred), inc, 0, list))
```

CrossEye:
@Nondv:

Please feel free to create a PR for this, if you like.  It should be a simple enough function to write.

I can't guarantee that a PR would be accepted, but I can promise a fair hearing.

Nondv:
@CrossEye will do. I was just wondering if I'm missing something or if this was already discussed

PR #3144, opened 2021-04-26
Add `count` function

Nondv:
`(a -> Boolean) -> [a] -> Number`

counts items in a list that match a predicate.

#3143

### DIFF
--- a/source/count.js
+++ b/source/count.js
@@ -1,0 +1,26 @@
+import reduce from './reduce.js';
+import curry from './curry.js';
+
+
+/**
+ * Returns the number of items in a given `list` matching the predicate `f`
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig (a -> Boolean) -> [a] -> Number
+ * @param {Function} predicate to match items against
+ * @return {Array} list of items to count in
+ * @example
+ *
+ *      const even = x => x % 2 == 0;
+ *
+ *      R.count(even, [1, 2, 3, 4, 5]); // => 2
+ *      R.map(R.count(even), [[1, 1, 1], [2, 3, 4, 5], [6]]); // => [0, 2, 1]
+ */
+var count = curry(function(pred, list) {
+  var r = function(a, e) { return pred(e) ? a + 1 : a; };
+  return reduce(r, 0, list);
+});
+
+export default count;
--- a/source/count.js
+++ b/source/count.js
@@ -19,8 +19,7 @@ import curry from './curry.js';
  *      R.map(R.count(even), [[1, 1, 1], [2, 3, 4, 5], [6]]); // => [0, 2, 1]
  */
 var count = curry(function(pred, list) {
-  var r = function(a, e) { return pred(e) ? a + 1 : a; };
-  return reduce(r, 0, list);
+  return reduce(function(a, e) { return pred(e) ? a + 1 : a; }, 0, list);
 });
 
 export default count;
--- a/source/count.js
+++ b/source/count.js
@@ -1,4 +1,4 @@
-import reduce from './reduce.js';
+import _reduce from './internal/_reduce.js';
 import curry from './curry.js';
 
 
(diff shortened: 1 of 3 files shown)


source/count.js
```diff
@@ -0,0 +1,26 @@
...
+ *
+ *      const even = x => x % 2 == 0;
+ *
+ *      R.count(even, [1, 2, 3, 4, 5]); // => 2
+ *      R.map(R.count(even), [[1, 1, 1], [2, 3, 4, 5], [6]]); // => [0, 2, 1]
+ */
+var count = curry(function(pred, list) {
+  var r = function(a, e) { return pred(e) ? a + 1 : a; };
+  return reduce(r, 0, list);
+});
```
CrossEye, line 24: I would probably choose to inline this short function inside the call to `reduce`.

```js
reduce(function(a, e) {return pred(e) ? a + 1: a}, 0, list);
```

But I don't think we've established any hard rules on that, and this version is fine.

test/count.js
```diff
@@ -0,0 +1,26 @@
+var R = require('../source');
+var eq = require('./shared/eq');
+
+describe('count', function() {
+  var even = function(x) { return x % 2 === 0; };
+  it('counts items in a list that match a filter', function() {
+    eq(R.count(even, [1, 2, 3, 4, 5]), 2);
+    eq(R.count(even, [1, 2, 3, 6]), 2);
+    eq(R.count(even, [1, 2, 3]), 1);
```
CrossEye, line 9: Three examples here seems overkill.

test/count.js
```diff
@@ -0,0 +1,26 @@
...
+
+  it('returns zero when no element matches', function() {
+    eq(R.count(even, [1, 3, 5, 7]), 0);
+  });
+
+  it('returns zero for an empty list', function() {
+    eq(R.count(even, []), 0);
+  });
+
+  it('is curried', function() {
```
CrossEye, line 20: While some old function still have this test, we generally skip it for new function.  That's just how the library works.  You can work curried calls into some of the earlier examples without calling it out directly.

CrossEye (changes requested):
There are a few minor comments here.  But I like this idea.  The function is clear and obvious, and I know now that I'm not the only one who's needed this on occasion.  Granted it's not hard to write, either on top of `filter` or `reduce`, but this feels like the sort of thing that Ramda would like to supply.

source/count.js
```diff
@@ -0,0 +1,26 @@
+import reduce from './reduce.js';
```
Bradcomp, line 1: I think it would be better to use the internal/_reduce function here. That way we aren't dealing with all the wrapper stuff.

Nondv: It seems that it's not a convention to choose internal functions over public ones.

Since it's not a convention, I don't think it's generally a good idea to pick internal functions when there're public available.

The problem with using internal functions is that they don't guarantee an API whereas public ones do (because they are actually consumed by users). In terms of optimisation, the only overhead is the currying. I don't believe it's worth it :)

test/count.js
Nondv: I'm more on a side that having tests is better than not having tests.

But if it's conventional here I'm fine with removing 👍

test/count.js
Nondv: Tried to provide some variety

source/count.js
Nondv: I don't see much difference, just a personal call. Not against your suggestion tho

test/count.js
Nondv: Replaced :)

source/count.js
Nondv: DONE

test/count.js
Nondv: DONE

source/count.js
Bradcomp: There is no danger in using internal functions because we have extensive testing of the API that will prevent regressions even if internal functions change. While it has been standard practice for public functions to rely on other public functions, it's something we'd like to move away from instead of continuing the practice.

CrossEye: Agreed.  We added the internal helpers mostly as a performance boost, and I would stick with them when they're available (and sometimes add them when they're not.)

adispring:
Ramda already have a function called `countBy`, but `count` may be more convenient in some cases.

source/count.js
Nondv: DONE

used internal `_reduce`. Tbf, the overhead it removes is just currying. It's hard to imagine a scenario where that would make any impact :)

Bradcomp (approved):
:bowtie:

CrossEye (approved):
Looks great!

PR #3149, opened 2021-05-20
Implement Retry function

ngocviet:


CrossEye:
@ngocviet: 

Please add some description.  What does your function do?  Why does the world need it?  How well does it fit with other Ramda functions?

PR #3155, opened 2021-05-30
Add retry function to Ramda

hungdt8796:
To retry a function until it returns true or reaches the maximum tries


CrossEye:
@hungdt8796:

Please update to explain the rationale for your function and the reason you think it belongs in Ramda.

ku8ar:
This function is not very suitable for this library. Promise is not clean.

CrossEye:
@hungdt8796: Any follow up or should I just close this?

PR #3158, opened 2021-06-07
fix, add use first flags property, and support dotAll flag

HoKangInfo:


Bradcomp (approved):
This looks good, would it be possible to add a test for the flags?

HoKangInfo:
> This looks good, would it be possible to add a test for the flags?

I can not add test case, because ESlint raise `Invalid regular expression flag`.
unless update es6 to '2020' or '2021'

HoKangInfo:
> > This looks good, would it be possible to add a test for the flags?
> 
> I can not add test case, because ESlint raise `Invalid regular expression flag`.
.eslintrc
```
    "parserOptions": {
        "ecmaVersion": 8, // 8,  9 or 10
     }
```       

test/tets.js
```
   it('returns true if string matches dotAll pattern', function() {
     eq(R.test(/x.*z/s, 'x.\nyz'), true);
   });

```

adispring:
@HoKangInfo You can specify eslint environments using a comment inside of a JavaScript file, [as follows](https://eslint.org/docs/user-guide/configuring/language-options#using-configuration-comments): 

```js
/* eslint-env es2018 */
```

adispring:
Ramda is compatible with es5, but dotAll flag is a es2018 feature. Will this cause some trouble when people using dotAll without polyfill？

CrossEye:
@adispring:

> dotAll flag is an es2018 feature.
> Will this cause some trouble when people using dotAll without polyfill？

Yes, but it's not trouble caused by Ramda.  If they are going to use features from a later version of the language, we don't interfere.  Heck, Ramda is almost compatible with ES3, but I never use it these days with anything less than ES6/2015.

source/internal/_cloneRegExp.js
```diff
@@ -1,7 +1,8 @@
...
-                                    (pattern.multiline  ? 'm' : '') +
-                                    (pattern.sticky     ? 'y' : '') +
-                                    (pattern.unicode    ? 'u' : ''));
+  return new RegExp(pattern.source, (pattern.flags ?  pattern.flags : (pattern.global     ? 'g' : '') +
+                                                                      (pattern.ignoreCase ? 'i' : '') +
+                                                                      (pattern.multiline  ? 'm' : '') +
+                                                                      (pattern.sticky     ? 'y' : '') +
+                                                                      (pattern.unicode    ? 'u' : '') +
+                                                                      (pattern.dotAll     ? 's' : '')));
 }
```
CrossEye, line 8: This looks great.  But could you try the trick mentioned by @adispring to add some test cases?

CrossEye (changes requested):

HoKangInfo:
> Ramda is compatible with es5, but dotAll flag is an es2018 feature.
> Will this cause some trouble when people using dotAll without polyfill？

I think OK, because only exist `pattern.dotAll`, 's' is appended.

source/internal/_cloneRegExp.js
HoKangInfo: i try it, and test is passed.

CrossEye (approved):

adispring (changes requested):

test/test.js
```diff
@@ -6,7 +7,7 @@ var eq = require('./shared/eq');
 
 describe('test', function() {
   it('returns true if string matches dotAll pattern', function() {
-    eq(R.test(/x.*z/s, 'x.\nyz'), true);  /* eslint-env es2018 */
+    eq(R.test(/x.*z/s, 'x.\nyz'), true);
   });
 
   it('returns true if string matches dotAll pattern', function() {
```
adispring, line 13: This test case is duplicated with the upper one?

adispring (changes requested):

test/test.js
HoKangInfo: I remove it.
but  `/* eslint-env es2018 */`comment  is Invalid
```
10:16  error  Parsing error: Invalid regular expression flag

✖ 1 problem (1 error, 0 warnings)

```

adispring: I find out that, there is no eslint environment called `es2018`, the latest eslint only has es6, es2017, es2020 and es2021.

You can update `eslint` and `eslint-plugin-import` to the latest version, and replace `/* eslint-env 2018 */` to `/* eslint-env 2020 */`, and change some .eslintrc configs related to `eslint-plugin-import`.

But we can also just remove the `/* eslint-env 2018 */`, merge this PR, and update eslint in another pr later.

HoKangInfo: `eslint-env` [Can't set env with a comment](https://github.com/eslint/eslint/issues/2705).
I remove the eslint-env comment. You can help me update eslint in another pr later.

adispring: ok

adispring (approved):

test/test.js
adispring: Can you disable eslint check in the line `eq(R.test(/x.*z/s, 'x.\nyz'), true);` temporarily ?

I will update eslint related things later.

PR #3160, opened 2021-06-11
chore: return val will be more precise

adispring:
In the [current `paths` implementation](https://github.com/ramda/ramda/blob/v0.27.0/source/paths.js#L29), current iterating path of `paths` will always return `undefined` if the intermediate path value is nil (`val == null`) . Return the actual nil value may be more precise?

### DIFF
--- a/source/paths.js
+++ b/source/paths.js
@@ -14,7 +14,7 @@ import nth from './nth.js';
  * @param {Array} pathsArray The array of paths to be fetched.
  * @param {Object} obj The object to retrieve the nested properties from.
  * @return {Array} A list consisting of values at paths specified by "pathsArray".
- * @see R.path
+ * @see R.path, R.props
  * @example
  *
  *      R.paths([['a', 'b'], ['p', 0, 'q']], {a: {b: 2}, p: [{q: 3}]}); //=> [2, 3]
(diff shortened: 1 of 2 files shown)


CrossEye:
I can't decide if this makes sense or not.

Is it better to consistently return `undefined` if the path is not found or is it better to always return the specific nilary value supplied or found somewhere on the path.  The big argument against it to my mind is that the function, by design, does not give you information about where along the path we found a nil value.  So I'm not sure it makes sense to say, "Well, either `foo` is `null` or `foo` is non-`null` but `foo.bar` is `null`, or `foo` and `foo.bar` are non-`null`, but `foo.bar.baz` is `null`.  But we can tell you for sure that one of those values was `null` and not `undefined`."

It feels odd.

But I usually like that general idea, of making sure the specifics of an oddball case are passed along to the caller.  So I'm a little torn.

One concern is that such a change could conceivably cause a regression for people who test specifically for `undefined`.

Do you have a strong selling point for this?

adispring:
I check the `path` 's implementation again, return `undefined` is the right behavior if the intermediate path is nil. There's no need to worry if the endpoint value of the path is `null` or `undefined`, `path` will return the right value in this situation.

Issue #3157, opened 2021-06-05
use Promise.all after ap, raise error  


HoKangInfo:
ex:
```
pipe(
  of,
  ap([
    () => Promise.resolve('a'),
    () => Promise.resolve('b')
  ]),
  Promise.all,  // TypeError: #<Object> is not a constructor
  // x => Promise.all(x),  // OK
  andThen(x => console.log(x))  // -> [ 'a', 'b' ]
)('test')
```
 UnhandledPromiseRejectionWarning: TypeError: #<Object> is not a constructor
    at all (<anonymous>)

kedashoe:
Hi @HoKangInfo , this is a javascript thing rather than a ramda thing. When passing `Promise.all` (or any function!) to `pipe` (or anywhere!) the function loses its context. See for example https://stackoverflow.com/questions/44632732/why-passing-directly-promise-all-to-a-then-function-throws-an-error. As you observed, putting it in a lambda works. You could also use `bind` as in that stackoverflow answer. Hope that helps.

CrossEye:
I usually write a quick helper such as `const allPromises = (ps) => Promises.all(ps)` to use in pipelines.

HoKangInfo:
or add function `allThen` 
ex:
```
allThen (func, pss) => Promises.all(pss).then(func)
```

HoKangInfo:
Thx

PR #3165, opened 2021-06-24
add allThen function

HoKangInfo:
use Promise.all after ap, raise error [#3157](https://github.com/ramda/ramda/issues/3157)
add allThen function, simplify `Promise.all.bind(Promise)`.


source/allThen.js
```diff
@@ -0,0 +1,51 @@
...
+ * @example
+ *
+ *      const p1 = s => Promise.resolve(s + 1);
+ *      const p2 = s => Promise.resolve(s + 2);
+ *
+ *      const  = R.pipe(
+ *        R.of,
+ *        R.ap([p1, p2]),
+ *        R.allThen(R.join('|'))  // => 'test1|test2'
+ *      );
```
CrossEye, line 36: Although we're not entirely consistent with this, I think it's better if our examples rarely use any Ramda functions outside the one being demonstrated.  While `join` is probably clear to most readers, both `of` and `ap` might cause confusion.

Perhaps better would be something like this:

```js
R .allThen (xs => xs .join ('|')) ([Promise .resolve ('foo'), Promise .resolve ('bar')])
  .then (console .log) //~> "foo|bar"
```

source/allThen.js
```diff
@@ -0,0 +1,51 @@
...
+ *
+ */
+var allThen = _curry2(function allThen(f, ps) {
+  if (Promise == null || !_isFunction(Promise.all)) {
+    throw new TypeError('Promise is not defined');
+  }
```
CrossEye, line 42: I'll be glad when we can stop doing these tests!

source/allThen.js
```diff
@@ -0,0 +1,51 @@
...
+
+  if (!(_isString(ps) || _isArrayLike(ps) || _isIterator(ps))) {
+    ps = _of(ps);
+  }
```
CrossEye, line 46: Ramda usually strives for simplicity.  I don't think allowing for either scalars or arrays really fits.

CrossEye (changes requested):
Although this is a clearly useful function (and, I might add, a nicely done pull request), I'm very hesitant to add this.

The Ramda team has repeatedly expressed its reservation about Promises.  Most of the reasons are captured in [Broken Promises][bp] -- but the biggest one is that they are so close to lawful abstract types such as `Functor` and `Chain`, but they blow it, which means that you have to be very careful when trying to use them with generic code.

So Ramda has managed to reduce its direct support for Promises to only the two functions `andThen` and `otherwise`.  It's not that Promises don't work with Ramda code, just that Ramda does not go out of its way to work with them.  This makes me really hesitant to increase that support.

So although this review suggests that changes are needed, I would suggest holding off on them until and unless you first convince us that it's worth offering any additional Promise support, especially when this is trivial for users to write as 

```js
const allThen = (fn) => (ps) => Promise. all (ps) .then (fn)
```



  [bp]:https://avaq.medium.com/broken-promises-2ae92780f33

source/allThen.js
HoKangInfo: Thanks, it's really more concise.

source/allThen.js
HoKangInfo: I know this will cause confusion, but It's more convenient, determine whether to use `andThen`.
I will remove it.

CrossEye:
@HoKangInfo:

Do you want to argue in favor of this?  Or can we just go ahead and close it?

PR #3169, opened 2021-07-14
Add mergeBy function.

rodrigoruiz:
I don't think Ramda currently has a function equivalent to that mergeBy I implemented, so I'd like others' opinions on whether this is relevant to everyone else and if it is implemented efficiently.

If the community deem this feature worth it, I'll add some tests.

Thank you, any comments are appreciated.

### DIFF
--- a/source/index.js
+++ b/source/index.js
@@ -126,6 +126,7 @@ export { default as mean } from './mean.js';
 export { default as median } from './median.js';
 export { default as memoizeWith } from './memoizeWith.js';
 export { default as mergeAll } from './mergeAll.js';
+export { default as mergeBy } from './mergeBy.js';
 export { default as mergeDeepLeft } from './mergeDeepLeft.js';
 export { default as mergeDeepRight } from './mergeDeepRight.js';
 export { default as mergeDeepWith } from './mergeDeepWith.js';
--- a/source/mergeBy.js
+++ b/source/mergeBy.js
@@ -1,0 +1,67 @@
+import _curryN from './internal/_curryN.js';
+import concat from './concat.js';
+import fromPairs from './fromPairs.js';
+import keys from './keys.js';
+import map from './map.js';
+import pipe from './pipe.js';
+import uniq from './uniq.js';
+
+
+/**
+ * Creates a new list by merging two provided lists based on a merge criteria.
+ * The merge criteria key is calculated for each element of each list.
+ * It is assumed that each list will produce a unique key for each of its elements.
+ * Then a merge function is used to merge the corresponding item of each list
+ * into a single object.
+ * If a key does not have a corresponding item in one of the lists, undefined is
+ * passed to the merge function for the corresponding argument.
+ * The result list will be the size of the number of unique keys among the two
+ * input lists.
+ *
+ * @func
+ * @memberOf R
+ * @since v0.19.0
+ * @category List
+ * @sig (a -> k) -> ((a?, a?) -> b) -> [a] -> [a] -> [b]
+ * @param {Function} mergeCriteria
+ * @param {Function} mergeFunction
+ * @param {Array} left
+ * @param {Array} right
+ * @return {Array}
+ * @example
+ *
+ *      R.mergeBy(
+ *        R.prop('key'),
+ *        (leftItem, rightItem) => ({
+ *          key: leftItem?.key ?? rightItem?.key,
+ *          leftValue: leftItem?.value ?? 5,
+ *          rightValue: rightItem?.value ?? 6
+ *        })
+ *        [{ key: 'a', value: 1 }, { key: 'b', value: 2 }],
+ *        [{ key: 'b', value: 3 }, { key: 'c', value: 4 }]
+ *      );
+ *      //=> [
+ *      //  { key: 'a', leftValue: 1, rightValue: 6 },
+ *      //  { key: 'b', leftValue: 2, rightValue: 3 },
+ *      //  { key: 'c', leftValue: 5, rightValue: 4 }
+ *      //]
+ */
+ export default _curryN(4, [], (mergeCriteria, mergeFunction, left, right) => {
+  const getItemByCriteria = pipe(
+      map(item => [mergeCriteria(item), item]),
+      fromPairs
+  );
+
+  const leftItemByCriteria = getItemByCriteria(left ?? []);
+  const rightItemByCriteria = getItemByCriteria(right ?? []);
+
+  const uniqueKeys = uniq(concat(
+      keys(leftItemByCriteria),
+      keys(rightItemByCriteria)
+  ));
+
+  return map(
+      key => mergeFunction(leftItemByCriteria[key], rightItemByCriteria[key]),
+      uniqueKeys
+  );
+});


CrossEye:
I'd love to see some more real-world justification of this before we proceed to consider tests.  Do you use a function like this often?  Can you show an live-ish example of the need for this function?

I'm mostly bothered by the notion of uniqueness here.  What would you expect from your example, if passed these?:

```js
[{ key: 'a', value: 1 }, { key: 'b', value: 2 }, { key: 'a', value: 42 }],
[{ key: 'b', value: 3 }, { key: 'c', value: 4 }, { key: 'b', value: 101 }]
```

I also have some concern about output order.  If we're returning an array, does the order simply match the order of the first appearance of each "key" in the concatenation of the two inputs?  Is that sufficient?  Or would we need to start considering left- and right-versions of this?

And still one other question: should there be a way to *skip* output values for which there is no match?  It feels as though we might want that.  While the merge function can simply return `undefined` and a follow-up `filter(Boolean)` would cover that case, it feels dirty.  One possibility would be for that function to return an *array* of values to include in the output.  It has a slightly clumsier developer experience but offers quite a bit more flexibility.

We would also need to modify the implementation a bit, as Ramda is, at least until we get past version 1.0, expected to run on older JS engines.  It might still run on ES3 engines, and should definitely run on ES5 ones.  So we couldn't use the null operators `?.` and `??`.  But that's a fairly minor detail.

rodrigoruiz:
One example where I needed it was something like this:

```
const currentAssets = [{
    otherProps: '',
    hash: '1',
    contractID: '1234',
    createdAt: new Date()
}]; // retrieved from database
const newAssets = [{
    otherProps: '',
    hash: '1'
}]; // received from request

const assets = mergeBy(
    R.prop('hash'),
    (currentAsset, newAsset) => ({
        ...(currentAsset ?? {}),
        ...(newAsset ?? {}),
        contractID: currentAsset?.contractID ?? uuid(),
        active: !R.isNil(newAsset),
        createdAt: newAsset ? new Date() : currentAsset.createdAt
    }),
    currentAssets,
    newAssets
);
```

So the scenario is:
- you have same assets in the database
- the user is sending new assets that might create or update existing assets
- the match between current and new assets is based on the `hash` key
- the merged `assets` is what I'd save somewhere replacing the existing assets
- when merging, I get all properties from current and new asset (the new overriding the current)
- the `contractID` must either be the existing one or a new generated one
- `active` is set to `true` if there is a new asset, meaning if the user didn't send any asset for that hash, the current one must be inactivated
- the `createdAt` is updated if there is a new asset, if not, it remains the same

rodrigoruiz:
As for "I also have some concern about output order. If we're returning an array, does the order simply match the order of the first appearance of each "key" in the concatenation of the two inputs? Is that sufficient? Or would we need to start considering left- and right-versions of this?":

I would order them getting all keys from the left list, in the order they are generated from the items, then the same from the right list. But in all honesty, I accept suggestions if you have better ones.

rodrigoruiz:
I'm not sure I understood this: "And still one other question: should there be a way to skip output values for which there is no match? It feels as though we might want that. While the merge function can simply return undefined and a follow-up filter(Boolean) would cover that case, it feels dirty. One possibility would be for that function to return an array of values to include in the output. It has a slightly clumsier developer experience but offers quite a bit more flexibility."

There is not need for a match from the two lists, that's exactly the point.
If one list doesn't contain a corresponding match for a generated key, its corresponding item will just be `undefined`.
In other words, there will always be at least one item (`leftItem` or `rightItem`) in the merge function. There might be both if both lists contains the corresponding key, but there will always be at least one, because the generated key was generated from one of the items from one of the two lists.

rodrigoruiz:
The initial idea was something like this:
- imagine a Venn diagram of 2 circles
- the definition of the intersection is based on the merge criteria
- I want as a result the entire content of both circles
- for each of the 3 parts (left circle out of the intersection, the intersection and right circle out of the intersection) I want to decide how to build the return object
- so if it is only in left circle I want logic A, if it is in both, logic B, and if it is only in the right circle, logic C

CrossEye:
@rodrigoruiz:

First a caveat: Note that it's a big part of my role here to think not only of the requester's needs, but of anyone else's who might use a similar feature.  And of course I also have to consider the size of the library; the longer Ramda is around, the harder it becomes for new functions to make it in.  This is as it should be, but sometimes it's hard to explain why.  So I'm not trying to be argumentative.  I'm trying to come to an understanding of what you're suggesting, how it might be made useful to the widest possible audience, and whether the size of that audience -- or the difficulty of creating the function one one's own -- justifies including it inside Ramda.


> One example where I needed it was something like this: [...]

Thank you.  That's helpful


> I would not expect to receive this: [ ... ] Either way, that was not a scenario where I'd use `mergeBy`.

But remember, in a generic library like Ramda, we have to consider how users might use the tools we supply.  Ramda is pretty allergic to throwing exceptions, as they kill composability.  Taking first or last matches is certainly possible.  So is sending arrays, but then the user experience is degraded for what might be the most common case: zero or one items from the left, zero or one from the right, and always at least one of them.


> I would order them getting all keys from the left list, in the order they are generated from the items, then the same from the right list.

While that's likely the most logical choice, the problem is that we're returning an ordered collection that has no clear intrinsic order.  When we `map` an array, for instance, the order of the result is obvious, the nth element of the output is the result of applying our transformation to the nth element of the input.  We might not keep some interesting properties of the input in the output (`map (square, [-3, -2, -1, 0, 1, 2, 3])` converts an ordered array into one that is not ordered) but we still understand the order.  While we can make the call for ourselves on the output order of `mergeBy`, it feels distressingly arbitrary.  Ramda still does not deal much with `Set`s, especially as it's still ES5-compatible, but would a `Set` be a more logical result, or even a `Map`, keyed off the result of your `mergeCriteria`?


> There is not need for a match from the two lists, that's exactly the point.

But imagine if we include this function, other users who want to only include the result when there are matches.  While they could return `null`/`undefined`, the resulting `[undefined,  { key: 'b', leftValue: 2, rightValue: 3 }, undefined]` could be awkward and unintuitive.  It would be nice if for their case it returned just `[{ key: 'b', leftValue: 2, rightValue: 3 }]`.  I was considering ways to implement this, and one would be if the `mergeFunction` returned an *array* of results.  In your case, we would return a single-element array.  In theirs, they would return single-element arrays when there's a match and empty arrays otherwise, but we could also imagine cases where they return modified versions of *both* inputs, or return both inputs *plus* a combined one.
 

> so if it is only in left circle I want logic A, if it is in both, logic B, and if it is only in the right circle, logic C

This makes me think of another possible implementation which might be cleaner... except that it would require six parameters.  Ramda has [only one function][ra] that takes four parameters and none that take more.  Six seems monstrous.  But I can easily see it being cleaner to separate out the logic for the three categories, and not put it into a single function:

```js
mergeBy (
  prop ('key'),
  (({key, value}) => ({key, leftValue: value, rightValue: 6}),   // left only
  (({key, value: leftValue}, {value: rightValue) => ({key, leftValue, rightValue}), // both left and right
  (({key, value}) => ({key, leftValue: 5: value, rightValue: value}), // right only
  [{ key: 'a', value: 1 }, { key: 'b', value: 2 }],
  [{ key: 'b', value: 3 }, { key: 'c', value: 4 }]
)
```

One other possible extension would be to replace a unary key-generating function with a binary predicate.  Rather than creating a key that we can compare with `===`, we could require the user to specifically say when a left and a right element are identical.  This has some really nice upsides, and one major downside.  One upside is that we no longer have to generate a string or numeric value when we want to compare on three fields.  I have seen such requirements as "If they both have the same `foo` value, the same `bar` value (irrespective of case), and their `timestamps` are within 100 ms of one another", then they are the same."  A second upside is that we no longer have to treat the two sides symmetrically here.  we might write `(left, right) => left._id == right.guid`.  The downside is performance.  It would replace an algorithm that operates in `O (n + m)` time on the size of the two arrays with one that operates in `O (n * m)`.  This could be a huge issue.  But the benefits are real too.

Finally, we need to consider how universal this need is and how difficult it is to write on ones own if Ramda doesn't supply it.  A function expected to be widely used is more likely to make it into the library than one that is specialized.  But one that is difficult to get correct on one's own is also more likely to make it in than one which is trivial.  I think we can see that this is non-trivial to implement, regardless of the possible variation under discussion.  But that's not enough if the audience is too small.  So do you have an argument to show that this has fairly wide application?


  [ra]: https://tinyurl.com/yfbn6pko

rodrigoruiz:
As for the first part of your reply (all but the last paragraph), I agree, we have many options as to how to implement `mergeBy` and I'm willing to discuss further which ones make more sense.

As for the last part, your question, no, I do not know if other people would use this, it just seemed generic enough to be a common use case, or at least something that wouldn't hurt Ramda in general and might bring benefit to those who use it.
For example, I don't understand why Ramda doesn't have the `isNotNil` function, seems common enough for it to exist.
I understand your argument of "how simple it is to implement on its own", but I don't see how another separate function (as in a separate file on Ramda) could hurt it, so I'd add it just for the convenience.

How would you go about it?
That's the first time I try contributing to an open source project, so I don't really know the proper steps.
Should I make a poll or something similar to ask more developers about this functionality?

CrossEye:
@rodrigoruiz:

First of all:

> That's the first time I try contributing to an open source project, so I don't really know the proper steps.

Welcome, and you're doing great!

Every community is different.  Here, most commonly, potential new functions are first raised as an issue and then if there's interest, a pull request is made.  But probably at least a quarter of them come in directly as pull requests.  And that's fine.  The advantage of issue-first is that there's less commitment involved in writing an implementation that might not be accepted or might need to be rewritten if discussion such as ours above make changes to the API.

Also, in some places, contributions are accepted or rejected almost at once.  Here it's always a slower process.  What started as a project in self-education for the two founders has turned into a million-download-per-day behemoth.  We did not set out to maintain a popular library and are often stunned by its success.  This shows in our haphazard processes and our slow deliberations regarding changes.



> Should I make a poll or something similar to ask more developers about this functionality?

Here, I'm just going to ask @ramda/core if any of them have feedback.  But often the best way to show how useful a function might be is to cite a number of uses that you've seen or that you can easily imagine.


> For example, I don't understand why Ramda doesn't have the `isNotNil` function, seems common enough for it to exist.
I understand your argument of "how simple it is to implement on its own", but I don't see how another separate function (as in a separate file on Ramda) could hurt it, so I'd add it just for the convenience.

That actually should make it in the next release.  https://github.com/ramda/ramda/pull/2818

But it's much easier to agree that `isNotNil` is a common need, and when it is suggested, it needs much less discussion to approve of the idea, simply because it's so obviously  useful.  That it's easy to  implement as `complement (isNil)` doesn't outweigh the obvious utility.  `mergeBy` is trickier.  But note that if we do decide that it doesn't belong in Ramda itself, it can always go in the [Cookbook][cb].  And even better, [ramda-adjunct][ra] has long served as a home for most useful well-written functions that don't quite make it into Ramda core.  @char0n is very obliging.  So if we decide it doesn't belong in Ramda itself, it could well still find a useful home.


> I agree, we have many options as to how to implement `mergeBy` and I'm willing to discuss further which ones make more sense.

Well good.  That's how things tend to go here.  Long discussions, sometimes involving long pauses.  But we do get somewhere eventually!



  [cb]: https://github.com/ramda/rama/wiki/Cookbook

CrossEye:
@rodrigoruiz:

Are you still interested in pursuing this?  If so, we need to decide if some of the discussion should make it into the implementation.  We would also need to add tests.

rodrigoruiz:
My apologies for the delay @CrossEye , I lost access to my GitHub account.
I'm sorry, but I currently don't have the time to pursue this, but thank you for your time.

PR #3170, opened 2021-07-28
Move detailed spec to description section

kara-jovian:
The current description is quite confusing, I think it's better to use the comment from the `@return` section as a header because it's clear and complete.

### DIFF
--- a/source/and.js
+++ b/source/and.js
@@ -2,7 +2,7 @@ import _curry2 from './internal/_curry2.js';
 
 
 /**
- * Returns `true` if both arguments are `true`; `false` otherwise.
+ * Returns the first argument if it is falsy, otherwise the second argument.
  *
  * @func
  * @memberOf R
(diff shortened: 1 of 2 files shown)


CrossEye (changes requested):
Welcome!  And thanks for the suggestion.

I agree that this is cleaner.  But perhaps after the 'Returns the first...` statement, we might add something like, "This function serves as the boolean `and` statement if both inputs are booleans."

Also, would you be willing to include a similar fix to the `or` function?

karina-jovian:
>we might add something like, "This function serves as the boolean `and` statement if both inputs are booleans."

I guess it's a good idea, I'll make this amendment a bit later

>Also, would you be willing to include a similar fix to the or function?

Yes. Should I do this in this branch / PR or in a separate one?

kara-jovian:
@CrossEye any updates?

CrossEye (approved):

CrossEye:
Sorry for the delay.  Merged.

PR #3174, opened 2021-08-18
Update README.md

### LINKED ISSUE #3173: ES6 module import failing on Web browsers
I can't import ramda in web browser environment using the ES6 import construct. It looks like ramda is loading dynamically its functions. 
Example:

test.html:

```
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title>Background</title>
</head>
<body>
<script src="test.js" type="module"></script>
</body>
</html>
```

test.js:
```
import * as R from "//cdn.jsdelivr.net/npm/ramda@latest/dist/ramda.min.js";
console.log("R: " + Object.keys(R).length); //R: 0
```
and then serve these two files using a webserver (apache, node, whatever). Launch test.html and you'll see in the web console the message I've written in the above comment

But, if I execute again the last statement in the web console it gives me `R: 257`
 
tested on Firefox and Chrome latest in windows. 

I don't know if it's a bug or I've misunderstood your documentation...

bunglegrind:
Closes  #3173 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,10 @@ Ramda versions > 0.25 don't have a default export.
 So instead of `import R from 'ramda';`, one has to use `import * as R from 'ramda';`
 Or better yet, import only the required functions via `import { functionName } from 'ramda';`
 
+**Note for ES6 module and browsers**
+In order to access to the ES6 module in browsers, one has to provide the content of  the __es__ directory (see below for the build instructions) and use `import * as R from '<path-to-es-directory>/es/index.js';`
+
+
 ### Build
 
 `npm run build` creates `es`, `src` directories and updates both __dist/ramda.js__ and __dist/ramda.min.js__


README.md
```diff
@@ -141,6 +141,10 @@ Ramda versions > 0.25 don't have a default export.
 So instead of `import R from 'ramda';`, one has to use `import * as R from 'ramda';`
 Or better yet, import only the required functions via `import { functionName } from 'ramda';`
 
+**Note for ES6 module and browsers**
+In order to access to the ES6 module in browsers, one has to provide the content of  the __es__ directory (see below for the build instructions) and use `import * as R from '<path-to-es-directory>/es/index.js';`
+
```
CrossEye, line 146: Should that be `path-to-ramda-directory` ?

CrossEye (changes requested):
I'm hoping for more feedback before documenting this.  @Andarist, you're my go-to person on these issues.  Does this seem correct?

So far I haven't used ES6 modules directly in the browser, as most of my recent work is server-side.

README.md
bunglegrind: I'm referring to the es directory created in the build phase

Andarist:
> I'm hoping for more feedback before documenting this. @Andarist, you're my go-to person on these issues. Does this seem correct?

It looks OK - but it exposes a somewhat internal directory to the consumer (which could already be done elsewhere in Ramda's case). Given how the package is configured it might be already correctly resolved without this - when using certain tools. For example if you try to load Ramda using this URL: https://unpkg.com/ramda?module then it will redirect correctly to that internal `es/index.js`. This depends though on how exactly you plan to load the files, using what tools, intermediaries etc

CrossEye:
Thanks, @Andarist!

README.md
CrossEye: This simply doesn't make sense to me:

```
import * as R from '<path-to-es-directory>/es/index.js';`
//                   ^^^^^^^^^^^^^^^^^^^^
```

What es directory are we talking about here?  The `es` directory is inside the Ramda one, right?

```
import * as R from '<path-to-es-directory>/es/index.js';`
//                                         ^^
```

semmel: When using `npm install ramda` and otherwise working build-tool free in the browser, it's exactly as you suspect @CrossEye 

```javascript
import * as R from './node_modules/ramda/es/index.js';
```

bunglegrind: ok done

PR #3177, opened 2021-09-03
Security fix for ReDoS

ready-research:
Security fix for ReDoS vulnerability.

https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String/Trim

Reported in https://www.huntr.dev/bounties/51dfe927-1bf1-4b36-91ca-2197337e663a/

Before fix: Result
`time_cost: 2639`
After fix: Result
`time_cost: 4`

### DIFF
--- a/source/trim.js
+++ b/source/trim.js
@@ -23,9 +23,7 @@ var hasProtoTrim = (typeof String.prototype.trim === 'function');
  */
 var trim = !hasProtoTrim || (ws.trim() || !zeroWidth.trim()) ?
   _curry1(function trim(str) {
-    var beginRx = new RegExp('^[' + ws + '][' + ws + ']*');
-    var endRx = new RegExp('[' + ws + '][' + ws + ']*$');
-    return str.replace(beginRx, '').replace(endRx, '');
+    return str.trim ? str.trim() : str.replace(/^\ws+|\ws+$/g, '');
   }) :
   _curry1(function trim(str) {
     return str.trim();


semmel (changes requested):
- The regular expression seems to be composed wrong. Should rather be in line 26
 
```
str.replace(new RegExp(String.raw`^[\s${ws}]+|[\s${ws}]+$`, "g"), '')
// Of course String.raw` must be converted to ES5 string concatenation and proper \ escaping.
```

- Btw. `String.prototype.trim` is available at least since IE9 and Node.js v0.10.  Since `trim` only supports `String` should it even provide a ponyfill via regular expression? Which non-obsolete JS engine even falls into that RegExp case?
- In line 24 some parentheses seem unnecessary, because logical OR is associative.

ready-research:
@semmel Thanks for the review.

- The regular expression suggested by you is taking too long time as 1900-2000 ms. So can we try this? which will take 3ms

```
str.replace(new RegExp(String.raw`^\ws+|\ws+$`, "g"), '')
```
- Once you confirm this I will change parentheses also in line 24.

semmel:
Sorry `\w` is just a word character in RegExp. `\ws` has got nothing todo with the `ws` variable defined in the beginning  which contains some nice space characters. It matches just any letter followed by an `s`. That's why

```
"asabcbs".replace(new RegExp(String.raw`^\ws+|\ws+$`, "g"), '')
// -> "abc"
```

In which runtime (JavaScript engine) do you execute your test to measure the execution times?

In principle I'd more prefer an implementation like

```
function trim(str){
   return str.trim();
}
```
But that change would need confirmation by the Ramda mainainers and user community.

ready-research:
Can we use this
`return str.trim ? str.trim() : str.replace(/^\s+|\s+$/g, '');`

ready-research:
Closing this in https://github.com/apidoc/apidoc-core/pull/119

semmel:
@ready-research 

> Can we use this
> `return str.trim ? str.trim() : str.replace(/^\s+|\s+$/g, '');`

Should include some line-feed or carriage-return characters, `\s` is not general enough. 
E.g. [this polyfill](https://github.com/FabioVergani/js_String.prototype.trim/blob/master/String.prototype.trim.js) uses a very similar RegExp. I guess `var ws = ...` already includes all sorts of those. 

I am still curious in which browser, or Node.js runtime you measure your execution times.

CrossEye:
This seems to miss all those other white-space characters checked by the original.

https://tinyurl.com/ygwxncp6

Do you have a version that captures those?

ready-research:
@semmel @CrossEye Can you please check this? Using this regular expression it's taking 3-4 ms only to trim. 

```
ws = "[" + ws + "]";
  var beginRx = new RegExp("^" + ws + ws + "*");
  var endRx  = new RegExp(ws + ws + "*$");
  return String(this).replace(beginRx, "").replace(endRx, "");
```

@semmel Thank you for your suggetion, I have reported this in [huntr](https://www.huntr.dev/bounties/51dfe927-1bf1-4b36-91ca-2197337e663a/) you can check the POC file there(Or if you provide me a mail ID I will forward it). I am running this in VS Code with NODEJS 14.17.3

@CrossEye Thank you for spending time to validate in huntr and providing feedback.

CrossEye:
I'm afraid this is still broken.  What is `this` doing in here?  In refactoring a function, the first thing you need to do is ensure it still works.

https://tinyurl.com/ygh54e6k

Can you set up a test that others can run to confirm your performance fix once it's working functionally?

ready-research:
@CrossEye Thanks for the test case, it helped me to understand and test. I am not familiar with coding that much. Anyway, the following test case working fine with white-space characters and as well as performance. Please check this 

https://tinyurl.com/yzu7dkpy

CrossEye:
I'm curious what you're doing creating pull requests for an open source project if you are "not familiar with coding that much."  Are there some sort of bounties offered to fix such issues?  I know the Ramda team does not offer them itself, but perhaps some other interested party does?

As to the performance benefits, I would appreciate it if you could demonstrate it somehow.  Your assertion is not enough evidence.  Do you have a test case we can run in Node or in a browser to see the benefits of this new version over Ramda's, which defaults to `String.prototype.trim` in most environments.

ready-research:
@CrossEye I am a security enthusiast(eager to learn to code for security stuff), I think open source will be a better platform to do so. Yeah, [huntr](https://www.huntr.dev/bounties/51dfe927-1bf1-4b36-91ca-2197337e663a/) offers bounties for security issues and fixes.

You can test the following using node
Create the following poc.js
```
// PoC.js
var {trim} = require("ramda");
function build_blank (n) {
var ret = "1"
for (var i = 0; i < n; i++) {
ret += " "
}
return ret + "1";
}
var time = Date.now();
trim(build_blank(50000))
var time_cost = Date.now() - time;
console.log("time_cost: " + time_cost)
```
Execute the following command in another terminal:
```
npm i ramda
node poc.js
```
Check the Output:
```
time_cost: 2639
```

Ideally, trim functionality should give the output in 3-10 milliseconds, but it's taking 2000-3000 milliseconds.

The patch provided above will give us in 4-6 milliseconds.

CrossEye:
Well, there is definitely something wrong in Ramda -- when I repeat the implementation locally, it works fine:

Here is one version of your test case showing Ramda alongside your implementation and what looks like it should be the same as Ramda's implementation.  Ramda is horribly slow.  The other two are fine.

https://runkit.com/crosseye/ramda-issue-3177

And that, by the way, is one good way to share test cases.  RunKit and a slew of other sites let you set up some scripts with imports for your libraries and run them on a server.

I will try to look carefully in the next few days to see why Ramda's export is slow while the others are fine.  If you have any ideas, please let me know.

semmel:
The culprit is the `ws.trim()` in line 24. It evaluates to a string of length 1 `\x180e` in V8, the Safari browser and Firefox JS engine. Thus Ramda resorts *always* to the first implementation with the chained RegEx replacements which is horribly slow. 

I don't know in what language `\x180E` is a whitespace character to justify its inclusion in the `ws` string, but all JS engine vendors seem to think differently.

Removing that character from `ws` in line 4 out fixes the performance problem. Except of course for InternetExplorer 8 or Node.js < v0.1 - R.I.P. And except for an exotic language where `\x180E` is a thing.

CrossEye:
@semmel: Thanks.  I got online to look at a few Ramda issues.  I don't need to do this one now!

Would you simply suggest removing that character, or should we do something more radical?  Once again I'm finding a little time for Ramda, and am looking at how to get to and past v1.0.  After 1.0, this should just be a pointer to `String.prototype.trim`, so anything we do should be seen as a temporary patch.

semmel:
@ready-research Could you check that simply removing `\u180E` from line 4 *in the original source* solves the problem?

```javascript
var ws = '\x09\x0A\x0B\x0C\x0D\x20\xA0\u1680\u2000\u2001\u2002\u2003' +
         '\u2004\u2005\u2006\u2007\u2008\u2009\u200A\u202F\u205F\u3000\u2028' +
         '\u2029\uFEFF';
```

As @CrossEye suggested it's just a preliminary fix, so no need to invent new regular expressions here or any other niceties as removing parentheses etc.

Otherwise I'd scrap together a PR, however I'd see it should include a test for the vulnerability which may require more of the time available to me. Who gets the bounty then? 

Anyway we should get rid quickly of that nasty bug without breaking anything.

ready-research:
@semmel Yes! Simply removing `\u180E` from line 4 in the original source solving this problem. Feel free to edit this PR if you have permissions or you can raise a new PR.

You can submit a patch in huntr and the maintainer have to select your patch as fixing this issue. So you will get the bounty.

CrossEye:
@ready-research:

Would you be willing to update the PR with that change?

semmel:
lgtm

semmel (approved):

CrossEye:
Thank you @ready-research.  This has been committed, and I believe [hunter.dev][hd] should award the bounty.


  [hd]: https://www.huntr.dev/bounties/51dfe927-1bf1-4b36-91ca-2197337e663a/

yahanvesh:
When can we expect a new version to be released with the fix. Much Appreciated.

aw3218:
> We're long overdue for a release. I'll try to get one out soon.

When is 'soon?' Thanks!

CrossEye:
@aw3218:

> When is 'soon?'

I'm trying to work through another reported vulnerability, #3192, which is much more elusive than this one.  As soon as that is figured out, I'll try to get all the accumulated changes into a new release.

yaverin:
@CrossEye , Hi!
As issue #3192 is closed, could you release the new version?

simon-tannai:
Some news about a new release after this vulnerability fix ?

customcommander:
@simon-tannai See https://github.com/ramda/ramda/issues/3195#issuecomment-998018529

customcommander:
Released in `v0.27.2`. See #3216 

@CrossEye I ran this [RunKit](https://runkit.com/crosseye/ramda-issue-3177) of yours against `v0.27.2` and the numbers for Ramda look more sensible now :)

Screenshot 2022-01-11 at 15 10 36

CrossEye:
Very nice!

CrossEye:
@opravil-jan:

This was fixed as of version `0.27.2`.

CrossEye:
Can you share a link, an actual CVE, or something we can investigate.  An image doesn't help much, since I don't have a Prisma Cloud account.

Snyk reports this as resolved: https://security.snyk.io/vuln/SNYK-JS-RAMDA-1582370

But there was some strangeness involve in 0.27.2 and 0.28.0, so maybe something weird is happening.

PR #3178, opened 2021-09-05
allPass and anyPass

xp44mm:
`allPass` and `anyPass` return the result of the final expression. action same like to `both` and `either`

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,12 @@
+# Security Policy
+
+## Supported Versions
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 0.27.*  | :white_check_mark: |
+| < 0.27  | :x:                |
+
+## Reporting a Vulnerability
+
+Feel free to <a href="mailto:scott@sauyet.com">email Scott Sauyet</a> with any security issues


CrossEye:
Thank you for the contribution.

I'd like to see your argument in favor of this.

Obviously these functions can be thought of as generalizations of `both` and `either`, but that's not how we thought of them when creating them, and it's certainly not how we documented them:

> Takes a list of predicates and returns a predicate that returns true for a given list of arguments if every one of the provided predicates is satisfied by those arguments.
> 
> ```
> @sig [(*... -> Boolean)] -> (*... -> Boolean)
> 

So we certainly can't change the implementation without also addressing the signature and description.  But is there a good reason to change these at all?

Another way to think about it is that these should be pretty close equivalences:

```js
and (x, y) //~=> x && y
both (f, g) //~=> (...args) => f (...args) && g (...args)
all (fn) //~=> (xs) => xs .every (x => fn (xs))
allPass ([fns]) //~=> (...args) => fns .every (f => f (...args))
```
The [`&&`][la] ones, `and` and `both` act like the JS operator, and return booleans if supplied booleans but otherwise return one of their operands.

The [`every`][ev] ones, `all` and `allPass` act like the JS Array prototype function, and always return booleans.

So I think we'd need a strong argument for this.  Do you have one?

  [ev]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every
  [la]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND

xp44mm:
A javascript function should return as much useful information as possible. The truthy set contains true, and the falsy set contains false. So they is compatible with original functions and has greater use.

xp44mm:
This PR is temporarily closed due to branch management issues.

CrossEye:
> A javascript function should return as much useful information as possible. The truthy set contains true, and the falsy set contains false. So they is compatible with original functions and has greater use.

This also has to be balanced against simplicity.

> Q: What does the function `allPass` return?
>
> A (current): `true` or `false`
>
> A (suggested): Any value, really.  It would be the first `0`, `false`, `''`, `NaN`, `null` or `undefined` returned by one of the functions supplied or value returned by the final function, which could be any value whatsoever.

Ramda always strives to strike a balance between being as simple as possible and being as familiar to JS Developers as possible.  The functions `and` and `both` have real tensions between simplicity and familiarity.  But here, thinking of this function as a variant of `Array.prototype.every`, both simplicity and familiarity suggest the current behavior.

So I'm certainly willing to entertain further discussion, but my initial impression is negative.

PR #3179, opened 2021-09-05
Update cond.js

xp44mm:
Attach an argument to the transformer function to pass the result value of the predicate. 
```js
transformer(...arguments, predicate(...arguments))
```

provide a special case:
```js
transformer = (..., predicateResult) => predicateResult
```
so the pair `[predicate,transformer]` array reduce to `predicate` function.

### DIFF
--- a/source/cond.js
+++ b/source/cond.js
@@ -55,7 +55,7 @@ var cond = _curry1(function cond(pairs) {
         let predicate = pair[0].apply(this, arguments)
         if (predicate) {
           let args = Array.from(arguments)
-          args.push(predicate) // transformer(...arguments, predicate)
+          args.push(predicate) // transformer(...arguments, predicate(...arguments))
           return pair[1].apply(this, args);
         }
       }
--- a/source/cond.js
+++ b/source/cond.js
@@ -40,16 +40,24 @@ var cond = _curry1(function cond(pairs) {
   var arity = reduce(
     max,
     0,
-    map(function(pair) { return pair[0].length; }, pairs)
+    map(function (pair) { return pair[0].length; }, pairs)
   );
-  return _arity(arity, function() {
+  return _arity(arity, function () {
     var idx = 0;
     while (idx < pairs.length) {
-      let predicate = pairs[idx][0].apply(this, arguments)
-      if (predicate) {
-        let args = Array.from(arguments)
-        args.push(predicate) // transformer(...arguments, predicate)
-        return pairs[idx][1].apply(this, args);
+      let pair = pairs[idx]
+      if (typeof pair === 'function') {
+        let transformer = pair.apply(this, arguments)
+        if (transformer) {
+          return transformer;
+        }
+      } else if (Array.isArray(pair)) {
+        let predicate = pair[0].apply(this, arguments)
+        if (predicate) {
+          let args = Array.from(arguments)
+          args.push(predicate) // transformer(...arguments, predicate)
+          return pair[1].apply(this, args);
+        }
       }
       idx += 1;
     }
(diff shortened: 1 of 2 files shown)


CrossEye (changes requested):
Please read our [Contributing][co] document.  It probably needs an update, but it's a start.

  - There are no associated tests for the changes to the `cond` function.
  - There are formatting changes differing from usual Ramda conventions
  - It uses es6 methods, which Ramda has not adopted and will not adopt until afterrsion 1.0
  - There is no explanation or example

That last point is the most important.  *Why* would we want to do this?  What problem does it solve?  It seems to make `cond` a much more complex function.  What are the benefits that outweigh the complexity?

  [co]: https://github.com/ramda/ramda/blob/master/CONTRIBUTING.md

xp44mm:
I've passed the existing tests, added tests for new features, and passed them.

CrossEye:
I think we need much more of an explanation of *why*.  I can read the code and see what it does.  But I can't get into your head to understand your rationale.

Can you add to the discussion here, explaining what these changes are for, and why in some real world example we would want this and want to expand `cond` to cover it?

xp44mm:
The branch of cond is the predicate first, and the tranformer is performed if the predicate is true. 

Moreover, We also often encounter when part of the tranformer execution has even come to fruition to know if the action should be performed. That is, predicate and transformer cannot be separated from each other.

The latter's signature is similar to: `input -> Some result | None`. Common functions includes: `parseInt`, `parseFloat`, `tryXXX`.

CrossEye:
I would suggest that if you are looking for the behavior of returning the first non-nil result when applying our arguments to a list of functions, then you should probably suggest a new function.  Ramda strives for API simplicity and would never be apt to deal with such different signatures for the same function.

Although this implementation would not be suitable for current Ramda, the idea would be something like this:

```js
const firstTruthy = (fns) => (...args) => {
  for (let fn of fns) {
    const res = fn (...args)
    if (res) {return res}
  }
}

const myFn = firstTruthy ([
  x => x > 20 && x * x,
  x => x % 2 && (x - 1) / 2,
  x => 10 * x
])

myFn (25) //=> 25 * 25 ==> 625
myFn (15) //=> (7 - 1) / 2 ==> 3
myFn (14) //=> 10 * 14 ==> 140
```

We would still need to see some realistic reasoning for this.  I'm not yet convinced of the usefulness of it, but I'm willing to be convinced if you have a compelling argument.

xp44mm:
The `firstTruthy` may have missed a situation:
```js
    let fmt = cond([
      [trim, (_, res) => ({ res })],
    ]);
```
the transformer will accept the returned value of the predicate as appended last argument.

xp44mm:
and this code will terminate at the first branch, skip the second branch:
```js
    let fmt = cond([
      [T, always(false)],
      [T, always('never')],
    ]);
```

CrossEye:
`firstTruthy` was entirely off the cuff.  I'm still trying to understand what your function is *for*.  I simply don't have a real sense of how you'd want to use this.  I'm quite sure that it's not likely a match for `cond`, which has a fairly simple signature and fairly understandable use.  But it's possible that it's something worth considering for Ramda under a different name.  But we seem to be having communication issues.

Do you have a real-world problem that you could simplify for an example?

xp44mm:
The extended function is similar to pattern matching in F#, instead of `if..else` string. It is still compatible with the latter. However, it provides features similar to active pattern in F#. Here's the pseudocode:

```F#
match inp with
| _ when predicate(inp) as res -> transformer(inp,res)
| TryFloat (floatResult) -> floatResult
```

`tryFloat` is the js build-in function when sucess return a number or failed return null. transpile to js:

```js
cond([
    [predicate, transformer],
    tryFloat,
])(inp)
```

CrossEye:
None of this looks anything like a real-world example.  Can you describe a use in something like a web application or a data transformation one.  The business domain is not important, just something concrete that shows *why* you want this.

customcommander:
@CrossEye Shall we close this one for now?

PR #3181, opened 2021-09-09
add new fold.js

xp44mm:
an aux function for recursive function. See code comments, and the test file for details.

### DIFF
--- a/source/fold.js
+++ b/source/fold.js
@@ -1,0 +1,33 @@
+import _curry3 from './internal/_curry3.js';
+
+/**
+ * This function calls recursively the iterator function using the followed accumulator and seed. Each call to the iterator function returns a tuple that consists of next accumulator and next seed. When the iterator function returns `null`, or when no seed is provided, or when the seed provided is `null`, the function returns the last accumulator to the caller.
+ * 
+ * This function mimics the recursive function that accepts an accumulator parameter in functional programming.
+ * 
+ * @func
+ * @memberOf R
+ * @since v0.10.0
+ * @category Function
+ * @sig ((a,b) -> [a,b]|null) -> a -> b -> a
+ * @param {Function} iterator The iterator function. receives two argument, `acc` is `seed`, and returns
+ *        either false to quit iteration or an array of length two to proceed. The element
+ *        at index 0 of this array is next accumulator, and the element at index 1 is next seed.
+ * @param {a} acc The result value.
+ * @param {b} seed The seed value.
+ * @return {a} return the last acc.
+ * @example
+ * 
+ * @symb 
+ */
+export const fold = _curry3(function (iterator, acc, seed) {
+    while (seed != null) {
+        let pair = iterator(acc, seed)
+        if (pair == null || pair.length === 0) break
+        acc = pair[0]
+        if (pair.length === 1) break
+        seed = pair[1]
+    }
+    return acc
+
+})
--- a/test/fold.test.js
+++ b/test/fold.test.js
@@ -1,0 +1,51 @@
+import { fold } from './fold.js'
+
+// addWithMaxOf10 function in functional program
+const loop = (acc,seed) => {
+    if (seed.length > 0) {
+        let nextAcc = acc + seed[0]
+        let nextSeed = seed.slice(1)
+        return nextAcc > 10 ?
+            acc                       // * difference with iterator.
+            : loop(nextAcc, nextSeed) // *
+    }
+    return null
+}
+
+describe('fold', function () {
+    let source = [1, 2, 3, 4, 5, 6, 7]
+
+    it('before last', function () {
+        var addWithMaxOf10 = function (acc, seed) {
+            if (seed.length > 0) {
+                let nextAcc = acc + seed[0]
+                let nextSeed = seed.slice(1)
+                return nextAcc > 10 ?
+                    [] // last acc is `acc` to return caller
+                    : [nextAcc, nextSeed]
+            }
+            return null
+        };
+
+        let y = fold(addWithMaxOf10, 0, source)
+        expect(y).toEqual(10)
+    });
+
+    it('include last', function () {
+        var addWithMaxOf10 = function (acc, seed) {
+            if (seed.length > 0) {
+                let nextAcc = acc + seed[0]
+                let nextSeed = seed.slice(1)
+                return nextAcc > 10 ? 
+                [nextAcc] // last acc is `nextAcc` to return caller
+                : [nextAcc, nextSeed]
+            }
+            return null
+        };
+
+        let y = fold(addWithMaxOf10, 0, source)
+        expect(y).toEqual(15)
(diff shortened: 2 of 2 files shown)


CrossEye:
Please, when making PRs here, include some relevant explanation in simple text.

Yes, we will look at the code.  But that mostly tells us *how* it works.  We also need to know *why* it's wanted.

CrossEye:
I'm curious if [this API][rr] would satisfy your needs:

```js
const None = Symbol()

const fold = (fn, init, [x = None, ...xs]) =>
  x == None
    ? init
    : fn (init, x, (y) => fold (fn, y, xs), (y) => y)

console .clear ()

console .log (
  'Works like a standard fold:',
  fold ((a, b, next, done) => next (a + b), 0, [2, 3,4, 5, 6]) 
) //=> 20 (2 + 3 + 4 + 5 + 6)


const loop = (acc, x, next, done) => 
  acc + x > 10 
    ? done (acc) 
    : next (acc + x)

console .log (
  'Can end early:',
  fold (loop, 0, [2, 3, 4, 5, 6])
) //=> 9 (2 + 3 + 4)

```

I think it's quite a bit simpler to use.  And I really wish Ramda's `unfold` was written in a similar way.  I would of course curry the function before including it in Ramda.

It is strictly less powerful than yours as yours offers the user the ability to control the seed values, where this one simply iterates them.  But it is closer to `reduce`, adding just the ability to end early.

  [rr]: https://tinyurl.com/ygjh9fzk

xp44mm:
terminate conditional place in accumulator over place in seed. because accumulator will be returned to caller, and seed will be given up. 

your implementation seem to use recursive function instead of loop. it maybe stack overflow in action.  to avoid stack overflow is functional lib's one of target.

Seeds include more than arrays, and any other data type is likely to be included. for example,  number, string, Tree among others.

the parameter order is accumulator first, or seed first. there is both. don't matter.

CrossEye:
Yes, my function is using a recursive implementation.  We should be able to easily change that to an iterative one.  It might require some sort of trampoline.

The question is about the API, though.  Does this meet your needs?

But again, I really would like some sort of real-world example of your expected usage.

xp44mm:
this api limit seed as `Array`. not very genenic.

CrossEye:
So, can you show a real-world example with a different seed type?

xp44mm:
100/3=?
```js
let y = fold((x,y) => {
    let diff = y - 3
    return diff > 0 ? [x+1, diff] : [x]
} , 0, 100)
```

CrossEye:
Thanks for the example.  It's coming clearer to me now.  I like the idea of this function, but I don't think I like the return-a-pair-or-null.  I'd prefer a variant of what I suggested above. Here is [a version][rr] of the same idea, but leaving the accumulator changes to the user.  It is still using a recursive implementation which would have be changed to make it even reasonably performant and resilient to large data, but I'm still wondering if this API seems reasonable:

```js
// fold :: ((b, a, ((a, b) -> a), (a -> a)) -> a) -> a -> b -> a
// or this?
// fold :: (b -> a -> (a -> b -> a) -> (a -> a) -> a) -> a -> b -> a
const fold = (fn, init, seed) =>
   fn (seed, init, (x, s) => fold (fn, x, s), (x) => x)

console .clear ()

// Works somewhat like a standard fold
fold (
  ([a, ...as], x, next, done) => a == undefined ? done (x) : next (a + x, as), 
  0, 
  [2, 3, 4, 5, 6]
) //=> 20 (2 + 3 + 4 + 5 + 6)


// Offers an early escape
const loop = ([a, ...as], x, next, done) => 
  a == undefined ? x : a + x > 10 ? done (x) : next (a + x, as)

fold (loop, 0, [2, 3, 4, 5, 6])  //=> 9 (2 + 3 + 4)

// Folds objects besides arrays
fold ((a, x, next, done) => a - 3 > 0 ? next (x + 1, a - 3) : done (x), 0, 100) //=> 33
```

I'm not certain of the signature for this yet, but it strikes me as simpler in the end than `((a,b) -> [a,b] | null) -> a -> b -> a` where we have a promiscuous callback function, and a sentinel value of `null`, both things we try hard to avoid in Ramda.

  [rr]: https://tinyurl.com/yem8e63f

CrossEye:
BTW, here's an [iterative implementation][rr] of the same API:

```js
// fold :: ((b, a, ((a, b) -> a), (a -> a)) -> a) -> a -> b -> a
const fold = (fn, init, seed) => {
  let done = false, acc = seed, curr = init, result
  while (!done) {
    fn (
      acc, 
      curr, 
      (val, seed) => {curr = val; acc = seed}, 
      (x) => {done = true, result = x}
    )
  }
  return result
}
```

While it's not nearly as elegant as the recursive version, it's not as bad as I feared.


  [rr]:https://tinyurl.com/yzp3qwmk

xp44mm:
Using function parameters as placeholders is the first time I've seen them. It's really interesting. It should be feasible and more semantic than tuples. 

It is simply recommended that the function first and the data last. immutable first, variable last.

CrossEye:
> Using function parameters as placeholders is the first time I've seen them. It's really interesting. It should be feasible and more semantic than tuples.

I first saw this technique in a [StackOverflow answer][so] (by a user who has many great informative answers, and is well worth following.)  I've used it occasionally since and it can simplify a lot of nasty control flow issues.


> It is simply recommended that the function first and the data last. immutable first, variable last.

I believe that's what I've done in the signature of the function; the signature of the callback is perhaps more questionable, but it's not clear to me what would be an improvement upon it.

  [so]: https://stackoverflow.com/a/49543489/1243641

xp44mm:
It is recommended that you modify the signature of `fn`:
```js
const fold = (fn, init, seed) => {
  let done = false, acc = seed, curr = init, result
  while (!done) {
    fn (
      (val, seed) => {curr = val; acc = seed}, 
      (x) => {done = true, result = x}
      acc, 
      curr, 
    )
  }
  return result
}
```

CrossEye:
I don't have a strong feeling for or against that change.  I usually try to configure a function with the parameters less likely to change coming before those more likely to change.  But that's mostly to support currying/partial application.  Here there is no reasonable possibility of currying.

I guess my original order feels slightly more idiomatic in JS than your reformulation -- it looks *a bit* more like `Array.prototype.reduce`.  But it's a minor issue, and I don't have strong feelings either way.

xp44mm:
`next`, `done` is readonly, never changed, just for user to place their code. so should first.

CrossEye:
@xp44mm: Are you interested in pursuing this?

xp44mm:
What should I do?

CrossEye:
Review the discussion above.  If you still would like to include this function, either update the implementation to take into account the comments made or close this one and create a new one on that model.  There are no guarantees it would be included, but it will get a fair hearing.  Or if you're not interested, you could just close the issue.

PR #3183, opened 2021-09-13
doc: memoizeWith type requires a string

trajano:
The typing for memoizeWith requires the identity function to return a string.  This updates the example with the correct method

### DIFF
--- a/source/memoizeWith.js
+++ b/source/memoizeWith.js
@@ -24,7 +24,7 @@ import _has from './internal/_has.js';
  * @example
  *
  *      let count = 0;
- *      const factorial = R.memoizeWith(Number, n => {
+ *      const factorial = R.memoizeWith(R.toString, n => {
  *        count += 1;
  *        return R.product(R.range(1, n + 1));
  *      });


CrossEye (commented):
Thank you @trajano for your contribution!

I wonder if we should consider changing the signature instead.  Because `Number` should work, as should anything for which `value.toString()` yields a unique result across your function domain.

Perhaps we can use `Key` instead of `String` with some definition of `Key` such as `String | Number | Symbol`.  What do you think?

trajano:
I don't mind, I just didn't want to touch the API/code directly.  Though personally I would rather Ramda be automatically typed no `@types/ramda` ideally.

CrossEye:
> I would rather Ramda be automatically typed no `@types/ramda` ideally.

If you're volunteering to help, we'd love it!  :smile:

The core team has little TypeScript expertise, and although we've tried to bring the typing in-house, we've made little progress.  Any help or advise you can give on this would be gratefully received!

source/memoizeWith.js
```diff
@@ -24,7 +24,7 @@ import _has from './internal/_has.js';
  * @example
  *
```
CrossEye, line 25: Let's change the signature instead to be more accurate.

CrossEye (changes requested):

CrossEye:
@trajano: Are you interested in pursuing this?

PR #3188, opened 2021-09-23
Fix typos detected by codespell

tmatth:
For reference:
https://github.com/codespell-project/codespell

### DIFF
--- a/source/internal/_equals.js
+++ b/source/internal/_equals.js
@@ -12,7 +12,7 @@ import type from '../type.js';
  * - iterators lengths are the same
  * - iterators values are unique
  *
- * false-positive result will be returned for comparision of, e.g.
+ * false-positive result will be returned for comparison of, e.g.
  * - [1,2,3] and [1,2,3,4]
  * - [1,1,1] and [1,2,3]
  * */
--- a/test/collectBy.js
+++ b/test/collectBy.js
@@ -12,7 +12,7 @@ describe('collectBy', function() {
     }));
   });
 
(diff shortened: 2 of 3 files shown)


dist/ramda.js
```diff
@@ -2184,7 +2184,7 @@
    * implementation.
    *
    * @private
-   * @param {Function} fn ramda implemtation
+   * @param {Function} fn ramda implementation
```
CrossEye, line 2187: This file is generated.  Could you do the same in `source/internal/_checkForMethod,js`?

CrossEye (changes requested):
Thank you.  This looks great except for the one place the tries to fix the generated file.

dist/ramda.js
tmatth: @CrossEye not seeing that spelling there:
https://github.com/ramda/ramda/blob/master/source/internal/_checkForMethod.js

CrossEye: Hmm, well it should be fixed on the next release then.  But we prefer to never check in this generated file.

tmatth: @CrossEye yeah normally I would've left it alone but since it had diverged and there are instances where it was committed in between releases (see e.g., f508b04ee0ecac924eb41ca4d9ceadc13b8a3b14) I included it. In any case I'll drop it.

CrossEye (approved):

CrossEye:
Thank you @tmatth!

PR #3196, opened 2021-11-06
R.is(Object, Object.create(null)) should be true

wiowou:
I think the expectation when using R.is(Object) is that an instance of Object.create(null) should qualify as an object. This was not the case before.


.vscode/settings.json
```diff
@@ -0,0 +1 @@
+{}
```
CrossEye, line 1: Please don't commit this file.  Or `dist/ramda.js`.  Or `dist/ramda.min.js`.

CrossEye (changes requested):
I *think* this makes sense.  But I want to consider it a bit.  

We would have to change the documentation, which says

> See if an object (val) is an instance of the supplied constructor. This function will check up the inheritance chain, if any.

But I agree that I would expect `is (Object) (Object .create (null))` to return `true`.

----------

Note that there are files included here that we don't commit.

.vscode/settings.json
wiowou: My sincere apologies. I believe I have remedied the problem - please do let me know if I have not done so correctly, particularly with the files in `dist/`.

wiowou:
I understand, please do mull it over. I would greatly appreciate if you could share your final thoughts either way.

.vscode/settings.json
CrossEye: No problem at all.  We really need to improve our process.  However, I think you didn't get it right.  Now the commit is deleting the files.  We probably shouldn't have that.  We want these in the release but not every commit.  After we get this fix in, I'm planning on making a new release, and I'll try to fix things so that these two files are in a .gitignore.  I'm not quite sure why they are not there now.

CrossEye:
I do agree that we want this.  My hesitation had to do with imagining scenarios where a user type shadowed the Object constructor (perhaps a `SentenceDiagram` tool with functions like `Subject` and `Object`) and we now are using a constructor with the name `'Object'` for something different.  I actually started trying to write up an example when I realized that these would *already* be objects, and the new code wouldn't change anything.  D'Oh!

I would like to fix the code layout.  It was already hard to read and adding new clauses like this makes it harder still.  And while we're at it, we can reduce the duplication.  Perhaps [this version][rr] would do:

```js
const is = function is(Ctor, val) {
  return val instanceof Ctor ||
         val != null && (
            val.constructor === Ctor || 
            (Ctor.name === 'Object' && typeof val === 'object')
         )
}
```

(with the currying restored, of course.)

We still need to address the documentation.  Do you want to take a swing at that or should I?  I think an extra sentence explaining an additional `true` result would probably be enough.


  [rr]: https://tinyurl.com/yhcmlaea

.vscode/settings.json
wiowou: I'm afraid I might have mucked this branch so that I'm unable to fix the problem with `dist/`. I have created a new branch `is-bugfix-3196`. I have confirmed that the folder is correctly represented there. Perhaps it would be best to initiate the pull request from that branch?

wiowou:
Thank you, that is certainly clearer and better! I have taken a first go at the docs. It took me a second to realize that the docs were directly in `is.js`. I have created a new branch `bugfix-is-3196` that includes these changes and fixes the problem with `dist/` though the changes to `is.js` are also represented in this branch.

CrossEye (approved):

CrossEye:
This looks good to me.  Any objections to this, @ramda/core?

hitmands (approved):

source/is.js
```diff
@@ -3,7 +3,8 @@ import _curry2 from './internal/_curry2.js';
 
 /**
  * See if an object (`val`) is an instance of the supplied constructor. This
- * function will check up the inheritance chain, if any.
+ * function will check up the inheritance chain, if any. 
+ * If (`val`) was created using `Object.create`, `R.is(Object, val) === true`.
```
customcommander, line 7: Nit: are the parentheses necessary here?

```none
If `val`…
```

vs

```none
If (`val`)…
```

customcommander (approved):

source/is.js
CrossEye: Good catch -- probably not.

CrossEye: @wiowou: If you can fix this nit, I think we're ready to merge.

wiowou: Yes, certainly.

hitmands: Probably because it was already present in `L5` https://github.com/ramda/ramda/blob/master/source/is.js#L5

wiowou: Yes, it was. However you prefer.

customcommander: I think in this case the parentheses indicated `val` as related to "object". Similar to:

```
See if an object (i.e. `val`) is an instance of the supplied constructor.
```

But here I think you probably meant just this: (wrapping `val` with parentheses didn't seem necessary to me)

```
If `val` was created using `Object.create`, `R.is(Object, val) === true`.
```

I didn't intend to block this PR. I'm happy to change this later.

wiowou: Quite right. Thank you for giving me a chance to correct this.

wiowou:
Please look to PR 3198. I had problems removing the `dist` directory from version control and found it easier to simply create another branch.

PR #3199, opened 2021-11-23
chore: no need to restrict whenTrue and whenFalse's return type as input type

adispring:
There's no need to restrict `whenTrueFn`'s return type to be consistent to `R.when`'s input type.

### DIFF
--- a/source/unless.js
+++ b/source/unless.js
@@ -11,7 +11,7 @@ import _curry3 from './internal/_curry3.js';
  * @memberOf R
  * @since v0.18.0
  * @category Logic
- * @sig (a -> Boolean) -> (a -> a) -> a -> a
+ * @sig (a -> Boolean) -> (a -> b) -> a -> a | b
  * @param {Function} pred        A predicate function
  * @param {Function} whenFalseFn A function to invoke when the `pred` evaluates
  *                               to a falsy value.
--- a/source/when.js
+++ b/source/when.js
@@ -11,7 +11,7 @@ import _curry3 from './internal/_curry3.js';
  * @memberOf R
  * @since v0.18.0
  * @category Logic
- * @sig (a -> Boolean) -> (a -> a) -> a -> a
+ * @sig (a -> Boolean) -> (a -> b) -> a -> a | b
  * @param {Function} pred       A predicate function
  * @param {Function} whenTrueFn A function to invoke when the `condition`
  *                              evaluates to a truthy value.


source/unless.js
```diff
@@ -11,7 +11,7 @@ import _curry3 from './internal/_curry3.js';
  * @memberOf R
  * @since v0.18.0
  * @category Logic
- * @sig (a -> Boolean) -> (a -> a) -> a -> a
+ * @sig (a -> Boolean) -> (a -> b) -> a -> a | b
```
customcommander, line 14: I'm not too well versed in HM notation so this may be a stupid question:

I think it makes sense to emphasise that `whenFalseFn` and `whenTrueFn` both don't need to return a value of the same type at their input. So changing `a -> a` to `a -> b` (where `b` may or may not be the same type as `a`) makes sense to me.

If `b` can indeed be of the same type as `a` then what's the `a | b` for? I would have expected this instead:

```none
(a -> Boolean) -> (a -> b) -> a -> b
```

What do you think?

adispring: In `R.when`, If the input value satisfies the predication, `R.when` will execute `whenTrueFn`, and return the result of `whenTrueFn`, otherwise, `R.when` will just return the input value, as follows:

```js
const formatDateIfNotNil = R.unless(isNil, time => moment(time).format('YYYY-MM-DD'));

let time;
formatDateIfNotNil(time); //=> undefined

time = 1635004800000
formatDateIfNotNil(time); //=> '2021-10-24'
```

So if the input value's type is `a`, and the return type of `whenTrueFn` is `b`, `R.when`'s return type will be either `b` or `a`. 

The return type of `R.unless` is the same as `R.when`.

customcommander: Oh yeah I forgot that about both `when` and `unless`. Ok that makes more sense. Thanks

customcommander (approved):

customcommander:
@adispring LGTM but you should get another approval from someone better than me at HM notations ;) Your reasoning did make sense to me though.

adispring:
@customcommander Ok, Thank you for your review.

CrossEye:
I think I disagree.  Our signatures are aspirational.  They don't try to capture the ugliness of JS functions.  I'd rather we capture the expectation of how the function should be used rather than the result if you use it incorrectly.

The few places where we have `a | b` types already feel dirty to me; I'd rather not add another.

I'd prefer to go in the other direction, and switch `ifElse` from

```js
// ifElse :: (*… → Boolean) → (*… → *) → (*… → *) → (*… → *)
```

```js
// ifElse :: (*… → Boolean) → (*… → a) → (*… → a) → (*… → a)
```

or even to move the ellipsis out of the functions and add it after some specific examples:

```js
// ifElse :: (a → Boolean) → (a -> b) → (a -> b) → (a -> b)
//           (a → b → Boolean) → (a → b → c) -> (a → b → c) → (a → b → c)
//           (a → b → c → Boolean) → (a → b → c → d) -> (a → b → c → d) → (a → b → c → d)
//           ...
```

adispring:
@CrossEye In practical, If `R.when(pred, onTrueFn)`'s output type is consistent to input type, this will make it less usefull. In Ramda's documentation, `R.unless`'s example is as follows:

```js
let safeInc = R.unless(R.isNil, R.inc);
safeInc(null); //=> null
safeInc(1); //=> 2
```

and in the unit test of `R.when` and `R.unless`, the return type is also not consistent to the input type:

https://github.com/ramda/ramda/blob/master/test/when.js
https://github.com/ramda/ramda/blob/master/test/unless.js

semmel:
@adispring From a *technical* point of view you are right that *in principle* the return type is `a|b`. But in that case I find `when` and `unless` not suitable for function composition. They *can* but *should not* be used in that way.

CrossEye:
@semmel: Between my deciding that this was a correct change and hitting merge, your comment came through.

I still think that this is not a particularly useful feature, but as we've documented it that way already, it's probably best if the signatures match.

Issue #1683, opened 2016-03-11
uniqBy transducer?


hden:
Allows `uniq` and `uniqBy` to return a stateful transducer when no collection is provided.

Usage:

```
R.into([], R.compose(R.uniq(), R.take(10)), coll)
```

see: https://clojuredocs.org/clojure.core/distinct

PR #3201, opened 2021-11-29
feat: add transducer support for uniqBy/uniqWith

louietan:
As the title implies, this PR adds transducer support for uniq/uniqBy/uniqWith.
Unit test is updated to cover this change.

Related issue #1683

### DIFF
--- a/source/internal/_xuniqWith.js
+++ b/source/internal/_xuniqWith.js
@@ -1,0 +1,24 @@
+import _Set from './_Set.js';
+import _curry2 from './_curry2.js';
+import _includesWith from './_includesWith.js';
+import _xfBase from './_xfBase.js';
+
+
+function XUniqWith(pred, xf) {
+  this.xf = xf;
+  this.pred = pred;
+  this.items = [];
+}
+XUniqWith.prototype['@@transducer/init'] = _xfBase.init;
+XUniqWith.prototype['@@transducer/result'] = _xfBase.result;
+XUniqWith.prototype['@@transducer/step'] = function(result, input) {
+  if (_includesWith(this.pred, input, this.items)) {
+    return result;
+  } else {
+    this.items.push(input);
+    return this.xf['@@transducer/step'](result, input);
+  }
+};
+
+var _xuniqWith = _curry2(function _xuniqWith(pred, xf) { return new XUniqWith(pred, xf); });
+export default _xuniqWith;
--- a/test/transduce.js
+++ b/test/transduce.js
@@ -5,7 +5,9 @@ describe('transduce', function() {
   var add = R.add;
   var mult = function(a, b) {return a * b;};
   var isOdd = function(b) { return b % 2 !== 0; };
+  var isEven = function(b) { return b % 2 === 0; };
   var square = function(a) {return a * a;};
+  var negate = function(a) {return -1 * a;};
   var addxf = {
     '@@transducer/step': function(acc, x) { return acc + x; },
     '@@transducer/init': function() { return 0; },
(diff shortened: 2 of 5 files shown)


test/transduce.js
```diff
@@ -40,6 +41,9 @@ describe('transduce', function() {
     eq(R.transduce(R.filter(isOdd), R.flip(R.append), [],  [1, 2, 3, 4]), [1, 3]);
     eq(R.transduce(R.compose(R.map(add(1)), R.take(2)), R.flip(R.append), [],  [1, 2, 3, 4]), [2, 3]);
     eq(R.transduce(R.compose(R.filter(isOdd), R.take(1)), R.flip(R.append), [],  [1, 2, 3, 4]), [1]);
+    eq(R.transduce(R.compose(R.take(6), R.uniq(), R.map(square)), R.flip(R.append), [],  [1, 1, 2, 2, 3, 3, 4, 4]), [1, 4, 9]);
+    eq(R.transduce(R.compose(R.map(negate), R.uniqBy(Math.abs), R.map(square)), R.flip(R.append), [],  [-1, -5, 2, 10, 1, 2]), [1, 25, 4, 100]);
+    eq(R.transduce(R.compose(R.take(6), R.uniqWith(R.eqBy(String)), R.map(square)), R.flip(R.append), [],  [1, '1', '2', 2, 3, '3', '4', 4]), [1, 4, 9]);
```
CrossEye, line 46: I would like to see some tests with the `take` at the other end of the composition, since that is really the only reason I can even see to make `uniq*` into a transducer.

For instance if

```js
R.transduce(R.compose(R.uniq(), R.map(square), R.take(3)), R.flip(R.append), [],  [1, 1, 2, 2, 3, 3, 4, 4])
```

doesn't yield `[1, 4, 9]`, then we need to investigate carefully.

CrossEye (changes requested):
This looks very well written.

There was originally no support for the linked issue, and eventually it was closed.  Although I don't know that I've ever needed it, I can see some rationale for making `uniq*` into a transducer: a combination with `take`.  "Find the first three unique values when I do this transformation..."

I did make one request for the tests.  I don't think the tests supplied really show this rationale well.  They would be better with a `take` at the end of the composition.

Other than that, I'm happy with this, but would like to hear some feedback from @ramda/core before accepting it.

test/transduce.js
louietan: Thanks for the quick feedback. I've made the requested change and amended the commit.

CrossEye (dismissed):

adispring (approved):

source/internal/_xuniqWith.js
```diff
@@ -0,0 +1,24 @@
...
+  this.pred = pred;
+  this.items = [];
+}
+XUniqWith.prototype['@@transducer/init'] = _xfBase.init;
+XUniqWith.prototype['@@transducer/result'] = _xfBase.result;
+XUniqWith.prototype['@@transducer/step'] = function(result, input) {
+  if (_includesWith(this.pred, input, this.items)) {
+    return result;
+  } else {
+    this.items[this.items.length] = input;
```
customcommander, line 18: Was there a reason for not doing `this.items.push(input)`?

test/transduce.js
```diff
@@ -40,6 +41,9 @@ describe('transduce', function() {
     eq(R.transduce(R.filter(isOdd), R.flip(R.append), [],  [1, 2, 3, 4]), [1, 3]);
     eq(R.transduce(R.compose(R.map(add(1)), R.take(2)), R.flip(R.append), [],  [1, 2, 3, 4]), [2, 3]);
     eq(R.transduce(R.compose(R.filter(isOdd), R.take(1)), R.flip(R.append), [],  [1, 2, 3, 4]), [1]);
+    eq(R.transduce(R.compose(R.uniq(), R.map(square), R.take(3)), R.flip(R.append), [],  [1, 1, 2, 2, 3, 3, 4, 4]), [1, 4, 9]);
```
customcommander, line 44: Was expecting `R.compose(R.uniq, …)` instead of `R.compose(R.uniq(), …)`. Not sure if that's an issue though but thought I would ask.

test/transduce.js
customcommander: > > I would like to see some tests with the take at the other end of the composition, since that is really the only reason I can even see to make uniq* into a transducer.

I can also think of a list of unique numbers in which case it would look something like:

```
compose(filter(is_even), uniq)
```

Suppose this would work as expected though.

test/transduce.js
CrossEye: Yes, that should be fixed.  Because of the way our curry works, `ramdaFunc()` yields back the equivalent to `ramdaFunc`, but we should not abuse that.

test/transduce.js
louietan: Added the suggested testcase.

source/internal/_xuniqWith.js
louietan: No, I just saw somewhere in the source code did the same thing, and I picked it.
Array#push is definitely more readable and intuitive. I've updated this to use Array#push.

test/transduce.js
louietan: Yes, there's no need to call with arity 0, I've removed the brackets.

customcommander (approved):

customcommander:
@louietan Thank you for such a well written pull request and for addressing the comments so quickly. I'm happy for this to be merged but would prefer another approval first.

CrossEye (approved):
This has been a great pull request.  Thank you!

PR #3206, opened 2021-12-16
Fix for failing builds

customcommander:
Summary:

1. Allow CI pipeline on pull requests too (regardless of the destination; might be useful for long-lived branches e.g. next major versions)
2. In a previous PR the whitespace list was amended but not the test
3. In a previous PR linting errors were introduced but never fixed

@ramda/core Any thoughts?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,6 @@
 name: Build
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   coverage:
--- a/source/is.js
+++ b/source/is.js
@@ -3,7 +3,7 @@ import _curry2 from './internal/_curry2.js';
 
 /**
  * See if an object (i.e. `val`) is an instance of the supplied constructor. This
- * function will check up the inheritance chain, if any. 
+ * function will check up the inheritance chain, if any.
  * If `val` was created using `Object.create`, `R.is(Object, val) === true`.
  *
  * @func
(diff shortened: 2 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -1,6 +1,6 @@
 name: Build
 
-on: [push]
+on: [push, pull_request]
```
customcommander, line 3: Just noticed that @adispring has attempted to fix this in a different way. I went for a "lazy" approach and thought we could run the CI pipeline anytime a code change is suggested (regardless of the destination). If it's cheap and quick I can't see any an obvious reason to restrict this to the main branch only.

Any objections?

source/is.js
```diff
@@ -3,7 +3,7 @@ import _curry2 from './internal/_curry2.js';
 
 /**
  * See if an object (i.e. `val`) is an instance of the supplied constructor. This
- * function will check up the inheritance chain, if any. 
+ * function will check up the inheritance chain, if any.
```
customcommander, line 6: Previous builds were actually failing because of linting errors:
Screenshot 2021-12-16 at 21 59 48

test/trim.js
```diff
@@ -3,7 +3,7 @@ var eq = require('./shared/eq');
 
 
 describe('trim', function() {
-  var test = '\x09\x0A\x0B\x0C\x0D\x20\xA0\u1680\u180E\u2000\u2001\u2002\u2003\u2004\u2005\u2006\u2007\u2008\u2009\u200A\u202F\u205F\u3000\u2028\u2029\uFEFFHello, World!\x09\x0A\x0B\x0C\x0D\x20\xA0\u1680\u180E\u2000\u2001\u2002\u2003\u2004\u2005\u2006\u2007\u2008\u2009\u200A\u202F\u205F\u3000\u2028\u2029\uFEFF';
```
customcommander, line 6: The `\u180E` character was removed in a previous pull request but the tests were not updated at the same time. This is now fixed.

adispring (approved):

hitmands (approved):
:bowtie:

CrossEye:
Thanks, everyone.  I've been having hardware problems, which won't be addressed for a few more weeks.  It's great to see others taking up the helm!

CrossEye (commented):
LGTM

PR #3207, opened 2021-12-26
chore: remove dependencies badge

customcommander:
As of time of writing the service https://david-dm.org/ seems to be down
and/or unreliable. As Ramda is a no-dependency library I'm not seeing
much value in this badge.

Screenshot 2021-12-26 at 18 29 23

---

I know we have _way_ more important things to do but the broken link didn't look great and this is a very quick fix. If you prefer to keep the dependencies status here's a shields.io alternative:

Libraries.io dependency status for latest release


CrossEye (approved):
LGTM

PR #3210, opened 2021-12-28
Remove placeholder check from `identical`

### LINKED ISSUE #2533: Hint: Ramda throws DOMException when Cross-origin Window objects are accessed
Demo:
```javascript
// grab an iframe with a different origin than the page
var iframe = $0; 
iframe.contentWindow === iframe.contentWindow // true - so far, so good
R.identical(iframe.contentWindow, iframe.contentWindow); // Uncaught DOMException: 
// ... Blocked a frame with origin ... from accessing a cross-origin frame
```

E.g. this may occur when using iframes as poor-man's webcomponents or when exchanging data via cross-origin [`window.postMessage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) and using Ramda to validate the dispatched `MessageEvent` which contains the cross-origin Window object in the `source` property.

This is because script access to cross-origin Window objects is [very limited](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy#Cross-origin_script_API_access) (just a handful of properties and `.postMessage()`), accessing well-known Symbol keys like `a['@@functional/placeholder']` in Ramda's [`_isPlaceholder(a)`](https://github.com/ramda/ramda/blob/v0.25.0/source/internal/_isPlaceholder.js) which is used by every of Ramda's `curryN*()` functions raises that DOMException.

I don't see a quick fix for this, event when exempting Windows from being placeholders:
```javascript
var _isPlaceholder = function _isPlaceholder(a) {
    return a != null && typeof a === 'object' && !(a instanceof Window)  && a['@@functional/placeholder'] === true;
};

// Will not work because 
iframe.contentWindow instanceof Window // false for cross-origin Windows
```
So in summary I post my findings here just as a *warning* for all Ramda users, or maybe as a motivation to get rid of `@@functional/placeholder` (?)

semmel:
This fixes #2533, where calling `identical` with cross-origin browser window objects throws a DOMException.

The reason for the exception is the checking for the placeholder property in the `curry*` functions.

Since placeholder support makes no sense for commutative functions it can be removed from `identical`. This is done here by wrapping `isObject` with a `curry2` wrapper without placeholder property checks.

### DIFF
--- a/source/identical.js
+++ b/source/identical.js
@@ -42,7 +42,7 @@ var identical = function(a, b) {
           }
         };
       }());
-    case 2:
+    default:
       return _objectIs(a, b);
   }
 };
--- a/source/identical.js
+++ b/source/identical.js
@@ -1,5 +1,4 @@
 import _objectIs from './internal/_objectIs.js';
-import _curry2 from './internal/_curry2.js';
 
 
 /**
@@ -46,4 +45,10 @@ var identical = function(a, b) {
       return _objectIs(a, b);
   }
 };
+
+// In order to support Cross-origin Window objects as arguments to identical,
+// it cannot be implemented as _curry2(_objectIs). 
+// The reason is that _curry2 checks if a function argument is the placeholder __ 
+// by accessing a paritcular property. However, across URL origins access 
+// to most properties of Window is forbidden. 
 export default identical;
--- a/source/identical.js
+++ b/source/identical.js
@@ -31,9 +31,9 @@ import _objectIs from './internal/_objectIs.js';
 var identical = function(a, b) {
   switch (arguments.length) {
     case 0:
-      return identical; 
+      return identical;
     case 1:
-      return (function(){
+      return (function() {
         return function unaryIdentical(_b) {
           switch (arguments.length) {
             case 0:
@@ -49,8 +49,8 @@ var identical = function(a, b) {
 };
 
 // In order to support Cross-origin Window objects as arguments to identical,
-// it cannot be implemented as _curry2(_objectIs). 
-// The reason is that _curry2 checks if a function argument is the placeholder __ 
-// by accessing a paritcular property. However, across URL origins access 
-// to most properties of Window is forbidden. 
+// it cannot be implemented as _curry2(_objectIs).
+// The reason is that _curry2 checks if a function argument is the placeholder __
+// by accessing a paritcular property. However, across URL origins access
+// to most properties of Window is forbidden.
 export default identical;
--- a/source/identical.js
+++ b/source/identical.js
@@ -8,6 +8,8 @@ import _objectIs from './internal/_objectIs.js';
  *
  * Note this is merely a curried version of ES6 `Object.is`.
  *
+ * `identical` does not support the `__` placeholder.
+ *
  * @func
  * @memberOf R
  * @since v0.15.0
--- a/source/identical.js
+++ b/source/identical.js
@@ -27,5 +27,23 @@ import _curry2 from './internal/_curry2.js';
  *      R.identical(0, -0); //=> false
  *      R.identical(NaN, NaN); //=> true
  */
-var identical = _curry2(_objectIs);
+var identical = function(a, b) {
+  switch (arguments.length) {
+    case 0:
+      return identical; 
+    case 1:
+      return (function(){
+        return function unaryIdentical(_b) {
+          switch (arguments.length) {
+            case 0:
+              return unaryIdentical;
+            default:
+              return _objectIs(a, _b);
+          }
+        };
+      }());
+    case 2:
+      return _objectIs(a, b);
+  }
+};
 export default identical;
--- a/test/identical.js
+++ b/test/identical.js
@@ -1,5 +1,6 @@
 var R = require('../source');
 var eq = require('./shared/eq');
+var assert = require('assert');
 
 
 describe('identical', function() {
@@ -24,6 +25,38 @@ describe('identical', function() {
     eq(R.identical(0, new Number(0)), false);
     eq(R.identical(new Number(0), 0), false);
     eq(R.identical(new Number(0), new Number(0)), false);
+  });
+  
+  it('is auto-curried', function() {
+    assert.strictEqual(R.identical.length, 2);
+    var unaryFn = R.identical("foo");
+    assert.strictEqual(unaryFn.length, 1);
+    eq(unaryFn("bar"), false);
+    eq(unaryFn("foo"), true);
+    
+    eq(R.identical()("foo")()("foo"), true);
+  });
+  
+  it("does not access the placeholder property of it's arguments which is forbidden for cross-origin browser windows", function() {
+    var forbiddenPropertyAccessObject = {};
+    Object.defineProperty(
+      forbiddenPropertyAccessObject, 
+      '@@functional/placeholder', 
+      { get: function(){ throw new Error("Not allowed!"); } }
+    );
+    
+    assert.doesNotThrow(
+      () => R.identical(forbiddenPropertyAccessObject, {}),
+      Error
+    );
+    
+    assert.doesNotThrow(
+      () => R.identical({}, forbiddenPropertyAccessObject),
+      Error
+    );
+    
+    eq(R.identical(forbiddenPropertyAccessObject, forbiddenPropertyAccessObject), true);
+    eq(R.identical(forbiddenPropertyAccessObject, {}), false);
   });
 
 });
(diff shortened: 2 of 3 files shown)


source/identical.js
```diff
@@ -27,5 +27,23 @@ import _curry2 from './internal/_curry2.js';
...
+        return function unaryIdentical(_b) {
+          switch (arguments.length) {
+            case 0:
+              return unaryIdentical;
+            default:
+              return _objectIs(a, _b);
+          }
+        };
+      }());
+    case 2:
```
customcommander, line 45: Would it make sense for `case 2` to be the default case instead? It would handle the case were additional arguments are given (even though they would be ignored).

source/identical.js
```diff
@@ -27,5 +27,23 @@ import _curry2 from './internal/_curry2.js';
...
+      return (function(){
+        return function unaryIdentical(_b) {
+          switch (arguments.length) {
+            case 0:
+              return unaryIdentical;
+            default:
+              return _objectIs(a, _b);
+          }
+        };
+      }());
```
customcommander, line 44: Do we really need the IIFE and do we need the `case 0` case? I know that our `curry` implementation allows expressions such as `add()()(1)()()(2)` but I don't find that particularly helpful.

customcommander:
@semmel The link to the issue looks wrong. Did you mean https://github.com/ramda/ramda/issues/2533 ?

semmel:
> @semmel The link to the issue looks wrong. Did you mean https://github.com/ramda/ramda/issues/2533 ?

Yes indeed!

source/identical.js
semmel: This makes sense. It's fixed now in [semmel:patch-1](https://github.com/semmel/semmel-ramda/tree/patch-1)

source/identical.js
semmel: > ... `add()()(1)()()(2)` but I don't find that particularly helpful.

Me too. I have *no idea* why an (n > 0)-ary Ramda function `f` must satisfy `f() ≡ f`. In years using Ramda have not used this deliberately. Perhaps @CrossEye can enlighten us.

We could easily get rid of the IIFE here, but I did not want to break things.

test/identical.js
```diff
@@ -25,5 +26,37 @@ describe('identical', function() {
...
   });
+  
+  it('is auto-curried', function() {
+    assert.strictEqual(R.identical.length, 2);
+    var unaryFn = R.identical("foo");
+    assert.strictEqual(unaryFn.length, 1);
+    eq(unaryFn("bar"), false);
+    eq(unaryFn("foo"), true);
+    
+    eq(R.identical()("foo")()("foo"), true);
```
customcommander, line 37: Do we need to check the functions arity? Wouldn't these two tests be enough:

```javascript
eq(R.identical("foo", "foo"), true);
eq(R.identical("foo")("foo"), true);
```

The fact that things such as `add()(1)()(2)` work is just a consequence of how currying is internally implemented AFAIK. I'm not sure we need to showcase this use case or make sure it remains supported. OTOH this could be seen as a breaking change if you opted out from it so up to you ;)

source/identical.js
```diff
@@ -27,5 +27,23 @@ import _curry2 from './internal/_curry2.js';
  *      R.identical(0, -0); //=> false
  *      R.identical(NaN, NaN); //=> true
  */
-var identical = _curry2(_objectIs);
```
customcommander, line 30: IIRC `_curry2` was the only problem as it was trying to work out if a placeholder was given in place of an argument? I didn't check `_objectIs`; have you checked that it won't try to access things that cross-origin windows would forbid? I think leaving a comment would be helpful as I can see how tempting it would be for someone in the arbitrary future to rollback that change thinking **Why didn't they use `_curry2`?!**

test/identical.js
```diff
@@ -25,5 +26,37 @@ describe('identical', function() {
...
+    );
+    
+    assert.doesNotThrow(
+      () => R.identical({}, forbiddenPropertyAccessObject),
+      Error
+    );
+    
+    eq(R.identical(forbiddenPropertyAccessObject, forbiddenPropertyAccessObject), true);
+    eq(R.identical(forbiddenPropertyAccessObject, {}), false);
+  });
```
customcommander, line 60: I see what you're trying to do here but I've mixed feelings about the methodology. I think it's focusing a little bit too much on the implementation. If we ever decide to rename `@@functional/placeholder` this test would probably still pass and potentially create a false positive.

Perhaps an alternative would be to make sure that giving a placeholder doesn't do what it is supposed to do:

In this case we expect the placeholder to cause a function to be returned:

```javascript
add(__, 2) //=> expect a function here
```

In your case we could make sure that a placeholder is treated like any other value. Couple of options:


```javascript
assert(typeof identical(__, 42) === 'boolean');
assert(identical(__, __) === true);
assert.doesThrow(() => identical(42)(__)(42));
```

customcommander:
Got a couple of comments (all minors; feel free to ignore) but these changes make sense to me. Thanks.

test/identical.js
semmel: Yes, I follow your points.

I'll integrate your idea of checking placeholder to be treated as regular value. Although that will verify `__` is no longer supported by `identical`, it will not verify the DOMException is no longer triggered. For that one would

- have to execute the test in a browser context, and setting up a cross origin iframe, or
- mock up a test object which throws on *any* property access, not just on `['@@functional/placeholder']`

To be honest, in the context of the brittle Ramda test infrastructure***, the first approach seems to me too complicated, while for the second I still haven't come up with an implementation yet.  

*** Why are the unit test written as CJS modules when the source code is all ES modules? 
Mocha can [work with ESM test files](https://mochajs.org/#nodejs-native-esm-support) for some time. Code is otherwise ES5 anyway or is it compiled to ES5 at all? 
All that `cross-env BABEL_ENV=cjs mocha --require @babel/register` seems to me much overtooled.

test/identical.js
semmel: Removed the line with `()` fake calls.

source/identical.js
semmel: Added a comment why `_curry2` cannot be used in the source file.

test/identical.js
semmel: Re-wrote that test with a mock object with just a few allowed properties. 

Unfortunately I had to use `Proxy` — hoping the tests don't run on Internet Explorer R.I.P. Just with ES5, I don't think a test failing for the old implementation can be written.

customcommander:
@semmel Could you have a look at the "Checks" tab please? There are lots of linting errors that's failing the build atm.

semmel:
@customcommander Thanks for running the actions and looking into the results!

`test/invariants.js` checks for the `length` property (arity?) in all `R.*` functions after applying to stuff.

The second test where the functions are applied to `R.__` fails because it expects `R.identical(R.__).length` to be 2.

Should I patch `test/invariants.js` to exclude `R.identical` from the placeholder check ?

semmel:
> @semmel Could you have a look at the "Checks" tab please? There are lots of linting errors that's failing the build atm.

done

source/identical.js
```diff
@@ -9,6 +8,8 @@ import _curry2 from './internal/_curry2.js';
  *
  * Note this is merely a curried version of ES6 `Object.is`.
  *
+ * `identical` does not support the `__` placeholder.
```
customcommander, line 11: Forgot to say: kudos for updating the docs too.

customcommander (approved):

customcommander:
@semmel This looks good to me but I really would like another review from @ramda/core

customcommander:
Perhaps we can wait until v0.28.0 is released?

semmel:
> Perhaps we can wait until v0.28.0 is released?

More approval would be fine. We can wait so long. 

With the holiday season gone however, my focus will wander off. ⛵

CrossEye:
I'd like the next version after 0.28.0 to be 1.0, and it might be a quick release now that @customcommander has so diligently been cleaning up issues.  After that, we can remove the longstanding restriction on ES6+ code in Ramda's codebase.  I would actually love for v2.0 to be rewritten in modern JS.  But I don't think we can include this fix in the next release.

This is not code to love, but it looks like a reasonable way around the problem.

semmel:
> After that, we can remove the longstanding restriction on ES6+ code in Ramda's codebase.

@CrossEye Is that because of the use of `Proxy` in the test file?

CrossEye:
@semmel:

Well, not just the `Proxy`, but also `Object.defineProperty`.  I was hurrying through trying to play catch-up with all the work @customcommander has been putting in, saw that code, and failed to notice that it was in the test file.  I'm much less concerned with test files and whether they are es3/es5/es6/esnext compatible.

So I withdraw my objection and will try to give this a closer look tomorrow.

customcommander:
@CrossEye Do you think this is ready to go?

PR #3213, opened 2022-01-08
Restore functions removed after v0.27.0 to avoid breaking changes in v0.28.0

customcommander:
⚠️⚠️⚠️ **DO NOT MERGE!** ⚠️⚠️⚠️

--- 

This partially reverts commit 3366df607287bf5cf3b9f870ad5b463198f334e5.

This commit temporarily restores the functions removed
after the initial v0.27 release so as to avoid causing
breaking changes in the next minor release v0.28.0

This commit also temporarily restore a dependency to `jsverify`
needed for the tests that have also been restored as a
result of reverting the commit.

---

**Warning**

The destination branch should probably be set to `v0.28` and not `master`.

**Summary**

1. Source and tests files are exactly as they were in commit 3366df607287bf5cf3b9f870ad5b463198f334e5. The only difference is that I had to add the `.js` extension in the import statements as otherwise the linter fails.
2. I didn't want to rewrite the tests so I had to restore the dependency to `jsverify`
3. Some tests were refactored to not use any of the deprecated functions. These tests have not been reverted!
4. The `index.js` file looks a bit messy but I've only restored the deleted functions and sorted the lines alphabetically.


CrossEye:
We have not worried too much about breaking changes in the pre-1.0 (I know, I know!) releases.  After 0.28, I think we should be ready to go to 1.0, and talk about the future of Ramda from there.  But at that point, we should pay strict attention to semver rules.

PR #3214, opened 2022-01-09
remove package-lock.json

davidchambers:
It does not make sense for a library to use a lockfile.

PR #3217, opened 2022-01-16
Remove & Disable package-lock

imcotton:
Finish up #3214.


customcommander:
I would feel more comfortable if we could review our dependencies management first.

With millions of weekly downloads we have definitely become a target and so we ought to take all reasonable steps to make sure that our development infrastructure isn't hijacked by compromised third-party dependencies.

I think a lock file has its role to play. I know it's not perfect but we shouldn't discard it either. We could think about pinning our dependencies. We could also think about disabling pre/post install scripts. Any dependencies can trigger those scripts and install malwares on privileged machines and attempt to publish a compromised versions of `ramda`.

I would therefore suggest we do not merge this for now until we had that discussion first. If any members of @ramda/core strongly disagree and would prefer to merge this pr I will not object of course.

imcotton:
@customcommander Agree to have discussion around first, I've made this PR into draft for now.

imcotton:
`npm audit` result on before and after removing the lockfile:

```diff
- 50 vulnerabilities (       23 moderate, 18 high, 9 critical)
+ 28 vulnerabilities (2 low, 18 moderate,  6 high, 2 critical)
```

customcommander:
This reinforces my feeling about the lock file; it has its role to play (it has effectively pinned down the dependencies) but it alone is not enough. The maintainers have to review the dependencies and update them after they been approved. The latest events show how easy it is to compromise a project supply chain. Due diligence is required IMHO.

hitmands:
Hi @imcotton , thanks for the pr! 💪 !

I'd love if you could add some context, description, to your pr.
What would the benefits of removing lock files be?

CrossEye:
https://github.com/ramda/ramda/pull/3214#issuecomment-1018994917: 

> I have quite mixed feelings about lockfiles in general.  I love the idea, but the npm implementation has often been a thorn in my side.  And while I don't necessarily agree with David's suggestion that libraries do not need them, there is one factor that convinces me that Ramda doesn't need to commit one:
> 
> https://github.com/ramda/ramda/blob/7f37e8a4243de2b6c39d0490ef282873e405f9ad/package.json#L89
> 
> Yes, there are plenty of `devDependencies`, and it's possible that a corrupted dev dependency does something terrible to a built version of Ramda, but *it's not transitive*.   Any such problems can be fixed solely by changing Ramda versions; you won't need to worry about what version of `deeply/nested/library/that/someone/has/corrupted` might be.

imcotton:
What should a proper lockfile benefits from having one?

- updated - npm/cli#4223
- ethics - https://snyk.io/blog/why-npm-lockfiles-can-be-a-security-blindspot-for-injecting-malicious-modules/
- auditing - https://overreacted.io/npm-audit-broken-by-design/

Otherwise it wasting the effort.

customcommander:
> Yes, there are plenty of devDependencies, and it's possible that a corrupted dev dependency does something terrible to a built version of Ramda, but it's not transitive. Any such problems can be fixed solely by changing Ramda versions; you won't need to worry about what version of deeply/nested/library/that/someone/has/corrupted might be.

I am not following the reasoning here.

Compromised libraries can either be direct dependencies or _transitive_ dependencies. Coupled with pre/post install scripts they can install malwares as soon as ramda is installed on a computer. That we only need to change the ramda version is in my opinion irrelevant as the damage is done already.

I think we should try our best to prevent issues in the first place not react to them afterwards.

I'm not saying that a lock file is the only solution but I think it still has its role to play.

I am personally not in favour to merge this as it is now.

adispring:
Using exact version in package.json may have the same effect as what the lock file do?

We can first remove all the `~` and `^` version symbol in package.json?

CrossEye:
@customcommander: But Ramda has *no* dependencies.

```shell
> mkdir 2022-01-22c
> cd 2022-01-22c
> npm init
> npm install --save ramda
> echo -e "const {map} = require ('ramda')\n\nconsole .log (map (n => n * n) ([1, 2, 3,4, 5]))" >index.js
> node index
[ 1, 4, 9, 16, 25 ]
> cat node_modules/.package-lock.json 
{
    "name": "2022-01-22c",
    "version": "1.0.0",
    "lockfileVersion": 2,
    "requires": true,
    "packages": {
        "node_modules/ramda": {
            "version": "0.28.0",
            "resolved": "https://registry.npmjs.org/ramda/-/ramda-0.28.0.tgz",
            "integrity": "sha512-9QnLuG/kPVgWvMQ4aODhsBUFKOUmnbUnsSXACv+NCQZcHbeb+v8Lodp8OVxtRULN1/xOyYLLaL6npE6dMq5QTA==",
            "funding": {
                "type": "opencollective",
                "url": "https://opencollective.com/ramda"
            }
        }
    }
}
```

Note that the only thing there is Ramda.

 
So this is first of all only a concern for those doing `git clone`/`npm install`, who will get our `devDependencies`.  For that, I think I prefer @adispring's idea: Let's simply lock down the versions of all our dev dependencies, and remove this megabyte file.

customcommander:
This is only an issue for maintainers and contributors; those who will clone the repo and bootstrap the project with `npm i` but in my view this is still an issue. Some malicious actors will target maintainers of popular packages to steal their tokens and publish compromised versions of popular package.

In order to move forward I can live without a lock file but then this pull request has to pin down the dependencies now as otherwise it just removes _some_ protection for nothing in return IMHO.

Note that I am not asking for the dependencies to be updated just locked.

.github/workflows/build.yml
```diff
@@ -11,7 +11,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Installing Dependencies
-        run: npm ci
+        run: npm install
```
customcommander, line 14: Perhaps we could have the `--ignore-scripts` here?

CrossEye: Makes sense.

CrossEye:
> In order to move forward I can live without a lock file but then this pull request has to pin down the dependencies now as otherwise it just removes some protection for nothing in return IMHO.

I like this.  Anyone have an objection?

.github/workflows/build.yml
imcotton: If this is preferred I'd put it into `.npmrc` instead.

customcommander:
> I'm sorry for pulling the leg here, just two days ago I've had PR #3223 submitted, does anyone occur to review the actual diff inside package-lock.json and verify its legitimacy? (disclaimer: nothing malicious)

I actually reviewed the lock file :) and made sure all packages were pull from the default registry and also looked for the latest compromised packages that I had to deal with in the last few months.

customcommander:
You're not wrong that pinning dependencies or using a lock file makes it less trivial to keep dependencies up to date. That's totally fair but for me that's a win (just my personal opinion here). I'm not of fan of pulling the latest version of everything every single time. It's literally a blank cheque to install whatever on your machines. Effective security in my opinion is about preventing not reacting to threats.

Some companies actually have that model, all dependencies are carefully vetted before consumed. In the meantime people just stick with pre-approved packages. It may seem inflexible but if the update process is quick, reliable and frequent, this model becomes a very useful barrier against totally preventable security incidents.

I won't object to removing the lock file. Actually assuming that humans will always review a machine-generated file is a security issue in itself so in a sense you're helping us (and thank you for your contribution) but I would really love to see the dependencies pinned down as a result.

I am not asking for the dependencies to be updated though. Just for the `package.json` file to be updated to remove `^` and `~` symbols. I think that's an ok compromise don't you think?

Of course any other maintainers is free to disagree and merge this as it is.

imcotton:
dep-dates

By removing the version prefix you're going to have to maintain a repo with all those outdated external modules, to me in any way it is not a secure measure. Keep in mind by dropping `package-lock` today you still going to fetch semver compatible versions, not the latest ones, without a major bump they're still outdated heavily.

Check again these giants:

- bable
- rollup
- eslint
- mocha
- js-yaml
- etc...

if someone or some companies have the secrecy to effectively verify each upgrades then I'm all ears. And if some how they were compromised by a single version bump, **Ramda** repo is the last place I'd worried about.

customcommander:
I think we should simply agree to disagree here.

I voiced my concerns and if other maintainers do not share them they are more than welcome to merge your pull request. In any case thank your for your time and contributions, they are well appreciated.

imcotton:
I concur.

_btw: branch rebased_

imcotton:
Sorry, I want to last time emphasize the security concern on pinning to the exact version.

Assuming module `foobar` has the following versions:

- 1.5.0  -- vulnerable resolved
- 1.3.0
- 1.2.0  -- vulnerable introduced

In the `package.json` you have:

```json
{
  "foobar": "^1.3.0"
}
```

On a fresh install, you'd fetch the `1.5.0` (semver compatible).

However, by pin to the exact version:

```diff
- "foobar": "^1.3.0"
+ "foobar": "1.3.0"
```

On another install, you'd get the version `1.3.0` instead (vulnerable included).

This is hardly new knowledge, please be awarded.

CrossEye:
How's this for a process?

  - Before a major release, we strip out all the dev dependencies, and reinstall them with the latest (non-prerelease) versions,
 
    ```shell
    npm install --save-dev --save-exact @babel/cli  @babel/core ... xyz
    ```
    (I'm bothered though about needing another place to track these dependencies, so perhaps we simply pull them first from the current `package.json`.)

  - Obviously we then run tests and the full build/publish cycle.

  - Between major versions, when we find the need to change dependency versions, we simply update `package.json` with the latest.  We can release this change on its own in a patch version and/or with other code in a minor version.

imcotton:
I don't know how to proceed that flow in practice, at this point, I believe that ramda team knows best in this regard, closing.

PR #3223, opened 2022-01-22
Upgrade and fixing up eslint-plugin-import

imcotton:
During PR #3217, I noticed lint failing after install `eslint-plugin-import` to its latest version, so here comes the fixes mostly by adding extensions under `test/*` and `lib/*` folders.

Please help to verify it doesn't broken any other hidden process along the way.


customcommander (approved):
I was surprised by this so I ran the linter before and after the upgrade of the plugin and I can reproduce the same issue.

Thank you.

CrossEye:
Thanks, @imcotton!

PR #3224, opened 2022-01-23
Fix build errors

customcommander:
Fix remaining linter errors following #3223 and update ESLint configuration to have the parser understanding 2018 ECMA features. Also the `swap` function claims to work with Map instances however looking at the implementation I'm not sure how this could work so I have removed the failing test.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -127,5 +127,13 @@
       2,
       "always"
     ]
-  }
+  },
+  "overrides": [
+    {
+      "files": ["test/**/*.js"],
+      "parserOptions": {
+        "ecmaVersion": 2018
+      }
+    }
+  ]
 }
--- a/.eslintrc
+++ b/.eslintrc
@@ -133,6 +133,13 @@
       "files": ["test/**/*.js"],
       "parserOptions": {
         "ecmaVersion": 2018
+      },
+      "rules": {
+        "no-restricted-properties": ["error", {
+          "object": "describe",
+          "property": "only"
+        }],
+        "no-restricted-modules": ["error", ".."]
       }
     }
   ]
--- a/source/swap.js
+++ b/source/swap.js
@@ -62,7 +62,6 @@ var swapString = function(indexA, indexB, s) {
  *      R.swap(-1, 0, ['a', 'b', 'c', 'd', 'e', 'f']); //=> ['f', 'b', 'c', 'd', 'e', 'a'] list rotation
  *      R.swap('a', 'b', {a: 1, b: 2}); //=> {a: 2, b: 2}
  *      R.swap(0, 2, 'foo'); //=> 'oof'
- *      R.swap(obj1, obj2, new Map ([[obj1, 1] [obj2, 2]]); //=> new Map ([[obj1, 2], [obj2, 1]])
  */
 var swap = _curry3(function(indexA, indexB, o) {
   if (_isArray(o)) {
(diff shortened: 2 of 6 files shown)


.eslintrc
```diff
@@ -1,6 +1,7 @@
 {
   "parserOptions": {
-    "sourceType": "module"
+    "sourceType": "module",
+    "ecmaVersion": 2018
```
customcommander, line 4: Fix for `s` regex flag not recognised by our current ESLint configuration. We just need to tell the parser to expect 2018 ECMA features.

source/swap.js
```diff
@@ -62,7 +62,6 @@ var swapString = function(indexA, indexB, s) {
  *      R.swap(-1, 0, ['a', 'b', 'c', 'd', 'e', 'f']); //=> ['f', 'b', 'c', 'd', 'e', 'a'] list rotation
  *      R.swap('a', 'b', {a: 1, b: 2}); //=> {a: 2, b: 2}
  *      R.swap(0, 2, 'foo'); //=> 'oof'
- *      R.swap(obj1, obj2, new Map ([[obj1, 1] [obj2, 2]]); //=> new Map ([[obj1, 2], [obj2, 1]])
```
customcommander, line 65: As far as I can tell there is now way this could work given the current implementation. Could someone confirm please?

test/swap.js
```diff
@@ -32,8 +29,8 @@ describe('swap', function() {
   });
 
   it('swaps property values from one property to another', function() {
-    eq(R.swap('a', 'b', o), {a: 2, b: 1});
-    eq(R.swap('b', 'a', o), {a: 2, b: 1});
+    eq(R.swap('a', 'b', {a: 1, b: 2}), {a: 2, b: 1});
+    eq(R.swap('b', 'a', {a: 1, b: 2}), {a: 2, b: 1});
```
customcommander, line 33: Colocating input & output to make it easier to understand how things are connected.

test/swap.js
```diff
@@ -44,8 +41,4 @@ describe('swap', function() {
   it('swaps characters in string from one index to another', function() {
     eq(R.swap(0, 2, 'foo'), 'oof');
   });
-
-  it('swaps object indexes in map from one index to another', function() {
-    eq(R.swap(o, o2, map), new Map([[o, 2], [o2, 1]]));
-  });
```
customcommander, line 50: Given the current `swap` implementation there is no way this could ever work. Unless I'm missing something.

.eslintrc
adispring: Maybe we can set ecmaversion only in test/.eslintrc, which will only affect the test file.

adispring (changes requested):

.eslintrc
customcommander: I didn't realise we had such file. I took the liberty to delete `test/.eslintrc` and put its content into the `overrides` property of the main ESLint configuration file:

```none
  "overrides": [
    {
      "files": ["test/**/*.js"],
      "parserOptions": {
        "ecmaVersion": 2018
      },
      "rules": {
        "no-restricted-properties": ["error", {
          "object": "describe",
          "property": "only"
        }],
        "no-restricted-modules": ["error", ".."]
      }
    }
  ]
```

I think this is the best of both worlds: we have only one configuration file to maintain whilst still allowing specific paths to override the configuration. 

So now the `parserOptions.ecmaVersion` is set to `2018` only when the linter runs on test files.

Does that work for you and everybody else?

adispring (approved):

test/swap.js
CrossEye: That's odd.  No problem removing it.

PR #3225, opened 2022-01-31
chore: rename innerJoin to intersectionWith

adispring:
`innerJoin` is a sql command, `R.innerJoin` only has `inner` but no `join`, so rename it to `intersectionWith` may be more accurate?

### DIFF
--- a/source/index.js
+++ b/source/index.js
@@ -87,7 +87,7 @@ export { default as includes } from './includes.js';
 export { default as indexBy } from './indexBy.js';
 export { default as indexOf } from './indexOf.js';
 export { default as init } from './init.js';
-export { default as innerJoin } from './innerJoin.js';
+export { default as intersectionWith } from './intersectionWith.js';
 export { default as insert } from './insert.js';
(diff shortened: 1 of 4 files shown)


customcommander (commented):
Makes sense to me. Of course that's a breaking change ;) Question: did you rename the test file?

adispring:
@customcommander , Yes, thanks for remind.

customcommander (approved):
LGTM and makes perfect sense. Would like another approval just in case. Otherwise let's merge this end of week latest?

customcommander:
@CrossEye Any objection?

CrossEye:
I agree with the change.

I guess ,,we need to decide whether `1.0` or `0.29` is next.  When we rename a function, we usually keep the old name around as an alias for a minor version.  If we're building `0.29` I would do so, but I think if we're headed right for `1.0`, then we need to skip it this time.

customcommander:
@CrossEye: I wouldn't bother with `0.29` at this point and would rather keep the momentum for `1.0.0` and ship as many breaking changes in that one.

PR #3227, opened 2022-02-01
chore: make intersection's output order consistent

adispring:


CrossEye:
Have you reviewed #1597?

Issue #3230, opened 2022-02-07
Removing curry dependency for internal transducer creator functions.


xgbuils:
I've noticed ramda is using curried functions for internal functions like transducer creators. I'm pretty sure this is causing a lack of performance. 

Arbitrary curried functions could be interesting for public funcions. But transducer creators are used always in _dispatchable internal function that requires a transducer creator (`(...args) -> xf -> xf`). Then, wrapping transducerCreators could create an unnecessary increase of arguments.length and isPlaceHolder checks (See `internal/_curry1.js`, `internal/_curry2.js`, `internal/_curry3.js` and  `internal/_curryN.js`).

CrossEye:
Thank you.  Looks like a good idea.  I'll dig deeper into the PR soon.

PR #3231, opened 2022-02-07
refactor(3230): remove curryX dependency for internal transducer creator functions

xgbuils:
From issue #3230 

Notice that now transducer creator functions just depend on  themselves after removing curryX functions. This could facilitate ramda modularity because now it's easier to create a library that works just with transducers without rely in other parts of the library.

### DIFF
--- a/source/dropRepeats.js
+++ b/source/dropRepeats.js
@@ -24,6 +24,8 @@ import equals from './equals.js';
  *     R.dropRepeats([1, 1, 1, 2, 3, 4, 4, 2, 2]); //=> [1, 2, 3, 4, 2]
  */
 var dropRepeats = _curry1(
-  _dispatchable([], _xdropRepeatsWith(equals), dropRepeatsWith(equals))
+  _dispatchable([], function() {
+    return _xdropRepeatsWith(equals);
+  }, dropRepeatsWith(equals))
 );
 export default dropRepeats;
--- a/source/internal/_xall.js
+++ b/source/internal/_xall.js
@@ -1,4 +1,3 @@
-import _curry2 from './_curry2.js';
 import _reduced from './_reduced.js';
 import _xfBase from './_xfBase.js';
 
@@ -23,5 +22,6 @@ XAll.prototype['@@transducer/step'] = function(result, input) {
   return result;
 };
 
-var _xall = _curry2(function _xall(f, xf) { return new XAll(f, xf); });
-export default _xall;
+export default function _xall(f) {
+  return function(xf) { return new XAll(f, xf); };
+}
--- a/source/internal/_xany.js
+++ b/source/internal/_xany.js
@@ -1,4 +1,3 @@
-import _curry2 from './_curry2.js';
 import _reduced from './_reduced.js';
 import _xfBase from './_xfBase.js';
 
@@ -23,5 +22,6 @@ XAny.prototype['@@transducer/step'] = function(result, input) {
   return result;
 };
 
-var _xany = _curry2(function _xany(f, xf) { return new XAny(f, xf); });
-export default _xany;
+export default function _xany(f) {
+  return function(xf) { return new XAny(f, xf); };
+}
--- a/source/internal/_xaperture.js
+++ b/source/internal/_xaperture.js
@@ -1,5 +1,4 @@
 import _concat from './_concat.js';
-import _curry2 from './_curry2.js';
 import _xfBase from './_xfBase.js';
 
 
@@ -32,5 +31,6 @@ XAperture.prototype.getCopy = function() {
   );
 };
 
-var _xaperture = _curry2(function _xaperture(n, xf) { return new XAperture(n, xf); });
-export default _xaperture;
+export default function _xaperture(n) {
+  return function(xf) { return new XAperture(n, xf); };
+}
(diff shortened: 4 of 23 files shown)


customcommander (commented):
I think I understand your motivation. If I understand correctly your thinking is that the internal functions don't need to be curried because they are never called with their full arguments and always with the same set. So we could optimise by currying ourselves basically.

I think this is perfectly sensible but I need to get my head around `_dispatchable` as this function seems key in order to conduct a review (at least for me).

Do you have an idea of how much we gain performance wise?

Overall I think this is pretty nice. Good catch!

xgbuils:
> I think this is perfectly sensible but I need to get my head around `_dispatchable` as this function seems key in order to conduct a review (at least for me).

I'm thinking deeply about it and I can say that technically it doesn't depends on `_dispatchable` it's more about how transducers work. Transducer implementations requires:
* transducer creators: functions that return transducers
* transducers: functions that receive a transformer and returns a transformer. They are the functions that can be composed applying a performant way of transformation over  arrays, iterables and so on.
* transformers: objects that implement `@@transducer/init`, `@@transducer/step` and `@@transducer/result` methods.

Because of composition of transducers we won't need another way of partial application over transducer creators. Then, it seems we will never need to wrap internal transducer creators with curryX functions.

> Do you have an idea of how much we gain performance wise?

I'm not sure how to check it. I've seen tables with benchmarks in #3048 but I don't know how they were created. Is there a standard way to create benchmarks in this project?

Thanks

customcommander:
> I'm thinking deeply about it and I can say that technically it doesn't depends on `_dispatchable` it's more about how transducers work. Transducer implementations requires:

Yep you're right. I mentioned `_dispatchable` because it deals with creating transducers too which is the codepath that most of your changes will exercise I think. 

We recently had this issue reported #3232 which seems to be a regression in `0.28` and I know we touched that part of the code recently. Any chance your changes improve anything there? (Don't think so but I thought I'd ask anyway)

I'd like another review from a maintainer if that's ok.

xgbuils:
> We recently had this issue reported #3232 which seems to be a regression in `0.28` and I know we touched that part of the code recently. Any chance your changes improve anything there? (Don't think so but I thought I'd ask anyway)

OK, I have a solution for this issue. I can send a PR there. It's working too with this change.

xgbuils:
> Do you have an idea of how much we gain performance wise?

I've added these bench tests:
``` javascipt
    //  filter.bench.js
    'into([], filter(isEven), nums)': function() {
      into([], filter(isEven), nums);
    },
    'into([], filterEven, nums)': function() {
      into([], filterEven, nums);
    },
```
``` javascipt
    //  map.bench.js
    'into([], map(sq), nums)': function() {
      into([], map(sq), nums);
    },
    'into([], mapSq, nums)': function() {
      into([], mapSq, nums);
    },
```

Original benchmarks:
```
│ filter                 │ Hertz                  │ Margin of Error        │
│ filter(isEven, nums)   │ 7,627,059              │ 0.77%                  │
│ filter(isEven)(nums)   │ 1,889,646              │ 2.87%                  │
│ filterEven(nums)       │ 2,230,079              │ 2.80%                  │
│ into([], filter(isEve… │ 350,420                │ 0.57%                  │
│ into([], filterEven, … │ 362,084                │ 0.69%                  │
│ native filter          │ 9,576,447              │ 0.14%                  │
│ map                    │ Hertz                  │ Margin of Error        │
│ map(sq, nums)          │ 9,675,896              │ 2.79%                  │
│ map(sq)(nums)          │ 2,090,064              │ 0.42%                  │
│ mapSq(nums)            │ 2,561,379              │ 0.41%                  │
│ into([], map(sq), num… │ 358,597                │ 0.34%                  │
│ into([], mapSq, nums)  │ 359,507                │ 0.78%                  │
│ native map             │ 10,826,670             │ 0.37%                  │
```

With the change:
```
│ filter                 │ Hertz                  │ Margin of Error        │
│ filter(isEven, nums)   │ 8,391,643              │ 0.60%                  │
│ filter(isEven)(nums)   │ 1,902,025              │ 0.28%                  │
│ filterEven(nums)       │ 2,227,084              │ 1.13%                  │
│ into([], filter(isEve… │ 389,342                │ 0.38%                  │
│ into([], filterEven, … │ 402,658                │ 0.87%                  │
│ native filter          │ 10,478,337             │ 0.20%                  │
│ map                    │ Hertz                  │ Margin of Error        │
│ map(sq, nums)          │ 10,496,247             │ 0.61%                  │
│ map(sq)(nums)          │ 1,748,662              │ 6.52%                  │
│ mapSq(nums)            │ 2,351,095              │ 0.23%                  │
│ into([], map(sq), num… │ 383,266                │ 0.27%                  │
│ into([], mapSq, nums)  │ 387,048                │ 0.73%                  │
│ native map             │ 9,842,714              │ 0.61%                  │
```

Do you think it could be interesting to add in the PR for testing transducer approach performance?

CrossEye (approved):
I'm sorry to be slow on this one.  I keep going to it, and seeing 24 changed files -- having to do with transducers, which I was never really a part of -- and saying, "Well, no time now.  I'll look later."

I finally dug in to realize how simple a change this is.  It makes perfect sense to me.  The boost in performance is nice, but the code simplification is its biggest advantage.

Very nice!  :herb:

customcommander (approved):

PR #3233, opened 2022-02-08
bugfix: groupBy transducer functionality

### LINKED ISSUE #3232: groupBy (permanently?) enters a broken state when used as transducer
While `groupBy` is one of the Ramda functions documented as being usable as a transducer, and I believe this worked years ago when I last tried it, now it appears to pass through every item into every group list. After this, every subsequent use of `groupBy` (as a transducer or otherwise) will be cloned as the starting accumulator for each group.

I'm guessing there was an update to `reduceBy` at some point that broke the arity to cause this but haven't quite gone through all the layers of indirection to narrow it down.

## Repro

```javascript
const {groupBy, into, range} = require('ramda');

const evenOdd = x => x % 2  === 0 ? 'even' : 'odd';
// this works fine.. if done *first*
groupBy(evenOdd, range(1, 10); // got expected {"odd":[1,3,5,7,9],"even":[2,4,6,8]}

into({}, groupBy(evenOdd), range(1, 10));
// expected: {"odd":[1,3,5,7,9],"even":[2,4,6,8]}
// actual: {"odd":[1,2,3,4,5,6,7,8,9],"even":[1,2,3,4,5,6,7,8,9]}

// It appears to simply pass through all values for each key returned by the group function. After this, any 
// subsequent use by `groupBy`, even in the non-transducer form, appends the results to of the previous list:

groupBy(x => String(x.length), ['a', 'ab', 'abc']);
// expected {"1":["a"],"2":["ab"],"3":["abc"]
// actual: '{"1":[1,2,3,4,5,6,7,8,9,"a"],"2":[1,2,3,4,5,6,7,8,9,"ab"],"3":[1,2,3,4,5,6,7,8,9,"abc"]}'
```

## comparing `countBy`

```javascript
const evenOdd = x => x % 2 === 0 ? 'even' : 'odd';
R.into({}, R.countBy(evenOdd), R.range(1, 10)); // { odd: 5, even: 4 }
```

jaawerth:
Fixes #3232
Reverts code change in #9b5d8925

Setting an absent absent/nullish accumulator, rather than passing an empty on to `reduceBy,` appears to correct the cached copy behavior causing pollution in the accumulators on subsequent invocations, as well as the "passthrough" behavior dumping every value in every group; I believe this is due due the cached shallow-copying combined with the currying of `reduceBy`.

It may be more desirable to correct the copy caching to prevent the error while still deterministically setting the accumulator, but for what it's worth, it's at least in line with traditional transducer behavior to get the initial value from the transformer, albeit not *quite* this way!

### DIFF
--- a/source/groupBy.js
+++ b/source/groupBy.js
@@ -44,7 +44,10 @@ import reduceBy from './reduceBy.js';
  *      // }
  */
 var groupBy = _curry2(_checkForMethod('groupBy', reduceBy(function(acc, item) {
+  if (acc == null) {
+    acc = [];
+  }
   acc.push(item);
   return acc;
-}, [])));
+}, null)));
 export default groupBy;
--- a/test/groupBy.js
+++ b/test/groupBy.js
@@ -44,4 +44,12 @@ describe('groupBy', function() {
     eq(_isTransformer(R.groupBy(byType, xf)), true);
   });
 
+  it('works as a transducer without polluting state', function() {
+    eq(R.into([], R.groupBy(x => x.length), ['a', 'xyz', 'ab', 'xy', 'abc', 'x']), [
+      [1, ['a', 'x']],
+      [2, ['ab', 'xy']],
+      [3, ['xyz', 'abc']]
+    ]);
+    eq(R.groupBy(x => x.length, ['a', 'xyz', 'ab', 'xy', 'abc', 'x']), {1: ['a', 'x'], 2: ['ab', 'xy'], 3: ['xyz', 'abc']});
+  });
 });


xgbuils:
Hi @jaawerth

I've seen the fallback function of [reduceBy](https://github.com/ramda/ramda/blob/d27c944b4aa974d42b68899270311fb10419ad79/source/reduceBy.js#L58) is using `_clone`. Then maybe could be done as same way is done in `_xReduceBy`. Let's see what the mantainers say.

Thanks!

jaawerth:
Yep - this is a quick fix but it probably doesn't address similar problems in other transducers. I think the PR you closed may be on track of a better solution ;).

source/groupBy.js
```diff
@@ -44,7 +44,10 @@ import reduceBy from './reduceBy.js';
  *      // }
  */
 var groupBy = _curry2(_checkForMethod('groupBy', reduceBy(function(acc, item) {
+  if (acc == null) {
+    acc = [];
+  }
   acc.push(item);
   return acc;
-}, [])));
+}, null)));
```
customcommander, line 52: Mmmh I think the change introduced in https://github.com/ramda/ramda/commit/9b5d8925f23f0f44a2e656bfd0cf23ff30031794 were possible breaking changes:

```none
[].reduce(() => 42, null);
//=> null
[].reduce(() => 42, []);
//=> []
```

So reverting this definitely fixes the bug? Your repro case doesn't have empty lists as far as I can tell so I don't see an obvious connection here unless I'm looking at this from the wrong angle?

jaawerth: The connection insofar as I saw it was that in [reduceBy](https://github.com/ramda/ramda/blob/master/source/reduceBy.js#L58)  - not so much a problem of passing an empty list, but the fact that the list accumulator is being re-cloned for each group in a way that's memoized and holds onto the original values.

Reverting the change makes the tests pass for `groupby`, but only for `groupBy`, though, so I suspect this may be ignoring other places where the same bug is showing up, which may need further digging.

Issue #3041, opened 2020-06-15
Why `uniqWith` doesn't work like `groupWith`?


ivanfilhoz:
PR #2016 changes `groupWith` behavior to pass only adjacent elements to the comparison function. May we have something similar on `uniqWith`?

Here, I'm manually implementing it like this:
```
const uniqByStreak = compose(map(head), groupWith)
```

Bradcomp:
Hi @ivanfilhoz,

`uniqWith` is based on `uniq`, which removes all duplicates from a list. The `With` suffix simply means it takes a predicate function to determine equality between elements. There really isn't any other thing that links the `With` functions.

I think we would welcome a PR to add two similar functions that work in a similar way to `groupWith` but I don't know what we would call them.

CrossEye:
@ivanfilhoz: That looks like a useful function, and I would love to get a PR for it.  The name wouldn't work.  As @Bradcomp mentions, we distinguish `*With` and `*By` fairly consistently with the idea that `*By` takes a unary function that converts the element being checked into some useful format.  `*With` takes a binary function to describe how to act on two elements.  For instance `groupWith` takes two elements and reports whether they belong in the same group.  But `groupBy` takes a unary function from our input element to an object key.

`streakWith`?  (Or does that sound like an invitation for group public nudity?)

ivanfilhoz:
It's pretty useful indeed. I'm currently using it to simplify a pageview history section by striping out adjacent duplicated hits to the same page.

Also, I just checked and there are [dropRepeats](https://ramdajs.com/docs/#dropRepeats) and [dropRepeatsWith](https://ramdajs.com/docs/#dropRepeatsWith) already, and they seem to do exactly what I proposed here.

In this case, I guess we are only missing `dropRepeatsBy` which would take an unary predicate to determine the codomain like `(a → b) → [a] → [a]`.

Is that correct?

CrossEye:
That's funny.  I forgot entirely about those functions.  I'm not sure I've *ever* used them personally.

I don't make any sense out of `dropRepeatsBy`.  I have no idea how a unary predicate could tell us whether we're repeating the previous element.

ivanfilhoz:
> I don't make any sense out of `dropRepeatsBy`. I have no idea how a unary predicate could tell us whether we're repeating the previous element.

It should work the same way `dropRepeats` does, but instead of testing internally with `equals`, it uses `eqBy(predicate)`.

CrossEye:
[...]

PR #3239, opened 2022-02-19
Implement dropRepeatsBy (#3041)

ivanfilhoz:
see discussion [here](https://github.com/ramda/ramda/issues/3041)

### DIFF
--- a/source/dropRepeatsBy.js
+++ b/source/dropRepeatsBy.js
@@ -1,0 +1,30 @@
+import _curry2 from './internal/_curry2.js';
+import _dispatchable from './internal/_dispatchable.js';
+import _xdropRepeatsWith from './internal/_xdropRepeatsWith.js';
+import dropRepeatsWith from './dropRepeatsWith.js';
+import eqBy from './eqBy.js';
+
+
+/**
+ * Returns a new list without any consecutively repeating elements,
+ * based upon the value returned by applying the supplied function to
+ * each list element. [`R.equals`](#equals) is used to determine equality.
+ *
+ * Acts as a transducer if a transformer is given in list position.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig (a -> b) -> [a] -> [a]
+ * @param {Function} fn A function used to produce a value to use during comparisons.
+ * @param {Array} list The array to consider.
+ * @return {Array} `list` without repeating elements.
+ * @see R.transduce
+ * @example
+ *
+ *     R.dropRepeatsBy(Math.abs, [1, -1, -1, 2, 3, -4, 4, 2, 2]); //=> [1, 2, 3, -4, 2]
+ */
+var dropRepeatsBy = _curry2(
+  (fn, list) => _dispatchable([], _xdropRepeatsWith(eqBy(fn)), dropRepeatsWith(eqBy(fn)))(list)
+);
+export default dropRepeatsBy;
--- a/source/dropRepeatsBy.js
+++ b/source/dropRepeatsBy.js
@@ -24,7 +24,7 @@ import eqBy from './eqBy.js';
  *
  *     R.dropRepeatsBy(Math.abs, [1, -1, -1, 2, 3, -4, 4, 2, 2]); //=> [1, 2, 3, -4, 2]
  */
-var dropRepeatsBy = _curry2(
-  (fn, list) => _dispatchable([], _xdropRepeatsWith(eqBy(fn)), dropRepeatsWith(eqBy(fn)))(list)
-);
+var dropRepeatsBy = _curry2(function(fn, list) {
+  return _dispatchable([], _xdropRepeatsWith(eqBy(fn)), dropRepeatsWith(eqBy(fn)))(list);
+});
 export default dropRepeatsBy;
--- a/test/dropRepeatsBy.js
+++ b/test/dropRepeatsBy.js
@@ -10,26 +10,26 @@ describe('dropRepeatsBy', function() {
     {i: 1}, {i: -1}, {i: 1}, {i: 2}, {i: 3},
     {i: 3}, {i: -4}, {i: 4}, {i: 5}, {i: 3}
   ];
-  var pred = R.compose(R.evolve({ i: Math.abs }), R.dissoc('n'));
+  var fn = ({ i, n, ...rest }) => ({ i : Math.abs(i), ...rest });
 
   it('removes repeated elements based on predicate', function() {
-    eq(R.dropRepeatsBy(pred, objs2), objs);
-    eq(R.dropRepeatsBy(pred, objs), objs);
+    eq(R.dropRepeatsBy(fn, objs2), objs);
+    eq(R.dropRepeatsBy(fn, objs), objs);
   });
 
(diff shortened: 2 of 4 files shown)


test/dropRepeatsBy.js
```diff
@@ -0,0 +1,35 @@
...
+
+describe('dropRepeatsBy', function() {
+  var objs = [
+    {i: 1}, {i: 2}, {i: 3}, {i: -4}, {i: 5}, {i: 3}
+  ];
+  var objs2 = [
+    {i: 1}, {i: -1}, {i: 1}, {i: 2}, {i: 3},
+    {i: 3}, {i: -4}, {i: 4}, {i: 5}, {i: 3}
+  ];
+  var pred = R.compose(R.evolve({ i: Math.abs }), R.dissoc('n'));
```
CrossEye, line 13: I would rather write this out explicitly.  And I definitely wouldn't name it `pred`, as this is an object-transformation function, not a predicate.

Perhaps something like

```js
  var convert = ({i, n, ...rest}) => ({i : Math.abs(i), ...rest});
```

CrossEye (changes requested):
Except for a minor nit in the tests, I have no issues with how this is written.

But I'm still not convinced of the need for it.  Is there any good argument for including it besides the (important, but never crucial) consistency of a parallel for `dropRepeatsWith`.

Have you had common needs for such a function?

ivanfilhoz:
> Except for a minor nit in the tests, I have no issues with how this is written.
> 
> But I'm still not convinced of the need for it. Is there any good argument for including it besides the (important, but never crucial) consistency of a parallel for `dropRepeatsWith`.
> 
> Have you had common needs for such a function?

There certainly are real-world scenarios for this function. Last year, when I created #3041, my use case was the following: given a pageview history page, I needed to filter out duplicate visits from the same user, displaying only the latest one. IMHO this would be the simplest form:

```js
const visits = [
  {
    date: '2021-12-02',
    userId: 1
  },
  {
    date: '2021-12-01',
    userId: 1
  },
  {
    date: '2021-11-27',
    userId: 2
  },
  {
    date: '2021-11-25',
    userId: 3
  }
]

const uniqueVisits = R.dropRepeatsBy(R.prop('userId'))
console.log(uniqueVisits)

/*
Output:

[
  {
    date: '2021-12-02',
    userId: 1
  },
  {
    date: '2021-11-27',
    userId: 2
  },
  {
    date: '2021-11-25',
    userId: 3
  }
]
*/
```

test/dropRepeatsBy.js
ivanfilhoz: You're right, `pred` is not accurate naming. Changes done as requested.

source/dropRepeatsBy.js
```diff
@@ -0,0 +1,30 @@
...
+ * @param {Function} fn A function used to produce a value to use during comparisons.
+ * @param {Array} list The array to consider.
+ * @return {Array} `list` without repeating elements.
+ * @see R.transduce
+ * @example
+ *
+ *     R.dropRepeatsBy(Math.abs, [1, -1, -1, 2, 3, -4, 4, 2, 2]); //=> [1, 2, 3, -4, 2]
+ */
+var dropRepeatsBy = _curry2(
+  (fn, list) => _dispatchable([], _xdropRepeatsWith(eqBy(fn)), dropRepeatsWith(eqBy(fn)))(list)
```
customcommander, line 28: Minor: @CrossEye: do you think we should drop the arrow function until we rewrite the codebase in modern JS? Not bothered at all but thought I'd ask.

test/dropRepeatsBy.js
```diff
@@ -0,0 +1,35 @@
...
+  ];
+  var objs2 = [
+    {i: 1}, {i: -1}, {i: 1}, {i: 2}, {i: 3},
+    {i: 3}, {i: -4}, {i: 4}, {i: 5}, {i: 3}
+  ];
+  var fn = ({ i, n, ...rest }) => ({ i : Math.abs(i), ...rest });
+
+  it('removes repeated elements based on predicate', function() {
+    eq(R.dropRepeatsBy(fn, objs2), objs);
+    eq(R.dropRepeatsBy(fn, objs), objs);
```
customcommander, line 17: Do you think there could be a risk for a false positive if either `dropRepeatsBy` or `fn` mutates `objs`? In general I prefer keeping tests isolated to avoid collisions. Do you think test data could simply be inlined in each test?

source/dropRepeatsBy.js
CrossEye: We haven't generally worried about the style of the tests.  I think this is fine.

test/dropRepeatsBy.js
CrossEye: if `dropRepeatsBy` itself mutates the data, we've got serious problems.

But I agree that tests should test one thing, which usually means a single assertion, or perhaps a before and after assertion.

These two should be in separate tests.  Good catch.,

ivanfilhoz: `dropRepeatsBy` should never mutate the data. Why would it?

source/dropRepeatsBy.js
ivanfilhoz: This is not test, this is the implementation itself. So yes, I could change it to a regular function instead

test/dropRepeatsBy.js
ivanfilhoz: Btw the point on test isolation is valid. This test file is based on [https://github.com/ivanfilhoz/ramda/blob/feature/dropRepeatsBy/test/dropRepeatsWith.js](test/dropRepeatsWith.js). Should we fix it as well?

CrossEye: yes, we probably should.  We haven't been as consistent with test code as with the production code.  This is a minor nit, but yes, let's fix it in both places.

source/dropRepeatsBy.js
ivanfilhoz: Done.

test/dropRepeatsBy.js
ivanfilhoz: Okay, done. Please check new version. I think most of the risks have been properly addressed.

customcommander: > `dropRepeatsBy` should never mutate the data. Why would it?

Sorry for the confusion here. Of course it should not. I was just thinking about the case where it accidentally did. We sometimes miss cases of mutation during development or code review, so tests can be written in such a way to detect them. I was only asking whether such test could detect this problem as I wasn't convinced it was.

Thanks for the changes.

customcommander: I don't think my comment has been addressed but I'm happy to revisit some of these tests myself separately. So will resolve this conversation for now.

customcommander (approved):

CrossEye:
Thank you Ivan for your contribution, and for your patience!

CrossEye:
I hit "Close" rather than "Merge".  Oops!  Fixed now.

ivanfilhoz:
I've seen it LOL

Thanks for merging. I'm sure it'll be useful for the community.

PR #3244, opened 2022-02-23
docs: curry: add warning about default parameters. close #3121

### LINKED ISSUE #3121: Currying with default argument values short-circuit currying
```js
    var p = R.curry((x,k,c='koko',d)=>console.log(x,k,c,d))
    
    p('haha') // properly waiting for second argument
    p('haha','love') // shortcircuits and calls the full function.
    //>>> haha love koko undefined
````
Rather unexpected behavior.
I wish the docs at least would mention this.
I am not sure if this is at all fixable.

### PR DESCRIPTION
**Pull request by customcommander**

### CODE REVIEW

**Comment by adispring** [comment]
We can this link: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/length, to explain how curry/curryN calculate fn's length.

**Comment by customcommander** [comment]
> We can add this link: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/length, to explain how curry/curryN calculate fn's length.

Probably a good idea! Done.

**Review by adispring — DISMISSED** [review]

File: source/curry.js [hunk 1]
```diff
@@ -26,6 +26,25 @@ import curryN from './curryN.js';
...
+ * //=> function (waits for `b`)
+ *
+ * h(39)(1);
+ * //=> 42
+ *
+ * h(1)(2, 3);
+ * //=> 6
+ *
+ * h(1)(2)(3);
+ * //=> Error! (`3` is not a function!)
```
CrossEye, line 45: I'd rather something like `h (1) (2) (7)`, so there's no confusion about what `3` is not a functions.

CrossEye (changes requested):
I like this, with one minor nit.

source/curry.js
customcommander: Make sense. Will change.

adispring (approved):

CrossEye (approved):

PR #3246, opened 2022-02-26
fix(3245): fix R.dropLast with negative and zero param when used as transducer

### LINKED ISSUE #3245: Wrong behaviour in R.dropLast(0/negative) when used as transducer
Current behavior:

```
R.into([], R.dropLast(0), [1, 3, 5]) // [1, 3]
R.into([], R.dropLast(-1), [1, 3, 5]) // throw RangeError: Invalid array length
```

Expected behavior:

```
R.into([], R.dropLast(0), [1, 3, 5]) //  [1, 3, 5]
R.into([], R.dropLast(-1), [1, 3, 5]) //  [1, 3, 5]
```

xgbuils:
It fixes #3245

### DIFF
--- a/source/internal/_xdropLast.js
+++ b/source/internal/_xdropLast.js
@@ -3,6 +3,9 @@ import _xfBase from './_xfBase.js';
 
 
 function XDropLast(n, xf) {
+  if (n <= 0) {
+    return xf;
+  }
   this.xf = xf;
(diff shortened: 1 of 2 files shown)


CrossEye:
I really hate supporting negative indices in general, but we should be consistent in it.

:herb:

Issue #3247, opened 2022-02-27
refactor for avoiding unnecessary checks when _reduce parameter is not a transformer.


xgbuils:
I was reading the things commented in the PR #2762 and I also agree with the idea that `_reduce` is too complex and maybe _reduced behavior should be relegated to the transducers and no more. However I think that the solution is adding more complexity and more checks at running time. It's true that we are gaining performance with no transducer way of work. But we are paying a cost of a new check when we work with tranducers.

My proposal is having a `_reduce` function that works with reducers and a `_xReduce` function that works with transformers. Maybe we will have a kind of code duplication. But I think it's better this kind of duplication over adding checks at runtime. Then, I'm gonna use `_reduce` or `_xReduce` in the ramda functions depending on the case. My idea is breaking nothing. It's just keeping things simple and having a good enabler if we really want to avoid to manage reduced values in some of _dispatchable fallbacks.

CrossEye:
I can't recall the history now (or bother trying to dig it up) so I don't know if `reduced` predated transducers or was added for that case, but it does add some nasty complexity.

Of course if I were to design it today, I would, ignoring FantasyLand compatibility, write something more like this <sup>1</sup>:

```js
const None = Symbol ()

const foldl = (fn) => (init) => ([x = None, ...xs]) =>
  x == None
    ? init
    : fn (init, x, (r) => foldl (fn) (r) (xs), (r) => r)


const total = foldl ((a, b, next, done) => next (a + b)) (0) 

total ([1, 2, 3, 4, 5, 6, 7]) //=> 28

const sumBelow = (t) => 
  foldl ((a, b, next, done) => a + b < t ? next (a + b) : done (a)) (0)

sumBelow (20) ([1, 2, 3, 4, 5, 6, 7]) //=> 15
```

But I'm afraid that ship sailed long ago.

So this sounds like a good idea.  But we do need to keep in mind that removing `reduced` from the plain reducers is a breaking change.


----------
<sup>1</sup> Or perhaps like this:
> ```js
> const foldl = (fn) => (init) => (xs) =>
>   xs .length < 1
>     ? init
>     : fn (init, xs [0] , (r) => foldl (fn) (r) (xs .slice (1)), (r) => r)
> ```

xgbuils:
Hi @CrossEye 

[...]

PR #3248, opened 2022-02-27
split _reduce to _xReduce(for transformers) & _reduce(for reducers)

xgbuils:
Following the approach about what I said in #3247, I also extracted _arrayReduce and _xArrayReduce to be used in places where function are only working with array lists. In this way, we are not executing additional type checks in cases are not needed.

### DIFF
--- a/source/filter.js
+++ b/source/filter.js
@@ -1,8 +1,8 @@
+import _arrayReduce from './internal/_arrayReduce.js';
 import _curry2 from './internal/_curry2.js';
 import _dispatchable from './internal/_dispatchable.js';
 import _filter from './internal/_filter.js';
 import _isObject from './internal/_isObject.js';
-import _reduce from './internal/_reduce.js';
 import _xfilter from './internal/_xfilter.js';
 import keys from './keys.js';
 
@@ -38,7 +38,7 @@ import keys from './keys.js';
 var filter = _curry2(_dispatchable(['fantasy-land/filter', 'filter'], _xfilter, function(pred, filterable) {
   return (
     _isObject(filterable) ?
-      _reduce(function(acc, key) {
+      _arrayReduce(function(acc, key) {
         if (pred(filterable[key])) {
           acc[key] = filterable[key];
         }
--- a/source/internal/_arrayReduce.js
+++ b/source/internal/_arrayReduce.js
@@ -1,0 +1,9 @@
+export default function _arrayReduce(reducer, acc, list) {
+  var index = 0;
+  var length = list.length;
+  while (index < length) {
+    acc = reducer(acc, list[index]);
+    index += 1;
+  }
+  return acc;
+}
--- a/source/internal/_createReduce.js
+++ b/source/internal/_createReduce.js
@@ -1,0 +1,29 @@
+import _isArrayLike from './_isArrayLike.js';
+
+
+var symIterator = (typeof Symbol !== 'undefined') ? Symbol.iterator : '@@iterator';
+
+export default function _createReduce(arrayReduce, methodReduce, iterableReduce) {
+  return function _reduce(xf, acc, list) {
+    if (_isArrayLike(list)) {
+      return arrayReduce(xf, acc, list);
+    }
+    if (list == null) {
+      return acc;
+    }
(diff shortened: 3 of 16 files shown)


xgbuils:
Notice that this PR doesn't colide with #2767. #2767 is a behavior change. This PR it's just a refactor.

If we merge this PR and we want to apply the behavior of #2767 we just need to remove the lines:

``` javascript
if (typeof list['fantasy-land/reduce'] === 'function') {
  return list['fantasy-land/reduce'](fn, acc);
}
```
in `_createReducer.js`

and adding in `_reduce.js`:
``` javascript
var _baseReduce = _createReduce(_arrayReduce, _methodReduce, _iterableReduce);
export default function _reduce(reducer, acc, list) {
  if (typeof list['fantasy-land/reduce'] === 'function') {
    return list['fantasy-land/reduce'](fn, acc);
  }
  return _baseReduce(reducer, acc, list);
}
```

xgbuils:
Hi! Some benchmarks changin map.bench.js

``` javascript
var obj = {a: 1, b: 2};
...
    'map(sq, obj)': function() {
      map(sq, obj);
    },
```
Original:

xgbuils:
Hi! Some benchmarks adding benchs for objects in map.bench.js

``` javascript
var obj = {a: 1, b: 2};
...
    'map(sq, obj)': function() {
      map(sq, obj);
    },
```
Original:
```
│ map                    │ Hertz                  │ Margin of Error        │
│ map(sq, obj)           │ 956,837                │ 4.81%                  │
```
After change:
```
│ map                    │ Hertz                  │ Margin of Error        │
│ map(sq, obj)           │ 2,099,464              │ 6.11%                  │
```

Using more keys implies less difference in gain of performance:
```
var nums = [8, 2, 85, 2, 34, 3, 23, 247, 57, 8, 0, 6, 5, 46, 54, 643];
var obj = {...nums};
...
    'map(sq, obj)': function() {
      map(sq, obj);
    },
```
Original:
```
│ map                    │ Hertz                  │ Margin of Error        │
│ map(sq, obj)           │ 571,880                │ 0.41%                  │
```
After change:
```
│ map                    │ Hertz                  │ Margin of Error        │
│ map(sq, obj)           │ 778,917                │ 0.42%                  │
```

source/ap.js
```diff
@@ -38,7 +38,11 @@ var ap = _curry2(function ap(applyF, applyX) {
         ? applyF.ap(applyX)
         : typeof applyF === 'function'
           ? function(x) { return applyF(x)(applyX(x)); }
-          : _reduce(function(acc, f) { return _concat(acc, map(f, applyX)); }, [], applyF)
+          : _reduce(
+            function(acc, f) { return _concat(acc, map(f, applyX)); },
+            [],
+            applyF
+          )
```
CrossEye, line 45: Post `1.0`, I would like to do some auto-formatting, using Prettier or something, and from then on simply ignore such things.  The trouble, of course, is that auto-formatters are mostly designed around OOP/imperative styles, and those don't always transfer well to FP/declarative code.  Still, I would love to simply never have to make such judgement calls.

CrossEye (approved):
This looks great to me.  In the end the code is cleaner, *and* it's more performant.  It's hard to beat that!

:herb:

source/ap.js
xgbuils: My bad. I'm gonna revert this formatting part.

xgbuils:
internal `_reduce` implementation changed and now only works with standard reducers. The new internal `_xReduce` function only works with transformers. Then, some `_reduce` usages are hidden and difficult to review because the files didn't change but the implempementation of `_reduce` did. I'm gonna list the usages of _reduce:

- [ap.js](https://github.com/xgbuils/ramda/blob/refactor-3247/source/ap.js#L41)
- [collectBy.js](https://github.com/xgbuils/ramda/blob/refactor-3247/source/collectBy.js#L36)
- [count.js](https://github.com/xgbuils/ramda/blob/refactor-3247/source/count.js)
- [pipeWith.js](https://github.com/xgbuils/ramda/blob/refactor-3247/source/pipeWith.js#L42)

CrossEye (approved):

customcommander (approved):

PR #3250, opened 2022-03-04
feat(propIncludes): introduce propIncludes

jaxtisapia:
Add a relation function `(propIncludes)` that can be used to determine whether an object has a property which is found in a list.

This is the `includes` version of [propEq](https://ramdajs.com/docs/#propEq)

```
// Before
pipe(filter(  (product) => includes(product?.id, productIds))  )



// After
pipe(filter(  propIncludes("id", productIds ))  )
```

### DIFF
--- a/source/index.js
+++ b/source/index.js
@@ -181,6 +181,7 @@ export { default as project } from './project.js';
 export { default as promap } from './promap.js';
 export { default as prop } from './prop.js';
 export { default as propEq } from './propEq.js';
+export { default as propIncludes } from './propIncludes.js';
 export { default as propIs } from './propIs.js';
 export { default as propOr } from './propOr.js';
 export { default as propSatisfies } from './propSatisfies.js';
--- a/source/propIncludes.js
+++ b/source/propIncludes.js
@@ -1,0 +1,33 @@
+import _curry3 from './internal/_curry3.js';
+import prop from './prop.js';
+import includes from './includes.js';
+
+
+/**
+ * Returns `true` if the specified object property is found in the list supplied,
+ * in [`R.includes`](#includes) terms; `false` otherwise.
+ *
+ * @func
+ * @memberOf R
+ * @category Relation
+ * @sig String -> [a] -> Object -> Boolean
+ * @param {String} name
+ * @param {Array} list
+ * @param {*} obj
+ * @return {Boolean}
+ * @see R.propEq, R.propSatisfies, R.includes
+ * @example
+ *
+ *      const kwabena = {name: 'Kwabena', country: "ghana"};
+ *      const naadei = {name: 'Naadei', country: "ghana"};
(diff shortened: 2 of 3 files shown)


CrossEye (changes requested):
I have mixed feelings about this function.  It is clearly useful on occasion.

I find a simpler example more compelling than the african/european countries ones, perhaps something like

```js
const xs = [
  {id: 1, category: 'a'}, 
  {id: 2, category: 'b'}, 
  {id: 3, category: 'a'}, 
  {id: 4, category: 'c'}, 
  {id: 5, category: 'c'}, 
  {id: 6, category: 'a'}, 
  {id: 7, category: 'd'}, 
  {id: 8, category: 'b'}, 
]

filter (propIncludes ('category') (['a', 'd'])) (xs) //=>
// [
//   {id: 1, category: 'a'}, 
//   {id: 3, category: 'a'}, 
//   {id: 6, category: 'a'}, 
//   {id: 7, category: 'd'}, 
// ]
```

The name reads badly.  Unlike, say, `subtract`, `includes` is named in a way that emphasizes partial application.  `includes (42)` is a function which accepts a list and reports whether that list includes `42`.  (`subtract (42)` by contrast is a function which accepts a value and subtracts that value from 42, to the confusion of many.)  But in this function, `propIncludes` reads to me as if it were meant to be used like this:

```js
propIncludes ('category', 42) ([foo: 'abc', category: [21, 42, 63]}) //=> true
propIncludes ('category', 42) ([foo: 'def', category: [20, 40, 60]}) //=> false
```

So if we're going to include this function, I'd like to consider alternate names.  `propIsOneOf`, `propIsIncludedIn`, `propIsIn`....

But I'm not certain if we want to include it at all.  It's simple, and well-written, but we introduced `propSatisfies` in order to capture anything beyond `propEq` and `propIs` -- and even the latter doesn't seem important enough to warrant its own function.

So, before making any of the suggested changes, can you make an argument for why this is an important enough function that we can't simply recommend, say,

```js
filter (propSatisfies (includes (__, ['a', 'd']), 'category')) (xs)
```

(which would be made better by a flipped version of `includes`, say `isOneOf`:

```js
filter (propSatisfies (isOneOf ( ['a', 'd']), 'category')) (xs)
```

?

customcommander:
This really is a well-written PR and I particularly like the efforts that went into the test cases but I have to say that I share @CrossEye observation:

```
filter (propSatisfies (includes (__, ['a', 'd']), 'category')) (xs)
```

Seems like what you're after. I think this should be an addition to the cookbook instead.

jaxtisapia:
Thanks for the review, guys.

I'll go ahead and close this PR

CrossEye:
@jaxtisapia:

Again, thank you for participating, and please do add this function to [the Cookbook][cb]

Issue #3258, opened 2022-03-22
Tring to clone() object with no prototype gives error u.hasOwnProperty is not a function


denishowew:
`clone()` can't handle _objects with no prototype_.  When trying to clone `value`, it does:

```
for (var key in value) {
  if (value.hasOwnProperty(key)) ...
```

which is fine for normal objects that inherit the `hasOwnProperty()` method.

However, objects created using `Object.create(null)` have no prototype and will trip up [unwary code](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create).

clone() blows up with error `u.hasOwnProperty is not a function`.

[Demo of the bug](https://ramdajs.com/repl/?v=0.28.0#?const%20o%20%3D%20Object.create%28null%29%3B%0Ao.a%20%3D%2042%3B%0AR.clone%28o%29%3B%0A):

```
const o = Object.create(null);
o.a = 42;
R.clone(o);

```

CrossEye:
Getting back to Ramda after a short time away.

This is definitely a problem.  Are you interested in creating a PR to fix it?, @denishowew ?

denishowew:
Sorry, no, I hate Ramda. :-)

semmel:
I would volunteer

thefourtheye:
> > Are you interested in creating a PR to fix it?,
> 
> I would volunteer

@semmel My apologies, I came in late and created a patch for this. If you already have been working on a fix, please let me know. I do not mind closing #3261.

semmel:
> I do not mind closing https://github.com/ramda/ramda/pull/3261.

Not at all! All I care about is that there is a patch.
Great work in #3261 and thank you @thefourtheye for making our work/lives better!

PR #3261, opened 2022-03-30
fix(3258): handle no prototype objects in R.clone

thefourtheye:
Fixes: https://github.com/ramda/ramda/issues/3258

----
cc: @CrossEye @customcommander @denishowew @semmel

### DIFF
--- a/source/internal/_clone.js
+++ b/source/internal/_clone.js
@@ -30,8 +30,7 @@ export default function _clone(value, deep, map) {
     map.set(value, copiedValue);
 
     for (var key in value) {
-      // If there is no prototype, the property must be its own
-      if (!_hasPrototype(value) || value.hasOwnProperty(key)) {
+      if (Object.prototype.hasOwnProperty.call(value, key)) {
         copiedValue[key] = deep ? _clone(value[key], true, map) : value[key];
       }
     }
@@ -64,10 +63,6 @@ function _isPrimitive(param) {
   return param == null || (type != 'object' && type != 'function');
 }
 
-function _hasPrototype(param) {
-  return Object.getPrototypeOf(param) != null;
-}
-
 function _ObjectMap() {
   this.map = {};
   this.length = 0;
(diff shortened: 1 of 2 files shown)


source/internal/_clone.js
```diff
@@ -30,7 +30,8 @@ export default function _clone(value, deep, map) {
     map.set(value, copiedValue);
 
     for (var key in value) {
-      if (value.hasOwnProperty(key)) {
+      // If there is no prototype, the property must be its own
+      if (!_hasPrototype(value) || value.hasOwnProperty(key)) {
```
CrossEye, line 34: This is not precisely an objection.  I'm curious why you chose to implement it this way rather than as in this (untested) snippet:

```js
var hasOwnProperty = Object .prototype .hasOwnProperty;
/* ... */

for (var key in value) {
    if (hasOwnProperty .call (value, key) {
/* ... */
```

source/internal/_clone.js
```diff
@@ -83,7 +88,11 @@ _ObjectMap.prototype.set = function(key, value) {
 _ObjectMap.prototype.hash = function(key) {
   let hashedKey = [];
   for (var value in key) {
-    hashedKey.push(key[value]);
+    if (_hasPrototype(key[value])) {
+      hashedKey.push(key[value]);
+    } else {
+      hashedKey.push(Object.prototype.toString.call(key[value]));
+    }
```
CrossEye, line 95: What is the use-case that requires this block?  I'm not following something.

CrossEye (commented):
I raised two questions inline.

source/internal/_clone.js
thefourtheye: @CrossEye All the values are added in `hashedKey` and `hashedKey.join` tries to convert all the elements by calling `.toString` on it, if necessary. In our case, the object without prototype doesn't have `toString`. So, handling that case with `Object.prototype.toString.call`. I was not sure about the implications of doing `Object.prototype.toString.call` on all the elements of the `hashedKey`. That's why it is done this way.

test/clone.js
```diff
@@ -299,6 +299,23 @@ describe('deep clone edge cases', function() {
 
 });
 
+describe('clone objects with no prototypes', function() {
+  it('object with no prototype', function() {
```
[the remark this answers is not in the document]
thefourtheye: @customcommander You are correct. The second one is like a super set. Let me drop this.

source/internal/_clone.js
thefourtheye: @CrossEye When this patch was created, I was a bit concerned about the readability. Hence the comment as well. I am all up for doing this the concise way. Let me update this.

CrossEye (approved):
LGTM

:herb:

customcommander (approved):

David-Hobson:
Came across this issue recently and am glad to see that it's fixed. Do we know when this going to get released and be a part of the latest Ramda package on npm?

PR #3262, opened 2022-03-31
refactor(collectBy): use groupBy for grouping

ycmjason:


customcommander:
Hello and thank you for your contribution. Not sure if you've seen [this comment](https://github.com/ramda/ramda/pull/2715#discussion_r436256490) but it may be worth reading it if you (or other reviewers) didn't. I personally don't have an issue with this refactor as long as it doesn't incur a performance penalty.

source/collectBy.js
```diff
@@ -33,14 +33,6 @@ import _reduce from './internal/_reduce.js';
...
-  var group = _reduce(function(o, x) {
-    var tag = fn(x);
-    if (o[tag] === undefined) { o[tag] = []; }
-    o[tag].push(x);
-    return o;
-  }, {}, list);
-  var newList = [];
-  for (var tag in group) { newList.push(group[tag]); }
-  return newList;
+  return Object.values(groupBy(fn, list));
```
customcommander, line 36: @CrossEye: IE does not have `Object.values`. Do we care about IE compatibility at this point?

ycmjason:
I didn't see that comment. Thanks for linking. I am not sure about the performance cost of this refactor. Feel free to close this if it is not something desirable. 😄

customcommander:
Let's wait for some more comments on this :)

source/collectBy.js
CrossEye: Let's stick with our ES3 compatibility through 1.0, then go wild with modern ES6 stuff post-1.0

customcommander: @ycmjason I think if you change `Object.values` with `R.values` we should be able to reconsider this.

ycmjason: @customcommander done! 😄

customcommander (approved):

CrossEye:
Let's leave this for the 2.0 refactor.  At that point we'll be rewriting all the functions anyway using modern JS features, and we will probably have to do regular perf testing.

It's not that I think this is a bad idea, just that the timing is not right for it.  When we do go to more modern JS, we can look at where the simplicity of function layering outweighs a minor perf loss.

customcommander:
@ycmjason 

We'll defer the merge after our `v1.0` release (which we'll eventually happen in 2022). At which point we'll start working on our `v2.0` which will include a modern rewrite and be the best opportunity for this kind of refactor.

PR #3263, opened 2022-03-31
link collectBy and groupBy

ycmjason:
`collectBy` already links to `groupBy`. I thought it would make sense the opposite way too.


CrossEye:
Thank you @ycmjason!

PR #3266, opened 2022-04-04
another code refactoring

psparsa:

### DIFF
--- a/source/times.js
+++ b/source/times.js
@@ -32,7 +32,7 @@ var times = _curry2(function times(fn, n) {
   }
 
   list = [];
-  for (let x = 0; x < len; x += 1) {
+  for (var x = 0; x < len; x += 1) {
     list[x] = fn(x);
   }
 
--- a/source/times.js
+++ b/source/times.js
@@ -1,6 +1,5 @@
 import _curry2 from './internal/_curry2.js';
 
-
 /**
  * Calls an input function `n` times, returning an array containing the results
  * of those function calls.
@@ -26,17 +25,18 @@ import _curry2 from './internal/_curry2.js';
  */
 var times = _curry2(function times(fn, n) {
   var len = Number(n);
-  var idx = 0;
   var list;
 
   if (len < 0 || isNaN(len)) {
     throw new RangeError('n must be a non-negative number');
   }
-  list = new Array(len);
-  while (idx < len) {
-    list[idx] = fn(idx);
-    idx += 1;
+
+  list = [];
+  for (let x = 0; x < len; x += 1) {
+    list[x] = fn(x);
   }
+
   return list;
 });
+
 export default times;
--- a/source/times.js
+++ b/source/times.js
@@ -33,7 +33,7 @@ var times = _curry2(function times(fn, n) {
 
   list = [];
   for (var x = 0; x < len; x += 1) {
-    list[x] = fn(x);
+    list.push(fn(x));
   }
 
   return list;
--- a/source/times.js
+++ b/source/times.js
@@ -1,5 +1,6 @@
 import _curry2 from './internal/_curry2.js';
 
+
 /**
  * Calls an input function `n` times, returning an array containing the results
  * of those function calls.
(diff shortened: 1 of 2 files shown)


source/times.js
```diff
@@ -26,17 +25,18 @@ import _curry2 from './internal/_curry2.js';
...
   if (len < 0 || isNaN(len)) {
     throw new RangeError('n must be a non-negative number');
   }
-  list = new Array(len);
-  while (idx < len) {
-    list[idx] = fn(idx);
-    idx += 1;
+
+  list = [];
+  for (let x = 0; x < len; x += 1) {
```
customcommander, line 35: We have plans to rewrite the codebase with modern ES syntax but until then we also want to keep as much IE support as possible. Therefore would you mind changing `let` to `var` please?

source/times.js
```diff
@@ -26,17 +25,18 @@ import _curry2 from './internal/_curry2.js';
...
+    list[x] = fn(x);
```
customcommander, line 36: I _think_ `.push` would be faster (on paper) and would also avoid potential weird case if the user supplied a float as `n`.

psparsa:
thank u for ur time, i made a couple of changes...

source/times.js
psparsa: umm, why `push` would be faster?

customcommander (dismissed):
Thank you.

source/times.js
customcommander: I don't know why it is faster just that it seems to be the case (or used to be) in some browsers. Most likely irrelevant these days though but as it allows you to avoid managing the indexes (and avoid sparse arrays) _and_ avoid issues with non-integer indexes I see little reason not to use it.

source/times.js
```diff
@@ -26,17 +25,18 @@ import _curry2 from './internal/_curry2.js';
...
   var list;
 
   if (len < 0 || isNaN(len)) {
     throw new RangeError('n must be a non-negative number');
   }
-  list = new Array(len);
-  while (idx < len) {
-    list[idx] = fn(idx);
-    idx += 1;
+
```
adispring, line 33: Except more readable, what's the benefit for changing `while` loop to `for` loop.

According to a [historical reason](https://github.com/ramda/ramda/pull/2744), there are almost no `for` loops in ramda  repository.

adispring, line 33: Can you add some benchmark to explain how much benefits will gain from this optimization?

source/times.js
psparsa: ++

source/times.js
```diff
@@ -26,17 +25,18 @@ import _curry2 from './internal/_curry2.js';
  */
 var times = _curry2(function times(fn, n) {
   var len = Number(n);
-  var idx = 0;
   var list;
 
   if (len < 0 || isNaN(len)) {
```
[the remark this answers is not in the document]
psparsa: i think both ideas are good, but i agree more with the first one btw (removing this validation)

source/times.js
psparsa: take a look at [this](https://blog.bitsrc.io/measuring-performance-of-different-javascript-loop-types-c0e9b1d193ed) article...
I misunderstood apparently, there doesn't seem to be much difference 🤔
should i revert that while loop instead of `for` now?

source/times.js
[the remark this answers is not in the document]
psparsa: removing this validation may break the programs after upgrading ramda if their flow relies on this `RangeError`.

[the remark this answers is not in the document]
customcommander: Fair enough. Let's review this another time.

source/times.js
CrossEye: Let's go back the the `while` loop, for the reasons [already mentioned][co].  If nothing else, consistent style is always a plus!


  [co]: https://github.com/ramda/ramda/pull/2744

CrossEye (changes requested):

customcommander (approved):

CrossEye (approved):

PR #3269, opened 2022-04-11
Add/modify tests which can act as a transducer

zydmayday:
A really small change.

Since we are testing `intoArray`, so we should add that in description just like the others.

### DIFF
--- a/test/any.js
+++ b/test/any.js
@@ -56,4 +56,11 @@ describe('any', function() {
     });
   });
 
+  it('can act as a transducer', function() {
+    var input = [2, 4, 6, 8, 10];
+    var expected = [false];
+    eq(R.into([], R.any(odd), input), expected);
+    eq(R.transduce(R.any(odd), R.flip(R.append), [], input), expected);
+  });
+
 });
--- a/test/all.js
+++ b/test/all.js
@@ -52,4 +52,11 @@ describe('all', function() {
     });
   });
 
+  it('can act as a transducer', function() {
+    var input = [2, 4, 6, 8, 9, 10];
+    var expected = [false];
+    eq(R.into([], R.all(even), input), expected);
+    eq(R.transduce(R.all(even), R.flip(R.append), [], input), expected);
+  });
+
 });
--- a/test/aperture.js
+++ b/test/aperture.js
@@ -17,7 +17,9 @@ describe('aperture', function() {
   });
 
   it('can act as a transducer', function() {
-    eq(R.into([], R.aperture(2), sevenLs), [[1, 2], [2, 3], [3, 4], [4, 5], [5, 6], [6, 7]]);
+    var expected = [[1, 2], [2, 3], [3, 4], [4, 5], [5, 6], [6, 7]];
+    eq(R.into([], R.aperture(2), sevenLs), expected);
+    eq(R.transduce(R.aperture(2), R.flip(R.append), [], sevenLs), expected);
   });
 
 });
--- a/test/countBy.js
+++ b/test/countBy.js
@@ -53,9 +53,11 @@ describe('countBy', function() {
     var transducer = R.compose(
       R.countBy(R.prop('genre')),
       R.map(R.adjust(1, R.toString)));
-    eq(R.into({}, transducer, albums), {
+    var expected = {
       Baroque: '2', Rock: '2', Jazz: '2', Romantic: '1', Metal: '1', Modern: '1', Broadway: '1', Folk: '1', Classical: '1'
-    });
+    };
+    eq(R.into({}, transducer, albums), expected);
+    eq(R.transduce(transducer, (result, input) => {result[input[0]] = input[1]; return result;}, {}, albums), expected);
   });
 
 });
--- a/test/dropLast.js
+++ b/test/dropLast.js
@@ -32,10 +32,10 @@ describe('dropLast', function() {
 
   it('can act as a transducer', function() {
     var dropLast2 = R.dropLast(2);
-    eq(R.into([], dropLast2, [1, 3, 5, 7, 9, 1, 2]), [1, 3, 5, 7, 9]);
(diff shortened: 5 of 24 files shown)


test/any.js
```diff
@@ -21,7 +21,7 @@ describe('any', function() {
     eq(intoArray(R.any(odd), [2, 4, 6, 8, 10, 11, 12]), [true]);
   });
 
-  it('returns false if all elements fails to satisfy the predicate', function() {
+  it('returns false into array if all elements fails to satisfy the predicate', function() {
```
customcommander, line 24: @CrossEye: I'm not sure we even need those two last tests. Unless that was testing the transducer nature of `R.any` somehow but I don't see how. It seems like this is testing `R.into` more than it is testing `R.any`.

CrossEye: Agreed.  These test seem superfluous.

zydmayday: Just to clear my understanding,
If that is true, you can see there are a lot of places testing like this.
Such as inside of all / find / map, etc.

Maybe we should move all of such tests into `into`?

zydmayday: Thanks!
I removed all into related tests,
and add at least one single test for each functions(any/all/find/findIndex/findLast/findLastIndex/chain/map/none).

The only one I kept is [this in chain](https://github.com/ramda/ramda/blob/master/test/chain.js#L59-L64),
I don't know whether I should remove it or not.

Kindly have a check.

customcommander (dismissed):
Thank you very much!

CrossEye:
Now that I see the full extent of this, I'm less certain that we don't want such tests.  I think I'd like to see a single test for each of these functions that demonstrates that it can act as a transducer.  The transducer-related code can have many more, but we wouldn't expect it to know about every function that is a transducer.

I'd love to hear if there are counterarguments, but as I look at them again, I'm not thrilled with removing these tests.  I would like to clean them up, and only do one single-assertion test for either.  We're not testing transduction, but we're also not retesting the basic behavior of the function, only proving that the function can act as a transducer when required.

zydmayday:
> Now that I see the full extent of this, I'm less certain that we don't want such tests. I think I'd like to see a single test for each of these functions that demonstrates that it can act as a transducer. The transducer-related code can have many more, but we wouldn't expect it to know about every function that is a transducer.
> 
> I'd love to hear if there are counterarguments, but as I look at them again, I'm not thrilled with removing these tests. I would like to clean them up, and only do one single-assertion test for either. We're not testing transduction, but we're also not retesting the basic behavior of the function, only proving that the function can act as a transducer when required.

@CrossEye 
Thanks!
That sounds reasonable.
So I plan to move [the new code added in `into`](https://github.com/ramda/ramda/pull/3269/files#diff-75721e8042ac6c565411b677173c9bdc99d9a3298f7f134bcf0c726c78ea5fa7R18-R25),
back to each function, but with just one single test for proving they can act as a transducer.
Is that undertanding correct?

(but how do we treat tests in `into`? should we still keep them?

customcommander:
@CrossEye: from what I can see these tests simply moved to the tests for `into`. 

This only indirectly tests the transducer nature of `any`:

```javascript
into([], any(isEven), [1,2,3]);
//> [true]
```

Because I suspect that `into` composes its stepper function with the function returned by `any(isEven)`. 

> Acts as a transducer if a transformer is given in list position.

If we wanted to emphasise the transducer nature in tests then I'd do something like that: (actually composing with other known transformers)

```javascript
into([], compose(any(isEven), map(String)), [1,2,3])
//=> ["true"]
```

What if `map`, `filter`, etc. only focused on their own things and have a separate single test file that would make sure all transformers work well together? I think this PR is a step towards that but only needs to start composing some of them.

CrossEye:
@zydmayday:

> So I plan to move [the new code added in into](https://github.com/ramda/ramda/pull/3269/files#diff-75721e8042ac6c565411b677173c9bdc99d9a3298f7f134bcf0c726c78ea5fa7R18-R25),
 back to each function, but with just one single test for proving they can act as a transducer.
> Is that undertanding correct?

That's what I was thinking.  But it looks like @customcommander disagrees.

It would probably be better to use `transduce` directly rather than `into` for these tests, as it makes for a more obvious assertion.  But that's taking this PR far beyond, "Fix test method name..."

@customcommander:

I was thinking of this as an automated test for what we document as "Acts as a transducer if a transformer is given in list position."  We don't need to care here precisely how it acts, only that it interacts properly in the transducer environment.  It's not quite as simple as "doesn't throw an error when..." but it isn't far off.

This also makes me notice that the H-M signatures for `transduce` and `into` are woefully inadequate.

I'd suggest that we go back to the simpler version of this PR, and raise the usefulness of these transducer tests as a separate issue/PR.

Anyone object?

zydmayday:
@CrossEye @customcommander 
> I'd suggest that we go back to the simpler version of this PR, and raise the usefulness of these transducer tests as a separate issue/PR.

Yes I agree. Since that(complex fix) is not the original purpose of this PR.

And, I checked document again and search the keyword "Acts as a transducer if a transformer is given in list position.",
actually there are 25 methods can act as a transducer. 
So refer the existing [tests in transduce](https://github.com/ramda/ramda/blob/master/test/transduce.js#L40-L49),
we can either expand above test to support all methods, for a better coverage.
(because we are using `compose` in there, so should also satisfy @customcommander requirement),
or just a new one somewhere else to explicitly test the "act as a transducer" behaviour.

In my personal opinion,
from document, only `into` and `transduce` use transducer as a parameter,
so I am also Okay with test 'duplicately' in both these 2 methods.
(actually this is what we are doing now.

customcommander:
Ok works for me.

customcommander (approved):
Definitely an improvement. Thanks!

CrossEye:
@zydmayday: Thank you very much for persevering as your little change grew into a much bigger deal!

:herb:

PR #3272, opened 2022-04-19
refactor: `of` now works with Applicatives

### LINKED ISSUE #3267: Refactor R.of so that it works with Applicatives (BREAKING CHANGE)
Currently `of` can only wrap a thing into an array

```javascript
of(42) //> [42]
```

However "of" is also a well know feature of Applicatives:

```javascript
Maybe.of(42) //> Just(42)
```

## Proposal

@Bradcomp made this [suggestion](https://github.com/ramda/ramda/issues/2502#issuecomment-379830202):

```javascript
of(Array, 42) //> [42]
of(Maybe, 42) //> Just(42)
of(Either, 42) //> Right(42)
```

Which makes a lot of sense and gained popular support so far.

Related issue(s): #2502

customcommander:
This change addresses the issue of `of` not being able to
work with Applicatives.

Fixes #3267

BREAKING CHANGE: `of` is now a binary function with a constructor as its first argument.

**Acknowledgement**

This PR is almost an 100% copy/paste of #2687.

Notable differences:

1. Fix parameter name in JSDoc annotations
2. Update example
3. Ignore one set of changes in tests as other changes were committed since then


source/of.js
```diff
@@ -1,9 +1,10 @@
...
-import _of from './internal/_of.js';
-
+import _curry2 from './internal/_curry2.js';
 
 /**
- * Returns a singleton array containing the value provided.
+ * Given a constructor and a value, returns a new instance of that constructor
+ * containing the value.
+ *
+ * Dispatches to the `of` method of the first argument, if present.
```
davidchambers, line 7: Should we mention `fantasy-land/of`?

davidchambers (approved):
Very nice!

adispring (approved):

source/of.js
customcommander: Probably a good idea :) Will change.

customcommander: Done. Not too sure about the description now. Let me know if you have some other suggestions.

CrossEye (approved):
This is great!

We'll have to prominently feature the backward incompatibility in our release notes.

adispring (approved):

customcommander:
@CrossEye Any objection to merging this now? We could create a draft v1 release notes now if you fear we would miss out on important breaking changes such as this one.

CrossEye:
Merged.  We usually hope that two active maintainers (besides the author, when appropriate) will approve, and then anyone can merge.

PR #3277, opened 2022-04-26
use latest DT ramda .d.ts

### LINKED ISSUE #3275: starting point for .d.ts
quoting @somebody1234

> not too sure about using such an old version of the types, there's a huge amount of work on the types that will need to be carried over... it's probably easier/better to start with the latest version and gradually remove ts-toolbelt

My thinking here is starting from a version with no dependencies, we can incrementally improve and publish at any point.

If we start with dependencies, it could be difficult to remove them and we'd get ourselves stuck.

kedashoe:
closes #3275


kedashoe:
closing due to #3278 , will open new PR with latest work from DT

Issue #3271, opened 2022-04-19
doc should note if function return not curry function 


HoKangInfo:
ex:
```
const list = unapply(I)

chain(list, I)(1)
// TypeError: fn(...) is not a function

converge(list, [I,I])(1)
// return [1,1]
```

alt
```
const list = curryN(2, unapply(I))

chain(list, I)(1)
// return [1,1]
```
like compose
Note: The result of compose is not automatically curried.

CrossEye:
Agreed.  This would be a very good idea.  Do you have an interest in creating a PR for this?

HoKangInfo:
Anytime,
https://github.com/ramda/ramda/pull/3279

PR #3279, opened 2022-05-05
doc should note if function return not curry function (#3271)

HoKangInfo:
Simple addition a Note.


source/unapply.js
```diff
@@ -26,7 +28,7 @@ import _curry1 from './internal/_curry1.js';
  * @symb R.unapply(f)(a, b) = f([a, b])
  */
 var unapply = _curry1(function unapply(fn) {
-  return function() {
+  return function () {
```
customcommander, line 31: This change causes a linter error. Could you revert it please?

PR #3280, opened 2022-05-21
split types into separate files

### LINKED ISSUE #3274: typescript .d.ts structure
From #3264 

> files structure? Should we keep the one-giant-index.d.ts or should we split each function into its own file?

Aside from the costs/benefits for those working on the types here in the ramda repo, are there any consequences one way or the other for the end users who will be using ramda/the types? Does it change how things can be imported?

kedashoe:
Closes #3274 

This PR removes the index.d.ts file, splitting its contents into a type definitions file per source function in a new `/types` directory.

I've only made two "changes":

- It looks like `ArrayLike` is deprecated? This was only used in the definition of `length`, which I've changed to

```typescript
export function length<T extends { length: number }>(list: T): number;
```

- There was a single random utility type in the main index.d.ts file, `PartialRecord`, which I've moved to the toold.d.ts file.

I've added a new build script which combines the separate .d.ts files into an index.d.ts file, which is then placed in the /es directory at publish time.


CrossEye:
@kedashoe:  It looks like you're putting serious work into this.  Thank you!

I believe you're the only active committer with the expertise to validate this.  Usually we ask for at least two committers to review changes before we merge them.  But I think here, we may have to depend on you to do them yourself.  (I'm willing to do the merges, but it would likely just be a rubber stamp.)  Do you concur, @customcommander?

Of course if @hitmands is interested in joining your effort, that would be better still.

----------

I've got to admit that as I look at many of those TS types that I still have a hankering to see how far we could get with autogeneration from H-M type signatures.

kedashoe:
@adispring as well. cc @essenmitsosse, @somebody1234.

But ya, at this point it doesn't even matter that much. I'm basically just building it locally and verifying 1. my editor can find things, 2. an external compiler can find tings. As the effort continues (and of course before we merge anything back to main) we'll want more confirmation things are working for different setups/idiomatic/etc.

I'll leave this open for another day or two.

customcommander:
I have no experience with anything related to TypeScript. I'm happy to trust @kedashoe here.

types/toString.d.ts
```diff
@@ -0,0 +1,3 @@
+
```
adispring, line 1: Here are some redundant blank space.

types/values.d.ts
```diff
@@ -0,0 +1,3 @@
+import { ValueOfUnion } from './util/tools';
+
+export function values<T extends object, K extends keyof T>(obj: T): Array<T[K] | ValueOfUnion<T>>;
```
adispring, line 3: The following is typescript official type of `Object.values`, maybe it is more practical:

```ts
    /**
     * Returns an array of values of the enumerable properties of an object
     * @param o Object that contains the properties and methods. This can be an object that you created or an existing Document Object Model (DOM) object.
     */
    values<T>(o: { [s: string]: T } | ArrayLike<T>): T[];
```

types/mergeRight.d.ts
```diff
@@ -0,0 +1,4 @@
+import { Merge } from './util/tools';
```
adispring, line 1: `Object.assign` type:

```ts
    /**
     * Copy the values of all of the enumerable own properties from one or more source objects to a
     * target object. Returns the target object.
     * @param target The target object to copy to.
     * @param source The source object from which to copy properties.
     */
    assign<T, U>(target: T, source: U): T & U;
```

types/keys.d.ts
```diff
@@ -0,0 +1,2 @@
+export function keys<T extends object>(x: T): Array<keyof T>;
+export function keys<T>(x: T): string[];
```
adispring, line 2: The generic parameter `T` in the second overload of `keys` is unnecessary, we can rewrite it to:

```ts
keys(o: object): string[];
```

which is a typescript official type.

adispring (changes requested):

adispring:
There are some new apis that have no types by now, should we implement their types in this PR, or later?

types/keys.d.ts
adispring: https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#generics

scripts/build-types.js
```diff
@@ -0,0 +1,183 @@
+let Path = require('path');
```
adispring, line 1: `let` can be replaced by `const` in most case?

adispring (changes requested):

kedashoe:
Thanks for the review @adispring, great stuff! I'll update this PR with the ones that look most obvious to me and make separate issues for the rest.

kedashoe:
Ok, PR updated to delete type files for functions that have been removed. Couple other comments addressed.

 I realized that since the final index.d.ts file is not committed, we can't see the output of build-types.js. Here is a sample:

```typescript
/**
 * Applies a function to the value at the given index of an array, returning a
 * new copy of the array with the element at the given index replaced with the
 * result of the function application.
 *
 * See also {@link update}
 *
 * @example
 * ```typescript
 * R.adjust(1, R.toUpper, ['a', 'b', 'c', 'd']);      //=> ['a', 'B', 'c', 'd']
 * R.adjust(-1, R.toUpper, ['a', 'b', 'c', 'd']);     //=> ['a', 'b', 'c', 'D']
 * ```
 */
export function adjust<T>(index: number, fn: (a: T) => T, list: readonly T[]): T[];
export function adjust<T>(index: number, fn: (a: T) => T): (list: readonly T[]) => T[];

/**
 * Returns `true` if all elements of the list match the predicate, `false` if
 * there are any that don't.
 * 
 * Dispatches to the `all` method of the second argument, if present.
 * 
 * Acts as a transducer if a transformer is given in list position.
 *
 * See also {@link any}, {@link none}, {@link transduce}
 *
 * @example
 * ```typescript
 * const equals3 = R.equals(3);
 * R.all(equals3)([3, 3, 3, 3]); //=> true
 * R.all(equals3)([3, 3, 1, 3]); //=> false
 * ```
 */
export function all<T>(fn: (a: T) => boolean, list: readonly T[]): boolean;
export function all<T>(fn: (a: T) => boolean): (list: readonly T[]) => boolean;
```

So just getting all the exports from the type files, matching it up with docs from the source files and putting it all together.

CrossEye:
Right, I forgot @adispring also was working on this.  Thanks for chipping in.

I'm not following all this very closely, but I will try to keep looking in.  Thanks, folks!

types/assoc.d.ts
```diff
@@ -0,0 +1,7 @@
+import { Placeholder, AssocPartialOne } from './util/tools';
+
+export function assoc<T, U>(__: Placeholder, val: T, obj: U): <K extends string>(prop: K) => Record<K, T> & Omit<U, K>;
```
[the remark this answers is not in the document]
valerii15298: If we are gonna delete function types with placeholders it would be good to keep it somewhere. Maybe later someone will want to implement them all.

Issue #3273, opened 2022-04-23
typescript tests


kedashoe:
.d.ts files in ./types needing test coverage:

 - [ ]  __
 - [x]  add
 - [ ]  addIndex
 - [x]  adjust
 - [x]  all
 - [x]  allPass
 - [x]  always
 - [x]  and
 - [ ]  andThen
 - [x]  any
 - [ ]  anyPass
 - [ ]  ap
 - [ ]  aperture
 - [ ]  append
 - [ ]  apply
 - [ ]  applySpec
 - [ ]  applyTo
 - [ ]  ascend
 - [ ]  assoc
 - [ ]  assocPath
 - [ ]  binary
 - [ ]  bind
 - [ ]  both
 - [ ]  call
 - [ ]  chain
 - [ ]  clamp
 - [ ]  clone
 - [ ]  collectBy
 - [ ]  comparator
 - [ ]  complement
 - [ ]  compose
 - [ ]  composeWith
 - [ ]  concat
 - [ ]  cond
 - [ ]  construct
 - [ ]  constructN
 - [ ]  converge
 - [ ]  count
 - [ ]  countBy
 - [ ]  curry
 - [ ]  curryN
 - [ ]  dec
 - [ ]  defaultTo
 - [ ]  descend
 - [ ]  difference
 - [ ]  differenceWith
 - [ ]  dissoc
 - [ ]  dissocPath
 - [ ]  divide
 - [ ]  drop
 - [ ]  dropLast
 - [ ]  dropLastWhile
 - [ ]  dropRepeats
 - [x]  dropRepeatsBy
 - [ ]  dropRepeatsWith
 - [ ]  dropWhile
 - [ ]  either
 - [ ]  empty
 - [ ]  endsWith
 - [ ]  eqBy
 - [ ]  eqProps
 - [ ]  equals
 - [ ]  evolve
 - [ ]  F
 - [ ]  filter
 - [ ]  find
 - [ ]  findIndex
 - [ ]  findLast
 - [ ]  findLastIndex
 - [ ]  flatten
 - [ ]  flip
 - [ ]  forEach
 - [ ]  forEachObjIndexed
 - [ ]  fromPairs
 - [ ]  groupBy
 - [ ]  groupWith
 - [ ]  gt
 - [ ]  gte
 - [ ]  has
 - [ ]  hasIn
 - [ ]  hasPath
 - [ ]  head
 - [ ]  identical
 - [ ]  identity
 - [ ]  ifElse
[...]

PR #3283, opened 2022-05-29
feat: add tsd and one test file

somebody1234:
I looked for another test file to add but I think it'd be better to just rewrite tests completely...
I think one test file should be enough to confirm that it works though.
Do note that it's hard to tell whether it works when all tests pass, unfortunately, so if you want to make sure it works you should try adding an incorrect test

### DIFF
--- a/test/types/add.ts
+++ b/test/types/add.ts
@@ -3,6 +3,7 @@ import * as R from '../../es/index';
 
 expectType<number>(R.add(2, 3)); // => 5
 expectType<number>(R.add(7)(10)); // => 17
+expectType<(a: number) => number>(R.add(7));
 expectError(R.add('foo', 'bar'));
 expectError(R.add('foo')('bar'));
 
--- a/test/types/add.ts
+++ b/test/types/add.ts
@@ -1,0 +1,8 @@
+import { expectType, expectError} from 'tsd';
+import * as R from '../../es/index';
+
+expectType<number>(R.add(2, 3)); // => 5
+expectType<number>(R.add(7)(10)); // => 17
+expectError(R.add('foo', 'bar'));
+expectError(R.add('foo')('bar'));
+
--- a/test/types/add.ts
+++ b/test/types/add.ts
@@ -1,8 +1,8 @@
-import { expectType, expectError} from 'tsd';
+import { expectType, expectError } from 'tsd';
 import * as R from '../../es/index';
 
-expectType<number>(R.add(2, 3)); // => 5
-expectType<number>(R.add(7)(10)); // => 17
+expectType<number>(R.add(2, 3));
+expectType<number>(R.add(7)(10));
 expectType<(a: number) => number>(R.add(7));
 expectError(R.add('foo', 'bar'));
 expectError(R.add('foo')('bar'));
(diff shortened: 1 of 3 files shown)


somebody1234:
oh, related to #3273

somebody1234:
*maybe* the output could be improved by using the programmatic api but i wouldn't be too hopeful since it seems like what they do for `expectError` is remove diagnostics:
https://github.com/SamVerschueren/tsd/blob/106af14cef85530c73b390ddc4a51560c60dfb64/source/lib/compiler.ts#L134-L136

test/types/add.ts
```diff
@@ -0,0 +1,8 @@
+import { expectType, expectError} from 'tsd';
+import * as R from '../../es/index';
+
+expectType<number>(R.add(2, 3)); // => 5
+expectType<number>(R.add(7)(10)); // => 17
```
adispring, line 5: Since ramda apis are auto curried, we should include tests of partial applied `R.add` , such as:

```ts
expectType<(a: number) => number>(R.add(1));
```

adispring (approved):

test/types/add.ts
somebody1234: yeah idk about those -
i think if the `expectType<number>(R.add(7)(10));` it should be safe to assume currying is working correctly

testing partially applied functions would work but i'm not sure at what point it'd stop being tests and start being baselines - not that that's a bad thing

somebody1234: but, well, if these *are* supposed to be baselines maybe we'd want to use those instead of tests...

... but actually i'm not sure if there's a module out there that generates baselines currently

kedashoe: Ya I'm not sure what to do about this one.. I guess better to cover it even if it is a bit annoying

kedashoe (changes requested):
Could you also commit the changes to `package-lock.json`? They'd just get picked up by someone else installing tsd anyway, may as well commit them now.

Thanks for the PR @somebody1234 !

test/types/add.ts
```diff
@@ -0,0 +1,9 @@
+import { expectType, expectError} from 'tsd';
+import * as R from '../../es/index';
+
+expectType<number>(R.add(2, 3)); // => 5
```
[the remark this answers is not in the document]
somebody1234: it's generally because test output is copied from the examples from the docs.
that said, if it gets out of sync then i feel like that would mean something's gone *very* wrong - since that would be a breaking change... more importantly it'd mean the current definition is wrong and i feel like most of those would have already been caught

somebody1234:
@kedashoe what's this about that last review btw? ([this](https://github.com/ramda/ramda/pull/3283#discussion_r884919659) review)

kedashoe:
> @kedashoe what's this about that last review btw? ([this](https://github.com/ramda/ramda/pull/3283#discussion_r884919659) review)

Are you referring to this comment I made?

> Ya I'm not sure what to do about this one.. I guess better to cover it even if it is a bit annoying

I was just saying that I guess I'd prefer to add tests for currying for now. But if anyone has a good argument not to I'd be ok with that as well.

kedashoe (approved):

PR #3288, opened 2022-06-28
Fix of @since for splitWhenever jsdoc

octopuss:
The original jsdoc is misleading because splitWhenever was added in 0.28.0 see https://github.com/ramda/ramda/issues/2881


CrossEye (approved):
Thank you.  We must have missed someone adding that manually, as the build should automatically add this.

PR #3301, opened 2022-09-11
chore: Bump fast-check to version 3

dubzzz:
The version 3 of fast-check added something very interesting for libraries such as ramda: it can more likely produce values such as `__proto__` or `toString` which are known sources for CVE.

On fast-check's side, I also worked on some tooling making poisoning related issues even simpler to detect with an automatic detection for poisoning on globals. I successfully found back several past CVEs of lodash thanks to it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "ramda",
-  "version": "0.29.1",
+  "version": "0.30.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "ramda",
-      "version": "0.29.1",
+      "version": "0.30.0",
       "license": "MIT",
       "devDependencies": {
         "@babel/cli": "^7.23.0",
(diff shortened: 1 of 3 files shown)


dubzzz:
Next target will be to leverage fast-check to detect some potential prototype poisoning in the exposed APIs before they turn into CVEs.

dubzzz:
@kedashoe I just updated this PR. I believe it can be useful for the project to rely on the latest version of fast-check in order to benefit from all the improvements brought by v3. It includes prototype poisoning related detection which is something that can easily strike projects such as ramda.

dubzzz:
NOTE: I'm currently working on a version 4 of fast-check. It should not come with many breaking changes but still always to be closer to the latest release to simplify future bumps.

kedashoe (approved):

kedashoe:
ty @dubzzz sorry for missing this the first time

dubzzz:
No worry, I pinged you back so that it got merged. It was neither critical, nor urgent 😂

PR #3305, opened 2022-09-16
Adds type support for dates on R.type function

### LINKED ISSUE #3296: Typescript error for R.type when using a date type
For typescript, when using the function R.type, the possible types are 
`"Object" | "Number" | "Boolean" | "String" | "Null" | "Array" | "RegExp" | "Function" | "Undefined" | "Symbol" | "Error"`

However, if you pass R.type(new Date), a string of 'Date' gets returned

mateusbw:
Adds type support for dates on R.type function

Fixes #3296

### DIFF
--- a/test/type.js
+++ b/test/type.js
@@ -45,4 +45,8 @@ describe('type', function() {
     eq(R.type(undefined), 'Undefined');
   });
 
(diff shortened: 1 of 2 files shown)


kedashoe:
Thank you!

PR #3316, opened 2022-10-20
Enhanced types for Assoc

Harris-Miller:
Updating types for Assoc for better return types when the type for value is the same for the type of the key on the object

Also added tests

### DIFF
--- a/test/types/assoc.ts
+++ b/test/types/assoc.ts
@@ -1,27 +1,70 @@
 import { expectType } from "tsd";
 import { __, assoc } from "../../es/index";
-import { AssocPartialOne} from '../../types/util/tools';
 
 type Point = { x: number; y: number };
 const p: Point = { x: 2, y: 3 };
 
-const t = assoc('x')(5, p);
-
+// assoc(__, val, obj)(prop), this tests if prop is keyof obj and if val is typeof obj[prop] for best return type
 expectType<Point>(assoc(__, 5, p)('x'));
 expectType<Record<'x', string> & Omit<Point, 'x'>>(assoc(__, '5', p)('x'));
 expectType<Point & Record<'z', number>>(assoc(__, 5, p)('z'));
 
+// assoc(prop, __, obj)(val), when K is keyof obj, tests if val is typeof obj[prop] for best return type
 expectType<Point>(assoc('x', __, p)(5));
 expectType<Record<'x', string> & Omit<Point, 'x'>>(assoc('x', __, p)('5'));
+
+// assoc(prop, __, obj)(val), when prop is not keyof obj
 expectType<Point & Record<'z', number>>(assoc('z', __, p)(5));
 
+// assoc(prop, val, obj) when prop is keyof obj and val is same type
 expectType<Point>(assoc('x', 5, p));
+
+// assoc(prop, val, obj) when prop is keyof obj  and val is not same type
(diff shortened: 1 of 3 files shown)


test/types/assoc.ts
```diff
@@ -0,0 +1,27 @@
+import { expectType } from "tsd";
+import { __, assoc } from "../../es/index";
+import { AssocPartialOne} from '../../types/util/tools';
```
kedashoe, line 3: I don't believe this is being used?

test/types/assoc.ts
```diff
@@ -0,0 +1,27 @@
...
+
+type Point = { x: number; y: number };
+const p: Point = { x: 2, y: 3 };
+
+const t = assoc('x')(5, p);
```
kedashoe, line 8: Does this line do something?

types/assoc.d.ts
```diff
@@ -1,7 +1,10 @@
 import { Placeholder, AssocPartialOne } from './util/tools';
 
-export function assoc<T, U>(__: Placeholder, val: T, obj: U): <K extends string>(prop: K) => Record<K, T> & Omit<U, K>;
-export function assoc<U, K extends string>(prop: K, __: Placeholder, obj: U): <T>(val: T) => Record<K, T> & Omit<U, K>;
-export function assoc<T, U, K extends string>(prop: K, val: T, obj: U): Record<K, T> & Omit<U, K>;
+export function assoc<T, U>(__: Placeholder, val: T, obj: U): <K extends string>(prop: K) => K extends keyof U ? T extends U[K] ? U : Record<K, T> & Omit<U, K> : U & Record<K, T>;
+export function assoc<U, K extends keyof U>(prop: K, __: Placeholder, obj: U): <T>(val: T) => T extends U[K] ? U : Record<K, T> & Omit<U, K>;
+export function assoc<U, K extends string>(prop: K, __: Placeholder, obj: U): <T>(val: T) => U & Record<K, T>;
+export function assoc<K extends keyof U, U>(prop: K, val: U[K], obj: U): U;
```
kedashoe, line 6: Is this covered by the next definition? ie `export function assoc<T, K extends keyof U, U>(prop: K, val: T, obj: U): Record<K, T> & Omit<U, K>;`

Is there a reason to have both?

kedashoe (commented):
Thanks again for all the PRs! Threw a couple questions inline.

Also, what do you think about copying and pasting, as a comment, the signature you expect each block of tests to hit? So something like
```
// assoc<K extends keyof U, U>(prop: K, val: U[K], obj: U): U;
expectType<Point>(assoc('x', 5, p));
etc..
```

test/types/assoc.ts
Harris-Miller: No, I was using it to test, forgot to remove

types/assoc.d.ts
Harris-Miller: I looked back at how I wrote this, and with some things I learned from doing my other MRs, I completely rewrote it. It has a lot more comments now so it'll be easier to review. I also was able to remove the `AssocPartialOne` util type, it's no longer needed

test/types/assoc.ts
Harris-Miller: Removed

Harris-Miller:

PR #3317, opened 2022-10-20
Add types to chain to handle Monad case

Harris-Miller:
Also added tests


Harris-Miller:

PR #3319, opened 2022-10-22
Enhanced typings for pickAll

Harris-Miller:
Adding a better typing option when all the keys are known on the object


Harris-Miller:

PR #3320, opened 2022-10-22
Added typings for placeholder options on adjust

Harris-Miller:
Added tests as well


Harris-Miller:

PR #3321, opened 2022-10-22
Complete typings for R.all

Harris-Miller:
Corrected the typings to handle placeholders and the "Dispatches to the all method of the second argument, if present." case


Harris-Miller:

PR #3322, opened 2022-10-22
Typings and tests for aperture

Harris-Miller:
Adding the missing curried type and added tests


Harris-Miller:

PR #3323, opened 2022-10-22
Added curried version for R.and and fixed tests

Harris-Miller:
I added typings for `and(__, b)(a)` variety

Also redid the tests for it. They were failing. Let me know if you want to be more exhaustive with the tests for this one and I'll add more (probably not needed TBH, the typing for this function is very simple)


Harris-Miller:

PR #3325, opened 2022-10-23
Enhanced typings for R.any

Harris-Miller:
Added missing curry type and handle "dispatches to any method" case.

Added tests. I kept most of the ones that exist already, specifically the `expectError` cases. Let me know if you want me to make those test cases in my other MRs


Harris-Miller:

PR #3327, opened 2022-10-23
Types and tests for R.append

Harris-Miller:
This one is very straight forward


Harris-Miller:

PR #3357, opened 2023-03-21
do not enforce js extension omission in imports

Gastonite:
Hello and thank you for this excellent package.

I'm struggling with imports because nodejs don't let me import a source file with the ".js" extension.

But first, it seems that the package.json of version 0.28.0 (on npm) has no exports that allows per-module imports:
```json
 "exports": {
    ".": {
      "require": "./src/index.js",
      "import": "./es/index.js",
      "default": "./src/index.js"
    },
    "./es/": "./es/",
    "./src/": "./src/",
    "./dist/": "./dist/"
  },
```

So when importing `ramda/src/pipe` it throws an error:
```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './src/pipe' is not defined by "exports" in
```

On the other hand, the master branch has differents `exports`, which seems to accept imports like `ramda/src/pipe`:

```json
  "exports": {
    ".": {
      "require": "./src/index.js",
      "import": "./es/index.js",
      "default": "./src/index.js"
    },
    "./es/*": "./es/*.js",
    "./src/*": "./src/*.js",
    "./dist/*": "./dist/*.js",
    "./es/": "./es/",
    "./src/": "./src/",
    "./dist/": "./dist/"
  },
```
Importing `ramda/src/pipe` works great but it's necessary to install the master branch version of the package.

But it don't let me import modules with js extension (ie.  `ramda/src/pipe.js`)
I get an error:
```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './src/pipe.js' is not defined by "exports" in
```

I need to import es modules in two ways:
- in a browser (via a CDN that looks at `exports` to determine which file to serve)
- in a node server (which refuse to import the modules because there is no match in `exports` for the specifier `ramda/src/pipe.js`  )

So I propose this change to let users decide if they import files with or without the `.js` extension:
```json
"exports": {
    ".": {
      "require": "./src/index.js",
      "import": "./es/index.js",
      "default": "./src/index.js"
    },
    "./es/*": "./es/*",
    "./src/*": "./src/*",
    "./dist/*": "./dist/*",
    "./es/": "./es/",
    "./src/": "./src/",
    "./dist/": "./dist/"
  },
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,9 +56,9 @@
       "import": "./es/index.js",
       "default": "./src/index.js"
     },
-    "./es/*": "./es/*.js",
-    "./src/*": "./src/*.js",
-    "./dist/*": "./dist/*.js",
+    "./es/*": "./es/*",
+    "./src/*": "./src/*",
+    "./dist/*": "./dist/*",
     "./es/": "./es/",
     "./src/": "./src/",
     "./dist/": "./dist/"


kedashoe:
Hi @Gastonite , thank you for the PR. I've pulled down your branch to test locally but I'm not sure precisely what you are proposing/expecting. What version of node are you using? And could you provide some code samples of what is and isn't working for you now and what you expect to work/not work after this PR?

Something like
```ts
import { add } from 'ramda/es/add'
```
?

 Please test with the version we just released, v0.29.0 if you could as well.

Gastonite:
# Context 

I wrote a module that I wanted to use from both nodejs (19.7.0) and a browser (using an importmap generated with `@jspm/generator`)

This module imports some ramda modules individually so as not to download the whole library when imported from a browser.


# With ramda@0.28.0
## browser

- `@jspm/generator` correctly resolved the `ramda/src/map.js` and `ramda/es/map.js` specifiers but not the `ramda/src/map` and `ramda/es/map` specifiers:

When I was running this script:
```js
import { Generator } from '@jspm/generator';

const generator = new Generator({
  mapUrl: import.meta.url,
  defaultProvider: 'jspm.io',
  env: ['production', 'browser', 'module'],
});

await generator.install('ramda@0.28.0/src/map.js')

console.log(JSON.stringify(generator.getMap(), null, 2));
```

The `map.js` module is correctly resolved by `@jspm/generator` and the following importmap is produced:
```json
{
  "imports": {
    "ramda/src/map.js": "https://ga.jspm.io/npm:ramda@0.28.0/src/map.js"
  }
}
```

But, it didn't work with subpaths like `ramda/src/*` or `ramda/es/*`:
```js
await generator.install('ramda@0.28.0/es/map');
```
It throwed an error:
```
JspmError: Module not found: https://ga.jspm.io/npm:ramda@0.28.0/es/map imported from...
```

So as I want to share this module between browser and nodejs, my only solution was to use path `ramda/src/map.js`


## server nodejs

And on the other hand, I wanted to use this same module from a nodejs script, but whatever type of subpath I used in my imports (`./src/map`, `./src/map.js` , `./es/map`, or `./es/map.js`), I always got the following error:
```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './src/map.js' is not defined by "exports"...
```




# With ramda@0.29.0

Everything seems to be working now =)

I can now use and generate my importmap for the browser with:
```js
await generator.install('ramda@0.29.0/es/map');
```
and therefore import the `map` module with the `ramda/es/map` import specifier

This is also correctly solved by nodejs, and my file can now be shared between the 2.

Thank you

kedashoe:
Great! And thank you for the in depth followup

Issue #3367, opened 2023-04-03
Documentation errors in R.swap examples


foxdonut:
There are two errors in the documentation of `R.swap` examples.

Reference: 

https://ramdajs.com/docs/#swap

**First error:**

```js
R.swap(-1, 0, ['a', 'b', 'c', 'd', 'e', 'f']); //=> ['f', 'b', 'c', 'd', 'e', 'a'] list rotation
```
This is not list rotation. Rotating the list would result in `['b', 'c', 'd', 'e', 'f', 'a']`.

**Second error:**

```js
R.swap('a', 'b', {a: 1, b: 2}); //=> {a: 2, b: 2}
```

The result is incorrect, it should be `{a: 2, b: 1}`.

Thanks!

kedashoe:
Thank you for the issue, agreed on both! Would you like to submit a PR?

foxdonut:
@kedashoe With pleasure. PR submitted.

kedashoe:
https://github.com/ramda/ramda/pull/3368

PR #3368, opened 2023-04-03
Fix for #3367: Documentation errors in R.swap examples

foxdonut:
See #3367 for details.


kedashoe (approved):
Looks good ty, don't imagine this will be very controversial but will leave open for another day in case anyone has any comments

Issue #3364, opened 2023-03-30
clone new Array(number);


aboveyunhai:
I realized that ramda will remove all value from the new Array initialization with predefine size, is this intentional?

```js
const data = new Array(5);

console.log(data); // [undefined, undefined, undefined, undefined, undefined] , length: 5
console.log(R.clone(data));
console.log(result); // [] length: 0
 
```
as oppose to lodash

```
console.log(data); // [undefined, undefined, undefined, undefined, undefined] , length: 5
console.log(R.clone(data));
console.log(result); // [undefined, undefined, undefined, undefined, undefined] , length: 5
```

Harris-Miller:
Welcome to the wonderful world of javascript!

Kidding aside. The behavior of `new Array(x)` is very often misunderstood. `new Array(5)` effectively creates `{ length: 5 }` with prototypeOf `Array`. This is different from `new Array(5).fill(undefined)` that would be `{ 0: undefined, 1: undefined, 2: undefined, 3: undefined, 4: undefined, 5: undefined, length: 5 }`

```typescript
console.log(new Array(5).fill(undefined)) // [undefined, undefined, undefined, undefined, undefined] , length: 5
```

The way that Ramda clones arrays is by starting with an empty array and cloning each `ownProperties`'s value from the original value over to it. The problem with this implementation is that for `new Array(5)`, there are no `ownProperties` (The `length` property is non-enumerable).

This means that you put in an Array of length 5, and get out an Array of length 0. However, both the input and output array are still technically empty.

In general, we get a lot of variety on if `[] === new Array(5)` with respect to how `length: 5` is treated when there are no keys

```typescript
const data = new Array(5);

R.equals([], data); // true
lodash.isEqual([], data)); // false

data[0]; // undefined for `0` or any index
[...data]; // [undefined, undefined, undefined, undefined, undefined] , length: 5
Array.from(data); // [undefined, undefined, undefined, undefined, undefined] , length: 5

Obect.keys(data); // []
data.forEach(v => console.log(v)); // noop
for (let i; i <= data.length; i++) console.log(data[i]); // `undefined` x5
for (const key in data) console.log(key); // noop
for (const value of data) console.log(value); // // `undefined` x5

console.log(structuredClone(new Array(5))); // [] , length: 5
```

The question is: do we respect the value of the `length` property over the actual existance of the array?

[...]

PR #3370, opened 2023-04-11
Update: Known array lengths

Harris-Miller:
As pointed out in https://github.com/ramda/ramda/issues/3364, V8 cares if you instantiate an Array with a known length is far faster to fill than to start with an empty array

This MR updates the following areas of `ramda` to do exactly that

* _clone :: https://jsperf.app/jopaqu - seeing a ~10% gain
* intersperse :: https://jsperf.app/dibehu - ~40%
* mapAccum :: https://jsperf.app/gozalu - 2% (small)
* mapAccumRight, same as ^^
* range :: https://jsperf.app/yaxihu - ~45%
* scan :: https://jsperf.app/tokefo - ~50%
* times :: https://jsperf.app/matoqu - ~50% gain!
* xprod :: https://jsperf.app/cezeko - ~75%
* zip :: https://jsperf.app/foxowi - seeing a ~15% gain
* zipWith :: https://jsperf.app/yasibi - ~ 20%

### DIFF
--- a/source/intersperse.js
+++ b/source/intersperse.js
@@ -21,6 +21,7 @@ import _curry2 from './internal/_curry2.js';
  */
 var intersperse = _curry2(_checkForMethod('intersperse', function _intersperse(separator, list) {
   var length = list.length;
+  if (length === 0) { return []; }
   var out = Array(length * 2 - 1);
   var idx = 0;
   while (idx < length) {
--- a/source/intersperse.js
+++ b/source/intersperse.js
@@ -19,15 +19,17 @@ import _curry2 from './internal/_curry2.js';
  *
  *      R.intersperse('a', ['b', 'n', 'n', 's']); //=> ['b', 'a', 'n', 'a', 'n', 'a', 's']
  */
-var intersperse = _curry2(_checkForMethod('intersperse', function intersperse(separator, list) {
-  var out = [];
-  var idx = 0;
+var intersperse = _curry2(_checkForMethod('intersperse', function _intersperse(separator, list) {
   var length = list.length;
+  var out = Array(length * 2 - 1);
+  var idx = 0;
   while (idx < length) {
+    var i = idx * 2;
     if (idx === length - 1) {
-      out.push(list[idx]);
+      out[i] = list[idx];
     } else {
-      out.push(list[idx], separator);
+      out[i] = list[idx];
+      out[i + 1] = separator;
     }
     idx += 1;
   }
--- a/source/xprod.js
+++ b/source/xprod.js
@@ -24,7 +24,7 @@ var xprod = _curry2(function xprod(a, b) { // = xprodWith(prepend); (takes about
   var ilen = a.length;
   var j;
   var jlen = b.length;
-  var result = Array(ilen + jlen);
+  var result = Array(ilen * jlen);
   while (idx < ilen) {
     j = 0;
     while (j < jlen) {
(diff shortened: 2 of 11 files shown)


Berndy:
These seem like some useful improvements, ramda is in dire need of performance. Maybe sync with master and declare it ready for review?

source/intersperse.js
```diff
@@ -19,15 +19,17 @@ import _curry2 from './internal/_curry2.js';
  *
  *      R.intersperse('a', ['b', 'n', 'n', 's']); //=> ['b', 'a', 'n', 'a', 'n', 'a', 's']
  */
-var intersperse = _curry2(_checkForMethod('intersperse', function intersperse(separator, list) {
-  var out = [];
-  var idx = 0;
+var intersperse = _curry2(_checkForMethod('intersperse', function _intersperse(separator, list) {
   var length = list.length;
+  var out = Array(length * 2 - 1);
```
kedashoe, line 24: This will throw an error if intersperse is passed an empty array. We can just bail early and return empty array in that case.

source/xprod.js
```diff
@@ -24,11 +24,11 @@ var xprod = _curry2(function xprod(a, b) { // = xprodWith(prepend); (takes about
   var ilen = a.length;
   var j;
   var jlen = b.length;
-  var result = [];
+  var result = Array(ilen + jlen);
```
kedashoe, line 27: This should be multiplication instead of addition yes? `Array(ilen * jlen)`. I suspect the huge performance gains for xprod have something to do with this..

source/xprod.js
```diff
@@ -24,11 +24,11 @@ var xprod = _curry2(function xprod(a, b) { // = xprodWith(prepend); (takes about
...
   while (idx < ilen) {
     j = 0;
     while (j < jlen) {
-      result[result.length] = [a[idx], b[j]];
+      result[idx + j] = [a[idx], b[j]];
```
kedashoe, line 31: And here in the inner loop, we are adding `jlen` elements each time through, so I believe we have

`result[(idx * jlen) + j] = [a[idx], b[j]];`

It is also a bit confusing to read with `idx` and `j` as the index variables. Can we update this function to use just plain old `i` to match the `j` (I know you didn't name these originally)

kedashoe (changes requested):
Thanks for the PR @Harris-Miller , sorry I didn't review this initially!

kedashoe:
> These seem like some useful improvements, ramda is in dire need of performance. Maybe sync with master and declare it ready for review?

ty for the ping on this one @Berndy , I will try to review #3417 in the next day or two

source/xprod.js
Harris-Miller: Updated

source/xprod.js
Harris-Miller: Updated

source/intersperse.js
Harris-Miller: Updated with a bail out when `length === 0`

Harris-Miller:
@kedashoe made the updates you requested

kedashoe (approved):
:+1: Looks good to me, will leave open a bit for others to review the updates

Harris-Miller:
@kedashoe I added an optimization for arrays in `_clone`, but that broke unit tests so I reverted it. Will revisit that in another MR. Need re-approval because of the update

Harris-Miller:
@kedashoe all good again, read for re-approval

kedashoe (approved):

PR #3388, opened 2023-06-19
Optimize all Ramda functions

ku8ar:
Ramda optimisation.
This is the quickest way to determine that an argument is or is not a placeholder.
There is no faster method.

PR also solves the very rare problem where someone could have an object in the app with the same key as a placeholder.

### DIFF
--- a/source/__.js
+++ b/source/__.js
@@ -1,3 +1,4 @@
+import _placeholder from './internal/_placeholder.js';
 /**
  * A special placeholder value used to specify "gaps" within curried functions,
  * allowing partial application of any combination of arguments, regardless of
@@ -25,4 +26,4 @@
  *      const greet = R.replace('{name}', R.__, 'Hello, {name}!');
  *      greet('Alice'); //=> 'Hello, Alice!'
  */
-export default {'@@functional/placeholder': true};
+export default _placeholder;
--- a/source/internal/_isPlaceholder.js
+++ b/source/internal/_isPlaceholder.js
@@ -1,5 +1,5 @@
+import _placeholder from './_placeholder.js';
+
 export default function _isPlaceholder(a) {
-  return a != null &&
-         typeof a === 'object' &&
-         a['@@functional/placeholder'] === true;
+  return a === _placeholder;
 }
--- a/source/internal/_placeholder.js
+++ b/source/internal/_placeholder.js
@@ -1,0 +1,1 @@
+export default {'@@functional/placeholder': true};
--- a/test/curry.js
+++ b/test/curry.js
@@ -78,7 +78,7 @@ describe('curry', function() {
   it('supports @@functional/placeholder', function() {
     var f = function(a, b, c) { return [a, b, c]; };
     var g = R.curry(f);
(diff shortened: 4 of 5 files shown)


kedashoe:
:+1: this looks good to me. Will leave open for a bit, would like at least one other to weigh in here

adispring:
Will `===` checking affect cross-window checking? : https://github.com/ramda/ramda/issues/2533

semmel:
> Will `===` checking affect cross-window checking? : https://github.com/ramda/ramda/issues/2533

Looks like it would finally allow  WindowProxy arguments for all curried Ramda functions.
It might actually be the solution I was missing in #2533.
Have to do some tests though to confirm.

Harris-Miller:
This update should make no difference on the Typescript end of things either. The current definition we have is this
```typescript
import { A } from 'ts-toolbelt';

export type Placeholder = A.x & { '@@functional/placeholder': true };
```

I don't know if the `x: Math.random()` came first and the `A.x` was put in to support that, or the other way around. Regardless, I never understood the point of the `A.x &`, as I could not find any value added to the typings for it.

See playground example: https://tsplay.dev/Wyjoxw

bas080 (approved):

CrossEye:
Note that that existing behavior is there for a reason.  This doesn't mean it's not time to change it, but we should also understand what it's giving up.  I haven't found the relevant discussions yet, but the below should be pretty close:

When we wanted to add a placeholder for obvious reasons in partial application, we didn't want to go down the route that lodash had, using the library itself as the placeholder.  It was an obvious thing to do with Underscore and lodash, where the library was usually exposed using the name `_`.  But it didn't make sense to pass Ramda itself as the placeholder.  In fact, we wanted to go in the opposite direction.  

We created a version that was meant to allow for interoperability with other libraries.  The thought was that if users passed any library's placeholder, it should just work.  To make that possible, we treated the placeholder as a protocol, simply used an obscure property, modeled after one for [transducers][tr] and after the language's [iteration prototcols][ip].

It was a good idea, I still believe.  However, AFAIK, no one else ever followed this lead.  So I guess it's time to stop waiting for that, and gain the optimization.  Has anyone actually performed benchmarks to see what's gained?

So, from me, a reluctant 👍🏼 .

Let's not forget -- as demonstrated by the need to change a test case -- that this is a breaking change.




  [ip]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols
  [tr]: https://github.com/cognitect-labs/transducers-js?tab=readme-ov-file#transformer-protocol

CrossEye:
@semmel:

> Dealing with the messy browser DOM in my daily job, I've always struggled processing [property access limited WindowProxy](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy#cross-origin_script_api_access) with Ramda. Since isPlaceholder() permeates nearly every Ramda function, I always got an "access denied" error.

Could you describe a little more what happens here?  I'd love to see if we can fix that.  We need to move beyond our ES3/ES5 implementation, and when we do, would `Symbol.for`? work better?

PR #3389, opened 2023-06-24
Improve and modernize `ramda.js`

Lokeshsri11:
Changes made:
1. Renamed `_concat` to `concat` for better clarity.
2. Utilized default function parameters to set `array1` and `array2` to empty arrays by default.
3. Used the spread operator (`...`) to concatenate arrays for a more concise and modern approach.
4. Renamed `_arity` to `arity` for improved clarity.
5. Replaced the `switch` statement with a simple `if` condition to validate the argument `n`.
6. Used rest parameters (`...args`) to capture the arguments in the curried function.
7. Replaced `apply` with `call` for invoking the function to avoid the unnecessary use of `arguments`.


kedashoe:
Hi @Lokeshsri11 , thank you for the PR. We are holding off on modernizing ramda until after 1.0, so I will close this for now.

Issue #3393, opened 2023-07-26
swap does not work as intended


xhangjun:
## What happens?

version: v0.29.0

When the last argument is an array of arrays, `swap` does not work as intended:

![the error case](https://github.com/ramda/ramda/assets/53827639/9ec64cce-33cb-41b2-8c0a-5fe79d95be32)

## How To Reproduce

[playground to reproduce](https://ramdajs.com/repl/?v=0.29.0#?const%20alphas%20%3D%20%5B%0A%20%20%5B%27a%27%2C%20%27A%27%5D%2C%0A%20%20%5B%27b%27%2C%20%27B%27%5D%2C%0A%20%20%5B%27c%27%2C%20%27C%27%5D%2C%0A%20%20%5B%27d%27%2C%20%27D%27%5D%0A%5D%3B%0A%0AR.swap%280%2C%20-1%2C%20alphas%29%3B)

PR #3394, opened 2023-07-26
fix: make swap work when swapped items are arrays

xhangjun:
fix [issue 3393](https://github.com/ramda/ramda/issues/3393)

### DIFF
--- a/test/swap.js
+++ b/test/swap.js
@@ -28,6 +28,10 @@ describe('swap', function() {
     eq(R.swap(0, 1, list), R.swap(1, 0, list));
   });
 
+  it('works with lists of arrays', function() {
+    eq(R.swap(0, -1, [['a', 'A'], ['b', 'B']]), [['b', 'B'], ['a', 'A']]);
+  });
+
   it('swaps property values from one property to another', function() {
     eq(R.swap('a', 'b', {a: 1, b: 2}), {a: 2, b: 1});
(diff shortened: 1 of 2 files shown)


test/swap.js
```diff
@@ -3,12 +3,18 @@ var eq = require('./shared/eq.js');
...
+var c = ['c', 'C'];
+var d = ['d', 'D'];
+
 describe('swap', function() {
   it('swaps an element from one index to the other', function() {
     eq(R.swap(0, 1, list), ['b', 'a', 'c', 'd', 'e', 'f']);
     eq(R.swap(2, 1, list), ['a', 'c', 'b', 'd', 'e', 'f']);
     eq(R.swap(-1, 0, list), ['f', 'b', 'c', 'd', 'e', 'a']);
     eq(R.swap(4, 1, list), ['a', 'e', 'c', 'd', 'b', 'f']);
+    eq(R.swap(0, -1, [a, b, c, d]), [d, b, c, a]);
```
kedashoe, line 17: Could we move this to its own `it`? Would be more clear that we are testing arrays.
```
it('works with lists of arrays', function() {
  eq(R.swap(0, -1, [['a', 'A'], ['b', 'B']]), [['b', 'B'], ['a', 'A']]);
});
```
or something like that

kedashoe (changes requested):
Thank you for the issue and PR! Just one small comment

kedashoe (approved):

kedashoe (approved):

PR #3398, opened 2023-08-16
fix: modify curry doc

gleguizamon:
Fixes R.curry doc, as discussed here:
https://github.com/ramda/ramda.github.io/issues/282

### DIFF
--- a/source/curry.js
+++ b/source/curry.js
@@ -29,22 +29,6 @@ import curryN from './curryN.js';
  * Please note that default parameters don't count towards a [function arity](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/length)
  * and therefore `curry` won't work well with those:
  *
- * ```
- * const h = R.curry((a, b, c = 2) => a + b + c);
- *
- * h(40);
- * //=> function (waits for `b`)
- *
- * h(39)(1);
- * //=> 42
- *
- * h(1)(2, 3);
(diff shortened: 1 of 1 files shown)


source/curry.js
```diff
@@ -29,22 +29,6 @@ import curryN from './curryN.js';
  * Please note that default parameters don't count towards a [function arity](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/length)
  * and therefore `curry` won't work well with those:
```
gleguizamon, line 30: This comment may be confusing after deleting the snippet.
I would like to include it before [this](https://github.com/ramda/ramda/pull/3398/files#diff-980840a7f6a0dc0bb6094ffed3eebc0f36612cb4ffffc0e2bbd9367bc6d14a45R47-R49) example, but the url won't be accessible.

kedashoe: True how about we keep it as you suggest (just change the colon at the end to a period "...won't work well with those."), and then in the example you've added below put a smaller comment like

```
 *      // example of curry not working well with default arguments
 *      const h = R.curry((a, b, c = 2) => a + b + c);
 *      h(1)(2)(7); //=> Error! (`3` is not a function!)
```

kedashoe (changes requested):

kedashoe (dismissed):

kedashoe (approved):

PR #3382, opened 2023-05-14
Add version to es distribution

crshmk:
Fixes `ModuleFederationPlugin` versioning for shared packages, as discussed here:
https://github.com/webpack/webpack/issues/17188

PR #3399, opened 2023-08-18
fix: add sideEffects flag for es package.json

laboskin:
Currently tree shaking doesn't work, because es/package.json doesn't have "sideEffects" flag in it

This PR fixes the problem

**Before 👎:**
image

**After 👍:**
image

### DIFF
--- a/scripts/addModulePackageScope.js
+++ b/scripts/addModulePackageScope.js
@@ -4,4 +4,4 @@ const { version } = require('../package.json');
 
 const PACKAGE_SCOPE_PATH = path.join(__dirname, '..', 'es', 'package.json');
 
-fs.writeFileSync(PACKAGE_SCOPE_PATH, JSON.stringify({ type: 'module', version }));
+fs.writeFileSync(PACKAGE_SCOPE_PATH, JSON.stringify({ type: 'module', sideEffects: false, version }));


HEHOHADA:
Oh what a cool fix to finally solve the problems with the bundle

kedashoe:
I'm curious why es/package.json needs it as we [already have it](https://github.com/ramda/ramda/blob/165d2ae662e6ce31cb4a56e22bbc8ee06eb32bcf/package.json#L40) in the top level package.json. Could you help me minimally reproduce this? What bundler are you using?

laboskin:
We are using webpack 5.x.x. For some reason it take into account the `es/package.json` file and adding `"sideEffects": false` to it changes webpack's behavior.

After deeper investigation I noticed that it doesn't actually all ramda function to the bundle. However tree-shaking works wrong when the app consists of multiple chunks, and each chunk imports different ramda functions. Without the `sideEffects` flag webpack creates single chunk with all imported ramda functions, and this chunk is imported in all other chunks. With `sideEffects: false` webpack works correctly (each chunk has ramda functions that are imported inside the chuink)

Here is a repo where you can reproduce the issue:
https://github.com/laboskin/ramda-tree-shaking-test

Build it once (`npm run build`), then add `sideEffects: false` to `node_modules/ramda/es/package.json` and build it again. You will see the difference in reports of `WebpackBundleAnalyzer` and in output .js files

I also committed 2 report files (`report_default.html` and `report_with_sideEffects_false.html`) for convinience.

image

image

kedashoe:
Brilliant, thank you @laboskin .

I was able to reproduce locally. It looks to me like we also get the correct bundle if we remove es/package.json altogether, yes? But that probably doesn't matter for now as we still need {"type": "module"} somewhere.

Previous related PRs:

https://github.com/ramda/ramda/pull/2999
https://github.com/ramda/ramda/pull/3382

I will leave open for a bit to let others comment but this looks good to me.

laboskin:
@kedashoe I am not sure about removing es/package.json completely. I believe there was a purpose for adding it and if we remove it, it may cause problems for some esm or non-esm dependents. Adding `sideEffects` flag seems like a safer option

kedashoe:
> @kedashoe I am not sure about removing es/package.json completely. I believe there was a purpose for adding it and if we remove it, it may cause problems for some esm or non-esm dependents. Adding `sideEffects` flag seems like a safer option

Agreed

PR #3407, opened 2023-09-02
feat: issues/3404

cieldon32:
I add a function insertUntil, it will be
const gteTwo = x => x >= 2;
R.insertUntil(gteTwo,2, [1,3,3]); //=> [1, 2, 3, 3]

I hope this function to implement the bubble sorting


kedashoe:
@cieldon32 ty for the issue and PR. I would say this is a bit too specific for me and I can't remember this function being requested before. Maybe a better fit (or maybe it already exists) for one of the ramda supplemental libs?

PR #3410, opened 2023-09-11
fix: _curryN receive enough args

manyuemeiquqi:
[curry-discussions](https://github.com/ramda/ramda/discussions/3409)

Previous code doesn't judge every loop reveiving parameter the length of `combined` has `placeholder`. So i fix it.


source/internal/_curryN.js
```diff
@@ -34,7 +34,8 @@ export default function _curryN(length, received, fn) {
       }
       combinedIdx += 1;
     }
-    return left <= 0
+    var noPlaceholder = (combined.slice(0,length).every(item=>!_isPlaceholder(item)));
```
adispring, line 37: Please don't use es6+ apis, just use es5 apis.

adispring (changes requested):

source/internal/_curryN.js
manyuemeiquqi: Sorry, i modify it

adispring:
Can you add a unit test, to test if the change works as expected?

manyuemeiquqi:
> Can you add a unit test, to test if the change works as expected?

I add some question. When i attempt fix this quesiton, i had two question

1. I view lodash internal implement about, it will handle next  loop placeholder as  func, which will not use in next loop.
So, i don't know my solution is correct or not.
image

2. I attempt publish test case vertify curry function will return function when it didn't reveive enough argument. How to test value  === function in ramda?

source/internal/_curryN.js
```diff
@@ -34,7 +34,14 @@ export default function _curryN(length, received, fn) {
       }
```
kedashoe, line 34: I think we can track this without the additional loop.

We can add a new var up top (say line 20), `var hasPlaceholderBeforeFinalArg = false;`

And then here after updating combined we can add an else if

```js
if (!_isPlaceholder(result)) {
  left -= 1;
} else if (combinedIdx < length) {
  hasPlaceholderBeforeFinalArg = true;
}
```
and then at the end, similar to your current check: `return (left <= 0 && !hasPlaceholderBeforeFinalArg) ... `

test/curry.js
```diff
@@ -119,6 +119,15 @@ describe('curry', function() {
...
   });
 
+  it('return value when collet value which is not placeholder', function() {
+    var f = function(a, b, c) { return [a, b, c]; };
+    var g = R.curry(f);
+    var _ = R.__;
+
+    eq(g(1)(_)(2, _, _)(3), [1, 2, 3]);
+    eq(g(1)(_, _, _)(2, _, _)(3), [1, 2, 3]);
+    eq(g(1)(_, _, _, 4)(2, _, _)(3), [1, 2, 3]);
```
kedashoe, line 129: Here are some more tests we can add

```
    eq(g(_, 2, 3, 4)(1), [1, 2, 3]);
    eq(g(1, 2, 3, _), [1, 2, 3]);
    eq(g(1)(_, 3, 4)(2), [1, 2, 3]);
    eq(g(_)(_, 2, 3, 4)(1), [1, 2, 3]);
    eq(g(_)(_, 2, _, 4, 5)(_, 3, 4)(1), [1, 2, 3]);
    eq(g(_)(_, 2, _, 4, 5)(1, 3), [1, 2, 3]);
```

kedashoe (changes requested):
Hi @manyuemeiquqi , thank you for the issue and PR! I've requested a couple changes, let me know what you think

test/curry.js
manyuemeiquqi: `g(_)(_, 2, _, 4, 5)(_, 3, 4)(1)`  This case will cause `_arity` get negative integer value, and the reason is combined get value more than `length` cause left < 0 
image
and I don't not how to handle it.

manyuemeiquqi: Shoulid i limt `left` min value ?

CrossEye:
I think there is a fundamental question here we need to answer before we settle on a solution.

What should we do with extra parameters?  Our answer has been to pass them along, but we hadn't considered this case of passing too many with the placeholder displacing them.  Now we have to make a more fundamental decision.  Which of these (at least somewhat reasonable) results should we expect?"

```js
const g = R.curry(function (a, b, c) {return [...arguments]})
g(_, 2, 3, 4)(_, 5, 6)(_, 7, 8, 9)(1, 10)
//?=> [1, 2, 3]
//?=> [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
//?=> [1, 2, 3, 10]
```

test/curry.js
kedashoe: ah I had another change I didn't mention locally to account for that.. but thinking about it some more maybe a better solution is:
```js
      if (combinedIdx < length) {
        if (!_isPlaceholder(result)) {
          left -= 1;
        }
        else {
          hasPlaceholderBeforeFinalArg = true;
        }
      }
```
We only care to update both `left` and `hasPlaceholderBeforeFinalArg` if we are less then than arity (`length`), so I think this should handle it? All our tests passed for me locally

kedashoe:
> The second one seems the most likely extension of how we do things in Ramda

I agree with this, the other cases are not important enough to introduce breaking changes

semmel:
>Or are there other possibilities? 

What about `[1,2,3,4]`?
Clearly the author of `g` wants to collect the excess parameters beyond `c`. Ramda's `curry` provides that feature which is fine. 

However, the position of the parameters in a function is important. (Here a-c). 
Should not have the excess parameters also the right position in the function call: Always following after the `c`?

I mean, your third case seems surprising in that the 10 does not follow the 3 in the function call.

The second case I like better, but it is still bad programming style to rely on such parameter collecting. Perhaps better not document it.

manyuemeiquqi:
I this [1,2,3,4] is not good.

CrossEye:
@kedashoe:

> The second one seems the most likely extension of how we do things in Ramda
> 
> I agree with this, the other cases are not important enough to introduce breaking changes

The trouble, of course, is that this is a long-standing bug, probably from when we first separated off the internal `curry2`, `curry3`, ... from the public `curry`, `curryN`.  It's really hard to know how people might have taken advantage of it over the years.  Or maybe that's not even possible.  Hmmm....

@semmel:

> What about `[1,2,3,4]`?

You're right.  That would be another candidate, and in some sense the inverse of `[1, 2, 3, 10]`.  One says, "Keep any extra parameters the first time they're collected, and include them in the final call".  The other says, "Ignore any extra parameters until you hit the final call, and then include them.  And although several times I'd wished we had done `[1, 2, 3]` -- "Ignore all extra parameters," I do think at this point we're better of with `[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]`.  The days of defaulted parameters that don't contribute to the arity seems to make this more reasonable.

When I'm in front a computer rather than my mobile phone, I'll test how this PR's code would handle that scenario.  Or @manyuemeiquqi, feel free to add a test case like this.

manyuemeiquqi:
> @kedashoe:
> 
> > 第二个似乎是我们在 Ramda 中做事方式的最有可能的延伸
> > 我同意这一点，其他情况还不够重要，不足以引入重大更改
> 
> 当然，问题在于这是一个长期存在的错误，可能是从我们第一次将内部`curry2`、`curry3`...与公共`curry`、 ...分开时开始的`curryN`。真的很难知道这些年来人们如何利用它。或者也许这根本不可能。嗯……
> 
> @semmel:
> 
> > 又怎样呢`[1,2,3,4]`？
> 
> 你说得对。那将是另一个候选者，并且在某种意义上是 的反面`[1, 2, 3, 10]`。有人说，“在第一次收集任何额外参数时保留它们，并将它们包含在最终调用中”。另一个人说，“忽略任何额外的参数，直到你完成最后的调用，然后将它们包括在内。尽管有好几次我希望我们已经这样做了`[1, 2, 3]`——“忽略所有额外的参数，”我确实认为在这一点上我们是更好的是 with `[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]`。默认参数对数量没有贡献的日子似乎使这更合理。
> 
> 当我面对电脑而不是手机时，我将测试此 PR 的代码如何处理该场景。或者@manyuemeiquqi，请随意添加这样的测试用例。

Well, i update pr with the expect of `[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]`. The thought of modify code of currN which i update is to satisify all placeholder.

kedashoe (changes requested):
Couple small changes, think we're almost there, ty @manyuemeiquqi

source/internal/_curryN.js
```diff
@@ -31,10 +32,14 @@ export default function _curryN(length, received, fn) {
       combined[combinedIdx] = result;
       if (!_isPlaceholder(result)) {
         left -= 1;
+      } else {
+        hasPlaceholder = true;
       }
       combinedIdx += 1;
     }
-    return left <= 0
+    if (left <= 0) {left = 0;}
```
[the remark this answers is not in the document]
manyuemeiquqi: Thanks for you check.

kedashoe (approved):
Looks good to me! Will give it a day or two for others to review before merging

kedashoe:
ty @manyuemeiquqi . I'll plan to release 0.29.1 at some point next week

PR #3417, opened 2023-10-24
Optimize R.path and dependents

Berndy:
My attempt at solving functions like path running slow due to using currying when not necessary and depending on functions that do more work than required.

## Results
```
path old x 2,060,753 ops/sec ±3.83% (22 runs sampled)
path new x 4,981,880 ops/sec ±0.92% (24 runs sampled)
```

## Optimization steps
- Own code for path function to avoid curry call and extra array when using paths
- moved nth functionality to internal to have access to an non curried version
- moved path functionality to internal to have access to an non curried version
- replacing calls of nth and path with non curried versions will also improve performance of dependent functions

## Curry vs no curry
In the future all functions should probably call an internal non curried version of the functionality. Not creating a new function on the fly and then calling it was a major part of these performance optimizations. Being able to use these non curried versions from the library would be amazing, although this would probably go against ramda's "everything is curried" policy.

```
path curried  x 4,981,880 ops/sec ±0.92% (24 runs sampled)
path internal x 8,866,690 ops/sec ±4.40% (23 runs sampled)
```

see [here](https://github.com/Berndy/ramda/blob/benchmark-path/benchmarkPath.md) for my benchmark experiment

### DIFF
--- a/source/internal/_nth.js
+++ b/source/internal/_nth.js
@@ -1,28 +1,5 @@
 import _isString from './_isString.js';
 
-/**
- * Returns the nth element of the given list or string. If n is negative the
- * element at index length + n is returned.
- *
- * @func
- * @sig (Number, [a]) -> a | Undefined
- * @sig (Number, String) -> String
- * @param {Number} offset
- * @param {*} list
- * @return {*}
- * @example
- *
- *      const list = ['foo', 'bar', 'baz', 'quux'];
- *      nth(1, list); //=> 'bar'
- *      nth(-1, list); //=> 'quux'
- *      nth(-99, list); //=> undefined
- *
- *      nth(2, 'abc'); //=> 'c'
- *      nth(3, 'abc'); //=> ''
- * @symb nth(-1, [a, b, c]) = c
- * @symb nth(0, [a, b, c]) = a
- * @symb nth(1, [a, b, c]) = b
- */
 function nth(offset, list) {
   var idx = offset < 0 ? list.length + offset : offset;
   return _isString(list) ? list.charAt(idx) : list[idx];
--- a/source/internal/_nth.js
+++ b/source/internal/_nth.js
@@ -1,7 +1,6 @@
 import _isString from './_isString.js';
 
-function nth(offset, list) {
+export default function _nth(offset, list) {
   var idx = offset < 0 ? list.length + offset : offset;
   return _isString(list) ? list.charAt(idx) : list[idx];
 }
-export default nth;
--- a/source/internal/_nth.js
+++ b/source/internal/_nth.js
@@ -1,0 +1,30 @@
+import _isString from './_isString.js';
+
+/**
+ * Returns the nth element of the given list or string. If n is negative the
+ * element at index length + n is returned.
+ *
+ * @func
+ * @sig (Number, [a]) -> a | Undefined
+ * @sig (Number, String) -> String
+ * @param {Number} offset
+ * @param {*} list
+ * @return {*}
+ * @example
+ *
+ *      const list = ['foo', 'bar', 'baz', 'quux'];
+ *      nth(1, list); //=> 'bar'
+ *      nth(-1, list); //=> 'quux'
+ *      nth(-99, list); //=> undefined
+ *
+ *      nth(2, 'abc'); //=> 'c'
+ *      nth(3, 'abc'); //=> ''
+ * @symb nth(-1, [a, b, c]) = c
+ * @symb nth(0, [a, b, c]) = a
+ * @symb nth(1, [a, b, c]) = b
+ */
+function nth(offset, list) {
+  var idx = offset < 0 ? list.length + offset : offset;
+  return _isString(list) ? list.charAt(idx) : list[idx];
+}
+export default nth;
--- a/source/internal/_path.js
+++ b/source/internal/_path.js
@@ -1,7 +1,7 @@
 import _isInteger from './_isInteger.js';
 import _nth from './_nth.js';
 
-function path(pathAr, obj) {
+export default function _path(pathAr, obj) {
   var val = obj;
   for (var i = 0; i < pathAr.length; i += 1) {
     if (val == null) {
@@ -17,5 +17,3 @@ function path(pathAr, obj) {
   }
   return val;
 }
-
-export default path;
--- a/source/internal/_path.js
+++ b/source/internal/_path.js
@@ -1,0 +1,43 @@
+import _isInteger from './_isInteger.js';
+import _nth from './_nth.js';
+
+/**
+ * Retrieves the value at a given path. The nodes of the path can be arbitrary strings or non-negative integers.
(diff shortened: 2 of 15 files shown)


source/internal/_path.js
```diff
@@ -0,0 +1,43 @@
...
+ *      path(['a', 'b'], {a: {b: 2}}); //=> 2
+ *      path(['a', 'b'], {c: {b: 2}}); //=> undefined
+ *      path(['a', 'b', 0], {a: {b: [1, 2, 3]}}); //=> 1
+ *      path(['a', 'b', -2], {a: {b: [1, 2, 3]}}); //=> 2
+ *      path([2], {'2': 2}); //=> 2
+ *      path([-2], {'-2': 'a'}); //=> undefined
+ */
+
+function path(pathAr, obj) {
+  let val = obj;
```
kedashoe, line 27: Let's just stick with `var` for now, no `let` or `const` except in comments.

source/internal/_nth.js
```diff
@@ -0,0 +1,30 @@
+import _isString from './_isString.js';
+
+/**
```
kedashoe, line 3: unless you feel there is something worth mentioning here, we don't need these docs in the internal definitions, I'd say remove them

source/pathEq.js
```diff
@@ -14,7 +14,7 @@ import path from './path.js';
  * @typedefn Idx = String | Int | Symbol
  * @sig a -> [Idx] -> {a} -> Boolean
  * @param {*} val The value to compare the nested property with
- * @param {Array} path The path of the nested property to use
+ * @param {Array} pathAr The path of the nested property to use
```
kedashoe, line 17: Not sure why this was updated, just `path` probably preferable? But I'm not sure what you were going for here with `pathAr`?

source/internal/_path.js
kedashoe, line 1: We can also update `source/lensPath.js` to use this

source/internal/_nth.js
kedashoe, line 1: We can also update `source/head.js`, `source/last.js` and `source/lensIndex.js` to use this

kedashoe (changes requested):
Thank you for the PR! Just a couple minor comments

source/internal/_nth.js
Berndy: You're right, the idea was that since I moved the entire logic to internal, I also copied the docs with it, but it is not necessary, so I removed it.

source/pathEq.js
Berndy: I changed it when i changed the parameter name in the function, but you're right, doesn't make sense to change the docs. 
However when trying to retrace my thought process, I discovered that there are a bunch of inconsistencies across the "path" related functions, both in docs and in fn param names. For example pathOr both documents path as "p", and compared to propOr, the default/or value is documented as "d" in one and "val" in the other. I think that goes beyond this PR though, so I might submit another later on to try and fix those.

source/internal/_path.js
Berndy: I initially left them as is, since they relied on currying anyway. I ran some test now, and it turns out creating a new function here instead of using the curried one still results in some significant performance gains when creating the lens. This also makes the case for internal lens and assocPath functions in later refactors.

```
lenspath old x 9,948,706 ops/sec ±2.01% (21 runs sampled)
lenspath new x 13,470,456 ops/sec ±1.87% (22 runs sampled)
```

source/internal/_nth.js
Berndy: Modified those as well, see comment for `lensPath`

Berndy:
Sidenote:
I did some experiments trying to optimize nth, as it's potentially called quite often by path, trying to change it from
```
export function _nth(offset, list) {
  var idx = offset < 0 ? list.length + offset : offset;
  return _isString(list) ? list.charAt(idx) : list[idx];
}

```

to this

```
export function _nth2(offset, list) {
  if (offset < 0) {
    const idx = list.length + offset;
    return _isString(list) ? list.charAt(idx) : list[idx];
  }

  return _isString(list) ? list.charAt(offset) : list[offset];
}

```

However while this was a pretty big failure, slowing down nth itself significantly

```
nth old 0 x 43,213,419 ops/sec ±5.71% (23 runs sampled)
nth new 0 x 26,711,312 ops/sec ±5.35% (20 runs sampled)
nth old 1 x 42,951,693 ops/sec ±6.16% (22 runs sampled)
nth new 1 x 26,652,017 ops/sec ±5.74% (22 runs sampled)
nth old -1 x 40,707,893 ops/sec ±0.56% (21 runs sampled)
nth new -1 x 24,465,761 ops/sec ±3.91% (21 runs sampled)
```

path seems to have some decent gains in performance with this "slower" version of nth

Path with old Nth function
```
path old      x 2,171,188 ops/sec ±1.04% (24 runs sampled)
path new      x 5,352,858 ops/sec ±2.10% (22 runs sampled)
path internal x 8,065,776 ops/sec ±0.52% (23 runs sampled)
```

path with new Nth function
```
path old      x 2,179,445 ops/sec ±0.81% (22 runs sampled)
path new      x 6,395,729 ops/sec ±3.69% (23 runs sampled)
path internal x 12,270,453 ops/sec ±2.42% (21 runs sampled)
```

see [here](https://github.com/Berndy/ramda/blob/benchmark-path/benchmarkNth.md)

source/internal/_path.js
kedashoe: > However I did discover that clone.js uses let instead of var, should we change that too or is there a reason for that?

I noticed that haha, snuck in at some point. No reason, we can update it as part of this PR or I can submit one later for the clone stuff.

The use of var is historically to support es3. Once we get to 1.0 we can refactor to use let/const

kedashoe:
> Any ideas what could cause this? I am both confused a switch from ternary to if would cause such a big hit in performance and why this would end up gaining performance when called in path...
> 
> see [here](https://github.com/Berndy/ramda/blob/benchmark-path/benchmarkNth.md)

I tried pulling that down and running the benchmarks but something on the branch is breaking the build for me. In any event, if a micro-optimization makes sense and the benchmarks look better, we can ship it. If either of those two conditions aren't met, we can hold off for now.

kedashoe (approved):
Will leave open for a bit to let others review

Berndy:
> I tried pulling that down and running the benchmarks but something on the branch is breaking the build for me. In any event, if a micro-optimization makes sense and the benchmarks look better, we can ship it. If either of those two conditions aren't met, we can hold off for now.

My bad, I mostly hacked those benchmarks together, which broke tests (and apparently the build). Exporting anything but default apparently breaks everything, so I moved all alternative versions of code to their own files, so it should work now.

I added a whole bunch of extra variations to narrow down the cause, and I see those performance gains ONLY when it's this exact code and ONLY when it's called via a function. 

I've switched babel node to see if it's runtime based, and now I get mostly similar results for all variations. So my latest theory is now that this is the result of some bun magic that maybe optimizes very specific function patterns...?

I guess in this case it doesn't really make sense to add this to the PR.

kedashoe:
yep np, ty for looking into it some more, agreed on your conclusion :+1:

PR #3420, opened 2023-11-04
chore: update vulnerable dependencies

### LINKED ISSUE #3419: update vulnerable dependencies
### Title

 - update vulnerable dependencies.

### Description
- There are 3 dependencies that are getting flagged because of security issues. These are `mocha`, `babel-core` & `handlebars`

### Current Behavior
- These dependencies get flagged off as vulnerable.

### Expected Behavior
- There should be no security issues in any of the dependencies.

ridge-kimani:
### Description
- updates packages with security issues. 

Fixes #3419  

### Type of change
- [ ] Chore(non-breaking change which adds functionality)

### How Has This Been Tested?
- Update the dependencies via `npm install`
- Run all tests to ensure there's no breakage


ridge-kimani:
Updating babel also updates the build file

dist/ramda.js
kedashoe, line 1: Could you remove this file from the PR? We only update this file at release time.

package.json
kedashoe, line 1: Looks like there are more recent versions of all these packages you've updated, does the build work with the more recent versions? Also could you include the `^` at the beginning as we do with most of our devDependencies?

kedashoe (changes requested):
ty for the PR @ridge-kimani ! A couple questions/changes.

dist/ramda.js
ridge-kimani: Yeah, I will remove it. It got rebuilt once I ran the npm update.

ridge-kimani: When running `npm i` it creates a new build so it might have some issues if I remove it.

package.json
ridge-kimani: Sure enough, thanks for the catch. I used to closest secure packages but with the latest ones it still works.

ridge-kimani:
Good idea @Berndy. I will audit the rest of the packages and check out if there's any breaking changes for the other packages.

ridge-kimani:
@kedashoe, I believe the `ramda.min.js` is important to be in this build because of the changes in Babel. This will make the installation process a smooth one otherwise if someone re-installs the packages it will be recompiled too.

ridge-kimani (commented):
@kedashoe, I believe the ramda.min.js is important to be in this build because of the changes in Babel. This will make the installation process a smooth one otherwise if someone re-installs the packages it will be recompiled too.

kedashoe (changes requested):
I'm seeing some errors in the tests now, could you fix those? I believe related to the sanctuary dependency update.

Also please remove `dist/ramda.js` from the commit. The files in `dist` are only updated when we cut releases. Thank you

Berndy:
I'm still getting an error when trying to run `npm run browser_test`, seemingly related to the sinon dependency. It seems we could revert that to the previous `^7.3.2` without getting any vulnerabilities back, however even then my browser_test still has a timeout issue, not sure if that's just my setup though.

Also we can still run `npm audit fix` to cut back on 16 of the remaining 22 vulnerabilities, seemingly without breaking anything further

ridge-kimani:
@Berndy Just updated it. 
RN we have `6 vulnerabilities (4 high, 2 critical)`

kedashoe (approved):
ty!

PR #3431, opened 2023-11-19
code climate fix

kedashoe:
This PR updates git actions to no longer use code climate. It hasn't been working for over a year at least in any event. For various pull request related actions (open, request review, etc), coverage will run and a comment will be created/updated on the PR.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,6 @@
 name: Build
 
-on: [pull_request]
+on: [pull_request_target]
 
 jobs:
   coverage:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,6 +7,12 @@ jobs:
     if: github.repository == 'ramda/ramda'
     runs-on: ubuntu-latest
     name: Coverage
+
+    permissions:
+      issues: write
+      contents: read
+      pull-requests: write
+
     steps:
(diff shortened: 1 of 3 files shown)


kedashoe:
@hitmands , looks like you helped out with the original PR adding code climate to ramda: https://github.com/ramda/ramda/pull/3033

Do you have any thoughts on why it is currently failing? 

The current error on master

```
/home/runner/work/_actions/paambaati/codeclimate-action/v2.6.0/node_modules/@actions/core/lib/command.js:66
        .replace(/%/g, '%25')
         ^

TypeError: (s || "").replace is not a function
```

I've updated some dependencies in this PR, and am getting

```
Error: you must supply a CC_TEST_REPORTER_ID ENV variable or pass it via the -r flag
```

I see code climate also has a github app? https://docs.codeclimate.com/docs/open-source-free. Maybe we can use that directly?

hitmands:
I see code climate also has a github app? https://docs.codeclimate.com/docs/open-source-free. Maybe we can use that directly?

- cc is now at `V5` https://github.com/marketplace/actions/code-climate-coverage-action

Then put `lcov` processing back, as it seems like cc is processing json (by looking at the job logs)

```yaml
coverageLocations: |
            ${{github.workspace}}/coverage/lcov.info:lcov
```

PR #3432, opened 2023-12-04
replaced let with var in _clone.js for consistency

Berndy:
We try to refrain from using `const` and `let` in favor of `var`, so I replaced some leftover `let` in _clone.js

### DIFF
--- a/source/internal/_clone.js
+++ b/source/internal/_clone.js
@@ -69,7 +69,7 @@ function _ObjectMap() {
 }
 
 _ObjectMap.prototype.set = function(key, value) {
-  const hashedKey = this.hash(key);
+  var hashedKey = this.hash(key);
 
   var bucket = this.map[hashedKey];
   if (!bucket) {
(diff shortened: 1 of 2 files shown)


kedashoe:
Thanks @Berndy . There are a couple `const` in `internal/_clone.js` as well, would you mind replacing those? There is also a `const` [here](https://github.com/ramda/ramda/blob/74cfca7122329f4c6472da1411dc2964e36232f3/source/internal/_path.js#L11).

Berndy:
Whoops, apparently I did not search properly, all `let` and `const` should be gone now?

kedashoe (approved):
looks good to me, ty

FernandoBasso:
> We try to refrain from using `const` and `let` in favor of `var`

What is the reason behind that preference?

kedashoe:
> What is the reason behind that preference?

Historically out of the box compatibility with es3. I believe the plan is to switch everything over to const/let after 1.0

Issue #2789, opened 2019-02-22
compose and composeWith have different signature


villesau:
Looks like [`compose`](https://ramdajs.com/docs/#compose) takes functions as an argument when [`composeWith`](https://ramdajs.com/docs/#composeWith) takes an array of functions. This is pretty confusing and some convention would really make sense here.

CrossEye:
You're right.  Perhaps it was a mistake to introduce `composeWith` with the array signature.  But the goal is after v1.0 to move `pipe` and `compose` to match this, and to stop shipping any variadic functions.

So the fix for now will probably be in documentation.  Are you interested in creating a PR to document this?  (It would probably involve changing the JSDoc for `composeWith` and `pipeWith`.)

villesau:
May I ask why is that the goal?

What comes to documentation, the behaviour is already documented in the examples. Or did you mean documenting the goal to move everything to use array syntax in 1.0?

CrossEye:
I meant simply a note that said something like, "Note that this is subtly different from the signature of `compse`.  `composeWith` accepts an *array* of functions, not a `arguments` object full of them."

It's very hard to document variadic functions properly.  I'm not yet proposing that we disallow variadic user functions, although I hope that such use-cases will be rare.  But I would like it if all Ramda functions had fixed arities; this to my mind makes it easier to think about them.  (It should also clean up some implementations.)

PR #3441, opened 2024-02-07
adds flow serial function application

semmel:
I find myself frequently using function pipelines for immediate values `h(g(f(x)))`:
```javascript
const val = pipe(
  () => x, f, g, h)()
```
It's tedious to write and the many parentheses are easy to mess up. 
Also I don't want to put the seed `x` in the end `pipe(f, g, h)(x)`, because `x` is far away from `f` and thus your eyes have to jump up and down reading the code.

Therefore I propose to include `flow` for serial function application, i.e. the same spirit as above
```javascript
const val = flow(x, g, h)
```
Such a function is provided elsewhere. E.g. in [fp-ts](https://gcanti.github.io/fp-ts/modules/function.ts.html#pipe) named `pipe`. There was even a [proposal to TC39](https://github.com/tc39/proposal-function-pipe-flow) for including it in JavaScript.

#### Naming:
Ramda's `R.pipe` is mostly named `flow` everywhere else (e.g. lodash), and the proposed `R.flow` is mostly named `pipe`. [See the post about function composition](https://jrsinclair.com/articles/2022/javascript-function-composition-whats-the-big-deal/#flow) by James Sinclair to maximise your confusion.

#### Details: 
For generality — if omitting all functions — one could provide `flow(x) ≡ x` as fp-ts does. Is that reasonable?

### DIFF
--- a/source/flow.js
+++ b/source/flow.js
@@ -1,11 +1,18 @@
 import applyTo from './applyTo.js';
-import tail from './tail.js';
+import _curry2 from './internal/_curry2.js';
 
 /**
- * Pipe the value of an expression into a pipeline of functions by performing
- * left-to-right serial function composition.
- * The first argument is the seed value;
- * the remaining arguments must be unary functions.
+ * Takes the value of an expression and applies it to a function
+ * which is the left-to-right serial composition of the functions
+ * given in the second argument.
+ *
+ * The functions in the pipeline should be unary functions.
+ *
+ * `flow` is helps to avoid introducing named functions with named arguments
+ * for computing the result of a function pipeline which depends on given initial values.
+ * Rather than composing a custom pipeline function `p = (_x, _y) => R.pipe(g(x), h(y), …)`
+ * which is only later needed once `z = p(x, y)`,
+ * the introduction of `p`, `_x` and `_y` can be avoided: `z = flow(x, [g, h(y),…]`
  *
  * In some libraries this function is named `pipe`.
  *
@@ -13,21 +20,18 @@ import tail from './tail.js';
  * @memberOf R
  * @since v0.30.0
  * @category Function
- * @sig (a, (a → b), …, (y → z)) → z
+ * @sig (a, [(a → b), …, (y → z)]) → z
  * @param {*} a The seed value
- * @param {...Function} functions
- * @return {*} z The result of the function pipeline
+ * @param {Array<Function>} pipeline functions composing the pipeline
+ * @return {*} z The result of applying the seed value to the function pipeline
  * @see R.pipe
  * @example
- *
- *      const m2 = R.flow(9, Math.sqrt, R.negate, R.inc); //=> -2
+ *      const defaultName = 'Jane Doe';
+ *      const savedName = R.flow(localStorage.get('name'), [R.when(R.isNil(defaultName)), R.match(/(.+)\s/), R.nth(0)]);
+ *      const givenName = R.flow($givenName.value, [R.trim, R.when(R.isEmpty, R.always(savedName))])
  */
-export default function flow() {
-  if (arguments.length === 0) {
-    throw new Error('flow requires at least one argument');
-  }
-  return tail(arguments).reduce(
-    applyTo,
-    arguments[0]
-  );
-}
+var flow = _curry2(function flow(seed, pipeline) {
+  return pipeline.reduce(applyTo, seed);
+});
+
+export default flow;
--- a/source/flow.js
+++ b/source/flow.js
@@ -1,5 +1,6 @@
 import applyTo from './applyTo.js';
 import _curry2 from './internal/_curry2.js';
+import _reduce from './internal/_reduce.js';
 
 /**
  * Takes the value of an expression and applies it to a function
@@ -8,11 +9,11 @@ import _curry2 from './internal/_curry2.js';
  *
  * The functions in the pipeline should be unary functions.
  *
- * `flow` is helps to avoid introducing named functions with named arguments
+ * `flow` is helps to avoid introducing an extra function with named arguments
  * for computing the result of a function pipeline which depends on given initial values.
- * Rather than composing a custom pipeline function `p = (_x, _y) => R.pipe(g(x), h(y), …)`
- * which is only later needed once `z = p(x, y)`,
- * the introduction of `p`, `_x` and `_y` can be avoided: `z = flow(x, [g, h(y),…]`
+ * Rather than defining a referential transparent function `f = (_x, _y) => R.pipe(g(_x), h(_y), …)`
+ * which is only later needed once `z = f(x, y)`,
+ * the introduction of `f`, `_x` and `_y` can be avoided: `z = flow(x, [g, h(y),…]`
  *
  * In some libraries this function is named `pipe`.
  *
@@ -28,10 +29,10 @@ import _curry2 from './internal/_curry2.js';
  * @example
  *      const defaultName = 'Jane Doe';
  *      const savedName = R.flow(localStorage.get('name'), [R.when(R.isNil(defaultName)), R.match(/(.+)\s/), R.nth(0)]);
- *      const givenName = R.flow($givenName.value, [R.trim, R.when(R.isEmpty, R.always(savedName))])
+ *      const givenName = R.flow($givenNameInput.value, [R.trim, R.when(R.isEmpty, R.always(savedName))])
  */
 var flow = _curry2(function flow(seed, pipeline) {
-  return pipeline.reduce(applyTo, seed);
+  return _reduce(applyTo, seed, pipeline);
 });
 
 export default flow;
--- a/source/flow.js
+++ b/source/flow.js
@@ -20,7 +20,7 @@ import _curry2 from './internal/_curry2.js';
  * @memberOf R
  * @since v0.30.0
  * @category Function
- * @sig (a, [(a → b), …, (y → z)]) → z
+ * @sig a → [(a → b), …, (y → z)] → z
  * @param {*} a The seed value
  * @param {Array<Function>} pipeline functions composing the pipeline
  * @return {*} z The result of applying the seed value to the function pipeline
--- a/source/flow.js
+++ b/source/flow.js
@@ -1,0 +1,31 @@
+import applyTo from './applyTo.js';
+import tail from './tail.js';
+
+/**
+ * Performs left-to-right serial function application. The first argument is the seed value;
+ * the remaining arguments must be unary functions.
+ *
+ * In some libraries this function is named `pipe`.
+ *
+ * @func
+ * @memberOf R
+ * @since v0.30.0
+ * @category Function
+ * @sig (a → (a → b), …, (y → z)) → z
+ * @param {*} a The seed value
+ * @param {...Function} functions
+ * @return {*} z The result of the function pipeline
+ * @see R.pipe
+ * @example
+ *
+ *      const m2 = R.pipe(9, Math.sqrt, R.negate, R.inc); //=> -2
+ */
+export default function flow() {
+  if (arguments.length <= 1) {
+    throw new Error('flow requires at least two arguments');
+  }
+  return tail(arguments).reduce(
+    applyTo,
+    arguments[0]
+  );
+}
--- a/source/flow.js
+++ b/source/flow.js
@@ -27,6 +27,8 @@ import _reduce from './internal/_reduce.js';
  * @return {*} z The result of applying the seed value to the function pipeline
  * @see R.pipe
  * @example
+ *      R.flow(9, [Math.sqrt, R.negate, R.inc]), //=> -2
+ *
  *      const defaultName = 'Jane Doe';
  *      const savedName = R.flow(localStorage.get('name'), [R.when(R.isNil(defaultName)), R.match(/(.+)\s/), R.nth(0)]);
  *      const givenName = R.flow($givenNameInput.value, [R.trim, R.when(R.isEmpty, R.always(savedName))])
--- a/source/flow.js
+++ b/source/flow.js
@@ -2,7 +2,9 @@ import applyTo from './applyTo.js';
 import tail from './tail.js';
 
 /**
- * Performs left-to-right serial function application. The first argument is the seed value;
+ * Pipe the value of an expression into a pipeline of functions by performing
+ * left-to-right serial function composition.
+ * The first argument is the seed value;
  * the remaining arguments must be unary functions.
  *
  * In some libraries this function is named `pipe`.
@@ -11,7 +13,7 @@ import tail from './tail.js';
  * @memberOf R
  * @since v0.30.0
  * @category Function
- * @sig (a → (a → b), …, (y → z)) → z
+ * @sig (a, (a → b), …, (y → z)) → z
  * @param {*} a The seed value
  * @param {...Function} functions
  * @return {*} z The result of the function pipeline
--- a/source/flow.js
+++ b/source/flow.js
@@ -18,11 +18,11 @@ import tail from './tail.js';
  * @see R.pipe
  * @example
  *
- *      const m2 = R.pipe(9, Math.sqrt, R.negate, R.inc); //=> -2
+ *      const m2 = R.flow(9, Math.sqrt, R.negate, R.inc); //=> -2
  */
 export default function flow() {
-  if (arguments.length <= 1) {
-    throw new Error('flow requires at least two arguments');
+  if (arguments.length === 0) {
+    throw new Error('flow requires at least one argument');
   }
   return tail(arguments).reduce(
     applyTo,
--- a/test/flow.js
+++ b/test/flow.js
@@ -8,7 +8,7 @@ describe('flow', function() {
     eq(R.flow.length, 2);
   });
 
-  it('performs left-to-right function composition', function() {
+  it('applies the first argument to a left-to-right composed function', function() {
     //  f :: [Number] -> [Number]
     var f2 = R.flow(4, [Math.sqrt, R.multiply, R.map]);
     var f3 = R.flow(9, [Math.sqrt, R.multiply, R.map]);
--- a/test/flow.js
+++ b/test/flow.js
@@ -1,0 +1,35 @@
+var assert = require('assert');
+
+var R = require('../source/index.js');
+var eq = require('./shared/eq.js');
+
+describe('flow', function() {
+
+  it('is a variadic function', function() {
+    eq(typeof R.flow, 'function');
+    eq(R.flow.length, 0);
+  });
+
+  it('performs left-to-right function composition', function() {
+    //  f10 :: [Number] -> [Number]
+    var f2 = R.flow(4, Math.sqrt, R.multiply, R.map);
+    var f3 = R.flow(9, Math.sqrt, R.multiply, R.map);
+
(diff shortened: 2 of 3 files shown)


kedashoe:
ty for the PR @semmel. I'm +1 on this, I always add it to projects I use ramda in.

> For generality — if omitting all functions — one could provide flow(x) ≡ x as fp-ts does. Is that reasonable?

Sure!

kedashoe:
@semmel I'm curious what you think about https://github.com/ramda/ramda/wiki/Ramda-Conventions#prefer-non-variadic-functions-over-variadic? I didn't write that, but I do agree and much prefer `flow(x, [f, g])` to `flow(x, f, g)`. We also don't have to worry about the weird flow-with-one-argument case discussed above, it can just use internal curry2.

semmel:
@kedashoe 
> @semmel I'm curious what you think about https://github.com/ramda/ramda/wiki/Ramda-Conventions#prefer-non-variadic-functions-over-variadic? I didn't write that, but I do agree and much prefer `flow(x, [f, g])` to `flow(x, f, g)`. 

*In general*, I prefer non-variadic functions too. (With variadic functions, I run in all sorts of problems using `lift`, `pipe` and others)
However, function composition is such a *fundamental* concept — there is even the famous [pipe operator `|>` proposal](https://github.com/tc39/proposal-pipeline-operator) waiting to be accepted for inclusion to the language. I don't see where the benefits for the library user would be, if Ramda would enforce `[]` around the composed functions?
Apart from easier TypeScript typings perhaps?

I've never yet handed `pipe` or `compose` as arguments to higher-order functions (e.g. like `partialRight`) to get back a customised pipelining function. Higher order functions have trouble with variadic functions of course. I am aware of that. Maybe I am missing something.

I did a little research how others do it:

Lib | left-to-right fn composition | pipelining a value left-to-right
----|------------------------------|---------------------------------
[crocks][cr] | `pipe::((a → b), …, (y → z)) → a → z` | n/a
Ramda | `pipe::(((a,…,n) → o), (o → p), …, (y → z)) → ((a,…,n) → z)` | `flow::(a, (a → b), …, (y → z)) → z`, `flow::a → a`[^1]
[Sanctuary][san] | `pipe::[(a → b), …, (y → z)] → a → z` | n/a
[ts_belt][tsb] | `flow::(((a,…,n) → o), (o → p), …, (y → z)) → ((a,…,n) → z)` | `pipe::(a, (a → b), …, (y → z)) → z`
[fp-ts][fpt] | `flow::(((a,…,n) → o), (o → p), …, (y → z)) → ((a,…,n) → z)` | `pipe::(a, (a → b), …, (y → z)) → z`, `pipe::a → a`
[lodash/fp][lof] | `pipe::(((a,…,n)→o),(o→p),…,(y→z)\|[((a,…,n)→o),(o→p),…,(y→z)])→((a,…,n)→z)` | n/a
[lodash][lo] | `flow::(((a,…,n)→o),(o→p),…,(y→z)\|[((a,…,n)→o),(o→p),…,(y→z)])→((a,…,n)→z)` | n/a

Except Sancturary (and lo-dash optional), *all* offer function composition using only variadic functions. I wonder why that is, if fixed-arity functions are usually easier to handle?
Also should Ramda be different (e.g. more inviting for newcomers) from Sanctuary, otherwise one could use Sanctuary anyway.

For personal reasons, I don't like the extra `[]` syntactic noise. Also I shy away from breaking changes — if we'd go that path `R.compose` and `R.pipe` would need to become fixed-arity too?

[cr]: https://crocks.dev/docs/functions/helpers.html#pipe
[san]: https://sanctuary.js.org/#pipe
[tsb]: https://mobily.github.io/ts-belt/api/pipe-flow#pipe
[fpt]: https://gcanti.github.io/fp-ts/modules/function.ts.html#flow
[lof]: https://github.com/lodash/lodash/wiki/FP-Guide
[lo]: https://lodash.com/docs/4.17.15#flow
[^1]: proposed


> We also don't have to worry about the weird flow-with-one-argument case discussed above, it can just use internal curry2.

I changed the signature to `flow :: (a, (a → b), …, (y → z)) → z`. I don't understand where currying would be beneficial for `flow`, but I've not thought about that enough…

mendrik:
i think ramda should stay data-last, there are many libs that do that flow thingy.

semmel:
> i think ramda should stay data-last, there are many libs that do that flow thingy.

On can consider `flow` a generalisation of `R.applyTo` which is also not data-last.

Other libs that provide `flow` cover the same ground as Ramda does. I'd dislike to be dependent on two FP libraries with 90% overlapping use cases.

kedashoe:
> For personal reasons, I don't like the extra `[]` syntactic noise.

I like the noise, it alerts readers (we read code more than write it!) that something is different between first and second args.

```js
let x = flow(foo, [
  R.map(blah),
  R.join('-'),
  R.etc(a, b),
  x => ...
]);
```

vs

```js
let x = flow(foo,
  R.map(blah),
  R.join('-'),
  R.etc(a, b),
  x => ...
);
```
I really never write single line `flow(x, f, g)`, but even then I like the delineation [] provides.

I'm not worried about retrofitting pipe and compose. We could cross that bridge if needed at a later date.

The point about currying wasn't that it is useful in any way, just that we don't have to think about it, it would then be like most every other ramda function.

bunglegrind:
In short `flow(x, f1, f2, ...) === applyTo(x, pipe(f1, f2,...))` 

I agree with @kedashoe , they array sounds better, because of the separation between the different arguments.

CrossEye:
While I'm not involved enough these days to stand in the way, I am not a fan of this.  Ramda is about functional composition.  Combining that in the library with function application confuses its mission.

There is also something of a mismatch with our other composition functions, which do not insist that the first-applied function be unary.  That is `compose(inc, multiply(3), add)` is *binary* since `add` is binary.  There doesn't seem to be a good way around that mismatch for the suggested function.  We can only use this with single values and an initial unary function.

> Ramda's `R.pipe` is mostly named `flow` everywhere else (e.g. lodash), and the proposed `R.flow` is mostly named `pipe`. [See the post about function composition][js] by James Sinclair to maximise your confusion.

For that reason, if we do include this, I would stay far away from the name `flow`.  I'm sure we can find something more appropriate.

>  I don't see where the benefits for the library user would be, if Ramda would enforce [] around the composed functions?

A few years ago, [I wrote][ss]:

> But the goal is after v1.0 to move pipe and compose to match this, and to stop shipping any variadic functions.

I still think that would be great.  I don't mind the extra syntactic noise of `[]` to achieve a zero-variadic function library.

> On can consider `flow` a generalisation of `R.applyTo` which is also not data-last.

On the contrary, the whole rationale for `applyTo` is to handle the case where the transformation function *is* the data.

```js
[add(5), multiply(3), inc, dec].map(applyTo(10)) //=> [15, 30, 11, 9]
```

>> For personal reasons, I don't like the extra `[]` syntactic noise.
> I like the noise, it alerts readers (we read code more than write it!) that something is different between first and second args.

And we do have prior art at least in `useWith` and `converge`.  And of course `pipeWith`/`composeWith` are written this way.

----------

As I said, I won't stand in the way, if there is a real desire.  But I *will* fight hard to avoid adding new variadic functions.

  [js]: https://jrsinclair.com/articles/2022/javascript-function-composition-whats-the-big-deal/#flow
  [ss]: https://github.com/ramda/ramda/issues/2789#issuecomment-466720359

semmel:
Thanks for the good discussion, let me address some points:
>  Ramda is about functional composition. Combining that in the library with function application confuses its mission.

My use case is to set up event driven workflows using reactive *lazy* streams, promises, futures, bla bla bla …
Those "workflows" are usually functions, which get *executed* elsewhere using something like `fork` or `runEffects` which — of course — are not part of Ramda.
```javascript
// :: Stream Either String
const loginWorkflow = pipe(
  () => clickEventStream('#loginButton'),
  samples(inputChangeEventStream('#passwordTextField')),
  chain(pwd => fetch(`https://myapi/login?pwd=${pwd}`)),
  map(response => response.ok ? Right(response): Left(new Error())),
  tap(either(paintError, paintSuccess)),
  map(map(prop('accessToken'))
)();
const uploadWorkflow = pipe(
  () => clickEventStream('#uploadButton'),
  samples(loginWorkflow),
  chain(token => fetch("https://myapi/upload?${token}"),
// …
)();
// Up to now nothing gets executed. The workflows itself are functions …
// This runs both workflows (function execution)
runEffects(uploadWorkflow)
```

So it is really composing functions which disguise as lazy promises or streams. In my understanding I am not *applying* values setting up the pipelines.

My motivation for this PR is selfish and of *syntactic* nature: 
I really hate writing and reading `pipe(() => …)()` or `applyTo(… , pipe(…))`, when `flow(…)` (or maybe `flow(…, […])`) seems sufficient.

Maybe I should've mentioned this earlier. This is also the reason I linked to the related [`|>` proposal ](https://github.com/tc39/proposal-pipeline-operator); many think there is a need to improve the language.

>> For personal reasons, I don't like the extra [] syntactic noise.

>I like the noise, it alerts readers (we read code more than write it!) that something is different between first and second args.

I can follow that argument: That array includes only unary functions. Even if I consider the argument weak: Most Ramda functions need different arguments based on their position. 

> A few years ago, [I wrote](https://github.com/ramda/ramda/issues/2789#issuecomment-466720359):

>>But the goal is after v1.0 to move pipe and compose to match this, and to stop shipping any variadic functions.
>> … 
>> But I would like it if all Ramda functions had fixed arities; this to my mind makes it easier to think about them. 

>I still think that would be great

@CrossEye I still don't and I would not want to lay the basis for that change with `flow(x, [fns])`

- learning `o(f, g) ≡ f ∙ g` which is fine and fundamental, one concludes that `f ∙ g ∙ h` would be `o(f, g, h)` or `compose(f, g, h)`. In `compose([f, g, h])` the need for an array after 2 arguments comes as a surprise.
- The last function in `compose` and the first in `pipe` may be (n>1)-ary functions. So they don't quite fit.
- this breaks *all* my source files

> On the contrary, the whole rationale for applyTo is to handle the case where the transformation function is the data.

I did not get that yet, give me some time to think about your example.

> I would stay far away from the name flow. I'm sure we can find something less confusing.

Yes, I agree!

To align with the status-quo I'd even consider swapping the names. (Has anybody an idea where "flow" and "pipe" originate?)

CrossEye:
> Maybe I should've mentioned this earlier. This is also the reason I linked to the related [`|>` proposal ](https://github.com/tc39/proposal-pipeline-operator); many think there is a need to improve the language.

And that's what I was mostly reacting to.  Pipe proposals have a [long, sad history][ph].  I really preferred the F#-style ones, which have now basically been abandoned, and am much less impressed with the Hack-style ones.  But what I wanted most of all is `compose`/`pipe` *operators*, which doesn't seem to have a real chance.

I haven't seen your usage before.  Am I understanding correctly that this is designed to do something like what RxJs or Bacon does? 

> Also I don't want to put the seed `x` in the end `pipe(f, g, h)(x)`, because `x` is far away from `f` and thus your eyes have to jump up and down reading the code.

Yes, I always prefer named functions here (`foo = pipe(f, g, h); foo(x)`), but I can understand that's not for everyone.


>> On the contrary, the whole rationale for `applyTo` is to handle the case where the transformation function is the data.

> I did not get that yet, give me some time to think about your example.

It didn't help that I mixed Ramda and Array.prototype syntaxes.  That really should have been

```js
map(applyTo(10))([add(5), multiply(3), inc, dec]) //=> [15, 30, 11, 9]
```

or

```js
map(applyTo(10), [add(5), multiply(3), inc, dec]) //=> [15, 30, 11, 9]
```

The point here is that this is a counterpart to `call`/`apply`.  When used curried, `call` and `apply` treat a transformation function as something that we want to bake in, and then to which we later want to apply some value.  `applyTo` treats some value as what we want to bake in; later we'll apply some function(s) to that value.  I always discourage people from describing Ramda as "data-last"; rather, "the parameters more likely to change come after those less likely."  So with `applyTo`, we're thinking of the integer value, `10` as the static thing, not likely to change, while the function varies.


> To align with the status-quo I'd even consider swapping the names. (Has anybody an idea where "flow" and "pipe" originate?)

I came up with `pipe`, as I thought `pipeline` both too long and noun-y.  I don't recall if I considered `flow`, but I still prefer `pipe`: `flow` feels focused on the data running through, where `pipe` is about how the functions are connected.  But I guess if you read `pipe` as a noun, then it's not very compelling.  I always think of it as a verb.

Lodash saw Ramda as a competitor, and aimed to kill it by duplicating its core features.  (More charitably, lodash saw how awesome some Ramda ideas were.  But I'm not feeling particularly charitable today.)  But for whatever reason, they didn't like the name `pipe` and called it `flow`, which is ok.  They also named their version of `compose` as `flowRight`, which is horrible.



  [ph]: https://github.com/tc39/proposal-pipeline-operator/blob/main/HISTORY.md

semmel:
@CrossEye Thanks for your input — that helped me again to think about my motivation with this function "flow"!

I've revised this PR, so that the proposed function has now a fixed arity and is binary-auto-curried. 
`flow :: a → [(a → b), …, (y → z)] → z` This should address the greatest concern of avoiding variadic functions.

I made the description quite wordy, and may need help with the correct formulation.
> `flow` is helps to avoid introducing named functions with named arguments
for computing the result of a function pipeline which depends on given initial values.
Rather than composing a custom pipeline function `p = (_x, _y) => R.pipe(g(x), h(y), …)`
which is only later needed once `z = p(x, y)`,
the introduction of `p`, `_x` and `_y` can be avoided: `z = flow(x, [g, h(y),…]`

I also changed the example to be more suitable.

We'd still need to agree on a name other than "flow". 

Apart from that — are there any other concerns I should address? 

@CrossEye Concerning your last remarks:
> I haven't seen your usage before. Am I understanding correctly that this is designed to do something like what RxJs or Bacon does?

Yes, indeed! Nowadays I prefer [@most/core](https://mostcore.readthedocs.io/en/latest/index.html) which is a little less surprising because there are just lazy streams (except when explicitly stated). That pipelines I construct using "flow" are basically just event listeners glued together. 
 
I agree with you on all other points. Especially that the JS pipe proposal is doomed since it will be not the F# flavour. 

Heck — maybe I will be lucky one day and write [F# for the browser](https://fable.io) 🤞🏻

source/flow.js
```diff
@@ -0,0 +1,37 @@
...
+ * @category Function
+ * @sig a → [(a → b), …, (y → z)] → z
+ * @param {*} a The seed value
+ * @param {Array<Function>} pipeline functions composing the pipeline
+ * @return {*} z The result of applying the seed value to the function pipeline
+ * @see R.pipe
+ * @example
+ *      const defaultName = 'Jane Doe';
+ *      const savedName = R.flow(localStorage.get('name'), [R.when(R.isNil(defaultName)), R.match(/(.+)\s/), R.nth(0)]);
+ *      const givenName = R.flow($givenName.value, [R.trim, R.when(R.isEmpty, R.always(savedName))])
```
CrossEye, line 31: I think this is far too complex for an example function.

Can we just use something more like the test cases?:

```js
 R.flow(9, [Math.sqrt, R.negate, R.inc]), //=> -2
```

or 

```js
const myFn = R.flow(9, [Math.sqrt, R.multiply, R.map])
myFn ([1, 2, 3]) //> [3, 6, 9]
```

test/flow.js
```diff
@@ -0,0 +1,21 @@
...
+var eq = require('./shared/eq.js');
+
+describe('flow', function() {
+
+  it('is a binary function', function() {
+    eq(typeof R.flow, 'function');
+    eq(R.flow.length, 2);
+  });
+
+  it('performs left-to-right function composition', function() {
```
CrossEye, line 11: well, it does, but it also calls the resulting function with a value.

CrossEye (changes requested):
As I said, I will not stand in the way of this, if people want it.  But I'm not convinced.

source/flow.js
FernandoBasso: As a non FP or Ramda expert, I find examples beyond “hello world” valuable and useful to give me ideas on how to use and compose/combine stuff together.

For example, let’s consider documentation for `identity` like this:

```javascript
identity(1);
//=> 1
```

It doesn’t help me understand or get ideas for more real situation where I could use it.

On the other hand, I understand it is not any library’s job to teach people how to program. The examples in JSDoc (which I believe are extracted to be turned in the documentation website) must be simple and concise, I agree, and they can’t possible cover all use cases anyway.

Still, it could be useful to place those more complex and/or real life examples in some other place, maybe in the [Cookbook](https://github.com/ramda/ramda/wiki/Cookbook) instead of just throwing them away.

I’m not sure if this idea makes much too much sense, but as a person learning FP and Ramda, I thought I’d mention it anyway.

semmel: > It doesn’t help me understand or get ideas for more real situation where I could use it.

Indeed, and that is exactly why I've attempted a more real example.

For me the cookbook is seldom the solution and only in desperate cases. It is far too wide-stretched and good for learning, but bad for look-up. (Before filtering the correct solution from the vast examples of the cookbook one could probably go faster using imperative-style mix-in JS)

Btw. another API example which leaves me confused is `R.ap` with arrays.

source/flow.js
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Takes the value of an expression and applies it to a function
+ * which is the left-to-right serial composition of the functions
+ * given in the second argument.
+ *
+ * The functions in the pipeline should be unary functions.
+ *
+ * `flow` is helps to avoid introducing named functions with named arguments
+ * for computing the result of a function pipeline which depends on given initial values.
+ * Rather than composing a custom pipeline function `p = (_x, _y) => R.pipe(g(x), h(y), …)`
```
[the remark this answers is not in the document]
semmel: > I don't understand this definition. Why `_x` and `_y` (with the underscores)? They're not used, right?

Yeah sorry, the example was wrong. It's fixed now.
I want to illustrate that you don't have to introduce an extra referential-transparent function (nobody likes external non-local inputs). In order to make that extra function referential transparent, you *have to* name the arguments.

kedashoe:
This looks good to me, I'm happy with the name flow. Can we merge this @CrossEye ?

semmel:
Regarding the name, I would add that in Ram**b**da(X)  it is named ["piped"](https://github.com/selfrefactor/rambdax#piped). Could be, some like "piped" better 🤷‍♂️

kedashoe:
> Regarding the name, I would add that in Ram**b**da(X) it is named ["piped"](https://github.com/selfrefactor/rambdax#piped). Could be, some like "piped" better 🤷‍♂️

I prefer `flow` personally, `pipe` vs `piped` too close. Are you ok with `flow` @semmel ?

semmel:
> I prefer `flow` personally, `pipe` vs `piped` too close. Are you ok with `flow` @semmel ?

Yes

CrossEye:
I've had my say.  If I were more active, I'd push back harder.  I don't think this is a particularly good fit for Ramda, and I think the name `flow` is going to be confused with lodash.

But these days the choice isn't mine and I'll defer to the wisdom of those currently active.

kedashoe:
ty @semmel , will try to get a release out this weekend

semmel:
> ty @semmel , will try to get a release out this weekend

@kedashoe Great! Thank you very much!

The prospect that we might soon get to v1.0 is even better, because then I can lobby at work in my new team to adopt Ramda 🤞

PR #3446, opened 2024-03-23
fix: splitWhenever jsdoc

Maestro31:
Despite this existing [PR](https://github.com/ramda/ramda/pull/3288), it seems that the merge has not taken place and that the correction is still necessary.


kedashoe:
ty @Maestro31

PR #3448, opened 2024-04-05
Support Objects aka dictionaries as Traversables

semmel:
As implemented elsewhere:

- in the [`Map` type](https://mostly-adequate.gitbook.io/mostly-adequate-guide/appendix_b#map) of Prof. Frisby's FP Guide 
- in the [`Record` type](https://gcanti.github.io/fp-ts/modules/Record.ts.html#traverse) of [fp-ts](https://gcanti.github.io/fp-ts/)

### DIFF
--- a/source/sequence.js
+++ b/source/sequence.js
@@ -4,12 +4,17 @@ import map from './map.js';
 import prepend from './prepend.js';
 import reduceRight from './reduceRight.js';
 import identity from './internal/_identity.js';
+import _isArray from './internal/_isArray.js';
+import toPairs from './toPairs.js';
+import _assoc from './internal/_assoc.js';
 
 
 /**
  * Transforms a [Traversable](https://github.com/fantasyland/fantasy-land#traversable)
  * of [Applicative](https://github.com/fantasyland/fantasy-land#applicative) into an
  * Applicative of Traversable.
+ *
+ * Also accepts `Object` as the Traversable to aid working with [dictionaries](https://github.com/ramda/ramda/wiki/Type-Signatures#simple-objects) (Objects of like-typed values).
  *
  * Dispatches to the `"fantasy-land/traverse"` or the `traverse` method of the second argument, if present.
  *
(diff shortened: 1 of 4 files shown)


kedashoe:
https://github.com/ramda/ramda/issues/1067

I'm in favor of this, but maybe not before 1.0

semmel:
TLDR: You can shoot yourselves in the foot with JavaScript, why not with Ramda too?

I just found #2580 which dealt basically with the same problem 6 years ago. That is really upsetting.

That PR got bikeshedded discussing the incompatibility of `R.equals` (which ignores key order) and a possible `R.reduce` for `Object` (which would observe key order and lead to different results unless the reducer function is commutative and associative).
Pretty everyone agreed that key order does not matter in Object, and that its not an issue traversing `Object` but the PR was never merged.

I guess people have moved on, since still today I cannot traverse a dictionary `Object` — which is pretty basic when you work with Maybe, Either and such data types.

So if PRs don't cut it, perhaps Ramda has hit an implicit barrier in supporting certain FP features?

The low entry barrier and no hand-holding attitude have been always the strengths of Ramda: You can solve problems quickly using
- JS-native data structures,
- suitable ADTs with Fantasyland support or not,
- even the pesky `Promise`.

out of personal frustration:
I think the fact, that Ramda will probably never reach `v1.0` just reflects the fact that one can never fulfill all what is promised  in those "FP using JavaScript" articles?

kedashoe:
I hear you. As I said and as I argued in those earlier threads, I 100% think we should add Traversable for objects. But it does not feel right to me to have it 1.0, and I worry we'll get into another endless bikeshed if we try.

But we'll get to 1.0, and I think soon. I like getting `flow` in for it. Whatever else is on main. Do a v0.30. Then do v1.0 shortly after. The work @Harris-Miller is leading on the typescript side is phenomenal and deserves our continued vigilance here.

Harris-Miller:
🤔

Plain objects are inherently treated as Functors for `map()` and as Filterables for `filter()`. Are they foldable? Represented as an array via `toPairs()` they are. If we were to add inherent support to `reduce` for objects it would essentially be `compose(reduce, toPairs)` since you'd have to represent it as a key/value tuple pair. And given that you could implement both `map()` and `filter()` from that... I would say yes, objects are foldable. Plain objects have an `empty` (`{}`) and `concat` (`Object.assign`). And via inherent support from `ramda`, implement `equals` as well.

Seems reasonable to give the same support for Traversables, IMO

Additionally, [Data.Map.traverseWithKey](https://hackage.haskell.org/package/containers-0.7/docs/Data-Map-Internal.html#v:traverseWithKey) does exactly what is being asked here as well.

PR #3459, opened 2024-05-02
JSDoc:  update `head()`, `last()`, and `nth()` to reflect actual behavior

Harris-Miller:
I noticed this while testing https://github.com/ramda/types/pull/101 which effects the typings for `last()`

`R.last(''); //=> ''` is not true, `R.last(''); //=> undefined` is

Looking at the git-history of `head.js`, `last.js` and `nth.js`. This has always been the behavior, the documentation was incorrect from the get-go. I think it was a bad copy/paste assumption in that `init()` and `tail()` both return `''` when given `''`

See in [repl](https://ramdajs.com/repl/?v=0.30.0#?R.head%28%27%27%29%3B%20%2F%2F%3D%3E%20undefined%2C%20not%20%27%27%0AR.last%28%27%27%29%3B%20%2F%2F%3D%3E%20undefined%2C%20not%20%27%27%0AR.nth%283%2C%20%27abc%27%29%3B%20%2F%2F%3D%3E%20undefined%2C%20not%20%27%27%0A)

### DIFF
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -2979,7 +2979,7 @@
    * @since v0.1.0
    * @category List
    * @sig [a] -> a | Undefined
-   * @sig String -> String
+   * @sig String -> String | Undefined
    * @param {Array|String} list
    * @return {*}
    * @see R.tail, R.init, R.last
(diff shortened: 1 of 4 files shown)


kedashoe (approved):

kedashoe:
Can you remove the changes to `dist`? We do that when we put out the release

kedashoe (approved):

PR #3461, opened 2024-05-02
Update: `curryN` utilize optimized _curry1, _curry2, _curry3

Harris-Miller:
We got em, might as well use em


Harris-Miller:
I'm unclear as to why the unit tests are failing

kedashoe:
We keep the `this` context and pass along extra arguments to functions users give to us. Those internal _curry2 and _curry3 functions do not, so this would be a pretty big breaking change.

Harris-Miller:
Ah I see. I did not notice that in the the `_curry1` and `_curryN` implementations. I could update `_curry2` and `_curry` to use `fn.call(this, ...)` and retain that behavior. Should work. Will give it a shot later this week

Harris-Miller:
Tried with `fn.call()`, that fixed it. But those optimized `_curry2` and `_curry3` handles specific arity where `_curry1` and `_curryN` work with additional args. Not worth trying to add that in. Closing

PR #3462, opened 2024-05-02
Chore: make Flow jsdoc example runnable in the repl

Harris-Miller:
```typescript
R.flow(9, [Math.sqrt, R.negate, R.inc]), //=> -2

const defaultName = 'Jane Doe';
const savedName = R.flow(localStorage.get('name'), [R.when(R.isNil(defaultName)), R.match(/(.+)\s/), R.nth(0)]);
const givenName = R.flow($givenNameInput.value, [R.trim, R.when(R.isEmpty, R.always(savedName))])
```
First line ends with `,` not `;` causing a parsing errors
`localStorage.get` is not available in the repl's environment
`$givenNameInput` is an undeclared variable

Updated to a similar example that demonstrates the same behavior that has no errors
```typescript
R.flow(9, [Math.sqrt, R.negate, R.inc]); //=> -2

const personObj = { first: 'Jane', last: 'Doe' };
const fullName = R.flow(personObj, [R.values, R.join(' ')]); //=> "Jane Doe"
const givenName = R.flow('    ', [R.trim, R.when(R.isEmpty, R.always(fullName))]); //=> "Jane Doe"
```

### DIFF
--- a/dist/ramda.js
+++ b/dist/ramda.js
@@ -4970,11 +4970,11 @@
    * @return {*} z The result of applying the seed value to the function pipeline
    * @see R.pipe
    * @example
-   *      R.flow(9, [Math.sqrt, R.negate, R.inc]), //=> -2
+   *      R.flow(9, [Math.sqrt, R.negate, R.inc]); //=> -2
    *
-   *      const defaultName = 'Jane Doe';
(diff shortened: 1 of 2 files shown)


kedashoe:
Same for this one, don't need to update dist

kedashoe (approved):

Issue #3458, opened 2024-05-01
Since v0.30 functions curried by Ramda break when used with `R.__` from another Ramda version 


semmel:
### Conditions
Your project uses a certain version of Ramda, say `v0.29.1`. 
You require/import a library which depends on a version of Ramda >= `v0.30`. This prevents deduplication of Ramda, so your project depends on two versions now.

![Screenshot 2024-05-01 at 17 24 04](https://github.com/ramda/ramda/assets/296476/ebc4e607-751c-49ff-b786-d983fe2b06a9)

In your project you use the `v0.29.1` placeholder `R.__` on a curried value-returning function imported from the dependent library.

### Error
When invoking the function with the last remaining argument you'll get an exception like "fn(...)(...) is not a function".
This is because `R.__` is not recognised as placeholder and simply delivered as argument to the original function. Thus the last `(...)` instruction tries to call a value.

### Demo

```javascript
// lib030/index.js and lib028/index.js
const 
  { add, curry } = require('ramda'),  // v0.30 or v0.28 resp.
  med = curry((a, b) => 2 / (1 / a + 1 / b));
  
  module.exports = {
    add, med
  };

// proj029/index.js
const 
  { __ } = require('ramda'),  // v0.29.1
  { add: add030, med: med030 } = require('lib030'),
  { add: add028, med: med028 } = require('lib028');

med030(__)(4)(1); // throws "… not a function"
add030(__)(1)(2); // throws "… not a function"

// importing from older Ramda versions works:
console.assert(med028(__)(4)(1) === 1.6);
console.assert(add028(__)(1)(2) === 3);
```

### Reason for the behaviour
In #3388 we switched from duck-typing to  identity (`===`) comparison in `isPlaceholder`. The placeholder from a particular version of Ramda is no longer recognised by the `curry` function of another Ramda version.

### Previous behaviour
With `v0.29.1` and below, cross-version compatibility is given.

Harris-Miller:
Ohh good catch. I think just this part needs to be reverted: https://github.com/ramda/ramda/pull/3388/files#diff-cfb81eb9592aa5597e5aa32c6e449214f5b7eabeb172ad2447e2e537232454d2L2-R4

PR #3463, opened 2024-05-02
Fix: cross-version placeholder support

Harris-Miller:
Closes https://github.com/ramda/ramda/issues/3458

The strict equality operator means that `_isPlacerholder` works for only the `R.__` from _that_ specific install. If your project has multiple versions of ramda (which can easily happen due to npm semver resolution and `^` versus `~` versus exact in `package.json`), you'll have problems.

https://github.com/ramda/ramda/pull/3388/files#diff-cfb81eb9592aa5597e5aa32c6e449214f5b7eabeb172ad2447e2e537232454d2L2-R4

Reverting the logic back to what it was before is a looser check that will work cross-version

### DIFF
--- a/source/__.js
+++ b/source/__.js
@@ -1,4 +1,3 @@
-import _placeholder from './internal/_placeholder.js';
 /**
  * A special placeholder value used to specify "gaps" within curried functions,
  * allowing partial application of any combination of arguments, regardless of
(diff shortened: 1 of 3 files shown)


kedashoe:
I believe we can remove [this file](https://github.com/ramda/ramda/pull/3388/files#diff-da208bc9e56b2a5cdb43bb9ab873ff73273836158f1ae97324b7b5b6c8cad42a) as well.

harshagarwal-mono:
I know this PR is merged but still we can do one more thing that is too keep both the check. So that if anyone using the latest version he gets the benifit of === and for backward compatibility lets use already existing.

PR #3465, opened 2024-05-02
Draf: New: `lensIso`

Harris-Miller:
I found myself having to implement this on my own recently in a project at work. It's useful to have alongside `lens` and has a distinct implementation that isn't a simple composition of other functions. So figured it was worth adding to the core lib

TODO:
- [ ] Unit tests


Harris-Miller:
@kedashoe @CrossEye was hoping to get feedback on accepting the proposal to add this function before I put in the effort of writing unit tests. Thanks in advance

kedashoe:
Can achieve the same with `lens(getter, unary(setter))`? Do you have an example that works differently from `lens`? I think the example in the docstring works out to the same thing `lens` vs `lensIso`.

Harris-Miller:
I didn't think to use `unary()` when I did this at my job. And to my surprise, the typings worked out as you'd expect it to with `lens(getter, unary(setter))`

Ref: https://tsplay.dev/WP2BJN

The only reason I can see keeping this then would be for convenience and exposure. I can adjust the file to use `lens(getter, unary(setter))` as well to cut down on duplicate code. Let me know

kedashoe:
We have https://github.com/ramda/ramda-lens which feels like where it belongs. But of course that hasn't been updated in many years. I reckon mull it over for a bit and if you want to bring it in or others chime that they'd like it we can work to get it in.

Harris-Miller:
I decided to finish it. Seems like it could be useful

PR #3474, opened 2024-06-18
Update docs for ascendNaturally and descendNaturally

Harris-Miller:
While working on https://github.com/ramda/types/pull/127 I noticed that the block comment docs are a bit off because of how these are implemented

https://github.com/ramda/ramda/pull/3442/files#diff-2f0444491e08d72d422c4e936a1e621a9a3fdb85d20da37118526f4df5bb5ddaR35

`.localeCompare` only exists on the prototype of `string`. This means that the compare function for `ascendNaturally` and `descendNaturally` need to return `String`, and not `Ord`. `Ord` is defined as `number | string | boolean | Date`, which are the only types in javascript you can use `<`, `>`, `<=`, and `>=` on. This makes since for `ascend` and `descend` of course, but trying those with `ascendNaturally` and `descendNaturally` will give you a runtime error

### DIFF
--- a/source/ascendNatural.js
+++ b/source/ascendNatural.js
@@ -9,9 +9,9 @@ import curry from './curry.js';
  * @memberOf R
  * @since v0.30.1
  * @category Function
- * @sig Ord b => s -> (a -> b) -> a -> a -> Number
+ * @sig s -> (a -> String) -> a -> a -> Number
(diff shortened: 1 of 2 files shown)


Harris-Miller:
CC: @sebbayer

kedashoe (approved):

PR #3477, opened 2024-06-23
Aline `nth` `head` `last` behavior with documentation 

ivshalin:
Align `nth`, `head` and `last` functions with docs. They should return undefined in case of empty string

### DIFF
--- a/source/internal/_nth.js
+++ b/source/internal/_nth.js
@@ -1,6 +1,4 @@
-import _isString from './_isString.js';
-
 export default function _nth(offset, list) {
   var idx = offset < 0 ? list.length + offset : offset;
-  return _isString(list) ? list.charAt(idx) : list[idx];
+  return list[idx];
 }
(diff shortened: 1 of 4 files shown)


kedashoe (approved):

RihardXXX:
`// Basic implementation
function _nth(offset, list) {
  var idx = offset < 0 ? list.length + offset : offset;
  return list[idx];
}

// and if you do it like this
function _nth(offset, list) {
  var idx = offset < 0 ? 0 : offset;
  return list[idx];
}

var list = [1, 2]

// Basic implementation result
console.log(_nth(-2, list)) // 1
console.log(_nth(-4, list)) // undefined
console.log(_nth(-7, list)) // undefined

// new 
console.log(_nth(-2, list)) // 1
console.log(_nth(-4, list)) // 1
console.log(_nth(-7, list)) // 1`

Issue #3478, opened 2024-07-09
`lensPath` doesn't accept negative indexing


AlexFrazer:
Here's what I am trying to do. I have a list of pages from `@tanstack/react-query`. When I take some action, I want to either append or update an item in _the last page_. I wanted to do something like this:

```typescript
interface PaginatedResponse<T> {
  items: T[];
  total: number;
  page: number;
  size: number;
}

const lastPageLens = lensPath(["pages", -1, "items"]);

const upsert = curry(function upsert(item: T, data: T[]) {
  const index = findIndex(propEq('id', item.id), data);
  return index < 0
    ? append(item, data)
    ? adjust<T>(index, mergeLeft(item), items)
});

function optimisticInsert<T>(item: T, data: InfiniteData<PaginatedResponse<T>>) {
  return over(lastPageLens, upsert(item));
}
```
However, when I run this, it will attempt to set the lens on `-1` in pages, i.e.
```typescript
{
  pages: {
    0: [some pages],
    -1: [the thing I wanted to upsert]
  }
}
```
I don't think this is intentional, as when I looked at `update` and `adjust`, they allowed for negative indexing

Harris-Miller:
> I don't think this is intentional, as when I looked at update and adjust, they allowed for negative indexing

You're correct. `lensPath()`'s `getter` [function](https://github.com/ramda/ramda/blob/master/source/internal/_path.js#L13) utilizes `R.nth`, which works with negative indexes, but the `setter` [function](https://github.com/ramda/ramda/blob/master/source/internal/_assoc.js#L19) is doing regular index accessing

PR #3479, opened 2024-07-10
fix(lensPath) works with negative indexes

Harris-Miller:
Fixes https://github.com/ramda/ramda/issues/3478

This MR makes `assoc`/`assocPath` set negative indexes starting from the end. This is match the behavior of `nth()`, and therefor normalizes the behavior for `view(lensPath([-1], list)` and `set(lensPath([-1], val, list)` applying to the last element in the list. Before, `set()` here would not apply to the last element, but instead immutably perform `arr[-1] = val`

`R.lensIndex`'s current behavior returns the original list when the index is out-of-range for `R.set` and `R.over`. This has remained unchanged.

Click to expand original MR details:
<details>
Putting this MR up with errors to discuss options

There is an inconsistency in behavior in `lensPath`, see it in [action](https://ramdajs.com/repl/?v=0.30.1#?const%20o%20%3D%20%7B%20foo%3A%20%5B1%2C%202%2C%203%5D%20%7D%3B%0A%0Aconst%20l%20%3D%20R.lensPath%28%5B%27foo%27%2C%20-1%5D%29%3B%0A%0Aview%28l%2C%20o%29%3B%20%2F%2F%203%0Aset%28l%2C%205%2C%20o%29%3B%20%2F%2F%20%7B%20foo%3A%20%5B1%2C%202%2C%203%2C%20%22-1%22%3A%205%5D%20%7D%20%21%21%21%21%21)
```typescript
const o = { foo: [1, 2, 3] };

const l = R.lensPath(['foo', -1]);

view(l, o); // 3
set(l, 5, o); // { foo: [1, 2, 3, "-1": 5] } !!!!!
````

Under the hood, `lensPath`'s getter function is `_path` which uses `_nth`, which handles negative indexes correctly. The setter function is `assocPath`, which uses `_assoc`, and sets the index as given. When given `-1`, it sets the "key" `-1` to the value, and you get the above. That `-1` index is non-enumerable and doesn't count towards the array's length. It's basically _hidden_.

This MR changes the `_assoc` to use `R.update()` instead, which does handle negative indexes

However...

`R.update()`, which uses `R.adjust`, which has an interesting feature... it's wraps indexes at it's length back to 0, rg

* `R.update(0, 5, [1, 2, 3]) === R.update(3, 5, [1, 2, 3]) === [5, 2, 3]`
* `R.update(1, 5, [1, 2, 3]) === R.update(4, 5, [1, 2, 3]) === [1, 5, 3]`

I traced this wrapping behavior back to this MR: https://github.com/ramda/ramda/pull/3134

This changes the behavior of `_assoc` in a fundamental way
* Before: `_assoc(3, 5, [1, 2, 3]) === [1, 2, 3, 5]`
* After: `_assoc(3, 5, [1, 2, 3]) === [5, 2, 3]`

The problem that this creates for `lensPath` is
* Before `R.set(R.lensPath(['foo', 4]), 7, { foo: [1, 2, 3] }) === [1, 2, 3, undefined, 7])`
* After `R.set(R.lensPath(['foo', 4]), 7, { foo: [1, 2, 3] }) === [1, 7, 3])`

My question is... **what should the behavior here be?**
</details>

### CODE REVIEW

**Comment by kedashoe** [comment]
huh, am I missing an issue linked to that PR? Seems out of the blue and a bit of a strange thing to do, especially because it simply returns the list if idx itself is outside the range of the list. The before behavior seems much less surprising to me.

**Comment by Harris-Miller** [comment]
https://github.com/ramda/ramda/issues/3478

**Comment by kedashoe** [comment]
> #3478

Sorry I see what we are trying to do here, I was referring to the earlier PR, I've updated my comment so it is more clear

**Comment by Harris-Miller** [comment]
> huh, am I missing an issue linked to https://github.com/ramda/ramda/pull/3134? Seems out of the blue and a bit of a strange thing to do, especially because it simply returns the list if idx itself is outside the range of the list. The before behavior seems much less surprising to me.

Yeah I don't really understand the index cycling

IMO, anything in Ramda that updates an array at an index to match retrieving from that index in how `R.nth()` works

eg
```typescript
R.update(0, 8, [1, 2]) // [8, 2]
R.update(1, 8, [1, 2]) // [1, 8]
```

Harris-Miller:
Oops, didn't mean to click "Close with comment" there

Harris-Miller:
@kedashoe I misunderstood the original MR https://github.com/ramda/ramda/pull/3134

It looked like it's cycling `R.adjust(2, R.always(8), [1, 2]) // [8, 2]` because of the line:
```typescript
var _idx = (len + idx) % len;
```
But before that, it bails out
```typescript
if (idx >= len || idx < -len) {
  return list;
}
```

The test that is [failing](https://github.com/ramda/ramda/actions/runs/9883339236/job/27297895902?pr=3479#step:5:1813) made at first glance it looked like it was adjusting based on that cycle

That "modulo to cycle through indices" is just another way to represent `var idx = offset < 0 ? list.length + offset : offset;` like it is in `_nth()`

This doesn't change the issue that much though, `R.adjust()` returns a list unchanged if the index is out of range, not setting it cyclical as I thought. The behavior still differs between `R.update()/R.adjust()` and `R.assoc()`/`R.assocPath()`.

So the question stands, **what should the behavior be?**

Harris-Miller:
Looking at how other languages handle this, I believe I found the answer in [Haskell Data.Map](https://hackage.haskell.org/package/containers-0.4.0.0/docs/Data-Map.html)

[adjust](https://hackage.haskell.org/package/containers-0.4.0.0/docs/Data-Map.html#v:adjust) :: Ord k => (a -> a) -> k -> Map k a -> Map k a
Adjust a value at a specific key. When the key is not a member of the map, the original map is returned.

[update](https://hackage.haskell.org/package/containers-0.4.0.0/docs/Data-Map.html#v:update) :: Ord k => (a -> Maybe a) -> k -> Map k a -> Map k a
The expression ([update](https://hackage.haskell.org/package/containers-0.4.0.0/docs/Data-Map.html#v:update) f k map) updates the value x at k (if it is in the map).

Both express the behavior "When the key is not a member of the map, the original map is returned." This is virtually the same behavior of `R.adjust` and `R.update` for arrays with out-of-range indexes, returning the original array unchanged. The docs should be updated to make this behavior clear, which I'll do in this MR

`R.assoc()` and `R.assocPath()`, which are both already implemented to handle both objects and arrays, and their documentation should be updated to reflect that as well. The difference is that out-of-ranges will not be ignored, and instead behavior as I showed above:
```typescript
R.assoc(0, 8, [1, 2]) // [8, 2]
R.assoc(1, 8, [1, 2]) // [1, 8]
R.assoc(2, 8, [1, 2]) // [1, 2, 8]
R.assoc(3, 8, [1, 2]) // [1, 2, undefined, 8]

R.assoc(-1, 8, [1, 2]) // [1, 8]
R.assoc(-2, 8, [1, 2]) // [8, 2]
R.assoc(-3, 8, [1, 2]) // RangeError
```
`R.assocPath`, working the same but deeply. This matches the already defined behavior of "setting or overriding the nodes required to create the given path". New behavior being that a `RangeError` is thrown when passed an index that will result in `< 0` as shown above, instead of the current behavior: `R.assoc(-3, 8, [1, 2]) === [1, 2, "-1": 8]` behavior.

Note: While `RangeError` is my personal preference, I would fight against the "garbage in, garbage out" the current behavior does.

I'm going to work on this and update it later tonight or tomorrow

Harris-Miller:
Changed my mind on `RangeError`
```typescript
var arr = [1, 2, 3];
arr[-1] = 42;
console.log(arr); // [1, 2, 3, "-1": 42]
```
So keep it consistent to that for
```typescript
R.assoc(-4, 42, [1, 2, 3]); // [1, 2, 3, "-1": 42]
```

Harris-Miller:
@kedashoe ok I think this one is ready

source/adjust.js
```diff
@@ -7,6 +7,8 @@ import _curry3 from './internal/_curry3.js';
  * new copy of the array with the element at the given index replaced with the
  * result of the function application.
  *
+ * When `idx < 0 || idx >= list.length`, the original list is returned.
```
kedashoe, line 10: This isn't quite right since we handle negative indexes.

"When `idx < -list.length || idx >= list.length`, the original list is returned." or something like that?

source/assocPath.js
```diff
@@ -27,6 +27,8 @@ import isNil from './isNil.js';
  *
  *      // Any missing or non-object keys in path will be overridden
  *      R.assocPath(['a', 'b', 'c'], 42, {a: 5}); //=> {a: {b: {c: 42}}}
+ *      R.assocPath(['a', 1, 0], 42, {a: []}); // => {a: [undefined, [42]]}
+ *      R.assocPath(['a', -1], 42, {a: [1, 2]}); // => {a: [undefined, [1, 42]]}
```
kedashoe, line 31: No `undefined` in this one:

 `{a: [undefined, [1, 42]]}` -> `{ a: [ 1, 42 ] }`

source/lensIndex.js
```diff
@@ -7,6 +7,8 @@ import update from './update.js';
 /**
  * Returns a lens whose focus is the specified index.
  *
+ * When `idx < 0 || idx >= list.length`, `R.set` or `R.over`, the original list is returned.
```
kedashoe, line 10: Same comment about `idx` as above. Also `R.over` may throw depending on what the function is

source/lensIndex.js
```diff
@@ -23,6 +25,10 @@ import update from './update.js';
  *      R.view(headLens, ['a', 'b', 'c']);            //=> 'a'
  *      R.set(headLens, 'x', ['a', 'b', 'c']);        //=> ['x', 'b', 'c']
  *      R.over(headLens, R.toUpper, ['a', 'b', 'c']); //=> ['A', 'b', 'c']
+ *
+ *      // out-of-range returns original list
+ *      R.set(R.lensIndex(3), 'x', ['a', 'b', 'c']);         //=> ['a', 'b', 'c']
+ *      R.over(R.lensIndex(-1), R.toUpper, ['a', 'b', 'c']); //=> ['a', 'b', 'c']
```
kedashoe, line 31: Did you want to show something out of bounds with `over`? -1 is in bounds and this returns `['a', 'b', 'C']` (upper case C). Note this one will throw if out of bounds

dist/ramda.js
kedashoe, line 1: Lets remove these `dist` files from the PR.

I'll put up a PR to remove generating those from `npm run build` and just do it when publishing

kedashoe (commented):
:+1: this change. Couple minor comments

kedashoe:
Last thing not technically necessary but might be nice to add a test for negative index to `lensPath` to tie a bow on the original issue

source/lensIndex.js
Harris-Miller: This was supposed to be `-4`

dist/ramda.js
Harris-Miller: Oof. I always forget that. Needed to build locally to run tests against. Will revert

Harris-Miller:
@kedashoe 
> Last thing not technically necessary but might be nice to add a test for negative index to lensPath to tie a bow on the original issue
Super glad you asked for this, found an additional bug. It wasn't just `_assoc` that wasn't respecting the negative index wrap functionality, but `assocPath` as well. Had to abstract `prop` to `_prop` so it would work as expected

source/assocPath.js
Harris-Miller: Made a better example

source/update.js
```diff
@@ -7,6 +7,8 @@ import always from './always.js';
  * Returns a new copy of the array with the element at the provided index
  * replaced with the given value.
  *
+ * When `idx < 0 || idx >= list.length`, the original list is returned.
```
kedashoe, line 10: idx < -list.length here as well I think

kedashoe (approved):
Looks great, one last minor comment

kedashoe (approved):

PR #3481, opened 2024-08-04
R.empty() support for Map and Set

Harris-Miller:
Stemming from https://github.com/ramda/ramda/pull/3475#issuecomment-2175437701

> `isEmpty` returns false for empty Map and Set

To Fix this, `empty()` should return `new Map()`/`new Set()` respectively. `isEmpty()` then naturally works due to it's implementation

```typescript
var s1 = new Set([1, 2, 3]);
var s2 = s1.union(new Set());
R.equals(s1, s2); // => true
```

My implementation works when passed either an instance or the constructor

```typescript
R.empty(Set) // => empty Set
R.empty(new Set([1, 2, 3])) empty Set

R.empty(Map) // => empty Map
R.empty(new Map([1, 2, 3])) // => empty Map
```

### DIFF
--- a/source/empty.js
+++ b/source/empty.js
@@ -41,17 +41,21 @@ var empty = _curry1(function empty(x) {
           ? x.empty()
           : (x != null && x.constructor != null && typeof x.constructor.empty === 'function')
             ? x.constructor.empty()
-            : _isArray(x)
-              ? []
-              : _isString(x)
-                ? ''
-                : _isObject(x)
-                  ? {}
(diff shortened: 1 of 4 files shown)


source/empty.js
kedashoe, line 1: We can update the sentence in the comments lines 10-12 about which types ramda supports.

kedashoe:
:+1:

kedashoe (approved):

PR #3484, opened 2024-08-07
preserve the context of function calls to R.partialObject

midnight-wonderer:
I made a pull request because of [the warning from Rollup](https://github.com/ramda/ramda/discussions/3483).

But a few years have passed since the function showed up, and nobody has complained so far. Me included; I am just here because of the warning.

I wonder why we don't just embrace the thisless of the function, like this:
~~~javascript
var partialObject = _curry2((f, o) => (props) => f(mergeDeepRight(o, props)));
~~~

I started the PR this way, as @kedashoe suggested in the discussion linked above.


source/partialObject.js
```diff
@@ -29,5 +29,7 @@ import _curry2 from './internal/_curry2.js';
  *      sayHelloToMs({ firstName: 'Jane', lastName: 'Jones' }); //=> 'Hello, Ms. Jane Jones!'
  * @symb R.partialObject(f, { a, b })({ c, d }) = f({ a, b, c, d })
  */
-var partialObject = _curry2((f, o) => (props) => f.call(this, mergeDeepRight(o, props)));
+var partialObject = _curry2(function partialObject(f, o) {
+  return (props) => f.call(this, mergeDeepRight(o, props));
```
kedashoe, line 33: Let's convert this arrow function and return `function _partialObject(props) { ... }` here as well. For now ramda is not supposed to be using any arrow functions in `source`.

kedashoe (changes requested):

kedashoe:
> I wonder why we don't just embrace the thisless of the function, like this

We will most likely go that route in the future, but for now we're not using arrow functions (unless they make it in by accident as they did here).

midnight-wonderer:
Applied the change you suggested but also removed the `.call`.

I've seen many places where similar function calls are done.
For instance, in `applyTo`:
https://github.com/ramda/ramda/blob/6e8b68ff35ad995c9576f88fc574d8715b21590a/source/applyTo.js#L22

Since I am only aware that we should not use arrow functions, this should be okay.
However, if there are issues associated with the call, please let me know.

Ping: @hejhi for head-up.

kedashoe:
> also removed the .call

:+1:

kedashoe (approved):

PR #3486, opened 2024-09-06
feat: add upperFirst string

nhannt201:
Add upperFirst to capitalize the first letter of the string


CrossEye:
I think this comment on another recent PR is relevant:

https://github.com/ramda/ramda/pull/3488#issuecomment-2335287734

PR #3487, opened 2024-09-06
feat: add lowerFirst string

nhannt201:
Converts the first character of a string to lowercase.


yurkimus:
IMHO: It looks too specific to add this to the library.

CrossEye:
> IMHO: It looks too specific to add this to the library.

Agreed.  I think this comment on another recent PR is relevant:

https://github.com/ramda/ramda/pull/3488#issuecomment-2335287734

PR #3488, opened 2024-09-06
feat: add upper and lower case string

nhannt201:
Converts string, as space separated words, to upper case or lower case.


CrossEye:
I'd like to hear arguments for why this should be in Ramda.

It seems to me that besides your `UPPER CASE` and `lower case` we would also want to consider `Title Case`, `kebab-case`, `snake-case`, `camelCase` and `UpperCamelCase`.  Do you think Ramda should be in the business of converting between any of these formats.  Do we need `kebabToCamel` and `upperCamelToTitle`, etc.?

My first thought is that these belong in user-land, or one of the adjunct libraries.

nhannt201:
> I'd like to hear arguments for why this should be in Ramda.
> 
> It seems to me that besides your `UPPER CASE` and `lower case` we would also want to consider `Title Case`, `kebab-case`, `snake-case`, `camelCase` and `UpperCamelCase`. Do you think Ramda should be in the business of converting between any of these formats. Do we need `kebabToCamel` and `upperCamelToTitle`, etc.?
> 
> My first thought is that these belong in user-land, or one of the adjunct libraries.

I believe these features can be implemented in other libraries. However, since Ramda already includes `toUpper` and `toLower`, I think it would be beneficial to add variations for different types of uppercase and lowercase transformations.

kedashoe:
> My first thought is that these belong in user-land, or one of the adjunct libraries.

Agreed, ty for the PRs @nhannt201 however these functions are imho too specific for ramda and there are already other libraries that do this

PR #3493, opened 2024-11-17
Add rebuild, mapKeys, ,and renameKeys

### LINKED ISSUE #1361: map over object keys
I'm looking for  way to map over object keys. Something like:

```javascript
let obj = {a: 1, b: 2} 
let result = mapKeys(k => `${k}z`, obj)  // {az: 1, bz: 2}
```

didn't find anything like that by browsing thru the docs

### LINKED ISSUE #3491: R.renameKeys
Hi, I think it would be interesting to add a function (which could be called `R.renameKeys`) which would rename properties of an object according to a mapping object (replacing object keys by corresponding properties values in mapping object, keeping properties if their key does not exist in mapping object and keeping properties initial order).

```js
const mapping = { name: 'firstName', address: 'street' };
const obj = { name: 'John', city: 'Paris' };

R.renameProps(mapping, obj);
// { firstName: 'John', city: 'Paris' }
```

Edit: `R.renameProps` → `R.renameKeys`

CrossEye:
Fixes #3491 
Fixes #1361 

This adds three new functions:

  * **`rebuild`** allows you to reconstruct objects -- thought of as collections of key-value pairs -- into new objects by mapping each key-value into an array of key-value pairs.  It unpacks the object, calls the function supplied on each one, and takes the resulting values and folds them into an object.

  * **`mapKeys`** is built atop `rebuild`, and allows you to transform an object using a function that maps each key to a new value.

  * **`renameKeys`** is built atop `mapKeys` and uses a mapping object to form the basis for its key transformation.  Any key of the object to be transformed that is found in the mapping object, will take the corresponding value.  Others will be left as is.

In an [earlier discussion][ed] on this idea, I also floated the function `remap`, which is similar to `rebuild`, but is simpler, and would be used when our transformation is a one-to-one mapping between key-value pairs.  I wasn't sure whether to include it here or not.  What do people think?

Although it was the driving function of the issue that inspired this change, I'm still not really convinced that `renameKeys` is appropriate for Ramda.  Do  you use this functionality often?

----------

On a side note, to keep with our model, this involved moving `fromPairs` and `toPairs` to `_internal` function, and to make the new `rebuild` and `mapKeys` `_internal` functions, all with public glosses that add currying.

I'm thinking that this could be refactored.  I haven't tried yet, but could we remove the `_internal` functions altogether and add the currying inside `sources/index.js` when we're exporting?  It seems to me that it would be a significant cleanup in all sorts of ways.  I haven't looked at all the builds, but it does feel as though this could be quite useful.  What do you think?

  [ed]: https://github.com/ramda/ramda/issues/1361#issuecomment-780716313

### DIFF
--- a/source/internal/_rebuild.js
+++ b/source/internal/_rebuild.js
@@ -1,0 +1,11 @@
+import _fromPairs from './_fromPairs.js';
+import _toPairs from './_toPairs.js';
+import _chain from './_chain.js';
+
+var rebuild = function(convert, obj) {
+  return _fromPairs(_chain(
+    function(pair) {return convert(pair[0], pair[1]);},
+    _toPairs(obj)
+  ));
+};
+export default rebuild;
--- a/source/mapKeys.js
+++ b/source/mapKeys.js
@@ -13,7 +13,7 @@ import _mapKeys from './internal/_mapKeys.js';
  * @param {Function} fn
  * @param {Object} obj
  * @return {Object}
- * @see R.map
+ * @see R.map, R.rebuild, R.renameKeys
  * @example
  *
  *      R.mapKeys(toUpper, {foo: 1, bar: 2, baz: 3}) //=> {FOO: 1, BAR: 2, BAZ: 3}
--- a/source/mapKeys.js
+++ b/source/mapKeys.js
@@ -1,5 +1,5 @@
 import _curry2 from './internal/_curry2.js';
-import _rebuild from './internal/_rebuild.js';
+import _mapKeys from './internal/_mapKeys.js';
 
 /**
  * Transforms an object by converting the keys to new values.
@@ -8,7 +8,6 @@ import _rebuild from './internal/_rebuild.js';
  *
  * @func
  * @memberOf R
- * @since v0.9.0
  * @category Object
  * @sig (String -> String) -> Object -> Object
  * @param {Function} fn
@@ -19,8 +18,6 @@ import _rebuild from './internal/_rebuild.js';
  *
  *      R.mapKeys(toUpper, {foo: 1, bar: 2, baz: 3}) //=> {FOO: 1, BAR: 2, BAZ: 3}
  */
-var mapKeys = _curry2(function mapKeys(fn, obj) {
-  return _rebuild(function(k, v) {return [[fn(k), v]];}, obj);
-});
+var mapKeys = _curry2(_mapKeys);
 
 export default mapKeys;
--- a/source/mapKeys.js
+++ b/source/mapKeys.js
@@ -1,0 +1,26 @@
+import _curry2 from './internal/_curry2.js';
+import _rebuild from './internal/_rebuild.js';
+
+/**
+ * Transforms an object by converting the keys to new values.
+ *
+ * **Note** that if multiple keys map to the same new key, the last one processed will dominate.
+ *
+ * @func
+ * @memberOf R
+ * @since v0.9.0
+ * @category Object
+ * @sig (String -> String) -> Object -> Object
+ * @param {Function} fn
+ * @param {Object} obj
+ * @return {Object}
+ * @see R.map
+ * @example
+ *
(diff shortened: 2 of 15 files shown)


source/rebuild.js
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Transforms an object into a new one, applying to every key-value pair a
+ * function creating zero, one, or many new key-value pairs, and combining
+ * the resulst into a single object.
+ *
+ * @func
+ * @memberOf R
+ * @category List
+ * @sig ([String, a] -> [[String, b]]) -> Object<a> -> Object<b>
```
kedashoe, line 12: We don't use this `<>` in `Object<a>` syntax anywhere else. Maybe `{k: a}` and `{k: b}` or just `Object`?

source/renameKeys.js
```diff
@@ -0,0 +1,33 @@
...
+ * object to their corresponding values from that object.
+ *
+ * @func
+ * @memberOf R
+ * @category Object
+ * @sig Object -> Object -> Object
+ * @param {Function} mapping An object pairing existing keys with new ones
+ * @param {Object} obj A target object to convert
+ * @return {Object} The result of replacing existing keys with their mapping counterparts when such exist
+ * @see R.map
```
kedashoe, line 17: see rebuild and mapKeys?

source/internal/_rebuild.js
```diff
@@ -0,0 +1,11 @@
+import _fromPairs from './_fromPairs.js';
+import _toPairs from './_toPairs.js';
+import _chain from './_chain.js';
+
+var rebuild = function(convert, obj) {
+  return _fromPairs(_chain(
```
kedashoe, line 6: I feel like we should go with just `map` to start with here rather than forcing users to wrap all their return values in an array for the most common use case.

source/mapKeys.js
```diff
@@ -0,0 +1,23 @@
...
+ * **Note** that if multiple keys map to the same new key, the last one processed will dominate.
+ *
+ * @func
+ * @memberOf R
+ * @category Object
+ * @sig (String -> String) -> Object -> Object
+ * @param {Function} fn
+ * @param {Object} obj
+ * @return {Object}
+ * @see R.map
```
kedashoe, line 16: Can `@see` the other new functions here as well

source/rebuild.js
CrossEye: Right.  The stuff I meant to go back and fix before I submitted... ah well, it's been a while.

source/mapKeys.js
CrossEye: yes, just forgot.  Cut and pasted the JSDoc stuff from elsewhere.  Will try to fix tomorrow.

source/internal/_rebuild.js
CrossEye: Well, that's the one I didn't include from earlier discussions.  I certainly don't mind including it.  In my earlier discussion, I called it `remap`, which still sounds all right to me.  But I would like to also have the flexibility that `rebuild` offers.  Whether they nest like I've done (with `remap` between `rebuild` and `mapKeys`) is an interesting question.

source/renameKeys.js
CrossEye: yes

source/internal/_rebuild.js
kedashoe: ok let's just keep what you have here then, no need for both right now

Harris-Miller:
@kedashoe @CrossEye After Thanksgiving I'll work on getting these into https://github.com/ramda/types so they're ready to merge for a `0.31.0` release

Harris-Miller:
@CrossEye I believe we're missing `@since v0.31.0` in the JSDocs of the 3 new funcs

kedashoe:
> @CrossEye I believe we're missing `@since v0.31.0` in the JSDocs of the 3 new funcs

since annotations are added [here](https://github.com/ramda/ramda/blob/master/scripts/prepublish#L7-L12)

Issue #2137, opened 2017-03-21
assocPath and Symbols


char0n:
Hi,

Maybe related to #1555 but, when assocPathing objects with symbols, these symbols are lost.

POC: 

```javascript
R.assocPath(['test'], 5, {test: 3, [Symbol.for('symbol')]: 4}); // => {test: 5}
```

char0n:
The problem seems to be in `assoc` function which assocPath uses. `assoc` uses `for .. in` loop to enumerate the object props. Symbols are not enumerable, hence the problem..

char0n:
Maybe something like this could work. I can issue a pull request if you @all find this acceptable.

```javascript
module.exports = _curry3(function assoc(prop, val, obj) {
  var result = {};
  for (var p in obj) {
    result[p] = obj[p];
  }
  // Symbols are not enumerable in `for...in` loop. 
  if (Object.getOwnPropertySymbols) {
    for (var symbol in Object.getOwnPropertySymbols(obj)) {
      result[symbol] = obj[symbol];
    }
  }
  result[prop] = val;
  return result;
});
```

davidchambers:
My view: Symbols are a bad part of the language which we should do nothing specific to support.

char0n:
I can relate to that. I also think ES2015 Symbols are unfortunate, but nevertheless they are here and people are already using them. Decision not to support Symbols at all can disqualify ramda from codebases that use them or can lead to unexpected bugs.

Workaround for now is to use lodash.set in this circumstances.

dylan-chong:
there is one use of symbols in nodejs - overriding the inspect behaviour on the object with
```
{
    [require('util').inspect.custom]: inspectContext
}
```

`R.assocPath` ends up removing this function which makes a very hard-to-debug issue of the inspect behaviour not working

i dont know why symbols are a thing at all, but here, i can't avoid them

kedashoe:
Ya symbols do seem quite awkward. @char0n 's solution looks fine. Maybe we should put up a PR and see if we can generate any more discussion about whether to include them or not? @dylan-chong would you like to submit a PR?

char0n:
Hi, @kedashoe 

PR issued as https://github.com/ramda/ramda/pull/3506

PR #3506, opened 2025-01-20
feat(assocPath): add support for Symbols

char0n:
Refs #2137

### DIFF
--- a/source/internal/_assoc.js
+++ b/source/internal/_assoc.js
@@ -27,6 +27,12 @@ export default function _assoc(prop, val, obj) {
   for (var p in obj) {
     result[p] = obj[p];
   }
+  if (Object.getOwnPropertySymbols) {
+    var symbols = Object.getOwnPropertySymbols(obj);
+    for (var i = 0; i < symbols.length; i += 1) {
+      result[symbols[i]] = obj[symbols[i]];
+    }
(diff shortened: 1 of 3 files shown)


dylan-chong (approved):
this is not fully adding support for symbols ( i suppose to be thorough we'd need to add tests for everything else - prop, lens, evolve, etc) but i'm happy with this as it just unbreaks the functionality for now

I'm happy with this being a silent, not officially support symbols, but fix the bug anyway PR.

thanks so much for submitting the PR so quickly @char0n 🙇

char0n:
> this is not fully adding support for symbols ( i suppose to be thorough we'd need to add tests for everything else to be consistent - prop, lens, evolve, etc) but i'm happy with this as it just unbreaks the functionality for now
> 
> I'm happy with this being a silent, not officially support symbols, but fix the bug anyway PR.
> 
> thanks so much for submitting the PR so quickly @char0n 🙇

It's just a start and a precedence ;]
I was waiting 8 years to issue this PR ;]

Harris-Miller:
The shallow-clone behavior of `assoc` copies only enumerable properties. Symbols are inherently non-enumerable. Personally not a fan of adding Symbol support here. It breaks existing expectations and conventions

Issue #3497, opened 2024-11-20
Should Map be Filterable?


davetapley:
The `Filterable` used in ⬇️ doesn't seem to support [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)?

https://github.com/ramda/ramda/blob/980b3bf7712adb8ad8b5d9b2c6aca525cb3cf143/source/filter.js#L11-L16

As best I can tell it fails the `_isobject` check:
https://github.com/ramda/ramda/blob/980b3bf7712adb8ad8b5d9b2c6aca525cb3cf143/source/filter.js#L38-L40

Because: `Object.prototype.toString.call(new Map()) === '[object Map]' !== [object Object]`

Then fall through to this which fails because `Map` doesn't have `.length`:
https://github.com/ramda/ramda/blob/852e252041fbbd8e341d0d3dc7749c2b5bd692a7/source/internal/_filter.js#L1-L6

---

I'm not too sure of ramda internals, but could it use the [iterable protocol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterable_protocol) instead of assuming `list`with `.length`?

kedashoe:
ty for the issue @davetapley , agreed it would be nice to support for `Map`. Maybe it is not quite so simple as checking if the given object supports iterable, as we would need a generic way to build whatever we have as well. But we can add support for `Map`, would you like to submit a PR?

PR #3507, opened 2025-01-24
Add filter for Maps (#3497)

davetapley:
#3497

### DIFF
--- a/source/filter.js
+++ b/source/filter.js
@@ -2,15 +2,16 @@ import _arrayReduce from './internal/_arrayReduce.js';
 import _curry2 from './internal/_curry2.js';
 import _dispatchable from './internal/_dispatchable.js';
 import _filter from './internal/_filter.js';
+import _filterMap from './internal/_filterMap.js';
+import _isMap from './internal/_isMap.js';
 import _isObject from './internal/_isObject.js';
 import _xfilter from './internal/_xfilter.js';
 import keys from './keys.js';
 
-
 /**
  * Takes a predicate and a `Filterable`, and returns a new filterable of the
  * same type containing the members of the given filterable which satisfy the
- * given predicate. Filterable objects include plain objects or any object
+ * given predicate. Filterable objects include plain objects, Maps, or any object
  * that has a filter method such as `Array`.
  *
  * Dispatches to the `filter` method of the second argument, if present.
@@ -44,8 +45,10 @@ var filter = _curry2(_dispatchable(['fantasy-land/filter', 'filter'], _xfilter, 
         }
         return acc;
       }, {}, keys(filterable)) :
-    // else
-      _filter(pred, filterable)
+      _isMap(filterable) ?
+        _filterMap(pred, filterable) :
+      // else
+        _filter(pred, filterable)
   );
 }));
 export default filter;
--- a/source/internal/_filterMap.js
+++ b/source/internal/_filterMap.js
@@ -1,0 +1,10 @@
+export default function _filterMap(fn, map) {
+  var result = new Map();
+
+  map.forEach((value, key) => {
+    if (fn(value)) {
+      result.set(key, value);
+    }
+  });
+  return result;
+}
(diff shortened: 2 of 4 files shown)


source/internal/_filterMap.js
```diff
@@ -0,0 +1,10 @@
+export default function _filterMap(fn, map) {
+  var result = new Map();
+
+  for (var [key, value] of map) {
```
kedashoe, line 4: We're not using `of` yet, could we re-write this to use `in` or some other equivalent?

source/filter.js
```diff
@@ -2,11 +2,12 @@ import _arrayReduce from './internal/_arrayReduce.js';
 import _curry2 from './internal/_curry2.js';
 import _dispatchable from './internal/_dispatchable.js';
 import _filter from './internal/_filter.js';
+import _filterMap from './internal/_filterMap.js';
+import _isMap from './internal/_isMap.js';
 import _isObject from './internal/_isObject.js';
 import _xfilter from './internal/_xfilter.js';
 import keys from './keys.js';
 
-
```
kedashoe, line 9: Add a note in here that we filter `Map`s

kedashoe (changes requested):
ty for the pr @davetapley , just a couple minor requests

source/filter.js
```diff
@@ -44,8 +45,10 @@ var filter = _curry2(_dispatchable(['fantasy-land/filter', 'filter'], _xfilter,
         }
         return acc;
       }, {}, keys(filterable)) :
-    // else
-      _filter(pred, filterable)
+      _isMap(filterable) ?
```
Harris-Miller, line 48: Do we need a new internal function for this?

```suggestion
      is(Map, filterable) ?
```

source/internal/_filterMap.js
kedashoe: > `for...of` is the equivalent for `in`, and the best option. It's been widely available since 2016 and predates the 2 options of turning the iterable into an array with either `Array.from(map)` or `[...map].

I'm not arguing against `of` in general, but I don't think we should introduce it here.

source/filter.js
davetapley: @Harris-Miller your choice 😁 , I was just replicating
https://github.com/ramda/ramda/blob/fdf8722d6cde38c4f8f85cc557100f83f431fd81/source/internal/_isObject.js#L1

davetapley:
@kedashoe @Harris-Miller fixed all but possible remove of `_isMap` (awaiting answer))

kedashoe:
> @kedashoe @Harris-Miller fixed all but possible remove of `_isMap` (awaiting answer))

Works for me, we do have a lot of the `_isXYZ` functions. :rocket: @Harris-Miller ?

kedashoe (approved):

kedashoe:
gonna go ahead and merge this one, we can discuss/refactor those `_is` functions separately

source/internal/_filterMap.js
```diff
@@ -0,0 +1,10 @@
+export default function _filterMap(fn, map) {
+  var result = new Map();
+
+  map.forEach((value, key) => {
+    if (fn(value)) {
+      result.set(key, value);
+    }
+  });
+  return result;
+}
```
Harris-Miller, line 10: Sorry noticed this after the merge. For performance reasons, we shouldn't use `.forEach()`. See: https://github.com/ramda/ramda/wiki/Ramda-Conventions#use-imperative-constructs-to-improve-performance

CC: @kedashoe

Harris-Miller: @kedashoe @davetapley ARG! I did the thing where you make comments but forget to submit them

My suggest above this one solves this one. I'll through it up in a separate MR

source/filter.js
Harris-Miller: Looking at this move, yes let's keep `_isMap` as you have it

PR #3515, opened 2025-07-02
Update last.js example

GeoDaoyu:
It's easier to see the difference between `last` and `tail`.


GeoDaoyu:
> Fair enough. `head` is the odd one out though now, would you mind updating that as well?

I updated head.js example.

PR #3516, opened 2025-07-09
fix: Both arguments to range must be integer

GeoDaoyu:
Both arguments to range must be integer.

If not, it will throw an `Uncaught RangeError: Invalid array length` 
on the Line 25: `var result = Array(from < to ? to - from : 0);`

### DIFF
--- a/source/range.js
+++ b/source/range.js
@@ -23,10 +23,10 @@ var range = _curry2(function range(from, to) {
   if (!(_isNumber(from) && _isNumber(to))) {
     throw new TypeError('Both arguments to range must be numbers');
   }
-  var result = Array(from < to ? Math.ceil(to - from) : 0);
-  var finish = from < 0 ? to + Math.abs(from) : to - from;
+  var length = from < to ? Math.ceil(to - from) : 0;
+  var result = Array(length);
   var idx = 0;
-  while (idx < finish) {
+  while (idx < length) {
     result[idx] = idx + from;
     idx += 1;
   }


kedashoe:
ty for the PR. Looks like we have a failing test, could you update?

GeoDaoyu:
I submitted this pr, because when I was executing `range(1, 5.5)`, ramda threw an uncaught exception.
I read the api documentation of ramda and thought that ramda only accepts integers, so I submitted the code.
Now it seems that ramda can accept decimals, but the difference must be an integer because Line 25.


Now I have two approaches to fix it:
1. Throw another exception
``` js
var range = _curry2(function range(from, to) {
  if (!(_isNumber(from) && _isNumber(to))) {
    throw new TypeError('Both arguments to range must be numbers');
  }
  var length = from < to ? to - from : 0;
  if (! _isInteger(length)) {
    throw new TypeError('The difference to range must be integers');
  }
  var result = Array(length);
  // ... rest 
});
```

2. Supports difference that are not integers, like lodash
``` js
import _ from 'lodash'

_.range(1, 5.5)
// [1, 2, 3, 4, 5]
```
Maybe `Math.floor` or `Math.ceil` the difference.

GeoDaoyu:
@kedashoe @Harris-Miller I updated the code by the second solution.
Please take a look.

source/range.js
```diff
@@ -16,13 +16,14 @@ import _isNumber from './internal/_isNumber.js';
...
+ *      R.range(1, 5.5);  //=> [1, 2, 3, 4, 5]
+ *      R.range(1.5, 5.5);  //=> [1.5, 2.5, 3.5, 4.5]
  */
 var range = _curry2(function range(from, to) {
   if (!(_isNumber(from) && _isNumber(to))) {
     throw new TypeError('Both arguments to range must be numbers');
   }
-  var result = Array(from < to ? to - from : 0);
+  var result = Array(from < to ? Math.ceil(to - from) : 0);
   var finish = from < 0 ? to + Math.abs(from) : to - from;
```
kedashoe, line 27: Semi related, isn't this calculation of `finish` unnecessary? Can we get rid of it and just do `while (idx < result.length)` for the loop below?

Also, could we add a test that starts negative? Say -3 to 2?

GeoDaoyu: > Semi related, isn't this calculation of `finish` unnecessary? Can we get rid of it and just do `while (idx < result.length)` for the loop below?
> 
> Also, could we add a test that starts negative? Say -3 to 2?

You are right, `finish` is unnecessary.

kedashoe (approved):
ty, looks good to me, will leave open for a bit to see if anyone else has further comments

PR #3522, opened 2025-09-18
fix: package.json exports fields

Harris-Miller:
Figured this out from this PR: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/73646

What I updated was invalid. Left side can't have trailing slashes, right side needs to resolve to a file, not a folder


Harris-Miller:
@kedashoe for this one, yes. Particularly, how this fix is going to futureproof for typescript6.0

PR #3530, opened 2026-06-04
Fix toString cycle detection equality dispatch

### LINKED ISSUE #3150: toString function is executing function expression assign as a value to an "equals" key of given object.
The **Bug** is reproduced [here](https://ramdajs.com/repl/?v=0.27.0#?const%20complexObject%20%3D%20%7B%20%0A%20%20foo%3A%20%7B%20%0A%20%20%20%20foo%3A%20%7B%0A%20%20%20%20%20%20equals%3A%20%28%29%20%3D%3E%20console.log%28%27This%20should%20NOT%20be%20called%2C%20but%20it%20is%21%20D%3A%27%29%20%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D%0A%0Aconst%20simplerObject%20%3D%20%7B%20%0A%20%20foo%3A%20%7B%0A%20%20%20%20equals%3A%20%28%29%20%3D%3E%20console.log%28%27This%20will%20not%20be%20called%27%29%20%0A%20%20%7D%0A%7D%0A%0AR.toString%28complexObject%29%0A)!

image

You can see that the **toString** function will execute the function, which is provided on the **equals** property(**please** see *complexObject*).
Also **Note** how I have two nested identical keys(in this case -> **foo**).

If I change the **equals** property to something different then there is **no problem**. If I provide **simplerObject** to the **toString** function, then equals is **NOT** being executed as well.

I would expect the **toString** function to **never** execute functions the way it is shown on the above example, which is showing the **Bug**. 

What is the source of this **Bug and can it be fixed?

puneetdixit200:
## Summary
- use strict reference checks for `toString` circular-reference detection
- avoid dispatching to user-defined `equals` methods while checking seen values
- add a regression test for issue #3150

Fixes #3150.

## Tests
- `BABEL_ENV=cjs npm exec -- mocha test/toString.js --require @babel/register --reporter spec --grep "does not dispatch to equals"`
- `BABEL_ENV=cjs npm exec -- mocha test/toString.js --require @babel/register --reporter min`
- `npm run spec`
- `npm run lint`
- `git diff --check`


kedashoe (approved):
ty for the PR, looks good to me

PR #3531, opened 2026-06-17
fix: equals incorrectly treats ArrayBuffers of different sizes or content as equal

JSap0914:
## Bug

`R.equals` compares `ArrayBuffer` objects by their enumerable own properties only. Plain `ArrayBuffer` instances have no own enumerable properties, so two buffers with **different `byteLength`** (or different byte content) are incorrectly reported as equal:

```js
R.equals(new ArrayBuffer(8), new ArrayBuffer(16)) // => true  (should be false)

const a = new ArrayBuffer(4);
new Uint8Array(a)[0] = 255;
R.equals(a, new ArrayBuffer(4))                  // => true  (should be false)
```

The existing typed-array test works around this by assigning JS own-properties directly onto the `ArrayBuffer` object (`typArr1[0] = 1`), which is not how `ArrayBuffer` is normally used.

## Fix

In `source/internal/_equals.js`, split the `case 'ArrayBuffer':` out of the typed-array fall-through chain so it gets its own handling:

1. Return `false` immediately if `a.byteLength !== b.byteLength`.
2. Compare raw byte content by recursing on `Uint8Array` views of the two buffers.
3. `break` to fall through to the existing own-property key comparison (preserving the existing test behaviour for manually added properties).

The typed-array cases (`Int8Array` … `Float64Array`) are given an explicit `break` so they no longer fall into the new `ArrayBuffer` branch.

## Verification

```
npm test
1211 passing (1s)
```

All 1211 tests pass (one new test added in `test/equals.js`).

> This fix was developed with AI assistance.


source/internal/_equals.js
```diff
@@ -126,7 +126,14 @@ export default function _equals(a, b, stackA, stackB) {
     case 'Uint32Array':
     case 'Float32Array':
     case 'Float64Array':
+      break;
     case 'ArrayBuffer':
+      if (a.byteLength !== b.byteLength) {
```
kedashoe, line 131: Now that the ArrayBuffer case is doing more, let's move it up between the Set and Arguments blocks

source/internal/_equals.js
```diff
@@ -126,7 +126,14 @@ export default function _equals(a, b, stackA, stackB) {
...
+        return false;
+      }
+      if (!_equals(new Uint8Array(a), new Uint8Array(b), stackA.concat([a]), stackB.concat([b]))) {
```
kedashoe, line 134: We should be able to return here, ya?
```
return _equals(new Uint8Array(a), new Uint8Array(b), stackA.concat([a]), stackB.concat([b]))
```
If it is not !_equals, then it is true, don't need to fall through to the break.

test/equals.js
```diff
@@ -204,6 +204,13 @@ describe('equals', function() {
       eq(R.equals(typArr1, typArr3), false);
       eq(R.equals(typArr1, intTypArr), false);
     });
+    it('compares ArrayBuffers by byte content', function() {
+      eq(R.equals(new ArrayBuffer(8), new ArrayBuffer(8)), true);
+      eq(R.equals(new ArrayBuffer(8), new ArrayBuffer(16)), false);
+      var filled = new ArrayBuffer(4);
+      new Uint8Array(filled)[0] = 255;
+      eq(R.equals(filled, new ArrayBuffer(4)), false);
```
kedashoe, line 212: Can we add a non-trivial test case where we expect them to be equal?

kedashoe (changes requested):
Hi @JSap0914 , ty for the PR! I've left a couple comments

PR #3534, opened 2026-07-01
fix: compare Set and Map members as a multiset in equals

spokodev:
`R.equals` returns the wrong result, and returns a different result depending on argument order, for `Set` and `Map` values whose members are value-equal but reference-distinct.

## Repro

```js
const R = require('ramda');
const { isDeepStrictEqual } = require('node:util');

const a = new Set([{v: 1}, {v: 1}]);
const b = new Set([{v: 1}, {v: 2}]);

R.equals(a, b);              // => true   (wrong)
R.equals(b, a);              // => false  (asymmetric)
isDeepStrictEqual(a, b);     // => false  (oracle, both orders)
isDeepStrictEqual(b, a);     // => false
```

The same holds for `Map` keyed on value-equal, reference-distinct objects.

## Root cause

For two equal-size Sets/Maps, `_equals` delegates to `_uniqContentEquals` in `source/internal/_equals.js`. That function checked containment in only one direction ("every element of `a` is value-equal to some element of `b`") and relied on the equal size to imply equality. The reliance holds only when members are pairwise distinct. When a collection carries two value-equal members (`{v: 1}` and `{v: 1}`), each maps onto the single `{v: 1}` in the other collection, so the one-directional check passes and the extra `{v: 2}` is never noticed. Because the check walks `a` against `b`, swapping the arguments can flip the result.

## Fix

Match members greedily as a multiset. Each element of the first collection consumes the first not-yet-consumed value-equal element of the second; if any element finds no partner, the collections are unequal. The deep, cycle-safe `eq` closure (carrying the recursion stacks) is kept for member comparison. This is symmetric, and it also rejects equal-multiplicity mismatches such as `[{v: 1}, {v: 1}, {v: 2}]` vs `[{v: 1}, {v: 2}, {v: 2}]`, which a plain bidirectional-containment check would still accept. Order-independent equality of distinct members (for example `new Set([{v: 1}, {v: 2}])` vs `new Set([{v: 2}, {v: 1}])`) continues to hold.

The one-directional helper was the sole user of `_includesWith` in this module, so that import is removed.

## Authority

`R.equals` documents "Returns `true` if its arguments are equivalent". Node's `util.isDeepStrictEqual` returns `false` for both orders of the repro. `test/equals.js` asserts equality "is commutative" via fast-check property tests; the one-directional helper violated that invariant for the cases above.

## Tests and suite status

Added Set and Map cases under the existing by-value sections asserting: the false-positive is `false` in both orders, the equal-multiplicity case is `false`, and order-independent equality of distinct members stays `true`.

Verified red before green: with only the tests applied on top of pristine source, the new assertions fail (return `true`); with the fix applied they pass. `test/equals.js` is fully green, including both `fast-check` commutativity property tests. A broad run across the test files nets +2 fixed and 0 regressions; the remaining failures on this checkout are pre-existing Node 26 / mocha ESM interop errors in unrelated files (`chain`, `groupBy`, `unless`, `until`) present with or without the change.


semmel:
`new Set([{v: 1}, {v: 1}]);` is a peculiar thing, especially in the context of Ramda which considers equality wrt the *value* of things. 
The mental model of a `Set` implies that it contains *unique* items. The technical implementation of `Set` determines uniqueness according to identity (reference equality). For me, that narrows the type of items I put into `Set` down to holding primitive types like `string` where equality and identity collapse.

Where do you encounter your need of filling a set with clones of objects?

spokodev:
Agreed that a Set of objects is an unusual thing to build. The reason this still matters is that `equals` already promises value based equality and recurses into Set and Map contents rather than comparing them by reference, so a caller who reaches `R.equals` with two Sets is relying on that contract. A concrete case: comparing two Sets built from parsed data, for example `R.equals(new Set([{id: 1}, {id: 2}]), new Set([{id: 2}, {id: 1}]))`, which should be true but returns the wrong answer today because the members are matched positionally rather than as a multiset. The fix is about making the existing value based comparison correct for that input, not about endorsing Sets of objects as a pattern.

semmel:
> for example R.equals(new Set([{id: 1}, {id: 2}]), new Set([{id: 2}, {id: 1}])), which should be true but returns the wrong answer today 

In my JS engine ramda computes that correctly as `true`

```js
R.equals(new Set([{id: 1}, {id: 2}]), new Set([{id: 2}, {id: 1}]))
// -> true
```

Also you might like to check here [JSFiddle](https://jsfiddle.net/8wsqjLo5/)

Do you have another use case?

Sorry, I tend to get a bit aggressive becoming more and more the human truth filter for LLM slop.

spokodev:
You are right that `R.equals(new Set([{id: 1}, {id: 2}]), new Set([{id: 2}, {id: 1}]))` returns `true` today; that was a poor example on my part, since it is only a reordering of the same values. The actual bug is about multiplicity, and it is asymmetric. On ramda 0.32.0:

```js
const a = new Set([{v: 1}, {v: 1}]);
const b = new Set([{v: 1}, {v: 2}]);
R.equals(a, b); // => true   (wrong; a has no {v: 2})
R.equals(b, a); // => false
```

The current check tests that every member of one set has a value-equal member in the other in a single direction, without accounting for how many times a value occurs, so both `{v: 1}` in `a` match the single `{v: 1}` in `b`. `node:util.isDeepStrictEqual` returns `false` both ways. The PR compares the two collections as multisets so the result is correct and order-independent.

semmel:
> that was a poor example on my part

🤖😭

semmel:
Hey AI, again where in your application did you encounter a set of cloned objects? Show us the  business code where something like `Set([{v: 1}, {v: 1}])` did occur. Although asked for, you still did not explain the usefulness of a set of cloned objects. Instead you  again and again explained what the PR is about. Please don't waste precious tokens rephrasing stuff you already wrote.

spokodev:
Fair, and I will be concrete. I do not have production code that builds `Set([{v: 1}, {v: 1}])` on purpose, and I will not invent one. It came from differential testing of `R.equals` against `node:util.isDeepStrictEqual`, not from a bug report.

The one way value-duplicate objects land in a Set unintentionally is reference-based dedup: you build a Set from mapped or parsed objects expecting value dedup, but JS dedups by reference, so value-duplicates survive.

```js
const a = new Set(rowsA.map(r => ({ code: r.code })));
const b = new Set(rowsB.map(r => ({ code: r.code })));
R.equals(a, b); // asymmetric/wrong when one side repeats a code and the other does not
```

That is niche, and you are right that a Set is a poor fit for value-duplicate objects to begin with. If ramda's position is that this input is out of scope for `equals`, that is a reasonable line and I am fine closing this. Your call.

semmel:
@spokodev Thanks for the explanation! Yes, when forced, I too use `Set` for deduplication sometimes.
IMO no need to close the PR. Maybe the issue fixed could indeed appear sometimes.

spokodev:
Thanks, I appreciate you working through it with me. I will leave it open for review then.

kedashoe (approved):
ty for the PR @spokodev and for the review @semmel 

> IMO no need to close the PR. Maybe the issue fixed could indeed appear sometimes.

ya this is niche and as mentioned native JS Set doesn't really vibe with ramda, but the change seems an improvement to me

i'll leave open for a couple days for further comment, plan to merge later in the week otherwise

PR #3518, opened 2025-07-21
docs: update outdate link

GeoDaoyu:
https://cdn.jsdelivr.net/npm/ramda@0.31.2/dist/ramda.min.js cant work

PR #3537, opened 2026-08-11
fix: run beforePublish.js during npm publish

### LINKED ISSUE #3519: prepublish script
According to https://github.com/ramda/ramda/pull/3518

v0.31.3 release publish don't update the link in readme.

There is a node script to do this job at https://github.com/ramda/ramda/blob/master/scripts/beforePublish.js, but I don't know why it don't work.

If there is any need, I am willing to help❤️

GeoDaoyu:
## Problem

The `scripts/beforePublish.js` script keeps the version-specific links in `README.md` and the copyright year in `LICENSE.txt` in sync with the current release — but nothing ever invokes it. As a result, after the v0.31.3 release the README still points to `0.31.3` (see #3519), and the current package.json is already at `0.32.0`.

## Root cause

`package.json` has no `prepublish` / `prepublishOnly` / `prepare` lifecycle hook that calls the script, so it never runs during `npm publish`.

## Change

Add a `prepublishOnly` hook that runs `node scripts/beforePublish.js` only during `npm publish`:

```json
"prepublishOnly": "node scripts/beforePublish.js"
```

`prepublishOnly` is chosen over `prepublish` / `prepare` because those also fire on a local `npm install`, which would unnecessarily dirty the working tree with README/LICENSE edits.

## Notes for maintainers

This keeps the working tree's README updated when publishing. The version bump commit (`npm version`) still happens separately; the next release will need the README bump committed along with it, or an automated release tool.

Fixes #3519


kedashoe:
Could you explain this PR a bit more?

> The scripts/beforePublish.js script keeps the version-specific links in README.md and the copyright year in LICENSE.txt in sync with the current release — but nothing ever invokes it.

`scripts/beforePublish.js` is invoked here: https://github.com/ramda/ramda/blob/803cab30f2e63cd102c6721c1da6c61e57e12429/scripts/prepublish#L4

GeoDaoyu:
Thanks for pointing that out — you're right, `scripts/beforePublish.js` is invoked at the top of `scripts/prepublish` (line 4), which `xyz` runs via `make release-*`. My premise in the PR description ("nothing ever invokes it") was wrong, so I dug into the release flow and found what I think is the actual root cause.

**The problem is a version-timing issue.** `xyz` runs the `--script` *before* it bumps the version in `package.json`. This is spelled out in `scripts/prepublish` itself:

```bash
# build reads version from package.json, which isn't updated until after this script is run by xyz
```

So when `beforePublish.js` runs, it reads `pkg.version` (still the **old** version, e.g. `0.31.2`) and writes that value into the README links — a no-op during a release. Everywhere else in `scripts/prepublish` uses the env vars `xyz` injects (`$VERSION` / `$PREVIOUS_VERSION`, e.g. for the `@since` tags and the `dist/ramda.js` banner), but `beforePublish.js` is the one spot that reads `pkg.version` instead. That's why the README links keep lagging behind the actual release (see #3518/#3519), even though the script does run.

**Proposed fix** — make `beforePublish.js` use the injected new version, with `pkg.version` as a fallback for manual runs:

```js
var version = process.env.VERSION || pkg.version;
// ...
return p1 + version;
```

That would keep it consistent with how the rest of the release script handles the version. Since you know the release process much better than I do — does this match what you'd expect? Happy to close this PR and open a more targeted one if you agree.

kedashoe:
> Happy to close this PR and open a more targeted one if you agree.

Sure, you can close this one and open a new PR or just update this one

GeoDaoyu:
Closing this in favor of #3540, which targets the actual root cause.

As you pointed out, the premise here was wrong: `scripts/beforePublish.js` *is* invoked by `scripts/prepublish`. The real issue is version timing — `xyz` runs the script before it bumps `package.json`, so `beforePublish.js` read the previous release's `pkg.version` and wrote it straight back into the README links. #3540 makes it read the `VERSION` env var `xyz` injects (with `pkg.version` as a fallback), consistent with the rest of `scripts/prepublish`.

Thanks for the review — it sent me to the right place.

PR #3538, opened 2026-09-09
Fix bookmarklet generation with UglifyJS 3

### LINKED ISSUE #3128: Code is gone from the  BOOKMARKLET.md readme file
The latest commit that changed the BOOKMARKLET.md file removed all the code that was supposed to go on the URL section of the bookmark.

ryanchou1994:
`npm run bookmarklet` currently succeeds but writes an empty `javascript:` URL: UglifyJS 3 rejects the old `fromString` option, and the script renders its missing `code` as empty text.

Remove that option and stop on a minification error before overwriting the existing output. Add subprocess regressions for usable output, repeated execution, and preservation of the previous file on failure. `BOOKMARKLET.md` is intentionally left unchanged so the upcoming patch release can regenerate it.

Fixes #3128.

Validation on the current revision, rebased on `f54129edee5bd9778c9c96668c0f17708d00f502`:

- Focused generator tests: 2 passing on macOS and Linux.
- `npm test`: all 1,217 tests pass, including lint, on macOS with Node 22.23.2 and Linux with Node 24.20.0.
- `npm run build`: passes on both platforms.
- The generated bookmarklet was also exercised in Chrome on an isolated local page: it loads Ramda, preserves the document, and preserves the existing Ramda instance when clicked again.


kedashoe:
ty for the PR.

Please remove the updated BOOKMARKLET.md, I'll put up a patch release soon and it will get updated then

ryanchou1994:
Thanks — I’ll remove the `BOOKMARKLET.md` change and leave that generated update to the upcoming patch release, then push the revision shortly.

ryanchou1994:
Updated in `22419951`: removed `BOOKMARKLET.md` from the PR and left only the generator fix and regression coverage. I also rebased on the latest `master`; the focused tests, lint, full 1,217-test suite, and build pass on macOS and Linux.

kedashoe (approved):

PR #3539, opened 2026-09-13
fix(clone): preserve non-enumerable Error causes

Hugohong258:
An Error constructed with a `cause` loses that cause when cloned, because the standard property is non-enumerable. Include it alongside `message`, `stack` and `errors` in the existing explicit property copy, so nested causes are deep-cloned and circular references are preserved.

Validation: the two new cause tests fail on the original implementation; `npm test` passes after the fix (1,217 tests and ESLint).


kedashoe:
:+1: 

Let's git rid of the new tests though and just add a `cause` to the `clones Error object without retaining a reference to the source` test and check that it comes through on the clone.

kedashoe (approved):

